PR #27, opened 2022-08-18
Add llvm-project as a submodule.

GleasonK:
Adding llvm-project as a submodule has a few advantages:

1. No need to maintain `llvm_version.txt`, that information is encoded in the commit that comes with the repo now.
2. Helps with GH Actions, the version of llvm-project needed for CI is now a part of the repo checked out in the cloud.
3. Helps with bazel build, now has a constant path from project root to llvm-project. (@sdasgup3)


GleasonK:
Adding @sdasgup3 as a reviewer. He is looking into the bazel build. 

I am curious if there is any issue regarding submodules and internal integration. Perhaps that is a reason to avoid this approach?

joker-eph:
In general git submodule is full of pitfall.... Now if this repo never get branches, release, etc. you can get away from most issues. I think IREE got some experience with it, you may want to ask them.

sdasgup3:
bazel  allows [fetching external dependencies](https://bazel.build/docs/external#fetching-dependencies) and that is what mlir-hlo [currently does](https://github.com/tensorflow/mlir-hlo/blob/master/WORKSPACE#L20). I am trying to have something similar. With that the bazel way of building should not need the submodule dependency. 

May be we can keep the submodule for the standalone builds.

GleasonK:
Got some feedback from @GMNGeoffrey, @joker-eph, @burmako, @sdasgup3. 

Opinions on submodules vary, on one hand submodules are a bit gross, on the other they're standard and built into git. 

If there is no strong use case in favor of submodules for stablehlo, it probably makes sense to keep the current build approach, since it is consistent with tensorflow/mlir-hlo, and makes it trivial to keep build machinery in sync - keep a file with the proper LLVM revision, and users checkout their own LLVM to the proper revision before building, updates to LLVM revision will be kept in-sync as we have some systems for maintaining consistency between parts of stablehlo / mlir-hlo.

This can all be revisited in the future if we come across a better use case for a submodule / community requests.

Issue #15, opened 2022-08-17
Define code style


burmako:
Currently, StableHLO doesn't have a formally-defined code style, but it will be good to have one. We could start with inheriting MLIR-HLO's code style to maximize familiarity for existing CHLO/MHLO users.

bhack:
We will need a code style also for https://github.com/openxla/stablehlo/issues/32

jpienaar:
Note: CHLO/MHLO has mixed code style (which would be good to rectify). I'd say adopting either MLIR's style guide or Google's style guide (C++ side, Python side MLIR doesn't have en explicit one but close to Google's Python one in practice) would allow most familiarity. With the former being more consistent for end to end. (Builders and the like could also follow different style but that ends up being confusing).

burmako:
Further questions to handle when defining code style, via jpienaar's review of a recent PR:
  * Use of doxygen and which constructs or conventions are followed ([comment](https://github.com/openxla/stablehlo/pull/44#pullrequestreview-1085862966)).
  * Error conventions ([comment](https://github.com/openxla/stablehlo/pull/44#pullrequestreview-1085864483)).

burmako:
The last two weeks have been pretty hectic as we've been bootstrapping a GitHub-centric StableHLO development process from the Google-centric MLIR-HLO development process. During that time, we have set up the basic infrastructure (pull requests, issues, CI) and are almost done setting up regular integrates from StableHLO into MLIR-HLO. 

I anticipate that we'll get to defining code style fairly soon (within the next 1-2 weeks).

In the meantime, @GleasonK has set up a CI check with runs `git-clang-format` (#49) and opened a ticket to also run `clang-tidy` likely bootstrapping it from MLIR-HLO (#61). In #76, @joker-eph has explored what it would take to switch the codebase from Google's clang-format style that our codebase currently follows to LLVM's clang-format style.

burmako:
Another part of code style is Markdown formatting. Let's find a formatter / linter, so that we can automate changes like https://github.com/openxla/stablehlo/pull/128. (Also see Kevin's comment in that PR).

sdasgup3:
Regarding formatting markdowns: Adding links to the markdown might break the 80 lines constraint. The script can ignore those.

burmako:
We're getting closer to the point when all the urgent work is done (e.g. our code now ships as part of MLIR-HLO which is great, but we still need an RFC process which is blocking quite a few things), at which point I'm planning to write something up about this.

[...]

PR #37, opened 2022-08-24
Implement pretty printing for SelectOp, ClampOp, AfterAllOp types. Clean up type prints.

GleasonK:
Pretty print for SelectOp based on vector display of `arith.select`: https://mlir.llvm.org/docs/Dialects/ArithmeticOps/#arithselect-mlirarithselectop

```
// Element-wise vector selection in arith:
%vx = arith.select %vcond, %vtrue, %vfalse : vector<42xi1>, vector<42xf32>

// StableHLO display:
%0 = stablehlo.select %arg0, %arg1, %arg1 : tensor<2x3xi1>, tensor<2x3xi32>
```

Note that if branch types do not match exactly, `functional-type` is used:

```
%1 = stablehlo.select %arg0, %arg2, %arg3 : (tensor<2x3xi1>, tensor<2x?xi32>, tensor<?x2xi32>) -> tensor<2x?xi32>
```

Now that `SameOperandsAndResultType` is merged into main, we can use it in other ops that _may_ have the same operands and result types like ClampOp, or ops that _must_ have matching types like AfterAllOp:

```
%0 = stablehlo.clamp %arg0, %arg0, %arg0 : tensor<4xf32>
%5 = stablehlo.after_all %arg1, %arg1 : !stablehlo.token
%6 = stablehlo.after_all : !stablehlo.token
```

Added print tests and invalid MLIR tests.

Additionally this change cleans up some type prints, removing parentheses around operands / including parens around inputs in type printing for consistency with the rest of the dialect's assembly:

```
%0 = stablehlo.cstr_reshapable %arg0, %arg1 : (index, tensor<3xi32>) -> !shape.witness
%0 = stablehlo.compute_reshape_shape %arg0, %arg1 : (index, tensor<2xi32>) -> tensor<2xi32>
%0 = stablehlo.complex %arg0, %arg0 : (tensor<4xf32>, tensor<4xf32>) -> tensor<4xcomplex<f32>>
```

### DIFF
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -24,7 +24,6 @@ limitations under the License.
 #include <array>
 #include <cstdint>
 #include <functional>
-#include <iostream>  // FIXME
 #include <numeric>
 #include <set>
 #include <unordered_map>
@@ -2892,33 +2891,30 @@ ParseResult assignFromFunctionType(OpAsmParser& parser, llvm::SMLoc loc,
   return success();
 }
 
-// getInferredComplexType takes a complex tensor type and returns a
-// type that maintains the shape, but removes the complex type for the
-// underlying data type
+// createRealType takes a tensor type that may have complex elements and
+// returns a type that maintains the shape, but with real numeric data types.
 //   Ex: tensor<4xcomplex<f32>>  -->  tensor<4xf32>
-Type getInferredComplexType(Type result) {
-  assert(result.isa<TensorType>() &&
-         result.cast<TensorType>().getElementType().isa<ComplexType>());
-  TensorType tensorTy = result.cast<TensorType>();
-  ComplexType complexTy = tensorTy.getElementType().cast<ComplexType>();
-  Type elementTy = complexTy.getElementType();
-
-  return hlo::getSameShapeTensorType(tensorTy, elementTy);
+Type createRealType(TensorType type) {
+  auto elementTy = type.getElementType();
+  if (auto complexTy = elementTy.dyn_cast<ComplexType>()) {
+    elementTy = complexTy.getElementType();
+  }
+  return hlo::getSameShapeTensorType(type, elementTy);
 }
 
 // ComplexOpType - only print result type if the inferred complex type
 // matches all operand types.
 //
 // Inferring operand types for complex ops:
-//  %0 = mhlo.complex %1, %2 : tensor<4xcomplex<f32>>
+//  %0 = stablehlo.complex %1, %2 : tensor<4xcomplex<f32>>
 //    %0 : tensor<4xcomplex<f32>>
 //    %1 : tensor<4xf32>
 //    %2 : tensor<4xf32>
 void printComplexOpType(OpAsmPrinter& p, Operation* op, Type lhs, Type rhs,
                         Type result) {
-  Type inferredResult = getInferredComplexType(result);
+  Type realType = createRealType(result.cast<TensorType>());
 
-  if (lhs != inferredResult || rhs != inferredResult) {
+  if (lhs != realType || rhs != realType) {
     p.printFunctionalType(op);
     return;
   }
@@ -2928,7 +2924,6 @@ void printComplexOpType(OpAsmPrinter& p, Operation* op, Type lhs, Type rhs,
 
 ParseResult parseComplexOpType(OpAsmParser& parser, Type& lhs, Type& rhs,
                                Type& result) {
-  // Operand and result types are the same, use copy constructor
   llvm::SMLoc loc = parser.getCurrentLocation();
   Type type;
   if (failed(parser.parseType(type))) {
@@ -2941,14 +2936,14 @@ ParseResult parseComplexOpType(OpAsmParser& parser, Type& lhs, Type& rhs,
   }
 
   // Otherwise, operand type is inferred from complex type
-  if (!type.isa<TensorType>() ||
-      !type.dyn_cast<TensorType>().getElementType().isa<ComplexType>()) {
+  auto tensorType = type.dyn_cast<TensorType>();
+  if (!tensorType || !tensorType.getElementType().isa<ComplexType>()) {
     return parser.emitError(loc, "expected tensor with complex element type");
   }
 
   // Assign LHS and RHS to inferred type
-  Type inferredTy = getInferredComplexType(type);
-  lhs = rhs = inferredTy;
+  Type realType = createRealType(type.cast<TensorType>());
+  lhs = rhs = realType;
   result = type;
   return success();
 }
@@ -2968,16 +2963,6 @@ LogicalResult ComplexOp::inferReturnTypes(
 //===----------------------------------------------------------------------===//
 // ImagOp
 //===----------------------------------------------------------------------===//
-
-namespace {
-Type createRealType(TensorType type) {
-  auto elementTy = type.getElementType();
-  if (auto complexTy = elementTy.dyn_cast<ComplexType>()) {
-    elementTy = complexTy.getElementType();
-  }
-  return hlo::getSameShapeTensorType(type, elementTy);
-}
-}  // namespace
 
 LogicalResult ImagOp::inferReturnTypes(
     MLIRContext*, Optional<Location>, ValueRange operands, DictionaryAttr,
@@ -4328,7 +4313,6 @@ void printSelectOpType(OpAsmPrinter& p, Operation* op, Type pred, Type onTrue,
 
 ParseResult parseSelectOpType(OpAsmParser& parser, Type& pred, Type& onTrue,
                               Type& onFalse, Type& result) {
-  // Operand and result types are the same, use copy constructor
   llvm::SMLoc loc = parser.getCurrentLocation();
   SmallVector<Type> types;
   if (parser.parseTypeList(types)) {
@@ -4352,7 +4336,7 @@ ParseResult parseSelectOpType(OpAsmParser& parser, Type& pred, Type& onTrue,
   }
 
   // stablehlo.select %0, %1 : (<op_types> ...) -> <result_type>
-  auto fnType = types[0].dyn_cast<FunctionType>();
+  auto fnType = types[0].cast<FunctionType>();
   return assignFromFunctionType(parser, loc, {&pred, &onTrue, &onFalse}, result,
                                 fnType);
 }
@@ -5778,9 +5762,9 @@ LogicalResult UniformDequantizeOp::inferReturnTypeComponents(
 //   type($operand2))
 //
 //   Generic:
-//     %0 = "mhlo.operation"(%0, %1) : (tensor<i1>, tensor<i1>) -> tensor<i1>
+//     %0 = "stablehlo.op"(%0, %1) : (tensor<i1>, tensor<i1>) -> tensor<i1>
 //   Custom:
-//     %0 = mhlo.operation(%0, %1) : tensor<i1>
+//     %0 = stablehlo.op(%0, %1) : tensor<i1>
 //
 // Falls back to `printFunctionalType` if all operands do not match result type.
 //
@@ -5877,7 +5861,7 @@ ParseResult parseVariadicSameOperandsAndResultType(
 // TuplesOp - only print result type. Operand type is trivially inferrable.
 //
 // Inferring operand types from tuple type:
-//  %3 = mhlo.tuple %1, %2 : tuple<tensor<i1>, tensor<f32>>
+//  %3 = stablehlo.tuple %1, %2 : tuple<tensor<i1>, tensor<f32>>
 //    %1 : tensor<i1>
 //    %2 : tensor<f32>
 //    %3 : tuple<tensor<i1>, tensor<f32>>
@@ -5908,7 +5892,7 @@ ParseResult parseTupleOpType(OpAsmParser& parser,
 // inferrable.
 //
 // Inferring operand types for pairwise ops:
-//  %3, %4 = mhlo.operation %1, %2 : tensor<i1>, tensor<f32>
+//  %3, %4 = stablehlo.operation %1, %2 : tensor<i1>, tensor<f32>
 //    %1 : tensor<i1>
 //    %2 : tensor<f32>
 //    %3 : tensor<i1>
@@ -5921,7 +5905,6 @@ void printPairwiseOpType(OpAsmPrinter& p, Operation*, TypeRange operands,
 ParseResult parsePairwiseOpType(OpAsmParser& parser,
                                 SmallVectorImpl<Type>& operands,
                                 SmallVectorImpl<Type>& results) {
-  // Operand and result types are the same, use copy constructor
   llvm::SMLoc loc = parser.getCurrentLocation();
   if (parser.parseTypeList(operands)) {
     return parser.emitError(loc, "expected type list");
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -24,6 +24,7 @@ limitations under the License.
 #include <array>
 #include <cstdint>
 #include <functional>
+#include <iostream>  // FIXME
 #include <numeric>
 #include <set>
 #include <unordered_map>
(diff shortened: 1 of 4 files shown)


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -4164,6 +4164,71 @@ LogicalResult RngOp::reifyReturnTypeShapes(
...
+  // Fail if not two types, or single functional type
+  if (types.size() != 2 &&
+      (types.size() != 1 || !types.front().dyn_cast<FunctionType>())) {
+    return parser.emitError(loc,
+                            "expected functional type or list of two types");
+  }
+
+  if (types.size() == 2) {
+    pred = types.front();
+    onTrue = onFalse = result = types.back();
```
burmako, line 4223: Maybe it's just me but I find it easier to read `types[0]` and `types[1]` instead of `types.front()` and `types.back()`. The former is shorter, and the latter additionally doesn't require keeping track of the length of `types`. Similarly, for other parts of this PR.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -4164,6 +4164,71 @@ LogicalResult RngOp::reifyReturnTypeShapes(
...
+  llvm::SMLoc loc = parser.getCurrentLocation();
+  SmallVector<Type> types;
+  if (parser.parseTypeList(types)) {
+    return failure();
+  }
+
+  // Error handling for invalid types
+  // Fail if not two types, or single functional type
+  if (types.size() != 2 &&
+      (types.size() != 1 || !types.front().dyn_cast<FunctionType>())) {
```
burmako, line 4216: `.isa<FunctionType>()` is more idiomatic.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -4164,6 +4164,71 @@ LogicalResult RngOp::reifyReturnTypeShapes(
...
+  if (types.size() == 2) {
+    pred = types.front();
+    onTrue = onFalse = result = types.back();
+    return success();
+  }
+
+  FunctionType fnType = types.front().dyn_cast<FunctionType>();
+  return assignFromFunctionType(parser, loc, {&pred, &onTrue, &onFalse}, result,
+                                fnType);
+}
```
burmako, line 4230: How about restructuring this function so that each of the `types.size() == 1` and `types.size() == 2` cases is handled contiguously? E.g. right now, additional mental effort is required to realize that ending up on line 4227 means that `types.front()` must be `FunctionType`. E.g.

```
if (types.size() == 2) {
  ...
} else if (types.size() == 1 && types[0].isa<FunctionType>()) {
  auto fnType = types[0].cast<FunctionType>();
  ...
} else {
  ...
}
```

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -4164,6 +4164,71 @@ LogicalResult RngOp::reifyReturnTypeShapes(
...
+                            "expected functional type or list of two types");
+  }
+
+  if (types.size() == 2) {
+    pred = types.front();
+    onTrue = onFalse = result = types.back();
+    return success();
+  }
+
+  FunctionType fnType = types.front().dyn_cast<FunctionType>();
```
burmako, line 4227: It is idiomatic to say `auto` when the type is obvious from the context: https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -4164,6 +4164,71 @@ LogicalResult RngOp::reifyReturnTypeShapes(
...
+                                   FunctionType& fnType) {
+  assert(fnType);
+  if (fnType.getInputs().size() != operands.size()) {
+    return parser.emitError(loc)
+           << operands.size() << " operands present, but expected "
+           << fnType.getInputs().size();
+  }
+  if (fnType.getResults().size() != 1) {
+    return parser.emitError(loc, "expected single output");
+  }
```
burmako, line 4181: Perhaps move this check next to `result = fnType.getResults().front()`, so that readers have an easier time making sure that `fnType.getResults()` only has one element?

stablehlo/tests/print_stablehlo.mlir
```diff
@@ -130,21 +130,37 @@ func.func @type_convert_ops(%arg0 : tensor<2xf32>) -> () {
...
+  // CHECK-NEXT: %6 = stablehlo.after_all : !stablehlo.token
+  // CHECK-NEXT: %7 = stablehlo.cstr_reshapable %arg3, %arg2 : (index, tensor<4xi32>) -> !shape.witness
+  // CHECK-NEXT: %8 = stablehlo.compute_reshape_shape %arg3, %arg2 : (index, tensor<4xi32>) -> tensor<4xi32>
   %0 = "stablehlo.clamp"(%arg0, %arg0, %arg0) : (tensor<4xf32>, tensor<4xf32>, tensor<4xf32>) -> tensor<4xf32>
   %1 = "stablehlo.complex"(%arg0, %arg0) {} : (tensor<4xf32>, tensor<4xf32>) -> tensor<4xcomplex<f32>>
   %2 = "stablehlo.compute_reshape_shape"(%arg3, %arg2) : (index, tensor<4xi32>) -> tensor<4xi32>
   %3 = "stablehlo.uniform_quantize"(%arg0) : (tensor<4xf32>) -> tensor<4x!quant.uniform<ui8:f32, 34.0:16>>
   %4 = "stablehlo.uniform_dequantize"(%3) : (tensor<4x!quant.uniform<ui8:f32, 34.0:16>>) -> tensor<4xf32>
   %5 = "stablehlo.after_all"(%arg1, %arg1) : (!stablehlo.token, !stablehlo.token) -> !stablehlo.token
+  %6 = "stablehlo.after_all"() : () -> !stablehlo.token
```
burmako, line 148: Looks like we can remove `// TODO(gleasonk): Unit test these lines once after_all is converted` now?

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -5670,6 +5721,32 @@ ParseResult parseSameOperandsAndResultType(OpAsmParser& parser,
...
+                                            TypeRange opTypes, Type result) {
+  return printSameOperandsAndResultTypeImpl(p, op, opTypes, result);
+}
+
+ParseResult parseVariadicSameOperandsAndResultType(
+    OpAsmParser& parser,
+    SmallVectorImpl<OpAsmParser::UnresolvedOperand>& operands,
+    SmallVectorImpl<Type>& opTypes, Type& result) {
+  // Insert a type for each operand. Need to do this since passing the type of
+  // a variadic op gives no indication of how many operands were provided.
```
burmako, line 5737: Oof this is unfortunate. Let me take a closer look at this during the next pass over this PR. I'll submit my review for now because there's already plenty of comments to address.

burmako (changes requested):
Looks like I clicked "Comment" in my previous review, instead of "Request changes".

stablehlo/dialect/StablehloOps.td
```diff
@@ -1068,14 +1073,17 @@ def StableHLO_AfterAllOp : StableHLO_Op<"after_all", [NoSideEffect]> {
     Example:
 
     ```mlir
-    %0 = stablehlo.after_all %arg0, %arg1 : (!stablehlo.token, !stablehlo.token) -> !stablehlo.token
+    %0 = stablehlo.after_all %arg0, %arg1 : !stablehlo.token
```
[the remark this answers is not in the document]
GleasonK: I guess not. There are very few places in mlir where we can make a "same input and output type" as opposed to just compatible operand and result, however this being a Token in and out, we can guarantee same. I would suggest a trait to make the guarantee explicit, but it looks like `verifySameOperandsAndResultType` requires at least one operand, which isn't guaranteed for after all.

Modifying the PR description.

stablehlo/dialect/StablehloOps.cpp
GleasonK: Only advantage is that front/back allows swapping of the underlying list type to one that doesn't implement indexing, i.e. `SmallVector -> std::list`. Generally preferred for polymorphic functions, but this function is not polymorphic, nor is `SmallVector` likely to change, so I see no harm in preferring the more readable `types[0]`.  

A note on the "doesn't require keeping track of the length of types", it still requires keeping track of the minimum length of size 2 to avoid undefined indexing OOB behavior, right?

Will update.

stablehlo/dialect/StablehloOps.cpp
GleasonK: Would be useful to have a coding standard for things like this. 

In the past I worked with a lint for "no else after return" since it's considered unnecessary, and functions could end with a return in an else statement as opposed to a top-level return.

I also tend to prefer writing validation at the beginning of a function, to act as a sort of precondition for the function. Currently there are very few functions in `StablehloOps.cpp` (I found 2) that end in the error case as opposed to the success case.

GleasonK: I think I'm describing some combination of the below:

https://llvm.org/docs/CodingStandards.html#don-t-use-else-after-a-return

https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code

stablehlo/dialect/StablehloOps.cpp
GleasonK: Makes sense!

stablehlo/tests/print_stablehlo.mlir
GleasonK: good catch.

stablehlo/dialect/StablehloOps.td
```diff
@@ -678,12 +678,17 @@ def StableHLO_ComplexOp: StableHLO_BinaryElementwiseOp<"complex", [NoSideEffect,
   let description = [{
     Performs element-wise conversion of a pair of real and imaginary values to
     a complex value.
+
+    Example:
+
+    ```mlir
+    %0 = stablehlo.complex %arg0, %arg0 : (tensor<4xf32>, tensor<4xf32>) -> tensor<4xcomplex<f32>>
```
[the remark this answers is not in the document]
GleasonK: Agree. I'll start working on that. There are some cases where the types don't match and functional type should be used that will need to be handled.

```
%0 = "mhlo.complex"(%arg0, %arg1) : (tensor<10x10xf32, #CSR>, tensor<10x10xf32, #CSR>) -> tensor<10x10xcomplex<f32>>
```

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -24,6 +24,7 @@ limitations under the License.
 #include <array>
 #include <cstdint>
 #include <functional>
+#include <iostream>  // FIXME
```
burmako, line 27: What does "FIXME" stand for?

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2861,6 +2861,98 @@ LogicalResult ClampOp::reifyReturnTypeShapes(
...
+//   Ex: tensor<4xcomplex<f32>>  -->  tensor<4xf32>
+Type getInferredComplexType(Type result) {
+  assert(result.isa<TensorType>() &&
+         result.cast<TensorType>().getElementType().isa<ComplexType>());
+  TensorType tensorTy = result.cast<TensorType>();
+  ComplexType complexTy = tensorTy.getElementType().cast<ComplexType>();
+  Type elementTy = complexTy.getElementType();
+
+  return hlo::getSameShapeTensorType(tensorTy, elementTy);
+}
```
burmako, line 2905: Consider using `createRealType` which is also defined in this file and does the same thing modulo the assert.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2861,6 +2861,98 @@ LogicalResult ClampOp::reifyReturnTypeShapes(
...
+
+// ComplexOpType - only print result type if the inferred complex type
+// matches all operand types.
+//
+// Inferring operand types for complex ops:
+//  %0 = mhlo.complex %1, %2 : tensor<4xcomplex<f32>>
```
burmako, line 2911: `stablehlo.complex`.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2861,6 +2861,98 @@ LogicalResult ClampOp::reifyReturnTypeShapes(
...
+//    %0 : tensor<4xcomplex<f32>>
+//    %1 : tensor<4xf32>
+//    %2 : tensor<4xf32>
+void printComplexOpType(OpAsmPrinter& p, Operation* op, Type lhs, Type rhs,
+                        Type result) {
+  Type inferredResult = getInferredComplexType(result);
```
burmako, line 2917: This doesn't look like a result, since we're matching it to with lhs and rhs types. Consider renaming to `realType` or `inferredOperandType`. (Same for `Type inferredTy` below).

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2861,6 +2861,98 @@ LogicalResult ClampOp::reifyReturnTypeShapes(
...
+  }
+
+  // Handle if function type, all operand types did not match result type.
+  if (auto fnType = type.dyn_cast<FunctionType>()) {
+    return assignFromFunctionType(parser, loc, {&lhs, &rhs}, result, fnType);
+  }
+
+  // Otherwise, operand type is inferred from complex type
+  if (!type.isa<TensorType>() ||
+      !type.dyn_cast<TensorType>().getElementType().isa<ComplexType>()) {
```
burmako, line 2943: It is not super idiomatic to have `isa` immediately followed by a cast. Consider reformulating e.g.

```
auto tensor_type = type.dyn_cast<TensorType>();
if (!tensor_type || !tensor_type.getElementType().isa<ComplexType>()) {
  ...
}
```

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2861,6 +2861,98 @@ LogicalResult ClampOp::reifyReturnTypeShapes(
...
+    p.printFunctionalType(op);
+    return;
+  }
+
+  p.printType(result);
+}
+
+ParseResult parseComplexOpType(OpAsmParser& parser, Type& lhs, Type& rhs,
+                               Type& result) {
+  // Operand and result types are the same, use copy constructor
```
burmako, line 2929: Not sure what this comment means. Might've been copy/pasted from elsewhere?

stablehlo/dialect/StablehloOps.cpp
burmako: Fair enough! Let's put this on our todo list in #15, and keep the existing style in this PR.

Can you change `types[0].dyn_cast<FunctionType>()` to `types[0].cast<FunctionType>()` though? At this point, we know for sure that it's a `FunctionType`, and `cast` will help emphasize that.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -5595,7 +5733,6 @@ void printSameOperandsAndResultTypeImpl(OpAsmPrinter& p, Operation* op,
                                         TypeRange operands, Type result) {
   // Handle zero operand types `() -> a` prints `a`
   if (operands.empty()) {
-    // TODO(gleasonk): Unit test these lines once after_all is converted, with a
     // call that has no operands and single output.
```
burmako, line 5736: This line also needs to be deleted?

stablehlo/dialect/StablehloOps.cpp
burmako: I've been staring at how this function creates a temporary vector of pointers and realized that reuse is probably overrated in this case. Then I rewrote this function without leveraging `parseSameOperandsAndResultTypeImpl`, and I think I like the result better:

```c++
ParseResult parseVariadicSameOperandsAndResultType(
    OpAsmParser& parser,
    SmallVectorImpl<OpAsmParser::UnresolvedOperand>& operands,
    SmallVectorImpl<Type>& opTypes, Type& result) {
  llvm::SMLoc loc = parser.getCurrentLocation();

  Type type;
  if (parser.parseType(type)) {
    return failure();
  }

  // Handle if function type, all operand types did not match result type.
  if (auto fnType = type.dyn_cast<FunctionType>()) {
    if (fnType.getResults().size() != 1) {
      return parser.emitError(loc, "expected single output");
    }
    llvm::append_range(opTypes, fnType.getInputs());
    result = fnType.getResults()[0];
    return success();
  }

  // Handle bare types. ` : type` indicating all input/output types match.
  opTypes.append(operands.size(), type);
  result = type;
  return success();
}
```

Yes, there is some duplication with `parseSameOperandsAndResultTypeImpl` (more precisely, the parseType and dyn_cast and fnType.getResults().size() != 1 parts), but I think this version of code is easier to understand. What's yout take on this?

stablehlo/dialect/StablehloOps.cpp
GleasonK: Does clang tidy help with this? I can add a comment to #61 to consider if so.

burmako: Hmm I'm not sure. Your comment has already registered in #61, so we're all good :)

stablehlo/dialect/StablehloOps.cpp
GleasonK: fixed

stablehlo/dialect/StablehloOps.cpp
GleasonK: Hmm.. So the _real_ advantage of reuse in printer/parser code in my opinion is the parity:

```
let x = printSameOperandsAndResultTypeImpl(input)
let y = parseSameOperandsAndResultTypeImpl(x)
assert(y == input)
```

I feel like there is extra gravity to code duplication in printer/parser code. Duplicating this code adds the overhead of "if we change anything about the printer code, we need to update the parser code in two places." I'll keep thinking on how to improve this.

stablehlo/dialect/StablehloOps.cpp
GleasonK: Nice! I moved the function definition from the `ImagOp` section to the `ComplexOp` section and kept the code comments.

stablehlo/dialect/StablehloOps.cpp
GleasonK: Replaced all instances: `s/mhlo\./stablehlo.`

stablehlo/dialect/StablehloOps.cpp
GleasonK: True. Going with `realType` since it is accurate and precise.

burmako (approved):

Issue #7, opened 2022-08-17
Create a skeleton of the interpreter


burmako:
Let's opensource the internal skeleton of the StableHLO interpreter. Following up on that, we'll be creating individual tickets for individual ops to keep track of the ongoing work (perhaps, jointly with spec tickets for individual ops).

sdasgup3:
This issue would be preceded by the switch to signed-ness semantics https://github.com/openxla/stablehlo/issues/22.

fortianyou:
Hi @burmako, does it mean that stablehlo will have a runtime? With the interpreter, will XLA or any other compiler be required to run stablehlo module?

burmako:
Hi @fortianyou! The idea is for this interpreter to serve as the reference implementation of StableHLO which accompanies the spec. It would also be nice to use it for constant folding and/or partial evaluation at compile time, but that'll need further design discussion which has started in #44. We don't have other use cases planned for the interpreter at the moment.

> does it mean that stablehlo will have a runtime? With the interpreter, will XLA or any other compiler be required to run stablehlo module?

At the moment, StableHLO is positioned as "an operation set that expresses ML computations" and "a portability layer between ML frameworks and ML compilers". Within that positioning, StableHLO doesn't have an opinion on how it's executed beyond providing specs for individual ops, and it's up to individual consumers to decide how to compile StableHLO, what runtime to use together with that, etc.

At [the OpenXLA community meeting on 8/9](https://github.com/openxla/community/tree/main/meetings/20220809-community-meeting), we talked a bit about "framework optionality" and "hardware optionality" as important design goals for StableHLO. This is a fancy way of saying that we want StableHLO to work very well with a variety of frameworks and hardware, and not having an opinion on specific runtimes or compilers makes this a bit easier.

Issue #22, opened 2022-08-18
StableHLO to adopt signed-ness integers semantics


sdasgup3:
StableHLO, as bootsraped from MHLO, inherits signless integers which was added in MHLO for some legacy reasons. Being a signful dialect, StableHLO should adopt signed-ness semantics instead.

burmako:
The spec draft is now signful, the dialect (forked from MHLO) and the interpreter (synced with the dialect) are signless at the moment.

sdasgup3:
Last week worked mostly on adding specs and designing interpreters. No work related to this issue. I will try to prioritize this by the end of this week.

burmako:
Updating the StableHLO dialect will be a lot of work, especially given the implications for the interop, so it'll take quite a bit of time. Given the near-term focus on completing the spec, unassigning this ticket for now.

PR #44, opened 2022-08-25
Interpreter skeleton

sdasgup3:
Here is the skeleton of interpreter supporting a single `stablehlo.add` op. 
resolve https://github.com/openxla/stablehlo/issues/7


reference/tests/AddOpTest.cpp
```diff
@@ -0,0 +1,290 @@
...
+      %result = stablehlo.add %arg0, %arg1 : tensor<2xcomplex<f64>>
+      func.return %result : tensor<2xcomplex<f64>>
+    }
+  })";
+  runTestCase(kModule, {/*operands*/
+                        {"1.5", "2.5", "7.5", "5.5"},
+                        {"1.5", "2.5", "7.5", "5.5"},
+                        /*expected result*/
+                        {"3.0", "5.0", "15.0", "11.0"}});
+}
```
joker-eph, line 287: Can we find a way to use a more MLIR friendly way of testing? That is a test bench that fits in lit with .mlir files?

joker-eph (dismissed):
Is there a design doc? I don't find anything in the PR right now.

My main focus is how we can use the interpreter mechanism to replace the constant folders.

reference/Element.cpp
```diff
@@ -0,0 +1,163 @@
...
+  return absFloat.compare(err) == APFloat::cmpLessThan;
+}
+
+bool isSupportedUnsignedIntegerType(Type type) {
+  return type.isUnsignedInteger(4) || type.isUnsignedInteger(8) ||
+         type.isUnsignedInteger(16) || type.isUnsignedInteger(32) ||
+         type.isUnsignedInteger(64);
+}
+
+// TODO(https://github.com/openxla/stablehlo/issues/22): Support signed integers
```
burmako, line 47: Also, can you leave a slightly longer comment that explains why we currently map signless integers on signed integers?

reference/Interpreter.cpp
```diff
@@ -0,0 +1,99 @@
...
+#include "Ops.h"
+#include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/DenseMap.h"
+#include "llvm/ADT/STLExtras.h"
+#include "llvm/Support/Errc.h"
+#include "mlir/IR/Block.h"
+#include "mlir/IR/Operation.h"
+#include "mlir/IR/Value.h"
+#include "mlir/Support/DebugStringHelper.h"
+#include "dialect/StablehloOps.h"
```
burmako, line 28: Let's order includes alphabetically. (Here and below, including `Ops.cpp`).

reference/Tensor.cpp
```diff
@@ -0,0 +1,365 @@
...
+
+}  // namespace detail
+
+Tensor::Tensor() {}
+
+Tensor::Tensor(ShapedType type)
+    : impl_(llvm::makeIntrusiveRefCnt<detail::Buffer>(type)) {}
+
+Tensor::Tensor(DenseElementsAttr attr) {
+  // Note: Soon we will be deprecating the use of
```
burmako, line 67: How about providing more context here, e.g.: `TODO(sdasgup3): We're using DenseElementsAttr::getRawData() here for simplicity, because it provides a contiguous representation of underlying data in most cases. However, this doesn't always work (e.g. for splat or for i1), so we'll be migrating to something more reliable in the near future`?


 but its return value is not always the expected byte array, atleast for splat and boolean values

reference/Tensor.cpp
```diff
@@ -0,0 +1,365 @@
...
+  }
+
+  if (elementType.isF64()) {
+    auto elementData = reinterpret_cast<double *>(elementPtr);
+    return Element(elementType,
+                   FloatAttr::get(elementType, APFloat(*elementData)));
+  }
+
+  // Handle signed integer types.
+  // TODO(https://github.com/openxla/stablehlo/issues/22): Support signed
```
burmako, line 108: Also `TODO(#22)`? (And below as well).

reference/Tensor.cpp
```diff
@@ -0,0 +1,365 @@
...
+          return APFloat(floatType.getFloatSemantics(), strNum);
+        }));
+
+    auto complexData = llvm::makeArrayRef(
+        reinterpret_cast<std::complex<APFloat> *>(floatValues.data()),
+        floatValues.size() / 2);
+    return Tensor(DenseElementsAttr::get(type, complexData));
+  }
+
+  if (auto floatType = elemType.dyn_cast<FloatType>()) {
```
burmako, line 343: I think we had a useful comment here internally:

```
  // Unlike parsing Int literals, we are not using parseAttribute for parsing
  // Float literals mainly because it does not seem to parse special float
  // values like nan, +/-inf.
```

reference/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
...
+  MLIRSupport
+)
+
+add_mlir_library(StablehloReferenceTensor
+  PARTIAL_SOURCES_INTENDED
+  Tensor.cpp
+
+  LINK_LIBS PUBLIC
+  StablehloReferenceElement
+  MLIRIR
```
burmako, line 31: Let's order alphabetically. (Here and below).

reference/Tensor.cpp
```diff
@@ -0,0 +1,365 @@
...
+}
+
+Tensor makeTensor(ShapedType type, ArrayRef<StringRef> strData) {
+  auto elemType = type.getElementType();
+
+  if (auto complexTy = elemType.dyn_cast<ComplexType>()) {
+    auto complexElemTy = complexTy.getElementType();
+    auto floatType = complexElemTy.dyn_cast<FloatType>();
+    if (!floatType)
+      llvm_unreachable("Unsupported element type for commplex type");
```
burmako, line 330: Typo: complex.

reference/tests/AddOpTest.cpp
```diff
@@ -0,0 +1,290 @@
...
+                        {"0.0", "-0.0", "1.0", "0.125", "0.1", "3.141", "inf",
+                         "inf", "-inf", "inf"},
+                        {"0.0", "-0.0", "7.0", "0.75", "0.3", "3.141", "0",
+                         "inf", "-inf", "-inf"},
+                        /*expected result*/
+                        {"0.0", "-0.0", "8.0", "0.875", "0.4", "6.282", "inf",
+                         "inf", "-inf", "nan"}});
+}
+
+TEST(AddOpInterpreterTest, F32) {
```
burmako, line 41: Reorder test cases so that F32 comes after BF16, to be consistent with the order everywhere else?

reference/tests/AddOpTest.cpp
```diff
@@ -0,0 +1,290 @@
...
+                        {"0.0", "-0.0", "1.0", "0.125", "0.1",
+                         "3.14159265358979323846", "inf", "inf", "-inf", "inf"},
+                        {"0.0", "-0.0", "7.0", "0.75", "0.3",
+                         "3.14159265358979323846", "0", "inf", "-inf", "-inf"},
+                        /*expected result*/
+                        {"0.0", "-0.0", "8.0", "0.875", "0.4",
+                         "6.283185307179586", "inf", "inf", "-inf", "nan"}});
+}
+
+TEST(AddOpInterpreterTest, SInt4WithOv) {
```
burmako, line 96: Rename to just `SInt4` to simplify names? (Here and below).

reference/tests/AddOpTest.cpp
```diff
@@ -0,0 +1,290 @@
...
+    func.func @main(%arg0: tensor<2x2x2xi4>, %arg1: tensor<2x2x2xi4>) -> tensor<2x2x2xi4> {
+      %result = stablehlo.add %arg0, %arg1 : tensor<2x2x2xi4>
+      func.return %result : tensor<2x2x2xi4>
+    }
+  })";
+  runTestCase(kModule, {/*operands*/
+                        {"0", "1", "2", "-3", "7", "-8", "7", "-8"},
+                        {"-8", "-1", "2", "-3", "1", "-1", "7", "-8"},
+                        /*expected result*/
+                        {"-8", "0", "4", "-6", "-8", "7", "-2", "0"}});
```
burmako, line 108: The spec draft says that overflow is implementation-defined. Should we be testing it in that case?

reference/tests/RunTestUtils.h
```diff
@@ -0,0 +1,39 @@
+/* Copyright 2019 The TensorFlow Authors. All Rights Reserved.
```
burmako, line 1: Rename to just `TestUtils.h` to simplify names?

reference/Element.h
```diff
@@ -0,0 +1,74 @@
...
+#include "mlir/IR/Attributes.h"
+#include "mlir/IR/Types.h"
+
+namespace mlir {
+namespace stablehlo {
+
+class Element {
+ public:
+  /// \name Constructors
+  /// @{
```
jpienaar, line 30: Part of style guide potentially is use of doxygen and which constructs or conventions are followed.

reference/Interpreter.cpp
```diff
@@ -0,0 +1,99 @@
...
+inline llvm::Error invalidArgument(char const *Fmt, const Ts &...Vals) {
+  return createStringError(llvm::errc::invalid_argument, Fmt, Vals...);
+}
+
+}  // namespace
+
+llvm::Expected<SmallVector<Tensor>> eval(func::FuncOp func,
+                                         ArrayRef<Tensor> args) {
+  if (func->getNumRegions() != 1) {
+    return invalidArgument("Expected one region in func %s",
```
jpienaar, line 45: Will StableHLO/interpreter follow MLIR/LLVM's error conventions?

reference/tests/AddOpTest.cpp
sdasgup3: I have the following way in mind. 

```
// RUN: stablehlo-interpreter-runner -split-input-file  %s | FileCheck %s

// CHECK-LABEL:       tensor<i32> {
// CHECK-NEXT:          3 : i32
// CHECK-NEXT:        }
func.func @main() -> tensor<i32> {
  %0 = stablehlo.constant dense<1> : tensor<i32>
  %1 = stablehlo.constant dense<2> : tensor<i32>
  %2 = stablehlo.add %0, %1 : tensor<i32>
  func.return %2 : tensor<i32>
}

// --- 

// More tests

```

 - The tool `stablehlo-interpreter-runner` will consume the IR snippet and print the result tensor(s) to be matched uisng the CHECK directives.
 - The input operand values will be provided using `stable.constant` ops. Implementation wise, that includes defining the `eval` function for  `stablehlo.constant` op.
 
What do you think?

reference/tests/AddOpTest.cpp
sdasgup3: I agree. Also, given that, we have plans to include the tests link from here  in the spec, it does not make sense to include such cases.

sdasgup3:
Hi @joker-eph 
Please find the design doc:
 
https://github.com/openxla/stablehlo/pull/44/files#diff-91eba6136d3c0ea41ebc662b32c2db97d1502acfa7ad16e0765b82398c9ee9d2

reference/tests/AddOpTest.cpp
sdasgup3: Removed the gtests and added  the lit tests instead. 

https://github.com/openxla/stablehlo/pull/44/files#diff-0842b6fc7199e52d0740a7b91341e7659fe4007820194b536a2bb57e22b395b3

build_tools/README.md
```diff
@@ -26,4 +26,7 @@ $ cmake .. -GNinja \
    -DLLVM_ENABLE_ASSERTIONS=On \
    -DMLIR_DIR=${PWD}/../llvm-build/lib/cmake/mlir
 $ ninja check-stablehlo
+
+# To run tests for the reference implementation
+$ ninja && ninja test
```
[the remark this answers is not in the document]
sdasgup3: Replaced the gtests with lit tests. No need for `ninja test`.

reference/Element.h
sdasgup3: Noted. Handled via https://github.com/openxla/stablehlo/issues/15

reference/Interpreter.cpp
sdasgup3: Will be addressed via 
https://github.com/openxla/stablehlo/issues/15

docs/interpreter_design_doc.md
```diff
@@ -0,0 +1,100 @@
+# Interpreter Design
```
burmako, line 1: Rename to `interpreter.md` for conciseness?

docs/interpreter_design_doc.md
```diff
@@ -0,0 +1,100 @@
...
+
+
+## Class Hierarchy
+
+Stablehlo programs are computations over Tensors (n-dimensional arrays), hence
```
burmako, line 6: The canonical spelling that we've been using is "StableHLO".

docs/interpreter_design_doc.md
```diff
@@ -0,0 +1,100 @@
...
+most of the input/output data values for an op are Tensors with the exception of
+constant
+[Attribute](https://docs.google.com/document/d/1GAAK_yHk7q2cZdJ3zBhsiNcrqzfpyjwQ-oWLD7q__Ss/edit#heading=h.5nem09i7p9va)
```
burmako, line 9: I don't think this document is available outside Google.

docs/interpreter_design_doc.md
burmako, line 6: Do we have to capitalize "Tensor" here? It's a common noun, so it probably should be written as "tensor"?

burmako, line 6: I'd qualify this sentence with "at the moment", e.g. "At the moment, StableHLO programs are computations over tensors". To represent programs with dynamic shapes, we'll need to add support for shape computations, and they may not necessarily use tensors.

docs/interpreter_design_doc.md
```diff
@@ -0,0 +1,100 @@
...
+values of primitive types (e.g., Int, float etc.).
```
burmako, line 10: Are these attributes used as values, or they are part of value metadata?

docs/interpreter_design_doc.md
```diff
@@ -0,0 +1,100 @@
+# Interpreter Design
+
+
+## Class Hierarchy
```
burmako, line 4: Perhaps consider "Data model", given that we don't really have class hierarchy?

docs/interpreter_design_doc.md
```diff
@@ -0,0 +1,100 @@
...
+
+Stablehlo programs are computations over Tensors (n-dimensional arrays), hence
+most of the input/output data values for an op are Tensors with the exception of
+constant
+[Attribute](https://docs.google.com/document/d/1GAAK_yHk7q2cZdJ3zBhsiNcrqzfpyjwQ-oWLD7q__Ss/edit#heading=h.5nem09i7p9va)
+values of primitive types (e.g., Int, float etc.).
+
+In the current design, the class `Tensor` implements Tensor data. The underlying
+storage class for `Tensor` objects, `detail:Buffer`, stores the type of the
+Tensor along with a contiguous byte array representing its data.
```
burmako, line 14: "...representing its data, laid out in [major-to-minor order](https://www.tensorflow.org/xla/shapes)".

burmako, line 14: "Buffers are reference-counted to simplify memory management".

docs/interpreter_design_doc.md
```diff
@@ -0,0 +1,100 @@
...
+## Class Hierarchy
+
+Stablehlo programs are computations over Tensors (n-dimensional arrays), hence
+most of the input/output data values for an op are Tensors with the exception of
+constant
+[Attribute](https://docs.google.com/document/d/1GAAK_yHk7q2cZdJ3zBhsiNcrqzfpyjwQ-oWLD7q__Ss/edit#heading=h.5nem09i7p9va)
+values of primitive types (e.g., Int, float etc.).
+
+In the current design, the class `Tensor` implements Tensor data. The underlying
+storage class for `Tensor` objects, `detail:Buffer`, stores the type of the
```
burmako, line 13: "...stores the `mlir::ShapedType` of the tensor".

docs/interpreter_design_doc.md
```diff
@@ -0,0 +1,100 @@
...
+Tensor along with a contiguous byte array representing its data.
+
+Individual elements of a Tensor are represented using `Element` class which uses
+`mlir::Attribute` for storage.
```
burmako, line 17: "Using `mlir::Attribute` simplifies things because this means that we don't have to implement our own machinery for storing values of different types inside `Element`".

docs/interpreter_design_doc.md
```diff
@@ -0,0 +1,100 @@
...
+
+`Tensor`
```
burmako, line 19: Spurious newline?

docs/interpreter_design_doc.md
```diff
@@ -0,0 +1,100 @@
...
+class has the following APIs to interact with it's individual elements:
```
burmako, line 20: Typo: its.

docs/interpreter_design_doc.md
```diff
@@ -0,0 +1,100 @@
...
+  - `Element Tensor::get(int64_t index)`: To extract an individual tensor
+  element at index `index` as `Element` object.
+  - `void Tensor::set(int64_t index, Element element);`: To insert an `Element`
+  object `element` into a Tensor at index `index`.
```
burmako, line 24: "For the skeleton of the interpreter, we chose a linearized API for accessing tensor elements to simplify the implementation. This is sufficient for elementwise ops, and in the future we're planning to expand this to cover more complicated op".

docs/interpreter_design_doc.md
```diff
@@ -0,0 +1,100 @@
...
+```C++
+SmallVector<Tensor> eval(func::FuncOp func, ArrayRef<Tensor> args);
+```
+
+which does the followings:
+1. Maps SSA values with `Tensor` values.
+2. Invokes `eval` on each of the ops within `func`.
+
+The op-level `eval` as  mentioned in (2) is responsible for implementing the
+execution semantics of the op. Following is an example for Stablehlo::AddOp.  In
```
burmako, line 39: \`stablehlo::AddOp\`?

docs/interpreter_design_doc.md
```diff
@@ -0,0 +1,100 @@
...
+
+The entry function to the interpreter is
+
+```C++
+SmallVector<Tensor> eval(func::FuncOp func, ArrayRef<Tensor> args);
+```
+
+which does the followings:
+1. Maps SSA values with `Tensor` values.
+2. Invokes `eval` on each of the ops within `func`.
```
burmako, line 36: "...in their SSACFG order".

docs/interpreter_design_doc.md
```diff
@@ -0,0 +1,100 @@
...
+## Working of the interpreter
+
+The entry function to the interpreter is
+
+```C++
+SmallVector<Tensor> eval(func::FuncOp func, ArrayRef<Tensor> args);
+```
+
+which does the followings:
+1. Maps SSA values with `Tensor` values.
```
burmako, line 35: What does this mean?

docs/interpreter_design_doc.md
```diff
@@ -0,0 +1,100 @@
...
+  for (auto i = 0; i < lhs.getNumElements(); ++i) {
+    result.set(i, lhs.get(i) + rhs.get(i));
+  }
+  return result;
+}
+```
+
+In general, the `Element` class simplifies the execution semantics as specified
+in the `eval` function by encapsulating details about how different types are
+handled.
```
burmako, line 57: "Overall, the design of the interpreter is optimized for readability of implementations of `eval` functions for individual ops` because it's meant to serve as a reference implementation for StableHLO. For example, instead of defining `eval` as a template function and parameterizing it with element types, we encapsulate details about how different element types are handled in `Element::+` etc, simplifying the implementation of `eval`".

docs/interpreter_design_doc.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Interpreter used for constant folding
+We can use the interpreter mechanism to fold operations with constant
+operand/attribute vales.
```
burmako, line 61: Typo: values.

docs/interpreter_design_doc.md
```diff
@@ -0,0 +1,100 @@
...
+## Testing the interpreter
+
+The interpreter takes as inputs (A) a
+StableHLO program, and (B) data values to be fed to the program, and generates
+output data values, which are matched against the user-provided expected data
+values.
+
+In the current implementation, we package the inputs (mlir program + input data
+    values) and outputs in a
+[lit-based]((https://llvm.org/docs/CommandGuide/lit.html)) unit-test as follows:
```
burmako, line 76: I probably wouldn't call these tests "unit tests" given that they test the entire interpreter, rather than its individual pieces". Rephrase as "test"?

docs/interpreter_design_doc.md
```diff
@@ -0,0 +1,100 @@
...
+
+func.func @add_op_test_ui4() -> tensor<2xui4> {
+  %0 = stablehlo.constant dense<[0, 2]> : tensor<2xui4>
+  %1 = stablehlo.constant dense<[15, 3]> : tensor<2xui4>
+  %2 = stablehlo.add %0, %1 : tensor<2xui4>
+  func.return %2 : tensor<2xui4>
+}
+```
+
+A test utility `stablehlo-interpreter-runner`
```
burmako, line 92: Rename to `stablehlo-interpreter` for conciseness?

docs/interpreter_design_doc.md
```diff
@@ -0,0 +1,100 @@
...
+([code](https://github.com/openxla/stablehlo/tree/main/stablehlo/reference/tests/StablehloInterpreterRunner.cpp))
+is responsible for parsing the program and interpret each function and return
```
burmako, line 94: Typo: interpreting and returning.

stablehlo/reference/tests/CMakeLists.txt
```diff
@@ -0,0 +1,98 @@
...
+#include(GoogleTest)
+#
+#foreach(TEST ${TESTS})
+#    add_executable(${TEST} ${TEST}.cpp)
+#    target_link_libraries(${TEST} PRIVATE ${TEST_LIBS})
+#    gtest_discover_tests(${TEST})
+#endforeach()
+#
+#add_custom_target(check-stablehlo-interpreter ${TESTS})
+#add_dependencies(check-stablehlo check-stablehlo-interpreter)
```
burmako, line 62: Remove commented code? Looks like it's no longer used?

stablehlo/reference/tests/CMakeLists.txt
```diff
@@ -0,0 +1,98 @@
...
+        ${CMAKE_CURRENT_BINARY_DIR}/lit.site.cfg.py
+        MAIN_CONFIG
+        ${CMAKE_CURRENT_SOURCE_DIR}/lit.cfg.py
+)
+
+set(STABLEHLO_INTERPRETER_TEST_DEPENDS
+        FileCheck count not
+        stablehlo-interpreter-runner
+)
+add_lit_testsuite(check-stablehlo-interpreter-lit "Running the StableHLO interpreter unit tests"
```
burmako, line 75: "interpreter tests"?

stablehlo/reference/tests/StablehloInterpreterRunner.cpp
```diff
@@ -0,0 +1,122 @@
+/* Copyright 2022 The StableHLO Authors.
```
burmako, line 1: Let's move this file to `stablehlo/tools`, next to `stablehlo-opt`, and rename to `StablehloInterpreterMain.cpp` for consistency?

stablehlo/reference/tests/add.mlir
```diff
@@ -0,0 +1,270 @@
+// RUN: stablehlo-interpreter-runner -split-input-file %s | FileCheck %s
```
burmako, line 1: Let's move these tests under `tests/`? This way we won't need to have another `lit.cfg.py` and `lit.site.cfg.py.in`.

stablehlo/reference/tests/add.mlir
```diff
@@ -0,0 +1,270 @@
...
+// CHECK-NEXT:   8.750000e-01 : f16
+// CHECK-NEXT:   3.999020e-01 : f16
+// CHECK-NEXT:   6.281250e+00 : f16
+// CHECK-NEXT:   0x7C00 : f16
+// CHECK-NEXT:   0x7C00 : f16
+// CHECK-NEXT:   0xFC00 : f16
+// CHECK-NEXT:   0x7C00 : f16
+
+func.func @add_op_test_f16() -> tensor<2x5xf16> {
+  %0 = stablehlo.constant dense<[[0.0, -0.0, 1.0, 0.125, 0.1], [3.141, 0x7C00, 0x7C00, 0xFC00, 0x7C00]]> : tensor<2x5xf16>
```
burmako, line 17: How about turning these constants into 1D tensors since this is an elementwise op anyway? This will simplify reading and maintaining these tests.

stablehlo/reference/tests/add.mlir
```diff
@@ -0,0 +1,270 @@
...
+// CHECK-NEXT:   0.000000e+00 : f16
+// CHECK-NEXT:   -0.000000e+00 : f16
+// CHECK-NEXT:   8.000000e+00 : f16
+// CHECK-NEXT:   8.750000e-01 : f16
+// CHECK-NEXT:   3.999020e-01 : f16
+// CHECK-NEXT:   6.281250e+00 : f16
+// CHECK-NEXT:   0x7C00 : f16
+// CHECK-NEXT:   0x7C00 : f16
+// CHECK-NEXT:   0xFC00 : f16
+// CHECK-NEXT:   0x7C00 : f16
```
burmako, line 14: What do you think about moving expected results below `stablehlo.constant`s? I.e. under `func.return %2 : tensor<2x5xf16>` etc. I think this will help with readability, because this way the reader will see inputs first and outputs last. E.g.

```
// CHECK-LABEL: add_op_test_f16
func.func @add_op_test_f16() -> tensor<2x5xf16> {
  %0 = stablehlo.constant dense<[[0.0, -0.0, 1.0, 0.125, 0.1], [3.141, 0x7C00, 0x7C00, 0xFC00, 0x7C00]]> : tensor<2x5xf16>
  %1 = stablehlo.constant dense<[[0.0, -0.0, 7.0, 0.75, 0.3], [3.141, 0.0, 0x7C00, 0xFC00, 0x7C00]]> : tensor<2x5xf16>
  %2 = stablehlo.add %0, %1 : tensor<2x5xf16>
  func.return %2 : tensor<2x5xf16>
  // CHECK-NEXT:  tensor<2x5xf16>
  // CHECK-NEXT:   0.000000e+00 : f16
  // CHECK-NEXT:   -0.000000e+00 : f16
  // CHECK-NEXT:   8.000000e+00 : f16
  // CHECK-NEXT:   8.750000e-01 : f16
  // CHECK-NEXT:   3.999020e-01 : f16
  // CHECK-NEXT:   6.281250e+00 : f16
  // CHECK-NEXT:   0x7C00 : f16
  // CHECK-NEXT:   0x7C00 : f16
  // CHECK-NEXT:   0xFC00 : f16
  // CHECK-NEXT:   0x7C00 : f16
}
```

stablehlo/reference/tests/StablehloInterpreterRunner.cpp
```diff
@@ -0,0 +1,122 @@
...
+    return failure();
+
+  // Keep the output file if the invocation of InterpreterMain was successful.
+  output->keep();
+  return success();
+}
+
+}  // namespace
+
+int main(int argc, char **argv) { return failed(InterpreterMain(argc, argv)); }
```
burmako, line 122: Can we use https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Tools/mlir-translate/MlirTranslateMain.h for this? It handles a bunch of boilerplate for us.

stablehlo/reference/tests/StablehloInterpreterRunner.cpp
```diff
@@ -0,0 +1,122 @@
...
+  ParserConfig config(&context);
+  OwningOpRef<ModuleOp> module(parseSourceFile<ModuleOp>(sourceMgr, config));
+  if (!module) return failure();
+
+  for (auto funcOp : module->getBodyRegion().getOps<mlir::func::FuncOp>()) {
+    os << "\nEvaluated results of function: " << funcOp.getSymName() << "\n";
+
+    // Run the test model.
+    auto results = mlir::stablehlo::eval(funcOp, {});
+    if (!(bool)results) toString(results.takeError());
```
burmako, line 69: `if (failed(results))`

burmako, line 69: You probably wanted to print the result of `toString`?

stablehlo/reference/tests/add.mlir
```diff
@@ -0,0 +1,270 @@
...
+
+func.func @add_op_test_f16() -> tensor<2x5xf16> {
+  %0 = stablehlo.constant dense<[[0.0, -0.0, 1.0, 0.125, 0.1], [3.141, 0x7C00, 0x7C00, 0xFC00, 0x7C00]]> : tensor<2x5xf16>
+  %1 = stablehlo.constant dense<[[0.0, -0.0, 7.0, 0.75, 0.3], [3.141, 0.0, 0x7C00, 0xFC00, 0x7C00]]> : tensor<2x5xf16>
+  %2 = stablehlo.add %0, %1 : tensor<2x5xf16>
+  func.return %2 : tensor<2x5xf16>
+}
+
+// -----
```
burmako, line 23: Given that the tool handles multiple functions at once, do you need the split markers?

burmako (changes requested):

joker-eph (commented):
(My review crossed wires with Eugene's, so I'll leave it there and resume after the next update)

stablehlo/reference/tests/add.mlir
sdasgup3: Will switch to[ third_party/tensorflow/compiler/mlir/lite/flatbuffer_translate.cc](https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Tools/mlir-translate/MlirTranslateMain.h). 

Still we support both of split files (comes for free with the MlirTranslate interface) and specifying multiple functions in a module. 

Handling multiple modules is just a convenience. Avoiding it needs us to add constraints like 
  - the module should have only one function, Or 
  - in case of multiple functions, only the first one will be interpreted Or  one with a specific name will be evaluated. 

From user perspective, it means that they do not have to any of the above constraints.

docs/interpreter_design_doc.md
sdasgup3: done.

stablehlo/reference/tests/StablehloInterpreterRunner.cpp
sdasgup3: done with 
```
if(!result) ...
```

docs/interpreter_design_doc.md
sdasgup3: >> with the exception of constant Attributes
yes, the  tributes here are not used as values, but they are the metadata related to operand values.  I think this statement unnecessarily complicates the discussion. Will be revised.

docs/interpreter_design_doc.md
sdasgup3: will revise this.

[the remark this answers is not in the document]
sdasgup3: revised. Please have a look.

docs/interpreter_design_doc.md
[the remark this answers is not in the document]
sdasgup3: >> Tensor has a "contiguous byte array representing its data"

This is how a `Tensor` object internally stores its data.

>> elements are using mlir::Attribute storage?

Class `Element ` is how an element of a tensor is represented.  We have an API (`Tensor::get`) to extract an element of a tensor ( an `Element` object) from `Tensor` object.

stablehlo/tests/interpret_add.mlir
```diff
@@ -0,0 +1,254 @@
...
+// CHECK-NEXT:   0.000000e+00 : f32
+// CHECK-NEXT:   -0.000000e+00 : f32
+// CHECK-NEXT:   8.000000e+00 : f32
+// CHECK-NEXT:   8.750000e-01 : f32
+// CHECK-NEXT:   4.000000e-01 : f32
+// CHECK-NEXT:   6.28318548 : f32
+// CHECK-NEXT:   0x7F800000 : f32
+// CHECK-NEXT:   0x7F800000 : f32
+// CHECK-NEXT:   0xFF800000 : f32
+// CHECK-NEXT:   0x7FC00000 : f32
```
burmako, line 56: Looks like leftover formatting from the previous version of the PR?

stablehlo/tests/interpret_add.mlir
```diff
@@ -0,0 +1,254 @@
...
+  // CHECK-NEXT:   0.000000e+00 : f16
+  // CHECK-NEXT:   -0.000000e+00 : f16
+  // CHECK-NEXT:   8.000000e+00 : f16
+  // CHECK-NEXT:   8.750000e-01 : f16
+  // CHECK-NEXT:   3.999020e-01 : f16
+  // CHECK-NEXT:   6.281250e+00 : f16
+  // CHECK-NEXT:   0x7C00 : f16
+  // CHECK-NEXT:   0x7C00 : f16
+  // CHECK-NEXT:   0xFC00 : f16
+  // CHECK-NEXT:   0x7C00 : f16
```
burmako, line 19: Is there a convention for the amount of whitespace between CHECK-NEXT and subsequent content? I see that sometimes it's 2 spaces, sometimes 3, sometimes 4. How about just 1 whitespace everywhere, e.g. `CHECK-NEXT: tensor<10xf16>` and `CHECK-NEXT: 0.000000e+00 : f16`.

stablehlo/tests/interpret_add.mlir
```diff
@@ -0,0 +1,254 @@
+// RUN: stablehlo-interpreter  --interpret -split-input-file %s | FileCheck %s
```
burmako, line 1: Typo: double whitespace.

stablehlo/tools/StablehloInterpreterMain.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include "mlir/Tools/mlir-translate/Translation.h"
+#include "stablehlo/dialect/StablehloOps.h"
+#include "stablehlo/reference/Interpreter.h"
+
+namespace mlir {
+
+TranslateFromMLIRRegistration stablehlo_interpreter(
+    "interpret",
+    [](ModuleOp module, raw_ostream &os) {
+      for (auto funcOp : module.getBodyRegion().getOps<mlir::func::FuncOp>()) {
```
burmako, line 26: How about:

```
module.walk([&](FuncOp funcOp) {
  ...
}
```

stablehlo/tools/StablehloInterpreterMain.cpp
```diff
@@ -0,0 +1,53 @@
+/* Copyright 2022 The StableHLO Authors.
```
burmako, line 1: Very nice simplification!

docs/interpreter.md
```diff
@@ -0,0 +1,128 @@
...
+  return DenseElementsAttr::get(result.getType(), values);
+}
+```
+
+## Testing the interpreter
+
+The interpreter takes as inputs (A) a
+StableHLO program, and (B) data values to be fed to the program, and generates
+output data values, which are matched against the user-provided expected data
+values.
```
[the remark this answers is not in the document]
sdasgup3: Will fix 97.
For 103, the formatter tried to keep the markdown hyperlink in the same line and hence makes the other line look off-centered.

docs/interpreter.md
```diff
@@ -0,0 +1,128 @@
...
+```
+
+A test utility `stablehlo-interpreter`
+([code](https://github.com/openxla/stablehlo/tree/main/stablehlo/reference/tests/StablehloInterpreterMain.cpp))
+is responsible for parsing the program, interpreting each function, and
+returning the resulting tensor(s) to be matched against the output tensor
+provided in [FileCheck
+directives](https://llvm.org/docs/CommandGuide/FileCheck.html). We have a
+dedicated test-suite, consisting of several unit-tests, for each StableHLO Op.
+The unit-suites can be found
```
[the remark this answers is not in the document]
sdasgup3: My bad! I must have overlooked this. Will fix.

stablehlo/tools/StablehloInterpreterMain.cpp
```diff
@@ -23,22 +24,23 @@ namespace mlir {
...
         }
 
         // Dump the results.
         for (auto &result : *results) result.print(os);
-      }
+        return WalkResult::advance();
+      });
 
-      return success();
+      return walkResult.wasInterrupted() ? failure() : success();
```
burmako, line 43: Can be simplified to `return success(!walkResult.wasInterrupted())`.

docs/interpreter_design_doc.md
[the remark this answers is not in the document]
joker-eph: Why would we use an Attribute for this though? This seems incredibly expensive to me?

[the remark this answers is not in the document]
sdasgup3: We use `mlir::Attribute` to simplify the implementation in a sense that we can avoid implementing our own mechanism for storing values of different types inside `Element`.
I am open to suggestions for improvement. Maybe we can land this as is and improve this in a follow up PR.

stablehlo/reference/Element.cpp
```diff
@@ -0,0 +1,164 @@
...
+        type,
+        ArrayAttr::get(
+            type.getContext(),
+            {FloatAttr::get(complexElemTy,
+                            leftComplexValue.real() + rightComplexValue.real()),
+             FloatAttr::get(complexElemTy, leftComplexValue.imag() +
+                                               rightComplexValue.imag())}));
+  }
+
+  llvm_unreachable("Unsupported element type");
```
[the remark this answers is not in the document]
sdasgup3: HI @joker-eph 

Our plan  was to have such  `llvm_unreachable` programmatic error  part of all the future operations over `Element` objects. IMO, having `llvm_unreachable`  simplifies the control flow as the error does not needs to be propagated up the stack.  With Expected<Element>,  we will loose the  following better readability while operating on the `Element` objects

Comparing

```
Tensor eval(AddOp op, const Tensor &lhs, const Tensor &rhs) {
  ...
  for (...) {
    **result.set(i, lhs.get(i) + rhs.get(i));**
  }
  return result;
}
```

Vs

```
Expected<Tensor> eval(AddOp op, const Tensor &lhs, const Tensor &rhs) {
  ...
  for (...) {
    if(auto lhsOrErr = lhs.get(i)) return ...some Error code .. 
    if(auto rhsOrErr = rhs.get(i)) return ...some Error code .. 
    result.set(i, *lhsOrErr + *rhsOrErr);
}
```

Let me know your opinion otherwise.

docs/interpreter_design_doc.md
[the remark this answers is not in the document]
joker-eph: Just a discriminated union over all the element types would do?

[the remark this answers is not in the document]
sdasgup3: Thanks! That sounds good! 
How about making the suggested change as a separate PR and track that via an issue?

burmako (approved):

stablehlo/reference/Element.cpp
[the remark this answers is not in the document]
joker-eph: I meant that instead of llvm_unreachable we could use llvm::report_fatal_error ; the only difference is that the later is effective without assertions enabled.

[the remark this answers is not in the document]
sdasgup3: Thanks for suggested! Addressed it in the current file and also in files `Tensor.cpp` and `Interpreter.cpp`

PR #50, opened 2022-08-26
Add type inference for stablehlo.reduce

zhouxin913:
The new `ReduceOp::inferReturnTypeComponents()` replaces the old `ReduceOp::verify()` with all the checking logic.

Note that a unit test is removed: it had a `llvm.add` inside its region, and intended to return a scale from its region and catch the error in `ReduceOp::verify()`. Now when we replace verifier with infer function, there will be two errors:
1. `'llvm.add' op operand #0 must be integer or LLVM dialect-compatible vector of integer, but got 'f32'`: can be fixed by change `f32` -> `i32`
2. In the last line of region, stablehlo.return supports return tensor only, instead of scalar: No way to fix this.

The HEAD code tolerated these errors because the `ReduceOp::verify()` fails first. Now decide to remove this unit test which is a case that will not happen.

### DIFF
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -458,7 +458,7 @@ LogicalResult verifyReducerShape(
   //  BV(j) : j-th init-value of reducer-function
   //  R(i)  : i-th return-type
   //
-  //  Note that: |I(i)| == V(j)| == |BI(i)| == |BV(j)| == |R(i)|
+  //  Note that: |I(i)| == |V(j)| == |BI(i)| == |BV(j)| == |R(i)|
   //
   //  Here are the type-constraints among V(j), BI(i), BV(j), and R(i).
   //    C1 : Check that BI(i) and R(i) have same shape and element-type.
@@ -537,7 +537,8 @@ LogicalResult verifyReducerShape(
          outputShapeIdx < static_cast<int64_t>(allowedDimensions.size()) &&
          argShapeIdx < static_cast<int64_t>(argShape.size());
          outputShapeIdx++)
-      if (allowedDimensions[outputShapeIdx] == argShape[argShapeIdx])
+      if (allowedDimensions[outputShapeIdx] == argShape[argShapeIdx] ||
+          argShape[argShapeIdx] == ShapedType::kDynamicSize)
         argShapeIdx++;
 
     if (argShapeIdx != static_cast<int64_t>(argShape.size()))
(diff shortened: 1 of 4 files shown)


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -3952,17 +3953,19 @@ ParseResult ReduceOp::parse(OpAsmParser& parser, OperationState& result) {
...
-  int64_t numInputs = getNumOperands() / 2;
+  if (operands.size() % 2 != 0 || operands.size() == 0)
+    return emitOptionalError(location, "expects the size of operands to be even and >= 2");
+  int64_t numInputs = operands.size() / 2;
   auto operandTensorTypes = llvm::to_vector<4>(llvm::map_range(
-      getOperandTypes(),
-      [](Type t) -> TensorType { return t.cast<TensorType>(); }));
+      operands.getTypes(),
+      [](Type t) -> TensorType
+      { return t.cast<TensorType>(); }));
```
burmako, line 3968: Probably shouldn't change formatting here? In the near future when #49 lands, we'll have `.clang-format` + we'll have CI complaining if formatting isn't compliant with `clang-format`. In the meanwhile, this part can be fixed manually.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -4024,46 +4029,19 @@ LogicalResult ReduceOp::verify() {
     }
   }
 
-  Block& block = body().front();
+  Block& block = adaptor.body().front();
   SmallVector<TensorType> accumulatorSubShapes;
-  if (failed(verifyReducerShape(this->getLoc(), block, inputArgTypes,
+  if (failed(verifyReducerShape(location.getValue(), block, inputArgTypes,
```
burmako, line 4034: This is not going to work well if location is empty. You'll need to change `verifyReducerShape` to take an optional location.

stablehlo/dialect/StablehloOps.td
```diff
@@ -1259,6 +1259,7 @@ def StableHLO_AllToAllOp : StableHLO_Op<"all_to_all",
 def StableHLO_ReduceOp: StableHLO_ShapedInterfaceOp<"reduce", [
       RecursiveSideEffects,
       SameVariadicOperandSize,
+      InferTensorTypeWithReify,
```
burmako, line 1262: Did you mean `InferTensorType`?

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -4024,46 +4029,19 @@ LogicalResult ReduceOp::verify() {
...
-        failed(verifyCompatibleShape(newDimensions, opResultType.getShape()))) {
-      Type expectedResultType = RankedTensorType::get(
-          newDimensions, accumulatorSubShapes[shapeIdx].getElementType());
-      return emitError()
-             << "Unexpected type for reduce-op's return value at index "
-             << shapeIdx << ": " << opResultType << " vs " << expectedResultType
-             << " (expected)";
-    }
+  
+  for (auto resultType : accumulatorSubShapes)
```
burmako, line 4039: Rename to `accumulatorResultType` to disambiguate from the op's result type?

burmako (changes requested):

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Sure, I tried this in my `settings.json` MS Code:
```
"clang-format.style": "{BasedOnStyle: google}",
// "C_Cpp.formatting": "Disabled"
// For vscode >= 1.33, try this instead of "C_Cpp.formatting": "Disabled"
"[cpp]": {
    "editor.defaultFormatter": "xaver.clang-format"
},
```
After auto-format, the code looks exactly you suggested.

stablehlo/dialect/StablehloOps.cpp
zhouxin913: This is an issue indeed. Change `verifyReducerShape` to accept `Optional<Location>`

stablehlo/dialect/StablehloOps.td
zhouxin913: There exists a `ReduceOp::reifyReturnTypeShapes()`. So `InferTensorTypeWithReify` seems more proper here as it is fit for op which has both: "inferReturnTypeComponents" and "reifyReturnTypeShapes".

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Yes.

Issue #53, opened 2022-08-29
Add the specification of bitwise operators


sdasgup3:
### Request description

This is first in the upcoming series of tickets documenting the specification of various StableHLO ops.

PR #54, opened 2022-08-29
Add specifications for and, or, xor, not operators

sdasgup3:
fixes https://github.com/openxla/stablehlo/issues/53

Has the following changes:
  - A few spec types:
       -  Fix the header of the table names from "Operand Names" to "Name" and "Result Name" to "Name". The redundant prefixes can be derived from the section name.
       - not consider tensor as keyword. 
  - Addition of the spec of the Bitwise ops.

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -9,7 +9,9 @@ Following are the supported element types in StableHLO:
     document as `si<N>`, where the bit-width N ∊ {4, 8, 16, 32, 64}
     * Unsigned integer referred to in the document as `ui<N>`, where the
     bit-width N ∊ {4, 8, 16, 32, 64}
-  * **Boolean types** referred to in the document as `pred`
+  * **Boolean types** referred to in the document as `pred`. Exact
+  representation of boolean types (e.g. 1 byte per boolean vs 1 bit per boolean)
+  is implementation-defined.
   * **Floating-point types**
     * Single precision `f32`, double precision `f64` and half precision `f16`
     floating-points complying with [IEEE 754
@@ -88,7 +90,7 @@ The specification of an op comprises of the following components (in the order
 
 Performs element-wise addition of two tensors `lhs` and `rhs` and produces a
 `result` tensor. For integer element types, if the element-wise sum has an
-unsigned/signed overflow/underflow, the result is implementation defined and one
+unsigned/signed overflow/underflow, the result is implementation-defined and one
 of the followings:
 
   * mathematical result modulo $2^n$, where n is the bit width of the result.
@@ -168,7 +170,7 @@ logical operation.
 // Logical operation with with boolean tensors
   // %x: [[false, false], [true, true]]
   // %y: [[false, true], [false, true]]
-  %z = stablehlo.or %x, %y : tensor<2x2xpred>
+  %z = stablehlo.and %x, %y : tensor<2x2xpred>
   // %z: [[false, false], [false, true]]
 ```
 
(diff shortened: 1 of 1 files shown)


docs/spec_draft.md
```diff
@@ -127,3 +127,183 @@ specification.
...
 // %z: [[6, 8], [10, 12]]
 ```
+
+## stablehlo.and
+
+`stablehlo.and(lhs, rhs) -> result`
+
+### Semantics
+
+An element-wise operation performing bitwise AND of two tensors `lhs` and `rhs`
```
burmako, line 137: "Performs element-wise..." to be consistent with `add`, `maximum` and `minimum`?

docs/spec_draft.md
```diff
@@ -127,3 +127,183 @@ specification.
...
+of integer types and produces a `result` tensor. For Boolean tensors, it
+computes the logical operation.
```
burmako, line 139: This raises the question of when bitwise AND would be different from logical AND, which raises the question of what's the representation of `pred`. We're talking about representations for integer and floating types, but not booleans.

docs/spec_draft.md
```diff
@@ -127,3 +127,183 @@ specification.
...
+
+### Results
+
+| Name | Type |
+|-|-|
+| `result` | tensor of Integer or Boolean element types |
+
+### Constraints
+
+  * Supported shapes: all static shapes.
```
burmako, line 202: Missing constraint.

burmako (changes requested):

docs/spec_draft.md
sdasgup3: Can you please point which one?

docs/spec_draft.md
sdasgup3: >> what's the representation of `pred`
In the specification, we have representations for ints and floats which just happened to match the actual llvm types we use to implement them. (Side note: The intention is not to give away that implementation info in the spec.) 
Similarly,  we have the Boolean types represented as `pred`. 

Let me know if I am missing anything.

docs/spec_draft.md
sdasgup3: done.

sdasgup3:
Note that https://github.com/openxla/stablehlo/pull/56 and https://github.com/openxla/stablehlo/pull/54 are not diff-based on each other. Merging one will have merge conflicts on the other, which I will fix.

docs/spec_draft.md
burmako: Sorry! Same as above: "lhs, rhs have the same type".

docs/spec_draft.md
burmako: "The intention is not to give away that implementation info in the spec". I mean, the spec is fairly precise for the representation of integer and floating-point types, referring to "signed integer with two’s complement representation", "IEEE 754 format", "Brain Floating-Point Format".

Perhaps we should provide more details about booleans as well. E.g. "Boolean types referred to in the document as `pred`. Exact representation of boolean types (e.g. 1 byte per boolean vs 1 bit per boolean) is implementation-define"?

docs/spec_draft.md
sdasgup3: This being a unary op we can skip this constraint.

docs/spec_draft.md
sdasgup3: I got your point. Added the detail.

burmako (approved):

Issue #55, opened 2022-08-29
Add the specification of minimum/maximum operators


sdasgup3:
### Request description

Add the specification of `stablehlo.minimum` and `stablehlo.maximum`.

PR #56, opened 2022-08-29
Add specifications for maximum and minimum operators

sdasgup3:
fixes https://github.com/openxla/stablehlo/issues/55

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -32,7 +32,7 @@ array, represented in the opset as `tensor<SxE>` such that
 
 ## Programs
 
-StableHLO programs consist of functions. Each function has arguments and results
+StableHLO programs consist of functions. Each function has operands and results
 of supported types and a list of ops in static single-assignment (SSA) form
 which is terminated by a return op which produces the results of the function.
 StableHLO ops take operands and produce results.
@@ -100,18 +100,18 @@ of the followings:
 For floating-point element types, corner cases are defined by the IEEE-754
 specification.
 
-### Arguments
+### Operands
 
-| Operand Name(s) | Type |
+| Name | Type |
 |-|-|
-| `lhs` | `tensor` of Integer, Floating-point and Complex element types |
-| `rhs` | `tensor` of Integer, Floating-point and Complex element types |
+| `lhs` | tensor of integer, floating-point, or complex element types |
+| `rhs` | tensor of integer, floating-point, or complex element types |
 
 ### Results
 
-| Result Name | Type |
+| Name | Type |
 |-|-|
-| `result` | `tensor` of Integer, Floating-point, and Complex element types |
+| `result` | tensor of integer, floating-point, or complex element types |
 
 ### Constraints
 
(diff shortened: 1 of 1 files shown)


docs/spec_draft.md
```diff
@@ -127,3 +127,79 @@ specification.
...
+Performs element-wise max operation on tensors `lhs` and `rhs` and produces a
+`result` tensor. For Floating-point element type, implements IEEE 754 semantics:
+Returns the larger of two arguments, propagating `NaN`s and treating `-0` as
+less than `+0`.
+
+### Arguments
+
+| Name | Type |
+|-|-|
+| `lhs` | tensor of Integer, Floating-point element types |
```
burmako, line 146: After reading the spec draft some more, I feel ambivalent about capitalization of "Integer types". Are these names special enough to be capitalized? For comparison, https://llvm.org/docs/LangRef.html doesn't capitalize them.

docs/spec_draft.md
```diff
@@ -127,3 +127,79 @@ specification.
...
+`stablehlo.maximum(lhs, rhs) -> result`
+
+### Semantics
+
+Performs element-wise max operation on tensors `lhs` and `rhs` and produces a
+`result` tensor. For Floating-point element type, implements IEEE 754 semantics:
+Returns the larger of two arguments, propagating `NaN`s and treating `-0` as
+less than `+0`.
+
+### Arguments
```
burmako, line 142: I've just noticed that the "Structure of an Op’s Specification" section says "Operands". I'd say that "Operands" is more idiomatic since we're using MLIR as the primary implementation vehicle.

burmako (changes requested):

docs/spec_draft.md
sdasgup3: Agreed. Done.

docs/spec_draft.md
sdasgup3: done.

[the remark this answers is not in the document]
sdasgup3: Added the semantics with lexicographic comparison.

sdasgup3:
Note that https://github.com/openxla/stablehlo/pull/56 and https://github.com/openxla/stablehlo/pull/54 are not diff-based on each other. Merging one will have merge conflicts on the other, which I will fix.

docs/spec_draft.md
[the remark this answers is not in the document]
burmako: Is this intended behavior though? I went ahead and checked what XLA does by using `run_hlo_module` on a program like this one:

```
HloModule max
ENTRY %max (x: c64[], y: c64[]) -> c64[] {
  %x = c64[] parameter(0)
  %y = c64[] parameter(1)
  ROOT %add = c64[] maximum(c64[] %x, c64[] %y)
}
```

Unfortunately, I didn't get a conclusive answer. `--platform=Interpreter` says "Unsupported type for maximum: C64" and `--platform=CPU` says `UNIMPLEMENTED: binary complex op 'maximum'`: https://github.com/tensorflow/tensorflow/blob/f0fb944fd60c12c0c0c9a96d1e8e4e802cad94c1/tensorflow/compiler/xla/service/elemental_ir_emitter.cc#L1352-L1431.

I think more investigation is needed.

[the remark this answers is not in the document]
sdasgup3: Thanks for the experiment! 
Actually we did the same for XLA and reached the similar conclusion. The current behaviour proposed here matches with [Jax](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.max.html).

burmako (approved):

PR #59, opened 2022-08-29
Bump llvm revision to use bytecode machinery

GleasonK:
Updated LLVM revision and rebuilt/tested. Looks like the display of `#sparse_tensor.encoding` has changed to no longer include `pointerBitWidth` and `indexBitWidth`.

This change will unblock StableHLO bytecode implementation.


burmako:
Manually checked that the build is passing. Approving.

burmako (approved):

Issue #34, opened 2022-08-23
Add serialization support for StableHLO


burmako:
Now that [the initial support for a binary serialization format](https://reviews.llvm.org/D131747) in MLIR upstream has landed, let's experiment with it to add serialization support for StableHLO!

PR #60, opened 2022-08-29
Bytecode implementation for StableHLO

GleasonK:
This PR plays a role in #34. There is still a little more work to do before closing this issue, as documented in `bytecode.md` of this PR.

- Implemented bytecode for StableHLO, all attributes and types. The changes here are largely based off [BuiltinDialectBytecode.cpp](https://github.com/llvm/llvm-project/blob/main/mlir/lib/IR/BuiltinDialectBytecode.cpp).  
- Added a test `bytecode.mlir`:
  + Ensures that `stablehlo-opt` and `stablehlo-opt -emit-bytecode file.mlir | stablehlo-opt` produce the same results.
  + Uses `stablehlo-bytecode` debug trace to ensure that all attrs/types have bytecoding implemented.
- Added documentation of work, findings, and open questions in `bytecode.md`.


GleasonK:
Adding Mehdi as a reviewer. I think my biggest area of concern is the datatypes chosen to represent data. Just about everything is represented as a `varint`, `svarint`, or array of `svarint[]`. I'm not sure if this is ideal for all the given datatypes, or if `APInt` should be used anywhere.

Additionally, note the test cases run in `bytecode.mlir`, they are bash-specific. Perhaps there is a better way to verify round-trip? I think I saw some conversation regarding this on the LLVM bytecode implementation review, not sure if anything ever came of that.

stablehlo/tests/bytecode.mlir
```diff
@@ -0,0 +1,769 @@
+// RUN: diff <(stablehlo-opt %s) <(stablehlo-opt -emit-bytecode %s | stablehlo-opt)
+// RUN: stablehlo-opt -emit-bytecode -debug-only=stablehlo-bytecode %s 2>&1 | (! grep 'Not Implemented')
```
joker-eph, line 2: I'm not sure that debug output is available in non assert builds

joker-eph:
> Adding Mehdi as a reviewer. I think my biggest area of concern is the datatypes chosen to represent data. Just about everything is represented as a varint, svarint, or array of svarint[]. I'm not sure if this is ideal for all the given datatypes, or if APInt should be used anywhere.

The (s)varint are the right choice right now for anything stored as integers (up to 64 bits).
If we have any attribute storing an APInt in memory, we shouldn't use this. But that seems unusual: do we have such cases?

stablehlo/tests/bytecode.mlir
GleasonK: This shouldn't cause the test to error on non-asserted builds, unless `debug-only` on non-asserted errors. On non-asserted grep doesn't find "Not Implemented" anywhere and silently pass. But this still would catch issues regarding non-bytecoded constructs in CI runs.

docs/bytecode.md
```diff
@@ -0,0 +1,208 @@
...
+depending on the change. Ensure that each location tagged with `TO ADD` 
+instructions is addressed. If so, bytecode for the attr/type should be generated
+on next call to `stablehlo-opt -emit-bytecode`.
+
+
+### Encoding `enum class` values
+Enum class values can be encoded as their underlying numeric types using `varint`. Currently all enums in StableHLO use `uint32_t` as the underlying value.
+
+## Open Questions
+### Is the compatibility guarantee of StableHLO dependent on other dialects?
```
jpienaar, line 199: Yes and no, but more yes it is dependent. I recommend to Eugene that an audit be done of all the attribute types used. This is also why for this layer one should either use things that aspire to be stable (as StableHLO does) such as ml_program (e.g., use ml_program.func rather than func.func) or are stable by "simplicity" (e.g., IntegerAttr) and curate. Now even with that, there are still levels of freedom to provide guarantees here independent of this, but starting with not overcomplicating ones life is good.

docs/bytecode.md
```diff
@@ -0,0 +1,208 @@
...
+If something about the way the builtin dialect serializes information changes,
+are all other dialects compatibility broken? Since our artifacts will include
+bits that need to be (de)serialized using builtin dialect information.
+
+### Any way to reduce the reliance on attribute/op name?
+What would happen if we renamed `fft_length` to `fft_length_int` for example,
```
jpienaar, line 205: Test it :-) Note there are going to be two levels of upgrade hooks here post versioning is added. But good to start with unversioned case and see what happens. You may want to try a couple of variants of renaming, of course one can look at the source code too but seeing executionally what happens and then considering that when the upgrade hooks are discussed (one under review, the other coming post) I think improves the feel.

It's good to also consider creating a policy for StableHLO updates to minimize these (as protobuf for example does too).

docs/bytecode.md
joker-eph: I suggested before using some verifier (possibly a pass) to check the content of a module and have an allowlist of type/attributes that can be present.

stablehlo/dialect/StablehloBytecode.cpp
```diff
@@ -0,0 +1,695 @@
...
+      "stablehlo-bytecode",                                                   \
+      llvm::errs() << "Called: " << _EXTRACT_AFTER(__PRETTY_FUNCTION__, func) \
+                   << '\n')
+
+#define LOG_WRITE_CALL _LOG_CALL_TO("write")
+#define LOG_READ_CALL _LOG_CALL_TO(__func__)
+#define LOG_NOT_IMPLEMENTED \
+  DEBUG_WITH_TYPE(          \
+      "stablehlo-bytecode", \
+      llvm::errs() << "***Not Implemented: " << __PRETTY_FUNCTION__ << '\n')
```
burmako, line 36: It's helpful to have these logging facilities to better understand what's going on! In the future, we'll likely want to standardize on our approach to logging as part of #15. Meanwhile this LGTM.

stablehlo/dialect/StablehloBytecode.cpp
```diff
@@ -0,0 +1,695 @@
...
+  // Use the attributes underlying type to get the numeric value.
+  // Note this may cause issues if enums use an int64_t and have a large value.
+  // All enums in StableHLO currently use int32_t.
+  template <typename EnumType, typename EnumTypeAttr, typename SymbolizeFn>
+  EnumTypeAttr readEnumAttribute(DialectBytecodeReader &reader,
+                                 SymbolizeFn symbolizeFn) const {
+    uint64_t code;
+    if (failed(reader.readVarInt(code))) return EnumTypeAttr();
+
+    llvm::Optional<EnumType> enumOpt = symbolizeFn(static_cast<uint32_t>(code));
```
burmako, line 265: You can say `auto enumOpt = ...` , and then you won't have to have the `EnumType` template parameter.

stablehlo/tests/bytecode.mlir
```diff
@@ -0,0 +1,780 @@
...
+func.func @test_alias_attribute (%arg0: tuple<tensor<f32>>
+    {stablehlo.result_alias = #stablehlo.result_alias<
+      tuple_indices = [0],
+      result_index = [0, 0],
+      must_alias>}
+    ) -> (tuple<tensor<f32>>) {
+  %0 = stablehlo.get_tuple_element %arg0[0] : (tuple<tensor<f32>>) -> tensor<f32>
+  %1 = stablehlo.tuple %0 : tuple<tensor<f32>>
+  func.return %1 : tuple<tensor<f32>>
+}
```
burmako, line 780: That's amazingly thorough! My only two requests would be to: 1) sort alphabetically to simplify human analysis, 2) have function names match the tested entity, e.g. `func.func @cross_replica_sum` instead of `func.func @ test_crs` (although we cannot say `func.func @cross-replica-sum` without double quotes which looks weird).

burmako (changes requested):

docs/bytecode.md
GleasonK: Created #69

docs/bytecode.md
GleasonK: Created #70.

docs/bytecode.md
```diff
@@ -0,0 +1,208 @@
+# StableHLO Bytecode
+
+## Basic Example
+A minimalist example which only encodes a single attribute in StableHLO can be found [at this commit](https://github.com/openxla/stablehlo/commit/00e3dc98d8d956e5e494be3022df973821e58e91).
+
+There is an image at the bottom of the page which shows difference in the bianry file.
+
+## Currently Encoded Attributes / Types
```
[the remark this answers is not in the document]
GleasonK: Agree. I think more so this section is not needed now that the "### Adding Bytecode for a New Type / Attribute" section exists. This was supposed to be a teaching tool more than anything, but the "how to add" section fills the same role.

docs/bytecode.md
```diff
@@ -0,0 +1,208 @@
...
+  isMustAlias: varint
+}
+
+ChannelHandleAttr {
+  handle: svarint
+  type: svarint
+}
+
+ComparisonDirectionAttr
+  ComparisonDirection: varint
```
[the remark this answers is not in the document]
GleasonK: Open to this change. I wanted to note in some way that these attributes are a `ComparisonDirection` enum value encoded as a `varint`. I'll think how to do this best, or if it's even necessary.

stablehlo/dialect/StablehloBytecode.cpp
```diff
@@ -0,0 +1,695 @@
...
+
+  // Enum reader and writer. Many attrs have a single enum type to serialize.
+  // Use the attributes underlying type to get the numeric value.
+  // Note this may cause issues if enums use an int64_t and have a large value.
+  // All enums in StableHLO currently use int32_t.
+  template <typename EnumType, typename EnumTypeAttr, typename SymbolizeFn>
+  EnumTypeAttr readEnumAttribute(DialectBytecodeReader &reader,
+                                 SymbolizeFn symbolizeFn) const {
+    uint64_t code;
+    if (failed(reader.readVarInt(code))) return EnumTypeAttr();
```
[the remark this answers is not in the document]
GleasonK: Is this suggesting to use `return {};` instead of the empty `return SomeAttr();`? If so this decision was based on how BuiltinDialect implemented this:
https://github.com/llvm/llvm-project/blob/main/mlir/lib/IR/BuiltinDialectBytecode.cpp#L443

stablehlo/dialect/StablehloBytecode.cpp
GleasonK: Agree - not sure if we want to make our own small logging library built on top of LLVM to help conform to our standards once they exist, but may be helpful.

stablehlo/dialect/StablehloBytecode.cpp
GleasonK: Nice, changed it.

stablehlo/tests/bytecode.mlir
GleasonK: Agree - I actually had a slightly different plan for this. We have `stablehlo/tests/ops_chlo_roundtrip.mlir` which is just an error-free exhaustive test of chlo ops. I think `bytecode.mlir` would be better suited as `ops_stablehlo_roundtrip.mlir` and bytecode testing can be built on top of that. Similarly chlo bytecode testing is on top of `ops_chlo_roundtrip.mlir`. I'll do this migration in a followup change if that's ok?

GleasonK: ^ Reason that work is non-trivial is that I want to convert it to a FileCheck test in addition to a bytecode test, which involves a bit of modification.

stablehlo/dialect/StablehloBytecode.cpp
```diff
@@ -0,0 +1,695 @@
...
+}
+
+ArgResultAliasAttr StablehloBytecodeInterface::readArgResultAliasAttr(
+    DialectBytecodeReader &reader) const {
+  LOG_READ_CALL;
+
+  llvm::SmallVector<int64_t> argTupleIndices;
+  int64_t resultIndex;
+  llvm::SmallVector<int64_t> resultTupleIndices;
+  uint64_t isMustAliasUint;
```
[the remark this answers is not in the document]
GleasonK: May have something to do with "Special case of IntegerAttr to represent boolean integers, i.e., signless i1 integers." ([ref](https://mlir.llvm.org/doxygen/classmlir_1_1BoolAttr.html#details)). Not sure if that conversion method resolves this?

Also this is just a bool value `"bool":$isMustAlias`, not a BoolAttr (unless there's something implicit going on).

stablehlo/tests/bytecode.mlir
GleasonK: Will alphabetize the test point in the meantime, that part is simple

docs/bytecode.md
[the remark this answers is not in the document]
GleasonK: Replaced with `value: varint (encoded enum)`

docs/bytecode.md
```diff
@@ -0,0 +1,208 @@
+# StableHLO Bytecode
+
+## Basic Example
+A minimalist example which only encodes a single attribute in StableHLO can be found [at this commit](https://github.com/openxla/stablehlo/commit/00e3dc98d8d956e5e494be3022df973821e58e91).
+
+There is an image at the bottom of the page which shows difference in the bianry file.
```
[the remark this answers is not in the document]
GleasonK: (moved from comment below, responded to incorrect comment) 
Agree. I think more so this section is not needed now that the "### Adding Bytecode for a New Type / Attribute" section exists. This was supposed to be a teaching tool more than anything, but the "how to add" section fills the same role.

docs/bytecode.md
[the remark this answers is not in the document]
GleasonK: Will be easier to fiddle with linking after this is merged. I'm thinking to link to a repository search result for "namespace stablehlo_encoding" since that should be reliable, as the enum declarations for attrs/types is the only thing in that namespace.

docs/bytecode.md
```diff
@@ -0,0 +1,208 @@
...
+### Is the compatibility guarantee of StableHLO dependent on other dialects?
+If something about the way the builtin dialect serializes information changes,
+are all other dialects compatibility broken? Since our artifacts will include
+bits that need to be (de)serialized using builtin dialect information.
+
+### Any way to reduce the reliance on attribute/op name?
+What would happen if we renamed `fft_length` to `fft_length_int` for example, 
+is this a breaking change?
+
+Similar quesiton for renaming `cross-replica-sum` to `cross_replica_sum`.
```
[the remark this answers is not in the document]
GleasonK: section was removed and issue created.

joker-eph (approved):

stablehlo/dialect/StablehloBytecode.h
```diff
@@ -0,0 +1,28 @@
...
+#ifndef STABLEHLO_DIALECT_STABLEHLO_BYTECODE_H
+#define STABLEHLO_DIALECT_STABLEHLO_BYTECODE_H
+
+namespace mlir {
+namespace stablehlo {
+class StablehloDialect;
+
+// Add the interfaces necessary for encoding the stablehlo dialect components
+// in bytecode.
+void addBytecodeInterface(StablehloDialect *dialect);
```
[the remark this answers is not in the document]
GleasonK: Based on BuiltinDialect: https://github.com/llvm/llvm-project/blob/df4e637ca7ef4ef17b662845120864921e65bb67/mlir/lib/IR/BuiltinDialectBytecode.h

Minimal risk of this going wrong, but I suppose using a more precise type prevents the accidental attempt to add the StableHLO bytecode interface to the CHLO dialect. Would be odd to attempt, but maybe with some bad copy-pasting it could happen.

docs/bytecode.md
```diff
@@ -0,0 +1,208 @@
...
+  inputBatchDimension: svarint
+  inputFeatureDimension: svarint
+  inputSpatialDimensions: svarint[]
+  kernelInputFeatureDimension: svarint
+  kernelOutputFeatureDimension: svarint
+  kernelSpatialDimensions: svarint[]
+  outputBatchDimension: svarint
+  outputFeatureDimension: svarint
+  outputSpatialDimensions: svarint[]
+}
```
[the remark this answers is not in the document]
GleasonK: Yeah so slight edge case on custom call api verision, unclear if this is intentional or not. This enum only defined as

Note the following line for `FftType`:
https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/StablehloEnums.td#L72
```
def StableHLO_FftTypeAttr : EnumAttr<StableHLO_Dialect, StableHLO_FftType, "fft_type">;
```

There is no corresponding declaration that ties `CustomCallApiVersion` to the StableHLO dialect, so it is only defined/used as a `Int32EnumAttr` and therefore handled by the builtin encoding it seems.

I added the following note to `bytecode.md`:

```
Special Cases:
- `StableHLO_ConvolutionAttributes`
  + Despite its name,  is not an attribute and is not encoded.
    Rather, it is a dag which gets expanded into several attributes
    which are all encoded separately.
- `StableHLO_CustomCallApiVersionAttr`
  + This enum is defined strictly as an attribute of `I32EnumAttr`
    and not an `EnumAttr` of the `StablehloDialect`. This differs from
   `FftType` and other enum attributes. Because of this, it is handled by
    the builtin encoding.
```

docs/bytecode.md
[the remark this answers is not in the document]
burmako: "Will be easier to fiddle with linking after this is merged". Sounds good.

docs/bytecode.md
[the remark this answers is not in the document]
burmako: Hmm I don't think that you pushed this update?

stablehlo/dialect/StablehloBytecode.cpp
```diff
@@ -0,0 +1,695 @@
+/* Copyright 2022 The StableHLO Authors.
+StablehloBytecode.cpp - StableHLO Bytecode Implementation */
```
burmako, line 2: Let's use the usual license blurb here?

stablehlo/dialect/StablehloBytecode.cpp
```diff
@@ -0,0 +1,695 @@
...
+//===----------------------------------------------------------------------===//
+// Encoding
+//===----------------------------------------------------------------------===//
+
+namespace {
+namespace stablehlo_encoding {
+
+/// This enum contains marker codes used to indicate which attribute is
+/// currently being decoded, and how it should be decoded. The order of these
+/// codes should generally be unchanged, as any changes will inevitably break
```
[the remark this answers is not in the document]
burmako: Do you anticipate situations where the order should be changed? (I guess, this is an eventuality that "generally" is preparing the reader for).

stablehlo/dialect/StablehloBytecode.cpp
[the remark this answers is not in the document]
burmako: Interesting! Nevermind then I guess.

stablehlo/dialect/StablehloBytecode.cpp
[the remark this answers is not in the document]
burmako: You're right, it's just a bool, not a BoolAttr. This was just a related thought that came to my mind as I was reading the code.

docs/bytecode.md
```diff
@@ -0,0 +1,202 @@
...
+```
+$ stablehlo-opt -emit-bytecode stablehlo/tests/print_stablehlo.mlir | stablehlo-opt
+```
+
+### Find out what attributes or types are not encoded:
+Since attributes and types that don't get encoded are instead stored as strings,
+the `strings` command can be used to see what attributes were missed:
+
+_Note: The following trace is from a previous revision where the `scatter` attribute 
+was not implemented. Currently all types/attrs are implemented and log only shows
```
burmako, line 159: How about we remove this sentence and its accompanying output in the printout below? "The following trace is from a previous revision where the `scatter` attribute was not implemented". Even without `stablehlo.scatter`, the output is educational enough?

stablehlo/tests/bytecode.mlir
burmako: Sounds like a plan!

burmako (approved):

docs/bytecode.md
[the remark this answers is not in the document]
GleasonK: Huh.. where did that go.. let me double check, may have been on the wrong branch making that edit

stablehlo/dialect/StablehloBytecode.cpp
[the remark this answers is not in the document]
GleasonK: Once we start offering stability guarantees then no, likely never. 

However if we decided to remove the arg alias attr in the near future before stability gets worked out, as it appears to be unused, I'd propose we remove the enum and decrement the rest by 1

PR #65, opened 2022-08-30
Add specifications for constant operator

### LINKED ISSUE #64: Add specification for constant operator
### Request description

Add the specification of `stablehlo.constant`.

### Additional context

_No response_

ghpvnist:
fixes #64

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -206,8 +206,8 @@ Produces a `result` tensor from a constant `value`.
 
 ### Examples
 
-```c
-%0 = stablehlo.constant dense<1> : tensor<pred>
+```mlir
+%0 = stablehlo.constant dense<true> : tensor<pred>
 // %0: true 
 
 %1 = stablehlo.constant dense<0> : tensor<i32>
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -134,19 +134,25 @@ specification.
 
 ### Semantics
 
-Creates a constant value or array from `value` and produces a `result` tensor.
+Produces a `result` tensor from a constant `value`.
 
 ### Attributes
 
 | Name | Type |
 |-|-|
-| `value` | `tensor` of Integer, Floating-point and Complex element types |
+| `value` | `tensor` of boolean, integer, floating-point or complex element types |
+
+### Operands 
+
+| Name | Type |
+|-|-|
+| `value` | Constant n-dimensional data of boolean, integer, floating-point, or complex element types |
 
 ### Results
 
 | Name | Type |
 |-|-|
-| `result` | `tensor` of Integer, Floating-point, and Complex element types |
+| `result` | `tensor` of boolean, integer, floating-point, or complex element types |
 
 ### Constraints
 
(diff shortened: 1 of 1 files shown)


sdasgup3:
1. For the semantics section, how about

Produces a tensor `result` using the constant data `value`.

2. Let us avoid mlir notation as much we can. What do yo think about:

### Operands
| Name | Type |
|-|-|
| `value` | constant n-dimensional data of integer, floating-point, or complex element types |

docs/spec_draft.md
```diff
@@ -127,3 +127,43 @@ specification.
...
+%x = "stablehlo.constant" dense<0> : (tensor<i32>) -> tensor<i32>
+// %x: 0
+
+%y = "stablehlo.constant" dense<[[0.0, 1.0], [2.0, 3.0]]> : (tensor<2x2xf32>) -> tensor<2x2xf32>
+// %y: [
+//       [0.0, 1.0],
+//       [2.0, 3.0]
+//     ]
+
+%z = "stablehlo.constant" dense<[(0.0, 1.0), (2.0, 3.0)]> : (tensor<2xcomplex<f32>>) -> tensor<2xcomplex<f32>>
```
GleasonK, line 167: Did you hand-write this or is this generated? It looks like a mix of generic and custom assembly.

Constant has pretty print, lets use the pretty where possible. Looks like the  `stablehlo.add` example above uses the binary-op pretty syntax.

```
%1 = stablehlo.constant dense<2> : tensor<i32>
```

GleasonK (changes requested):

docs/spec_draft.md
```diff
@@ -127,3 +127,43 @@ specification.
...
 // %z: [[6, 8], [10, 12]]
 ```
+
+## stablehlo.constant
+
+`stablehlo.constant(value) -> result`
+
+### Semantics
+
+Creates a constant value or array from `value` and produces a `result` tensor.
```
burmako, line 137: What does "value or array" mean here? Perhaps consider rewording to "Creates a `result` tensor from a constant `value`" to avoid the ambiguity?

docs/spec_draft.md
```diff
@@ -127,3 +127,43 @@ specification.
...
+
+### Attributes
+
+| Name | Type |
+|-|-|
+| `value` | `tensor` of Integer, Floating-point and Complex element types |
```
burmako, line 143: How about saying "tensor of any supported element type" to be more concise? I know this is how `stablehlo.add` is currently specced, and we have a few more specs in the pipeline, so perhaps this is not the right venue for discussing this wording. /cc @sdasgup3

burmako (changes requested):

burmako:
"constant n-dimensional data of integer, floating-point, or complex element types". The spec defines what a constant is ([link](https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md#constants)). Perhaps reformulate as "constant of any supported type"? There's no reason to exclude boolean constants, right?

docs/spec_draft.md
ghpvnist: @sdasgup3 made a similar comment earlier, and I'll update the wording to "Produces a `result` tensor from a constant `value`".

docs/spec_draft.md
```diff
@@ -127,3 +127,52 @@ specification.
...
+| `result` | `tensor` of boolean, integer, floating-point, or complex element types |
+
+### Constraints
+
+  * Supported shapes: all static shapes.
+
+### Examples
+
+```c
+%0 = stablehlo.constant dense<1> : tensor<pred>
```
burmako, line 164: `dense<true>` to follow the syntax for constants from above?

burmako (approved):

docs/spec_draft.md
```diff
@@ -127,3 +127,52 @@ specification.
...
+|-|-|
+| `result` | `tensor` of boolean, integer, floating-point, or complex element types |
+
+### Constraints
+
+  * Supported shapes: all static shapes.
+
+### Examples
+
+```c
```
sdasgup3, line 163: mlir?

docs/spec_draft.md
```diff
@@ -127,3 +127,52 @@ specification.
...
+%0 = stablehlo.constant dense<1> : tensor<pred>
+// %0: true 
+
+%1 = stablehlo.constant dense<0> : tensor<i32>
```
sdasgup3, line 167: (A) If we plan to keep the example snippet here, we can have better names `%result` etc., (B) Else we can replace it with links to the interpreter test file. 

Not relevant to the current PR: Both (A) or (B) can be handled in a serate PR.

docs/spec_draft.md
sdasgup3: Having variable names  relatable to spec definitely makes sense.

GleasonK (approved):

PR #67, opened 2022-08-31
Add type inference for stablehlo.batch_norm_* ops.

zhouxin913:
This change adds type inference method for `stablehlo.batch_norm_grad`,  `stablehlo.batch_norm_training` and `stablehlo.batch_norm_inference`, and revisit their `verify()`:
1. Existing comments in `verify()` are verbose and hard to read and maintain: ODS provides clear constrains for those many operands/atts/result and much more readable than the long comments in nature language. Besides, the existing comments also miss some points like element type match for some items. So they are removed, instead, this change states clearly that `verify()`only contains checks beyond ODS.
2. These ops shares quite similar logic, so refactor by reusing them.
3. XLA semantics is very long because it includes everything that the ODS of StableHLO already covers. The missing pieces are covered by `verify()`.

### DIFF
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -2294,65 +2294,58 @@ LogicalResult AllGatherOp::verify() {
 // BatchNormGradOp
 //===----------------------------------------------------------------------===//
 
-LogicalResult BatchNormGradOp::verify() {
-  // The following properties are already enforced by the ODS:
-  //  1. Inputs 'operand' & 'grad_output' and outputs 'grad_operand',
-  //     are ranked-tensors with floating-point (fp) type.
-  //  2. The shapes of inputs 'operand' & 'grad_output' match.
-  //  3. Inputs 'scale', 'mean', 'variance' and Outputs 'grad_scale',
-  //     'grad_offset'  are all 1D fp tensors with same shape.
-  //  4. The element-types of input 'operand' and outputs 'grad_scale',
-  //     'grad_offset' match.
-  //  5. The type of input 'operand' and output 'grad_operand' match.
-  //
-  // We intend to verify the following properties
-  //  P1. Inputs 'operand' & 'grad_output' has the same shape with fp
-  //      element-types, ignoring fp-precision : Inferred from (1) & (2).
-  //  P2. The feature dimension 'feature_index' is a valid index in 'operand':
-  //      Inferred from check C2 below.
-  //  P3. Inputs 'scale', 'mean', 'variance' must be 1D tensors with same shape
-  //      and fp element-type (ignoring precision) and the number of elements
-  //      in its sole-dimension == number of features in the 'operand's
-  //      feature-dimension 'feature_index': Inferred from (3) and check C3
-  //      below.
-  //  P4. Outputs 'grad_scale' & 'grad_offset' are 1D tensors with
-  //      element-type == element-type of(operand) and same shape as any of
-  //      the inputs 'scale', 'mean', or 'variance': Inferred from (3), (4) and
-  //      check C3 below.
-  //  P5. The type (shape + element-type) of input 'operand' and
-  //      output 'grad_operand' must match: Inferred from (5).
+LogicalResult verifyBatchNorm(Location loc, Value operand,
+                              int64_t feature_index, Value scale) {
+  auto operandType = operand.getType().cast<RankedTensorType>();
+  if (static_cast<int64_t>(feature_index) >= operandType.getRank())
+    return emitError(loc) << "expects feature_index to be smaller "
+                             "than the rank of operand type; got feature_index "
+                          << feature_index << ", and rank "
+                          << operandType.getRank() << ".";
 
-  // C2.
-  auto operandType = operand().getType().cast<RankedTensorType>();
-  if (static_cast<int64_t>(feature_index()) >= operandType.getRank())
-    return emitOpError() << "expects feature_index to be smaller "
-                            "than the rank of operand type; got feature_index "
-                         << feature_index() << ", and rank "
-                         << operandType.getRank() << ".";
+  if (static_cast<int64_t>(feature_index) < 0)
+    return emitError(loc) << "expects feature_index to be a "
+                          << "non-negative number, got "
+                          << static_cast<int64_t>(feature_index) << ".";
 
-  if (static_cast<int64_t>(feature_index()) < 0)
-    return emitOpError() << "expects feature_index to be a "
-                         << "non-negative number, got "
-                         << static_cast<int64_t>(feature_index()) << ".";
-
-  auto gradOutputType = grad_output().getType().cast<RankedTensorType>();
-  if (operandType.getRank() != gradOutputType.getRank())
-    return emitOpError() << "expects 'operand' and 'grad_output' to have the "
-                            "same rank. but got rank(oprand) "
-                         << operandType.getRank() << " and rank(grad_output) "
-                         << gradOutputType.getRank() << ".";
-
-  // C3.
-  const int64_t featureCount = operandType.getShape()[feature_index()];
+  const int64_t featureCount = operandType.getShape()[feature_index];
   const int64_t scaleShape =
-      scale().getType().cast<RankedTensorType>().getShape()[0];
+      scale.getType().cast<RankedTensorType>().getShape()[0];
+  // As ODS enforces `scale`, `mean`, `variance`, `offset` are AllShapesMatch,
+  // this also infers that featureCount is aligned with them.
   if (scaleShape != featureCount)
-    return emitOpError() << "expects the size of scale factor to be "
-                            "same as the feature count,"
-                            " but the size of scale factor is "
-                         << scaleShape << " and the feature count is "
-                         << featureCount << ".";
+    return emitError(loc) << "expects the size of scale factor to be "
+                             "same as the feature count,"
+                             " but the size of scale factor is "
+                          << scaleShape << " and the feature count is "
+                          << featureCount << ".";
 
+  return success();
+}
+
+// Refer ODS for properties that are already enforced including shapes and
+// element types. This verifier includes additional checks.
+LogicalResult BatchNormGradOp::verify() {
+  if (failed(verifyBatchNorm(getLoc(), operand(), feature_index(), scale())))
+    return failure();
+  return success();
+}
+
+LogicalResult BatchNormGradOp::inferReturnTypeComponents(
+    MLIRContext* context, Optional<Location> location, ValueShapeRange operands,
+    DictionaryAttr attributes, RegionRange regions,
+    SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
+  BatchNormGradOp::Adaptor adaptor(operands, attributes, regions);
+
+  RankedTensorType operandType =
+      adaptor.operand().getType().cast<RankedTensorType>();
+  inferredReturnShapes.emplace_back(operandType.getShape(),
+                                    operandType.getElementType());
+
+  const int64_t featureCount = operandType.getShape()[adaptor.feature_index()];
+  SmallVector<int64_t> featureShape(1, featureCount);
+  inferredReturnShapes.emplace_back(featureShape, operandType.getElementType());
+  inferredReturnShapes.emplace_back(featureShape, operandType.getElementType());
   return success();
 }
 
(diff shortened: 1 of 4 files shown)


stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -4040,6 +4040,15 @@ func.func @error_incompatible_alias_element_types (%arg0: tensor<2xf32> {stableh
 
 // stablehlo.batch_norm_training
 
+// CHECK-LABEL: @batch_norm_train 
+func.func @batch_norm_train(%input: tensor<2x2x2x2xf32>, %scale: tensor<2xf32>, %offset: tensor<2xf32>) -> tuple<tensor<2x2x2x2xf32>, tensor<2xf32>, tensor<2xf32>> {
+  %0:3 = "stablehlo.batch_norm_training" (%input, %scale, %offset) {epsilon = 0.001 : f32, feature_index = 1 : i64} : (tensor<2x2x2x2xf32>, tensor<2xf32>, tensor<2xf32>) -> (tensor<2x2x2x2xf32>, tensor<2xf32>, tensor<2xf32>)
+  %1 = "stablehlo.tuple"(%0#0, %0#1, %0#2) : (tensor<2x2x2x2xf32>, tensor<2xf32>, tensor<2xf32>) -> tuple<tensor<2x2x2x2xf32>, tensor<2xf32>, tensor<2xf32>>
```
sdasgup3, line 4046: Can we avoid the tuple just like the example you added for batch_norm_grad?

stablehlo/dialect/StablehloOps.td
```diff
@@ -1460,7 +1460,8 @@ def StableHLO_BatchNormGradOp : StableHLO_Op<"batch_norm_grad", [NoSideEffect,
         "grad_offset"]>,
     AllShapesMatch<["operand", "grad_output"]>,
     AllElementTypesMatch<["operand", "grad_scale", "grad_offset"]>,
-    AllTypesMatch<["operand", "grad_operand"]>]> {
+    AllTypesMatch<["operand", "grad_operand"]>,
+    InferTensorType]> {
```
burmako, line 1464: If you expand the definition of this op, then it'll say `let results` twice. What does this mean?

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2294,148 +2294,112 @@ LogicalResult AllGatherOp::verify() {
...
-                            "same rank. but got rank(oprand) "
-                         << operandType.getRank() << " and rank(grad_output) "
-                         << gradOutputType.getRank() << ".";
-
-  // C3.
-  const int64_t featureCount = operandType.getShape()[feature_index()];
+LogicalResult verifyBatchNorm(Location loc, Value operand,
+                              int64_t feature_index, Value scale) {
+  auto operandType = operand.getType().cast<RankedTensorType>();
+  if (static_cast<int64_t>(feature_index) >= operandType.getRank())
```
burmako, line 2300: `feature_index` is already `int64_t`, so the cast is not needed. (Here and below).

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2294,148 +2294,112 @@ LogicalResult AllGatherOp::verify() {
...
+                             "than the rank of operand type; got feature_index "
+                          << feature_index << ", and rank "
+                          << operandType.getRank() << ".";
+
+  if (static_cast<int64_t>(feature_index) < 0)
+    return emitError(loc) << "expects feature_index to be a "
+                          << "non-negative number, got "
+                          << static_cast<int64_t>(feature_index) << ".";
+
+  const int64_t featureCount = operandType.getShape()[feature_index];
```
burmako, line 2311: `.getDimSize(feature_index)`? (Here and below).

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2294,148 +2294,112 @@ LogicalResult AllGatherOp::verify() {
...
+
+  RankedTensorType operandType =
+      adaptor.operand().getType().cast<RankedTensorType>();
+  inferredReturnShapes.emplace_back(operandType.getShape(),
+                                    operandType.getElementType());
 
+  const int64_t featureCount = operandType.getShape()[adaptor.feature_index()];
+  SmallVector<int64_t> featureShape(1, featureCount);
+  inferredReturnShapes.emplace_back(featureShape, operandType.getElementType());
+  inferredReturnShapes.emplace_back(featureShape, operandType.getElementType());
```
burmako, line 2348: Can we say `{featureCount}` instead? (Here and below).

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2294,148 +2294,112 @@ LogicalResult AllGatherOp::verify() {
...
+LogicalResult BatchNormGradOp::inferReturnTypeComponents(
+    MLIRContext* context, Optional<Location> location, ValueShapeRange operands,
+    DictionaryAttr attributes, RegionRange regions,
+    SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
+  BatchNormGradOp::Adaptor adaptor(operands, attributes, regions);
+
+  RankedTensorType operandType =
+      adaptor.operand().getType().cast<RankedTensorType>();
+  inferredReturnShapes.emplace_back(operandType.getShape(),
+                                    operandType.getElementType());
```
burmako, line 2343: I think this can be simplified to `inferredReturnShapes.emplace_back(operandType)` because `operandType` is a `ShapedType`. (Here and below).

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2294,148 +2294,112 @@ LogicalResult AllGatherOp::verify() {
...
+  return success();
+}
+
+LogicalResult BatchNormGradOp::inferReturnTypeComponents(
+    MLIRContext* context, Optional<Location> location, ValueShapeRange operands,
+    DictionaryAttr attributes, RegionRange regions,
+    SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
+  BatchNormGradOp::Adaptor adaptor(operands, attributes, regions);
+
+  RankedTensorType operandType =
```
burmako, line 2340: `auto operandType = ...` given that the type is obvious from the context. (Here and below).

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2294,148 +2294,112 @@ LogicalResult AllGatherOp::verify() {
...
-                            "than the rank of operand type; got feature_index "
-                         << feature_index() << ", and rank "
-                         << operandType.getRank() << ".";
-
-  if (static_cast<int64_t>(feature_index()) < 0)
-    return emitOpError() << "expects feature_index to be a "
-                         << "non-negative number, got "
-                         << static_cast<int64_t>(feature_index()) << ".";
+  if (failed(verifyBatchNorm(getLoc(), operand(), feature_index(), scale())))
+    return failure();
```
burmako, line 2360: Are we checking that `operand`'s rank is at least 1? https://github.com/tensorflow/tensorflow/blob/d078f90b2833df402ff5c39e328f46e0bfd761b1/tensorflow/compiler/xla/service/shape_inference.cc#L1241-L1246

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2294,148 +2294,112 @@ LogicalResult AllGatherOp::verify() {
...
-  // Together with (4) implies that 'scale', 'offset', 'mean', &
-  // 'variance' all have the same shape.
-  if (scaleSize != featureCount)
-    return emitOpError() << "expects the size of scale factor to be "
-                            "same as the feature count,"
-                            " but the size of scale factor is "
-                         << scaleSize << " and the feature count is "
-                         << featureCount << ".";
+  if (failed(verifyBatchNorm(getLoc(), operand(), feature_index(), scale())))
+    return failure();
```
burmako, line 2390: Are we checking that `operand`'s rank is at least 1? https://github.com/tensorflow/tensorflow/blob/d078f90b2833df402ff5c39e328f46e0bfd761b1/tensorflow/compiler/xla/service/shape_inference.cc#L1347-L1352

stablehlo/dialect/StablehloOps.td
```diff
@@ -1460,7 +1460,8 @@ def StableHLO_BatchNormGradOp : StableHLO_Op<"batch_norm_grad", [NoSideEffect,
         "grad_offset"]>,
     AllShapesMatch<["operand", "grad_output"]>,
     AllElementTypesMatch<["operand", "grad_scale", "grad_offset"]>,
```
burmako, line 1462: Given that shape_inference.cc is checking for `SameElementTypeIgnoringFpPrecision`, do we need `AllElementTypesMatch` at all? Since ODS enforces all participating values to be tensors of `HLO_Float`, `SameElementTypeIgnoringFpPrecision` is achieved automatically, right? (Here and below).

burmako (changes requested):

stablehlo/tests/ops_stablehlo.mlir
zhouxin913: Sure,  I will delete it from the negative tests as well.

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Yes the signature already covert the uint64_t (from feature_index) to `int64_t` so these `static_cast<int64_t>` is not needed anymore.

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Done.

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Tried but not working. Though `SmallVector<int64_t> x = {1};` is valid expression, but `inferredReturnShapes.emplace_back({1}, operandType.getElementType());` is not.

burmako: Got it. Then how about simply changing the definition of `featureShape` to `SmallVector<int64_t> featureShape{featureCount}`?

stablehlo/dialect/StablehloOps.cpp
burmako: You probably fixed this locally but haven't pushed?

stablehlo/dialect/StablehloOps.td
zhouxin913: Removed the first line as sync with Sandeep offline. It was a typo.

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Sure!

stablehlo/dialect/StablehloOps.cpp
zhouxin913: This looks better but actually not working. Yes `operandType` is a `ShapedType`, but it is also a `Type`. `ShapedTypeComponents` (the element type of `inferredReturnShapes` has contructors:
```
ShapedTypeComponents(Type elementType) ...
ShapedTypeComponents(ShapedType shapedType) : attr(nullptr) ...
```
So the Type one is used rather than what we expect. As I see, `ShapedType` is not a `Type`. But our `operandType` is fit for both constructors, so it ends up to be used as elementType and error out.

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Done.

stablehlo/dialect/StablehloOps.cpp
zhouxin913: There is an old comment I should add (bring back): `// Note:A valid value of feature-index implies 'operand_type.getRank() >=1'.` 
So, after we assert `0 <= feature_index < operandType.getRank()` in `verifyBatchNorm`, it implies that operand's rank is at least 1

stablehlo/dialect/StablehloOps.cpp
zhouxin913: same as the comment above.

stablehlo/dialect/StablehloOps.td
zhouxin913: Good point!

stablehlo/dialect/StablehloOps.cpp
zhouxin913: pushed. Thanks!

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2294,148 +2294,112 @@ LogicalResult AllGatherOp::verify() {
...
+                          << "non-negative number, got " << feature_index
+                          << ".";
+
+  // Note: the above checks '0 <= feature-index < operandType.getRank()'
+  // imply 'operand_type.getRank() >= 1'.
+
+  const int64_t featureCount = operandType.getDimSize(feature_index);
   const int64_t scaleShape =
-      scale().getType().cast<RankedTensorType>().getShape()[0];
+      scale.getType().cast<RankedTensorType>().getShape()[0];
```
burmako, line 2316: `.getDimSize(0)`

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2294,148 +2294,112 @@ LogicalResult AllGatherOp::verify() {
...
+    DictionaryAttr attributes, RegionRange regions,
+    SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
+  BatchNormGradOp::Adaptor adaptor(operands, attributes, regions);
+
+  auto operandType = adaptor.operand().getType().cast<RankedTensorType>();
+  inferredReturnShapes.emplace_back(operandType.getShape(),
+                                    operandType.getElementType());
+
+  const int64_t featureCount = operandType.getDimSize(adaptor.feature_index());
+  SmallVector<int64_t> featureShape({featureCount});
```
burmako, line 2348: You can drop the parentheses and say `SmallVector<int64_t> featureShape{featureCount}`.

stablehlo/dialect/StablehloOps.cpp
burmako: What I meant is that we can `.cast<ShapedType>()` and then the simpler `emplace_back` will work.

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Nice!

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Done.

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Got it. That's clean and fixed.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2294,148 +2294,109 @@ LogicalResult AllGatherOp::verify() {
...
+}
 
+LogicalResult BatchNormGradOp::inferReturnTypeComponents(
+    MLIRContext* context, Optional<Location> location, ValueShapeRange operands,
+    DictionaryAttr attributes, RegionRange regions,
+    SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
+  BatchNormGradOp::Adaptor adaptor(operands, attributes, regions);
+
+  auto operandType = adaptor.operand().getType().cast<RankedTensorType>();
+  inferredReturnShapes.emplace_back(operandType.cast<ShapedType>());
```
burmako, line 2344: This can be simplified as follows:

```
  auto operandType = adaptor.operand().getType().cast<ShapedType>();
  inferredReturnShapes.emplace_back(operandType);
```

burmako (approved):
LGTM except for one minor comment. Not merging just yet pending Sandeep's review.

sdasgup3 (approved):

Issue #73, opened 2022-09-02
Add the specifications of negate/abs operators


sdasgup3:
### Request description

Add specification for stablehlo.abs and stablehlo.negate.

PR #74, opened 2022-09-02
spec-draft: negate/abs

sdasgup3:
closes https://github.com/openxla/stablehlo/issues/73

Also, fixes some typos in the specification docs.

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -107,8 +107,7 @@ The specification of an op comprises of the following components (in the order
 
 Performs element-wise absolute value of `operand` tensor and produces a `result`
 tensor. For floating-point element types, implements the `abs` operation from
-the IEEE-754 specification. For a complex value `(a, b)`, the absolute value is
-defined as $\sqrt(a^2 + b^2)$.
+the IEEE-754 specification.
 
 For n-bit signed integer, the absolute value of $-2^{n-1}$ is implementation
 defined and one of the following:
@@ -121,13 +120,13 @@ defined and one of the following:
 
 | Name | Type |
 |-|-|
-| `operand` | tensor of signed integer, floating-point, or complex element types |
+| `operand` | tensor of integer, floating-point, or complex element types |
 
 ### Results
 
 | Name | Type |
 |-|-|
-| `result` | tensor of signed integer, floating-point, or complex element types |
+| `result` | tensor of integer, floating-point, or complex element types |
 
 ### Constraints
 
@@ -467,14 +466,15 @@ lexicographic comparison on the (real, imaginary) pairs.
 
 Performs element-wise negation of `operand` tensor and produces a `result`
 tensor. For floating-point element types, implements the `nagate` operation from
-the IEEE-754 specification. For complex types, the operation negates both the
-real and the imaginary parts.
+the IEEE-754 specification. For integer types, performs the [two's complement](https://en.wikipedia.org/wiki/Two's_complement)
+operation.
 
-For n-bit signed integer, the negation of $-2^{n-1}$ is implementation
-defined and one of the following:
+For n-bit signed integer, the negation of $-2^{n-1}$ is implementation defined
+and one of the following:
 
   * Saturation to $2^{n-1}-1$
   * $-2^n-1$
+
 
 ### Operands
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -465,29 +465,30 @@ lexicographic comparison on the (real, imaginary) pairs.
 ### Semantics
 
 Performs element-wise negation of `operand` tensor and produces a `result`
-tensor. For floating-point element types, implements the `nagate` operation from
-the IEEE-754 specification. For integer types, performs the [two's complement](https://en.wikipedia.org/wiki/Two's_complement)
-operation.
-
-For n-bit signed integer, the negation of $-2^{n-1}$ is implementation defined
+tensor. For floating-point element types, implements the `negate` operation from
+the IEEE-754 specification. For signed integer types, performs the regular
+negation operation, where the negation of $-2^{n-1}$ is implementation defined
 and one of the following:
 
   * Saturation to $2^{n-1}-1$
   * $-2^n-1$
 
+For unsigned integer types, bitcasts to the corresponding signed integer type,
+    performs the regular negation operation and bitcasts back to the original
+    unsigned integer type.
 
 ### Operands
 
 | Name | Type |
 |-|-|
-| `operand` | tensor of signed integer, floating-point, or complex element types |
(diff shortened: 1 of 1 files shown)


docs/spec_draft.md
```diff
@@ -254,6 +311,54 @@ the (real, imaginary) pairs.
...
+defined and one of the followings:
+
+  * Saturation to $2^{n-1}-1$
+  * $-2^n-1$
+
+### Operands
+
+| Name | Type |
+|-|-|
+| `operand` | tensor of signed integer, floating-point, or complex element types |
```
sdasgup3, line 334: Unlike the MHLO version, the StableHLO op does not support unsigned types as it does not makes sense. 

The decision is similar to : https://www.tensorflow.org/api_docs/python/tf/math/negative
Moreover,  abs op has related somewhat semantics to negate and the supported types for the mhlo.abs operator has unsigned excluded.

docs/spec_draft.md
```diff
@@ -82,13 +82,70 @@ The specification of an op comprises of the following components (in the order
...
+`stablehlo.abs(operand) -> result`
+
+### Semantics
+
+Performs element-wise absolute value of `operand` tensor and produces a `result`
+tensor. For a complex value `(a, b)`, the absolute value is defined as
+$\sqrt(a^2 + b^2)$.
+
+For n-bit signed integer, the absolute value of $-2^{n-1}$ is implementation
+defined and one of the followings:
```
burmako, line 96: Typo: following.

docs/spec_draft.md
```diff
@@ -82,13 +82,70 @@ The specification of an op comprises of the following components (in the order
...
+
+  * Saturation to $2^{n-1}-1$
+  * $-2^n-1$
+
```
burmako, line 100: What about special cases for floating point? (Same question for `negate`).

docs/spec_draft.md
burmako: Any idea why [shape_inference.cc for kNegate](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/shape_inference.cc#L308-L317) supports unsigned types? Should we change it to not support them?

"as it does not makes sense". Does anyone use negate on unsigned types at the moment? E.g. TF? JAX? PyTorch/XLA? If someone does, then we'll need to support that as well for backward compatibility reasons.

burmako (changes requested):

docs/spec_draft.md
```diff
@@ -82,13 +82,70 @@ The specification of an op comprises of the following components (in the order
...
 
+### stablehlo.abs
+
+`stablehlo.abs(operand) -> result`
+
+### Semantics
+
+Performs element-wise absolute value of `operand` tensor and produces a `result`
+tensor. For a complex value `(a, b)`, the absolute value is defined as
+$\sqrt(a^2 + b^2)$.
```
burmako, line 93: How about we remove the formula, since it's defined in calculus textbooks anyway? 

This is similar to other related simplifications that we discussed recently - e.g. a proposal to remove details of how min/max is supposed to work on floats or to remove the formula for tanh. The less text we write, the less mistakes we can make :)

In comparison, I think that the `minimum`/`maximum` semantics for complex numbers is important to keep in our specification, because I don't think this is part of textbook calculus.

docs/spec_draft.md
```diff
@@ -254,6 +311,54 @@ the (real, imaginary) pairs.
...
 
+## stablehlo.negate
+
+`stablehlo.negate(operand) -> result`
+
+### Semantics
+
+Performs element-wise negation of `operand` tensor and produces a `result`
+tensor. For complex types, the operation negates both the real and the imaginary
+parts.
```
burmako, line 322: This also seems to be part of calculus textbooks. E.g. the first result in my Google search says: https://www2.clarku.edu/~djoyce/complex/plane.html#:~:text=Negation%20and%20Subtraction.&text=Of%20course%2C%20the%20negation%20of,left%20and%20one%20unit%20down.

docs/spec_draft.md
sdasgup3: added.

docs/spec_draft.md
sdasgup3: >> Any idea why supports unsigned types?
Still exploring.

Added the unsigned type. The current behavior is same as that of HLO. 
[Jax](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.neg.html#jax.lax.neg) and XLA support them. From [PyTorch spec](https://pytorch.org/docs/stable/generated/torch.neg.html?highlight=neg#torch.neg), it seems they do also support.

sdasgup3:
Unlike the HLO, added the unsigned type for abs op. [Jax](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.abs.html#jax.lax.abs) supports it.

docs/spec_draft.md
```diff
@@ -397,6 +458,58 @@ lexicographic comparison on the (real, imaginary) pairs.
...
 [Back to Ops](#index-of-documented-ops)
 
+## stablehlo.negate
+
+`stablehlo.negate(operand) -> result`
+
+### Semantics
+
+Performs element-wise negation of `operand` tensor and produces a `result`
+tensor. For floating-point element types, implements the `nagate` operation from
```
burmako, line 468: Typo: "negate".

docs/spec_draft.md
burmako: "Operands" and "Results" still say "tensor of signed integer...".

burmako (changes requested):

docs/spec_draft.md
```diff
@@ -397,6 +458,58 @@ lexicographic comparison on the (real, imaginary) pairs.
...
+the IEEE-754 specification. For integer types, performs the [two's complement](https://en.wikipedia.org/wiki/Two's_complement)
+operation.
```
burmako, line 470: Above, we're already saying that signed integers are using two’s complement representation, so it looks like for them this sentence redundant. Is this written for unsigned integers?

Perhaps reformulate to provide more details, e.g. "For signed integer types, performs the regular negation operation, where the negation of $-2^{n-1}$ is implementation defined and one of the following: <...>. For unsigned integer types, bitcasts to the corresponding signed integer type, performs the regular negation operation and bitcasts back to the original unsigned integer type"?

sdasgup3: I wrote them for both given that in two's complement arithmetic is same for the fundamental  operations  like add/sub/ etc for unsigned binary numbers (as long as the inputs are represented in the same number of bits as the output, and any overflow beyond those bits is discarded from the result).

I like your suggestion which is definitely more clearer.

burmako (approved):

Issue #75, opened 2022-09-02
Modify clang-format to match the current state of the repository


GleasonK:
### Request description

Currently the Lint action validates google syntax, while the `.clang-format` file specifies `LLVM`.

This is because it is based on TensorFlow's mhlo configuration:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/mlir_hlo/.clang-format

However, for some reason `mlir-hlo` doesn't seem to use LLVM formatting uniformly, it uses google formatting. Note the "two spaces between code and comment:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/mlir_hlo/lib/Dialect/mhlo/IR/hlo_ops.cc#L84 

Given this, we should use google formatting in the short term. 

In the near future we should resolve #15, as the longer we wait to change a coding style (if required), the more merge conflicts there will be.

joker-eph:
> Given this, we should likely use google formatting in the short term, and consider flipping the repo to LLVM formatting at a later date. This requires edits to almost every cpp/h file in stablehlo which will cause merge conflicts.

How will this get better? Wouldn't we have less merge conflicts while doing it ASAP before we get many many PRs and the codebase get larger?

burmako:
"How will this get better?". I read Kevin's comment not as a suggestion that this will get better, but rather that we should update `.clang-format` to reflect reality and have a discussion about which code style we want to follow (#15)

joker-eph:
Well I was reacting to the part about "consider flipping the repo to LLVM formatting at a later date"

GleasonK:
Both reasonable. I edited the issue title and description. Mehdi's comment does note a bit more urgency required by #15, the longer we wait, the more intrusive it will be to make any edits, if required. In the meantime, I'll modify the clang format to reflect the current state of this repo.

joker-eph:
Here is what the opposite would look like FYI: https://github.com/openxla/stablehlo/pull/76 (so that we don't talk in hypothetical)

GleasonK:
This is a tradeoff between consistency with MHLO and a preference to use LLVM formatting. Will it cause issues during integrations if the two have different formatting? I think that's the main thing I want to avoid.

burmako:
[...]

PR #76, opened 2022-09-02
Align the code with the clang-format spec (NFC)

joker-eph:
See #75


joker-eph:
(Closing: this was just opened as an example anyway)

PR #84, opened 2022-09-03
Fix examples to include operand and result names

### LINKED ISSUE #80: Spec: Fix examples to include operand and result names
The examples currently use single alphabet characters to represent variables. The examples should use the respective operand and result names.

ghpvnist:
fixes #80 and minor typos

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -134,10 +134,10 @@ specification.
 ### Examples
 
 ```mlir
-// %x: [[1, 2], [3, 4]]
-// %y: [[5, 6], [7, 8]]
-%z = stablehlo.add %x, %y : tensor<2x2xf32>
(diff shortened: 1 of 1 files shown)


docs/spec_draft.md
```diff
@@ -186,19 +186,19 @@ Produces a `result` tensor from a constant `value`.
 
 | Name | Type |
 |-|-|
-| `value` | `tensor` of boolean, integer, floating-point or complex element types |
+| `value` | `tensor` of any supported type |
```
sdasgup3, line 189: types?
here and below

sdasgup3 (changes requested):

docs/spec_draft.md
ghpvnist: Done

sdasgup3 (approved):

PR #86, opened 2022-09-03
Add specifications for sine, cosine, and tanh operators

### LINKED ISSUE #85: Add the specifications of sine/cosine/tanh operators

ghpvnist:
fixes #85

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -263,7 +263,7 @@ Produces a `result` tensor from a constant `value`.
 
 ```mlir
 %result = stablehlo.constant dense<true> : tensor<pred>
-// %result: true 
+// %result: true
 
 %result = stablehlo.constant dense<0> : tensor<i32>
 // %result: 0
@@ -287,8 +287,8 @@ Produces a `result` tensor from a constant `value`.
 ### Semantics
 
 Performs element-wise cosine operation on `operand` tensor and produces a
-`result` tensor, according to the IEEE-754 specification. Numeric precision is
-implementation-defined.
+`result` tensor, implementing the `cos` operation from the IEEE-754
+specification. Numeric precision is implementation-defined.
 
 ### Operands
 
@@ -535,8 +535,8 @@ operation.
 ### Semantics
 
 Performs element-wise sine operation on `operand` tensor and produces a `result`
-tensor, according to the IEEE-754 specification. Numeric precision is
-implementation-defined.
+tensor, implementing the `sin` operation from the IEEE-754
+specification. Numeric precision is implementation-defined.
 
 ### Operands
 
@@ -575,8 +575,8 @@ implementation-defined.
 ### Semantics
 
 Performs element-wise tanh operation on `operand` tensor and produces a `result`
-tensor, according to the IEEE-754 specification. Numeric precision is
-implementation-defined.
+tensor, implementing the `tanh` operation from the IEEE-754
+specification. Numeric precision is implementation-defined.
 
 ### Operands
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -577,7 +577,6 @@ implementation-defined.
 Performs element-wise tanh operation on `operand` tensor and produces a `result`
 tensor, according to the IEEE-754 specification. Numeric precision is
 implementation-defined.
-$$tanh(x)=\frac{e^{x}-e^{-x}}{e^{x}+e^{-x}}$$
 
 ### Operands
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -90,11 +90,14 @@ The specification of an op comprises of the following components (in the order
    * [and](#stablehloand)
    * [ceil](#stablehloceil)
    * [constant](#stablehloconstant)
+   * [cosine](#stablehlocosine)
    * [floor](#stablehlofloor)
    * [max](#stablehlomaximum)
    * [min](#stablehlominimum)
    * [not](#stablehlonot)
    * [or](#stablehloor)
+   * [sine](#stablehlosine)
+   * [tanh](#stablehlotanh)
    * [xor](#stablehloxor)
 
 ## stablehlo.add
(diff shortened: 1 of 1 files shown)


docs/spec_draft.md
```diff
@@ -223,6 +224,48 @@ Produces a `result` tensor from a constant `value`.
...
+
+| Name | Type |
+|-|-|
+| `value` | Constant n-dimensional data of floating-point, or complex element types |
+
+### Results
+
+| Name | Type |
+|-|-|
+| `result` | `tensor` of floating-point, or complex element types |
```
sdasgup3, line 251: remove the "," when having just two items? Here and other similar places

docs/spec_draft.md
```diff
@@ -223,6 +224,48 @@ Produces a `result` tensor from a constant `value`.
...
 
+## stablehlo.cosine
+
+`stablehlo.cosine(operand) -> result`
+
+### Semantics
+
+Performs element-wise cosine operation on `operand` tensor and produces a `result` tensor.
+
+### Attributes
```
sdasgup3, line 235: We do not need an Attribute for this op.

docs/spec_draft.md
```diff
@@ -223,6 +224,48 @@ Produces a `result` tensor from a constant `value`.
...
+
+| Name | Type |
+|-|-|
+| `value` | `tensor` of floating-point or complex element types |
+
+### Operands
+
+| Name | Type |
+|-|-|
+| `value` | Constant n-dimensional data of floating-point, or complex element types |
```
sdasgup3, line 245: operand instead of value?

sdasgup3, line 245: We should use tensor which is already defined as n dimensional data. Something like

`tensor of floating-point or complex element types`

sdasgup3: here and below

sdasgup3: here and below

docs/spec_draft.md
sdasgup3: here and below

sdasgup3 (changes requested):

docs/spec_draft.md
```diff
@@ -223,6 +224,48 @@ Produces a `result` tensor from a constant `value`.
 // %3: [(0.0, 1.0), (2.0, 3.0)]
 ```
 
+## stablehlo.cosine
+
+`stablehlo.cosine(operand) -> result`
+
+### Semantics
+
+Performs element-wise cosine operation on `operand` tensor and produces a `result` tensor.
```
burmako, line 233: What does it mean to perform a cosine operation? What kind of precision with respect to an ideal mathematical result are we talking about? What are the corner cases? (Same questions for other ops in this PR and for sqrt/rsqrt in the neighboring PR).

burmako (changes requested):

docs/spec_draft.md
ghpvnist: Fixed

docs/spec_draft.md
ghpvnist: Fixed

docs/spec_draft.md
ghpvnist: Fixed

ghpvnist: Fixed

docs/spec_draft.md
```diff
@@ -405,6 +448,80 @@ operation.
...
+[Back to Ops](#index-of-documented-ops)
+
+## stablehlo.tanh
+
+`stablehlo.tanh(operand) -> result`
+
+### Semantics
+
+Performs element-wise tanh operation on `operand` tensor and produces a `result` tensor.
+$$tanh(x)=\frac{e^{x}-e^{-x}}{e^{x}+e^{-x}}$$
```
burmako, line 496: What should be our policy for providing explanations on what some of these operations do? How about we consistently refer to IEEE-754 if they are part of it, or express them in terms of IEEE-754 operations if they aren't?

E.g. `tanh` is called out in IEEE-754, so we could just say "Performs element-wise tanh operation on `operand` tensor and produces a `result` tensor, according to the IEEE-754 specification"?

burmako, line 496: On a separate note, as discussed offline, let's also call out numeric precision for floating-point ops. Initially, that could be as simple as "Numeric precision is implementation-defined".

burmako (changes requested):

docs/spec_draft.md
ghpvnist: As mentioned in later comments, we will initially state, "Numeric precision is implementation-defined."

docs/spec_draft.md
```diff
@@ -405,6 +450,84 @@ operation.
...
+## stablehlo.tanh
+
+`stablehlo.tanh(operand) -> result`
+
+### Semantics
+
+Performs element-wise tanh operation on `operand` tensor and produces a `result`
+tensor, according to the IEEE-754 specification. Numeric precision is
+implementation-defined.
+$$tanh(x)=\frac{e^{x}-e^{-x}}{e^{x}+e^{-x}}$$
```
burmako, line 502: Let's drop the formula, given that we're referenced the specification?

burmako (changes requested):

docs/spec_draft.md
ghpvnist: Done

docs/spec_draft.md
ghpvnist: Done

burmako (approved):

docs/spec_draft.md
```diff
@@ -277,6 +280,46 @@ Produces a `result` tensor from a constant `value`.
...
 [Back to Ops](#index-of-documented-ops)
 
+## stablehlo.cosine
+
+`stablehlo.cosine(operand) -> result`
+
+### Semantics
+
+Performs element-wise cosine operation on `operand` tensor and produces a
+`result` tensor, according to the IEEE-754 specification. Numeric precision is
```
sdasgup3, line 290: `implementing to the `cos` operation from the IEEE-754 specification.`
Same for other ops below.

docs/spec_draft.md
```diff
@@ -277,6 +280,46 @@ Produces a `result` tensor from a constant `value`.
...
+implementation-defined.
+
+### Operands
+
+| Name | Type |
+|-|-|
+| `operand` | tensor of floating-point or complex element types |
```
sdasgup3, line 297: Can we say something for the complex numbers? We can check with HLO?

For example for sine,
```
sine(a-+ib) = sin(a)*cosh(b) +i cos(a)*sinh(b)
```
Same for other ops.

https://proofwiki.org/wiki/Sine_of_Complex_Number/Proof_1

sdasgup3: Ignore the previous comments. 
From the fact that this is a standard textbook information,  let us just skip mentioning it here in the spec. Same for ops below

docs/spec_draft.md
ghpvnist: Done.

sdasgup3 (approved):

PR #87, opened 2022-09-03
spec: Add an index of documented ops

sdasgup3:
The PR adds a tables indexing to all the documented ops and back to the index. This improves navigation through the doc.

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -82,6 +82,17 @@ The specification of an op comprises of the following components (in the order
   * **Constraints** Type constraints on the operand(s), result(s).
   * **Examples** Examples demonstrating the working of the op.
 
+
+## Index of documented ops
+   * [AddOp](#stablehloadd)
+   * [AndOp](#stablehloand)
+   * [ConstantOp](#stablehloconstant)
+   * [MaxOp](#stablehlomaximum)
+   * [MinOp](#stablehlominimum)
+   * [NotOp](#stablehlonot)
+   * [OrOp](#stablehloor)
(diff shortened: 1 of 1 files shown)


docs/spec_draft.md
```diff
@@ -82,6 +82,17 @@ The specification of an op comprises of the following components (in the order
   * **Constraints** Type constraints on the operand(s), result(s).
   * **Examples** Examples demonstrating the working of the op.
 
+
+## Index of documented ops
```
burmako, line 86: Let's just calls this "Ops" to match other similar headers which are called "Types", "Programs", etc?

docs/spec_draft.md
```diff
@@ -82,6 +82,17 @@ The specification of an op comprises of the following components (in the order
...
+   * [AddOp](#stablehloadd)
```
burmako, line 87: Let's use the lowercase names introduced later in the document, e.g. `add`, `constant`, etc? I don't think names like `AddOp` are used anywhere else.

docs/spec_draft.md
```diff
@@ -129,6 +140,8 @@ specification.
 // %z: [[6, 8], [10, 12]]
 ```
 
+[top](#index-of-documented-ops)
```
burmako, line 143: How about something more descriptive e.g. "Back to Ops"?

burmako (changes requested):
This is a great idea. I've left some minor comments, but overall LGTM.

burmako (approved):
Oops I meant to say "Approve". Please address the comments mentioned above though.

PR #89, opened 2022-09-03
Add specifications for sqrt and rsqrt operators

### LINKED ISSUE #88: Add the specifications of sqrt/rsqrt operators

ghpvnist:
fixes #88

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -98,7 +98,9 @@ The specification of an op comprises of the following components (in the order
    * [negate](#stablehlonegate)
    * [not](#stablehlonot)
    * [or](#stablehloor)
+   * [rsqrt](#stablehlorsqrt)
    * [sine](#stablehlosine)
+   * [sqrt](#stablehlosqrt)
    * [tanh](#stablehlotanh)
    * [xor](#stablehloxor)
 
@@ -642,6 +644,42 @@ operation.
 
 [Back to Ops](#index-of-documented-ops)
 
+## stablehlo.rsqrt
+
+`stablehlo.rsqrt(operand) -> result`
+
+### Semantics
+
+Performs element-wise reciprocal square root operation on `operand` tensor and produces a `result` tensor.
+$$rsqrt(x) = \frac{1}{\sqrt{x}}$$
+
+### Operands
+
+| Name | Type |
+|-|-|
+| `operand` | tensor of floating-point or complex element types |
+
+### Results
+
+| Name | Type |
+|-|-|
+| `result` | tensor of floating-point or complex element types |
+
+### Constraints
+
+  * Supported shapes: all static shapes.
+  * `result` must have the same type as that of `operand`.
+
+### Examples
+
+```mlir
+// %operand: [[1.0, 4.0], [9.0, 25.0]]
+%result = stablehlo.rsqrt %operand : tensor<2x2xf32>
+// %result: [[1.0, 0.5], [0.33333343, 0.2]]
+```
+
+[Back to Ops](#index-of-documented-ops)
+
 ## stablehlo.sine
 
 `stablehlo.sine(operand) -> result`
@@ -678,6 +716,41 @@ specification. Numeric precision is implementation-defined.
              ]
 %result = stablehlo.sine %operand : tensor<2x2xf32>
 // %result: [[0.0, 1.0], [0.0, -1.0]]
+```
+
+[Back to Ops](#index-of-documented-ops)
+
+## stablehlo.sqrt
+
+`stablehlo.sqrt(operand) -> result`
+
+### Semantics
+
+Performs element-wise square root operation on `operand` tensor and produces a `result` tensor.
+
+### Operands
+
+| Name | Type |
+|-|-|
+| `operand` | tensor of floating-point or complex element types |
+
+### Results
+
+| Name | Type |
+|-|-|
+| `result` | tensor of floating-point or complex element types |
+
+### Constraints
+
+  * Supported shapes: all static shapes.
+  * `result` must have the same type as that of `operand`.
+
+### Examples
+
+```mlir
+// %operand: [[0.0, 1.0], [4.0, 9.0]]
+%result = stablehlo.sqrt %operand : tensor<2x2xf32>
+// %result: [[0.0, 1.0], [2.0, 3.0]]
 ```
 
 [Back to Ops](#index-of-documented-ops)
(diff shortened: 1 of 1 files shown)


docs/spec_draft.md
```diff
@@ -385,6 +385,86 @@ operation.
...
+// %result: [[1.0, 0.5], [0.33333343, 0.2]]
+```
+
+## stablehlo.sqrt
+
+`stablehlo.sqrt(operand) -> result`
+
+### Semantics
+
+Performs element-wise sqrt operation on `operand` tensor and produces a `result` tensor.
```
sdasgup3, line 434: Perform element-wise square root operation on ... 

and remove the equation below?
As that is not conveying anything extra

docs/spec_draft.md
```diff
@@ -385,6 +385,86 @@ operation.
...
+// %operand: [[1.0, 4.0], [9.0, 25.0]]
+%result = stablehlo.rsqrt %operand : tensor<2x2xf32>
+// %result: [[1.0, 0.5], [0.33333343, 0.2]]
+```
+
+## stablehlo.sqrt
+
+`stablehlo.sqrt(operand) -> result`
+
+### Semantics
```
sdasgup3, line 432: Also can we say something about negation int values?

sdasgup3 (changes requested):

docs/spec_draft.md
```diff
@@ -385,6 +385,86 @@ operation.
...
+
+Performs element-wise rsqrt operation on `operand` tensor and produces a `result` tensor.
+$$rsqrt(x) = \frac{1}{\sqrt{x}}$$
+
+### Attributes
+
+| Name | Type |
+|-|-|
+| `value` | `tensor` of floating-point or complex element types |
```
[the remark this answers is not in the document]
ghpvnist: Fixed

docs/spec_draft.md
```diff
@@ -385,6 +385,86 @@ operation.
...
+
+### Operands
+
+| Name | Type |
+|-|-|
+| `value` | constant n-dimensional data of floating-point, or complex element types |
```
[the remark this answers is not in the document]
ghpvnist: Fixed

docs/spec_draft.md
ghpvnist: Fixed

docs/spec_draft.md
sdasgup3: Or just point out the appropriate operation from IEEE 754 standard that this op will be implementing. That way we can skip the details.

ghpvnist: Done

docs/spec_draft.md
```diff
@@ -682,6 +721,43 @@ specification. Numeric precision is implementation-defined.
...
 [Back to Ops](#index-of-documented-ops)
 
+## stablehlo.sqrt
+
+`stablehlo.sqrt(operand) -> result`
+
+### Semantics
+
+Performs element-wise square root operation on `operand` tensor and produces a
+`result` tensor, implementing the `sqrt` operation from the IEEE-754
```
sdasgup3, line 731: I could not find  `sqrt` as a standalone operation in the standard. Once it is used in the formula of `rsqrt`. 
However, I can find the name of the operation as `squareRoot`.

docs/spec_draft.md
```diff
@@ -642,6 +644,43 @@ operation.
...
+| Name | Type |
+|-|-|
+| `result` | tensor of floating-point or complex element types |
+
+### Constraints
+
+  * Supported shapes: all static shapes.
+  * `result` must have the same type as that of `operand`.
+
+### Examples
```
sdasgup3, line 674: An example with complex number? Here and below.

docs/spec_draft.md
```diff
@@ -642,6 +644,43 @@ operation.
...
 
+## stablehlo.rsqrt
+
+`stablehlo.rsqrt(operand) -> result`
+
+### Semantics
+
+Performs element-wise reciprocal square root operation on `operand` tensor and
+produces a `result` tensor, implementing the `rSqrt` operation from the IEEE-754
+specification. Numeric precision is implementation-defined.
```
sdasgup3, line 655: Numeric precision is more of a concern for the transcendental function. For the non-transcendental function we do not have to write it.

docs/spec_draft.md
ghpvnist: Fixed

docs/spec_draft.md
ghpvnist: Done

docs/spec_draft.md
ghpvnist: Table 9.1 of IEEE-754 (2008) has a list of "recommended correctly rounded functions." These do not have any guarantees for numeric precision, including `rSqrt`. However, I will remove it for sqrt since `squareRoot` is defined as an arithmetic operation in section 5.4.1.

sdasgup3 (approved):

docs/spec_draft.md
```diff
@@ -682,6 +725,47 @@ specification. Numeric precision is implementation-defined.
...
 
+## stablehlo.sqrt
+
+`stablehlo.sqrt(operand) -> result`
+
+### Semantics
+
+Performs element-wise square root operation on `operand` tensor and produces a
+`result` tensor, implementing the `squareRoot` operation from the IEEE-754
+specification.
```
burmako, line 736: I noticed that `rsqrt` says "Numeric precision is implementation-defined", but `sqrt` doesn't. Why is there a difference?

ghpvnist: `squareRoot` is defined as an arithmetic operation in section 5.4.1 (required to have 0.5 ulp) whereas `rSqrt` in Table 9.1 is *recommended* to have correctly rounded functions (this combines two arithmetic operations `division` and `squareRoot` which does not guarantee 0.5 ulp when combined together), thus having to state so for `rsqrt`.

sdasgup3: IMO, for `rsqrt` op, the IEEE standard delegates the precision to be defined by the language standard or implementation. For example, if implemented in a naive way, `sqrt` followed by `divide`, can have two rounding errors as opposed to one if implemented using very large precision computation.

OTOH, for ops like `sqrt` the precision is defined as

```
In this standard, a computational operation that returns a numeric result first produces an unrounded result
as an exact number of infinite precision. That unrounded result is then rounded to a destination format.

```
 
I think that is the reason `rsqrt` have its precision documented in this spec as implementation defined.

burmako: Thank you for the detailed explanation! I didn't realize this.

PR #91, opened 2022-09-05
Add type inference for stablehlo.reduce_window

zhouxin913:
The new `ReduceWindowOp::inferReturnTypeComponents()` replaces the old `ReduceWindowOp::verify()` with all the checking logic. And expect to cover all XLA checks as well.

And invalid test with `llvm.add` is removed in the same reason as the description of [the PR of "Type inference of ReduceOp"](https://github.com/openxla/stablehlo/pull/50). Please see the reason there.

`ReduceWindowOp` has similar but not same code as `ReduceOp` which is not ideal. So align them to be same in P1~P2 checks.

To address the legacy TODOs, `convertDenseIntAttr` add check for 1-D tensor, and change all caller respectively. Those ops like `SelectAndScatterOp` will be fully addressed in later PRs recently.

### DIFF
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -178,36 +178,44 @@ Value maybeCastTo(OpBuilder& b, Location loc, Value value, Type type) {
 //===----------------------------------------------------------------------===//
 
 // Convert a 1D dense int64 attribute to a list of values.
-SmallVector<int64_t> convertDenseIntAttr(
-    llvm::Optional<mlir::DenseIntElementsAttr> optionalAttr) {
+FailureOr<SmallVector<int64_t>> convertDenseIntAttr(
+    Optional<DenseIntElementsAttr> optionalAttr, Optional<Location> loc) {
   if (!optionalAttr.has_value()) return SmallVector<int64_t>{};
 
-  mlir::DenseIntElementsAttr attr = *optionalAttr;
+  DenseIntElementsAttr attr = *optionalAttr;
+  auto attrType = attr.getType().cast<RankedTensorType>();  // ensured by ODS.
+  if (attrType.getRank() != 1)
+    return (emitOptionalError(
+                loc, "expects the shape of attribute to be 1-D, but got {",
+                attrType.getShape(), "}."),
+            failure());
   auto values = attr.getValues<int64_t>();
-  return {values.begin(), values.end()};
+  return SmallVector<int64_t>{values.begin(), values.end()};
 }
 
-// Convert a 1D or Nx2 dense int64 attribute to a list of tuples.
-FailureOr<SmallVector<std::pair<int64_t, int64_t>>> convertNx2Attribute(
-    llvm::Optional<mlir::DenseIntElementsAttr> optionalAttr, Location loc) {
+// Convert a 1D or Nx2 dense int64 padding attribute to a list of tuples.
+FailureOr<SmallVector<std::pair<int64_t, int64_t>>> convertPaddingAttribute(
+    Optional<DenseIntElementsAttr> optionalAttr, Optional<Location> loc) {
   if (!optionalAttr.has_value())
     return SmallVector<std::pair<int64_t, int64_t>>{};
-  mlir::DenseIntElementsAttr attr = *optionalAttr;
 
+  DenseIntElementsAttr attr = *optionalAttr;
   auto attrType = attr.getType().cast<RankedTensorType>();  // ensured by ODS.
   if (attrType.getRank() > 1) {
     if (attrType.getRank() != 2 || attrType.getShape()[1] != 2)
-      return (mlir::emitError(loc) << "expects the shape of padding-attribute "
-                                      "to be {N, 2}, but got {"
-                                   << attrType.getShape() << "}.",
-              failure());
+      return (
+          emitOptionalError(
+              loc,
+              "expects the shape of padding-attribute to be {N, 2}, but got {",
+              attrType.getShape(), "}."),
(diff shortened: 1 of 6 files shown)


burmako (changes requested):

stablehlo/dialect/StablehloOps.td
```diff
@@ -2396,7 +2396,8 @@ def StableHLO_TriangularSolveOp: StableHLO_Op<"triangular_solve",
 def StableHLO_ReduceWindowOp: StableHLO_Op<"reduce_window", [
       RecursiveSideEffects,
       SameVariadicOperandSize,
-      SingleBlockImplicitTerminator<"ReturnOp">
+      SingleBlockImplicitTerminator<"ReturnOp">,
+      InferTensorType,
```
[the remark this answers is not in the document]
zhouxin913: Hi Eugene, thanks for this reminder. They are all addressed in the new code version. I also update PR description to help reviewers. Please take a look.

stablehlo/tests/verify_reduce_window.mlir
```diff
@@ -262,6 +268,27 @@ func.func @reduce_window_invalid_attributes(%arg0: tensor<4x2xf32>,
 
 // -----
 
+func.func @reduce_window_invalid_attributes(%arg0: tensor<4x2xf32>, %arg1: tensor<4x2xi32>,
+                    %init0: tensor<f32>, %init1: tensor<i32>) ->
+                      (tensor<2x2xf32>, tensor<2x2xi32>) {
+  // expected-error @+1 {{expects the shape of attribute to be 1-D}}
```
burmako, line 274: Of what attribute?

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -178,36 +178,44 @@ Value maybeCastTo(OpBuilder& b, Location loc, Value value, Type type) {
 //===----------------------------------------------------------------------===//
 
 // Convert a 1D dense int64 attribute to a list of values.
-SmallVector<int64_t> convertDenseIntAttr(
-    llvm::Optional<mlir::DenseIntElementsAttr> optionalAttr) {
+FailureOr<SmallVector<int64_t>> convertDenseIntAttr(
+    Optional<DenseIntElementsAttr> optionalAttr, Optional<Location> loc) {
```
burmako, line 182: It doesn't sound right to me that something with a generic name like `convertDenseIntAttr` would be picky about the shape of its argument. Perhaps rename to `convert1DAttribute`?

stablehlo/dialect/StablehloOps.td
```diff
@@ -2448,9 +2449,6 @@ def StableHLO_ReduceWindowOp: StableHLO_Op<"reduce_window", [
     See https://www.tensorflow.org/xla/operation_semantics#reducewindow.
   }];
 
-  // TODO(hinsu): Verify that padding attribute is 2-d and the remaining
-  // attributes are 1-d. Attributes' leading dimension should match rank of the
-  // operands.
```
burmako, line 2453: It looks like `convertPaddingAttribute` still accepts 1D tensors. Is that intended?

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -178,36 +178,44 @@ Value maybeCastTo(OpBuilder& b, Location loc, Value value, Type type) {
...
   if (!optionalAttr.has_value()) return SmallVector<int64_t>{};
 
-  mlir::DenseIntElementsAttr attr = *optionalAttr;
+  DenseIntElementsAttr attr = *optionalAttr;
+  auto attrType = attr.getType().cast<RankedTensorType>();  // ensured by ODS.
```
burmako, line 186: What ODS? Looks like this function could be called by anyone?

burmako (changes requested):

stablehlo/tests/verify_reduce_window.mlir
zhouxin913: To address this, I add arg "StringRef attrName" to the signature of `onvertDenseIntAttr` so that we have a more concrete error message.

stablehlo/dialect/StablehloOps.td
zhouxin913: nope, I can't find any evidence that 1D padding should be supported in StableHLO, so I'm removing it now. It may lead to some error when merging this back to MHLO (maybe break some tests there), but will keep in mind of this to fix/remove tests there too.

stablehlo/dialect/StablehloOps.cpp
zhouxin913: I think it meant `DenseIntElementsAttr`. However, I can't find the relationship between `DenseIntElementsAttr` VS `RankedTensorType` by digging into the class definition. But can't find a scenario either when a `DenseIntElementsAttr` object is NOT a `RankedTensorType`. Are you aware of that is there any case that this `cast<RankedTensorType>()` could fail? 

So leave this comment unresolved. (removing the miss-leading `// ensured by ODS.` too)

stablehlo/dialect/StablehloOps.td
burmako: Sounds good. Let's see if we can get away with this.

stablehlo/dialect/StablehloOps.cpp
burmako: "Are you aware of that is there any case that this cast<RankedTensorType>() could fail?". I don't think so. Resolving this thread - thank you for the explanation!

burmako (approved):

PR #92, opened 2022-09-06
Update the "Discussions" link

burmako:
At the GitHub organization level, we've restructured discussions, so
that they all end up at https://github.com/orgs/openxla/discussions,
similarly to how we have the OpenXLA Discord server and
multiple channels within that server.


GleasonK (approved):
Looks good - didn't know this page existed.

PR #93, opened 2022-09-06
Remove "Additional context" from bug and FR templates

burmako:
So far, additional context in our tickets has always been empty.
Let's remove it for the time being.


GleasonK (approved):
Looks good. We haven't had any bug reports yet, so unclear if we will need additional context there, but it can always be added back later if we find that the quality of a typical bug report is lacking.

PR #94, opened 2022-09-06
Clarify the floating point semantics of add/max/min

sdasgup3:
The semantics of ops on FP element types is currently specified slightly inconsistently. The idea of this PR is to have a consistent wording (if possible).

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -111,7 +111,7 @@ of the followings:
       underflow) and saturation to $2^n - 1$ (or $0$) for unsigned overflow (or
         unsigned underflow).
 
-For floating-point element types, implements the addition operation from the
+For floating-point element types, implements the `addition` operation from the
 IEEE-754 specification.
 
 ### Operands
@@ -244,7 +244,7 @@ Produces a `result` tensor from a constant `value`.
 ### Semantics
 
 Performs element-wise max operation on tensors `lhs` and `rhs` and produces a
-`result` tensor. For floating-point element types, implements the maxNum
+`result` tensor. For floating-point element types, implements the `maximum`
 operation from the IEEE-754 specification. For complex element type,  performs
 lexicographic comparison on the (real, imaginary) pairs.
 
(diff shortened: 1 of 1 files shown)


docs/spec_draft.md
```diff
@@ -109,8 +111,8 @@ of the followings:
       underflow) and saturation to $2^n - 1$ (or $0$) for unsigned overflow (or
         unsigned underflow).
 
-For floating-point element types, corner cases are defined by the IEEE-754
-specification.
+For floating-point element types, implements the addition operation from the
```
burmako, line 114: \`addition\`? (I.e. add backticks).

docs/spec_draft.md
```diff
@@ -242,10 +244,9 @@ Produces a `result` tensor from a constant `value`.
 ### Semantics
 
 Performs element-wise max operation on tensors `lhs` and `rhs` and produces a
-`result` tensor. For floating-point element type, implements IEEE 754 semantics:
-Returns the larger of two operands, propagating `NaN`s and treating `-0` as
-less than `+0`. For complex element type,  performs lexicographic comparison on
-the (real, imaginary) pairs.
+`result` tensor. For floating-point element types, implements the maxNum
```
burmako, line 247: Per https://en.wikipedia.org/wiki/IEEE_754, "As of 2019, the formerly required minNum, maxNum, minNumMag, and maxNumMag in IEEE 754-2008 are now deleted due to their non-associativity. Instead, two sets of new minimum and maximum operations are recommended... The first set contains minimum, minimumNumber, maximum and maximumNumber. The second set contains minimumMagnitude, minimumMagnitudeNumber, maximumMagnitude and maximumMagnitudeNumber. The history and motivation for this change are explained in a background document".

Should we use `maximum` or `maximumNumber` here? (Same question for `minimum` and `minimumNumber` below).

docs/spec_draft.md
```diff
@@ -242,10 +244,9 @@ Produces a `result` tensor from a constant `value`.
...
+operation from the IEEE-754 specification. For complex element type,  performs
+lexicographic comparison on the (real, imaginary) pairs.
```
burmako, line 249: Add backticks? (Here and below).

burmako (changes requested):

docs/spec_draft.md
sdasgup3: Thanks for the information!

IMO, we should go with the `maximum`/`minimum` (with NaN propagation) as that is the same behavior displayed by the corresponding HLO operation.

burmako: Following the behavior of the corresponding HLO operation SGTM for the spec draft.

burmako (approved):

PR #100, opened 2022-09-08
Add specification for log op

### LINKED ISSUE #97: Add the specification of log operator

ghpvnist:
fixes #97

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -93,6 +93,7 @@ The specification of an op comprises of the following components (in the order
    * [constant](#stablehloconstant)
    * [cosine](#stablehlocosine)
    * [floor](#stablehlofloor)
+   * [log](#stablehlolog)
    * [max](#stablehlomaximum)
    * [min](#stablehlominimum)
    * [negate](#stablehlonegate)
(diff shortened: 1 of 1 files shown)


docs/spec_draft.md
```diff
@@ -420,6 +421,47 @@ IEEE-754 specification.
...
 [Back to Ops](#index-of-documented-ops)
 
+## stablehlo.log
+
+`stablehlo.log(operand) -> result`
+
+### Semantics
+
+Performs element-wise logarithm operation on `operand` tensor and
+produces a `result` tensor, implementing the `log` operation from the IEEE-754
```
sdasgup3, line 431: As per the IEEE standard `log` is not meant for the complex op, can we re-frame as

`Performs element-wise logarithm operation on `operand` tensor and produces a `result` tensor. For floating-point element types, implements the `log` from ...`.

sdasgup3 (approved):

docs/spec_draft.md
sdasgup3: Also same comment for the already merged ops like sqrt,rsqrt,sine, cos which can be handled via a separate PR.

burmako: How does this apply to other ops which work on complex numbers, like sine and cosine?

sdasgup3: Say for sine, we have the following spec 

```
Performs element-wise sine operation on `operand` tensor and produces a `result`
tensor, implementing the `sin` operation from the IEEE-754
specification. Numeric precision is implementation-defined.
```

Again `sin` operation from IEEE is not meant for complex type.

ghpvnist: If we don't explicitly state the behavior of complex numbers, the element-wise x operation can be misinterpreted as performing x operation on the real and imaginary values separately when in reality it is computed together. Since complex numbers are not defined in IEEE-754, should the behavior be explicitly stated similar to that of [C++ documentation](https://en.cppreference.com/w/cpp/numeric/complex/log)? @burmako

burmako: Huh that's a good catch. 

How about deferring describing corner cases for complex numbers to a later point? It feels that this will take a significant chunk of time while not being essential for the first pass over the spec. We could say something along these lines:

"""
Performs element-wise logarithm operation on `operand` tensor and produces a `result` tensor. For floating-point element types, implements the `log` operation from the IEEE-754 specification. For complex element types, computes a complex logarithm, with corner cases TBD. Numeric precision is implementation-defined.
"""

burmako (changes requested):

docs/spec_draft.md
```diff
@@ -420,6 +421,47 @@ IEEE-754 specification.
...
+### Results
+
+| Name | Type |
+|-|-|
+| `result` | tensor of floating-point or complex element types |
+
+### Constraints
+
+  * Supported shapes: all static shapes.
+  * `result` must have the same type as that of `operand`.
```
burmako, line 449: Let's also update constraints, similarly to https://github.com/openxla/stablehlo/pull/103:

```
### Constraints

  * (C1) `operand` and `result` have the same type.
```

burmako (changes requested):

sdasgup3:
We need to rebase the earlier PRs properly? This includes changes from earlier PRs.

burmako (approved):

Issue #79, opened 2022-09-02
Revisit the use of mlir::Attribute as the underlying storage of Element Class


sdasgup3:
### Request description

This is based on review comment from @joker-eph 
https://github.com/openxla/stablehlo/pull/44#discussion_r960741645

At present, mlir::Attribute is used as the underlying storage for Element class [code](https://github.com/openxla/stablehlo/blob/e947f10d45003669fa7cb6e42efb125b45041389/stablehlo/reference/Element.h#L61) . This seems an expensive choice and needed to be revisited.

PR #101, opened 2022-09-08
Revisit use of mlir::Attribute storage of Element Class

sdasgup3:
The proposal here is to use a discriminated union of APInt and APFloat as the storage mechanism for Element objects.

fixes https://github.com/openxla/stablehlo/issues/79

### DIFF
--- a/docs/interpreter.md
+++ b/docs/interpreter.md
@@ -69,7 +69,7 @@ We can use the interpreter mechanism to fold operations with constant operand
 values. The following code snippet demonstrates an idea of the implementation
 for folding `stablehlo::AddOp` with floating-point typed operands:
 
-```
+```C++
 OpFoldResult AddOp::fold(ArrayRef<Attribute> attrs) {
   DenseElementsAttr lhsData = attrs[0].dyn_cast<DenseElementsAttr>();
   DenseElementsAttr rhsData = attrs[1].dyn_cast<DenseElementsAttr>();
@@ -106,7 +106,7 @@ In the current implementation, we package the inputs (MLIR program + input data
 values) and outputs in a
 [lit-based](https://llvm.org/docs/CommandGuide/lit.html) test as follows:
 
-```c++
+```C++
 // CHECK-LABEL: Evaluated results of function: add_op_test_ui4
 func.func @add_op_test_ui4() -> tensor<2xui4> {
   %0 = stablehlo.constant dense<[0, 2]> : tensor<2xui4>
--- a/docs/interpreter.md
+++ b/docs/interpreter.md
@@ -12,9 +12,12 @@ representing its data laid out in
 objects are reference-counted to simplify memory management.
 
 Individual elements of a tensor are represented using `Element` class which uses
-`mlir::Attribute` for storage. Using `mlir::Attribute` simplifies things because
-this means that we don't have to implement our own machinery for storing values
-of different types inside `Element`.
+discriminated union holding one of `APInt` or `vector<APFloat>` for storage. The
+later is used for storing elements with floating-point or complex types. Note
+that we use `mlir::Attribute` while constructing an `Element` object (via
+    `Element(Type type, Attribute attr)`) or extracting value from `Element`
+object (via `Attribute Element::getValue()`), in order to simplify these
+interfaces.
 
 `Tensor` class has the following APIs to interact with its individual elements:
   - `Element Tensor::get(llvm::ArrayRef<int64_t> index)`: To extract an
--- a/docs/interpreter.md
+++ b/docs/interpreter.md
@@ -12,12 +12,8 @@ representing its data laid out in
 objects are reference-counted to simplify memory management.
 
 Individual elements of a tensor are represented using `Element` class which uses
-discriminated union holding one of `APInt` or `vector<APFloat>` for storage. The
-later is used for storing elements with floating-point or complex types. Note
-that we use `mlir::Attribute` while constructing an `Element` object (via
-    `Element(Type type, Attribute attr)`) or extracting value from `Element`
-object (via `Attribute Element::getValue()`), in order to simplify these
-interfaces.
+discriminated union holding one of `APInt`, `APFloat` or `pair<APFloat,APFloat>`
+for storage. The last one is used for storing elements with complex types.
 
 `Tensor` class has the following APIs to interact with its individual elements:
   - `Element Tensor::get(llvm::ArrayRef<int64_t> index)`: To extract an
--- a/stablehlo/reference/Element.cpp
+++ b/stablehlo/reference/Element.cpp
@@ -34,39 +34,6 @@ inline llvm::Error invalidArgument(char const *Fmt, const Ts &...Vals) {
   return createStringError(llvm::errc::invalid_argument, Fmt, Vals...);
 }
 
-bool isSupportedUnsignedIntegerType(Type type) {
-  return type.isUnsignedInteger(4) || type.isUnsignedInteger(8) ||
-         type.isUnsignedInteger(16) || type.isUnsignedInteger(32) ||
-         type.isUnsignedInteger(64);
-}
-
-bool isSupportedSignedIntegerType(Type type) {
-  // TODO(#22): StableHLO, as bootstrapped from MHLO, inherits signless
-  // integers which was added in MHLO for legacy reasons. Going forward,
-  // StableHLO will adopt signfull integer semantics with signed and unsigned
-  // integer variants.
-  return type.isSignlessInteger(4) || type.isSignlessInteger(8) ||
-         type.isSignlessInteger(16) || type.isSignlessInteger(32) ||
-         type.isSignlessInteger(64);
-}
-
-bool isSupportedIntegerType(Type type) {
-  return isSupportedUnsignedIntegerType(type) ||
-         isSupportedSignedIntegerType(type);
-}
-
-bool isSupportedFloatType(Type type) {
-  return type.isF16() || type.isBF16() || type.isF32() || type.isF64();
-}
-
-bool isSupportedComplexType(Type type) {
-  auto complexTy = type.dyn_cast<ComplexType>();
-  if (!complexTy) return false;
-
-  auto complexElemTy = complexTy.getElementType();
-  return complexElemTy.isF32() || complexElemTy.isF64();
-}
-
 template <typename IntegerFn, typename FloatFn, typename ComplexFn>
 Element map(const Element &el, IntegerFn integerFn, FloatFn floatFn,
             ComplexFn complexFn) {
@@ -157,6 +124,34 @@ Element mapWithUpcastToDouble(const Element &el, FloatFn floatFn,
 
 }  // namespace
 
+APInt Element::getIntegerValue() const {
+  if (!isSupportedIntegerType(type_))
+    report_fatal_error(StringRef("Accessing value of a type different than "
+                                 "what is stored in Element object") +
+                       LLVM_PRETTY_FUNCTION);
+
+  return std::get<APInt>(value_);
+}
+
+APFloat Element::getFloatValue() const {
+  if (!isSupportedFloatType(type_))
+    report_fatal_error(StringRef("Accessing value of a type different than "
+                                 "what is stored in Element object") +
+                       LLVM_PRETTY_FUNCTION);
+
+  return std::get<APFloat>(value_);
+}
+
+std::complex<APFloat> Element::getComplexValue() const {
+  if (!isSupportedComplexType(type_))
+    report_fatal_error(StringRef("Accessing value of a type different than "
+                                 "what is stored in Element object") +
+                       LLVM_PRETTY_FUNCTION);
+
+  auto floatPair = std::get<std::pair<APFloat, APFloat>>(value_);
+  return std::complex<APFloat>(floatPair.first, floatPair.second);
+}
+
 Element Element::operator+(const Element &other) const {
   return map(
       *this, other, [](APInt lhs, APInt rhs) { return lhs + rhs; },
@@ -227,17 +222,52 @@ void Element::print(raw_ostream &os) const {
   }
 
   if (isSupportedComplexType(type_)) {
-    auto complexTy = type_.dyn_cast<mlir::ComplexType>();
+    auto complexElemTy = type_.dyn_cast<mlir::ComplexType>().getElementType();
     auto complexVal = getComplexValue();
-    Attribute attr =
-        complex::NumberAttr::get(complexTy, complexVal.real().convertToDouble(),
-                                 complexVal.imag().convertToDouble());
-    attr.print(os, /*elideType*/ false);
+
+    os << "[";
+    FloatAttr::get(complexElemTy, complexVal.real()).print(os);
+    os << ", ";
+    FloatAttr::get(complexElemTy, complexVal.imag()).print(os);
+    os << "]";
+
     return;
   }
 }
 
 void Element::dump() const { print(llvm::errs()); }
+
+bool isSupportedUnsignedIntegerType(Type type) {
+  return type.isUnsignedInteger(4) || type.isUnsignedInteger(8) ||
+         type.isUnsignedInteger(16) || type.isUnsignedInteger(32) ||
+         type.isUnsignedInteger(64);
+}
+
+bool isSupportedSignedIntegerType(Type type) {
+  // TODO(#22): StableHLO, as bootstrapped from MHLO, inherits signless
+  // integers which was added in MHLO for legacy reasons. Going forward,
+  // StableHLO will adopt signfull integer semantics with signed and unsigned
+  // integer variants.
+  return type.isSignlessInteger(4) || type.isSignlessInteger(8) ||
+         type.isSignlessInteger(16) || type.isSignlessInteger(32) ||
+         type.isSignlessInteger(64);
+}
+bool isSupportedIntegerType(Type type) {
+  return isSupportedUnsignedIntegerType(type) ||
+         isSupportedSignedIntegerType(type);
+}
+
+bool isSupportedFloatType(Type type) {
+  return type.isF16() || type.isBF16() || type.isF32() || type.isF64();
+}
+
+bool isSupportedComplexType(Type type) {
+  auto complexTy = type.dyn_cast<ComplexType>();
+  if (!complexTy) return false;
+
+  auto complexElemTy = complexTy.getElementType();
+  return complexElemTy.isF32() || complexElemTy.isF64();
+}
 
 }  // namespace stablehlo
 }  // namespace mlir
--- a/stablehlo/reference/Element.cpp
+++ b/stablehlo/reference/Element.cpp
@@ -244,12 +244,46 @@ Element tanh(const Element &e) {
       [](std::complex<double> e) { return std::tanh(e); });
 }
 
-void Element::print(raw_ostream &os) const { getValue().print(os); }
+void Element::print(raw_ostream &os) const {
+  // Converting APFlots/APInts to mlir::Attribute to leverage the its
+  // pretty-printing.
+  auto convertToFloatAttr = [](Type type, APFloat apf) -> Attribute {
+    return FloatAttr::get(type, apf);
+  };
 
-void Element::dump() const {
-  print(llvm::errs());
-  llvm::errs() << "\n";
+  auto convertToIntAttr = [](Type type, APInt api) -> Attribute {
+    return IntegerAttr::get(type, api);
+  };
+
+  if (isSupportedSignedIntegerType(type_)) {
+    convertToIntAttr(type_, getIntegerValue()).print(os);
+    return;
+  }
+
+  if (isSupportedUnsignedIntegerType(type_)) {
+    convertToIntAttr(type_, getIntegerValue()).print(os);
+    return;
+  }
+
+  if (isSupportedFloatType(type_)) {
+    convertToFloatAttr(type_, getFloatValue()).print(os);
+    return;
+  }
+
+  if (isSupportedComplexType(type_)) {
+    auto complexTy = type_.dyn_cast<mlir::ComplexType>().getElementType();
+    auto complexVal = getComplexValue();
+
+    os << "[";
+    convertToFloatAttr(complexTy, complexVal.real()).print(os);
+    os << ", ";
+    convertToFloatAttr(complexTy, complexVal.imag()).print(os);
+    os << "]";
+    return;
+  }
 }
+
+void Element::dump() const { print(llvm::errs()); }
 
 }  // namespace stablehlo
 }  // namespace mlir
--- a/stablehlo/reference/Element.cpp
+++ b/stablehlo/reference/Element.cpp
@@ -244,7 +244,11 @@ Element tanh(const Element &e) {
       [](std::complex<double> e) { return std::tanh(e); });
 }
 
-void Element::print(raw_ostream &os) const { value_.print(os); }
+bool Element::operator!=(const Element &other) const {
+  return !(*this == other);
+}
+
+void Element::print(raw_ostream &os) const { getValue().print(os); }
 
 void Element::dump() const {
   print(llvm::errs());
--- a/stablehlo/reference/Element.cpp
+++ b/stablehlo/reference/Element.cpp
@@ -244,10 +244,6 @@ Element tanh(const Element &e) {
       [](std::complex<double> e) { return std::tanh(e); });
 }
 
-bool Element::operator!=(const Element &other) const {
-  return !(*this == other);
-}
-
 void Element::print(raw_ostream &os) const { getValue().print(os); }
 
 void Element::dump() const {
--- a/stablehlo/reference/Element.cpp
+++ b/stablehlo/reference/Element.cpp
@@ -20,8 +20,8 @@ limitations under the License.
 #include "llvm/ADT/APFloat.h"
 #include "llvm/Support/Errc.h"
 #include "llvm/Support/Error.h"
+#include "mlir/Dialect/Complex/IR/Complex.h"
 #include "mlir/IR/BuiltinAttributes.h"
-#include "mlir/IR/BuiltinTypes.h"
 #include "mlir/Support/DebugStringHelper.h"
 
 namespace mlir {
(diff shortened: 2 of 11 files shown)


stablehlo/reference/Element.h
```diff
@@ -52,7 +57,7 @@ class Element {
 
  private:
   Type type_;
-  Attribute value_;
+  std::variant<APInt, std::vector<APFloat>> value_;
```
burmako, line 60: We should use a variant of: `APInt`, `APFloat` and `std::complex<APFloat>` - to explicitly specify how we're modelling complex numbers.

docs/interpreter.md
```diff
@@ -12,9 +12,12 @@ representing its data laid out in
...
 Individual elements of a tensor are represented using `Element` class which uses
-`mlir::Attribute` for storage. Using `mlir::Attribute` simplifies things because
-this means that we don't have to implement our own machinery for storing values
-of different types inside `Element`.
+discriminated union holding one of `APInt` or `vector<APFloat>` for storage. The
+later is used for storing elements with floating-point or complex types. Note
+that we use `mlir::Attribute` while constructing an `Element` object (via
+    `Element(Type type, Attribute attr)`) or extracting value from `Element`
+object (via `Attribute Element::getValue()`), in order to simplify these
+interfaces.
```
burmako, line 20: To be honest, I'm not sure what we're getting out of replacing the storage with `std::variant` if we're still using `mlir::Attribute` as the API.

I think that we should go all the way and completely remove all uses of `Attribute` in this file:
  1) Replace `Element::Element(Type type, Attribute attr)` with three constructors: `Type+APInt`, `Type+APFloat`, `Type+std::complex<APFloat>`.
  2) Replace `Element::getValue` with three methods: `getIntValue`, `getFloatValue` and `getComplexValue` (which are already defined in Element.cpp anyway).
  3) Update `<<` to work without attributes.

This will significantly simplify existing code. E.g. we'll be able to replace swaths of code like `return Element(elementType, FloatAttr::get(elementType, value)` with `return Element(elementType, value)`.

burmako (changes requested):

docs/interpreter.md
sdasgup3: I agree with the point that using Attributes in Element's interface API incurs huge overhead.
DOne.

stablehlo/reference/Element.h
sdasgup3: Tried using the suggested one  `std::variant<APInt, APFloat, std::complex<APFloat>>` but it gives errors related to calling the [private constructor](https://github.com/llvm-mirror/llvm/blob/2c4ca6832fa6b306ee6a7010bfb80a3f2596f824/include/llvm/ADT/APFloat.h#L832) of APFloat. As per the [requirement](https://en.cppreference.com/w/cpp/utility/variant) of std::variant, the first alternative of the variant  do have a default constructor, so not sure what exactly the problem was. I will explore it later. 

Hence using the following instead.

```
std::variant<APInt, APFloat, std::pair<APFloat, APFloat>>
```

docs/interpreter.md
burmako, line 20: Let's align to left? We should be able to do the following, which fits into 80 columns:

```
discriminated union holding one of `APInt` or `vector<APFloat>` for storage. The
later is used for storing elements with floating-point or complex types. Note
that we use `mlir::Attribute` while constructing an `Element` object (via
`Element(Type type, Attribute attr)`) or extracting value from `Element` object 
(via `Attribute Element::getValue()`), in order to simplify these interfaces.
```

stablehlo/reference/Element.cpp
```diff
@@ -67,51 +67,25 @@ bool isSupportedComplexType(Type type) {
...
   auto type = getType();
-  assert(type == other.getType());
 
-  if (isSupportedFloatType(type)) {
-    auto left = getFloatValue(*this);
-    auto right = getFloatValue(other);
-    return Element(type, FloatAttr::get(type, left + right));
-  }
+  if (isSupportedFloatType(type))
+    return Element(type, this->getFloatValue() + other.getFloatValue());
```
burmako, line 76: Just `getFloatValue()` (without `this`) would be sufficient.

stablehlo/reference/Element.cpp
```diff
@@ -153,12 +123,46 @@ Element sine(const Element &e) {
...
+
+  if (isSupportedSignedIntegerType(type_)) {
+    convertToIntAttr(type_, getIntegerValue()).print(os);
+    return;
+  }
+
+  if (isSupportedUnsignedIntegerType(type_)) {
+    convertToIntAttr(type_, getIntegerValue()).print(os);
+    return;
+  }
```
burmako, line 145: This can be simplified to:

```
if (isSupportedIntegerType(type_)) {
  IntegerAttr::get(type, getIntegerValue()).print(os);
}
```

stablehlo/reference/Element.cpp
```diff
@@ -153,12 +123,46 @@ Element sine(const Element &e) {
...
 
-void Element::dump() const {
-  print(llvm::errs());
-  llvm::errs() << "\n";
+  if (isSupportedFloatType(type_)) {
+    convertToFloatAttr(type_, getFloatValue()).print(os);
```
burmako, line 148: I don't think that `convertToFloatAttr` buys us anything here. Let's just inline the implementation.

stablehlo/reference/Element.cpp
```diff
@@ -153,12 +123,46 @@ Element sine(const Element &e) {
...
+  if (isSupportedComplexType(type_)) {
+    auto complexTy = type_.dyn_cast<mlir::ComplexType>().getElementType();
+    auto complexVal = getComplexValue();
+
+    os << "[";
+    convertToFloatAttr(complexTy, complexVal.real()).print(os);
+    os << ", ";
+    convertToFloatAttr(complexTy, complexVal.imag()).print(os);
+    os << "]";
+    return;
```
burmako, line 161: Can we use `complex::NumberAttr`'s prettyprinting here?

stablehlo/tests/interpret_add.mlir
```diff
@@ -255,6 +255,6 @@ func.func @add_op_test_c128() -> tensor<2xcomplex<f64>> {
   %2 = stablehlo.add %0, %1 : tensor<2xcomplex<f64>>
   func.return %2 : tensor<2xcomplex<f64>>
   // CHECK-NEXT: tensor<2xcomplex<f64>>
-  // CHECK-NEXT: [3.000000e+00, 5.000000e+00]
-  // CHECK-NEXT: [1.500000e+01, 1.100000e+01]
+  // CHECK-NEXT: [3.000000e+00 : f64, 5.000000e+00 : f64]
+  // CHECK-NEXT: [1.500000e+01 : f64, 1.100000e+01 : f64]
```
burmako, line 259: Hmm I can't say that I love this change. I think it just adds boilerplate without buying us much?

stablehlo/reference/Element.h
```diff
@@ -27,7 +30,10 @@ class Element {
  public:
   /// \name Constructors
   /// @{
-  Element(Type type, Attribute value) : type_(type), value_(value) {}
+  Element(Type type, APInt val) : type_(type), value_(val) {}
+  Element(Type type, APFloat val) : type_(type), value_(val) {}
+  Element(Type type, std::complex<APFloat> val)
+      : type_(type), value_(std::make_pair(val.real(), val.imag())) {}
```
burmako, line 36: How about renaming `val` to `value` to be consistent with `value_` and `getFooValue()`?

stablehlo/reference/Element.h
```diff
@@ -38,8 +44,12 @@ class Element {
   /// Returns type of the Element object.
   Type getType() const { return type_; }
 
-  /// Returns the underlying storage of Element object.
```
burmako, line 41: This comment seems to have been lost in the refactoring.

stablehlo/reference/Tensor.cpp
```diff
@@ -50,6 +48,27 @@ int64_t getSizeInBytes(Type type) {
...
+  // integer variants.
+  return type.isSignlessInteger(4) || type.isSignlessInteger(8) ||
+         type.isSignlessInteger(16) || type.isSignlessInteger(32) ||
+         type.isSignlessInteger(64);
+}
+
+bool isSupportedIntegerType(Type type) {
+  return isSupportedUnsignedIntegerType(type) ||
+         isSupportedSignedIntegerType(type);
+}
```
burmako, line 70: Rather than duplicating these functions, I propose we move them to `namespace detail` in `Element.h` (including the `isSupportedFloatType` and `isSupportedComplexType`.

stablehlo/reference/Element.h
```diff
@@ -52,7 +62,7 @@ class Element {
 
  private:
   Type type_;
-  Attribute value_;
+  std::variant<APInt, APFloat, std::pair<APFloat, APFloat>> value_;
```
burmako, line 65: Why can't this be `std::variant<APInt, APFloat, std::complex<APFloat>>`? This works for me.

```
diff --git a/stablehlo/reference/Element.h b/stablehlo/reference/Element.h
index dfc454eca2..61f07ab01c 100644
--- a/stablehlo/reference/Element.h
+++ b/stablehlo/reference/Element.h
@@ -32,8 +32,7 @@ class Element {
   /// @{
   Element(Type type, APInt val) : type_(type), value_(val) {}
   Element(Type type, APFloat val) : type_(type), value_(val) {}
-  Element(Type type, std::complex<APFloat> val)
-      : type_(type), value_(std::make_pair(val.real(), val.imag())) {}
+  Element(Type type, std::complex<APFloat> val) : type_(type), value_(val) {}

   /// @}
@@ -47,8 +46,7 @@ class Element {
   APFloat getFloatValue() const { return std::get<APFloat>(value_); }
   APInt getIntegerValue() const { return std::get<APInt>(value_); }
   std::complex<APFloat> getComplexValue() const {
-    auto floatPair = std::get<std::pair<APFloat, APFloat>>(value_);
-    return std::complex<APFloat>(floatPair.first, floatPair.second);
+    return std::get<std::complex<APFloat>>(value_);
   }

   /// Overloaded + operator.
@@ -62,7 +60,7 @@ class Element {

  private:
   Type type_;
-  std::variant<APInt, APFloat, std::pair<APFloat, APFloat>> value_;
+  std::variant<APInt, APFloat, std::complex<APFloat>> value_;
 };
```

burmako (changes requested):

stablehlo/reference/Element.h
```diff
@@ -27,7 +30,10 @@ class Element {
  public:
```
joker-eph, line 30: Not directly related to this CL, but classes should be documented IMO

docs/interpreter.md
```diff
@@ -12,9 +12,12 @@ representing its data laid out in
 objects are reference-counted to simplify memory management.
 
 Individual elements of a tensor are represented using `Element` class which uses
-`mlir::Attribute` for storage. Using `mlir::Attribute` simplifies things because
-this means that we don't have to implement our own machinery for storing values
-of different types inside `Element`.
+discriminated union holding one of `APInt` or `vector<APFloat>` for storage. The
+later is used for storing elements with floating-point or complex types. Note
+that we use `mlir::Attribute` while constructing an `Element` object (via
+    `Element(Type type, Attribute attr)`) or extracting value from `Element`
```
joker-eph, line 18: This CL is actually removing this constructor, isn't it?

docs/interpreter.md
```diff
@@ -12,9 +12,12 @@ representing its data laid out in
...
+object (via `Attribute Element::getValue()`), in order to simplify these
```
joker-eph, line 19: Same for this accessor if I see correctly?

joker-eph (approved):
Overall seems like we're going in the right directly by removing dependencies on the MLIRContext for storage of things we can avoid to

docs/interpreter.md
sdasgup3: That is right. updated the document.

stablehlo/reference/Element.cpp
sdasgup3: done.

stablehlo/reference/Tensor.cpp
sdasgup3: Moving the definition to the include file, which gets included in other cpp file, gives link time error  related to duplicate symbol. Am I missing something here?

stablehlo/reference/Element.h
sdasgup3: Strange that I am still getting the same error with the above change.

```
In file included from /usr/local/google/home/sdasgup/Github/stablehlo/stablehlo/reference/Ops.cpp:16:                                                                                                                           
In file included from /usr/local/google/home/sdasgup/Github/stablehlo/stablehlo/reference/Ops.h:19:                                                                                                                             
In file included from /usr/local/google/home/sdasgup/Github/stablehlo/llvm-project/mlir/include/mlir/IR/BuiltinAttributes.h:12:                                                                                                 
In file included from /usr/local/google/home/sdasgup/Github/stablehlo/llvm-project/mlir/include/mlir/IR/BuiltinAttributeInterfaces.h:19:                                                                                        
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/complex:135:75: error: calling a private constructor of class 'llvm::APFloat'                                                                                
      _GLIBCXX_CONSTEXPR complex(const _Tp& __r = _Tp(), const _Tp& __i = _Tp())                                                                                                                                                
                                                                          ^                                                                                                                                                     
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/variant:774:36: note: in instantiation of default function argument expression for 'complex<llvm::APFloat>' required here                                    
                      void_t<decltype(_Arr<_Ti>{{std::declval<_Tp>()}})>>                                                                                                                                                       
                                                 ^                                                                                                                                                                              
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/variant:786:7: note: during template argument deduction for class template partial specialization '_Build_FUN<_Ind, _Tp, _Ti>' [with _Ind = 2, _Tp = llvm::AP
Float &, _Ti = std::complex<llvm::APFloat>]                                                                                                                                                                                     
    : _Build_FUN<_Ind, _Tp, _Ti>...                                                                                                                                                                                             
      ^                                                                                                                                                                                                                         
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/variant:786:7: note: in instantiation of template class 'std::__detail::__variant::_Build_FUN<2, llvm::APFloat &, std::complex<llvm::APFloat>>' requested her
e                                                                                                                                                                                                                               
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/variant:795:18: note: in instantiation of template class 'std::__detail::__variant::_Build_FUNs<llvm::APFloat &, std::variant<llvm::APInt, llvm::APFloat, std
::complex<llvm::APFloat>>, std::integer_sequence<unsigned long, 0, 1, 2>>' requested here                                                                                                                                       
      = decltype(_Build_FUNs<_Tp, _Variant>::_S_fun(std::declval<_Tp>()));                                                                                                                                                      
                 ^                                                                                                                                                                                                              
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/variant:804:51: note: in instantiation of template type alias '_FUN_type' requested here                                                                     
    struct __accepted_index<_Tp, _Variant, void_t<_FUN_type<_Tp, _Variant>>>                                                                                                                                                    
                                                  ^                                                                                                                                                                             
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/variant:1376:27: note: (skipping 1 context in backtrace; use -ftemplate-backtrace-limit=0 to see all)                                                        
          = __detail::__variant::__accepted_index<_Tp, variant>::value;                                                                                                                                                         
                                 ^                                                                                                                                                                                              
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/variant:1376:27: note: in instantiation of template class 'std::__detail::__variant::__accepted_index<llvm::APFloat &, std::variant<llvm::APInt, llvm::APFloa
t, std::complex<llvm::APFloat>>>' requested here                                                                                                                                                                                
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/variant:1382:36: note: in instantiation of static data member 'std::variant<llvm::APInt, llvm::APFloat, std::complex<llvm::APFloat>>::__accepted_index<llvm::
APFloat &>' requested here                                                                                                                                                                                                      
        using __accepted_type = __to_type<__accepted_index<_Tp>>;                                                                                                                                                               
                                          ^                                                                                                                                                                                     
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/variant:1412:24: note: in instantiation of template type alias '__accepted_type' requested here                                                              
               typename _Tj = __accepted_type<_Tp&&>,                                                                                                                                                                           
                              ^                                                                                                                                                                                                 
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/variant:1416:2: note: in instantiation of default argument for 'variant<llvm::APFloat &, void, void>' required here                                          
        variant(_Tp&& __t)                                                                                                                                                                                                      
        ^~~~~~~~~~~~~~~~~~                                                                                                                                                                                                      
/usr/local/google/home/sdasgup/Github/stablehlo/stablehlo/reference/Element.h:35:59: note: while substituting deduced template arguments into function template 'variant' [with _Tp = llvm::APFloat &, $1 = (no value), $2 = (no
 value), _Tj = (no value), $4 = (no value)]                                                                                                                                                                                     
  Element(Type type, APFloat value) : type_(type), value_(value) {}                                                                                                                                                             
                                                          ^                                                                                                                                                                     
/usr/local/google/home/sdasgup/Github/stablehlo/llvm-project/llvm/include/llvm/ADT/APFloat.h:844:3: note: implicitly declared private here                                                                                      
  APFloat() : U(IEEEdouble()) {                                                                                                                                                                                                 
  ^                             
```

stablehlo/reference/Element.h
sdasgup3: Point noted! Will make sure to document the classes `Tensor, Element, IndexSpaceIterator`.

sdasgup3: done.

stablehlo/reference/Tensor.h
```diff
@@ -61,7 +61,8 @@ class Buffer : public llvm::RefCountedBase<Buffer> {
 
 }  // namespace detail
 
-/// Helper class to access the tensor elements in a linearized layout.
+/// Class to model a tensor, an n-dimensional array. Provide access to
+/// individual elements of the tensor using n-dimensional indices.
```
joker-eph, line 65: This is missing some important aspects (lifetime, mutability, etc.)

stablehlo/reference/Element.h
burmako: As promised, here's a pull request that compiles on my machine. Can you take a look if it compiles on yours? https://github.com/sdasgup3/stablehlo/pull/2.

stablehlo/reference/Element.h
```diff
@@ -38,8 +48,20 @@ class Element {
   /// Returns type of the Element object.
   Type getType() const { return type_; }
 
-  /// Returns the underlying storage of Element object.
-  Attribute getValue() const { return value_; }
+  /// Returns the underlying integer value stored in an Element object with
+  /// integer type.
```
burmako, line 52: What happens if it's we call `getFloatValue` on an integer Element?

stablehlo/reference/Element.h
```diff
@@ -38,8 +48,20 @@ class Element {
   /// Returns type of the Element object.
   Type getType() const { return type_; }
 
-  /// Returns the underlying storage of Element object.
-  Attribute getValue() const { return value_; }
+  /// Returns the underlying integer value stored in an Element object with
```
burmako, line 51: Typo: "floating-point value".

stablehlo/reference/Element.h
```diff
@@ -38,8 +48,20 @@ class Element {
...
+  /// integer type.
+  APFloat getFloatValue() const { return std::get<APFloat>(value_); }
+
+  /// Returns the underlying floating-point value stored in an Element object
```
burmako, line 55: Typo: "integer value".

stablehlo/reference/Tensor.cpp
burmako: That's right. You'll need to move the declarations to `Element.h` and the definitions to `Element.cpp`. After that, you'll be able to delete these functions from `Tensor.cpp`.

stablehlo/reference/Element.h
burmako: Interesting! What compiler are you using? Here's a pull request that works on my machine: https://github.com/sdasgup3/stablehlo/pull/2 - just so that we're sure we're looking at the same code.

stablehlo/tests/interpret_add.mlir
```diff
@@ -242,8 +242,8 @@ func.func @add_op_test_c64() -> tensor<2xcomplex<f32>> {
   %2 = stablehlo.add %0, %1 : tensor<2xcomplex<f32>>
   func.return %2 : tensor<2xcomplex<f32>>
   // CHECK-NEXT: tensor<2xcomplex<f32>>
-  // CHECK-NEXT: [3.000000e+00 : f32, 5.000000e+00 : f32]
-  // CHECK-NEXT: [1.500000e+01 : f32, 1.100000e+01 : f32]
+  // CHECK-NEXT: #complex.number<:f32 3.000000e+00, 5.000000e+00> : complex<f32>
+  // CHECK-NEXT: #complex.number<:f32 1.500000e+01, 1.100000e+01> : complex<f32>
```
burmako, line 246: Hmmm that's a mouthful... I'm sorry for providing conflicting feedback, but can we get back to `[3.000000e+00 : f32, 5.000000e+00 : f32]`? Or even better - `[3.000000e+00, 5.000000e+00]` - but that can wait until a follow-up PR.

stablehlo/reference/Element.cpp
```diff
@@ -200,42 +171,59 @@ Element Element::operator+(const Element &other) const {
...
   return mapWithUpcastToDouble(
-      e, [](double e) { return std::tanh(e); },
+      el, [](double e) { return std::tanh(e); },
       [](std::complex<double> e) { return std::tanh(e); });
 }
 
-void Element::print(raw_ostream &os) const { value_.print(os); }
+void Element::print(raw_ostream &os) const {
+  if (isSupportedIntegerType(type_)) {
+    IntegerAttr::get(type_, getIntegerValue()).print(os);
```
burmako, line 206: To be honest, I think we should print attributes with `/*elideType=*/ true` to make things easier on the eyes, but I can make this change in a separate PR when this PR lands.

stablehlo/reference/Element.cpp
```diff
@@ -200,42 +171,59 @@ Element Element::operator+(const Element &other) const {
...
+  }
+
+  if (isSupportedFloatType(type_)) {
+    FloatAttr::get(type_, getFloatValue()).print(os);
+    return;
+  }
 
-void Element::dump() const {
-  print(llvm::errs());
-  llvm::errs() << "\n";
```
burmako, line 237: Any particular reason to remove `"\n"`? E.g. [Attribute's dump](https://github.com/llvm/llvm-project/blob/b947e15a577f72284d69d077cfdb2d3f211abe65/mlir/lib/IR/AsmPrinter.cpp#L3282-L3285) also prints `"\n"`.

burmako (changes requested):

stablehlo/reference/Element.h
sdasgup3:
```
Debian clang version 13.0.1-7
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

Unfortunately the error persists.

stablehlo/reference/Element.h
sdasgup3: Made changes so that we will throw a fatal error for those cases.

stablehlo/tests/interpret_add.mlir
sdasgup3: makes sense. Reverted.

stablehlo/reference/Element.h
sdasgup3: Let us take this up in a [new ticket](https://github.com/openxla/stablehlo/issues/226)

stablehlo/reference/Tensor.h
sdasgup3: I agree. Let is take this up in  https://github.com/openxla/stablehlo/pull/117#issuecomment-1262869759

stablehlo/reference/Element.cpp
sdasgup3: Thanks!

stablehlo/reference/Element.h
```diff
@@ -79,6 +98,13 @@ inline raw_ostream &operator<<(raw_ostream &os, Element element) {
   return os;
 }
 
+/// Check if the type 'type' comforms with what is supported in StableHLO.
```
burmako, line 101: Typo: "conforms".

stablehlo/reference/Element.h
```diff
@@ -79,6 +98,13 @@ inline raw_ostream &operator<<(raw_ostream &os, Element element) {
...
+bool isSupportedUnsignedIntegerType(Type type);
+bool isSupportedSignedIntegerType(Type type);
+bool isSupportedIntegerType(Type type);
+bool isSupportedFloatType(Type type);
+bool isSupportedComplexType(Type type);
```
burmako, line 106: Originally, we discussed putting these functions in the `mlir::stablehlo::detail` namespace similarly to where we put `Buffer` (since they aren't e.g. supposed to be used in Ops.cpp). What do you think about that?

stablehlo/reference/Element.h
burmako, line 101: "with what is supported in the StableHLO spec"?

burmako (changes requested):
One more minor thing, and we're good to go.

burmako (approved):
I'll address the namespace feedback in a follow-up pull request later today. Thank you for all the work on the review feedback!

PR #104, opened 2022-09-09
Add Python API Testing to GitHub Actions

### LINKED ISSUE #62: Add python bindings test to GitHub Actions
### Request description

Per feedback in #49 - Add tests for python API to CI workflows.

https://github.com/openxla/stablehlo/tree/main/build_tools#python-api

### Additional context

_No response_

GleasonK:
- Added build script for Python API `ci_build_stablehlo_api.sh`
- Added a github action to build and check Python API.
- Added to `buildAndTest.yaml`. 
- Modified `setup-build` to also install required python libraries for mlir bindings.

Timing (first time cost any time we bump llvm version):
- First time cost: ~1.5hr 
- With ccache, following builds: ~3min

_Open question:_
I have a feeling a lot of common LLVM machinery is being built twice. Is there a way to combine API bindings build with the LLVM build? Do we want to do this?

Closes #62.

### DIFF
--- a/.github/actions/setup-build/action.yml
+++ b/.github/actions/setup-build/action.yml
@@ -19,7 +19,7 @@ runs:
     with:
       repository: llvm/llvm-project
       ref: ${{ inputs.llvm-version }}
-      path: llvm-project
+      path: llvm-project 
       fetch-depth: 1
 
   # Get ninja for cmake build.
@@ -39,3 +39,18 @@ runs:
     with:
       key: ${{ runner.os }}-stablehlo_build_assets-${{ inputs.llvm-version }}
       max-size: 4G
+
+  # Install Python/Numpy for API tests
+  - name: Install Python and Pip
+    uses: actions/setup-python@v4
+    with:
+      python-version: '3.10.6'
+      cache: 'pip' # caching pip dependencies
+  - name: Install MLIR python requirements
+    shell: bash
+    run: |
+      pwd
+      ls
+      ls llvm-project
+      ls $GITHUB_WORKSPACE/llvm-project
+      pip install -r $GITHUB_WORKSPACE/llvm-project/mlir/python/requirements.txt
--- a/.github/actions/setup-build/action.yml
+++ b/.github/actions/setup-build/action.yml
@@ -19,7 +19,7 @@ runs:
     with:
       repository: llvm/llvm-project
       ref: ${{ inputs.llvm-version }}
-      path: llvm-project 
+      path: llvm-project
       fetch-depth: 1
 
   # Get ninja for cmake build.
@@ -49,8 +49,4 @@ runs:
   - name: Install MLIR python requirements
     shell: bash
     run: |
-      pwd
-      ls
-      ls llvm-project
-      ls $GITHUB_WORKSPACE/llvm-project
       pip install -r $GITHUB_WORKSPACE/llvm-project/mlir/python/requirements.txt
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -25,10 +25,10 @@ concurrency:
 jobs:
   build-test:
     env:
-      LLVM_PROJ_DIR: "llvm-project"
+      LLVM_PROJECT_DIR: "llvm-project"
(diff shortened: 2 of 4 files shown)


.github/actions/setup-build/action.yml
```diff
@@ -19,7 +19,7 @@ runs:
     with:
       repository: llvm/llvm-project
       ref: ${{ inputs.llvm-version }}
-      path: llvm-project
+      path: llvm-project
```
burmako, line 22: Typo: trailing whitespace.

.github/actions/setup-build/action.yml
```diff
@@ -39,3 +39,18 @@ runs:
...
+    with:
+      python-version: '3.10.6'
+      cache: 'pip' # caching pip dependencies
+  - name: Install MLIR python requirements
+    shell: bash
+    run: |
+      pwd
+      ls
+      ls llvm-project
+      ls $GITHUB_WORKSPACE/llvm-project
```
burmako, line 55: Why do we need these commands?

.github/workflows/buildAndTest.yml
```diff
@@ -24,6 +24,11 @@ concurrency:
 # Use Cached LLVM to improve build times.
 jobs:
   build-test:
+    env:
+      LLVM_PROJ_DIR: "llvm-project"
```
burmako, line 28: How about expanding this to `LLVM_PROJECT_DIR` given that it only saves a few characters?

.github/workflows/buildAndTest.yml
```diff
@@ -24,6 +24,11 @@ concurrency:
...
+      LLVM_BUILD_DIR: "llvm-build"
+      STABLEHLO_BUILD_DIR: "stablehlo-build"
+      STABLEHLO_API_BUILD_DIR: "stablehlo-api-build"
```
burmako, line 31: Perhaps we can call this "stablehlo-python-build" (and rename related things accordingly), given that the focus of this build is on Python, not on the API? (We also have C and C++ APIs which are already provided by the regular build).

build_tools/github_actions/ci_build_stablehlo_api.sh
```diff
@@ -0,0 +1,43 @@
...
+if [[ $# -ne 3 ]] ; then
+  echo "Usage: $0 <llvm_project_dir> <stablehlo_api_build_dir> <stablehlo_root_dir>"
+  exit 1
+fi
+
+LLVM_PROJ_DIR="$1"
+STABLEHLO_API_BUILD_DIR="$2"
+STABLEHLO_ROOT_DIR="$3"
+
+# Configure StableHLO Bindings
```
burmako, line 24: "StableHLO Python Bindings"?

burmako (changes requested):
Thank you for the quick turnaround! Good to know that the build is reasonably quick. In the worst case, we can change it to a postsubmit in the future.

.github/actions/setup-build/action.yml
GleasonK: Ah. Debug code. Was hitting an issue where it couldn't find `stablehlo/llvm-project/` directory, and I added these to figure out what was going on. Turns out if you pass an environment variable as an input argument, it escapes the string. Was creating `stablehlo/\$GITHUB_WORKSPACE/\$LLVM_PROJ_DIR/llvm-project` instead.

Will remove.

build_tools/github_actions/ci_build_stablehlo_python_api.sh
```diff
@@ -0,0 +1,43 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+if [[ $# -ne 3 ]] ; then
+  echo "Usage: $0 <llvm_project_dir> <stablehlo_api_build_dir> <stablehlo_root_dir>"
+  exit 1
+fi
+
+LLVM_PROJ_DIR="$1"
```
burmako, line 20: LLVM_PROJECT_DIR here as well?

burmako (approved):

Issue #77, opened 2022-09-02
Add the specification of divide/rem


sdasgup3:
### Request description

Also clarify:
  * mhlo.divide: Does it means floor_divide (returns an integer less than or equal to the quotient) or mathematical divide (a floating point number)?
  * mhlo.remainder(x,y): Does it mean the floor_mod (which is x - floor_div(x/y)*y) or modulus (which is x - (x/y)*y) where the '/' operator returns the integer closest to zero)?

PR #106, opened 2022-09-09
Add spec of div/rem

sdasgup3:
fixes https://github.com/openxla/stablehlo/issues/77

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -388,9 +388,9 @@ specification. Numeric precision is implementation-defined.
 Performs element-wise division of dividend `lhs` and divisor `rhs` tensors and
 produces a `result` tensor. For floating-point element types, implements the
 `division` operation from IEEE-754 specification. For integer element types,
-  implements integer division truncating any fractional part. For n-bit integer
-  types, division overflow (division by zero or division of $-2^{n-1}$ with
-      $-1$) produces an implementation-defined value.
+implements integer division truncating any fractional part. For n-bit integer
+types, division overflow (division by zero or division of $-2^{n-1}$ with $-1$)
+produces an implementation-defined value.
 
 ### Operands
 
@@ -770,15 +770,14 @@ operation.
 
 Performs element-wise remainder of dividend `lhs` and divisor `rhs` tensors and
 produces a `result` tensor. The sign of the result is taken from the dividend,
-         and the absolute value of the result is always less than the divisor's
-         absolute value. The remainder of the any division operation `x/y` is
-         calculated as `x - n*y`, where `n` is `x/y` with its fractional part
-         truncated. For floating-point element types, this is in contrast with
-         the `remainder` operation from IEEE-754 specification where `n` is an
-         integral value nearest the exact value `x/y` with ties to even. For
-         floating-point types, the corner cases are TBD. For n-bit integer,
-         division overflow (remainder by zero or remainder of $-2^{n-1}$ with
-             $-1$) produces an implementation-defined value.
+and the absolute value of the result is always less than the divisor's absolute
+value. The remainder is calculated as `lhs - d * rhs`, where
+`d = stablehlo.divide`. For floating-point element types, this is in contrast
+with the `remainder` operation from IEEE-754 specification where `d` is an
+integral value nearest to the exact value of `lhs/rhs` with ties to even. For
+floating-point types, the corner cases are TBD. For n-bit integer, division
+overflow (remainder by zero or remainder of $-2^{n-1}$ with $-1$) produces an
+implementation-defined value.
 
 ### Operands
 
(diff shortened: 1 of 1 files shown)


GleasonK:
Not a part of this review, but was just reading through the spec from this review and noticed:

`stablehlo.xor` states: 
 > (C1) `operand` and `result` have the same type.
  
But `xor` does not have an operand named `operand`. Probably meant "`lhs`, `rhs`, and `result` have the same type." ?

GleasonK:
Also, again not part of this review, just noticing `stablehlo.log` seems to use a different constraint sentence structure:
> (C1) `result` must have the same type as that of `operand`.

GleasonK:
Opened #108 for those minor changes mentioned above.

docs/spec_draft.md
```diff
@@ -376,6 +378,51 @@ specification. Numeric precision is implementation-defined.
...
+operation from IEEE-754 specification. For n-bit integer types, division
+overflow (division by zero or division of $-2^{n-1}$ with $-1$) produces an
+implementation-defined value.
+
+### Operands
+
+| Name | Type |
+|-|-|
+| `dividend` | tensor of integer, floating-point or complex element types |
+| `divisor` | tensor of integer, floating-point or complex element types |
```
burmako, line 398: ODS uses names `lhs` and `rhs`. Let's stick to them? (Same for remainder).

docs/spec_draft.md
```diff
@@ -376,6 +378,51 @@ specification. Numeric precision is implementation-defined.
...
+
+`stablehlo.divide(dividend, divisor) -> result`
+
+### Semantics
+
+Performs element-wise division of tensors `dividend` and `divisor` and produces
+a `result` tensor. For floating-point element types, implements the `division`
+operation from IEEE-754 specification. For n-bit integer types, division
+overflow (division by zero or division of $-2^{n-1}$ with $-1$) produces an
+implementation-defined value.
```
burmako, line 391: If integer division needs rounding, which way does it go?

docs/spec_draft.md
```diff
@@ -670,6 +717,54 @@ operation.
...
+`stablehlo.remainder(dividend, divisor) -> result`
+
+### Semantics
+
+Performs element-wise remainder of tensors `dividend` and `divisor` and produces
+a `result` tensor. The sign of the result is taken from the dividend, and the
+absolute value of the result is always less than the divisor's absolute value.
+For floating-point element types, remainder of the any division operation `x/y`
+is calculated as `x - n*y`, where `n` is `x/y` with its fractional part
+truncated. The corner cases are TBD. For n-bit integer, division overflow
```
burmako, line 731: I think it would be useful to call out that this is different from IEEE-754 `remainder` and explain why.

burmako (changes requested):

docs/spec_draft.md
sdasgup3: I will rephrase both the divide and remainder specs to clarify what the they mean for integer types.

docs/spec_draft.md
```diff
@@ -670,6 +718,57 @@ operation.
...
+`stablehlo.remainder(lhs, rhs) -> result`
+
+### Semantics
+
+Performs element-wise remainder of dividend `lhs` and divisor `rhs` tensors and
+produces a `result` tensor. The sign of the result is taken from the dividend,
+         and the absolute value of the result is always less than the divisor's
+         absolute value. The remainder of the any division operation `x/y` is
+         calculated as `x - n*y`, where `n` is `x/y` with its fractional part
+         truncated. For floating-point element types, this is in contrast with
```
burmako, line 732: Consider rephrasing so that this uses the names introduced above, e.g. "The result calculated as \`lhs - d*rhs\`, where \`d = stablehlo.divide(lhs, rhs)\`".

docs/spec_draft.md
```diff
@@ -670,6 +718,57 @@ operation.
...
+         the `remainder` operation from IEEE-754 specification where `n` is an
+         integral value nearest the exact value `x/y` with ties to even. For
```
burmako, line 734: "nearest to the exact value"?

docs/spec_draft.md
```diff
@@ -670,6 +718,57 @@ operation.
...
+         floating-point types, the corner cases are TBD. For n-bit integer,
+         division overflow (remainder by zero or remainder of $-2^{n-1}$ with
+             $-1$) produces an implementation-defined value.
```
burmako, line 737: Also, formatting looks really wonky here and above. Is there a particular reason to not align to the left and fill the whitespace more evenly between lines?

burmako (changes requested):

docs/spec_draft.md
sdasgup3: The automatic vim wrap (applied using a switch). Fixed it for the two ops in this PR. I will fix the others in a separate one.

burmako (approved):

Issue #95, opened 2022-09-07
Decide on a policy for dealing with IEEE-754 exceptions


burmako:
The current thinking in spec_draft.md involves floating-point ops following IEEE-754 semantics. What does this mean as far as exception handling goes? As suggested offline by @joker-eph, [the LLVM Language Reference Manual](https://llvm.org/docs/LangRef.html) provides one way of handling this:

> ### Floating-Point Environment
>
>The default LLVM floating-point environment assumes that floating-point instructions do not have side effects. Results assume the round-to-nearest rounding mode. No floating-point exception state is maintained in this environment. Therefore, there is no attempt to create or preserve invalid operation (SNaN) or division-by-zero exceptions.
>
> The benefit of this exception-free assumption is that floating-point operations may be speculated freely without any other fast-math relaxations to the floating-point model.
>
> Code that requires different behavior than this should use the [Constrained Floating-Point Intrinsics](https://llvm.org/docs/LangRef.html#constrainedfp).

Here's how this floating-point environment is then used to defined floating-point operations in LLVM.

> ### ‘fadd’ Instruction
>
> Syntax:
> <result> = fadd [fast-math flags]* <ty> <op1>, <op2>   ; yields ty:result
>
> Semantics:
> The value produced is the floating-point sum of the two operands. This instruction is assumed to execute in the default [floating-point environment](https://llvm.org/docs/LangRef.html#floatenv). This instruction can also take any number of [fast-math flags](https://llvm.org/docs/LangRef.html#fastmath), which are optimization hints to enable otherwise unsafe floating-point optimizations:

sdasgup3:
My proposal is to comply with the HLO's stance of no-exception-handling, i.e.,  no exception state maintained in the execution environment.

PR #111, opened 2022-09-12
Clarification on floating-point exception env.

sdasgup3:
fixes https://github.com/openxla/stablehlo/issues/95

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -56,11 +56,22 @@ implementation-defined, as long as ops are evaluated before their uses. Possible
 execution orders of the above example program are `%1` → `%2` → `%3` → `return`
 or `%2` → `%1` → `%3` → `return`.
 
-### Floating-point Exceptions
+### Errors
 
-The execution envorinment of StableHLO does not maintain any exception state.
-Therefore, there is no attempt to create or preserve invalid operation,
-  division-by-zero, overflow, underflow, or inexact exceptions.
+StableHLO programs are validated through an extensive set of constraints for
+individual ops, which rules out many classes of errors prior to run time.
+However, error conditions are still possible, e.g. through integer overflows,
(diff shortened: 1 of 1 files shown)


docs/spec_draft.md
```diff
@@ -56,6 +56,12 @@ implementation-defined, as long as ops are evaluated before their uses. Possible
 execution orders of the above example program are `%1` → `%2` → `%3` → `return`
 or `%2` → `%1` → `%3` → `return`.
 
+### Floating-point Exceptions
+
+The execution envorinment of StableHLO does not maintain any exception state.
+Therefore, there is no attempt to create or preserve invalid operation,
+  division-by-zero, overflow, underflow, or inexact exceptions.
+
```
burmako, line 64: How about making this more general and taking this chance to talk about errors? Not sure that's necessarily the right word to use, but let's iterate from there? For example:

## Errors

StableHLO programs are validated through an extensive set of constraints for individual ops, which rules out many classes of errors prior to run time. However, error conditions are still possible, e.g. through integer overflows, out-of-bounds accesses, etc. Unless explicitly called out, all these errors result in implementation-defined behavior. 

As an exception to this rule, floating-point exceptions in StableHLO programs have well-defined behavior. Operations which result in exceptions defined by the IEEE-754 standard (invalid operation, division-by-zero, overflow, underflow, or inexact exceptions) produce default results.

burmako (changes requested):
I think it's a good call to draft the spec using XLA's behavior for floating-point exceptions. When reviewing the spec, we can discuss this in more depth if necessary, but meanwhile this looks good as a first approximation.

docs/spec_draft.md
sdasgup3: Thanks! updated the text with some changes.

burmako (approved):

PR #112, opened 2022-09-12
Add test for constant op

### LINKED ISSUE #105: Add test for constant op

ghpvnist:
fixes #105

### DIFF
--- a/stablehlo/tests/interpret_constant.mlir
+++ b/stablehlo/tests/interpret_constant.mlir
@@ -2,30 +2,30 @@
 
 // CHECK-LABEL: Evaluated results of function: constant_op_test_i32
 func.func @constant_op_test_i32() -> tensor<2x3xi32> {
-  %0 = stablehlo.constant dense<[[-1, -1, -1], [-1, -1, -1]]>: tensor<2x3xi32>
+  %0 = stablehlo.constant dense<[[1, 2, 3], [4, 5, 6]]>: tensor<2x3xi32>
   func.return %0 : tensor<2x3xi32>
   // CHECK-NEXT: tensor<2x3xi32>
-  // CHECK-NEXT: -1 : i32
-  // CHECK-NEXT: -1 : i32
-  // CHECK-NEXT: -1 : i32
-  // CHECK-NEXT: -1 : i32
-  // CHECK-NEXT: -1 : i32
-  // CHECK-NEXT: -1 : i32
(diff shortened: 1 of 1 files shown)


stablehlo/tests/interpret_constant.mlir
```diff
@@ -0,0 +1,95 @@
...
+  // CHECK-NEXT: 3 : ui32
+  // CHECK-NEXT: 4 : ui32
+  // CHECK-NEXT: 5 : ui32
+  // CHECK-NEXT: 6 : ui32
+}
+
+// -----
+
+// CHECK-LABEL: Evaluated results of function: constant_op_test_bf16
+func.func @constant_op_test_bf16() -> tensor<bf16> {
```
sdasgup3, line 34: nit: Should we use the same values from interpret_add.mlir for each type? That way the only change will be in the operator (add vs constant) and the CHECK results. Just to be consistent.  Here and below.

Also, how about including the tests for all the supported integer types with values borrowed from add tests?

stablehlo/tests/interpret_constant.mlir
```diff
@@ -0,0 +1,95 @@
+// RUN: stablehlo-interpreter --interpret -split-input-file %s | FileCheck %s
```
sdasgup3, line 1: Can we use the same order of tupes as used in interpret_add.mlir ? First floating point types in increasing order of bitwidth and then integer with increasing order of bitwidth, and lastly the complex types? Just for consistency.

ghpvnist: Speaking of consistency, should we follow the order--integer, boolean, float, and complex--to match how the spec types are ordered? I will make another PR to fix `interpret_add.mlir` file along with any typos/grammar errors in the `spec_draft.md`.

sdasgup3: That should be fine as long as we stick to the same order. 
>>  I will make another PR to fix interpret_add.mlir file 
Makes sense. In this PR we could include all the tests (maybe using the same constant as in add test) and  a follow up PR can maintain the order for-all tests.

stablehlo/tests/interpret_constant.mlir
ghpvnist: Since ops can be unary or binary, I will borrow the special values from the lhs, rhs, and the result from `interpret_add.mlir`

sdasgup3 (approved):

PR #114, opened 2022-09-12
Add type inference for stablehlo.triangular_solve.

zhouxin913:


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -5004,56 +5004,60 @@ LogicalResult TransposeOp::inferReturnTypes(
...
 
-  // Result and argument b must have same shape.
-  auto resultType = getType().dyn_cast<RankedTensorType>();
-  if (!resultType) return success();
-  if (resultType != bType)
-    return emitOpError()
-           << "result and operand 'b' must have same shape, but got "
-           << resultType << " and " << bType;
+  inferredReturnShapes.emplace_back(bType.cast<ShapedType>());
   return success();
```
burmako, line 5061: It doesn't look like we're validating `transpose_a`? https://github.com/tensorflow/tensorflow/blob/0ad9b707d4404dee0e925f853f043edaf6f8a069/tensorflow/compiler/xla/service/shape_inference.cc#L2005-L2006

burmako (changes requested):

stablehlo/tests/infer_stablehlo.mlir
```diff
@@ -275,6 +276,16 @@ func.func @batch_norm_inference(%input: tensor<4x256xf32>, %scale: tensor<256xf3
   func.return %1 : tensor<4x256xindex>
 }
 
+// -----
+
+// CHECK-LABEL: func @triangular_solve
+func.func @triangular_solve(%arg0: tensor<10x5x4x4xf32>, %arg1: tensor<10x5x4x4xf32>) -> tensor<10x5x4x4xindex> {
+  %0 = "stablehlo.triangular_solve"(%arg0, %arg1) {left_side = true, lower = true, transpose_a = #stablehlo<transpose NO_TRANSPOSE>, unit_diagonal = true} : (tensor<10x5x4x4xf32>, tensor<10x5x4x4xf32>) -> tensor<10x5x4x4xf32>
```
GleasonK, line 283: Do we have code coverage for the branches in `inferReturnTypeComponents`, namely `aType` isn't ranked, and `aType` is ranked but `bType` isn't ranked?

zhouxin913: Good catch! I think that will hurt coverage, fixed.

burmako (approved):

PR #116, opened 2022-09-13
Add type inference for stablehlo.map.

zhouxin913:
The verifications are aligned with XLA semantics with only one difference: XLA [shape_inference.cc](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/shape_inference.cc#L1131) seems to allow scalar in inputs, but there isn't any test for it in XLA [shape_inference_test.cc](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/shape_inference_test.cc#L968). So no change in StableHLO: still only support tensor instead of scalars. 

Add a positive test and move all positive cases before negative ones.
Add support when 1) all inputs are unranked, or 2) some are unranked.

### DIFF
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -3357,84 +3357,95 @@ LogicalResult InfeedOp::verify() {
 // MapOp
 //===----------------------------------------------------------------------===//
 
-LogicalResult MapOp::verify() {
+LogicalResult MapOp::inferReturnTypeComponents(
+    MLIRContext*, Optional<Location> location, ValueShapeRange operands,
+    DictionaryAttr attributes, RegionRange regions,
+    SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
   // Checks if the number of `operands` match the arity of the map `computation`
   // region.
-  auto& computationBlock = computation().front();
+  MapOp::Adaptor adaptor(operands, attributes, regions);
+  auto& computationBlock = adaptor.computation().front();
   auto computationArgs = computationBlock.getArguments();
-  if (operands().size() != computationArgs.size())
-    return emitOpError() << "expects number of operands to match the arity "
-                            "of map computation, but got: "
-                         << operands().size() << " and "
-                         << computationArgs.size();
+  if (operands.size() != computationArgs.size())
+    return emitOptionalError(location,
+                             "expects number of operands to match the arity of "
+                             "map computation, but got: ",
+                             operands.size(), " and ", computationArgs.size());
 
   // The parameters of computation should all be scalars and match the element
   // type of operands.
   for (const auto& indexedArg : llvm::enumerate(computationArgs)) {
     auto argType = indexedArg.value().getType().dyn_cast<TensorType>();
     if (!argType || argType.getRank() != 0)
-      return emitOpError()
-             << "computation arguments must be 0-rank tensor, but got: arg #"
-             << indexedArg.index() << " of type "
-             << indexedArg.value().getType();
-    auto operandElemTy = operands()[indexedArg.index()]
+      return emitOptionalError(
+          location,
+          "computation arguments must be 0-rank tensor, but got: arg #",
+          indexedArg.index(), " of type ", indexedArg.value().getType());
+    auto operandElemTy = operands[indexedArg.index()]
                              .getType()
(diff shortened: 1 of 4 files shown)


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -3357,84 +3357,95 @@ LogicalResult InfeedOp::verify() {
...
+  if (operands.size() != computationArgs.size())
+    return emitOptionalError(location,
+                             "expects number of operands to match the arity of "
+                             "map computation, but got: ",
+                             operands.size(), " and ", computationArgs.size());
 
   // The parameters of computation should all be scalars and match the element
   // type of operands.
   for (const auto& indexedArg : llvm::enumerate(computationArgs)) {
     auto argType = indexedArg.value().getType().dyn_cast<TensorType>();
```
burmako, line 3378: This should be a cast to RankedTensorType. `getRank` on the next line will assert if called on an UnrankedTensorType: https://github.com/llvm/llvm-project/blob/1b56b2b2678cde21f7c20e83f881ded9b96518e4/mlir/include/mlir/IR/BuiltinTypeInterfaces.td#L136-L139. (Here and below).

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -3357,84 +3357,95 @@ LogicalResult InfeedOp::verify() {
...
+    if (operandType.hasRank()) {
+      if (dimensions.size() !=
+          static_cast<int64_t>(operandType.getShape().size()))
+        return emitOptionalError(
+            location,
+            "applied to a subset of dimensions currently not supported: "
+            "operand dimensions = ",
+            operandType.getShape().size(),
+            ", requested map dimensions size = ", dimensions.size());
+      resultShape = operandType.dyn_cast<ShapedType>().getShape();
```
burmako, line 3440: You don't need a cast here.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -3357,84 +3357,95 @@ LogicalResult InfeedOp::verify() {
...
+    }
   }
 
+  if (resultShape.empty())  // infer unranked when all inputs are unranked.
```
burmako, line 3444: This is not going to work for rank-0 operands. Instead, you can compute a flag that indicates whether the op has unranked operands and use that here.

burmako (changes requested):

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -3357,84 +3357,95 @@ LogicalResult InfeedOp::verify() {
...
-  auto& computationBlock = computation().front();
+  MapOp::Adaptor adaptor(operands, attributes, regions);
+  auto& computationBlock = adaptor.computation().front();
   auto computationArgs = computationBlock.getArguments();
-  if (operands().size() != computationArgs.size())
-    return emitOpError() << "expects number of operands to match the arity "
-                            "of map computation, but got: "
-                         << operands().size() << " and "
-                         << computationArgs.size();
+  if (operands.size() != computationArgs.size())
```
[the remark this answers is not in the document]
zhouxin913: Agreed.

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Done.

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Done.

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Good catch! Well the code works with all test because operands are not rank-0. Anywhere, current code is ambiguous between these 2 cases. Fixed.

burmako: Please also add a test for this. Otherwise LGTM.

burmako (approved):

stablehlo/dialect/StablehloOps.cpp
zhouxin913: There was a test there named "func.func @map_unranked(". Feel free to let me know if you were talking about something else?

burmako (approved):

Issue #3, opened 2022-08-17
Move decomposable StableHLO ops to CHLO


burmako:
In https://discourse.llvm.org/t/rfc-proposal-for-a-high-level-ml-dialect-in-mlir/64249/46, we did a survey of MHLO ops (almost of that applies to StableHLO, since we've bootstrapped it from MHLO) and identified the following ops which are decomposable to other MHLO ops: `batch_norm_grad`, `batch_norm_inference`, `batch_norm_training`, `broadcast`, `create_token`, `cross-replica-sum`, `dot`, `einsum`, `torch_index_select`, `trace`, `unary_einsum`. Let's move these ops to CHLO.

Note: In MHLO, `broadcast` and `dot` are referring to less general versions of `broadcast_in_dim` and `dot_general`. In HLO, `broadcast` and `dot` are used to refer to the more general versions, so we have a bit of confusing naming here which this move will also help to solve.

burmako:
In retrospect, this ticket was a bit ahead of its time:
  1) The fact that we haven't established a baseline specification prevented us from effectively planning the move. For example, as discussed above, it turned out that we cannot move `broadcast` to CHLO, because unranked broadcasts cannot be expressed via `broadcast_in_dim`.
  2) Our compatibility guarantees haven't yet been finalized (see #115), so it's unclear what exactly this move will entail. E.g. are we going to keep both `chlo.broadcast` and `stablehlo.broadcast` around? For how long? How do we migrate users?
  3) We don't yet have an RFC process to reason about changes to the StableHLO opset. #196 was only introduced a few days ago and will still need some additional work.

As a result, we'll be focusing on completing the StableHLO v1.0 spec in the next few months, and by then we expect that all the blockers mentioned above will be resolved. Until then, I'll be unassigning this ticket.

burmako:
Refactored this ticket to focus on ops which don't have equivalents in the HLO opset (not counting dynamism ops which are handled separately in #8). BatchNorm ops which were previously part of this ticket are now discussed in #603.

PR #118, opened 2022-09-14
Move CrossReplicaSum from StableHLO to CHLO.

subhankarshah:
Part 1 of Many
 - Create a copy of stablehlo.cross-replica-sum in CHLO, including changes in .td and .cc files.

CrossReplicaSum is a special case of AllReduce where the reduction is a summation operation.
It is therefore decomposable to AllReduce. We intend to move the op to CHLO and lower it to AllReduce.


stablehlo/dialect/ChloOps.td
```diff
@@ -945,4 +945,27 @@ def CHLO_DynamicReshapeOp: CHLO_Op<"dynamic_reshape", [NoSideEffect,
   let results = (outs HLO_Tensor:$result);
 }
 
+def CHLO_CrossReplicaSumOp : CHLO_Op<"cross-replica-sum",
```
GleasonK, line 948: Now that we have a (somewhat) better idea of what compatibility maintenance may look like (#115), I think it would be better to rename this to `cross_replica_sum`.

We want to eventually rename this because (1) it is the only op that uses hyphens in the mnemonic, and (2) hyphenated ops are not supported by pretty printing.

I think this PR is a good time to do this, there is no compatibility impact until we migrate to start using this CHLO op. And if we did the change separately (`mhlo.cross-replica-sum --> chlo.cross-replica-sum --> chlo.cross_replica_sum`), there are two points of migration that we will need to support for the duration of the forwards/backwards compatibility window.

Doing this all at once also has the advantage of only needing to update test files once, after we start migrating to use this op.

GleasonK (changes requested):

burmako (changes requested):
Thank you for your contribution! Since this PR proposes to modify the opset, it'll need to go through the RFC process before it can get merged. I'm finishing setting up the process and will follow up here later this week.

Also, I wanted to document our in-person discussion about what happens next if/when this pull request is approved. This is how I propose this should look like.

(M0) Current state:
  * There's stablehlo.cross-replica-sum.
  * There's mhlo.cross-replica.sum.
  * stablehlo.cross-replica-sum => [stablehlo-legalize-to-hlo] => mhlo.cross-replica.sum.
  * mhlo.cross-replica-sum => [hlo-legalize-to-stablehlo] => stablehlo.cross-replica.sum.

(M1) After this pull request (and a pending follow-on in MLIR-HLO):
  * There's chlo.cross-replica-sum.
  * chlo.cross-replica-sum => [chlo-legalize-to-hlo] => mhlo.all_reduce.
  * There's stablehlo.cross-replica-sum.
  * There's mhlo.cross-replica.sum.
  * stablehlo.cross-replica-sum => [stablehlo-legalize-to-hlo] => mhlo.cross-replica.sum.
  * mhlo.cross-replica-sum => [hlo-legalize-to-stablehlo] => stablehlo.cross-replica.sum.

(M2) Clean up MHLO (should be done shortly afterwards, because we don't provide stability guarantees for MHLO):
  * There's chlo.cross-replica-sum.
  * chlo.cross-replica-sum => [chlo-legalize-to-hlo] => mhlo.all_reduce.
  * There's stablehlo.cross-replica-sum.
  * stablehlo.cross-replica-sum => [stablehlo-legalize-to-hlo] => mhlo.all_reduce.

(M3) Clean up StableHLO (need to wait for the backward compatibility window)
  * There's chlo.cross-replica-sum.
  * chlo.cross-replica-sum => [chlo-legalize-to-hlo] => mhlo.all_reduce.

GleasonK:
From a compatibility perspective, [Migration Change](https://github.com/openxla/stablehlo/blob/68b5c357de519649dfef0561e4bb813bd0191aca/docs/compatibility.md#migration-change). We will need to keep both Ops around, unless we figure out a way to use a generic `Operation *` for upgrade pass, before verifiers are run. I'll be investigating this in the near future.

(M0)
- Current state, no impact.

(M1)
- Verifier (pre-downgrade) blocks serialization of stablehlo.cross-replica-sum
- chlo.cross_replica_sum => [downgrade] => stablehlo.cross-replica-sum (remove after forward compatibility window)
- stablehlo.cross-replica-sum => [upgrade] => chlo.cross_replica_sum

(M2) 
- No impact.

(M3)
- Remove upgrade pass, delete stablehlo.cross-replica-sum

burmako:
Closing this pull request for now as discussed in https://github.com/openxla/stablehlo/issues/3#issuecomment-1263057416. Will revisit once we pick up the associated ticket again.

PR #124, opened 2022-09-15
Add TableGen targets to mlir-headers

burmako:
Without this change, users of dialect headers will have to remember to add these targets to their dependencies, which can be easily forgotten and can lead to hard to reproduce CMake build errors. A similar issue in the MLIR-HLO repository only showed with `ninja -j1`.


GleasonK (approved):
@burmako asked for review - This fix lgtm, in line with the mlir-hlo fix that was recently merged to TF.

PR #127, opened 2022-09-15
Add specification for exp operator

### LINKED ISSUE #99: Add the specification of exp operator

ghpvnist:
fixes #99

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -452,8 +452,8 @@ produces an implementation-defined value.
 Performs element-wise exponential operation on `operand` tensor and produces a
 `result` tensor. For floating-point element types, it implements the `exp`
 operation from the IEEE-754 specification. For complex element types, it
-computes a complex form of exponential using Euler's formula, with corner cases
-TBD. Numeric precision is implementation-defined.
+computes a complex exponential, with corner cases TBD. Numeric precision is
+implementation-defined.
 
 ### Operands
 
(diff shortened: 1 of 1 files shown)


docs/spec_draft.md
```diff
@@ -442,6 +443,48 @@ produces an implementation-defined value.
...
+## stablehlo.exp
+
+`stablehlo.exp(operand) -> result`
+
+### Semantics
+
+Performs element-wise exponential operation on `operand` tensor and produces a
+`result` tensor. For floating-point element types, it implements the `exp`
+operation from the IEEE-754 specification. For complex element types, it
+computes a complex form of exponential using Euler's formula, with corner cases
```
burmako, line 455: Use a simplified version of this consistent with other ops, e.g. "For complex element types, it computes a complex exponential, with corner cases TBD"?

burmako (approved):

docs/spec_draft.md
sdasgup3: As the Euler's formal is taking care of the imaginary part only as
image

How about we explicit write our formula as
$e^{a+ix} = e^a . e^{ix}$ where $e^{ix}$ is the complex exponential and the multiplication is the `multiplication` operation from IEEE-754 standard.

sdasgup3: After looking into https://en.cppreference.com/w/cpp/numeric/complex/exp, I realized that "complex exponentiation function" is not can mean $e^{a+ix}$ as well. 

Let me take back my comment.

sdasgup3 (approved):

PR #131, opened 2022-09-15
Remove mlir_runner_utils_dir replacement from lit.cfg.py

burmako:
This is going away (or has already gone away) in MLIR upstream.
Context: https://reviews.llvm.org/D133270.


GleasonK (approved):
LGTM

PR #133, opened 2022-09-15
Backward compat with the old way of registering CHLO

burmako:
We aren't promising source compatibility of our APIs yet, but this one is easy to provide, and it was useful for downstream users of MLIR-HLO.


GleasonK (approved):
LGTM

PR #135, opened 2022-09-15
Add type inference for stablehlo.if stablehlo.case.

zhouxin913:
This PR includes both for `if` and `case` as they share similar implementation.
The spec of these ops are no ready, and no XLA implementation for type inference of those ops found to compare. So current implementation is only an optimization of existing implementation.


stablehlo/dialect/StablehloOps.td
```diff
@@ -1118,15 +1119,25 @@ def StableHLO_IfOp: StableHLO_Op<"if", [
...
+  let extraClassDeclaration = [{
+    // Method from InferTypeOpInterface interface.
+    static bool isCompatibleReturnTypes(TypeRange l, TypeRange r) {
+      if (l.size() != r.size()) return false;
+      for (auto [lt, rt] : llvm::zip(l, r))
+        if (!mlir::hlo::isCompatibleForHloTypeInference(lt, rt))
+          return false;
+      return true;
+    }
+  }];
```
burmako, line 1131: It'd be good to avoid copy/pasting these things over and over again, but let's discuss that in #145. LGTM for this pull request.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -4660,48 +4660,80 @@ LogicalResult ReplicaIdOp::inferReturnTypes(
...
-  for (unsigned i = 0; i < numBranches; ++i)
-    if (failed(verifyConditionalBranch(*this, branches()[i],
-                                       /*branchName=*/"branch " + Twine(i))))
+LogicalResult CaseOp::inferReturnTypeComponents(
+    MLIRContext*, Optional<Location> location, ValueShapeRange operands,
+    DictionaryAttr attributes, RegionRange regions,
+    SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
+  CaseOp::Adaptor adaptor(operands, attributes, regions);
+  ValueTypeRange<OperandRange> branch0ResultTypes =
+      adaptor.branches()[0]->front().getTerminator()->getOperandTypes();
```
burmako, line 4721: We should check that the number of branches is non-zero.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -4660,48 +4660,80 @@ LogicalResult ReplicaIdOp::inferReturnTypes(
...
+bool isCompatibleForTypeRange(ValueTypeRange<OperandRange> typeRange1,
+                              ValueTypeRange<OperandRange> typeRange2) {
+  if (typeRange1.size() != typeRange2.size()) return false;
+  for (auto it : llvm::zip(typeRange1, typeRange2)) {
+    if (!mlir::hlo::isCompatibleForHloTypeInference(std::get<0>(it),
+                                                    std::get<1>(it)))
+      return false;
   }
+  return true;
+}
```
burmako, line 4682: I think that having a function that takes `TypeRange` will be sufficient, because `TypeRange` has constructor that takes `ValueTypeRange`.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -4660,48 +4660,80 @@ LogicalResult ReplicaIdOp::inferReturnTypes(
...
+  IfOp::Adaptor adaptor(operands, attributes, regions);
+  if (failed(verifyConditionalBranch(adaptor.true_branch(),
+                                     /*branchName=*/"true_branch", location)))
     return failure();
-  }
+  if (failed(verifyConditionalBranch(adaptor.false_branch(),
+                                     /*branchName=*/"false_branch", location)))
+    return failure();
+
+  auto trueBranchRetureTypes =
```
burmako, line 4696: Type: "trueBranchReturnTypes". Same for falseBranch.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -4660,48 +4660,80 @@ LogicalResult ReplicaIdOp::inferReturnTypes(
...
+    auto branchResultTypes = region->front().getTerminator()->getOperandTypes();
+    if (!isCompatibleForTypeRange(branch0ResultTypes, branchResultTypes))
+      return emitOptionalError(location, "branch 0 and ", branchName,
+                               " have mismatched return types: ",
+                               branch0ResultTypes, " vs ", branchResultTypes);
+  }
+  for (auto resultType : branch0ResultTypes)
+    inferredReturnShapes.emplace_back(resultType.cast<ShapedType>());
   return success();
 }
```
burmako, line 4738: I think the body of this function should be factored out into a helper function called from both IfOp and CaseOp type inference, to deduplicate the existing implementation. Something like:

```
static TypeRange inferConditionalReturnTypeComponents(Optional<Location> location, RegionRange branches) {
  // 1) Verify that branches is non-empty.
  // 2) Pick the type of the first branch.
  // For each branch:
  // 3) Verify that the branch has zero arguments.
  // 4) Verify that it has a compatible type.
}
```

`verifyConditionalBranch` could be merged into this function - I don't think it's needed as a separate thing in addition to `inferConditionalReturnTypeComponents`.

stablehlo/tests/infer_stablehlo.mlir
```diff
@@ -275,6 +275,34 @@ func.func @batch_norm_inference(%input: tensor<4x256xf32>, %scale: tensor<256xf3
   func.return %1 : tensor<4x256xindex>
 }
 
+// -----
+
+// CHECK-LABEL: if 
+func.func @if(%pred : tensor<i1>, %branch_operand : tensor<2xf32>, %wrong_type : tensor<2xf32>) {
+  %0 = "stablehlo.if"(%pred) ({
+      "stablehlo.return"(%wrong_type) : (tensor<2xf32>) -> ()
```
burmako, line 283: What kind of terminators are we expecting in conditional ops? I think we should restrict them to just `stablehlo.return` to make sure we don't depend on other dialects (so that we can provide robust compatibility guarantees).

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -488,6 +488,54 @@ func.func @broadcast_in_dim_unranked_operand(%arg0 : tensor<*xf32>) -> tensor<2x
 
 // -----
 
+// CHECK-LABEL: if 
+func.func @if(%pred : tensor<i1>, %branch_operand : tensor<2xf32>, %wrong_type : tensor<2xf32>) {
+  %0 = "stablehlo.if"(%pred) ({
+      "stablehlo.return"(%wrong_type) : (tensor<2xf32>) -> ()
```
burmako, line 494: What's wrong about this type?

burmako (changes requested):

stablehlo/dialect/StablehloOps.td
zhouxin913: Sure, will submit after https://github.com/openxla/stablehlo/pull/145

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Good catch, added in the tests.

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Done.

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Done.

stablehlo/tests/ops_stablehlo.mlir
zhouxin913: Ah I think this is old exsiting naming, which I changed some names but not all. Fixed.

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Actually I was thinking of similar way and what's stop me was considering that I need pass in a different list of branch names like `branch_true` or `branch 0`, and need make sure the len(branch_names) == regionRange.size, or use Array<pair<branch, branch_name>>, which is complicated. 
But now I prefer your suggested way with general branch names for both IfOp and CaseOp: branch 0, branch 1 ... Because now I feel these also good names for IfOp too: in mlir test, true and false branches exists in similar way like CaseOp.
Fixed. And now almost all code are shared!

stablehlo/tests/infer_stablehlo.mlir
zhouxin913: Looks like all_reduce/reduce_scatter/map all have this problem. Could you explain the solution in your mind? I'm thinking of a check like: `region->front().getTerminator().isa<ReturnOp>()` but what the terminator in negative test looks like?
these are not valid:
"return"(%branch_operand) : (tensor<2xf32>) -> ()
"llvm.return"(%branch_operand) : (tensor<2xf32>) -> ()

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -4660,49 +4661,65 @@ LogicalResult ReplicaIdOp::inferReturnTypes(
...
+      return false;
+  }
+  return true;
+}
 
+static LogicalResult inferConditionalReturnTypeComponents(
+    RegionRange branches, Optional<Location> location,
+    SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
+  if (branches.empty())
+    return emitOptionalError(location, "expect as lease one branch");
```
burmako, line 4679: Typo: "at least".

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -4660,49 +4661,65 @@ LogicalResult ReplicaIdOp::inferReturnTypes(
...
-           << branchName << " must have 0 arguments, but found "
-           << region.getNumArguments();
-
-  TypeRange branchReturnTypes =
-      region.front().getTerminator()->getOperandTypes();
-  if (branchReturnTypes != op->getResultTypes())
-    return op->emitOpError()
-           << branchName << " returned types (" << branchReturnTypes
-           << ") do not match op result types (" << op->getResultTypes() << ")";
+// TODO: remove after https://github.com/openxla/stablehlo/pull/145 landed.
```
burmako, line 4664: Given that this PR will likely land before the mentioned PR, let's add this function to Base.h right here?

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -4660,49 +4661,65 @@ LogicalResult ReplicaIdOp::inferReturnTypes(
...
-    return failure();
-  }
-  return success();
+LogicalResult IfOp::inferReturnTypeComponents(
+    MLIRContext*, Optional<Location> location, ValueShapeRange operands,
+    DictionaryAttr attributes, RegionRange regions,
+    SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
+  IfOp::Adaptor adaptor(operands, attributes, regions);
+  return inferConditionalReturnTypeComponents(
+      ArrayRef<Region*>{&adaptor.true_branch(), &adaptor.false_branch()},
```
burmako, line 4708: This can be simplified to `adaptor.getRegions()`.

stablehlo/tests/infer_stablehlo.mlir
burmako: Huh you're right. Looks like this is already validated for us :)

stablehlo/dialect/StablehloOps.cpp
burmako: Very nice! 👏

burmako (changes requested):

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -4686,7 +4697,8 @@ LogicalResult IfOp::inferReturnTypeComponents(
       adaptor.true_branch().front().getTerminator()->getOperandTypes();
   auto falseBranchRetureTypes =
```
GleasonK: typo "return" ? for both the true and false branch variables

stablehlo/dialect/StablehloOps.td
```diff
@@ -1118,6 +1118,17 @@ def StableHLO_IfOp: StableHLO_Op<"if", [
                         SizedRegion<1>:$false_branch);
 
   let results = (outs Variadic<HLO_TensorOrToken>);
+
+  let extraClassDeclaration = [{
+    // Method from InferTypeOpInterface interface.
+    static bool isCompatibleReturnTypes(TypeRange l, TypeRange r) {
```
GleasonK, line 1124: Is this different from `isCompatibleForTypeRange` from above? Maybe that's what Eugene's feedback from above meant.

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Done.

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Yes I change this for both IfOp and CaseOp so that they are aligned.

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Yeah, I can't find it in my code now, maybe fixed earlier :-)

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Yes, I copy `isCompatibleForHloTypeInference` from`Base.h/cc`in https://github.com/openxla/stablehlo/pull/145 to the same location here in this PR.

stablehlo/dialect/StablehloOps.td
zhouxin913: Exactly, and this method is copied again and again among different ops in ODS. That's why the pending https://github.com/openxla/stablehlo/pull/145 is trying to address this issue by moving this function to `Base.h/cc` and adding to top level op then inherit by all ops. 
Like Eugene said: https://github.com/openxla/stablehlo/pull/145 may need some time to refine and land, so it does not hurt if I copy this method to `Base.h/cc` in advance in the PR right now.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -4660,49 +4661,54 @@ LogicalResult ReplicaIdOp::inferReturnTypes(
...
+  for (unsigned i = 0; i < branches.size(); ++i) {
+    Twine branchName = "branch " + Twine(i);
+    Region* region = branches[i];
+    if (region->getNumArguments() != 0)
+      return emitOptionalError(location, branchName,
+                               " must have 0 arguments, but found ",
+                               region->getNumArguments());
+
+    auto branchResultTypes = region->front().getTerminator()->getOperandTypes();
+    if (!mlir::hlo::isCompatibleForHloTypeInference(branch0ResultTypes,
```
burmako, line 4681: You can say `hlo::isCompatibleForHloTypeInference`, since we're already in the `mlir` namespace. Other than that, LGTM.

burmako (approved):

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Done.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -4664,49 +4664,54 @@ LogicalResult ReplicaIdOp::inferReturnTypes(
...
-           << branchName << " returned types (" << branchReturnTypes
-           << ") do not match op result types (" << op->getResultTypes() << ")";
-
+static LogicalResult inferConditionalReturnTypeComponents(
+    RegionRange branches, Optional<Location> location,
+    SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
+  if (branches.empty())
+    return emitOptionalError(location, "expect at least one branch");
+
+  ValueTypeRange<OperandRange> branch0ResultTypes =
```
gnecula, line 4673: I am not sure I understand the logic here. It seems that shape inference will always return the type of branch0. If branch0 returns f32[5] and branch1 returns f32[?] then the correct type to infer is the type of branch1.

Is it true that the correctness criterion for shape inference is that if it succeeds with f32[5] then any execution of the code is guaranteed to have that shape?

zhouxin913: Hi George, thanks for your comment! Yes a more complicated case would be branch0 returns tensor<5x?xf32> while branch1 returns tensor<?x5xf32>: the ideal expected result should be tensor<5x5xf32>.
This PR provides a basic version to include dynamism in type inference. I added a while ago an util function in [inferMostSpecificType](https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/Base.h#L218) and here is the new PR to fix this: https://github.com/openxla/stablehlo/pull/229

Yes, inferring dim=5 is always true no matter which branch is taken in executions.

gnecula: I would have said that in your example the inferred shape should be <?x?xf32>. I commented in #229.

PR #137, opened 2022-09-15
Build with all warnings as errors.

### LINKED ISSUE #125: Decide what to do with build warnings
MLIR-HLO has a problem with build warning spew (https://github.com/tensorflow/mlir-hlo/issues/43), and we're starting to have an issue with this as well. One possibility to address this would be to compile with `-Werror`.

GleasonK:
Fixes a build warning with `Tensor.h`:

```
In file included from /usr/local/google/home/gleasonk/Coding/openxla/stablehlo/stablehlo/reference/Tensor.cpp:16:
/usr/local/google/home/gleasonk/Coding/openxla/stablehlo/stablehlo/reference/Tensor.h:44:11: error: explicitly defaulted move assignment operator is implicitly deleted [-Werror,-Wdefaulted-function-deleted]
  Buffer &operator=(Buffer &&other) = default;
          ^
/usr/local/google/home/gleasonk/Coding/openxla/stablehlo/stablehlo/reference/Tensor.h:33:16: note: move assignment operator of 'Buffer' is implicitly deleted because base class 'llvm::RefCountedBase<Buffer>' has a deleted move assignment operator
class Buffer : public llvm::RefCountedBase<Buffer> {
               ^
/usr/local/google/home/gleasonk/Coding/llvm-project/llvm/include/llvm/ADT/IntrusiveRefCntPtr.h:82:19: note: 'operator=' has been explicitly marked deleted here
  RefCountedBase &operator=(const RefCountedBase &) = delete;
                  ^
1 error generated.
```

Also note: there is an issue with clang regarding `-Werror` and `lld`. Without the ` -Wno-unused-command-line-argument` flag, it errors with:

```
cmake -GNinja \
  -B"$STABLEHLO_BUILD_DIR" \
  -DLLVM_ENABLE_LLD=ON \
  -DCMAKE_BUILD_TYPE=Release \
  -DLLVM_ENABLE_ASSERTIONS=On \
  -DMLIR_DIR="$LLVM_BUILD_DIR/lib/cmake/mlir" \
  -DCMAKE_CXX_COMPILER=clang++ \
  -DCMAKE_C_COMPILER=clang \
  -DCMAKE_C_COMPILER_LAUNCHER=ccache

-- The CXX compiler identification is Clang 14.0.6
-- The C compiler identification is Clang 14.0.6
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/clang++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/clang - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Building StableHLO standalone
-- Performing Test Terminfo_LINKABLE
-- Performing Test Terminfo_LINKABLE - Success
-- Found Terminfo: /usr/lib/x86_64-linux-gnu/libtinfo.so  
-- Found ZLIB: /usr/lib/x86_64-linux-gnu/libz.so (found version "1.2.11") 
-- Found LibXml2: /usr/lib/x86_64-linux-gnu/libxml2.so (found version "2.9.14") 
-- Using MLIRConfig.cmake in: /usr/local/google/home/gleasonk/Coding/llvm-build/lib/cmake/mlir
-- Using LLVMConfig.cmake in: /usr/local/google/home/gleasonk/Coding/llvm-build/lib/cmake/llvm
-- Linker detection: GNU ld
-- Performing Test LLVM_LIBSTDCXX_MIN
-- Performing Test LLVM_LIBSTDCXX_MIN - Success
-- Performing Test LLVM_LIBSTDCXX_SOFT_ERROR
-- Performing Test LLVM_LIBSTDCXX_SOFT_ERROR - Success
-- Performing Test CXX_SUPPORTS_CUSTOM_LINKER
-- Performing Test CXX_SUPPORTS_CUSTOM_LINKER - Failed
CMake Error at /usr/local/google/home/gleasonk/Coding/llvm-build/lib/cmake/llvm/HandleLLVMOptions.cmake:309 (message):
  Host compiler does not support '-fuse-ld=lld'
Call Stack (most recent call first):
  CMakeLists.txt:75 (include)
...
-----------------------------------------------------------------
$ cat <build_dir>/CMakeFiles/CMakeError.log
Performing C++ SOURCE FILE Test CXX_SUPPORTS_CUSTOM_LINKER failed with the following output:
Change Dir: /usr/local/google/home/gleasonk/Coding/stablehlo-build/CMakeFiles/CMakeTmp

Run Build Command(s):/usr/bin/ninja cmTC_d5508 && [1/2] Building CXX object CMakeFiles/cmTC_d5508.dir/src.cxx.o
FAILED: CMakeFiles/cmTC_d5508.dir/src.cxx.o 
/usr/bin/clang++ -DCXX_SUPPORTS_CUSTOM_LINKER  -Wall -Werror  -fuse-ld=lld -std=gnu++17 -MD -MT CMakeFiles/cmTC_d5508.dir/src.cxx.o -MF CMakeFiles/cmTC_d5508.dir/src.cxx.o.d -o CMakeFiles/cmTC_d5508.dir/src.cxx.o -c /usr/local/google/home/gleasonk/Coding/stablehlo-build/CMakeFiles/CMakeTmp/src.cxx
clang: error: argument unused during compilation: '-fuse-ld=lld' [-Werror,-Wunused-command-line-argument]
ninja: build stopped: subcommand failed.


Source file was:
int main() { return 0; }
```

Closes #125.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,9 +47,8 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
   project(stablehlo LANGUAGES CXX C)
   set(CMAKE_C_STANDARD 11)
   set(CMAKE_CXX_STANDARD 17)
-  
-  # Issue with LLD and -Werror. Similar to golang/go#41527, iree-org/iree#7450
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Werror") 
+  set(LLVM_ENABLE_WARNINGS ON)
+  set(LLVM_ENABLE_WERROR ON)
 endif()
 
 #-------------------------------------------------------------------------------
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,9 +50,6 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
   
   # Issue with LLD and -Werror. Similar to golang/go#41527, iree-org/iree#7450
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Werror") 
-  if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-command-line-argument") 
-  endif()
 endif()
 
 #-------------------------------------------------------------------------------
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,12 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
   project(stablehlo LANGUAGES CXX C)
   set(CMAKE_C_STANDARD 11)
   set(CMAKE_CXX_STANDARD 17)
+  
+  # Issue with LLD and -Werror. Similar to golang/go#41527, iree-org/iree#7450
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Werror") 
+  if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-command-line-argument") 
+  endif()
 endif()
 
 #-------------------------------------------------------------------------------
--- a/stablehlo/reference/Tensor.h
+++ b/stablehlo/reference/Tensor.h
@@ -40,8 +40,8 @@ class Buffer : public llvm::RefCountedBase<Buffer> {
   Buffer(Buffer &&other) = default;
   /// @}
 
-  /// Assignment operator.
-  Buffer &operator=(Buffer &&other) = default;
+  /// Move assignment operator deleted in RefCountedBase
+  Buffer &operator=(Buffer &&other) = delete;
 
   /// Returns type of the Buffer object.
   ShapedType getType() { return type_; }


sdasgup3 (approved):

CMakeLists.txt
```diff
@@ -47,6 +47,12 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
   project(stablehlo LANGUAGES CXX C)
   set(CMAKE_C_STANDARD 11)
   set(CMAKE_CXX_STANDARD 17)
+  
+  # Issue with LLD and -Werror. Similar to golang/go#41527, iree-org/iree#7450
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Werror") 
+  if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-command-line-argument") 
+  endif()
```
burmako, line 55: We probably don't need this flag at the moment?

burmako (changes requested):

CMakeLists.txt
burmako: Oops, I failed to read the pull request description. Nvm.

burmako (approved):

CMakeLists.txt
```diff
@@ -47,6 +47,12 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
   project(stablehlo LANGUAGES CXX C)
   set(CMAKE_C_STANDARD 11)
   set(CMAKE_CXX_STANDARD 17)
+  
+  # Issue with LLD and -Werror. Similar to golang/go#41527, iree-org/iree#7450
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Werror") 
+  if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-command-line-argument")
```
joker-eph, line 54: This sounds weird to me: why is -fuse-ld part of CMAKE_CXX_FLAGS instead of CMAKE_EXE_LINKER_FLAGS ?

GleasonK: I have edited the PR description to include the build command. The `-fuse-ld` part comes from `-DLLVM_ENABLE_LLD=ON`, the `CMAKE_CXX_FLAGS` are separate, in the CMakeList in this case. 

At the bottom of the PR description it shows the eventual `clang++` call, editing out the files for brevity:

```
/usr/bin/clang++ -DCXX_SUPPORTS_CUSTOM_LINKER  -Wall -Werror  -fuse-ld=lld -std=gnu++17 -MD -MT ....
clang: error: argument unused during compilation: '-fuse-ld=lld' [-Werror,-Wunused-command-line-argument]
```

Perhaps there is a different configuration we should be using?

joker-eph: We should fix the cmake logic upstream. The intend when I implemented it was to just use it for the linker phase: https://reviews.llvm.org/D28738
But the test was a no-op and got "fixed" later: https://reviews.llvm.org/D39274

GleasonK: I roughly understand what you're suggesting, but not sure how to best implement it. I used the following, which seems to have the desired behavior (without knowing the nuances, i.e. not sure why we test before append in current setup):

```
if( LLVM_USE_LINKER )
  set(OLD_CMAKE_EXE_LINKER_FLAGS ${CMAKE_EXE_LINKER_FLAGS})
  append("-fuse-ld=${LLVM_USE_LINKER}" CMAKE_EXE_LINKER_FLAGS)
  check_cxx_source_compiles("int main() { return 0; }" CXX_SUPPORTS_CUSTOM_LINKER)
  if ( NOT CXX_SUPPORTS_CUSTOM_LINKER )
    message(FATAL_ERROR "Host compiler does not support '-fuse-ld=${LLVM_USE_LINKER}'")
  endif()
  set(CMAKE_EXE_LINKER_FLAGS ${OLD_CMAKE_EXE_LINKER_FLAGS})
  append("-fuse-ld=${LLVM_USE_LINKER}"
    CMAKE_EXE_LINKER_FLAGS CMAKE_MODULE_LINKER_FLAGS CMAKE_SHARED_LINKER_FLAGS)
endif()
```

And it seems like these problems are resolved, and unsupported linkers still error:

```
-- Performing Test CXX_SUPPORTS_CUSTOM_LINKER - Failed
CMake Error at cmake/modules/HandleLLVMOptions.cmake:309 (message):
  Host compiler does not support '-fuse-ld=notalinker'
Call Stack (most recent call first):
  CMakeLists.txt:870 (include)
```

Let me know your thoughts / any tests you think I should try. Happy to take this through the review process, should learn how to contribute to llvm-project sooner or later..

joker-eph: Can we even simplify this further?
```
if( LLVM_USE_LINKER )
  append("-fuse-ld=${LLVM_USE_LINKER}"
    CMAKE_EXE_LINKER_FLAGS CMAKE_MODULE_LINKER_FLAGS CMAKE_SHARED_LINKER_FLAGS)
  check_cxx_source_compiles("int main() { return 0; }" CXX_SUPPORTS_CUSTOM_LINKER)
  if ( NOT CXX_SUPPORTS_CUSTOM_LINKER )
    message(FATAL_ERROR "Host compiler does not support '-fuse-ld=${LLVM_USE_LINKER}'")
  endif()
endif()
```

OLD_CMAKE_EXE_LINKER_FLAGS is only useful if we make it not a FATAL error right?

joker-eph: But otherwise, LG :)

GleasonK: Changed this PR a bit. I found the following config flags when editing the cmake file: `LLVM_ENABLE_WARNINGS` and `LLVM_ENABLE_WERROR`. When setting these flags in our cmake file, we get the desired behavior without any issues regarding LLD. Tested that StableHLO build fails with compiler warnings when these flags are set.

There are also lines that have comments regarding `-Wall`, so we likely should use these LLVM flags instead of directly adding the `-W...` compiler flags, as they seem to be safer: https://github.com/llvm/llvm-project/blob/cbf81558ab541802d1b2a859d72680173cafcb10/llvm/cmake/modules/HandleLLVMOptions.cmake#L692

I will still make the suggested change to upstream, but I don't think merging this PR is blocked by that fix.

burmako (approved):
Very nice find!

CMakeLists.txt
GleasonK: Created: https://reviews.llvm.org/D134206

CMakeLists.txt
```diff
@@ -47,6 +47,8 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
   project(stablehlo LANGUAGES CXX C)
   set(CMAKE_C_STANDARD 11)
   set(CMAKE_CXX_STANDARD 17)
+  set(LLVM_ENABLE_WARNINGS ON)
+  set(LLVM_ENABLE_WERROR ON)
```
joker-eph, line 51: I have very strong concerns with this. 

Werror should never be on by default: this can make it fail on someone's machine just because they are using a different compiler, C++ library, or even C library!

My understanding from the issue was that you were looking to turn this on in CI, which is a very different value proposition (you control the environment entirely).

GleasonK: That sounds reasonable. @burmako any opposition to removing this from the CMakeLists and adding these as flags in `ci_build_stablehlo.sh` so we only build with warnings as errors in CI? We can still verify our changes locally with ease by calling the build script.

joker-eph: > We can still verify our changes locally with ease by calling the build script.

You should provide a docker image for this, otherwise you're still dependent on the environment (compiler version and libraries on the host)

Issue #96, opened 2022-09-07
Decide on a policy for specifying numeric precision


burmako:
In #86, we've gotten into a discussion of what it means exactly to perform a cosine operation. Or, more precisely, what kind of precision with respect to an ideal mathematical result are we talking about? Initially, the policy could be as simple as "Numeric precision is implementation-defined", but I think that eventually we could work out something more specific.

sdasgup3:
A few pointers:

For implementing transcendental operations, we plan to use use std variants, thus exploring them could give us some hint on the precision guarantees we can make.  One option could be to say that the numeric precision is the same as offered by the std variants. 

Please note that  std variants use GNU libm which looks system dependent. 

```
git clone git://sourceware.org/git/glibc.git
find sysdeps -name *cos.c
```

ghpvnist:
Section 9.2 of IEEE-754 has a list of “***recommended*** correctly rounded functions”. Following the IEEE-754 standard does not guarantee correctly rounded to 0.5 ulp (rounded exact to the last bit). Other std functions in C++ follow the standard by using double, thus any calculations involving floats should be cast to double.

`f64` has 53 bits of mantissa bits (52 explicit and 1 implied bit), which allows $2^{-53}\approx 1.11*10^{-16}$, or 15-17 significant decimal digits of precision. If we really want our best attempt at conforming to the standard with more than 17 digits of precision for `f64` or allow accumulation of results for single (pow, which does log of the exponent number of multiplies) or multiple operations, using `long double` is an option to consider, noting that different variants of the compiler has varying bits of storage for `long double`.

Related links:
https://en.wikipedia.org/wiki/Double-precision_floating-point_format#cite_note-whyieee-1
https://stackoverflow.com/questions/69380549/is-cosx-required-to-return-identical-values-in-different-c-implementations-t
https://stackoverflow.com/questions/4317988/ieee-754-floating-point-precision-how-much-error-is-allowed

sdasgup3:
As pointed that these transcendental functions are listed as “recommended correctly rounded functions”. Being a recommendation,  a conforming implementation can chose to move away from "correct rounding"  (rounding an "infinitely precise" result of the operation to the closest representable floating-point number) recommendation.  In our case, it allows to accept any reasonable precision provided by the existing implementations of transcendental functions, which ranges from approximating the values from Taylor's series to mapping to a [hardware instruction](https://www.felixcloutier.com/x86/fcos)

joker-eph:
[...]

PR #139, opened 2022-09-15
Add interpreter for sine op

### LINKED ISSUE #138: Interpreter for sine op

ghpvnist:
fixes #138

### DIFF
--- a/stablehlo/reference/Element.cpp
+++ b/stablehlo/reference/Element.cpp
@@ -120,6 +120,62 @@ Element Element::operator+(const Element &other) const {
   report_fatal_error(std::move(err));
 }
 
+Element sin(const Element &e) {
+  Type type = e.getType();
+  if (isSupportedComplexType(type)) {
+    auto complex = getComplexValue(e);
+    APFloat realAPF = complex.real();
+    APFloat imagAPF = complex.imag();
+    Type elementType = e.getValue().cast<ArrayAttr>()
+                        .getValue()[0].cast<FloatAttr>().getType();
+    if (elementType.isF32()) {
+      std::complex<float> complex(realAPF.convertToFloat(),
+                                  imagAPF.convertToFloat());
+      std::complex<float> sinVal = std::sin(complex);
+      return Element(elementType,
+          ArrayAttr::get(
+            elementType.getContext(),
+            {FloatAttr::get(elementType, APFloat(sinVal.real())),
+             FloatAttr::get(elementType, APFloat(sinVal.imag()))}));
+    } else if (elementType.isF64()) {
+      std::complex<double> complex(realAPF.convertToDouble(),
+                                   imagAPF.convertToDouble());
+      std::complex<double> sinVal = std::sin(complex);
+      return Element(elementType,
+          ArrayAttr::get(
+              elementType.getContext(),
+              {FloatAttr::get(elementType, APFloat(sinVal.real())),
+               FloatAttr::get(elementType, APFloat(sinVal.imag()))}));
+    }
+  } else if (isSupportedFloatType(type)) {
+    APFloat val = e.getValue().cast<FloatAttr>().getValue();
+    if (type.isBF16()) {
+      APFloat sinVal = APFloat(std::sin(val.convertToFloat()));
+      bool roundingErr;
+      sinVal.convert(APFloat::BFloat(),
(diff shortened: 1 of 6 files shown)


ghpvnist:
Converting all element types to `f64` to calculate `std::sin` can reduce the LOC without reducing any more precision than when compared to calculating `std::sin` of element types `bf16`, `f16`, and `f32` in `f32`.

One example where rounding errors occur is with `sine_op_test_c128`, where `(3.5, 4.5)` gives rounding errors for `f32` when calculated in `f32` as opposed to `f64`. The value when converted to `f32` is:
`[-15.7901974 : f32, -42.1433716 : f32]`
whereas value when converted to `f64` is:
`[-15.7901983 : f32, -42.1433716 : f32]`
Its `f64` counterpart correctly calculates up to its formatting digits:
`[-15.790198357309713, -42.143370741504995]`

Although the imaginary part still has rounding issues when cast back to `f32`, the real value portion is able to maintain more units of precision when calculation is done in `f64`. Thus, from this op and forward, any floating point calculation will be done in double and then cast back to its original semantics. As per #96, there is no ulp precision guarantee for ops not required by the IEEE-754 spec.

stablehlo/reference/Element.cpp
```diff
@@ -120,6 +120,45 @@ Element Element::operator+(const Element &other) const {
   report_fatal_error(std::move(err));
 }
 
+Element sine(const Element &e) {
+  Type type = e.getType();
+  if (isSupportedComplexType(type)) {
```
sdasgup3, line 125: Can we maintain the same order of processing types as in other ops? (floats -> ints -> complex)

stablehlo/reference/Element.cpp
```diff
@@ -120,6 +120,45 @@ Element Element::operator+(const Element &other) const {
...
+    auto complex = getComplexValue(e);
+    Type elementType = e.getValue()
```
sdasgup3, line 127: We can get the same effect using:
`Type elementType = type.getElementType();`

stablehlo/reference/Element.cpp
```diff
@@ -120,6 +120,41 @@ Element Element::operator+(const Element &other) const {
   report_fatal_error(std::move(err));
 }
 
+Element sine(const Element &e) {
+  Type type = e.getType();
+  if (isSupportedFloatType(type)) {
+    APFloat val = e.getValue().cast<FloatAttr>().getValue();
```
sdasgup3, line 126: Use getFloatValue instead?

stablehlo/tests/interpret_sine.mlir
```diff
@@ -0,0 +1,111 @@
+// RUN: stablehlo-interpreter --interpret -split-input-file %s | FileCheck %s
+
+// CHECK-LABEL: Evaluated results of function: sine_op_test_bf16
+func.func @sine_op_test_bf16() -> tensor<12xbf16> {
+  %0 = stablehlo.constant dense<[0.0, -0.0, 1.0, 0.125, 0.1, 3.140630, 0x7F80, 0xFF80, 0x7FFF, 0xFFFF, 0x0001, 0x8001]> : tensor<12xbf16>
```
sdasgup3, line 5: missing +/- Inf?
Keep one representation of NaN and denormals? 
here and below.

ghpvnist: The hex values in each test case currently are: +inf, -inf, +NaN, -NaN, +Denormal (smallest), -Denormal (largest). Since we are not differentiating the difference between +/- NaN, I will remove one. I will also create a separate issue to revisit and explore which tests are worth exploring (to add necessary tests and remove unnecessary tests).

stablehlo/reference/Element.cpp
```diff
@@ -120,6 +120,41 @@ Element Element::operator+(const Element &other) const {
   report_fatal_error(std::move(err));
 }
 
+Element sine(const Element &e) {
+  Type type = e.getType();
+  if (isSupportedFloatType(type)) {
+    APFloat val = getFloatValue(e);
+    const llvm::fltSemantics &oldSemantics = val.getSemantics();
+    bool roundingErr;
+    val.convert(APFloat::IEEEdouble(), APFloat::rmNearestTiesToEven,
```
sdasgup3, line 129: Do we need this as the following convertToDouble should do this anyways?

sdasgup3 (approved):

PR #142, opened 2022-09-17
Add additional special test cases for interpret_add.mlir and interpret_constant.mlir

ghpvnist:
The current tests do not address all special values for floats.

The list of special values are:
* +/- 0
* +/- inf
* +/- NaN
* +/- Subnormal

However, the spec does not differentiate signedness of the NaN, so only the +NaN is tested. An investigation will be done in the near future to determine whether the tests include all special cases which are addressed in the spec for all ops in question.


burmako (changes requested):
In the spirit of the recent discussion about commit messages, can we update the pull request description to explain what these special cases are?

sdasgup3 (approved):
Approve (with changes requested by @burmako )

burmako (approved):

PR #145, opened 2022-09-17
 Enhance result capability match for all ops.

zhouxin913:
Currently, some ops, like `BinaryElementwiseOp` and `ReduceOp`, have their own `extraClassDeclaration` part in ODS which calls `isCompatibleForHloTypeInference()`, providing a relaxed check for "inferred type" VS "real type". The intent is to support dynamism/sparsity/quantization. Now as more and more ops are going to support them, this PR moves this part upward to `StableHLO_Op` and `StableHLO_ShapedInterfaceOp` so that all ops can inherit this feature.

Besides, Some ops calls MLIR method `mlir::verifyCompatibleShapes()` directly in ODS which ONLY checks shapes but not elements: This is reasonable for `DotOp`, which by design tolerant result element mismatch. But these ops are not :
ConcatenateOp
CompareOp
GatherOp
SelectOp
Transpose
RngOp
I suppose the historical reason that ignoring the element checks was to support dynamism/sparsity/quantization. But now the `mlir::verifyCompatibleShapes()` call insides these ops can be removed, and implicitly use the inherited `isCompatibleForHloTypeInference()` instead.


stablehlo/dialect/StablehloOps.td
```diff
@@ -2491,11 +2433,7 @@ def StableHLO_ReduceWindowOp: StableHLO_Op<"reduce_window", [
...
     // for StableHLO-specific differences.
     static bool isCompatibleReturnTypes(TypeRange l, TypeRange r) {
-      if (l.size() != r.size()) return false;
-      for (auto [lt, rt] : llvm::zip(l, r))
-        if (!mlir::hlo::isCompatibleForHloTypeInference(lt, rt))
-          return false;
-      return true;
+      return mlir::hlo::isCompatibleForHloTypeInference(l, r);
     }
   }];
```
burmako, line 2438: Why do we still need this extraClassDeclaration here, if it's already defined in StableHLO_Op? Same question for StableHLO_UnaryElementwiseOp and StableHLO_BinaryElementwiseOp. I see that tests fail without that but why?

stablehlo/dialect/StablehloOps.td
```diff
@@ -2788,12 +2719,6 @@ def StableHLO_DynamicGatherOp: StableHLO_Op<"dynamic_gather",
...
   let results = (outs HLO_Tensor);
-
-  let extraClassDeclaration = [{
-    static bool isCompatibleReturnTypes(TypeRange l, TypeRange r) {
-      return succeeded(mlir::verifyCompatibleShapes(l, r));
-    }
-  }];
 }
 
 def StableHLO_DynamicConvOp : StableHLO_Op<"dynamic_conv", [NoSideEffect]> {
```
burmako, line 2724: What kind of tests do we want to add to test this new functionality?

burmako, line 2724: Also, should we make the same change for CHLO?

stablehlo/dialect/StablehloOps.td
```diff
@@ -50,6 +57,13 @@ include "stablehlo/dialect/StablehloAttrs.td"
 class StableHLO_ShapedInterfaceOp<string mnemonic, list<Trait> traits> :
     StableHLO_Op<mnemonic, traits # [DeclareOpInterfaceMethods<InferShapedTypeOpInterface,
     ["reifyReturnTypeShapes"]>]> {
+  let extraClassDeclaration = [{
+    // Relax the strict default implementation with one that allows
+    // for StableHLO-specific differences.
+    static bool isCompatibleReturnTypes(TypeRange l, TypeRange r) {
+      return mlir::hlo::isCompatibleForHloTypeInference(l, r);
+    }
+  }];
```
burmako, line 66: This extraClassDeclaration doesn't seem to be needed. Tests succeed without it.

burmako (changes requested):

stablehlo/dialect/StablehloOps.td
zhouxin913: Yes, this is a surprising point here is the reason: ops inherit everything including `let arguments = ..` or `let extraClassDeclaration = ` from parent `StableHLO_Op`, but if the op itself contains same thing, it will "overwrite" what are inherited.
For example, here `ReduceWindowOp` has 
```
let extraClassDeclaration = [{
    // Get the operation used for reduction applied to `result_index`th result.
    Operation *getReductionOp(int result_index);
   ...
```
I don't think there is a way that we can inherit `let extraClassDeclaration = [{ static bool isCompatibleReturnTypes() ...` from `StableHLO_Op` and somehow **merge** them together (`extraClassDeclaration` is pure string). The related code is https://github.com/llvm/llvm-project/blob/main/mlir/tools/mlir-tblgen/OpDefinitionsGen.cpp#L800 which just read this attribute and use it. Similar situation for StableHLO_UnaryElementwiseOp and StableHLO_BinaryElementwiseOp.

stablehlo/dialect/StablehloOps.td
zhouxin913: Sure.

stablehlo/dialect/StablehloOps.td
zhouxin913: All tests are expect to succeed both before and after this PR: the intend of this PR is to loose the checks, so that in the future when we add dynamism in ops, like `ClampOp` which inherit this `StableHLO_ShapedInterfaceOp`, there will be not need to change the `.td` file any more: all the ops already are equipped a powerful check (inferred type VS real result type) by this PR.

stablehlo/dialect/StablehloOps.td
zhouxin913: Not now: same as mentioned in another comment, this PR loose the checks (inferred type VS real return type), so that all tests are expected to pass without any changes. When we implement dynamism one by one for each op, finer change specific for each op can be done case by case only when necessary.

stablehlo/dialect/StablehloOps.td
burmako: Okay, how about a different idea. 

Can we use traits as a mechanism to inherit `isCompatibleReturnTypes` from a base class? E.g. this is how we inherit another static function - `inferReturnType` - in all ops which are marked with `[HLO_CompatibleOperandsAndResultType]`: https://github.com/openxla/stablehlo/blob/b08a70ced3883994b8c410a6439f7114f4325d85/stablehlo/dialect/Base.h#L167-L182
Basically, we could create a new trait which would be added to the list of traits of all StableHLO ops by StableHLO_Op (same for CHLO), and this trait wouldn't do anything except for declaring `isCompatibleReturnTypes`. If individual ops would want to provide their own implementation, they would be able to do that via `extraClassDeclaration`, just like now.

burmako (changes requested):

stablehlo/dialect/StablehloOps.td
zhouxin913: Thanks for your suggestion! However, after investigation, there is no clear solution for this approach:
Similar to `HLO_CompatibleOperandsAndResultType`, we can try add a default implementation of `isCompatibleReturnTypes` in `Base.h` or `stablehloOps.cc`, but the errors will be:
1. adding in `Base.h`. There will be implementation conflicts against https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Interfaces/InferTypeOpInterface.td#L87,
2. adding `Op::isCompatibleReturnTypes` in `stablehloOps.cc` will trigger the error that `Op::isCompatibleReturnTypes` is not declared. And so far there is no clear solution to declare it because of http://google3/third_party/llvm/llvm-project/mlir/tools/mlir-tblgen/OpDefinitionsGen.cpp;l=2269;rcl=475283550, even after we declear it in the trait like:
```
def HLO_CompatibleOperandsAndResultType : TraitList<
[
  DeclareOpInterfaceMethods<InferTypeOpInterface, ["isCompatibleReturnTypes"]>,
  DeclareOpInterfaceMethods<InferShapedTypeOpInterface, ["inferReturnTypeComponents"]>,
  HLO_NativeOpTrait<"CompatibleOperandsAndResultType">
]>;
```
So for now let's stick to the current implementation.

stablehlo/dialect/StablehloOps.td
```diff
@@ -50,6 +57,13 @@ include "stablehlo/dialect/StablehloAttrs.td"
 class StableHLO_ShapedInterfaceOp<string mnemonic, list<Trait> traits> :
     StableHLO_Op<mnemonic, traits # [DeclareOpInterfaceMethods<InferShapedTypeOpInterface,
     ["reifyReturnTypeShapes"]>]> {
+  let extraClassDeclaration = [{
+    // Relax the strict default implementation with one that allows
+    // for StableHLO-specific differences.
```
burmako, line 62: I understand that it's not easy to get rid of the need to duplicate `isCompatibleReturnTypes` everywhere, and that's fine for now, but let's delete the repetitive comments. We can keep one comment in `CHLO_Op`, one comment in `StableHLO_Op`, but I think that the rest should go.

burmako (changes requested):

stablehlo/dialect/StablehloOps.td
burmako, line 2724: By the way, I've just realized that [`HLO_PairwiseSameOperandAndResultType`](https://github.com/openxla/stablehlo/blob/388b134277b19b38293354cf7ae4ac0ad4b37447/stablehlo/dialect/Base.td#L179-L180) is still problematic. It needs to be changed to `HLO_PairwiseCompatibleOperandAndResultType` to implement the intent of this PR.

burmako (changes requested):

stablehlo/dialect/StablehloOps.td
burmako: We can land this PR as is, once the previous comment about deleting comments is addressed, and work on this comment in https://github.com/openxla/stablehlo/pull/207.

stablehlo/dialect/StablehloOps.td
zhouxin913: Make senses! Only keep 1 comment in each td file.

stablehlo/dialect/StablehloOps.td
zhouxin913: Oh sure. Still, let me change `[HLO_PairwiseSameOperandAndResultType](https://github.com/openxla/stablehlo/blob/388b134277b19b38293354cf7ae4ac0ad4b37447/stablehlo/dialect/Base.td#L179-L180)` to `HLO_PairwiseCompatibleOperandAndResultType` to make this PR consistent.

stablehlo/tests/ops_chlo_roundtrip.mlir
```diff
@@ -353,11 +353,11 @@ func.func @chlo_zeta(%arg0: tensor<2x3x4xf64>, %arg1: tensor<2x3x4xf64>) -> tens
...
-// CHECK:       return %[[T]] : tensor<2x3x4xcomplex<f32>>
-func.func @chlo_broadcast_complex(%arg0: tensor<2x3x4xf64>, %arg1: tensor<2x3x4xf64>) -> tensor<2x3x4xcomplex<f32>> {
-  %0 = chlo.broadcast_complex %arg0, %arg1 : (tensor<2x3x4xf64>, tensor<2x3x4xf64>) -> tensor<2x3x4xcomplex<f32>>
-  return %0 : tensor<2x3x4xcomplex<f32>>
+// CHECK:       %[[T:.*]] = chlo.broadcast_complex %[[A0]], %[[A1]] : (tensor<2x3x4xf64>, tensor<2x3x4xf64>) -> tensor<2x3x4xcomplex<f64>>
+// CHECK:       return %[[T]] : tensor<2x3x4xcomplex<f64>>
+func.func @chlo_broadcast_complex(%arg0: tensor<2x3x4xf64>, %arg1: tensor<2x3x4xf64>) -> tensor<2x3x4xcomplex<f64>> {
+  %0 = chlo.broadcast_complex %arg0, %arg1 : (tensor<2x3x4xf64>, tensor<2x3x4xf64>) -> tensor<2x3x4xcomplex<f64>>
+  return %0 : tensor<2x3x4xcomplex<f64>>
 }
```
burmako, line 361: Overall, I don't have an objection to this, but during Google3 integrate we'll need to see if this breaks someone downstream.

burmako (approved):

PR #148, opened 2022-09-19
Fix CMake build

ashay:
This patch reorders the libraries linked into the stablehlo-interpreter
target, since otherwise, the CMake build fails with the error:

    StablehloOps.cpp: undefined reference to `mlir::DataLayout::closest`

This patch also adds `target_include_directories()` for the
`StablehloBase` and `StablehloBroadcastUtils` so that path to the
Stablehlo header files is added to the include path of downstream
targets.

### DIFF
--- a/stablehlo/dialect/CMakeLists.txt
+++ b/stablehlo/dialect/CMakeLists.txt
@@ -72,6 +72,11 @@ add_mlir_dialect_library(ChloOps
   MLIRQuantDialect
 )
 
+target_include_directories(ChloOps INTERFACE
+  $<BUILD_INTERFACE:${STABLEHLO_SOURCE_DIR}>
+  $<BUILD_INTERFACE:${STABLEHLO_BINARY_DIR}>
+)
+
 add_mlir_dialect_library(StablehloRegister
   PARTIAL_SOURCES_INTENDED
   Register.cpp
@@ -106,6 +111,7 @@ add_mlir_dialect_library(StablehloOps
   LINK_LIBS PUBLIC
   StablehloBase
   MLIRArithDialect
+  MLIRDataLayoutInterfaces
   MLIRInferTypeOpInterface
   MLIRIR
   MLIRQuantDialect
@@ -113,4 +119,9 @@ add_mlir_dialect_library(StablehloOps
   MLIRSparseTensorDialect
   MLIRSupport
   MLIRTensorDialect
+)
+
+target_include_directories(StablehloOps INTERFACE
+  $<BUILD_INTERFACE:${STABLEHLO_SOURCE_DIR}>
+  $<BUILD_INTERFACE:${STABLEHLO_BINARY_DIR}>
 )
--- a/stablehlo/tools/CMakeLists.txt
+++ b/stablehlo/tools/CMakeLists.txt
@@ -38,7 +38,6 @@ mlir_check_all_link_libraries(stablehlo-opt)
 add_llvm_executable(stablehlo-interpreter StablehloInterpreterMain.cpp)
 llvm_update_compile_flags(stablehlo-interpreter)
 target_link_libraries(stablehlo-interpreter PRIVATE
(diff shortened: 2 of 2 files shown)


burmako:
Thank you for the fixes! I'll take a look later today.

stablehlo/tools/CMakeLists.txt
```diff
@@ -38,14 +38,14 @@ mlir_check_all_link_libraries(stablehlo-opt)
...
 target_link_libraries(stablehlo-interpreter PRIVATE
-  MLIRDataLayoutInterfaces
   MLIRFuncDialect
   MLIRIR
   MLIRSupport
   MLIRTranslateLib
   StablehloOps
   StablehloReferenceInterpreter
   StablehloReferenceTensor
+  MLIRDataLayoutInterfaces    # StablehloOps references MLIRDataLayoutInterfaces
```
GleasonK, line 48: Order dependent includes/dependencies usually indicates a problem elsewhere. Not sure if that rule applies to cmake files.

It is unclear why `stablehlo/tools/CMakeLists.txt` needs `MLIRDataLayoutInterfaces`, it looks like the only use of DataLayout is in `StablehloOps` as noted.

Could we move this to the dependencies of `StablehloOps` in `stablehlo/dialect/CMakeLists.txt`?

```
add_mlir_dialect_library(StablehloOps
  PARTIAL_SOURCES_INTENDED
  StablehloOps.cpp
  StablehloBytecode.cpp

  DEPENDS
  StablehloOpsIncGen

  LINK_LIBS PUBLIC
  StablehloBase
  MLIRArithmeticDialect
  MLIRDataLayoutInterfaces
  MLIRIR
  ...
)
```

I tried building with this change and everything worked locally.

GleasonK: Nice! Glad that worked, marking as resolved. I'll leave the final approval to @burmako. 

Appreciate the fix!

stablehlo/dialect/CMakeLists.txt
```diff
@@ -34,6 +34,11 @@ add_mlir_library(StablehloBase
   MLIRSupport
 )
 
+target_include_directories(StablehloBase INTERFACE
+  $<BUILD_INTERFACE:${STABLEHLO_SOURCE_DIR}>
+  $<BUILD_INTERFACE:${STABLEHLO_BINARY_DIR}>
+)
```
burmako, line 40: From our conversation on #torch-mlir, I understand this is not an urgent PR, so let's spend a little bit more time iterating on it. I'd like to better understand the requirements for `target_include_directories`.
  * On the one hand, in MLIR-HLO we are also depending on StableHLO and have the same use case, but we're addressing this differently - by [using `include_directories`](https://github.com/tensorflow/mlir-hlo/blob/6a9a3ce04f8c01cc0c7bd7d574f339f73706149b/CMakeLists.txt#L137-L139) on the MLIR-HLO side.
  * IREE is doing something similar as well for their StableHLO dependency in CMake ([link](https://github.com/iree-org/iree/blob/094ec6d183e769893ae6e9ee98e668712a19dd24/build_tools/third_party/mlir-hlo/CMakeLists.txt#L22-L25)).
  * On the other hand, `target_include_directories` is not unprecedented in MLIR-HLO, e.g. [it's also used](https://github.com/tensorflow/mlir-hlo/blob/6a9a3ce04f8c01cc0c7bd7d574f339f73706149b/lib/Dialect/mhlo/IR/CMakeLists.txt#L52-L56) for the `MhloDialect` target.

@ashay Can you talk a bit more on the choice of targets - `StablehloBase` and `StablehloBroadcastUtils`? Do you anticipate you'd want `target_include_directories` for more targets in the future? Also what do you think about potentially using `include_directories` instead?

burmako (approved):
Looks great! Thanks again for your patience and for taking the time to share the issue that you ran into and the fix for it 👏

PR #150, opened 2022-09-19
Switch CHLO and StableHLO to kEmitAccessorPrefix_Both

burmako:
https://github.com/openxla/stablehlo/commit/3511b6b2810d2f183ecf2e10c2441cc719874838 was supposed to do this, but I forgot to update TableGen files.

### DIFF
--- a/stablehlo/dialect/ChloOps.td
+++ b/stablehlo/dialect/ChloOps.td
@@ -54,7 +54,7 @@ def CHLO_Dialect : Dialect {
     dialects.
   }];
 
-  let emitAccessorPrefix = kEmitAccessorPrefix_Raw;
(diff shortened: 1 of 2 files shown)


tpopp (approved):
Thanks!

GleasonK (approved):
I think the commit name is off? Switch it away from using Raw, not to using it.

Also I'm guessing `Both` isn't the final desired state, is there a task for migrating from `Both->Prefixed`?

burmako:
Nice catch of the typo in the commit name! 

"Also I'm guessing Both isn't the final desired state, is there a task for migrating from Both->Prefixed?". Yeah, Tres is looking into that and has an internal CL on the MLIR-HLO side that I was planning to upstream here later this week.

PR #153, opened 2022-09-19
Add type inference for stablehlo.dot_general

### LINKED ISSUE #152: Type inference for stablehlo.dot_general

ghpvnist:
Add type inference for `stablehlo.dot_general` and fix trailing spaces

Note: a test for element type mismatch does not exist due to quantization for dot, hence why `SameOperandsAndResultElementType` interface is not present. 

closes #152

### DIFF
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -1068,43 +1068,6 @@ ParseResult parsePrecisionConfig(OpAsmParser& parser, mlir::ArrayAttr& attr) {
 // DotGeneralOp
 //===----------------------------------------------------------------------===//
 
-LogicalResult DotGeneralOp::reifyReturnTypeShapes(
-    OpBuilder& builder, ValueRange operands,
-    SmallVectorImpl<Value>& reifiedReturnShapes) {
-  auto lhsType = getLhs().getType().dyn_cast<ShapedType>();
-  auto rhsType = getRhs().getType().dyn_cast<ShapedType>();
-  if (!lhsType || !rhsType) {
-    return failure();
-  }
-
-  Adaptor adaptor(operands);
-  auto dimNumbers = getDotDimensionNumbers();
-  SmallVector<Value> dimensions;
-  for (const int64_t lhsDim : dimNumbers.getLhsBatchingDimensions()) {
-    dimensions.push_back(
-        builder.create<tensor::DimOp>(getLoc(), adaptor.getLhs(), lhsDim));
-  }
-
-  for (int64_t i = 0; i < lhsType.getRank(); i++) {
-    if (!llvm::is_contained(dimNumbers.getLhsContractingDimensions(), i) &&
-        !llvm::is_contained(dimNumbers.getLhsBatchingDimensions(), i)) {
-      dimensions.push_back(
-          builder.create<tensor::DimOp>(getLoc(), adaptor.getLhs(), i));
-    }
-  }
-  for (int64_t i = 0; i < rhsType.getRank(); i++) {
-    if (!llvm::is_contained(dimNumbers.getRhsContractingDimensions(), i) &&
-        !llvm::is_contained(dimNumbers.getRhsBatchingDimensions(), i)) {
-      dimensions.push_back(
-          builder.create<tensor::DimOp>(getLoc(), adaptor.getRhs(), i));
-    }
-  }
-
-  reifiedReturnShapes.push_back(
-      builder.create<tensor::FromElementsOp>(getLoc(), dimensions));
-  return success();
-}
-
 LogicalResult DotGeneralOp::inferReturnTypeComponents(
     MLIRContext*, Optional<Location> location, ValueShapeRange operands,
     DictionaryAttr attributes, RegionRange regions,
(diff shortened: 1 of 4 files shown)


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -1168,6 +1168,48 @@ LogicalResult DotGeneralOp::reifyReturnTypeShapes(
   return success();
 }
 
+LogicalResult DotGeneralOp::inferReturnTypes(
```
zhouxin913, line 1171: will `inferReturnTypeComponents` instead of `inferReturnTypes` work here?

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -1168,6 +1168,48 @@ LogicalResult DotGeneralOp::reifyReturnTypeShapes(
...
+    MLIRContext*, Optional<Location>, ValueRange operands,
+    DictionaryAttr attributes, RegionRange regions,
+    SmallVectorImpl<Type>& inferredReturnTypes) {
+  DotGeneralOp::Adaptor adaptor(operands, attributes, regions);
```
zhouxin913, line 1175: Hey, according to the verifiiers/type inference plan (Will share the RFC in Github later) how about we merge verify() in to this infer function? It can be done by this:
1. Remove `let hasVerifier = 1;` in `stablehlo/dialect/StablehloOps.td`;
2. Remove `DotGeneralOp::verify()` and copy all the code below this "adaptor(...)" line;
3. Change `this->dot_dimension_numbers()` -> `adaptor.dot_dimension_numbers()` and similar for others, because this is an static method and there's no `this` op.
4. Change `return emitOpError() << a << b << c...` to `return emitOptionalError(location, a, b, c...)
I think no changes need for tests or others. Please let me know any questions you may have.

stablehlo/tests/infer_stablehlo.mlir
```diff
@@ -300,6 +300,23 @@ func.func @triangular_solve(%arg0: tensor<10x5x4x4xf32>, %arg1: tensor<10x5x4x4x
   func.return %1 : tensor<10x5x4x4xindex>
 }
 
+// -----
+
+// CHECK-LABEL: func @dot_general
+func.func @dot_general(%arg0: tensor<2x3x4xf32>, %arg1: tensor<2x3x5xf32>) -> tensor<2x4x5xindex> {
```
zhouxin913, line 306: You may test this one separately by:
1. Make new file `/tmp/local_test.mlir` with only this test inside it.
2. after build successfully, run it by
```
test_mlir=/tmp/local_test.mlir

./build/bin/stablehlo-opt $test_mlir \
-verify-diagnostics -split-input-file -allow-unregistered-dialect
```
It will output the real output so that you can check that with your `// CHECK: ...` commends.

dont's forget to add `// -----` before and after each test to split with is necessary. 
After you find your `FileCheck`, the full test command will be:
```
test_mlir=/tmp/local_test.mlir

./build/bin/stablehlo-opt $test_mlir \
-verify-diagnostics -split-input-file -allow-unregistered-dialect \
| FileCheck $test_mlir
```
Feel free to have a gvc with me and debug this together.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -1029,37 +1029,77 @@ LogicalResult DotOp::verify() {
 // DotGeneralOp
 //===----------------------------------------------------------------------===//
 
-LogicalResult DotGeneralOp::verify() {
-  auto dimNumbers = this->dot_dimension_numbers();
+LogicalResult DotGeneralOp::reifyReturnTypeShapes(
```
burmako, line 1032: I see that originally we had `verify` which was followed by `reifyReturnTypeShapes`. Now we have `reifyReturnTypeShapes` which is followed by `inferReturnTypeComponents`.

Let's reorder the function, so that `reifyReturnTypeShapes` comes second again. This will allow the diff to be more meaningful.

burmako (changes requested):

zhouxin913 (changes requested):
We need at least 2 new negative tests for the new  `inferConditionalReturnTypeComponents()` to test whether the "inferred type" match "real result type". You may add:
1. A positive test (similar as your code in `infer_stablehlo.mlir`): 
```
func.func @dot_general(%arg0: tensor<2x3x4xf32>, %arg1: tensor<2x3x5xf32>) 
-> tensor<2x4x5xf32>
```
2. A negative case which mismatch the element type of real result (i64 for example)
```
func.func @dot_general(%arg0: tensor<2x3x4xf32>, %arg1: tensor<2x3x5xf32>) 
-> tensor<2x4x5xi64>
```
3. A negative case which mismatch the shape of the real result
```
func.func @dot_general(%arg0: tensor<2x3x4xf32>, %arg1: tensor<2x3x5xf32>)
-> tensor<2x4x6xf32>
```
Add add proper test error message for the (2) and (3) negative test cases, which will look like: https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/ops_stablehlo.mlir#L4541
`inferred type(s) 'XXX' are incompatible with return type(s) of operation XXX`

zhouxin913 (changes requested):

zhouxin913 (approved):

burmako (approved):

PR #155, opened 2022-09-19
adding links to elaborative testing

### LINKED ISSUE #36: Device better ways to link examples to an op specification
### Request description

Currently, the spec inline examples, demonstrating the working of an op, as follows:

```c
// %x: [[1, 2], [3, 4]]
// %y: [[5, 6], [7, 8]]
%z = stablehlo.add %x, %y : tensor<2x2xf32>
// %z: [[6, 8], [10, 12]]
```

In the near future, let's think about the best way of providing these examples. For example, can we just link to the interpreter implementation of the op, so that people can check out the test suite? Or, more generally, let's rename this section to "Implementation" and link to the source code?

### Additional context

_No response_

sdasgup3:
fixes #36 

The indent is to differentiate the added link to the following "Back to Ops" link.

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -231,6 +231,8 @@ the IEEE-754 specification.
 // %result: [[6, 8], [10, 12]]
 ```
 
+&nbsp;[More Examples](https://github.com/sdasgup3/stablehlo/blob/main/stablehlo/tests/interpret_add.mlir)
+
 [Back to Ops](#index-of-documented-ops)
 
 ## stablehlo.and
@@ -357,6 +359,8 @@ Produces a `result` tensor from a constant `value`.
 %result = stablehlo.constant dense<[(0.0, 1.0), (2.0, 3.0)]> : tensor<2xcomplex<f32>>
 // %result: [(0.0, 1.0), (2.0, 3.0)]
 ```
+
+&nbsp;[More Examples](https://github.com/sdasgup3/stablehlo/blob/main/stablehlo/tests/interpret_constant.mlir)
 
 [Back to Ops](#index-of-documented-ops)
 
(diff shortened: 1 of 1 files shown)


ghpvnist (approved):

docs/spec_draft.md
```diff
@@ -229,6 +229,8 @@ the IEEE-754 specification.
 // %result: [[6, 8], [10, 12]]
 ```
 
+&nbsp;[More Examples](https://github.com/sdasgup3/stablehlo/blob/main/stablehlo/tests/interpret_add.mlir)
```
burmako, line 232: Looks like this link is pointing to your fork of the repo :) How about using a relative link instead? I think `stablehlo/tests/interpret_add.mlir` will work correctly.

docs/spec_draft.md
```diff
@@ -229,6 +229,8 @@ the IEEE-754 specification.
...
+
```
burmako, line 233: Also what do you think about turning "Examples` into a link? We don't really need to keep examples in the markdown file once we have exhaustive tests, right?

burmako (changes requested):

docs/spec_draft.md
sdasgup3: We already thought about this and here are our thoughts:

We want to keep a short example right in the spec to demonstrate, using a trivial (no corner cases) example, the functioning of the op and ask the reader to follow the link for more detailed tests. I believe having that quick-example-view provides better experience.  Having said that, I agree that the example section of spec should be concise: One representative should be enough to capture the high level behavior unless the there are multiple behaviors like bitwise ops.

What do you think?

burmako: " believe having that quick-example-view provides better experience". Sounds good, let's give this a try and see how it works out.

burmako (approved):

PR #160, opened 2022-09-20
Move warning and error flags into CI build flags.

GleasonK:
Per Mehdi's comments on #137. It is safer to enforce warnings as errors in an environment that we have full control over.


burmako (approved):
LGTM, but let's also wait for a review from Mehdi.

joker-eph (approved):

PR #161, opened 2022-09-20
Pretty print for ops with single enum attribute

### LINKED ISSUE #25: Enhance printing of ops with single enum attribute
The following ops have a single enum attribute:

```
DotOp, RngBitGeneratorOp, RngOp
```

These ops can simply inline the enum value:

```
%1 = "stablehlo.dot"(%0, %arg1) {precision_config = [#mhlo<precision DEFAULT>, #mhlo<precision DEFAULT>]} : (tensor<4xf32>, tensor<4xf32>) -> tensor<f32>
%2, %3 = "stablehlo.rng_bit_generator"(%arg0) {rng_algorithm = #mhlo.rng_algorithm<PHILOX>} : (tensor<1x2xf32>) -> (tensor<1x2xf32>, tensor<2x2xui32>)
%4 = "stablehlo.rng"(%arg0, %arg1, %0) {rng_distribution = #mhlo.rng_distribution<NORMAL>} : (tensor<f32>, tensor<f32>, tensor<3xi64>) -> tensor<2x3x5xf32>
==>
%1 = stablehlo.dot %arg0, %arg1, precision = [DEFAULT, DEFAULT] : tensor<4xf32> -> tensor<f32>
%2, %3 = stablehlo.rng_bit_generator %arg0, algorithm = PHILOX : (tensor<3xui64>) -> (tensor<3xui64>, tensor<2x2xui32>)
%4 = stablehlo.rng %arg0, %arg1, %0, distribution = NORMAL : (tensor<f32>, tensor<f32>, tensor<3xi64>) -> tensor<2x3x5xf32>
```

GleasonK:
See #25 for information, or view test updates for examples of new prettyprint.

Implements prettyprint for the following ops:

```
DotOp, RngBitGeneratorOp, RngOp
```

Closes #25


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -1025,6 +1025,63 @@ LogicalResult DotOp::verify() {
   return success();
 }
 
+// Certain enum values have custom printers that include `<` `>`
+// i.e. `<PHILOX>` This method parses the enum value without `<` `>`.
```
joker-eph, line 1029: Seems like this should all be handled by default, can you file a bug upstream (or better: fix the ODS infra for having this work natively?)

joker-eph: I asked Jeff and he confirmed to me that this is supposed to be the default but it is explicitly overridden in StableHLO here: https://github.com/openxla/stablehlo/blob/b08a70ced3883994b8c410a6439f7114f4325d85/stablehlo/dialect/StablehloEnums.td#L182

joker-eph: I guess this is what you meant in the comment here actually maybe? I didn't get it earlier, but then why do we have this just to "undo" it later?

GleasonK: Pasting in a conversation from an offline thread from while back:

> Mehdi Amini
> These should display `#mhlo.rng_algorithm<PHILOX>` instead, I suspect there is a bug for this?
>
> Kevin Gleason
> That was something I changed then reverted. This paste is outdated.
>
> It still says `#mhlo.rng_algorithm<PHILOX>` now, only problem is that when inlined it reads `mhlo.rng_bit_generator %arg0, <PHILOX>` including angle brackets due to the custom assembly format for the enum.
>
> Thinking to either accept that, or make a custom printer/parser for these enum values? Is there a different/better way?

This display is inherited from the MHLO bootstrap. Adding ``"`<` $value `>`"`` causes the generic print to display `#stablehlo.rng_algorithm<PHILOX>` instead of `#stablehlo<comparison_direction LT>`. I'd be happy to undo this and then pretty print without custom machinery. These two `rng_` attrs are the only enum attrs that use this display, all the rest use the style of `comparison_direction`. It is easier and more consistent that way, but I'm not sure if there is any historic reason why we would want `rng_*` to display this way.

stablehlo/tests/print_types_invalid.mlir
```diff
@@ -172,4 +212,4 @@ func.func @reduce_precision_overflow_int32_m(%arg0: tensor<3x4xf32>) -> (tensor<
   // expected-error @+1 {{custom op 'stablehlo.reduce_precision' unable to parse mantissa '2147483648'}}
   %0 = stablehlo.reduce_precision %arg0, format = e1m2147483648 : tensor<3x4xf32>
   func.return %0 : tensor<?x?xf64>
-}
+}
```
burmako, line 215: Typo: no newline at the end of file?

stablehlo/dialect/StablehloOps.cpp
burmako: It looks like the `RngAlgorithm` assemblyFormat was introduced in https://github.com/tensorflow/tensorflow/pull/55752 where the author seems to have run into difficulties with default prettyprinting. `RngDistribution` was added a month later, and I think it just followed whatever `RngAlgorithm` did.

Overall, I think we should try to migrate `RngAlgorithm` and `RngDistribution` to be consistent with other attributes.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -1025,6 +1025,63 @@ LogicalResult DotOp::verify() {
...
+}
+
+// PrecisionConfig - Optional attribute, print the array as raw enums
+//
+// {precision_config = [#stablehlo<precision DEFAULT>,
+//                      #stablehlo<precision DEFAULT>]}
+// ==> ..., precision = [DEFAULT, DEFAULT]
+void printPrecisionConfig(OpAsmPrinter& p, Operation*,
+                          llvm::Optional<mlir::ArrayAttr> attrOpt) {
+  if (!attrOpt.has_value() || !attrOpt.value()) return;
```
burmako, line 1052: I don't think we should check for `!attrOpt.value()`. I would be surprised if an empty attribute (not an empty option, but an empty attribute) will be able to pass verification and end up here.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -1025,6 +1025,63 @@ LogicalResult DotOp::verify() {
...
+    return success();  // No precision config specified
+  }
+
+  if (failed(parser.parseKeyword("precision")) || failed(parser.parseEqual()))
+    return failure();
+
+  SmallVector<Attribute> attrs;
+  if (failed(
+          parser.parseCommaSeparatedList(AsmParser::Delimiter::Square, [&]() {
+            attrs.emplace_back();
```
burmako, line 1072: `attrs.emplace_back()` returns the element that's been emplaced. Given that, can we replace `attrs.back()` below with `attrs.emplace_back()`?

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -1025,6 +1025,63 @@ LogicalResult DotOp::verify() {
...
+            return parseRawEnum<PrecisionAttr>(
+                parser, attrs.back(), "precision", [](llvm::StringRef enumStr) {
+                  return symbolizePrecision(enumStr);
+                });
```
burmako, line 1076: Can we use builtin functionality for this, i.e. `PrecisionAttr::parse(parser, {})`? (Same question below).

burmako (changes requested):

stablehlo/dialect/StablehloOps.cpp
GleasonK: Yes, and the error messages are _way_ better if we go this route.

stablehlo/dialect/StablehloOps.cpp
GleasonK: Removed custom printing and fixed test cases. Will wait on feedback before submit. The integration work required to port this to MHLO as well is not too bad, ~25 test files in the TF repo it looked like.

stablehlo/dialect/StablehloOps.cpp
GleasonK: Fixed this for `PrecisionAttr`. Code was removed for the other custom attr parsers for RNG enums - can be directly inlined in ODS once we removed the `<` `>`.

stablehlo/dialect/StablehloOps.cpp
GleasonK: I added that out of necessity (crashed without). Dug into the generated code a bit to see what was going on:

```
::mlir::ArrayAttr DotOp::getPrecisionConfigAttr() {
  return ::mlir::impl::getAttrFromSortedRange((*this)->getAttrs().begin() + 0, (*this)->getAttrs().end() - 0, getPrecisionConfigAttrName()).dyn_cast_or_null<::mlir::ArrayAttr>();
}

::llvm::Optional< ::mlir::ArrayAttr > DotOp::getPrecisionConfig() {
  auto attr = getPrecisionConfigAttr();
  return attr ? ::llvm::Optional< ::mlir::ArrayAttr >(attr) : (::llvm::None);
}

void DotOp::print(::mlir::OpAsmPrinter &_odsPrinter) {
  _odsPrinter << ' ';
  _odsPrinter << getLhs();
  _odsPrinter << ",";
  _odsPrinter << ' ';
  _odsPrinter << getRhs();
  printPrecisionConfig(_odsPrinter, *this, getPrecisionConfigAttr());
  _odsPrinter.printOptionalAttrDict((*this)->getAttrs(), /*elidedAttrs=*/{"precision_config"});
  _odsPrinter << ' ' << ":";
  _odsPrinter << ' ';
  _odsPrinter.printFunctionalType(getOperation()->getOperandTypes(), getOperation()->getResultTypes());
}
```

This is expected [according to documentation](https://mlir.llvm.org/docs/OpDefinitions/#custom-directives). I had misread the `Attribute Variables > Optional` section. The `ArrayAttr` gets passed to the print function as null. I was accidentally creating an optional of null with this function signature.

Fixed to use:

```
void printPrecisionConfig(OpAsmPrinter& p, Operation*, mlir::ArrayAttr attr) {
   if (!attr) return;
  // ....
}
```

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -1069,11 +1069,9 @@ ParseResult parsePrecisionConfig(OpAsmParser& parser, mlir::ArrayAttr& attr) {
...
   if (failed(
           parser.parseCommaSeparatedList(AsmParser::Delimiter::Square, [&]() {
-            attrs.emplace_back();
-            return parseRawEnum<PrecisionAttr>(
-                parser, attrs.back(), "precision", [](llvm::StringRef enumStr) {
-                  return symbolizePrecision(enumStr);
-                });
+            attrs.push_back(PrecisionAttr::parse(parser, {}));
+            if (!attrs.back()) return failure();
+            return success();
```
burmako, line 1074: This could be simplified to `return success(/*isSuccess=*/ attrs.back())`.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -1025,6 +1025,61 @@ LogicalResult DotOp::verify() {
...
+  llvm::StringRef enumStr;
+  if (failed(parser.parseOptionalKeyword(&enumStr)))
+    return parser.emitError(loc) << "expected " << errStr << " enum";
+  auto enumValOpt = symbolizeFn(enumStr);
+  if (!enumValOpt.has_value())
+    return parser.emitError(loc)
+           << "invalid " << errStr << " enum value " << enumStr;
+  enumAttr = EnumAttrKind::get(parser.getContext(), enumValOpt.value());
+  return success();
+}
```
burmako, line 1043: Looks like this function is no longer used?

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -1025,6 +1025,61 @@ LogicalResult DotOp::verify() {
...
+
+// PrecisionConfig - Optional attribute, print the array as raw enums
+//
+// {precision_config = [#stablehlo<precision DEFAULT>,
+//                      #stablehlo<precision DEFAULT>]}
+// ==> ..., precision = [DEFAULT, DEFAULT]
+void printPrecisionConfig(OpAsmPrinter& p, Operation*,
+                          ::mlir::ArrayAttr attrArr) {
+  if (!attrArr) return;
```
burmako, line 1052: Perhaps add a brief comment for why this check is needed?

burmako (changes requested):

stablehlo/dialect/StablehloOps.cpp
GleasonK: `return success(/*isSuccess=*/ attrs.back())` dispatches to the impl of `success` that returns a `FailureOr` and fails type checking. Used `return success(/*isSuccess=*/ bool(attrs.back()))` since Attribute has an explicit conversion to bool implemented.

PR #162, opened 2022-09-21
Add interpreters for cosine and tanh op

### LINKED ISSUE #143: Interpreter for cosine op

### LINKED ISSUE #144: Interpreter for tanh op

ghpvnist:
closes #143
closes #144

### DIFF
--- a/stablehlo/reference/Element.cpp
+++ b/stablehlo/reference/Element.cpp
@@ -213,36 +213,9 @@ Element floor(const Element &e) {
 }
 
 Element cosine(const Element &e) {
-  Type type = e.getType();
-  if (isSupportedFloatType(type)) {
-    APFloat val = getFloatValue(e);
-    const llvm::fltSemantics &oldSemantics = val.getSemantics();
-    APFloat cosVal(std::cos(val.convertToDouble()));
-    bool roundingErr;
-    cosVal.convert(oldSemantics, APFloat::rmNearestTiesToEven, &roundingErr);
-    return Element(type, FloatAttr::get(type, cosVal));
-  } else if (isSupportedComplexType(type)) {
-    auto complex = getComplexValue(e);
-    Type elementType = type.cast<ComplexType>().getElementType();
-    const llvm::fltSemantics &oldSemantics = complex.real().getSemantics();
-    std::complex<double> complexVal(complex.real().convertToDouble(),
-                                    complex.imag().convertToDouble());
-    std::complex<double> cosVal = std::cos(complexVal);
-    bool roundingErr;
-    APFloat realAPF(cosVal.real());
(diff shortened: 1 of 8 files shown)


sdasgup3 (approved):

sdasgup3 (approved):

stablehlo/reference/Element.cpp
```diff
@@ -153,6 +186,39 @@ Element sine(const Element &e) {
...
+                   ArrayAttr::get(type.getContext(),
+                                  {FloatAttr::get(elementType, realAPF),
+                                   FloatAttr::get(elementType, imagAPF)}));
+  }
+
+  // Report error.
+  auto err = invalidArgument("Unsupported element type: %s",
+                             debugString(type).c_str());
+  report_fatal_error(std::move(err));
+}
```
burmako, line 220: Can we avoid writing duplicate code for this? Let's find a way to share as much as possible between the implementations of sine/cosine/tanh in `Element.cpp`. One potential way of going about this would be to define a function with an interface along the lines of:
```
withUpcastToDouble(element,
  [](double value) { return std::tanh(value); },
  [](std::complex<double> value) { return std::tanh(value); })
```

burmako (changes requested):

burmako (approved):

stablehlo/reference/Element.cpp
```diff
@@ -81,6 +81,41 @@ std::complex<APFloat> getComplexValue(Element element) {
                                arryOfAttr[1].cast<FloatAttr>().getValue());
 }
 
+Element calcWithUpcastToDouble(
```
sdasgup3, line 84: Can we add a comment on this function? 
```
The function applies calcFloat or calcComplex on the constant value represented by the Element object `e`, depending on its type - floating point or complex  resp., after promoting the value to double.     
```

sdasgup3 (changes requested):

stablehlo/reference/Element.cpp
burmako, line 84: Another thing that is interesting about related work is that function parameters are not limited to just functions but are instead callables. I think this extra bit of generality would be useful here as well.

burmako (changes requested):
I did some further thinking about the infrastructure part of this PR, in relationship with other PRs in the queue, and put together a generalization in https://github.com/openxla/stablehlo/pull/190. PTAL, and let's discuss there if the proposal makes sense.

stablehlo/reference/Element.cpp
ghpvnist: Thanks! I've updated PR in response to #190

burmako (approved):

sdasgup3 (approved):

PR #167, opened 2022-09-21
Use multi-dimensional indices to access tensor elements

sdasgup3:
Tensor class currently supports APIs like `Tensor::get(int64_t index)` and `Tensor::set(int64t index, const Element &e)` which provides linearized access to the underlying storage, that is, the `index` points directly to the underlying storage byte array. This kind of "raw" indexing on tensor was done because, that way, it was simple to bootstrap the interpreter,  and secondly, it works pretty well for the element-wise operations. We understood that with more complex ops these APIs might need to be revised, which is what this PR is about.

Instead of using the raw indexing, we can use a more intuitive logical multi-dimensional indices based on the shape of the tensor. Internally we can use dimension [strides](https://en.wikipedia.org/wiki/Stride_of_an_array) to map the muiti-dimensional index to the linearlized raw index. This has the following benefits:

1. Logical indexing is natural  and more intuitive than raw index.
2. Hides the implementation details (how elements of a tensor are actually laid out in memory) from Tensor APIs.

### DIFF
--- a/docs/interpreter.md
+++ b/docs/interpreter.md
@@ -9,7 +9,9 @@
 stores the `mlir::ShapedType` of the tensor along with a contiguous byte array
 representing its data laid out in
 [major-to-minor order](https://www.tensorflow.org/xla/shapes). `detail::Buffer`
-objects are reference-counted to simplify memory management.
+objects are reference-counted to simplify memory management. A `Tensor` object
+also stores [stride](https://en.wikipedia.org/wiki/Stride_of_an_array) for each
+dimension to interpret the underlying 1-D storage as a multi-dimensional tensor.
 
 Individual elements of a tensor are represented using `Element` class which uses
 `mlir::Attribute` for storage. Using `mlir::Attribute` simplifies things because
@@ -17,15 +19,15 @@ this means that we don't have to implement our own machinery for storing values
 of different types inside `Element`.
 
 `Tensor` class has the following APIs to interact with its individual elements:
-  - `Element Tensor::get(int64_t index)`: To extract an individual tensor
-  element at index `index` as `Element` object.
-  - `void Tensor::set(int64_t index, Element element);`: To insert an `Element`
-  object `element` into a tensor at index `index`.
+  - `Element Tensor::get(const std::vector<int64_t> &indices)`: To extract an
+     individual tensor element at multi-dimensional index `indices` as `Element`
+     object.
+  - `void Tensor::set(const std::vector<int64_t> &indices, Element element);`:
+  To insert an `Element` object `element` into a tensor at multi-dimensional
+  index `indices`.
 
-For the skeleton of the interpreter, we chose the above linearized APIs for
-accessing tensor elements to simplify the implementation. This is sufficient for
-element-wise ops, and in the future we're planning to expand this to cover more
-complicated op.
+The above multi-dimensional indices are mapped to the underlying 1-D storage of
+a tensor using the strides, as mentioned above.
 
 ## Working of the interpreter
 
@@ -52,9 +54,15 @@ an `Element` object, is stored in the final `result` tensor.
 ```C++
 Tensor eval(AddOp op, const Tensor &lhs, const Tensor &rhs) {
   Tensor result(op.getType());
-  for (auto i = 0; i < lhs.getNumElements(); ++i) {
-    result.set(i, lhs.get(i) + rhs.get(i));
+
+  // An iterator over the indices of a tensor.
+  DimensionIterator dimIter(lhs.getType().getShape());
+
+  while (dimIter.canIncrement()) {
+    auto nextIndex = dimIter.getNextIndex();
+    result.set(nextIndex, lhs.get(nextIndex) + rhs.get(nextIndex));
   }
+
   return result;
 }
 ```
--- a/docs/interpreter.md
+++ b/docs/interpreter.md
@@ -9,9 +9,7 @@
 stores the `mlir::ShapedType` of the tensor along with a contiguous byte array
 representing its data laid out in
 [major-to-minor order](https://www.tensorflow.org/xla/shapes). `detail::Buffer`
-objects are reference-counted to simplify memory management. A `Tensor` object
-also stores [stride](https://en.wikipedia.org/wiki/Stride_of_an_array) for each
-dimension to interpret the underlying 1-D storage as a multi-dimensional tensor.
+objects are reference-counted to simplify memory management.
 
 Individual elements of a tensor are represented using `Element` class which uses
 `mlir::Attribute` for storage. Using `mlir::Attribute` simplifies things because
@@ -19,15 +17,12 @@ this means that we don't have to implement our own machinery for storing values
 of different types inside `Element`.
 
 `Tensor` class has the following APIs to interact with its individual elements:
-  - `Element Tensor::get(const std::vector<int64_t> &indices)`: To extract an
+  - `Element Tensor::get(llvm::ArrayRef<int64_t> indices)`: To extract an
      individual tensor element at multi-dimensional index `indices` as `Element`
      object.
-  - `void Tensor::set(const std::vector<int64_t> &indices, Element element);`:
+  - `void Tensor::set(llvm::ArrayRef<int64_t> indices, Element element);`:
   To insert an `Element` object `element` into a tensor at multi-dimensional
   index `indices`.
-
-The above multi-dimensional indices are mapped to the underlying 1-D storage of
-a tensor using the strides, as mentioned above.
 
 ## Working of the interpreter
 
@@ -55,13 +50,8 @@ an `Element` object, is stored in the final `result` tensor.
 Tensor eval(AddOp op, const Tensor &lhs, const Tensor &rhs) {
   Tensor result(op.getType());
 
-  // An iterator over the indices of a tensor.
-  DimensionIterator dimIter(lhs.getType().getShape());
-
-  while (dimIter.canIncrement()) {
-    auto nextIndex = dimIter.getNextIndex();
-    result.set(nextIndex, lhs.get(nextIndex) + rhs.get(nextIndex));
-  }
+  for (auto it = lhs.indexBegin(); it != lhs.indexEnd(); ++it)
+    result.set(*it, lhs.get(*it) + rhs.get(*it));
 
   return result;
 }
--- a/docs/interpreter.md
+++ b/docs/interpreter.md
@@ -17,12 +17,12 @@ this means that we don't have to implement our own machinery for storing values
 of different types inside `Element`.
 
 `Tensor` class has the following APIs to interact with its individual elements:
-  - `Element Tensor::get(llvm::ArrayRef<int64_t> indices)`: To extract an
-     individual tensor element at multi-dimensional index `indices` as `Element`
+  - `Element Tensor::get(llvm::ArrayRef<int64_t> index)`: To extract an
+     individual tensor element at multi-dimensional index `index` as `Element`
      object.
-  - `void Tensor::set(llvm::ArrayRef<int64_t> indices, Element element);`:
+  - `void Tensor::set(llvm::ArrayRef<int64_t> index, Element element);`:
   To insert an `Element` object `element` into a tensor at multi-dimensional
-  index `indices`.
+  index `index`.
 
 ## Working of the interpreter
 
@@ -50,7 +50,7 @@ an `Element` object, is stored in the final `result` tensor.
 Tensor eval(AddOp op, const Tensor &lhs, const Tensor &rhs) {
   Tensor result(op.getType());
 
-  for (auto it = lhs.indexBegin(); it != lhs.indexEnd(); ++it)
+  for (auto it = lhs.index_begin(); it != lhs.index_end(); ++it)
     result.set(*it, lhs.get(*it) + rhs.get(*it));
 
   return result;
--- a/stablehlo/reference/Index.cpp
+++ b/stablehlo/reference/Index.cpp
@@ -30,7 +30,7 @@ IndexSpaceIterator &IndexSpaceIterator::operator++() {
   if (!index_)
     llvm::report_fatal_error("Incrementing a past-the-end iterator.");
 
-  if(shape_.empty()) index_.reset();
+  if (shape_.empty()) index_.reset();
 
   for (int64_t i = shape_.size() - 1; i >= 0; --i) {
     (*index_)[i] += 1;
--- a/stablehlo/reference/Index.cpp
+++ b/stablehlo/reference/Index.cpp
@@ -15,32 +15,34 @@ limitations under the License.
 
 #include "stablehlo/reference/Index.h"
 
-#include "llvm/ADT/StringRef.h"
+#include "llvm/Support/Error.h"
 
 namespace mlir {
 namespace stablehlo {
 
-IndexSpaceIterator::IndexSpaceIterator(llvm::ArrayRef<int64_t> shape,
-                                       int64_t counter)
-    : shape_(shape), counter_(counter) {
-  num_elements_ = 1;
-  for (auto shapeElm : shape) num_elements_ *= shapeElm;
-  indices_.resize(shape.size());
+llvm::ArrayRef<int64_t> IndexSpaceIterator::operator*() const {
+  if (!index_)
+    llvm::report_fatal_error("Dereferencing a past-the-end iterator.");
+  return *index_;
 }
 
 IndexSpaceIterator &IndexSpaceIterator::operator++() {
-  if (counter_ == num_elements_)
-    report_fatal_error(llvm::StringRef("Incrementing beyond the last index."));
+  if (!index_)
+    llvm::report_fatal_error("Incrementing a past-the-end iterator.");
+
   for (int64_t i = shape_.size() - 1; i >= 0; --i) {
-    indices_[i] += 1;
-    if (indices_[i] >= shape_[i]) {
-      indices_[i] = 0;
+    (*index_)[i] += 1;
+    if ((*index_)[i] >= shape_[i]) {
+      (*index_)[i] = 0;
+      if (i == 0) {
+        index_.reset();
+        break;
+      }
     } else {
       break;
     }
   }
 
-  counter_++;
   return *this;
 }
 
--- a/stablehlo/reference/Index.cpp
+++ b/stablehlo/reference/Index.cpp
@@ -15,23 +15,22 @@ limitations under the License.
 
 #include "stablehlo/reference/Index.h"
 
+#include "llvm/ADT/StringRef.h"
+
 namespace mlir {
 namespace stablehlo {
 
-IndexSpaceIterator::IndexSpaceIterator(const std::vector<int64_t> &shape,
-                                       bool lastIndex)
-    : shape_(shape) {
-  auto shapeSize = shape.size();
-  if (lastIndex) {
-    counter_ = 1;
-    for (auto shapeElm : shape) counter_ *= shapeElm;
-  } else {
-    counter_ = 0;
-    indices_.resize(shapeSize);
-  }
+IndexSpaceIterator::IndexSpaceIterator(llvm::ArrayRef<int64_t> shape,
+                                       int64_t counter)
+    : shape_(shape), counter_(counter) {
+  num_elements_ = 1;
+  for (auto shapeElm : shape) num_elements_ *= shapeElm;
+  indices_.resize(shape.size());
 }
 
 IndexSpaceIterator &IndexSpaceIterator::operator++() {
+  if (counter_ == num_elements_)
+    report_fatal_error(llvm::StringRef("Incrementing beyond the last index."));
   for (int64_t i = shape_.size() - 1; i >= 0; --i) {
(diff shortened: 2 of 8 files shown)


stablehlo/reference/Tensor.h
```diff
@@ -107,6 +122,53 @@ inline raw_ostream &operator<<(raw_ostream &os, Tensor tensor) {
   return os;
 }
 
+/// Iterator over the indices of a tensor in  major-to-minor order. As an
+/// example, for a tensor with shape [2,3], the iterator enumerates the
+/// following indices (0,0), (0,1), (0,2), (1,0), (1,1), and (1,2) in that
+/// order.
+class DimensionIterator {
```
burmako, line 129: Also, `DimensionIterator` looks sufficiently independent from `Tensor`, so I'd move it into a dedicated target, e.g. `Index.h` and `Index.cpp`.

stablehlo/reference/Tensor.h
```diff
@@ -107,6 +122,53 @@ inline raw_ostream &operator<<(raw_ostream &os, Tensor tensor) {
...
+      : dimensions_(shape), indices_(shape.size()) {
+    if (shape.size() == 0)
+      llvm::report_fatal_error(
+          StringRef("Shape with zero size not supported."));
+    indices_[shape.size() - 1] = -1;
+  }
+
+  /// Get the next index while iterating over the dimensions in major-to-minor
+  /// order.
+  std::vector<int64_t> getNextIndex() const { return indices_; }
```
burmako, line 142: This should be `ArrayRef<int64_t>` to avoid copying.

stablehlo/reference/Tensor.h
```diff
@@ -107,6 +122,53 @@ inline raw_ostream &operator<<(raw_ostream &os, Tensor tensor) {
...
+/// following indices (0,0), (0,1), (0,2), (1,0), (1,1), and (1,2) in that
+/// order.
+class DimensionIterator {
+ public:
+  /// \name Constructors
+  DimensionIterator(std::vector<int64_t> shape)
+      : dimensions_(shape), indices_(shape.size()) {
+    if (shape.size() == 0)
+      llvm::report_fatal_error(
+          StringRef("Shape with zero size not supported."));
```
burmako, line 136: Why not? We need to support 0-dimensional tensors (aka scalars).

stablehlo/reference/Tensor.h
```diff
@@ -99,6 +109,11 @@ class Tensor {
 
  private:
   llvm::IntrusiveRefCntPtr<detail::Buffer> impl_;
+  std::vector<int64_t> strides_;
+
+  // Flattens multi-dimensional index `indices` of a tensor to a linearized
+  // index into the underlying storage using the dimension strides `strides_`.
+  int64_t flattenIndices(const std::vector<int64_t> &indices) const;
```
burmako, line 116: Can we turn this function into a free function in Tensor.cpp? If there's no strong reason for it to be a member function, it shouldn't be.

burmako (changes requested):

stablehlo/reference/Tensor.h
```diff
@@ -99,6 +109,11 @@ class Tensor {
 
  private:
   llvm::IntrusiveRefCntPtr<detail::Buffer> impl_;
+  std::vector<int64_t> strides_;
```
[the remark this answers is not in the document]
sdasgup3: We are computing the strides from the type using a free function `computeStride` defined in Tensor.cpp

stablehlo/reference/Tensor.h
```diff
@@ -67,29 +70,36 @@ class Tensor {
...
 
   /// Assignment operator.
-  Tensor &operator=(const Tensor &other) = default;
+  Tensor &operator=(const Tensor &other);
 
   /// Returns type of the Tensor object.
   ShapedType getType() const;
 
+  /// Updates type of the Tensor object.
+  void setType(ShapedType type);
```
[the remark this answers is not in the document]
sdasgup3: No need for setting the type any more. 

>> By the way, we should probably call out value semantics in the spec as well.

Can you please elaborate a bit?
Did you mean the type cannot be change once the tensor is created?

stablehlo/reference/Tensor.h
sdasgup3: done

stablehlo/reference/Tensor.h
sdasgup3: We keep strides are part of Tensor object as recomputing it every-time we call Tensor::get/Tensor::set seems redundant. With `stride`, `get`, and `set` being part of Tensor class, it makes sense to me for  `flattenIndices` to be a (private) member function as well  as it uses the `strides` to provide index to the internal implementation of Tensor: Any change in the internal implementation might need changes in the function because of its close connection with how a tensor is internally stored.   
WDYT?

docs/interpreter.md
```diff
@@ -9,23 +9,25 @@
...
 `mlir::Attribute` for storage. Using `mlir::Attribute` simplifies things because
 this means that we don't have to implement our own machinery for storing values
 of different types inside `Element`.
 
 `Tensor` class has the following APIs to interact with its individual elements:
-  - `Element Tensor::get(int64_t index)`: To extract an individual tensor
-  element at index `index` as `Element` object.
-  - `void Tensor::set(int64_t index, Element element);`: To insert an `Element`
-  object `element` into a tensor at index `index`.
+  - `Element Tensor::get(const std::vector<int64_t> &indices)`: To extract an
```
burmako, line 22: `ArrayRef<int64_t>`? (Same below).

docs/interpreter.md
```diff
@@ -9,23 +9,25 @@
 stores the `mlir::ShapedType` of the tensor along with a contiguous byte array
 representing its data laid out in
 [major-to-minor order](https://www.tensorflow.org/xla/shapes). `detail::Buffer`
-objects are reference-counted to simplify memory management.
+objects are reference-counted to simplify memory management. A `Tensor` object
+also stores [stride](https://en.wikipedia.org/wiki/Stride_of_an_array) for each
+dimension to interpret the underlying 1-D storage as a multi-dimensional tensor.
```
burmako, line 14: Do we have to: 1) store strides in the `Tensor` instead of computing them on the fly, 2) call them out in the document? Strides are an internal implementation detail of major-to-minor order which doesn't even need to be computed explicitly (its only use is `flattenIndices`, and there you can just have a running counter instead of an entire vector).

docs/interpreter.md
```diff
@@ -9,23 +9,25 @@
...
+  - `void Tensor::set(const std::vector<int64_t> &indices, Element element);`:
+  To insert an `Element` object `element` into a tensor at multi-dimensional
+  index `indices`.
 
-For the skeleton of the interpreter, we chose the above linearized APIs for
-accessing tensor elements to simplify the implementation. This is sufficient for
-element-wise ops, and in the future we're planning to expand this to cover more
-complicated op.
+The above multi-dimensional indices are mapped to the underlying 1-D storage of
+a tensor using the strides, as mentioned above.
```
burmako, line 30: I think this paragraph is redundant with the paragraph above where we're saying that the data is laid out in major-to-minor order.

docs/interpreter.md
```diff
@@ -52,9 +54,15 @@ an `Element` object, is stored in the final `result` tensor.
...
-    result.set(i, lhs.get(i) + rhs.get(i));
+
+  // An iterator over the indices of a tensor.
+  DimensionIterator dimIter(lhs.getType().getShape());
+
+  while (dimIter.canIncrement()) {
+    auto nextIndex = dimIter.getNextIndex();
+    result.set(nextIndex, lhs.get(nextIndex) + rhs.get(nextIndex));
   }
+
```
burmako, line 65: Stale example.

stablehlo/reference/Index.h
```diff
@@ -0,0 +1,59 @@
...
+#include <vector>
+
+#include "llvm/ADT/ArrayRef.h"
+
+namespace mlir {
+namespace stablehlo {
+
+/// Iterator over the indices of a tensor in major-to-minor order. As an
+/// example, for a tensor with shape [2,3], the iterator enumerates the
+/// indices (0,0), (0,1), (0,2), (1,0), (1,1), and (1,2).
```
burmako, line 29: Let's use the terminology from the spec, e.g.

```
Iterates over the index space of a tensor with a given shape, producing indices 
in lexicographical order. As an example, ...
```

Same feedback for the mention of "major-to-minor order" in a comment below. Let's use "lexicographical order" instead.

stablehlo/reference/Index.h
```diff
@@ -0,0 +1,59 @@
...
+class IndexSpaceIterator {
+ public:
+  /// \name Constructors
+  explicit IndexSpaceIterator(const std::vector<int64_t> &shape,
+                              bool lastIndex = false);
+
+  /// Get the current index.
+  llvm::ArrayRef<int64_t> operator*() const { return indices_; }
```
burmako, line 37: I don't think we qualify `ArrayRef` in other files in the interpreter, so let's drop `llvm::` here. (Same below).

stablehlo/reference/Index.h
```diff
@@ -0,0 +1,59 @@
...
+/// Iterator over the indices of a tensor in major-to-minor order. As an
+/// example, for a tensor with shape [2,3], the iterator enumerates the
+/// indices (0,0), (0,1), (0,2), (1,0), (1,1), and (1,2).
+class IndexSpaceIterator {
+ public:
+  /// \name Constructors
+  explicit IndexSpaceIterator(const std::vector<int64_t> &shape,
+                              bool lastIndex = false);
+
+  /// Get the current index.
```
burmako, line 36: "indices"? (Same below).

stablehlo/reference/Index.h
```diff
@@ -0,0 +1,59 @@
...
+  llvm::ArrayRef<int64_t> operator*() const { return indices_; }
+
+  bool operator==(const IndexSpaceIterator &it) {
+    return shape_ == it.shape_ && counter_ == it.counter_;
+  }
+
+  bool operator!=(const IndexSpaceIterator &it) { return !(*this == it); }
+
+  /// Increment to the next valid index while iterating over the dimensions in
```
burmako, line 45: What happens if we increment an iterator which is already at the end? I'd probably `report_fatal_error`. Either way, let's document and implement this.

stablehlo/reference/Index.h
```diff
@@ -0,0 +1,59 @@
...
+  /// major-to-minor order.
+  IndexSpaceIterator &operator++();
+  IndexSpaceIterator operator++(int);
+
+ private:
+  std::vector<int64_t> shape_;
+  std::vector<int64_t> indices_;
+  int64_t counter_;
```
burmako, line 53: Do we have to store the counter? This creates opportunities for unexpected overflows. If we need to compare two iterators, we can just compare indices, right?

stablehlo/reference/Index.h
```diff
@@ -0,0 +1,59 @@
...
+
+  bool operator!=(const IndexSpaceIterator &it) { return !(*this == it); }
+
+  /// Increment to the next valid index while iterating over the dimensions in
+  /// major-to-minor order.
+  IndexSpaceIterator &operator++();
+  IndexSpaceIterator operator++(int);
+
+ private:
+  std::vector<int64_t> shape_;
```
burmako, line 51: `SmallVector`?

stablehlo/reference/Index.h
```diff
@@ -0,0 +1,59 @@
...
+namespace mlir {
+namespace stablehlo {
+
+/// Iterator over the indices of a tensor in major-to-minor order. As an
+/// example, for a tensor with shape [2,3], the iterator enumerates the
+/// indices (0,0), (0,1), (0,2), (1,0), (1,1), and (1,2).
+class IndexSpaceIterator {
+ public:
+  /// \name Constructors
+  explicit IndexSpaceIterator(const std::vector<int64_t> &shape,
```
burmako, line 33: `ArrayRef<int64_t>`.

stablehlo/reference/Index.h
```diff
@@ -0,0 +1,59 @@
...
+                              bool lastIndex = false);
```
burmako, line 34: `lastIndex` is a bit awkward because it's ambiguous with respect to what it means when it's `false`. Does this mean first index? Some other index? We can avoid this ambiguity by passing the `indices` directly.

stablehlo/reference/Tensor.h
```diff
@@ -67,38 +68,45 @@ class Tensor {
...
+  /// underlying storage pointed to by \a indices.
+  void set(llvm::ArrayRef<int64_t> indices, const Element &element);
 
   /// Print utilities for Tensor objects.
   void print(raw_ostream &os) const;
 
   /// Print utilities for Tensor objects.
   void dump() const;
 
+  /// Iterator utils.
```
burmako, line 99: Let's expand this comment beyond just "utils", because there's non-trivial semantics involved here. 

Also, calling things "utilities", e.g. `print` and `dump` above, kind of defeats the purpose of having comments for everything. I'm not sure if such comments buy us anything, except for increasing the line of code count :)

stablehlo/reference/Tensor.h
```diff
@@ -67,38 +68,45 @@ class Tensor {
...
+  IndexSpaceIterator indexBegin() const;
+  IndexSpaceIterator indexEnd() const;
```
burmako, line 101: LLVM's code style has a carveout for iterator methods - https://llvm.org/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly - so these should be called `index_begin` and `index_end`.

stablehlo/reference/Tensor.h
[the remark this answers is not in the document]
burmako: Why do we need to store strides if we don't provide an API to change them? Can we compute them on the fly? This way we'll be sure that they are always what we expect, and this will allow constructors to have default implementations again. 

Also, as a bonus, since strides are computed only for the purposes of flattening indices, you won't even need to have a dedicated function for that. You'll be able to inline this computation in the flattening logic.

stablehlo/reference/Tensor.h
```diff
@@ -67,38 +68,45 @@ class Tensor {
...
+
  private:
   llvm::IntrusiveRefCntPtr<detail::Buffer> impl_;
+  std::vector<int64_t> strides_;
+
+  // Flattens multi-dimensional index `indices` of a tensor to a linearized
+  // index into the underlying storage using the dimension strides `strides_`.
+  int64_t flattenIndices(llvm::ArrayRef<int64_t> indices) const;
```
burmako, line 109: Does this have to be a member function? Can be a free function in an anonymous namespace, similar to `computeStrides`?

stablehlo/reference/Tensor.cpp
```diff
@@ -50,6 +51,24 @@ int64_t getSizeInBytes(Type type) {
   report_fatal_error(std::move(err));
 }
 
+// Computes the stride of a tensor shape: The number of locations in memory
```
burmako, line 54: "strides"?

burmako (changes requested):

stablehlo/reference/Index.h
sdasgup3: You are right. In other right the namespace is made visible via some header file. In this file I need to qualify `ArrayRef` with 
 the namespace to avoid compiler error.

stablehlo/reference/Index.h
sdasgup3: The reason we have the `counter_` is because it also works for the scalar case. 

For the scalar case, both `index_begin` and `index_end` returns iterators with `indices_` member variable set to empty. `it != index_end()` would disallow the the use of the current `indices_` of iterator. 

`counter_` is initially set to 0 and it is supposed to counter till the number of elements which we set to 1 in case of scalars.  `index_begin` sets the `counter_` to 0 and `index_end` sets to to number of elements or 1. `lastIndex` helps us to distinguish the two cases.

I agree that the `lastIndex` might be confusing. Instead I can make all these as arguments to the constructor.
wdyt?

docs/interpreter.md
sdasgup3: removed stride as member variable. updated the document accordingly.

stablehlo/reference/Index.h
sdasgup3: Please see my comment above.

stablehlo/reference/Tensor.h
sdasgup3: done.

stablehlo/reference/Tensor.h
sdasgup3: not appllicable.

stablehlo/reference/Index.h
sdasgup3: Please have a look at the latest upload. Added comments to the member variables. 

In short,
The `index_end` should return an iterator past the end of valid iteration space. The member variable `counter_` helps us achieve that. The member variable `num_element_` help flag increment beyond the last index.

burmako: In the proposed design, `counter_` and `num_elements_` have to be kept in sync with `shape_`, we have to worry about potential overflows - that's a lot of work.

I think it's possible to model the domain more cleanly - without the need for accompanying class members at all. Here's how it could look like: https://github.com/sdasgup3/stablehlo/pull/1.

burmako (changes requested):

docs/interpreter.md
```diff
@@ -17,15 +17,12 @@ this means that we don't have to implement our own machinery for storing values
...
-  - `Element Tensor::get(int64_t index)`: To extract an individual tensor
-  element at index `index` as `Element` object.
-  - `void Tensor::set(int64_t index, Element element);`: To insert an `Element`
-  object `element` into a tensor at index `index`.
-
-For the skeleton of the interpreter, we chose the above linearized APIs for
-accessing tensor elements to simplify the implementation. This is sufficient for
-element-wise ops, and in the future we're planning to expand this to cover more
-complicated op.
+  - `Element Tensor::get(llvm::ArrayRef<int64_t> indices)`: To extract an
```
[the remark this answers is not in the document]
sdasgup3: Updated so as to call it `index` throughout.

stablehlo/reference/Index.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include "stablehlo/reference/Index.h"
+
+#include "llvm/ADT/StringRef.h"
+
+namespace mlir {
+namespace stablehlo {
+
+IndexSpaceIterator::IndexSpaceIterator(llvm::ArrayRef<int64_t> shape,
+                                       int64_t counter)
+    : shape_(shape), counter_(counter) {
```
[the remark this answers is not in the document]
sdasgup3: done.

sdasgup3:
https://github.com/sdasgup3/stablehlo/pull/1.

Added the increment behavior on empty index applicable for scalar tensors. The change is at `IndexSpaceIterator::operator++`

stablehlo/reference/Index.h
```diff
@@ -0,0 +1,80 @@
...
+  }
+  bool operator!=(const IndexSpaceIterator &it) { return !(*this == it); }
+
+  /// Increment to the next index while iterating over the index space
+  /// of a tensor in lexicographical order.
+  /// Incrementing past the last index will result in a past-the-end iterator
+  /// which cannot be dereferenced. Incrementing even further will result in
+  /// a fatal error.
+  /// For scalar tensor, which empty 'shape_', incrementing the index result in
+  /// past-the-end iterator.
```
burmako, line 64: Do we need this last paragraph? I think it's covered by the two paragraphs above? The index space is: `{[]}`, so incrementing past `[]` will result in a past-the-end iterator.

stablehlo/reference/Tensor.cpp
```diff
@@ -50,6 +51,32 @@ int64_t getSizeInBytes(Type type) {
   report_fatal_error(std::move(err));
 }
 
+// Flattens multi-dimensional index 'index' of a tensor to a linearized
+// index into the underlying storage using the dimension strides `strides_`.
+int64_t flattenindex(ArrayRef<int64_t> shape, ArrayRef<int64_t> index) {
```
burmako, line 56: Might be a typo? I think "flattenIndex" reads better, because it's two words. If you're looking for a single word, perhaps consider "linearize"?

stablehlo/reference/Tensor.cpp
```diff
@@ -50,6 +51,32 @@ int64_t getSizeInBytes(Type type) {
   report_fatal_error(std::move(err));
 }
 
+// Flattens multi-dimensional index 'index' of a tensor to a linearized
+// index into the underlying storage using the dimension strides `strides_`.
```
burmako, line 55: I don't think the comment is applicable to this function anymore. "to a linearized index into the underlying storage where elements are laid out in canonical order"?

burmako (changes requested):
A few more typos, and we're good to go. Also, let's change the title of the PR to something like "Use multi-dimensional indices to access tensor elements" and update the description to match?

stablehlo/reference/Index.h
sdasgup3: Agree.

burmako (approved):

Issue #119, opened 2022-09-14
Sync the specification and the ODS description 


sdasgup3:
### Request description

At this point the above sync is missing. Here are a few alternatives:

1. The ODS description of an op to have a hyperlink to the corresponding section of https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md. 
2. The spec of an op to be embedded in the ODS description and stableHLO specification (a markdown file) is auto generated (like https://www.tensorflow.org/mlir/hlo_ops).

sdasgup3:
I propose to go with (1) at the moment and can be implemented only when the spec and ODS description are in sync. 

Later, depending upon how we are going to publish the final spec, to may switch to (2).

sdasgup3:
Closing this ticket as we had a resolution on this as mentioned above.

PR #168, opened 2022-09-21
Populate progress of verifier and type inference.

zhouxin913:
The new values in Markdown table of all 114 ops are copied from StableHLO audit sheets with local handy script and the rule to map the old value in sheets to the new value in Markdown table is:
``` 
convert_value_for_md = {
    'Yes (Declarative)': 'yes(match-xla)',
    'Yes (Imperative)': 'yes(match-xla)',
    'Yes (need revisit)': 'yes(need-revisit)',
    'TODO': 'no',
    'Yes (revisit)': 'yes(need-revisit)',
    'No (Infeasible)': 'no-plan',
    'WIP': 'wip',
}
```
The values of `Type Inference` column is clear and fully up to date. The `Verifier` column actually means both ODS and `verify()`. As `verify()` are removed for a lot of ops, the value of them are still set to `yes(match-xla)`. 
If an op's type inference is done or revisited in the future: its value of `Verify` column will also be revisited and updated at the same time (even if it's set to `yes`) now.
The sum number in the column title includes all kinds of `yes` and `no-plan`.

Please review and comment: I can update the sheet/table accordingly.

### DIFF
--- a/docs/status.md
+++ b/docs/status.md
@@ -20,126 +20,127 @@ one of the following tracking labels.
     - **yes**: complete
     - **wip**: semi-complete: Work in progress or under review.
     - **no**: not complete yet, but part of [the roadmap](https://github.com/openxla/stablehlo#roadmap).
- - Customized labels
-    - Verifier
-       - **match-xla**:  verifier in sync with  [XLA semantics](https://www.tensorflow.org/xla/operation_semantics).
-       - **match-spec**: verifier in sync with [StableHLO semantics](https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md).
-
+ - Customized labels for Verifier and Type Inference 
+    - **yes(match-xla)**:  in sync with  [XLA semantics](https://www.tensorflow.org/xla/operation_semantics).
+    - **yes(match-spec)**: in sync with [StableHLO semantics](https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md).
+    - **yes(need-revisit)**: implemented but need revisit for the sync with XLA or spec
+    - **no-plan**: infeasible or no plan to implement by design
+    
 ## Status
 
-| StableHLO Op (114) | Specification (21) | Verification | Type Inference | Prettyprinting | Interpreter (2) |
+| StableHLO Op (114) | Specification (21) | Verifier(104) | Type Inference(79) | Prettyprinting | Interpreter (2) |
 |:--|:--:|:--:|:--:|:--:|:--:|
-| AbsOp |yes||||no|
-| AddOp |yes|||| yes|
-| AfterAllOp |no||||no |
-| AllGatherOp |no||||no|
-| AllReduceOp |no||||no|
-| AllToAllOp |no||||no|
-| AndOp |yes|||| no|
-| Atan2Op |no||||no|
-| BatchNormGradOp |no||||no|
-| BatchNormInferenceOp |no||||no|
-| BatchNormTrainingOp |no||||no|
-| BitcastConvertOp |no||||no|
-| BroadcastInDimOp |no||||no|
-| BroadcastOp |no||||no|
-| CaseOp |no||||no|
-| CbrtOp |no||||no|
-| CeilOp |yes||||no|
-| CholeskyOp |no||||no|
-| ClampOp |no||||no|
-| ClzOp |no||||no|
-| CollectivePermuteOp |no||||no|
-| CompareOp |no||||no|
-| ComplexOp |no||||no|
-| ComputeReshapeShapeOp |no||||no|
-| ConcatenateOp |no||||no|
-| ConstantOp |yes|||| yes|
-| ConvertOp |no||||no|
-| ConvolutionOp |no||||no|
-| CosineOp |yes||||no|
-| CreateTokenOp |no||||no|
-| CrossReplicaSumOp |no||||no|
-| CstrReshapableOp |no||||no|
-| CustomCallOp |no||||no|
-| DivOp |yes||||no|
-| DotGeneralOp |no||||no|
-| DotOp |no||||no|
-| DynamicBroadcastInDimOp |no||||no|
-| DynamicConvOp |no||||no|
-| DynamicGatherOp |no||||no|
-| DynamicIotaOp |no||||no|
-| DynamicPadOp |no||||no|
-| DynamicReshapeOp |no||||no|
-| DynamicSliceOp |no||||no|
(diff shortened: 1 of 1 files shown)


docs/status.md
```diff
@@ -20,126 +20,127 @@ one of the following tracking labels.
     - **yes**: complete
     - **wip**: semi-complete: Work in progress or under review.
     - **no**: not complete yet, but part of [the roadmap](https://github.com/openxla/stablehlo#roadmap).
- - Customized labels
-    - Verifier
-       - **match-xla**:  verifier in sync with  [XLA semantics](https://www.tensorflow.org/xla/operation_semantics).
-       - **match-spec**: verifier in sync with [StableHLO semantics](https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md).
-
+ - Customized labels for Verifier and Type Inference 
+    - **yes(match-xla)**:  in sync with  [XLA semantics](https://www.tensorflow.org/xla/operation_semantics).
```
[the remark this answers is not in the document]
burmako: Agreed that the current form is pretty verbose. I'd propose a tweak to Kevin's proposal:
  * Change `yes(match-spec)` to `yes`, given that our ultimate goal is to match the spec, and matching XLA is only an intermediate step in that direction.
  * Change `yes(match-xla)` to just `yes*` and explain what it means in the section that talks about customized labels. This will make the table much easier on the eyes, similarly to what Kevin proposes.
  * `yes(need-revisit)` sounds good.
  * Change `no-plan` to `infeasible`. "No plan" can be interpreted in multiple ways, including "it's feasible to implement but we won't".

burmako (changes requested):
It was a lot of work to collect all this information, and it'll be very helpful both as a reference to StableHLO users and as a way to keep track of ongoing work! I have a few stylistic comments.

burmako (changes requested):
Also, as far as `wip` goes, what would be a process to use that status? Do we send a PR to status.md every time someone starts to work on a ticket? That seems a bit too much. Perhaps we could drop `wip` altogether - both from the table and from the labels - and rely on our issue tracker for that?

GleasonK:
Do most of the WIP/Future items have issues? We could place the issue number in those slots if so.

sdasgup3:
>> Do we send a PR to status.md every time someone starts to work on a ticket? That seems a bit too much. 
I agree that WIP would be too much to track using status.md file. 

What should be the expectation on updating the file? After the fix of a issue (that way the PR solving the issue might include the status.md file change) or at regular interval in batches?

burmako:
"What should be the expectation on updating the file?". I'd propose that we don't use WIP at all to reduce the need to send boilerplate PRs which just flip "no" to "wip". 

In that case, `status.md` would only be updated by PRs which update the corresponding code. E.g. if we have a PR that implements type inference for if and case, then this very PR should update `status.md` as well.

zhouxin913:
For the "wip" discussion all above, I totally agree that:
1: dont' use "wip" at all
2. in each PR that fix an op, change the corresponding cell the in `status.md` from "no" to proper "yes/..." value.
Will fix that in next commit. And will label "if/case" op to yes* if https://github.com/openxla/stablehlo/pull/135 is landed first.

docs/status.md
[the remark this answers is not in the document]
zhouxin913: Sounds great, addressed.

docs/status.md
```diff
@@ -18,128 +18,128 @@ one of the following tracking labels.
...
     - **yes**: complete
-    - **wip**: semi-complete: Work in progress or under review.
     - **no**: not complete yet, but part of [the roadmap](https://github.com/openxla/stablehlo#roadmap).
- - Customized labels
-    - Verifier
-       - **match-xla**:  verifier in sync with  [XLA semantics](https://www.tensorflow.org/xla/operation_semantics).
-       - **match-spec**: verifier in sync with [StableHLO semantics](https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md).
-
+ - Customized labels for Verifier and Type Inference 
+    - **yes\***:  in sync with  [XLA semantics](https://www.tensorflow.org/xla/operation_semantics).
```
burmako, line 23: Typo: double whitespace in two spots on this line.

docs/status.md
```diff
@@ -18,128 +18,128 @@ one of the following tracking labels.
...
+    - **yes**: in sync with [StableHLO semantics](https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md).
+    - **yes(need-revisit)**: implemented but need revisit for the sync with XLA or spec
+    - **infeasible**: infeasible to implement by design
+    
 ## Status
 
-| StableHLO Op (114) | Specification (21) | Verification | Type Inference | Prettyprinting | Interpreter (2) |
+| StableHLO Op (114) | Specification (21) | Verifier(104) | Type Inference(82) | Prettyprinting | Interpreter (2) |
```
burmako, line 30: The current column titles are populated using GitHub label names. Let's stick to "Verification"?

burmako (changes requested):

sdasgup3:
In the context of 

>> Change yes(match-spec) to yes, given that our ultimate goal is to match the spec, and matching XLA is only an intermediate step in that direction.

I would bring a point in the context of putting a "yes" for the verifier in a row. Ideally, a "yes" means that the verifier, implemented partly in ODS and using imperative code, and spec are in sync. This also includes the description section of the ODS, which at the moment, very rightly,  has links to HLO semantics. We want that section to  reflect the specification somehow. 

We have a ticket  https://github.com/openxla/stablehlo/issues/119  to track the above and discuses a few alternative to  reflect the specification in the ODS description.

docs/status.md
zhouxin913: Done.

docs/status.md
zhouxin913: Done.

zhouxin913:
> In the context of
> 
> > > Change yes(match-spec) to yes, given that our ultimate goal is to match the spec, and matching XLA is only an intermediate step in that direction.
> 
> I would bring a point in the context of putting a "yes" for the verifier in a row. Ideally, a "yes" means that the verifier, implemented partly in ODS and using imperative code, and spec are in sync. This also includes the description section of the ODS, which at the moment, very rightly, has links to HLO semantics. We want that section to reflect the specification somehow.
> 
> We have a ticket #119 to track the above and discuses a few alternative to reflect the specification in the ODS description.

Great suggestion! let's keep track of this work on your new issue.

burmako (approved):

Issue #169, opened 2022-09-21
Add specification and Interpreter for ReshapeOp


sdasgup3:
### Request description

Add specification and Implement the interpreter for `stablehlo.reshape` op.

PR #170, opened 2022-09-21
Specification and Interpreter for  reshape op

sdasgup3:
fixes https://github.com/openxla/stablehlo/issues/169

~Diff based on https://github.com/openxla/stablehlo/pull/167~
upd: Checked that the spec matches with the verifier of the Op.

### DIFF
--- a/docs/interpreter.md
+++ b/docs/interpreter.md
@@ -9,7 +9,9 @@
 stores the `mlir::ShapedType` of the tensor along with a contiguous byte array
 representing its data laid out in
 [major-to-minor order](https://www.tensorflow.org/xla/shapes). `detail::Buffer`
-objects are reference-counted to simplify memory management.
+objects are reference-counted to simplify memory management. A `Tensor` object
+also stores [stride](https://en.wikipedia.org/wiki/Stride_of_an_array) for each
+dimension to interpret the underlying 1-D storage as a multi-dimensional tensor.
 
 Individual elements of a tensor are represented using `Element` class which uses
 `mlir::Attribute` for storage. Using `mlir::Attribute` simplifies things because
--- a/docs/interpreter.md
+++ b/docs/interpreter.md
@@ -9,9 +9,7 @@
 stores the `mlir::ShapedType` of the tensor along with a contiguous byte array
 representing its data laid out in
 [major-to-minor order](https://www.tensorflow.org/xla/shapes). `detail::Buffer`
-objects are reference-counted to simplify memory management. A `Tensor` object
-also stores [stride](https://en.wikipedia.org/wiki/Stride_of_an_array) for each
-dimension to interpret the underlying 1-D storage as a multi-dimensional tensor.
+objects are reference-counted to simplify memory management.
 
 Individual elements of a tensor are represented using `Element` class which uses
 `mlir::Attribute` for storage. Using `mlir::Attribute` simplifies things because
@@ -23,7 +21,7 @@ of different types inside `Element`.
      individual tensor element at multi-dimensional index `index` as `Element`
      object.
   - `void Tensor::set(llvm::ArrayRef<int64_t> index, Element element);`:
-  To insert an `Element` object `element` into a tensor at multi-dimensional
+  To update an `Element` object `element` into a tensor at multi-dimensional
   index `index`.
 
 ## Working of the interpreter
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -916,6 +916,8 @@ flatenned vector into the shape of  `result`.
 // %result: [[1,2],[3,4],[5,6]]
 ```
 
+&nbsp;[More Examples](../stablehlo/tests/interpret_reshape.mlir)
+
 [Back to Ops](#index-of-documented-ops)
 
 ## stablehlo.rsqrt
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -166,6 +166,7 @@ described below)
    * [not](#stablehlonot)
    * [or](#stablehloor)
    * [remainder](#stablehloremainder)
+   * [reshape](#stablehloreshape)
    * [rsqrt](#stablehlorsqrt)
    * [sine](#stablehlosine)
    * [sqrt](#stablehlosqrt)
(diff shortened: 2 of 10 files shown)


docs/spec_draft.md
```diff
@@ -874,6 +875,46 @@ implementation-defined value.
...
+## stablehlo.reshape
+
+`stablehlo.reshape(operand) -> result`
+
+### Semantics
+
+Performs reshape of `operand` tensor to a `result` tensor. Conceptually, it
+collapses the dimensions of `operand` starting from slowest-varying
+dimension `0` to fastest-varying dimension `rank-1`, and then refines the
+flatenned vector into the shape of  `result`.
```
burmako, line 887: This explanation uses some concepts, e.g. "collapse the dimension", "slowest-varying dimension", "fastest-varying dimension", "flattened vector", "refine into a shape", which haven't been defined above.

Strictly speaking, the "element-wise" concept in "performs element-wise <operation>" which we're using in the current draft is also not defined above, but its fairly intuitive already and easily searchable. However, the concepts used in this explanation are a mix of somewhat searchable things (collapsing a dimension) and outright implementation details (flattening into a vector, refining into a shape).

Let's define op semantics from first principles, using only the definitions in this spec and college maths. Before proceeding with that, let's first expand our vocabulary - something which I took a stab at in #193.

burmako, line 887: With the terminology from #193, this explanation could be rewritten as follows:

```
Performs reshape of `operand` tensor to a `result` tensor. Conceptually, it
amounts to keeping the same elements but potentially changing the shape, e.g. 
from `tensor<2x3xf32>` to `tensor<3x2xf32>` or `tensor<6xf32>`.

More formally, the canonical representation of the `result` tensor must be
the same as the canonical representation of the `operand` tensor, which means
that the element types and the number of elements in these tensors must be the same.
```

stablehlo/reference/Ops.cpp
```diff
@@ -36,6 +36,18 @@ Tensor eval(ConstantOp op, ElementsAttr value) {
   return Tensor(value.cast<DenseElementsAttr>());
 }
 
+Tensor eval(ReshapeOp op, const Tensor &operand) {
+  // The fact that the interpreter stores the tensor internally as contiguous
+  // byte arrays allows to reshape the dimensions of tensor merely by modifying
+  // it's strides, while storing the same internal memory as that of the
+  // operand.
```
burmako, line 43: Similarly to the spec proposal above, let's define op implementations from first principles, i.e. using only the definitions in the spec and APIs from `Tensor.h` and `Element.h` which implement these definitions. 

From that perspective, relying on exact representation of a tensor, i.e. something that the spec treats as implementation-defined (currently, this is implicit, #193 makes this explicit), doesn't sound right.

I think improving on this is fairly straightforward if we use the iterator that you proposed in #167 (with the modifications that I proposed in the comments), for example:

```
Tensor eval(ReshapeOp op, const Tensor &operand) {
  Tensor result(op.getType());
  for (auto operandIt = operand.index_begin(), resultIt = result.index_begin();
       operandIt != operand.index_end();
       ++operandIt, ++resultIt) {
    result.set(*it, operand.get(*it));
  }
  return result;
}
```

burmako (changes requested):

docs/status.md
```diff
@@ -110,7 +110,7 @@ one of the following tracking labels.
 | ReduceWindowOp |no||||no|
 | RemOp |yes||||no|
 | ReplicaIdOp |no||||no|
-| ReshapeOp |no||||no|
+| ReshapeOp |yes||||yes|
```
burmako, line 113: Let's also check whether ReshapeOp's verifier and shape function conform to the newly written spec. If they do, let's flip `yes*` in the table to `yes`. If they don't let's open tickets about this.

docs/spec_draft.md
burmako: Upon further thinking, I don't like the declarative style of the second paragraph. I think that changing it to imperative style would be better:

```
Performs reshape of `operand` tensor to a `result` tensor. Conceptually, it
amounts to keeping the same canonical representation but potentially changing 
the shape, e.g. from `tensor<2x3xf32>` to `tensor<3x2xf32>` or `tensor<6xf32>`.

More formally, `result[i0, ..., iR-1]` is equal to `operand[j0, ..., jR'-1]`
where `i` and `j` have the same position in the lexicographic ordering of
the index spaces of `result` and `operand`.
```

stablehlo/reference/Tensor.h
```diff
@@ -46,6 +46,9 @@ class Buffer : public llvm::RefCountedBase<Buffer> {
   /// Returns type of the Buffer object.
   ShapedType getType() { return type_; }
 
+  /// Updates type of the Buffer object.
```
ghpvnist, line 49: Any specific reason on the use of "///" instead of "//"?

stablehlo/reference/Ops.cpp
sdasgup3: Agreed. I will modify as suggested once https://github.com/openxla/stablehlo/pull/167 will land.

docs/status.md
sdasgup3: Done.

docs/spec_draft.md
```diff
@@ -874,6 +875,46 @@ implementation-defined value.
...
+
+### Constraints
+
+  * (C1) `operand` and `result` have the same element type.
+  * (C2) `operand` and `result` have the same number of elements.
+
+### Examples
+
+```mlir
+// %operand: [[1,2,3],[4,5,6]]]
```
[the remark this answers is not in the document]
sdasgup3: done.

docs/status.md
```diff
@@ -75,7 +75,7 @@ one of the following tracking labels.
 | DynamicUpdateSliceOp |no||||no|
 | EinsumOp |no||||no|
 | Expm1Op |no||||no|
-| ExpOp |no||||no|
+| ExpOp |yes||||no|
```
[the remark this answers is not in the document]
sdasgup3: Thanks!

stablehlo/reference/Tensor.h
sdasgup3: As per https://llvm.org/docs/CodingStandards.html#comment-formatting, I am trying to follow doxygen documentation comments for classes.

docs/interpreter.md
```diff
@@ -9,23 +9,25 @@
...
 `Tensor` class has the following APIs to interact with its individual elements:
-  - `Element Tensor::get(int64_t index)`: To extract an individual tensor
-  element at index `index` as `Element` object.
-  - `void Tensor::set(int64_t index, Element element);`: To insert an `Element`
-  object `element` into a tensor at index `index`.
+  - `Element Tensor::get(const std::vector<int64_t> &indices)`: To extract an
+     individual tensor element at multi-dimensional index `indices` as `Element`
+     object.
+  - `void Tensor::set(const std::vector<int64_t> &indices, Element element);`:
+  To insert an `Element` object `element` into a tensor at multi-dimensional
```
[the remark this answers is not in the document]
sdasgup3: Thanks! done.

docs/spec_draft.md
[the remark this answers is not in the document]
sdasgup3: removed the code as per https://github.com/openxla/stablehlo/pull/167

burmako (approved):

Issue #171, opened 2022-09-21
Add Specification and Interpreter for TransposeOp


sdasgup3:
### Request description

Add the specification and Implement the interpreter for `stablehlo.transpose` op.

PR #172, opened 2022-09-21
Specification and Interpreter for  transpose op

sdasgup3:
fixes https://github.com/openxla/stablehlo/issues/171

Diff based on https://github.com/openxla/stablehlo/pull/167


docs/status.md
```diff
@@ -135,7 +135,7 @@ one of the following tracking labels.
 | TanhOp |yes||||no|
 | TorchIndexSelectOp |no||||no|
 | TraceOp |no||||no|
-| TransposeOp |no||||no|
+| TransposeOp |yes||||yes|
```
burmako, line 138: Let's also check whether TransposeOp's verifier and shape function conform to the newly written spec. If they do, let's flip `yes*` in the table to `yes`. If they don't let's open tickets about this.

stablehlo/tests/interpret_transpose.mlir
```diff
@@ -0,0 +1,114 @@
...
+  // CHECK-NEXT:   28 : i32
+  // CHECK-NEXT:   29 : i32
+  // CHECK-NEXT:   30 : i32
+  // CHECK-NEXT:   31 : i32
+  // CHECK-NEXT:   32 : i32
+  // CHECK-NEXT:   33 : i32
+  // CHECK-NEXT:   34 : i32
+  // CHECK-NEXT:   35 : i32
+  // CHECK-NEXT:   36 : i32
+}
```
burmako, line 114: This test doesn't check anything new and looks fairly laborious for a human to follow. Perhaps delete?

docs/spec_draft.md
```diff
@@ -1029,6 +1030,52 @@ specification. Numeric precision is implementation-defined.
...
 
+## stablehlo.transpose
+
+`stablehlo.transpose(operand, permutation) -> result`
+
+### Semantics
+
+Permutes the dimensions of `operand` tensor using a permutation of its shape
+`permutation` and produces a `result` tensor. The $i^{th}$ dimension of the
+`result` is given by `permutation`$[i]^{th}$ dimension of the `operand`.
```
burmako, line 1041: This section doesn't really explain how to obtain the `result` tensor from the `operand` tensor. "The X dimension of a tensor is given by the Y dimension of another tensor" sounds fancy but is not actionable.

Here, I have the same feedback as in #170. Let's define op semantics from first principles, using only the definitions in this spec and college maths. Before proceeding with that, let's first expand our vocabulary - something which I took a stab at in #193.

burmako, line 1041: With the terminology from #193, this section could be rewritten as follows:

```
Permutes the dimensions of `operand` tensor using `permutation` and produces a 
`result` tensor. More formally, `result[i0, ..., iR-1]` is equal to 
`operand[j0, ..., jR-1]` where `i[d] = j[permutation[d]]`.
```

docs/spec_draft.md
```diff
@@ -1029,6 +1030,52 @@ specification. Numeric precision is implementation-defined.
...
+
+### Operands
+
+| Name | Type |
+|-|-|
+| `operand` | tensor of any supported types |
+| `permutation` | array of type `ui64` |
```
burmako, line 1048: For better or for worse, this is currently an array of `si64`, so let's start with reflecting the reality, and in the future we'll be able to change this to `ui64` if we decide.

docs/spec_draft.md
```diff
@@ -1029,6 +1030,52 @@ specification. Numeric precision is implementation-defined.
...
+### Results
+
+| Name | Type |
+|-|-|
+| `result` | tensor of any supported types |
+
+### Constraints
+
+  * (C1) `operand` and `result` have the same element type.
+  * (C2) `permutation` is a permutation of `operand`'s shape.
```
burmako, line 1059: "* (C2) \`permutation\` is a permutation of \`[0, 1, ..., R-1]\` where \`R\` is the rank of \`operand\`".

burmako, line 1059: "* (C3) \`result\`'s shape is a permutation of \`operand\`'s shape".

stablehlo/reference/Ops.cpp
```diff
@@ -47,5 +47,37 @@ Tensor eval(SineOp op, const Tensor &operand) {
...
+
+  // The operation itself does not require any copying but involves swapping
+  // strides.
+  std::vector<int64_t> stride(operand.getStrides());
+  std::vector<int64_t> permutation(perm.getValues<int64_t>().begin(),
+                                   perm.getValues<int64_t>().end());
+  applyInPlacePermutation(stride, permutation);
+  result.setStrides(stride);
+  return result;
+}
```
burmako, line 80: Similarly to #170,  let's define op implementations from first principles, i.e. using only the definitions in the spec and APIs from `Tensor.h` and `Element.h` which implement these definitions. 

From that perspective, relying on exact representation of a tensor, i.e. something that the spec treats as implementation-defined (currently, this is implicit, #193 makes this explicit), doesn't sound right.

I think improving on this is fairly straightforward if we use the iterator that you proposed in https://github.com/openxla/stablehlo/pull/167 (with the modifications that I proposed in the comments), for example:

```
Tensor eval(TransposeOp op, const Tensor &operand) {
  Tensor result(op.getType());
  for (auto resultIt = result.index_begin(); resultIt != result.index_end(); ++resultIt) {
    auto resultIndex = *resultIt;
    auto operandIndex = permute(resultIt, op.permutation());
    result.set(resultIndex, operandIndex);
  }
  return result;
}
```

stablehlo/reference/Ops.cpp
```diff
@@ -47,5 +47,37 @@ Tensor eval(SineOp op, const Tensor &operand) {
   return result;
 }
 
+namespace {
+
+// Appies permutation `perm` to an array `array` where perm[i] indicates the
+// location where the current array[i] goes.
+void applyInPlacePermutation(std::vector<int64_t> &array,
+                             const std::vector<int64_t> &perm) {
```
burmako, line 55: Is there a strong reason to apply the permutation in place? Functional style would be much easier to understand, e.g.:

```
llvm::SmallVector<int64_t> permute(ArrayRef<int64_t> array, DenseIntElementsAttr permutation) {
  llvm::SmallVector<int64_t> result(array.size());
  for (auto i = 0; i < result.size(); ++i) {
    result[permutation.getValues<int64_t>()[i]] = array[i];
  }
  return result;
}
```

stablehlo/reference/Ops.cpp
```diff
@@ -47,5 +47,37 @@ Tensor eval(SineOp op, const Tensor &operand) {
...
+  for (size_t i = 0; i < perm.size(); i++) {
+    swapIdx = perm[i];
+    while (swapIdx < i) {
+      swapIdx = perm[swapIdx];
+    }
+    std::swap(array[i], array[swapIdx]);
+  }
+}
+
+}  // namespace
```
burmako, line 66: Move to the top of the file?

docs/spec_draft.md
burmako, line 1048: "1-dimensional array".

burmako (changes requested):

docs/spec_draft.md
```diff
@@ -1029,6 +1030,52 @@ specification. Numeric precision is implementation-defined.
...
+
+### Examples
+
+```mlir
+// %operand: [
+//            [[1,2], [3,4], [5,6]],
```
ghpvnist, line 1065: Let's add spaces after comma to be consistent with the rest of the doc here and below.

docs/status.md
```diff
@@ -75,7 +75,7 @@ one of the following tracking labels.
 | DynamicUpdateSliceOp |no||||no|
 | EinsumOp |no||||no|
 | Expm1Op |no||||no|
-| ExpOp |no||||no|
+| ExpOp |yes||||no|
```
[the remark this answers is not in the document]
sdasgup3: Sure.

stablehlo/tests/interpret_transpose.mlir
sdasgup3: deleted.

stablehlo/reference/Ops.cpp
sdasgup3: The in-place permute is not needed. Replaced it with a simpler implementation.

docs/spec_draft.md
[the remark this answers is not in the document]
sdasgup3: not applicable with the latest revision.

docs/status.md
sdasgup3: done.

docs/spec_draft.md
```diff
@@ -1072,6 +1073,56 @@ Numeric precision is implementation-defined.
...
+
+`stablehlo.transpose(operand, permutation) -> result`
+
+### Semantics
+
+Permutes the dimensions of `operand` tensor using `permutation` and produces a
+`result` tensor. More formally, `result[i0, ..., iR-1]` is equal to
+`operand[j0, ..., jR-1]` where `i[d] = j[permutation[d]]`.
+
+
```
burmako, line 1086: Extraneous newline.

docs/spec_draft.md
```diff
@@ -1072,6 +1073,56 @@ Numeric precision is implementation-defined.
...
+| `result` | tensor of any supported types |
+
+### Constraints
+
+  * (C1) `operand` and `result` have the same element type.
+  * (C2) `permutation` is a permutation of `[0, 1, ..., R-1]` where `R` is the
+  rank of `operand`.
+  * (C3) `result`'s shape is a permutation of `operand`'s shape.
+
+
```
burmako, line 1107: Same here.

docs/status.md
```diff
@@ -135,7 +135,7 @@ one of the following tracking labels.
 | TanhOp |yes|yes*|yes*|yes|yes|
 | TorchIndexSelectOp |no|no|no|no|no|
 | TraceOp |no|yes*|no|yes|no|
-| TransposeOp |no|yes*|yes*|no|no|
+| TransposeOp |yes|yes|yes*|no|yes|
```
burmako, line 138: Is it intended for Type Inference to stay at "yes*" here? At a glance, the spec is already implemented correctly: https://github.com/openxla/stablehlo/blob/be42540a288a302b6e4071e7d29c0bf8b911589e/stablehlo/dialect/StablehloOps.cpp#L4983-L5022.

stablehlo/reference/Ops.cpp
```diff
@@ -81,5 +94,17 @@ Tensor eval(TanhOp op, const Tensor &operand) {
   return result;
 }
 
+Tensor eval(TransposeOp op, const Tensor &operand) {
+  Tensor result(op.getType());
+  SmallVector<int64_t> permutation =
+      to_vector(op.permutation().getValues<int64_t>());
```
burmako, line 100: Can we avoid `to_vector` and just pass `getValues` directly? (By appropriately changing the signature of `permute`).

burmako (changes requested):

docs/status.md
sdasgup3: I agree. Updated.

stablehlo/reference/Ops.cpp
sdasgup3: Sure. The signature looks verbose, but, being a static function, does not hurt.

burmako (approved):

PR #174, opened 2022-09-21
Add bytecode test for compare_type attr.

GleasonK:
There is currently no bytecode test for `compare_type` in CHLO. See [coverage report](https://htmlpreview.github.io/?https://raw.githubusercontent.com/GleasonK/stablehlo/ccov/reports/ccov_2022_09_21_14-51-15/stablehlo/dialect/ChloBytecode.cpp.gcov.html).

Added tests that uses all attributes of `chlo.broadcast_compare`.


burmako (approved):
This is a good incremental improvement. Do we know which parts of the existing roundtrip tests are not exhaustive yet?

GleasonK:
"Do we know which parts of the existing roundtrip tests are not exhaustive yet?"

By this do you mean what pathways are not exercised by this test file? None of the error paths by nature of being round trip, none of the type/reify conversions are run by this test (I don't think) and as for the rest, code coverage can largely determine what isn't run, since this is one of the only CHLO test files we have. I'll take a closer look and see what else from CHLO can be improved.

PR #179, opened 2022-09-22
Add specification and interpreter for SubtractOp

### LINKED ISSUE #177: Add specification and interpreter for SubtractOp

ghpvnist:
closes #177

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -169,6 +169,7 @@ described below)
    * [reshape](#stablehloreshape)
    * [rsqrt](#stablehlorsqrt)
    * [sine](#stablehlosine)
+   * [subtract](#stablehlosubtract)
    * [sqrt](#stablehlosqrt)
    * [tanh](#stablehlotanh)
    * [xor](#stablehloxor)
@@ -997,6 +998,55 @@ Numeric precision is implementation-defined.
 ```
 
 &nbsp;[More Examples](../stablehlo/tests/interpret_sine.mlir)
+
+[Back to Ops](#index-of-documented-ops)
+
+## stablehlo.subtract
+
+`stablehlo.subtract(lhs, rhs) -> result`
+
+### Semantics
+
+Performs element-wise subtraction of two tensors `lhs` and `rhs` and produces a
+`result` tensor. For integer element types, if the element-wise difference has
+an unsigned/signed overflow/underflow, the result is implementation-defined and
+one of the following:
+
+  * mathematical result modulo $2^n$, where n is the bit width of the result,
+  for unsigned overflow/underflow. For signed integer overflow/underflow, wraps
+  the result around the representable range $[-2^{n-1},\ \ 2^{n-1} - 1]$.
+  * saturation to $2^{n-1} - 1$ (or $-2^{n-1}$) for signed overflow (or signed
+      underflow) and saturation to $2^n - 1$ (or $0$) for unsigned overflow (or
+        unsigned underflow).
+
+For floating-point element types, it implements the `subtraction` operation from
+the IEEE-754 specification.
+
+### Operands
+
+| Name | Type |
+|-|-|
+| `lhs` | tensor of integer, floating-point, or complex element types |
+| `rhs` | tensor of integer, floating-point, or complex element types |
+
+### Results
+
+| Name | Type |
+|-|-|
+| `result` | tensor of integer, floating-point, or complex element types |
+
+### Constraints
+
+  * (C1) `lhs`, `rhs` and `result` have the same type.
+
+### Examples
+
+```mlir
+// %lhs: [[6, 8], [10, 12]]
+// %rhs: [[5, 6], [7, 8]]
+%result = stablehlo.subtract %lhs, %rhs : tensor<2x2xf32>
+// %result: [[1, 2], [3, 4]]
+```
 
 [Back to Ops](#index-of-documented-ops)
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -1016,8 +1016,8 @@ one of the following:
   for unsigned overflow/underflow. For signed integer overflow/underflow, wraps
   the result around the representable range $[-2^{n-1},\ \ 2^{n-1} - 1]$.
   * saturation to $2^{n-1} - 1$ (or $-2^{n-1}$) for signed overflow (or signed
-      underflow) and saturation to $2^n - 1$ (or $0$) for unsigned overflow (or
-        unsigned underflow).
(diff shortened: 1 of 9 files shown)


stablehlo/reference/Element.cpp
```diff
@@ -120,6 +120,43 @@ Element Element::operator+(const Element &other) const {
   report_fatal_error(std::move(err));
 }
 
+Element Element::operator-(const Element &other) const {
```
sdasgup3, line 123: It seems that most of the element-wise arithmetic operation will share the same code. How about extracting creating a function out of the common code. 

```
Element computeArithmaticOperation(
    const Element &e, std::function<APInt(APInt)> intOperation, std::function<APFloat(APFloat)> floatOperation);
```

stablehlo/tests/infer_subtract.mlir
```diff
@@ -0,0 +1,258 @@
+// RUN: stablehlo-interpreter --interpret -split-input-file %s | FileCheck %s
```
sdasgup3, line 1: Can't we reuse the same constant as in add? Only the expected outputs would be different.

docs/spec_draft.md
```diff
@@ -953,6 +954,55 @@ specification. Numeric precision is implementation-defined.
...
+
+  * (C1) `lhs`, `rhs` and `result` have the same type.
+
+### Examples
+
+```mlir
+// %lhs: [[6, 8], [10, 12]]
+// %rhs: [[5, 6], [7, 8]]
+%result = stablehlo.subtract %lhs, %rhs : tensor<2x2xf32>
+// %result: [[1, 2], [3, 4]]
```
sdasgup3, line 1001: Can we add the "More Examples" link as well?

sdasgup3 (commented):
Also add the updated status.md file.

docs/spec_draft.md
```diff
@@ -953,6 +954,55 @@ specification. Numeric precision is implementation-defined.
...
+`result` tensor. For integer element types, if the element-wise difference has
+an unsigned/signed overflow/underflow, the result is implementation-defined and
+one of the following:
+
+  * mathematical result modulo $2^n$, where n is the bit width of the result,
+  for unsigned overflow/underflow. For signed integer overflow/underflow, wraps
+  the result around the representable range $[-2^{n-1},\ \ 2^{n-1} - 1]$.
+  * saturation to $2^{n-1} - 1$ (or $-2^{n-1}$) for signed overflow (or signed
+      underflow) and saturation to $2^n - 1$ (or $0$) for unsigned overflow (or
+        unsigned underflow).
```
burmako, line 973: The formatting looks wonky here. Let's align to left?

stablehlo/reference/Element.cpp
```diff
@@ -120,6 +120,43 @@ Element Element::operator+(const Element &other) const {
...
+        type,
+        ArrayAttr::get(
+            type.getContext(),
+            {FloatAttr::get(complexElemTy,
+                            leftComplexValue.real() - rightComplexValue.real()),
+             FloatAttr::get(complexElemTy, leftComplexValue.imag() -
+                                               rightComplexValue.imag())}));
+  }
+
+  // Report error.
```
burmako, line 154: The comment pretty much repeats what the code below does (`report_fatal_error`), so we should drop it.

stablehlo/reference/Element.cpp
```diff
@@ -120,6 +120,43 @@ Element Element::operator+(const Element &other) const {
...
+  auto err = invalidArgument("Unsupported element type: %s",
+                             debugString(type).c_str());
```
burmako, line 156: Is there a strong reason for this to be a local variable instead of being inlined at the use site?

stablehlo/tests/infer_subtract.mlir
burmako: Hmm, how?

burmako (changes requested):

stablehlo/tests/infer_subtract.mlir
burmako, line 1: Why is the file called `infer_`?

stablehlo/reference/Element.cpp
burmako: +1. This is similar to our discussion in #162, so I proposed a pull request that generalizes this to use a single concept. Let's discuss in #190.

burmako (changes requested):

docs/status.md
```diff
@@ -131,7 +131,7 @@ one of the following tracking labels.
 | SliceOp |no||||no|
 | SortOp |no||||no|
 | SqrtOp |yes||||no|
-| SubtractOp |no||||no|
+| SubtractOp |yes||||yes|
```
burmako, line 134: Let's also check whether SubtractOp's verifier and shape function conform to the newly written spec. If they do, let's flip `yes*` in the table to `yes`. If they don't let's open tickets about this.

stablehlo/tests/infer_subtract.mlir
ghpvnist: I think he meant to say to reuse the values from lhs and rhs of add op. Will fix.

stablehlo/reference/Element.cpp
ghpvnist: Resolving comment since it's fixed in #190

stablehlo/reference/Element.cpp
ghpvnist: Resolving comment since it's fixed in #190

stablehlo/tests/infer_subtract.mlir
ghpvnist: Thanks for catching this! Fixed.

stablehlo/reference/Ops.cpp
```diff
@@ -81,5 +81,13 @@ Tensor eval(TanhOp op, const Tensor &operand) {
   return result;
 }
 
+Tensor eval(SubtractOp op, const Tensor &lhs, const Tensor &rhs) {
+  Tensor result(op.getType());
+  for (auto i = 0; i < lhs.getNumElements(); ++i) {
+    result.set(i, lhs.get(i) - rhs.get(i));
+  }
+  return result;
+}
```
burmako, line 90: Looks like this will need to be rebased on top of HEAD.

burmako (approved):

stablehlo/reference/Ops.cpp
ghpvnist: Unrelated to this PR, but it looks like cosine and tanh needs to be fixed as well. Will make PR addressing accumulated fixes after this one.

PR #187, opened 2022-09-24
Interpreter for MaxOp and MinOp

### LINKED ISSUE #185: Add Interpreter for MaxOp and MinOp
### Request description

We already have the spec published. Time to publish the interpreters.

sdasgup3:
fixes #185


stablehlo/reference/Element.cpp
```diff
@@ -120,6 +120,114 @@ Element Element::operator+(const Element &other) const {
...
+        isFloatGreaterThan(leftComplexValue.real(), rightComplexValue.real());
+    if (cmpReal) return *cmpReal ? e1 : e2;
+
+    // Real parts are equal, compare the imag parts.
+    auto cmpImag =
+        isFloatGreaterThan(leftComplexValue.imag(), rightComplexValue.imag());
+    if (cmpImag) return *cmpImag ? e1 : e2;
+
+    // Both real and imag parts are equal, return any (say the left operand).
+    return e1;
```
burmako, line 171: Since we're modelling lexicographic comparison of complex numbers after JAX, let's use [their implementation](https://github.com/google/jax/blob/7c85ca38f45e3ae0d8a80b2d33cf908de9c6564f/jax/_src/lax/lax.py#L2230-L2238). E.g. for `max` that would be:

```
auto pickLhs = lhs.real() == rhs.real() ? lhs.imag() < rhs.imag() : lhs.real() < rhs.real();
return pickLhs ? lhs : rhs.
```

burmako (changes requested):

GleasonK (commented):
Not required, but I think this could be greatly cleaned up using a `TypeSwitch` ([example](https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/StablehloBytecode.cpp#L505)).

Every unary op and binary op uses the exact same logic in the body of the branch by the looks of it. Would need to factor out Return, but otherwise could reduce a lot of redundancy.

stablehlo/reference/Element.cpp
sdasgup3: The reason we have extra code that that you have suggested is for the following reasons: 

1. The `==` operator of APFloat used `APFloat::compare` which treats `-0 == 0`. OTOH, the spec demands -0< +0.
2. The other part of the code is for propagating the NaNs. However as it is not clear how this should formulated for complex types, we can avoid that code.  At the moment I am outputting the value for which we see a NaN in its real or imag part.

sdasgup3:
I am confused by the error in  the build run. https://github.com/openxla/stablehlo/actions/runs/3132304609/jobs/5084509784


 We do have a use of the function `isSupportedIntegerType` in Elements.cpp.

GleasonK:
The only use of `isSupportedIntegerType` is in `Element map(const Element &el, IntegerFn integerFn, FloatFn floatFn, ComplexFn complexFn)` which appears to be unused. If the template is never expanded, it makes sense that the function is unused. All uses of `map` in that file take a LHS and RHS.

stablehlo/reference/Element.cpp
```diff
@@ -212,6 +223,76 @@ Element floor(const Element &e) {
...
+        // operand).
+        return lhs;
+      });
+}
+
+Element min(const Element &e1, const Element &e2) {
+  return map(
+      e1, e2,
+      [](APInt lhs, APInt rhs) { return llvm::APIntOps::smin(lhs, rhs); },
+      [](APInt lhs, APInt rhs) { return llvm::APIntOps::umin(lhs, rhs); },
```
burmako, line 265: Let's use a conditional instead of introducing another lambda parameter in `map`. This will avoid duplication in other ops which don't need to branch on signedness. The conditional could look along the lines of:

```
[&](APInt lhs, APInt rhs) {
  e1.getType().isSignedInteger() ? llvm::APIntOps::smin(lhs, rhs) : llvm::APIntOps::umin(lhs, rhs);
}
```

stablehlo/reference/Element.cpp
burmako: 1. Let's write a helper function that we'll share between `min` and `max` that does what JAX does to compute `lhs.real() == rhs.real()`. It would seem that in this case that means LLVM's `fcmp oeq` (see [elemental_ir_emitter.cc](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/elemental_ir_emitter.cc#L904-L905)), which is unfortunate because APFloat's `==` seems to be `ueq` (so we'll have to implement our own).

2. "However as it is not clear how this should formulated for complex types, we can avoid that code". Yeah, for the draft spec, let's just do whatever JAX does.

burmako (changes requested):

stablehlo/reference/Element.cpp
```diff
@@ -184,6 +184,30 @@ Element mapWithUpcastToDouble(const Element &el, FloatFn floatFn,
                                      debugString(type).c_str()));
 }
 
+bool compareEqual(APFloat A, APFloat B) {
```
burmako, line 187: Let's make this name more specific, since "compare" may mean many things for floats. How about following LLVM's nomenclature, i.e. `fcmpOeq`, `fcmpOlt` and `fcmpOgt`, for these three functions?

stablehlo/reference/Element.cpp
```diff
@@ -184,6 +184,30 @@ Element mapWithUpcastToDouble(const Element &el, FloatFn floatFn,
...
+  // APFloat::compare treats -0.0 == 0.0, hence following is an additional check
+  // to ensure -0.0 < +0.0.
+  if (A.isZero() && B.isZero() && (A.isNegative() != B.isNegative()))
+    return false;
+  return A.compare(B);
```
burmako, line 192: `A.compare(B) == cmpEqual`.

stablehlo/reference/Element.cpp
```diff
@@ -184,6 +184,30 @@ Element mapWithUpcastToDouble(const Element &el, FloatFn floatFn,
                                      debugString(type).c_str()));
 }
 
+bool compareEqual(APFloat A, APFloat B) {
+  // APFloat::compare treats -0.0 == 0.0, hence following is an additional check
+  // to ensure -0.0 < +0.0.
+  if (A.isZero() && B.isZero() && (A.isNegative() != B.isNegative()))
+    return false;
```
burmako, line 191: Have we established that this is what we need? 

From what I can see, here's how the JAX/XLA pipeline works:
  * JAX uses `lax.eq`, `lax.lt` and `lax.gt` for minmax on complex numbers ([link](https://github.com/google/jax/blob/7c85ca38f45e3ae0d8a80b2d33cf908de9c6564f/jax/_src/lax/lax.py#L2230-L2238)).
  *  These primitives lower directly to `mhlo.compare` with `compare_type = FLOAT` ([link](https://github.com/google/jax/blob/0919a6776afe83400957f2f91c62fd3337c53839/jax/_src/lax/lax.py#L2281-L2303)).
  * In XLA, `kCompare` lowers these further to `FCMP_OEQ`, `FCMP_OLT` and `FCMP_OGT` ([link](https://github.com/tensorflow/tensorflow/blob/29f709d27068b412cbeeecae6507c57462eef64f/tensorflow/compiler/xla/service/elemental_ir_emitter.cc#L896-L922)).

Now the question is how the aforementioned LLVM instructions compare `-0.0` and `+0.0`. Does this match `APFloat::compare` (in which case this special case is needed) or not (in which this special case is not needed)?

Methodologically, this is just one implementation of the potential StableHLO spec, so it's not like we must follow it. But we added min/max on complex numbers because JAX supports it, so we might as well completely follow what it does in the draft spec.

burmako (changes requested):

stablehlo/reference/Interpreter.cpp
```diff
@@ -93,6 +93,16 @@ llvm::Expected<SmallVector<Tensor>> eval(func::FuncOp func,
       Tensor runtimeOpr = fetchOperand(floorOp.operand());
       Tensor runtimeResult = eval(floorOp, runtimeOpr);
       populateResults({runtimeResult});
+    } else if (auto maxOp = dyn_cast<MaxOp>(op)) {
+      Tensor runtimeLhs = fetchOperand(maxOp.lhs());
+      Tensor runtimeRhs = fetchOperand(maxOp.rhs());
+      Tensor runtimeResult = eval(maxOp, runtimeLhs, runtimeRhs);
+      populateResults({runtimeResult});
+    } else if (auto minOp = dyn_cast<MinOp>(op)) {
+      Tensor runtimeLhs = fetchOperand(minOp.lhs());
```
GleasonK, line 102: All instances of `lhs` and `rhs` will need to be updated to `getLhs` and `getRhs` before submit now that #203 is submitted

sdasgup3: Thanks for the heads-up @GleasonK

stablehlo/reference/Element.cpp
sdasgup3: As per the HLO interpreter,
HLO::kcompare(-0.0, 0.0) == true and 
HLO:kmaximum(-0.0, 0.0) == 0.0

As Jax is using mhlo.maximum which in turn is using kMaximum, we are good with our implementation of max on floating-point type.

For the complex type, as you pointed out, Jax is using kCompare, which as per the HLO interpreter treats 0.0 == -0.0. 

Looking it from the angel of how kCompare is lowered to fcmp, from the following links we can say that fcmp has the same semantics as demonstrated by the interpreter w.r.t.  treating -0.0 == 0.0

https://godbolt.org/z/54c37fTzj : How llc treats the fcmp oeq
https://alive2.llvm.org/ce/z/dzbFCV : How ALive treats `fcmp oeq`. I guess Alive is a well trusted tool for reasoning about LLVM semantics.

Having said that we can drop the special check (to ensure +0.0 > -0.0) and simplify the code,

sdasgup3:
Also updated the spec draft to reflect out choice for max/min on complex types.

stablehlo/reference/Ops.cpp
```diff
@@ -90,6 +90,22 @@ Tensor eval(ReshapeOp op, const Tensor &operand) {
   return result;
 }
 
+Tensor eval(MaxOp op, const Tensor &lhs, const Tensor &rhs) {
+  Tensor result(op.getType());
+  for (auto i = 0; i < lhs.getNumElements(); ++i) {
```
ghpvnist, line 95: Update this to use IndexSpaceIterator here and below.

ghpvnist (changes requested):

stablehlo/reference/Ops.cpp
sdasgup3: Good catch. Thanks!

docs/spec_draft.md
```diff
@@ -638,7 +638,8 @@ function, with corner cases TBD. Numeric precision is implementation-defined.
 Performs element-wise max operation on tensors `lhs` and `rhs` and produces a
 `result` tensor. For floating-point element types, it implements the `maximum`
 operation from the IEEE-754 specification. For complex element type, it performs
-lexicographic comparison on the (real, imaginary) pairs.
+lexicographic comparison on the (real, imaginary) pairs using LLVM's `fcmp oeq`
+and `fcmp ogt` compare instructions.
```
burmako, line 642: Calling out LLVM here looks too low-level. How about we use our trusty "with corner cases TBD" formulation here and below?

burmako (changes requested):

burmako (approved):

docs/spec_draft.md
```diff
@@ -675,7 +677,7 @@ lexicographic comparison on the (real, imaginary) pairs.
 Performs element-wise max operation on tensors `lhs` and `rhs` and produces a
 `result` tensor. For floating-point element types, it implements the `minimum`
 operation from the IEEE-754 specification. For complex element type, it performs
-lexicographic comparison on the (real, imaginary) pairs.
+lexicographic comparison on the (real, imaginary) pairs  with corner cases TBD.
```
ghpvnist, line 680: Double space here.

docs/spec_draft.md.bak
```diff
@@ -0,0 +1,1223 @@
+# StableHLO Specification Draft
```
ghpvnist, line 1: Remove this file?

docs/status.md
```diff
@@ -90,8 +90,8 @@ one of the following tracking labels.
 | log_plus_one             |      no       |     yes*     |       yes*        |        yes        |     no      |
 | logistic                 |      yes      |     yes*     |       yes*        |        yes        |     no      |
 | map                      |      no       |     yes*     |        no         |        no         |     no      |
-| maximum                  |      yes      |     yes*     |       yes*        |        yes        |     no      |
-| minimum                  |      yes      |     yes*     |       yes*        |        yes        |     no      |
+| maximum                  |      yes      |     yes      |       yes         |        yes        |     yes     |
```
ghpvnist, line 93: Should both ops stay as `yes*` due to #262?

docs/spec_draft.md
```diff
@@ -666,6 +666,8 @@ lexicographic comparison on the (real, imaginary) pairs.
 // %result: [[5, 6], [7, 8]]
 ```
 
+&nbsp;[More Examples](../stablehlo/tests/interpret_max.mlir)
```
ghpvnist, line 669: Wdyt about renaming the file to `interpret_maximum.mlir` to match the mnemonic? Here and below.

stablehlo/tests/interpret_min.mlir
```diff
@@ -0,0 +1,263 @@
...
+  // CHECK-NEXT: tensor<4xcomplex<f32>> {
+  // CHECK-NEXT: [1.500000e+00 : f32, 2.500000e+00 : f32]
+  // CHECK-NEXT: [1.500000e+00 : f32, 2.500000e+00 : f32]
+  // CHECK-NEXT: [0.000000e+00 : f32, 1.500000e+00 : f32]
+  // CHECK-NEXT: [0.000000e+00 : f32, 1.500000e+00 : f32]
+}
+
+// -----
+
+// CHECK-LABEL: Evaluated results of function: min_op_test_c64
```
ghpvnist, line 252: Rename line 252, 253:
`min_op_test_c64` -> `min_op_test_c128`

stablehlo/tests/interpret_max.mlir
```diff
@@ -0,0 +1,263 @@
...
+  // CHECK-NEXT: tensor<4xcomplex<f32>> {
+  // CHECK-NEXT: [7.500000e+00 : f32, 1.500000e+00 : f32]
+  // CHECK-NEXT: [1.500000e+00 : f32, 7.500000e+00 : f32]
+  // CHECK-NEXT: [-0.000000e+00 : f32, 2.500000e+00 : f32]
+  // CHECK-NEXT: [0.000000e+00 : f32, 1.500000e+00 : f32]
+}
+
+// -----
+
+// CHECK-LABEL: Evaluated results of function: max_op_test_c64
```
ghpvnist, line 252: Rename line 252, 253:
`max_op_test_c64` -> `max_op_test_c128`

ghpvnist (changes requested):

docs/status.md
ghpvnist: Nvm, I didn't see the nearby PR. Linking for reference #264.

sdasgup3: Good catch! Yes, it should back at "yes*"

sdasgup3: Yes, I will switch it back once that PR is merged.

docs/spec_draft.md
ghpvnist, line 669: One last change, here and below:
`../stablehlo/tests/interpret_maximum.mlir`

ghpvnist (changes requested):

docs/spec_draft.md
sdasgup3: my bad! Thanks!

ghpvnist (approved):

stablehlo/reference/Element.cpp
sdasgup3: Given that APFloat::compare is unordered and the equality HLO's comparison operation used by HLO in this case is ordered, I never thought if that difference would effect the implementation we have in this PR. Here are my notes:

For an example:
```
func.func @main() -> tensor<complex<f32>> {
  %0 = mhlo.constant dense<(0x7fc0000, 2.5)> : tensor<complex<f32>>
  %1 = mhlo.constant dense<(7.5, 1.5)> : tensor<complex<f32>>
  %2 = mhlo.maximum %0, %1 : tensor<complex<f32>>
  func.return %2 : tensor<complex<f32>>
}
```

Our implementation
```
  auto cmpRes = lhs.real().compare(rhs.real()) == APFloat::cmpEqual
                          ? lhs.imag() > rhs.imag()
                          : lhs.real() > rhs.real();
return cmpRes ? lhs : rhs;
```
`lhs.real().compare(rhs.real())` would return `UNORDERED` and hence `lhs.real().compare(rhs.real()) == APFloat::cmpEqual` is false and hence the `cmpRes` is `lhs.real() > rhs.real()`. Now `ApFloat::operator>` is implemented as 
```
bool operator>(const APFloat &RHS) const {
    return compare(RHS) == cmpGreaterThan;
  }
```
which will return `false` and hence the final return is  `rhs` (`(7.5, 1.5)`).
 
Based on the fact that 
For HLO, for equality comparison they use LLVM's fcmp oeq (see [elemental_ir_emitter.cc](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/elemental_ir_emitter.cc#L904-L905)), the hlo implementation seems to be like

```
auto cmpRes = FCMP_OEQ(lhs.real(),rhs.real())
                          ? FMP_OGTl(hs.imag(), rhs.imag())
                          : FMO_OGT(lhs.real(), rhs.real());
return cmpRes ? lhs : rhs;
```
For the running example,
FCMP_OEQ(lhs.real(),rhs.real())  = `false`
FMO_OGT(lhs.real(), rhs.real()) = `fasle`

leading to the same behavior as StableHLO implementation.

PR #202, opened 2022-09-28
Fix build warning spew. Change $operands->$inputs.

GleasonK:
Currently the build spews many warnings of the form:

```
[14/42] Building StablehloOps.h.inc...
/usr/local/google/home/burmako/stablehlo/stablehlo/dialect/StablehloOps.td:981:1: note: Skipping generation of prefixed accessor `getOperands` as it overlaps with default one; generating raw form (`operands`) still
def StableHLO_OutfeedOp : StableHLO_Op<"outfeed", []> {
^
```

This is because we are switching from raw accessors `$a --> a()` to prefixed accessors `$a --> getA()`. This causes a conflict when the variable name `$operands` is used since `getOperands()` is an existing method on the `Operation` class.

Solution is to rename all instances of `$operands --> $inputs`, and all accessors `operands() --> getInputs()`.

Closes #195

PR #203, opened 2022-09-28
Update to use prefixed accessors

### LINKED ISSUE #146: Migrate MLIR instruction member accesses to prefixed form
[Raw accessors are going away](https://discourse.llvm.org/t/psa-the-default-accessor-naming-has-been-flipped-to-prefixed/65111), so we need to switch to `Prefixed` while there's still time. Recently our dialect definitions have switched from `Raw` to `Both`, and now it's time to migrate to `Prefixed`.

GleasonK:
Changes built on top of #202.

Adapted from a patch by @tpopp (thank you).

Closes #146

### DIFF
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -35,7 +35,7 @@ def StableHLO_Dialect : Dialect {
     a portability layer between ML frameworks and ML compilers.
   }];
 
-  let emitAccessorPrefix = kEmitAccessorPrefix_Both;
+  let emitAccessorPrefix = kEmitAccessorPrefix_Prefixed;
   let useDefaultAttributePrinterParser = 0;
   let useDefaultTypePrinterParser = 0;
 }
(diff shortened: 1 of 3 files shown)


stablehlo/dialect/StablehloOps.td
```diff
@@ -1980,7 +1980,7 @@ def StableHLO_CustomCallOp: StableHLO_Op<"custom_call",
   }];
 
   let arguments = (ins
-    Variadic<HLO_TensorOrTokenOrTuple>:$operands,
+    Variadic<HLO_TensorOrTokenOrTuple>,
```
burmako, line 1983: Let's change these to `inputs` as well?

burmako (changes requested):
Thanks a lot for cleaning this up! One minor comment, and we're good to go.

burmako (approved):

GleasonK:
Rebased on main. This change was built on #202, and github wasn't happy that I submitted that change first.

PR #209, opened 2022-09-28
Fix table format and other typos

ghpvnist:
This commit makes markdown tables more readable and fixes few typos and minor improvements in spec_draft.md, interpret_add.mlir, and Interpreter.cpp files.


stablehlo/reference/Interpreter.cpp
```diff
@@ -90,16 +90,16 @@ llvm::Expected<SmallVector<Tensor>> eval(func::FuncOp func,
       Tensor runtimeResult = eval(constantOp, constantOp.value());
       populateResults({runtimeResult});
     } else if (auto cosineOp = dyn_cast<CosineOp>(op)) {
-      Tensor runtimeOperand = fetchOperand(cosineOp.operand());
-      Tensor runtimeResult = eval(cosineOp, runtimeOperand);
+      Tensor runtimeOpr = fetchOperand(cosineOp.operand());
```
burmako, line 93: I haven't seen "operand" abbreviated to "opr", so how about we keep the existing name?

burmako (changes requested):
Thank you for making things more consistent! I've good one naming comment, and we're good to go.

stablehlo/reference/Interpreter.cpp
ghpvnist: Sure, then since CeilOp and FloorOp uses `runtimeOpr`, I will change them to `runtimeOperand`

burmako (approved):

PR #214, opened 2022-09-29
Add interpreter for NegOp

### LINKED ISSUE #178: Add interpreter for NegOp

ghpvnist:
closes #178


stablehlo/reference/Element.cpp
```diff
@@ -200,6 +200,12 @@ Element Element::operator+(const Element &other) const {
       });
 }
 
+Element Element::operator-() const {
+  return map(
+      *this, [&](APInt api) { return -api; }, [&](APFloat apf) { return -apf; },
```
burmako, line 205: Let's rename `api`, `apf` and `complex` to `val` similarly to how similar variables are called in `ceil` and `floor`?

stablehlo/tests/interpret_negate.mlir
```diff
@@ -0,0 +1,247 @@
+// RUN: stablehlo-interpreter --interpret -split-input-file %s | FileCheck %s
+
+// CHECK-LABEL: Evaluated results of function: negate_op_test_si4
+func.func @negate_op_test_si4() -> tensor<5xi4> {
+  %0 = stablehlo.constant dense<[-8, -1, 0, 1, 7]> : tensor<5xi4>
+  %1 = stablehlo.negate %0 : tensor<5xi4>
+  func.return %1 : tensor<5xi4>
+  // CHECK-NEXT: tensor<5xi4>
+  // CHECK-NEXT: -8 : i4
```
burmako, line 9: Sounds like the spec might need an update? Currently, it says that negation of `-8` should be either `7` or `-9`.

burmako (changes requested):

stablehlo/reference/Interpreter.cpp
```diff
@@ -97,6 +97,10 @@ llvm::Expected<SmallVector<Tensor>> eval(func::FuncOp func,
       Tensor runtimeOperand = fetchOperand(floorOp.operand());
       Tensor runtimeResult = eval(floorOp, runtimeOperand);
       populateResults({runtimeResult});
+    } else if (auto negOp = dyn_cast<NegOp>(op)) {
+      Tensor runtimeOperand = fetchOperand(negOp.operand());
```
GleasonK, line 101: Rebase this CL on main before submit, `operand()` will need to be updated to `getOperand()` with #203 submitted.

stablehlo/tests/interpret_negate.mlir
sdasgup3: acknowledged.

stablehlo/reference/Interpreter.cpp
ghpvnist: Thanks for the heads up!

burmako (approved):

GleasonK (approved):

stablehlo/reference/Ops.cpp
```diff
@@ -56,6 +56,14 @@ Tensor eval(FloorOp op, const Tensor &operand) {
   return result;
 }
 
+Tensor eval(NegOp op, const Tensor &operand) {
+  Tensor result(op.getType());
+  for (auto it = operand.index_begin(); it != operand.index_end(); ++it) {
+    result.set(*it, -operand.get(*it));
+  }
+  return result;
+}
```
joker-eph, line 65: Drive by: seems like we're creating a new tensor unconditionally, but @sdasgup3 was telling me yesterday that Tensor is mutable. 
Seems like we could check if we're the only (or better: last) uses of the operand and mutate it in-place.

PR #215, opened 2022-09-29
Enforce compatible indices and shape while indexing into Tensor

sdasgup3:
Ideally, we are supposed to use valid indices from the index space of a Tensor to access individual elements. An invalid index might lead of OOB access. While testing the interpreter we found that the eval function of cosine op is still using the linearized index to access the tensor elements. 

```
Tensor eval(CosineOp op, const Tensor &operand) {
  Tensor result(op.getType());
  for (auto i = 0; i < operand.getNumElements(); ++i) {
    result.set(i, cosine(operand.get(i)));
  }
  return result;
}
``` 

While calling the `Tensor::get(rrayRef<int64_t>)`  API, these linear indices are implicitly converted to an `ArrayRef<int64_t>` of size 1. For a tensor with >= 1 rank, these are illegal indices and might cause OOB based on the shape and choice of index. 
We realized that it is  left over  from https://github.com/openxla/stablehlo/pull/167 which we fix here (to use multi-dimensional index instead).

Next the PR add checks, in `flatenIndex` and `IndexSpaceIterator::IndexSpaceIterator`, to ensure that index is compatible with the shape of the tensor.

### DIFF
--- a/stablehlo/reference/Index.h
+++ b/stablehlo/reference/Index.h
@@ -22,9 +22,16 @@ limitations under the License.
 #include "llvm/ADT/Optional.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/Support/Error.h"
+#include "mlir/Support/LogicalResult.h"
 
 namespace mlir {
 namespace stablehlo {
+
+/// Check if the 'index' is a valid index in the index space of a tensor with
+/// shape 'shape'. Specifically, for a shape '(d0)x(d1)x...x(dR-1)' and an index
+/// '{i0, i1, ..., iR-1}', we check if 0 <= i[k] <= d[k] for k in
+/// {0, 1, ..., R-1}. Note that the check also implies that 'd[k]' >= 1.
+LogicalResult verifyIndex(ArrayRef<int64_t> shape, ArrayRef<int64_t> index);
 
 /// Iterates over the index space of a tensor with a given shape, producing
 /// indices in lexicographical order. As an example, for a tensor with shape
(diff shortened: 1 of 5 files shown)


stablehlo/reference/Tensor.cpp
```diff
@@ -51,9 +51,40 @@ int64_t getSizeInBytes(Type type) {
...
+  // 'index' should be a valid index in the index space of a tensor with shape
+  // 'shape'.
+  if (shape.size() != index.size() ||
+      any_of(llvm::zip(shape, index), [](std::tuple<int64_t, int64_t> zip) {
+        return std::get<1>(zip) < 0 || std::get<1>(zip) >= std::get<0>(zip);
+      })) {
+    llvm::report_fatal_error(
+        StringRef("Incompatible index and shape found in: ") +
+        LLVM_PRETTY_FUNCTION);
+  }
```
burmako, line 86: Let's deduplicate by adding a `LogicalResult verifyIndex(ArrayRef<int64_t> shape, ArrayRef<int64_t> index)` function in Index.h.

stablehlo/reference/Tensor.cpp
```diff
@@ -51,9 +51,40 @@ int64_t getSizeInBytes(Type type) {
...
+    llvm::report_fatal_error(
+        "Cannot index a tensor with shape having dimension sizes < 0.");
+  }
+
+  // 'index' should be a valid index in the index space of a tensor with shape
+  // 'shape'.
+  if (shape.size() != index.size() ||
+      any_of(llvm::zip(shape, index), [](std::tuple<int64_t, int64_t> zip) {
+        return std::get<1>(zip) < 0 || std::get<1>(zip) >= std::get<0>(zip);
+      })) {
```
burmako, line 82: There's quite a bit of noise here with the tuple type and std::get, if we compare this with an equivalent for loop:

```
for (auto [shapeDim, indexDim] : llvm::zip(shape, index)) {
  if (indexDim < 0 || indexDim >= shapeDim) 
    llvm::report_fatal_error(...)
}
```

burmako (changes requested):

sdasgup3:
Added some comments to verifyIndex.
Removed the check  `dimension-size < 0` as that can be inferred from the index check.

stablehlo/reference/Ops.cpp
```diff
@@ -42,8 +42,8 @@ Tensor eval(ConstantOp op, ElementsAttr value) {
 
 Tensor eval(CosineOp op, const Tensor &operand) {
   Tensor result(op.getType());
-  for (auto i = 0; i < operand.getNumElements(); ++i) {
-    result.set(i, cosine(operand.get(i)));
+  for (auto it = operand.index_begin(); it != operand.index_end(); ++it) {
+    result.set(*it, cosine(operand.get(*it)));
   }
```
burmako, line 47: Let's also change `tanh` which also uses the old style of writing eval functions?

stablehlo/reference/Index.h
```diff
@@ -22,10 +22,17 @@ limitations under the License.
...
 #include "llvm/Support/Error.h"
+#include "mlir/Support/LogicalResult.h"
 
 namespace mlir {
 namespace stablehlo {
 
+/// Check if the 'index' is a valid index in the index space of a tensor with
+/// shape 'shape'. Specifically, for a shape '(d0)x(d1)x...x(dR-1)' and an index
+/// '{i0, i1, ..., iR-1}', we check if 0 <= i[k] <= d[k] for k in
+/// {0, 1, ..., R-1}. Note that the check also implies that 'd[k]' >= 1.
```
burmako, line 33: That's a good point! Sounds like `index_begin` will now start crashing for tensors with dimensions equal to zero? Let's update `index_begin` to return `{}` in that case and update tests accordingly?

burmako (changes requested):

burmako (approved):

PR #221, opened 2022-09-29
Add type inference for stablehlo.sort

zhouxin913:
Align the sort verification logic with https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/hlo_verifier.cc and https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/shape_inference.cc. 
Remove some duplicated verification as ODS `SameOperandsAndResultShape` helps a lot.
The existing TODO of partial dynamism is removed. 

Tests: old tests is comprehensive, and add a dynamics case.


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -4860,80 +4860,62 @@ void SortOp::build(OpBuilder& builder, OperationState& state,
...
+    DictionaryAttr attributes, RegionRange regions,
+    SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
+  SortOp::Adaptor adaptor(operands, attributes, regions);
+  auto operandTypes = adaptor.operands().getTypes();
+  for (auto operandType : operandTypes) {
+    auto operandShapedType = operandType.cast<ShapedType>();
+    if (operandShapedType.hasRank()) {
+      int64_t cmpDim = adaptor.dimension();
+      int64_t rank = operandShapedType.getRank();
+      if (cmpDim < -rank || cmpDim >= rank)
```
burmako, line 4874: I am surprised that we're allowing negative axes here, because I think that everywhere else we require positive axes. I thought that this is accidental, but e.g. [the TF/XLA bridge relies on this](https://github.com/tensorflow/tensorflow/blob/f0bbe9a42cfd91cbf2551cdba7a1bab2d2cf5898/tensorflow/compiler/mlir/xla/transforms/legalize_tf.cc#L5940). Well, one more thing to clean up when working on the dynamism RFC.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -4860,80 +4860,62 @@ void SortOp::build(OpBuilder& builder, OperationState& state,
...
-    return emitOpError() << "comparator must return single output, but got: "
-                         << comparatorResult.size();
-
-  // The output of computation must be 0-ranked tensor with element-type i1.
+    return emitOptionalError(location,
+                             "comparator must return single output but got ",
+                             comparatorResult.size());
   auto comparatorResultType =
       comparatorResult[0].getType().dyn_cast<RankedTensorType>();
   if (!comparatorResultType || comparatorResultType.getRank() != 0 ||
```
burmako, line 4911: Same comment as in the nearby PR for `stablehlo.while`. If the comparator returns `tensor<*xi1>`, why is that an error?

stablehlo/tests/infer_stablehlo.mlir
```diff
@@ -330,6 +330,20 @@ func.func @case(%index : tensor<i32>, %branch_operand : tensor<2xf32>) {
...
 
+func.func @sort(%input0: tensor<16x16xf32>, %input1: tensor<16x16xi32>) {
+  // CHECK: stablehlo.sort
+  %0:2 = "stablehlo.sort"(%input0, %input1) ({
+  ^bb0(%arg0: tensor<f32>, %arg1: tensor<f32>, %arg2: tensor<i32>, %arg3: tensor<i32>):
+    %7 = "stablehlo.compare"(%arg0, %arg1) {comparison_direction = #stablehlo<comparison_direction GT>} : (tensor<f32>, tensor<f32>) -> tensor<i1>
+    "stablehlo.return"(%7) : (tensor<i1>) -> ()
+  }) {dimension = 1 : i64, is_stable = true} : (tensor<16x16xf32>, tensor<16x16xi32>) -> (tensor<16x16xf32>, tensor<16x16xi32>)
+  // CHECK: (tensor<16x16xf32>) -> tensor<16x16xindex>
+  %1 = "hlo_test_infer.get_return_type_components"(%0) : (tensor<16x16xf32>) -> tensor<16x16xindex>
```
burmako, line 341: Same comment as in the nearby PR for `stablehlo.while`. Why does `get_return_type_components` only call out the first element of the result?

burmako (changes requested):

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Yes, and in ODS of SortOp, there is `CArg<"int64_t", "-1">:$dimension`. So this is intended. Sure, include this in dynamism RFC sounds good.

stablehlo/tests/infer_stablehlo.mlir
zhouxin913: Fixed.

stablehlo/dialect/StablehloOps.td
```diff
@@ -2397,7 +2397,12 @@ def StableHLO_SortOp : StableHLO_Op<"sort", [RecursiveSideEffects,
     OpBuilder<(ins "ValueRange":$operands, CArg<"int64_t", "-1">:$dimension,
       CArg<"bool", "false">:$is_stable)>];
 
-  let hasVerifier = 1;
+  // TODO: Remove this after https://github.com/openxla/stablehlo/pull/145 landed.
+  let extraClassDeclaration = [{
+    static bool isCompatibleReturnTypes(TypeRange l, TypeRange r) {
+      return mlir::hlo::isCompatibleForHloTypeInference(l, r);
+    }
+  }];
```
burmako, line 2405: Looks like this can be removed now?

burmako (changes requested):

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Fixed in the similar way like https://github.com/openxla/stablehlo/pull/207

stablehlo/dialect/StablehloOps.td
zhouxin913: Rebased and removed.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -4937,8 +4937,8 @@ LogicalResult SortOp::inferReturnTypeComponents(
                              comparatorResult.size());
   auto comparatorResultType =
       comparatorResult[0].getType().dyn_cast<RankedTensorType>();
-  if (!comparatorResultType || comparatorResultType.getRank() != 0 ||
-      !comparatorResultType.getElementType().isInteger(1))
+  if ((comparatorResultType && comparatorResultType.getRank() != 0) ||
+      !getElementTypeOrSelf(comparatorResult[0].getType()).isInteger(1))
```
burmako, line 4941: Same comment as for `while`.

burmako (changes requested):

stablehlo/dialect/StablehloOps.cpp
zhouxin913: I'm changing this to:
```
auto comparatorResultType =
      comparatorResult[0].getType().dyn_cast<TensorType>();
if (!comparatorResultType ||
      (comparatorResultType.hasRank() && comparatorResultType.getRank() != 0) ||
      !comparatorResultType.getElementType().isInteger(1))
```
The first line of !comparatorResultType is necessary because we have a test that the return value of cond is a tuple<> which fails `dyn_cast<TensorType>()`. I think this is necessary even tho that we will deprecate tuples.

burmako: Let's check with @sdasgup3. I think tuples were supposed to be removed almost everywhere.

sdasgup3: Yes tuples are removed from op's prototypes. Also the regions within certains ops (like the SortOp here ) should not have tuples as args or return types. SortOp::verify is currently missing the [verifyReducerShape](https://github.com/openxla/stablehlo/blob/246265a89a7d5f657f6eee97839cac7a7a53364c/stablehlo/dialect/StablehloOps.cpp#L399) check which enforces the restriction.

zhouxin913: Thanks Sandeep! Yes I remove the case of tuple and update this PR.
[verifyReducerShape](https://github.com/openxla/stablehlo/blob/246265a89a7d5f657f6eee97839cac7a7a53364c/stablehlo/dialect/StablehloOps.cpp#L399) Checks a lot of things and fit for reduce-* ops. But for this SortOp, many parts of it does not fit: fore example it check the element type of return value same as input, while comparator in SortOp needs tensor<i1>. So will not use it in this PR.

sdasgup3: Just to be clear: It is perfectly fine with me not to do suggested check in the current PR and schedule it separately.

burmako (approved):
Doing veirifer update for sort in a follow-up PR sounds good to me, as long as we have a ticket for this.

PR #229, opened 2022-09-30
Infer most specific typs for if and case op.

zhouxin913:
This PR addresses https://github.com/openxla/stablehlo/pull/135#discussion_r983216479.
The unit test shows that more accurate dims are inferred according to all branches. This PR also potentially supports bounded dynamism as well.
Move a big function from `Base.h` to `Base.cpp` with NO change.

Other small change: Trim the ODS as https://github.com/openxla/stablehlo/pull/145 is landed.


stablehlo/dialect/Base.cpp
```diff
@@ -115,6 +115,90 @@ bool isCompatibleForHloTypeInference(TypeRange l, TypeRange r) {
   return true;
 }
 
+// Cases of infer return shape with bounds (lhs and rhs are commutative):
+//       Dim of lhs     Dim of rhs      Infer
+//  c0:  3              3               3
+//  c1:  3              ?               3
```
gnecula, line 121: It is not obvious to me that this should be the correct inferred type, or in fact that we should be inferring the most specific type for a conditional. I think that the classic approach in type inference is to infer the least specific type. The correctness condition that I have in mind (and I think is used in general in type inference) is that "if shape inference succeeds with type T for an expression E, then at any point at runtime the value of E matches the type T".  So, if type inference infers shape [3] then we know for sure that at runtime the shape will always be [3]. If shape inference infers shape [?] then we only know the rank but not the size.

By the above criterion, if a conditional has two branches with inferred types [3] and [?] respectively the correct inference is [?] because we cannot be sure what branch will be taken and that the result will have shape [3].

I suspect this code is written with a different correctness condition. Is this design decision documented somewhere?

stablehlo/tests/infer_stablehlo.mlir
```diff
@@ -303,15 +303,15 @@ func.func @triangular_solve(%arg0: tensor<10x5x4x4xf32>, %arg1: tensor<10x5x4x4x
 // -----
 
 // CHECK-LABEL: func @if 
-func.func @if(%pred : tensor<i1>, %branch_operand : tensor<2xf32>, %wrong_type : tensor<2xf32>) {
+func.func @if(%pred : tensor<i1>, %true_branch_eperand : tensor<5x?xf32>, %false_branch_eperand : tensor<?x5xf32>) -> tensor<*xindex> {
```
gnecula, line 306: eperand -> operand

stablehlo/tests/infer_stablehlo.mlir
```diff
@@ -303,15 +303,15 @@ func.func @triangular_solve(%arg0: tensor<10x5x4x4xf32>, %arg1: tensor<10x5x4x4x
...
     }, {
-      "stablehlo.return"(%branch_operand) : (tensor<2xf32>) -> ()
-    }) : (tensor<i1>) -> tensor<2xf32>
-  // CHECK: (tensor<2xf32>) -> tensor<2xindex>
-  %1 = "hlo_test_infer.get_return_type_components"(%0) : (tensor<2xf32>) -> tensor<2xindex>
-  func.return
+      "stablehlo.return"(%false_branch_eperand) : (tensor<?x5xf32>) -> ()
+    }) : (tensor<i1>) -> tensor<?x?xf32>
+  // CHECK: types0 = tensor<5x5xf32>
+  %1 = "hlo_test_infer.get_return_types"(%0) : (tensor<?x?xf32>) -> tensor<*xindex>
```
gnecula, line 313: What does *xindex mean?

stablehlo/dialect/Base.cpp
joker-eph: It depends on the type system: is this a case of early vs late binding? (I'm not sure I have the right terminology here)

zhouxin913: Hi George, "this design decision documented somewhere" Yes there is. I add you with @ in the doc so you should have got the link.
"because we cannot be sure what branch will be taken" this is why I suppose the "[3]" as inferred type, let's consider this situation: 
1. if_op: branch_0 returns [?], branch_1 returns [3], the op returns [?]
2. the_next_op: the input is [4]

If infer the least specific type [?], we will not catch any error in the compile time. Because the inferred type [?] is compatible with any 1D tensor, even with [4]. Instead, with inferring [3], it will compatible with [3] or [?] in the_next_op. 

Could you provide more concrete examples or concerns you may have in JAX? In my current understanding the goal is to propagate static dims among ops as accurate as possible.

joker-eph: It all depends if the program is valid when the side of the branch that is "incorrect" is never taken: can you have "impossible" shapes in unreachable code? (this is what I was trying to refer to with "early" vs "late" binding earlier). 
For example in python the situation you're describing is perfectly fine as long as the branch that returns [3] is never taken and the branch that return [?] statically actually always return [4].

On the other hand, if you statically enforce that all possible paths have to be possible, then some slight change can trigger shape inference to turn a program that actually "works" into a compile-time failure. Users will have to resolve into explicit "shape erasure" to trip the inference and please the compiler.

There are trade-offs :)

gnecula: Thank you for pointing the design document to me. I note though that the discussion in the design doc refers to binary ops, which is quite different from conditionals. For a binary op that requires both operands to have the same shape, it is correct to infer the result to have shape [3] when the operands have shape [3] and [?]. This is because the operation will either fail with a shape error or the result will have shape [3]. 

For a conditional with two branches [3] and [?], you should infer [?] because on some executions the shape may be something other than [3]. I understand that this means you may not catch a type error further down, but this type system is always going to be missing some type errors. 

If you infer [3] then you loose all guarantees about your type system, because you cannot be sure that an expression with inferred shape [3] will really have that shape at runtime.

The intuition that I am using for this is based on a classic type inference with subtypes. The type [3] is a subtype of [?] because the set of values with shape [3] are included in the values with shape [?]. And for conditionals you need to pick the common super type of the branches in order to ensure soundness.

burmako: "The intuition that I am using for this is based on a classic type inference with subtypes". Yeah, the tricky part here is that, to the best of my knowledge, we don't have a formal calculus for HLO, so we have to partly rely on intuition and partly figure things out on our own. Furthermore, I don't think that the algorithm used in `--tf-shape-inference` has been formally defined either, so combining the two is an exciting exercise!

The situation with `stablehlo.if` in the current type system that StableHLO inherited from MHLO is interesting in two major ways:
  1. StableHLO ifs are different from conditionals in many other languages in that all branches must return exactly the same runtime type, extrapolating from statically-typed HLO conditionals. So while in e.g. Java we can say `cond ? new A() : new B()`, in the StableHLO example from above, the tensors must both be `[3]` at runtime.
  2. In StableHLO, runtime type mismatches are undefined behavior. This means that when statically reasoning about dynamically-typed StableHLO programs, we can assume that all the question marks line up (because if they don't, that's UB, and we can do whatever we want with UB, including treating it as not our problem).

Taking 1) and 2) together, I think we can arrive at a conclusion that for `stablehlo.if` the least upper bound (what you proposed) and the greatest lower bound (what Xin proposed) are guaranteed to be the same thing.

"And for conditionals you need to pick the common super type of the branches in order to ensure soundness". Is what I wrote above sound? And what does soundness mean for StableHLO programs? (E.g. does preservation hold in such a type system, and does that matter? Or what does progress mean when we can wish away stuck programs by declaring them UB?). I don't know the answers to these questions yet, but I plan to know much more about that by EOQ4.

Specifically for this pull request, I propose that we don't merge it just yet and wait until some of these questions are worked out in more detail.

burmako (changes requested):

stablehlo/dialect/Base.cpp
gnecula: I am having a hard time thinking of a way to rationalize all this, and what it means for type inference. I am too used to think about type inference and abstract interpretation with a soundness criterion in mind without which I do not know what is correct and what is not. 

One way to understand your point (2) is that we are only interested in the executions where the caller passes input arguments that ensure that throughout the execution values match the static input shapes **and** will not result in runtime type errors. Therefore, shape inference should abort only when there is no such execution possible, e.g., for "add <3xf32> <5xf32>". My comments earlier assumed that all executions must match the static shapes, not that there exists one execution that matches the shape. 

Do I have the right intuition about "exists an execution"? I am worried that a static analysis that tries to verify that is undecidable. 

I do not quite understand your point (1): "all branches must return exactly the same runtime type". At runtime only one branch is taken so the notion of "same runtime type" is not defined.

I'll stop here hoping that we can clarify if I have the right model.

atondwal: > I am having a hard time thinking of a way to rationalize all this, and what it means for type inference.

Yes, this all quite strange. We inherit the strange control flow semantics from TF, where cond branches must "statically" return the same type.  I say "statically" here even though this is unchecked because it's assumed in the compilation pipeline from TF that given an input shape, all executions for inputs with that shape must result in the identical shapes at meet points (IIUC).

On the other hand, if you're looking for invariants in this type system (as it exists), you're not going to find many. (Hopefully we can improve that situation with time). MLIR's `tensor` type and HLO `cond`/`while` are based on how TF handles tensors and control flow. And AFAIK TF has no static semantics beyond "it runs the graph". This was not a problem for HLO/XLA in the past because we didn't have to compile every program: just a subset of programs with statically-determinable shapes we could imbue with semantics and compile. This wasn't the approach that was taken with MHLO in including support of dynamic shapes --- my understanding is that this was born out of the desire to support a wider (the full range?) of TF graphs.

gnecula: @atondwal Thank you for your clarification, meanwhile @burmako has explained offline the TF genealogy of this logic. Eugene also said that while MHLO will have to stick with this inherited semantics, we can try to do better with StableHLO (IIUC).

burmako: To expand on this: at the moment, the verifier of `mhlo::IfOp` [requires](https://github.com/tensorflow/mlir-hlo/blob/34028950c5fac34a074bb9da63659acea210bbcb/lib/Dialect/mhlo/IR/hlo_ops.cc#L6556-L6584) that the type of each branch exactly matches the result type of the op. This is the invariant that is currently assumed by MHLO consumers, e.g. XLA Next and IREE.

Dynamism support in MHLO has evolved organically and I wasn't around at the very beginning, so I don't actually know whether this invariant is intentional or accidental. As a result, my next step will be to reach out to folks who have more context, e.g. @sherhut and @stellaraccident, to find out for sure.

I don't think that we absolutely have to stick with this inherited semantics. It's just that once we come up with details of what semantics we want instead, we'll want to build the consensus across the community, and that would involve figuring out the story for StableHLO consumers (most of which are currently using MHLO). After the consensus is reached, we'll have a path forward for MHLO as well.

zhouxin913: Thanks for discussion of everyone! 
These notes are super useful for moving this on in the future. As discussed above: we have chance to define a proper behavior for this in StableHLO, but it makes more sense after we collect enough info from Stakeholders and understand a wider range of scenarios. 
Let me close this PR for now, and reopen when there is a clear direction shows up.

zhouxin913:
Thanks for discussion of everyone!
These notes are super useful for moving this on in the future. As discussed above: we have chance to define a proper behavior for this in StableHLO, but it makes more sense after we collect enough info from Stakeholders and understand a wider range of scenarios.
Let me close this PR for now, and reopen when there is a clear direction shows up.

PR #231, opened 2022-09-30
Assorted cleanup

burmako:
* Inlines various errors directly into report_fatal_error (there doesn't seem to be a use case for having a temporary variable which is then immediately moved).
* Uses SmallVector in permute (Sandeep, you were right - the actual type is such a mouthful!).

### DIFF
--- a/stablehlo/reference/Ops.cpp
+++ b/stablehlo/reference/Ops.cpp
@@ -24,8 +24,7 @@ namespace {
 
 // Appies the permutation `perm` to an array `array` where perm[i] indicates the
 // location where the current array[i] goes.
-std::vector<int64_t> permute(ArrayRef<int64_t> array,
-                             SmallVector<int64_t> perm) {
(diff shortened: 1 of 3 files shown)


stablehlo/reference/Ops.cpp
```diff
@@ -24,14 +24,10 @@ namespace {
 
 // Appies the permutation `perm` to an array `array` where perm[i] indicates the
 // location where the current array[i] goes.
-std::vector<int64_t> permute(
-    ArrayRef<int64_t> array,
-    mlir::detail::ElementsAttrRange<
-        mlir::DenseElementsAttr::ElementIterator<int64_t>>
-        perm) {
+std::vector<int64_t> permute(ArrayRef<int64_t> array,
+                             SmallVector<int64_t> perm) {
```
sdasgup3, line 28: ArrayRef<int64_t> perm?

burmako: Done.

sdasgup3 (approved):

PR #232, opened 2022-09-30
Refactor error handling code

ghpvnist:
Error handling is needed in several files, and currently, we are repeatedly defining the wrapper error handling function in each file, which duplicates code:
* Element.cpp
* Interpreter.cpp
* Tensor.cpp

This PR refactors the error handling code in `Errors.h`


sdasgup3:
It is better to update the PR description with the need of this change. Maybe like "Error handling is need in almost all the files and currently we are not reusing the error handling code"

burmako (changes requested):
It's a great idea to deduplicate error handling! I've left a few requested changes.

sdasgup3:
@burmako @ghpvnist 

At this point we are using a mix of the following variants of `report_fatal_error`
(1) https://github.com/llvm/llvm-project/blob/fdadb5c146960c526f247c829a360f3dae825d8a/llvm/include/llvm/Support/Error.h#L730
(2) https://github.com/llvm/llvm-project/blob/fdadb5c146960c526f247c829a360f3dae825d8a/llvm/include/llvm/Support/ErrorHandling.h#L71

I do not think the interpreter is handling the errors any differently in these two cases. We just report and abort. 
Why not to use (2) consistently, then we may not need any of the above boilerplate.

What do guys think?

ghpvnist:
> @burmako @ghpvnist
> 
> At this point we are using a mix of the following variants of `report_fatal_error` (1) https://github.com/llvm/llvm-project/blob/fdadb5c146960c526f247c829a360f3dae825d8a/llvm/include/llvm/Support/Error.h#L730 (2) https://github.com/llvm/llvm-project/blob/fdadb5c146960c526f247c829a360f3dae825d8a/llvm/include/llvm/Support/ErrorHandling.h#L71
> 
> I do not think the interpreter is handling the errors any differently in these two cases. We just report and abort. Why not to use (2) consistently, then we may not need any of the above boilerplate.
> 
> What do guys think?

I agree that we should also unify the use of error handling as well. Do you have preference over (2)? @burmako

burmako:
How would (2) look like exactly? What I like about (1) is that it hides some of the details of string formatting, e.g. I like being able to say `("Expected one region in func %s", func.getName().str().c_str())` instead of stuff like `StringRef("Unsupported type: ") + debugString(type)`.

By the way, it would be even better if we could say `("Expected one region in func %s", func.getName())`, but that's probably a different story.

sdasgup3:
>> How would (2) look like exactly? What I like about (1) is that it hides some of the details of string formatting, e.g. I like being able to say ("Expected one region in func %s", func.getName().str().c_str()) instead of stuff like StringRef("Unsupported type: ") + debugString(type).

Yes I was suggesting the concatenating StringRefs using + operators using (2). 

makes sense with your point.

stablehlo/reference/Errors.h
```diff
@@ -0,0 +1,36 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#ifndef STABLHLO_REFERENCE_ERRORS_H
+#define STABLHLO_REFERENCE_ERRORS_H
```
burmako, line 17: Typo: "STABLEHLO".

burmako (approved):

stablehlo/reference/Errors.h
ghpvnist: Thanks for the catch!

PR #234, opened 2022-10-01
Add specification and interpreter for IotaOp

### LINKED ISSUE #225: Add specification and interpreter for IotaOp

ghpvnist:
closes #225


docs/spec_draft.md
```diff
@@ -550,6 +551,64 @@ IEEE-754 specification.
 
 [Back to Ops](#index-of-documented-ops)
 
+## stablehlo.iota
+
+### Semantics
+Fills a `result` array with values in increasing order starting from zero along
+the `iota_dimension` dimension.
+
+Note: For integers, if the dimension size is larger than what the element type's
```
sdasgup3, line 560: Can we have an signed/unsigned underflow with values increasing starting from zero?

We can drop the `Note:` as we are not using anything like that for such special cases for other ops.

docs/spec_draft.md
```diff
@@ -550,6 +551,64 @@ IEEE-754 specification.
...
+maximum value can hold, an overflow occurs and the behavior is implementation-
+defined and one of the following:
+  * mathematical result modulo $2^n$, where n is the bit width of the result,
```
sdasgup3, line 563: I believe this is what the HLO interpreter is producing. We should do the experiment if not already done. Else ignore.

docs/spec_draft.md
```diff
@@ -550,6 +551,64 @@ IEEE-754 specification.
...
+### Results
+
+| Name     | Type                                                       |
+|----------|------------------------------------------------------------|
+| `result` | tensor of integer, floating-point or complex element types |
+
+### Constraints
+
+  * (C1) `iota_dimension` >= 0
+  * (C2) `result` rank >= 1
```
sdasgup3, line 585: Instead of C1, C2 and C3, we can just say
0 <= `iota_dimension` < R, where R is the rank of the `result`

The fact that R >= 1 is implied from the above.

stablehlo/reference/Ops.cpp
```diff
@@ -68,6 +81,50 @@ Tensor eval(FloorOp op, const Tensor &operand) {
   return result;
 }
 
+Tensor eval(IotaOp op) {
+  Tensor result(op.getType());
+  ArrayRef<int64_t> shape = result.getType().getShape();
+  Type type = result.getType().getElementType();
+  if (!isSupportedIntegerType(type) && !isSupportedFloatType(type) &&
```
sdasgup3, line 88: This check can be removed. Instead, for every type cases enumerated below, return the `result` just within the `if` conditional and at the end of the function report_fatal_error.

stablehlo/tests/interpret_iota.mlir
```diff
@@ -0,0 +1,1213 @@
...
+  // CHECK-NEXT: 1 : i4
+  // CHECK-NEXT: 2 : i4
+  // CHECK-NEXT: 3 : i4
+  // CHECK-NEXT: 0 : i4
+  // CHECK-NEXT: 1 : i4
+  // CHECK-NEXT: 2 : i4
+  // CHECK-NEXT: 3 : i4
+}
+
+// CHECK-LABEL: Evaluated results of function: iota_op_test_si4_dim_0_overflow
```
sdasgup3, line 41: We should drop tests exercising the implementation defined behavior. This is what we followed for other ops like AddOp.

stablehlo/tests/interpret_iota.mlir
```diff
@@ -0,0 +1,1213 @@
...
+  // CHECK-NEXT: 2 : i64
+  // CHECK-NEXT: 3 : i64
+  // CHECK-NEXT: 0 : i64
+  // CHECK-NEXT: 1 : i64
+  // CHECK-NEXT: 2 : i64
+  // CHECK-NEXT: 3 : i64
+}
+
+// CHECK-LABEL: Evaluated results of function: iota_op_test_ui64_dim_0
+func.func @iota_op_test_ui64_dim_0() -> tensor<3x4x5x3xui64> {
```
sdasgup3, line 240: How about keeping the dimension sizes <= 3 and rank == 3; an example shape could be 2x3x2, which is manageable w.r.t the element prints and still can demonstrates a relative complex case. The output here is too lengthy to follow.

sdasgup3: same for other `3x4x5x3` cases below.

stablehlo/tests/interpret_iota.mlir
```diff
@@ -0,0 +1,1213 @@
...
+  // CHECK-NEXT: -8 : i4
+  // CHECK-NEXT: -7 : i4
+  // CHECK-NEXT: -6 : i4
+  // CHECK-NEXT: -5 : i4
+  // CHECK-NEXT: -4 : i4
+  // CHECK-NEXT: -3 : i4
+  // CHECK-NEXT: -2 : i4
+  // CHECK-NEXT: -1 : i4
+}
+
```
sdasgup3, line 86: You may add the split markers. Just to follow what other tests are doing. Not required though.

stablehlo/reference/Ops.cpp
```diff
@@ -68,6 +81,50 @@ Tensor eval(FloorOp op, const Tensor &operand) {
   return result;
 }
 
+Tensor eval(IotaOp op) {
```
sdasgup3, line 84: In my opinion, this op's implementation can be  much simplified, without having any extra computation to determine the return value at a particular index, by  following the observation that 

```
The value of a result  Tensor T  with index I = [i0, i1, .., iR-1] is given by
R[I] = I[iota_dimension];
```

The following code to handle the integer type. The other types will follow the similar logic.  
```
Tensor eval(IotaOp op) {
  Tensor result(op.getType());

  Type elementType = result.getType().getElementType();
  uint64_t iotaDimension = op.getIotaDimension();

  for (auto it = result.index_begin(); it != result.index_end(); ++it) {
    result.set(*it,
               Element(elementType, APInt(elementType.getIntOrFloatBitWidth(),
                                          (*it)[iotaDimension], false)));
  }

  return result;
} 
```

sdasgup3 (changes requested):

docs/spec_draft.md
```diff
@@ -550,6 +551,64 @@ IEEE-754 specification.
...
+  the result around the representable range $[-2^{n-1},\ 2^{n-1} - 1]$.
+  * saturation to $2^{n-1} - 1$ (or $-2^{n-1}$) for signed overflow (or signed
+  underflow) and saturation to $2^n - 1$ (or $0$) for unsigned overflow (or
+  unsigned underflow).
+
+### Operands
+
+| Name             | Type   |
+|------------------|--------|
+| `iota_dimension` | `ui64` |
```
burmako, line 574: Should this be `si64` similarly to how we define dimension sizes in https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md#types?

docs/spec_draft.md
```diff
@@ -550,6 +551,64 @@ IEEE-754 specification.
 
 [Back to Ops](#index-of-documented-ops)
 
+## stablehlo.iota
+
+### Semantics
+Fills a `result` array with values in increasing order starting from zero along
```
burmako, line 557: Let's say "tensor" instead of "array"?

docs/spec_draft.md
```diff
@@ -550,6 +551,64 @@ IEEE-754 specification.
...
+the `iota_dimension` dimension.
```
burmako, line 558: As discussed in #248, let's define op semantics from first principles, using only the definitions in this spec and college maths. This sentence is fine for building intuition for what's happening, but we'll also need to describe what  it means specifically, e.g. similarly to what https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md#stablehlotranspose does.

docs/status.md
```diff
@@ -84,7 +84,7 @@ one of the following tracking labels.
 | IfOp                    |         no         |        yes*        |        yes*         |          no          |        no        |
 | ImagOp                  |         no         |        yes*        |        yes*         |         yes          |        no        |
 | InfeedOp                |         no         |        yes*        |         no          |          no          |        no        |
-| IotaOp                  |         no         |        yes*        |     infeasible      |         yes          |        no        |
+| IotaOp                  |        yes         |        yes*        |     infeasible      |         yes          |       yes        |
```
burmako, line 87: What about verification logic in the StableHLO dialect? Does it conform to the newly added spec?

burmako (changes requested):

docs/spec_draft.md
```diff
@@ -550,6 +551,64 @@ IEEE-754 specification.
...
+### Constraints
+
+  * (C1) `iota_dimension` >= 0
+  * (C2) `result` rank >= 1
+  * (C3) `result` rank > `iota_dimension`
+
+### Examples
+
+```mlir
+%result = "stablehlo.iota"() { iota_dimension = 0 } : (tensor<i64>) -> tensor<4x5xi32>
```
burmako, line 591: `%result = "stablehlo.iota"() { iota_dimension = 0 } : () -> tensor<4x5xi32>` is the correct syntax. In the future PRs, we should run `stablehlo-opt` on all code examples - I'll add that to the checklist.

burmako (changes requested):

docs/spec_draft.md
burmako: More specifically, here's how this could look like using Sandeep's suggestion from above and bringing it more in line with `concatenate`, `reshape` and `transpose`:

```
Fills a `result` tensor with values in increasing order starting from zero along
the `iota_dimension` dimension. More formally, 
`result[i0, ..., id, ..., iR-1] = id`, where `d` is equal to `iota_dimension`.
```

docs/spec_draft.md
ghpvnist: Since the numbers are incrementing, it is not possible to have an underflow. Will drop the `Note:` part.

stablehlo/reference/Ops.cpp
ghpvnist: Will keep this per offline discussion to terminate early upon encountering error. This also removes the need for reporting error in the loop.

docs/spec_draft.md
ghpvnist: Yes, the HLO interpreter has this behavior, and since this is implementation specific, I will leave the two options.

docs/status.md
ghpvnist: Yes, I've checked that the verifier conforms to the spec and the ODS.

stablehlo/reference/Ops.cpp
```diff
@@ -68,6 +81,44 @@ Tensor eval(FloorOp op, const Tensor &operand) {
...
+      !isSupportedComplexType(elType))
+    report_fatal_error(invalidArgument("Unsupported element type: %s",
+                                       debugString(elType).c_str()));
+  uint64_t iotaDimension = op.getIotaDimension();
+  llvm::RoundingMode rm = APFloat::rmNearestTiesToEven;
+  bool roundingErr;
+  for (auto it = result.index_begin(); it != result.index_end(); ++it) {
+    auto iota = (*it)[iotaDimension];
+    if (isSupportedIntegerType(elType)) {
+      result.set(*it, Element(elType, APInt(elType.getIntOrFloatBitWidth(),
```
sdasgup3, line 97: Is it fine to consider all int types as unsigned ints here? 
I agree that these values are handled property, in `Tensor::get and tensor::set`, with or w/o sign extension based on the type being signed or unsigned, but I prefer generating the `APints` here with proper signed. First it would be agnostic of how we handle these values later and brings the implementation sync with the spec. 

Wdyt?

ghpvnist: I'll create separate cases for signed and unsigned integers as that would make it more aligned with the spec's notion of implementation-defined for signed and unsigned integer element types.

stablehlo/reference/Ops.cpp
```diff
@@ -68,6 +81,49 @@ Tensor eval(FloorOp op, const Tensor &operand) {
   return result;
 }
 
+Tensor eval(IotaOp op) {
+  Tensor result(op.getType());
+  Type elType = result.getType().getElementType();
+  if (!isSupportedIntegerType(elType) && !isSupportedFloatType(elType) &&
+      !isSupportedComplexType(elType))
+    report_fatal_error(invalidArgument("Unsupported element type: %s",
+                                       debugString(elType).c_str()));
```
burmako, line 90: It looks like this check duplicates the checks that are already done in the loop. E.g. if this was happening inside the loop, the only thing needed would be `else report_fatal_error(...)`. Any specific reason to prefer the current arrangement?

stablehlo/reference/Ops.cpp
```diff
@@ -68,6 +81,49 @@ Tensor eval(FloorOp op, const Tensor &operand) {
...
+      } else if (isSupportedUnsignedIntegerType(elType)) {
+        result.set(*it, Element(elType, APInt(elType.getIntOrFloatBitWidth(),
+                                              iota, false)));
+      }
+    } else if (isSupportedFloatType(elType)) {
+      APFloat val = APFloat((double)iota);
+      if (elType.isBF16()) val.convert(APFloat::BFloat(), rm, &roundingErr);
+      if (elType.isF16()) val.convert(APFloat::IEEEhalf(), rm, &roundingErr);
+      if (elType.isF32()) val.convert(APFloat::IEEEsingle(), rm, &roundingErr);
+      if (elType.isF64()) val.convert(APFloat::IEEEdouble(), rm, &roundingErr);
```
burmako, line 109: `val.convert(elType.getFloatSemantics(), rm, &roundingErr)`

stablehlo/reference/Ops.cpp
```diff
@@ -68,6 +81,49 @@ Tensor eval(FloorOp op, const Tensor &operand) {
...
+Tensor eval(IotaOp op) {
+  Tensor result(op.getType());
+  Type elType = result.getType().getElementType();
+  if (!isSupportedIntegerType(elType) && !isSupportedFloatType(elType) &&
+      !isSupportedComplexType(elType))
+    report_fatal_error(invalidArgument("Unsupported element type: %s",
+                                       debugString(elType).c_str()));
+  uint64_t iotaDimension = op.getIotaDimension();
+  llvm::RoundingMode rm = APFloat::rmNearestTiesToEven;
+  bool roundingErr;
```
burmako, line 93: I think these variables should be declared next to their callsites, so that they are placed in the narrowest scope possible. Also, how much mileage are we getting from declaring a variable that carries `APFloat::rmNearestTiesToEven`? It's used only three times below, so perhaps we should just inline it?

stablehlo/reference/Ops.cpp
```diff
@@ -15,10 +15,23 @@ limitations under the License.
 
 #include "stablehlo/reference/Ops.h"
 
+#include "Tensor.h"
+#include "llvm/ADT/APFloat.h"
```
burmako, line 19: `#include "llvm/ADT/APInt.h"` as well?

burmako (changes requested):

stablehlo/reference/Ops.cpp
ghpvnist: Thanks! I was looking for ways to simplify this, and `FloatType` had the function you suggested `getFloatSemantics()`.

stablehlo/reference/Ops.cpp
ghpvnist: It was declared early to remove some clutter from the loops, but with your suggestions above, it is much more simplified, so I'll inline it.

stablehlo/reference/Ops.cpp
ghpvnist: It's a combination of early termination preference and an attempt to remove some clutter in the loop. With your suggestion, below, I'll move this in the loop.

stablehlo/reference/Ops.cpp
```diff
@@ -68,6 +81,43 @@ Tensor eval(FloorOp op, const Tensor &operand) {
...
+    if (isSupportedSignedIntegerType(elType)) {
+      result.set(*it, Element(elType, APInt(elType.getIntOrFloatBitWidth(),
+                                            iota, /*isSigned=*/true)));
+    } else if (isSupportedUnsignedIntegerType(elType)) {
+      result.set(*it, Element(elType, APInt(elType.getIntOrFloatBitWidth(),
+                                            iota, /*isSigned=*/false)));
+    } else if (isSupportedFloatType(elType)) {
+      APFloat val = APFloat((double)iota);
+      bool roundingErr;
+      val.convert(elType.dyn_cast<FloatType>().getFloatSemantics(),
```
burmako, line 99: `.cast<FloatType>` to express the fact that we know that this cast will succeed.

stablehlo/reference/Ops.cpp
```diff
@@ -68,6 +81,43 @@ Tensor eval(FloorOp op, const Tensor &operand) {
...
+                  APFloat::rmNearestTiesToEven, &roundingErr);
+      result.set(*it, Element(elType, val));
+    } else if (isSupportedComplexType(elType)) {
+      APFloat real((double)iota);
+      APFloat imag((double)0.0);
+      FloatType flType =
+          elType.dyn_cast<ComplexType>().getElementType().dyn_cast<FloatType>();
```
burmako, line 106: Same here with `.cast<ComplexType>()` and `.cast<FloatType>()`.

burmako (changes requested):

docs/status.md
```diff
@@ -84,7 +84,7 @@ one of the following tracking labels.
 | if                       |      no       |     yes*     |       yes*        |        no         |     no      |
 | imag                     |      no       |     yes*     |       yes*        |        yes        |     no      |
 | infeed                   |      no       |     yes*     |        no         |        no         |     no      |
-| iota                     |      no       |     yes*     |    infeasible     |        yes        |     no      |
+| iota                     |      yes      |     yes      |    infeasible     |        yes        |     yes     |
```
burmako, line 87: Since the new format for ODS description and the checklist were developed after this PR was sent out for review, I forgot one more thing: can you update the ODS description as well?

burmako (changes requested):

burmako (approved):

sdasgup3 (approved):

PR #248, opened 2022-10-03
Add spec for ConcatenateOp

### LINKED ISSUE #242: Add spec for ConcatenateOp

ghpvnist:
Closes #242

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -363,8 +363,9 @@ IEEE-754 specification.
 
 ### Semantics
 
-Concatenates sequence of tensors in `inputs` along `dimension` dimension in the
-same order as the given arguments and produces a `result` tensor. More formally,
+Concatenates a variadic number of tensors in `inputs` along `dimension`
+dimension in the same order as the given arguments and produces a `result`
+tensor. More formally,
 `result[i0, ..., id, ..., iR-1] = inputs[k][i0, ..., kd, ..., iR-1]`, where:
   1. `id = d0 + ... + dk-1 + kd`.
   2. `d` is equal to `dimension`, and `d0`, ... are `d`'th dimension sizes
@@ -389,12 +390,10 @@ same order as the given arguments and produces a `result` tensor. More formally,
   * (C2) All tensors in `inputs` have the same shape except for the size of the
   `dimension` dimension.
   * (C3) `inputs` have N tensors where N >= 1.
-  * (C4) `inputs` have tensors of rank > 0.
-  * (C5) rank of tensors in `inputs` and `result` are the same.
-  * (C6) `dimension` is non-negative and less than the rank of the tensors in
+  * (C4) `dimension` is non-negative and less than the rank of the tensors in
   `inputs`.
-  * (C7) `result` has the same element type as the tensors in `inputs`.
-  * (C8) `result` has the same shape as the tensors in `inputs` except for the
+  * (C5) `result` has the same element type as the tensors in `inputs`.
+  * (C6) `result` has the same shape as the tensors in `inputs` except for the
   size of the `dimension` dimension, which is calculated as a sum of the sizes
   of this dimension in all tensors in `inputs`.
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -153,6 +153,7 @@ described below)
    * [add](#stablehloadd)
    * [and](#stablehloand)
    * [ceil](#stablehloceil)
+   * [concatenate](#stablehloconcatenate)
    * [constant](#stablehloconstant)
    * [cosine](#stablehlocosine)
    * [divide](#stablehlodivide)
@@ -355,6 +356,56 @@ IEEE-754 specification.
 ```
 
 &nbsp;[More Examples](../stablehlo/tests/interpret_ceil.mlir)
+
+[Back to Ops](#index-of-ops)
+
+## stablehlo.concatenate
+
+### Semantics
+
+Concatenates sequence of tensors in `val` along dimension `dimension` and
+produces a tensor as a result. The tensors are of the same rank as the input
+operands except the specified dimension.
+
+### Operands
+
+| Name        | Type                                                                     |
+|-------------|--------------------------------------------------------------------------|
+| `dimension` | `i64`                                                                    |
+| `val`       | Sequence of tensors of integer, floating-point, or complex element types |
+
+### Results
+
+| Name     | Type                                                        |
+|----------|-------------------------------------------------------------|
+| `result` | tensor of integer, floating-point, or complex element types |
+
+### Constraints
+
+  * (C1) `val` and `result` have the same type.
+  * (C2) All tensors in `val` have the same element type and rank.
+  * (C3) Tensors of `val` all have the same dimension except for `dimension`.
+
+### Examples
+
+```mlir
+// 1-dimensional concatenate
+
+// %val1: [1, 2]
+// %val2: [3, 4]
+// %val3: [5, 6]
+// %dimension: 0
+%result = "stablehlo.concatenate"(%val1, %val2, %val3, %dimension) : (tensor<2xi32>, tensor<2xi32>, tensor<2xi32>, i64) -> tensor<8xi32>
+// %result: [1, 2, 3, 4, 5, 6]
+
+// 2-dimensional concatenate
+
+// %val1: [[1, 2], [3, 4], [5, 6]]
+// %val2: [[7, 8]]
+// %dimension: 0
+%result = "stablehlo.concatenate"(%val1, %val2, %dimension) : (tensor<3x2xi32>, tensor<1x2xi32>, i64) -> tensor<4x2xi32>
+// %result: [[1, 2], [3, 4], [5, 6], [7, 8]]
+```
 
 [Back to Ops](#index-of-ops)
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -390,12 +390,11 @@ tensor. More formally,
   * (C2) All tensors in `inputs` have the same shape except for the size of the
   `dimension` dimension.
   * (C3) `inputs` have N tensors where N >= 1.
-  * (C4) `dimension` is non-negative and less than the rank of the tensors in
-  `inputs`.
+  * (C4) 0 $\le$ `dimension` $\lt$ rank of `inputs[0]`.
   * (C5) `result` has the same element type as the tensors in `inputs`.
   * (C6) `result` has the same shape as the tensors in `inputs` except for the
-  size of the `dimension` dimension, which is calculated as a sum of the sizes
-  of this dimension in all tensors in `inputs`.
+  size of the `dimension` dimension, which is calculated as a sum of the size of
+  `inputs[k][dimension]` for all `k` in `inputs`.
 
 ### Examples
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -368,7 +368,7 @@ dimension in the same order as the given arguments and produces a `result`
 tensor. More formally,
 `result[i0, ..., id, ..., iR-1] = inputs[k][i0, ..., kd, ..., iR-1]`, where:
   1. `id = d0 + ... + dk-1 + kd`.
-  2. `d` is equal to `dimension`, and `d0`, ... are `d`'th dimension sizes
+  1. `d` is equal to `dimension`, and `d0`, ... are `d`'th dimension sizes
      of `inputs`.
 
 ### Operands
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -363,47 +363,60 @@ IEEE-754 specification.
 
 ### Semantics
 
-Concatenates sequence of tensors in `val` along dimension `dimension` and
-produces a tensor as a result. The tensors are of the same rank as the input
-operands except the specified dimension.
+Concatenates sequence of tensors in `inputs` along `dimension` dimension in the
+same order as the given arguments and produces a `result` tensor. More formally,
+`result[i0, ..., id, ..., iR-1] = inputs[k][i0, ..., kd, ..., iR-1]`, where:
+  1. `id = d0 + ... + dk-1 + kd`.
+  2. `d` is equal to `dimension`, and `d0`, ... are `d`'th dimension sizes
+     of `inputs`.
 
 ### Operands
 
-| Name        | Type                                                                     |
-|-------------|--------------------------------------------------------------------------|
-| `dimension` | `i64`                                                                    |
-| `val`       | Sequence of tensors of integer, floating-point, or complex element types |
+| Name        | Type                                              |
+|-------------|---------------------------------------------------|
+| `inputs`    | Variadic number of tensors of any supported types |
+| `dimension` | `si64`                                            |
 
 ### Results
 
-| Name     | Type                                                        |
-|----------|-------------------------------------------------------------|
-| `result` | tensor of integer, floating-point, or complex element types |
+| Name     | Type                          |
+|----------|-------------------------------|
+| `result` | tensor of any supported types |
 
 ### Constraints
 
-  * (C1) `val` and `result` have the same type.
-  * (C2) All tensors in `val` have the same element type and rank.
-  * (C3) Tensors of `val` all have the same dimension except for `dimension`.
+  * (C1) All tensors in `inputs` have the same element type.
+  * (C2) All tensors in `inputs` have the same shape except for the size of the
+  `dimension` dimension.
+  * (C3) `inputs` have N tensors where N >= 1.
+  * (C4) `inputs` have tensors of rank > 0.
+  * (C5) `dimension` is non-negative and less than the rank of the tensors in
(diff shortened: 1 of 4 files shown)


docs/spec_draft.md
```diff
@@ -358,6 +359,56 @@ IEEE-754 specification.
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.concatenate
+
+### Semantics
+
+Concatenates sequence of tensors in `val` along dimension `dimension` and
+produces a tensor as a result. The tensors are of the same rank as the input
+operands except the specified dimension.
```
burmako, line 368: Similarly to the feedback that I left for #172, this section doesn't really explain how to obtain the `result` tensor from the `val` tensor. "Concatenates sequence of tensors in `val` along dimension `dimension`" sounds fancy but is not actionable.

Let's define op semantics from first principles, using only the definitions in this spec (see the "Types" section in  [spec_draft.md](https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md#types)) and college maths.

docs/spec_draft.md
```diff
@@ -358,6 +359,56 @@ IEEE-754 specification.
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.concatenate
+
+### Semantics
+
+Concatenates sequence of tensors in `val` along dimension `dimension` and
+produces a tensor as a result. The tensors are of the same rank as the input
```
burmako, line 367: "produces a \`result\` tensor"?

docs/spec_draft.md
```diff
@@ -358,6 +359,56 @@ IEEE-754 specification.
...
+operands except the specified dimension.
+
+### Operands
+
+| Name        | Type                                                                     |
+|-------------|--------------------------------------------------------------------------|
+| `dimension` | `i64`                                                                    |
+| `val`       | Sequence of tensors of integer, floating-point, or complex element types |
```
burmako, line 375: `val` is pretty confusing to read, so let's rename to `inputs`, similarly to #202? Can you take care of the renaming in the StableHLO dialect as well?

docs/spec_draft.md
```diff
@@ -358,6 +359,56 @@ IEEE-754 specification.
...
+
+Concatenates sequence of tensors in `val` along dimension `dimension` and
+produces a tensor as a result. The tensors are of the same rank as the input
+operands except the specified dimension.
+
+### Operands
+
+| Name        | Type                                                                     |
+|-------------|--------------------------------------------------------------------------|
+| `dimension` | `i64`                                                                    |
```
burmako, line 374: You mean `si64`, right?

docs/spec_draft.md
burmako, line 375: "Sequence" may mean many things, including a list type. Let's disambiguate by saying "Variadic number of tensors...".

burmako, line 375: "tensors of all supported types"? (Same feedback for `result`).

docs/status.md
```diff
@@ -53,7 +53,7 @@ one of the following tracking labels.
 | compare                  |         no         |        yes*        |        yes*         |         yes          |        no        |
 | complex                  |         no         |        yes*        |        yes*         |         yes          |        no        |
 | compute_reshape_shape    |         no         |         no         |         no          |         yes          |        no        |
-| concatenate              |         no         |        yes*        |        yes*         |         yes          |        no        |
+| concatenate              |        yes         |        yes*        |        yes*         |         yes          |        no        |
```
burmako, line 56: What about verification and type inference? Do they conform to the newly updated spec?

docs/spec_draft.md
```diff
@@ -358,6 +359,56 @@ IEEE-754 specification.
...
+### Examples
+
+```mlir
+// 1-dimensional concatenate
+
+// %val1: [1, 2]
+// %val2: [3, 4]
+// %val3: [5, 6]
+// %dimension: 0
+%result = "stablehlo.concatenate"(%val1, %val2, %val3, %dimension) : (tensor<2xi32>, tensor<2xi32>, tensor<2xi32>, i64) -> tensor<8xi32>
```
burmako, line 398: The actual MLIR syntax is different. `dimension` will be passed as an attribute, along the lines of `{ dimension = 0 : i64 }`. The type signature will be different as well.

docs/status.md
sdasgup3, line 56: Can we check our spec with verifier and shape inferencer?

docs/spec_draft.md
sdasgup3, line 375: `Sequence of n tensors ...`
In the constraint we can say n > =1

docs/spec_draft.md
sdasgup3, line 368: In which order the tensors are going to be concatenated?
Same order in which the arguments are specified.

docs/spec_draft.md
```diff
@@ -358,6 +359,56 @@ IEEE-754 specification.
...
+
+| Name     | Type                                                        |
+|----------|-------------------------------------------------------------|
+| `result` | tensor of integer, floating-point, or complex element types |
+
+### Constraints
+
+  * (C1) `val` and `result` have the same type.
+  * (C2) All tensors in `val` have the same element type and rank.
+  * (C3) Tensors of `val` all have the same dimension except for `dimension`.
```
burmako, line 387: I have a few feedback items, but I think they'll be easier to share via a proposed rewording:

  * (C1) All tensors in `val` have the same element type.
  * (C2) All tensors in `val` have the same shape except for the size of the `dimension` dimension.
  * (C3) `dimension` is non-negative and less than the rank of the tensors in `val`.
  * (C4) `result` has the same element type as the tensors in `val`.
  * (C5) `result` has the same shape as the tensors in `val` except for the size of the `dimension` dimension, which is calculated as a sum of the sizes of this dimension in all tensors in `val`.

Feedback items:
  1. Let's go from left to right, i.e. first `val`, then `dimension`, then `result`.
  2. I added C3 because constraints need to cover all operands and results.
  3. I added C5 because constraints need to cover everything needed for shape inference.

burmako (changes requested):

docs/spec_draft.md
```diff
@@ -358,6 +359,56 @@ IEEE-754 specification.
...
+
+### Results
+
+| Name     | Type                                                        |
+|----------|-------------------------------------------------------------|
+| `result` | tensor of integer, floating-point, or complex element types |
+
+### Constraints
+
+  * (C1) `val` and `result` have the same type.
```
sdasgup3, line 385: same element type?

docs/spec_draft.md
```diff
@@ -358,6 +359,56 @@ IEEE-754 specification.
...
+  * (C2) All tensors in `val` have the same element type and rank.
```
sdasgup3, line 386: same type? (that includes shape and element type)

docs/spec_draft.md
```diff
@@ -358,6 +359,56 @@ IEEE-754 specification.
...
 
+## stablehlo.concatenate
+
+### Semantics
+
+Concatenates sequence of tensors in `val` along dimension `dimension` and
+produces a tensor as a result. The tensors are of the same rank as the input
+operands except the specified dimension.
+
+### Operands
```
burmako, line 370: Oh, one more thing. Let's change the order of the items in this section to follow the order of definitions in ODS, i.e. `val` first and `dimension` second?

burmako (changes requested):

docs/spec_draft.md
ghpvnist: Ah, I just realized we also support booleans. Thanks for the catch!

docs/spec_draft.md
ghpvnist: I left the more formal notation out of the spec since calculating the actual indices of the `result` tensor is very ugly. Here are some proposals for how to approach this:
1. Keep the semantics the same as [XLA doc](https://www.tensorflow.org/xla/operation_semantics#concatenate) (aka not specifying the details).
2. Write the details as:
Screen Shot 2022-10-03 at 5 28 18 PM
which is a simplified version of:
Screen Shot 2022-10-03 at 5 29 37 PM

where:
$$Left=\prod_{k=0}^{dimension-1}inputs[k]$$
$$Right=\prod_{l=dimension+1}^{R-1}inputs[l]$$

2.5 The lexicographic ordering of the indices of each tensor `t_i` is calculated such that it takes `x` elements from each tensor `t_i` until all elements are used, where `x` is calculated as $$\prod_{l=dimension}^{R-1}inputs[l]$$

An example of the above:
```
inputs = {
[ 0,  1,  2,  3,
  7,  8,  9, 10,
 14, 15, 16, 17,
 21, 22, 23, 24], // 2x2x4
[ 4, 11, 18, 25], // 2x2x1
[ 5,  6, 12, 13,
 19, 20, 26, 27] //2x2x2
}
dimension=2
```
`Right` is the product of the shapes "right of" the specified dimension plus the size of `dimension`, in this case just 4, 1, and 2 respectively for each `t` in `input` ("Left of" `dimension=2` would be `2x2=4` for all `t` in `input`). We take 4 elements from `t0`, 1 from `t1`, 2 from `t2` until all elements are taken, which result in:
```
result = [[ 0,  1,  2,  3,  4,  5,  6,
            7,  8,  9, 10, 11, 12, 13],
          [14, 15, 16, 17, 18, 19, 20,
           21, 22, 23, 24, 25, 26, 27]]
```
3. If you have a much simpler way to express this, please let me know!

sdasgup3: May be something along this line:

Given,
`vals` = {V0, V1, ..., Vn}
$d_{i}^{j}$ = dimension size of jth dimension in V[j] 

With the concatenation of `vals` along dimension `d`, the value of `result` at index `i0, i1,...,  id, ..., iR-1` is given by
```
result[i0, i1,...,  id, ..., iR-1] = V[k][i0, i1,...,  id', ..., iR-1]
```
such that
$id <  \sum(i=1, n, d_{i}^{d})$ and 
there exist a `k` such that $id > \sum(i=1, k-1, d_{i}^{d}) - 1$

and id' is defined as $id - \sum(i=1, k-1, d_{i}^{d})$

docs/spec_draft.md
sdasgup3, line 374: `si64`?

docs/spec_draft.md
burmako: @ghpvnist @sdasgup3 Thank you for the analysis of how the more formal spec for `concatenate` might look like! I agree that formulas that involve the canonical representation are pretty much inscrutable, and I like the idea of using multidimensional indices instead. How about we simplify that a bit further along the following lines:

```
Concatenates sequence of tensors in `inputs` along dimension `dimension` and
produces a tensor as a result. More formally, 
`result[i0, i1, ..., id, ..., iR-1] = inputs[k][i0, i1, ..., kd, ..., iR-1]`,
where:
  1. `id = d0 + d1 + ... + dk-1 + kd`.
  1. `d` is equal to `dimension`, and `d0`, `d1`, ... are `d`th dimension 
     sizes of `inputs`.
```

burmako (changes requested):

docs/spec_draft.md
sdasgup3: Thanks @burmako for simplifying so much.

ghpvnist: Thanks both @burmako @sdasgup3 for your input!

docs/spec_draft.md
ghpvnist: Will fix to `si64` in the table, but keep `i64` in the examples, which is related to #22.

ghpvnist: Will fix to `si64` in the table, but keep `i64` in the examples, which is related to #22.

docs/status.md
ghpvnist: I checked that it follows the spec except one constraint. Currently, verify does not check that size of `inputs` is at least one:
* (C3) `inputs` have N tensors where N >= 1.

ghpvnist: No. Currently, verify does not check that size of `inputs` is at least one:
* (C3) `inputs` have N tensors where N >= 1.

sdasgup3: If HLO support concatenate with zero args, then maybe we should also support that? 

However it seems that the verifier does check that
https://github.com/openxla/stablehlo/blob/4859634f3d95851a7a55c16cf9d753327a57c860/stablehlo/dialect/StablehloOps.cpp#L3022

Also the HLO semantics.

ghpvnist: Ah I didn't catch that part. Yes, then concatenate is in sync.

ghpvnist: Looks like the verifier does check it actually, https://github.com/openxla/stablehlo/blob/4859634f3d95851a7a55c16cf9d753327a57c860/stablehlo/dialect/StablehloOps.cpp#L3022
so I'll mark it as `yes`.

docs/status.md
```diff
@@ -53,7 +53,7 @@ one of the following tracking labels.
 | compare                  |         no         |        yes*        |        yes*         |         yes          |        no        |
 | complex                  |         no         |        yes*        |        yes*         |         yes          |        no        |
 | compute_reshape_shape    |         no         |         no         |         no          |         yes          |        no        |
-| concatenate              |         no         |        yes*        |        yes*         |         yes          |        no        |
+| concatenate              |        yes         |        yes         |        yes*         |         yes          |        no        |
```
burmako, line 56: Let's also update the ODS for `concatenate` with the link to the spec?

burmako (changes requested):

stablehlo/dialect/StablehloOps.td
```diff
@@ -1797,19 +1797,11 @@ def StableHLO_ConcatenateOp : StableHLO_ShapedInterfaceOp<"concatenate",
...
   let description = [{
-     Concatenates a set of tensors along the specified dimension.
-
-     See https://www.tensorflow.org/xla/operation_semantics#concatenate.
-
-     Example:
-
-     ```mlir
-     %0 = stablehlo.concatenate %arg0, %arg1, dim = 1 : (tensor<4x1xf32>, tensor<4x2xf32>) -> tensor<4x3xf32>
-     ```
```
burmako, line 1808: Let's wait for @GleasonK to chime in on what we should do with the pretty syntax. Also see https://github.com/openxla/stablehlo/pull/250#discussion_r987523267.

burmako (changes requested):

docs/spec_draft.md
```diff
@@ -358,6 +359,70 @@ IEEE-754 specification.
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.concatenate
+
+### Semantics
+
+Concatenates sequence of tensors in `inputs` along `dimension` dimension in the
```
burmako, line 366: "a variadic number of tensors" here as well?

docs/spec_draft.md
```diff
@@ -358,6 +359,70 @@ IEEE-754 specification.
...
+|----------|-------------------------------|
+| `result` | tensor of any supported types |
+
+### Constraints
+
+  * (C1) All tensors in `inputs` have the same element type.
+  * (C2) All tensors in `inputs` have the same shape except for the size of the
+  `dimension` dimension.
+  * (C3) `inputs` have N tensors where N >= 1.
+  * (C4) `inputs` have tensors of rank > 0.
```
burmako, line 392: It looks like (C4) follows from (C6), so let's drop it?

docs/spec_draft.md
```diff
@@ -358,6 +359,70 @@ IEEE-754 specification.
...
+  * (C5) rank of tensors in `inputs` and `result` are the same.
```
burmako, line 393: Similarly, (C5) follows from (C8).

docs/status.md
burmako, line 56: Should we also change "Type Inference" to "yes"?

burmako (changes requested):

docs/spec_draft.md
```diff
@@ -358,6 +359,69 @@ IEEE-754 specification.
...
+| `result` | tensor of any supported types |
+
+### Constraints
+
+  * (C1) All tensors in `inputs` have the same element type.
+  * (C2) All tensors in `inputs` have the same shape except for the size of the
+  `dimension` dimension.
+  * (C3) `inputs` have N tensors where N >= 1.
+  * (C4) `dimension` is non-negative and less than the rank of the tensors in
+  `inputs`.
```
burmako, line 394: In a similar situation for `iota`, we say "0 $\le$ `iota_dimension` $\lt$ `R`, where `R` is the rank of the `result`". Let's say something similar here - that'd be easier to read.

burmako (approved):
While we're waiting for Kevin's reply, here's one more minor comment, and we're good to go.

stablehlo/dialect/StablehloOps.td
GleasonK: (Going through my notifications now, pasting the resolution from #250 for easy access)

> How about instead of removing everything from the description, we keep: 1) one sentence that describes the op, 2) pretty syntax, 3) link to spec?

PR #250, opened 2022-10-03
Update ODS links to reference StableHLO spec

ghpvnist:
Updates the ODS description to include:
* One sentence description from the spec
* Link to the spec
* Pretty print syntax

The changes in `status.md` for Verification and Type Inference columns are also now in sync with the spec.

### DIFF
--- a/docs/status.md
+++ b/docs/status.md
@@ -29,8 +29,8 @@ one of the following tracking labels.
 
 | StableHLO Op             | Specification | Verification |  Type Inference   |  Pretty Printing  | Interpreter |
 |:-------------------------|:-------------:|:------------:|:-----------------:|:-----------------:|:-----------:|
-| abs                      |      yes      |     yes*     |       yes*        |        yes        |     no      |
-| add                      |      yes      |     yes*     |       yes*        |        yes        |     yes     |
+| abs                      |      no       |     yes*     |        yes        |        yes        |     no      |
+| add                      |      no       |     yes*     |        yes        |        yes        |     yes     |
 | after_all                |      no       |      no      |        no         |        yes        |     no      |
 | all_gather               |      no       |     yes*     |        no         |        no         |     no      |
 | all_reduce               |      no       |      no      |        no         |        no         |     no      |
@@ -62,7 +62,7 @@ one of the following tracking labels.
 | cross-replica-sum        |      no       |      no      |       yes*        |        no         |     no      |
 | cstr_reshapable          |      no       |     yes*     |        no         |        yes        |     no      |
 | custom_call              |      no       |     yes*     |    infeasible     |        yes        |     no      |
-| divide                   |      yes      |     yes*     |       yes*        |        yes        |     no      |
+| divide                   |      yes      |      no      |        yes        |        yes        |     no      |
 | dot                      |      no       |     yes*     | yes(need-revisit) |        yes        |     no      |
 | dot_general              |      no       |     yes*     |       yes*        |        no         |     no      |
 | dynamic_broadcast_in_dim |      no       |     yes*     |        no         |        no         |     no      |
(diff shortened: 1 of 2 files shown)


sdasgup3 (approved):
I am assuming that we have gone through checking if the spec draft is in sync with the verifier/shape inferencer.

burmako (approved):

burmako (changes requested):
Sandeep made an excellent point! We can replace descriptions in ODS with the link to the spec only after Verification and Type Inference columns in `status.md` have been flipped to `yes`. Let's update `status.md` as well?

ghpvnist:
The PR has now been updated with the Verification and Type Inference columns in `status.md` to be in sync with the spec.

docs/status.md
```diff
@@ -54,15 +54,15 @@ one of the following tracking labels.
...
-| cosine                   |        yes         |        yes*        |        yes*         |         yes          |       yes        |
+| cosine                   |        yes         |        yes         |         yes         |         yes          |       yes        |
 | create_token             |         no         |        yes*        |         no          |         yes          |        no        |
 | cross-replica-sum        |         no         |         no         |        yes*         |          no          |        no        |
 | cstr_reshapable          |         no         |        yes*        |         no          |         yes          |        no        |
 | custom_call              |         no         |        yes*        |     infeasible      |         yes          |        no        |
-| divide                   |        yes         |        yes*        |        yes*         |         yes          |        no        |
+| divide                   |        yes         |        yes         |         yes         |         yes          |        no        |
```
burmako, line 65: StableHLO ODS allows boolean tensors in `divide` which doesn't make sense, while the spec and [shape_inference.cc](https://github.com/tensorflow/tensorflow/blob/c277b0b396b7709d4355d5794d380db90c22cc6a/tensorflow/compiler/xla/service/shape_inference.cc#L984-L999) don't. Looks like yet another ticket will be needed to align things. We're getting a lot of mileage out of this supposedly rountine PR! (Same for `remainder` and `subtract`).

burmako (changes requested):

stablehlo/dialect/StablehloOps.td
```diff
@@ -192,16 +192,7 @@ def StableHLO_AbsOp: StableHLO_UnaryElementwiseOp<"abs",
...
-    Returns `abs(operand)` element-wise.
-
-    See
-    https://www.tensorflow.org/xla/operation_semantics#element-wise_unary_functions.
-
-    Example:
-
-    ```mlir
-    %0 = stablehlo.abs %arg0 : tensor<3xi32>
-    ```
```
burmako, line 204: @GleasonK What's your take on how our documentation in the ODS should look like? I think it's a reasonable idea to reduce it to just a link to the spec if it exists, but this leaves an open question on what we should do with pretty syntax examples (because currently that's not part of the spec). What do you think?

burmako (changes requested):

stablehlo/dialect/StablehloOps.td
GleasonK: I see no reason to avoid pretty syntax in the doc examples, that could suffice. For example, the current doc has the following:

```
// integers
// %operand: [-2, 0, 2]
%result = "stablehlo.abs"(%operand) : (tensor<3xi32>) -> tensor<3xi32>
// %result: [2, 0, 2]
```

Could be:

```
// integers
// %operand: [-2, 0, 2]
%result = stablehlo.abs %operand : tensor<3xi32>
// %result: [2, 0, 2]
```

It is worth noting that removing this from `.td` files causes some incongruity between MHLO/StableHLO, since MHLO doc is generated directly from the ODS, so the pretty print example should appear in the `.td` for MHLO. However, given that this is not a design goal to keep StableHLO/MHLO docs in sync, this is fine with me.

ghpvnist: The reason for using the extended form is noted in #191, but indeed our ODS and spec doc do not provide consistency for which form is recommended.

burmako: I propose that we don't include prettyprinting in the spec because that'll be one more thing that spec readers will need to know about.

However, on a second thought, I think it would be good to keep pretty syntax in the ODS because that's part of the op's documentation. How about instead of removing everything from the description, we keep: 1) one sentence that describes the op, 2) pretty syntax, 3) link to spec?

ghpvnist: SGTM

sdasgup3:
Hi Gunhyun
Can you please accommodate the addition of spec links for minimum and maximum operations in this PR. They already have the  short description plus a pretty print example.

burmako (changes requested):
Final request, and we're good to go. 

Given the discoveries made during the review of this PR (#262 and #263), can we update status.md accordingly? Namely:
  * Change "Specification" for `abs`, `add`, `maximum` and `minimum` to "no" (their spec will need to be updated to account for booleans, and then this can change back to "yes").
  * Change "Verification" for `divide`, `remainder` and `subtract` to "no" (their ODS will need to be updated to disallow booleans, and then this can change back to "no").
  * Change "Type Inference" for `abs`, `add`, `divide`, `minimum`, `maximum`, `remainder` and `subtract` to "yes" (type inference is correct, what's incorrect is verification).

burmako (approved):

PR #253, opened 2022-10-04
Update status of AllReduce in type inference.

zhouxin913:
Update status of AllReduce in type inference.

- Change in column "Type Inference" for `all_reduce` op -> yes(need-revisit)
- Change in column "Verification" for many ops -> yes(need-revisit) whose "type inference" column is not yes or yes*. The reason is: while revisiting the type inference of an op, the responding verification will always be revisited. For infeasible ones, the column "Verification" is set to "yes(need-revisit)" also.
- Adjust the summary number in column title accordingly.


docs/status.md
```diff
@@ -27,21 +27,21 @@ one of the following tracking labels.
 
 ## Status
 
-| StableHLO Op (114)       | Specification (25) | Verification (104) | Type Inference (82) | Pretty Printing (77) | Interpreter (11) |
+| StableHLO Op (114)       | Specification (25) | Verification (114) | Type Inference (83) | Pretty Printing (77) | Interpreter (11) |
```
burmako, line 30: What does "Verification (114)" mean in this context? I would expect that it would be a number of ops, where we've done everything we can at the moment (i.e. "yes" or "yes*"), but it looks like it also includes ops that need revisiting? Same question for type inference.

docs/status.md
```diff
@@ -27,21 +27,21 @@ one of the following tracking labels.
```
burmako, line 27: The description of the PR seems out of date. E.g. "Change in column "Type Inference" for all_reduce op -> yes(need-revisit)" doesn't elaborate on why the "Type Inference" column has changed for other ops.

burmako (changes requested):
I'll take a detailed look tomorrow. Meanwhile, left a couple comments.

docs/status.md
```diff
@@ -52,94 +52,94 @@ one of the following tracking labels.
...
+| reduce_scatter           |         no         | yes(need-revisit)  |         no          |          no          |        no        |
 | reduce_window            |         no         |        yes*        |        yes*         |          no          |        no        |
 | remainder                |        yes         |        yes*        |        yes*         |         yes          |        no        |
-| replica_id               |         no         |        yes*        |  yes(need-revisit)  |         yes          |        no        |
-| reshape                  |        yes         |        yes         |     infeasible      |         yes          |       yes        |
-| return                   |         no         |        yes*        |         no          |         yes          |        no        |
+| replica_id               |         no         | yes(need-revisit)  |  yes(need-revisit)  |         yes          |        no        |
+| reshape                  |        yes         | yes(need-revisit)  |     infeasible      |         yes          |       yes        |
```
burmako, line 113: I don't think that we need to revisit `reshape`. #170 has already determined that it's a "yes".

burmako (changes requested):

docs/status.md
zhouxin913: I clarify this the new description. Please see accordingly. Numbers are removed previous PRs from other contributors.

docs/status.md
zhouxin913: "why the "Type Inference" column has changed for other ops"
There's no other changes in column "Type Inference" other than all_reduce op. Please let me know if the diff page shows diffidently on you side. I update (rebase) the PR, but no more updates on the column "Type Inference".

docs/status.md
zhouxin913: Thanks! in the new version updated today, it respects all the recent changes to the repo.

burmako:
Thank you for the thorough work in putting together this PR! I've finally found several hours to review it and make sure that it matches reality. This resulted in about a dozen fixes and a few stylistic changes, so I sent them out as a separate pull request (#331) which supersedes this PR.

PR #265, opened 2022-10-05
Fix lit checks with ops having implementation defined precision

sdasgup3:
The current mode of testing the implementation is to use LLVM's lit-test framework to syntactically match the output of the operation with a (user provided) golden output. We do have some ops (like sine, cosine, tanh) which might produce results with implementation defined precision, failing the lit checks. 

To avoid that this PR allows the lit checks to match the accuracy up to 12 places after the decimal point. The number 12 is chosen   arbitrary but provides a descent approximation of the checked values.

### DIFF
--- a/stablehlo/tests/interpret_cosine.mlir
+++ b/stablehlo/tests/interpret_cosine.mlir
@@ -71,9 +71,9 @@ func.func @cosine_op_test_f64() -> tensor<11xf64> {
   // CHECK-NEXT: tensor<11xf64>
   // CHECK-NEXT: 1.000000e+00 : f64
   // CHECK-NEXT: 1.000000e+00 : f64
-  // CHECK-NEXT: 0.54030230586813977 : f64
-  // CHECK-NEXT: 0.992197667229329 : f64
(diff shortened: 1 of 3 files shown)


burmako (changes requested):
How do we decide where in the tests we should make these changes? Perhaps we could update interpreter.md with more details?

Also, how do we guard ourselves against these failures in the future? E.g. would it be possible for our test suite to catch this too?

sdasgup3:
>> We could try an experiment of adding Mac tests.
I assume you meant running on a mac machine. Yes, that might have the same problem as well. Different hardware/compiler can trigger the same.

GleasonK (approved):
LGTM.

Is there any guarantee that this will work for the foreseeable future? I.e. how arbitrary is the number of decimal places?

burmako (approved):
Let's merge to unblock the integrate and have a ticket to address that in a more principled way later.

sdasgup3:
Have the [ticket](https://github.com/openxla/stablehlo/issues/268) in place to track this.

PR #266, opened 2022-10-05
Remove count from status.md column

ghpvnist:
Currently, we are updating the count for each column when we are updating the status of the columns. This can be useful, but it is error-prone and removing this avoids tedious work verifying the count.


burmako (approved):
GitHub's diff is very helpful in highlighting that no actual content in the table body is changed. Super nice that I don't have to verify this manually :)

PR #267, opened 2022-10-05
Add spec for PadOp

### LINKED ISSUE #243: Add spec for PadOp

ghpvnist:
closes #243

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -1001,11 +1001,12 @@ the interior-padded operand.
 ### Constraints
 
   * (C1) `operand`, `padding_value`, `result` have the same element type.
-  * (C2) `padding_value` has rank 0.
-  * (C3) `edge_padding_low`, `edge_padding_high`, `interior_padding` have the
+  * (C2) `edge_padding_low`, `edge_padding_high`, `interior_padding` have the
   size equal to `operand`'s rank.
-  * (C4) 0 $\le$ `interior_padding[i]` for all `i` values in `interior_padding`.
-  * (C5) `dim(result, i) = di + max(di - 1, 0) * interior_padding[i] + edge_padding_low[i] + edge_padding_high[i]`, where `di = dim(operand, i)`.
+  * (C3) 0 $\le$ `interior_padding[i]` for all `i` values in `interior_padding`.
+  * (C4) 0 $\le$ `dim(result, i)` for all `i`th dimension of `operand`, where
+  `dim(result, i) = di + max(di - 1, 0) * interior_padding[i] + edge_padding_low[i] + edge_padding_high[i]`
+  and `di = dim(operand, i)`.
 
 ### Examples
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -167,6 +167,7 @@ described below)
    * [negate](#stablehlonegate)
    * [not](#stablehlonot)
    * [or](#stablehloor)
+   * [pad](#stablehlopad)
    * [remainder](#stablehloremainder)
    * [reshape](#stablehloreshape)
    * [reverse](#stablehloreverse)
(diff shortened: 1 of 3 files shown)


docs/spec_draft.md
```diff
@@ -838,6 +839,74 @@ operation.
...
+elements in each dimension which may not be negative. Interior padding occurs
+before edge padding such that negative edge padding will remove elements from
+the interior-padded operand.
+
+### Operands
+
+| Name                | Type                               |
+|---------------------|------------------------------------|
+| `operand`           | tensor of any supported types      |
+| `padding_value`     | tensor of any supported types      |
```
sdasgup3, line 865: padding is a scalar value?

docs/spec_draft.md
```diff
@@ -838,6 +839,74 @@ operation.
...
+
+| Name     | Type                          |
+|----------|-------------------------------|
+| `result` | tensor of any supported types |
+
+### Constraints
+
+  * (C1) `operand` and `result` have the same element type.
+  * (C2) `operand` and `padding_value` have the same element type.
+  * (C3) `padding_value` should be rank 0.
```
sdasgup3, line 880: The table above should mention that the padding value is a scalar.

docs/spec_draft.md
```diff
@@ -838,6 +839,74 @@ operation.
...
+  * (C4) `edge_padding_low`, `edge_padding_high`, `interior_padding` have the
+  same rank as `operand`.
+  * (C5) `interior_padding` >= 0.
+  * (C6) Padding should not result in negative size for `result` dimensions.
```
sdasgup3, line 884: Can we infer the shape of the result as well?

docs/spec_draft.md
ghpvnist: Padding value is indeed scalar, but the ODS accepts HLOTensor, and the verifier checks that this is rank 0. This is the case with https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/shape_inference.cc#L566-L569 and https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/StablehloOps.cpp#L4451-L4456

docs/spec_draft.md
ghpvnist: I think this can be inferred from C3, since scalars must be rank 0.

docs/spec_draft.md
ghpvnist: Can you clarify? The shape of the result is inferred in https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/StablehloOps.cpp#L4518.

sdasgup3: Ah sorry that I was not clear:
For ops like add, we are putting constraint on the shape of its result like "same shape as the operand". Can we add a constraint on the result of PadOp based on the edge and and inferior padding values for each dimension.

>>The shape of the result is inferred in

Yes that is right that we are already inferring that but IMO spec is SoT from where all those other artifacts (verifier/shape inference) should be inferred. 

For example, the dimension size  of the ith dimension of the `result` is given by 

`di + max(di - 1, 0 ) *  interior_padding[i] + edge_padding_low[i] + edge_padding_high[i]`

wdyt?

ghpvnist: Ah thanks for clarifying.

>> For ops like add, we are putting constraint on the shape of its result like "same shape as the operand". Can we add a constraint on the result of PadOp based on the edge and and inferior padding values for each dimension.

Since pad will fundamentally change the size of each dimension, shape can be different. However, you did bring up a good point since the ranks should remain unchanged for `result`. I wonder whether this is worth stating in (C4) since this information can be somewhat inferred and inherent from the op itself since it is "padding" dimensions not "adding" dimensions.

>>For example, the dimension size of the ith dimension of the result is given by
>>`di + max(di - 1, 0 ) * interior_padding[i] + edge_padding_low[i] + edge_padding_high[i]`, where di is the dimension size of ith dimension of the operand`

Makes sense!

docs/status.md
```diff
@@ -98,7 +98,7 @@ one of the following tracking labels.
 | optimization_barrier     |      no       |     yes*     |        no         |        yes        |     no      |
 | or                       |      yes      |     yes*     |       yes*        |        yes        |     no      |
 | outfeed                  |      no       |     yes*     |        no         |        no         |     no      |
-| pad                      |      no       |     yes*     |       yes*        |        no         |     no      |
+| pad                      |      yes      |     yes      |        yes        |        no         |     no      |
```
burmako, line 101: Also update StablehloOps.td?

docs/spec_draft.md
```diff
@@ -838,6 +839,74 @@ operation.
...
+
+```mlir
+// %operand: [
+//            [1, 2, 3],
+//            [4, 5, 6]
+//           ]
+// %padding_value: 0
+%result = "stablehlo.pad"(%operand, %padding_value) {
+  edge_padding_low = dense<[0, 1]> : tensor<2xi64>,
+  edge_padding_high = dense<[2, 1]> : tensor<2xi64>
```
burmako, line 896: Missing comma at the end of the line

burmako (changes requested):

docs/spec_draft.md
ghpvnist: Ah I misunderstood you for adding a new constraint. I'll update the table to say:
```
scalar tensor of any supported types
```

burmako: I don't think that we introduce the notion of a "scalar tensor" in the spec. The closest we have is "Tensors of rank zero, e.g tensor<f32>, have 1 element. Such tensors are allowed and are useful to model scalars", but that's not a canonical definition imo. Let's stick to "0-dimensional tensor"?

A bigger question I think is which constraints should go where. At the moment, it looks like we are specifying element types + ranks of attributes in Operands/Results, and everything else in Constraints. Is the proposal to also specify ranks of tensors in Operands/Result?

docs/spec_draft.md
ghpvnist: I like your last suggestion. However, we will still need to keep the constraint that `dim(result, i)` cannot be negative since the formula will still allow negative dimensions. 

How about:
0 $\le$ `dim(result, i)` for all `i`th dimension of `operand`, where `dim(result, i) = di + max(di - 1, 0) * interior_padding[i] + edge_padding_low[i] + edge_padding_high[i]` and `di = dim(operand, i)`.

docs/spec_draft.md
ghpvnist: >Let's stick to "0-dimensional tensor"?

Sure.

>Is the proposal to also specify ranks of tensors in Operands/Result?

Imo, the element types should be in the table since that is what the "Type" part is meant to describe. As for the ranks, similar reasoning. Mostly, the constraints bullet points attempt to match the verification and type inference in ODS and StablehloOps.cpp, at least that's what I've been doing.

I agree that we don't have a clear boundary as to which should go where, but I propose we keep element type and rank stuff in the table description and add them to the Constraints section if the rank or element requirements need to address multiple rows of the Operands table (i.e. `operand` and `result` have the same element type vs. `operand` is a tensor of any supported types). Wdyt?

docs/spec_draft.md
```diff
@@ -999,13 +999,12 @@ the interior-padded operand.
...
 
-  * (C1) `operand` and `result` have the same element type.
-  * (C2) `operand` and `padding_value` have the same element type.
-  * (C3) `padding_value` should be rank 0.
-  * (C4) `edge_padding_low`, `edge_padding_high`, `interior_padding` have the
-  same rank as `operand`.
-  * (C5) `interior_padding` >= 0.
-  * (C6) Padding should not result in negative size for `result` dimensions.
+  * (C1) `operand`, `padding_value`, `result` have the same element type.
+  * (C2) `padding_value` has rank 0.
```
burmako, line 1003: Sounds like we no longer need (C2) now that it's part of the type?

docs/spec_draft.md
burmako: Sounds good!

docs/spec_draft.md
burmako: +1

burmako (approved):
A few minor comments, and we're good to go.

Issue #263, opened 2022-10-05
Supported type mismatch for AbsOp between spec and HLO semantics


sdasgup3:
### Request description

Follow up on [comment ](https://github.com/openxla/stablehlo/pull/250#discussion_r986278277)

As per HLO semantics, AbsOp only supports signed integers, where the StableHLO semantics supports unsigned as well. The later added  because of the similarity of AbsOp with NegateOp for which HLO semantics[ supports both](https://github.com/tensorflow/tensorflow/blob/475c1da0dcd0a3b13b5631b1ded8aaf16802bd01/tensorflow/compiler/xla/service/shape_inference.cc#L307).

We might have to figure out if we should diverge from HLO?

burmako:
"We might have to figure out if we should diverge from HLO?". For the initial version of the spec, I don't think we should diverge from HLO. Let's just align the StableHLO behavior with what HLO does and then think about changes after the spec is completed.

Issue #262, opened 2022-10-05
Ops unexpectedly supporting boolean types in the ODS


sdasgup3:
### Request description

This ticket will track the following 

https://github.com/openxla/stablehlo/pull/250#discussion_r986432897 
https://github.com/openxla/stablehlo/pull/240#discussion_r987531422 and
https://github.com/openxla/stablehlo/pull/250#discussion_r986436241

We have the following list of ops for which the  [StableHLO semantics](https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md)  does not support boolean type which seems expected.
However, the [StableHLO ODS](https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/StablehloOps.td)  and/or  the [HLO semantics](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/shape_inference.cc) differs in that respect. 

| Ops   | bool type supported in StableHLO semantics | bool type supported in ODS | bool type supported in HLO Semantics |
|-------|:---------------------:|:-----:|:---------------:|
| add | No                  | Yes | Yes           |
| multiply | No                  | Yes | Yes           |
| maximum | No                  | Yes | Yes           |
| minimum | No                  | Yes | Yes           |
| divide | No | Yes | No | 
| remainder | No |  Yes | No | 
| subtract | No | Yes | No | 

For the first 4 ops, StableHLO spec might support boolean types as well while clarifying the behavior in the spec.  
 - AddOp: ~For boolean the  behavior is like OrOp~ upd: The behavior like XorOp.
 -  MulOp: For boolean the  behavior is like AndOp
 -  Maximum/Minimum: For boolean the  behavior is `true` > `false`

For the others ODS might need to be updated.

ghpvnist:
Can you also add a row for `abs` op?

sdasgup3:
For abs op we have a separate [ticket](https://github.com/openxla/stablehlo/issues/263)

burmako:
Remaining work: disallow `i1` in DivOp/RemOp/SubOp.

PR #276, opened 2022-10-07
Flip Verification status of abs, add, minimum and maximum

burmako:
As established in #250, #262 and #263, verifiers for these ops are actually correct. It's the spec which is incorrect.


ghpvnist:
Since the doc mentions the following and the verify is ***not*** in sync with StableHLO semantics, should we still keep those as `yes*`?
https://github.com/openxla/stablehlo/blob/7babe1eb2dd85fd4c8888d1c0bd8fc28b9cd9283/docs/status.md?plain=1#L22-L26

burmako:
That's a good point. From that perspective, we should do it the other way around and downgrade Type Inference to yes*. I've updated the PR.

ghpvnist (approved):

PR #280, opened 2022-10-07
Add spec for SliceOp

### LINKED ISSUE #244: Add spec for SliceOp

ghpvnist:
closes #244

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -174,6 +174,7 @@ described below)
    * [reverse](#stablehloreverse)
    * [rsqrt](#stablehlorsqrt)
    * [sine](#stablehlosine)
+   * [slice](#stablehloslice)
    * [sqrt](#stablehlosqrt)
    * [subtract](#stablehlosubtract)
    * [tanh](#stablehlotanh)
@@ -1314,6 +1315,78 @@ Numeric precision is implementation-defined.
 ```
 
 &nbsp;[More Examples](../stablehlo/tests/interpret_sine.mlir)
+
+[Back to Ops](#index-of-ops)
+
+## stablehlo.slice
+
+### Semantics
+
+Extracts a sub-tensor from the `operand` using the dimension indices from the
+operands.
+
+More formally, `result[i0, ..., ik, ..., iR-1] = operand[j0+s0, ..., jk+sk, ..., jR-1+sR-1]`
+for all `sk` in `strides` for all `jk` in `start_indices[k]` $\leq$ `jk` $\lt$
+`limit_indices[k]` where 0 $\leq$ `k` $\lt$ rank(`operand`).
+
+Shape is calculated by `dim(result, i)` = $\lceil$`(limit_indices[i]-start_indices[i])/stride[i]`$\rceil$
+for `i`th dimension size of `operand`.
+
+### Operands
+
+| Name            | Type                                                                                                                                    |
+|-----------------|-----------------------------------------------------------------------------------------------------------------------------------------|
+| `operand`       | tensor of any supported types                                                                                                           |
+| `start_indices` | 1-dimensional array of integers containing the starting indices of the slice for each dimension                                         |
+| `limit_indices` | 1-dimensional array of integers containing the ending indices (exclusive) for the slice for each dimension                              |
+| `strides`       | 1-dimensional array of integers that decides the input stride of the slice. The slice picks every `strides[d]` element in `d` dimension |
+
+### Results
+
+| Name     | Type                          |
+|----------|-------------------------------|
+| `result` | tensor of any supported types |
+
+### Constraints
+
+  * (C1) `operand` and `result` have the same element type.
+  * (C2) `start_indices`, `limit_indices`, and `strides` have the same type.
+  * (C3) size(`start_indices`) = size(`limit_indices`) = size(`strides`) =
+  rank(`operand`).
+  * (C4) 0 $\le$ `start_indices[d]` $\le$ `limit_indices[d]` $\le$
+  size(`operand[d]`) for all dimension `d`.
+  * (C5) 0 $\lt$ `strides[d]` for all dimension `d`.
+
+### Examples
+
+```mlir
+// 1-dimensional slice
+
+// %operand: [0, 1, 2, 3, 4]
+%result = "stablehlo.slice" (%operand) {
+  start_indices = dense<2> : tensor<i64>,
+  limit_indices = dense<4> : tensor<i64>,
+  strides = dense<1> : tensor<i64>
+} : (tensor<5xi64>) -> tensor<2xi64>
+// %result: [2, 3]
+
+// 2-dimensional slice
+
+// %operand: [
+//            [0, 0, 0, 0],
+//            [0, 0, 1, 1],
+//            [0, 0, 1, 1]
+//           ]
+%result = "stablehlo.slice" (%operand) {
+  start_indices = dense<[1, 2]> : tensor<2xi64>,
+  limit_indices = dense<[3, 4]> : tensor<2xi64>,
+  strides = dense<1> : tensor<2xi64>
+} : (tensor<3x4xi64>) -> tensor<2x2xi64>
+// % result: [
+//            [1, 1],
+//            [1, 1]
+//           ]
+```
 
 [Back to Ops](#index-of-ops)
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -1354,9 +1354,9 @@ More formally, `result[i0, ..., iR-1] = operand[j0, ..., jR-1]` where
   * (C4) 0 $\le$ `start_indices[d]` $\le$ `limit_indices[d]` $\le$
   `dim(operand, d)` for all dimension `d`.
   * (C5) 0 $\lt$ `strides[d]` for all dimension `d`.
-  * (C6) Shape is formulated as `dim(result, i)` =
+  * (C6) `dim(result, d)` =
   $\lceil$`(limit_indices[d]-start_indices[d])/stride[d]`$\rceil$ for all
-  dimension `d`.
+  dimension `d` in `operand`.
 
 ### Examples
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -1348,13 +1348,12 @@ More formally, `result[i0, ..., iR-1] = operand[j0, ..., jR-1]` where
 ### Constraints
 
   * (C1) `operand` and `result` have the same element type.
-  * (C2) `operand` and `result` have the same rank.
-  * (C3) size(`start_indices`) = size(`limit_indices`) = size(`strides`) =
+  * (C2) size(`start_indices`) = size(`limit_indices`) = size(`strides`) =
   rank(`operand`).
-  * (C4) 0 $\le$ `start_indices[d]` $\le$ `limit_indices[d]` $\le$
+  * (C3) 0 $\le$ `start_indices[d]` $\le$ `limit_indices[d]` $\le$
   `dim(operand, d)` for all dimension `d`.
-  * (C5) 0 $\lt$ `strides[d]` for all dimension `d`.
-  * (C6) `dim(result, d)` =
+  * (C4) 0 $\lt$ `strides[d]` for all dimension `d`.
+  * (C5) `dim(result, d)` =
   $\lceil$`(limit_indices[d]-start_indices[d])/stride[d]`$\rceil$ for all
   dimension `d` in `operand`.
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -1322,24 +1322,22 @@ Numeric precision is implementation-defined.
 
 ### Semantics
 
-Extracts a sub-tensor from the `operand` using the dimension indices from the
-operands.
+Extracts a sub-tensor from the `operand` and produces a `result` tensor.
+`start_indices` contain the starting indices of the slice for each dimension,
+`limit_indices` contain the ending indices (exclusive) for the slice for each
+dimension, and `strides` contain the strides for each dimension.
 
-More formally, `result[i0, ..., ik, ..., iR-1] = operand[j0+s0, ..., jk+sk, ..., jR-1+sR-1]`
-for all `sk` in `strides` for all `jk` in `start_indices[k]` $\leq$ `jk` $\lt$
-`limit_indices[k]` where 0 $\leq$ `k` $\lt$ rank(`operand`).
-
-Shape is calculated by `dim(result, i)` = $\lceil$`(limit_indices[i]-start_indices[i])/stride[i]`$\rceil$
-for `i`th dimension size of `operand`.
+More formally, `result[i0, ..., iR-1] = operand[j0, ..., jR-1]` where
+`jd = start_indices[d] + id * strides[d]`.
 
 ### Operands
 
-| Name            | Type                                                                                                                                    |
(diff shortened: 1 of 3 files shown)


docs/spec_draft.md
```diff
@@ -1243,6 +1244,78 @@ Numeric precision is implementation-defined.
...
+### Semantics
+
+Extracts a sub-tensor from the `operand` using the dimension indices from the
+operands.
+
+More formally, `result[i0, ..., ik, ..., iR-1] = operand[j0+s0, ..., jk+sk, ..., jR-1+sR-1]`
+for all `sk` in `strides` for all `jk` in `start_indices[k]` $\leq$ `jk` $\lt$
+`limit_indices[k]` where 0 $\leq$ `k` $\lt$ rank(`operand`).
+
+Shape is calculated by `dim(result, i)` = $\lceil$`(limit_indices[i]-start_indices[i])/stride[i]`$\rceil$
```
sdasgup3, line 1258: Let us put in in the Constraints.

docs/spec_draft.md
```diff
@@ -1243,6 +1244,78 @@ Numeric precision is implementation-defined.
...
+### Results
+
+| Name     | Type                          |
+|----------|-------------------------------|
+| `result` | tensor of any supported types |
+
+### Constraints
+
+  * (C1) `operand` and `result` have the same element type.
+  * (C2) `start_indices`, `limit_indices`, and `strides` have the same type.
```
sdasgup3, line 1279: C3 can be derived from C2

docs/spec_draft.md
```diff
@@ -1243,6 +1244,78 @@ Numeric precision is implementation-defined.
...
+  * (C3) size(`start_indices`) = size(`limit_indices`) = size(`strides`) =
+  rank(`operand`).
+  * (C4) 0 $\le$ `start_indices[d]` $\le$ `limit_indices[d]` $\le$
+  size(`operand[d]`) for all dimension `d`.
+  * (C5) 0 $\lt$ `strides[d]` for all dimension `d`.
+
```
sdasgup3, line 1285: (C6) `operand` and `result` have the same rank.

docs/spec_draft.md
```diff
@@ -1243,6 +1244,78 @@ Numeric precision is implementation-defined.
...
+|----------|-------------------------------|
+| `result` | tensor of any supported types |
+
+### Constraints
+
+  * (C1) `operand` and `result` have the same element type.
+  * (C2) `start_indices`, `limit_indices`, and `strides` have the same type.
+  * (C3) size(`start_indices`) = size(`limit_indices`) = size(`strides`) =
+  rank(`operand`).
+  * (C4) 0 $\le$ `start_indices[d]` $\le$ `limit_indices[d]` $\le$
```
sdasgup3, line 1282: `limit_indices` are exclusive. So I guess when the start_indices == limit_indices, we might get a zero value as the dimension size.

sdasgup3:
I will add a few more comments on the semantics a bit later.

docs/spec_draft.md
```diff
@@ -1243,6 +1244,78 @@ Numeric precision is implementation-defined.
...
+  size(`operand[d]`) for all dimension `d`.
+  * (C5) 0 $\lt$ `strides[d]` for all dimension `d`.
```
sdasgup3, line 1284: can a stride be == 0

docs/spec_draft.md
```diff
@@ -1243,6 +1244,78 @@ Numeric precision is implementation-defined.
...
+
+Extracts a sub-tensor from the `operand` using the dimension indices from the
+operands.
+
+More formally, `result[i0, ..., ik, ..., iR-1] = operand[j0+s0, ..., jk+sk, ..., jR-1+sR-1]`
+for all `sk` in `strides` for all `jk` in `start_indices[k]` $\leq$ `jk` $\lt$
+`limit_indices[k]` where 0 $\leq$ `k` $\lt$ rank(`operand`).
+
+Shape is calculated by `dim(result, i)` = $\lceil$`(limit_indices[i]-start_indices[i])/stride[i]`$\rceil$
+for `i`th dimension size of `operand`.
```
sdasgup3, line 1259: of `result`?

docs/spec_draft.md
ghpvnist: What about the ```... = rank(`operand`)``` part? I wasn't sure how to merge (C2) and (C3) into one without making it seem to have multiple constraints in one line.

docs/spec_draft.md
ghpvnist: Yes, you are right, and this is allowed.

docs/spec_draft.md
ghpvnist: No, stride must be at least 1, otherwise there is nothing to "pick" from the `operand`.

docs/spec_draft.md
sdasgup3: What I meant was:  `start_indices`, `limit_indices`, and `strides` have the same type => have the same shape which for a 1-D tensor is its size.

ghpvnist: This constraint was more so meant to make sure that the rank of the `operand` is the same size as the shape/size of `start_indices`, etc. as opposed to making sure the type/shape of the three were the same. This constraint is needed since we could have arbitrarily large number of indices when the rank of `operand` is small. Maybe:

`start_indices`, `limit_indices`, and `strides` have the same type and their size is equal to rank of `operand`.

docs/spec_draft.md
ghpvnist: Sorry, it should say:
```
`i`th dimension of `operand`.
```
But since `operand` and `result` have the same rank, both are interchangeable. I used `operand` here since we are indexing using the `operand` to get the `result`.

docs/spec_draft.md
sdasgup3: I see.
Also stride is the one which help to step over the dimension. With 0 value there is nothing to step over beyond the start index value; kind of  an infinite loop.

Resolving this.

docs/spec_draft.md
sdasgup3: You are right.
With zero dimension size, in any dimension, the number of elements a tensor can store is zero: will be an empty tensor.

docs/spec_draft.md
sdasgup3: Ah! I see you point. I missed the part $==$ rank(`operand`).

The suggestion makes sense.

sdasgup3:
Here is my suggestion for writing the semantics

```
result[i0, ..., ik, ..., iR-1] = 
operand[i0 + start_indices[0]+stride[0], ..., ik + start_indices[k] + stride[k], ..., i(R-1) + start_indics[R-1] + stride(R-1)]


where stride[k] == stride_indices[k] if ik != 0
      else stride[k] = 0;
```

docs/spec_draft.md
sdasgup3: I  see you point. You are right.

My suggestion was because we are defining the dimension size at ith axis for result, thus to me it makes more sense to "apply" `i` in the context of result.

docs/spec_draft.md
ghpvnist: Correct, it would be an empty tensor. Though I think during implementation, we may do something along the lines of:
```
while (start_indices[i] < limit_indices[i] ...) {}
```
Since the indices must be less than or equal to size(`operand[i]`) (C4), the loop will terminate since `start_indices[i] == limit_indices[i] == 0` in which case there is no need to handle special cases of 0-dimension.

This is speculation though, so things might change once we actually start implementing this op.

Did you want me to add anything specific to the Semantics section to clarify empty tensors?

docs/spec_draft.md
```diff
@@ -1243,6 +1244,78 @@ Numeric precision is implementation-defined.
...
 [Back to Ops](#index-of-ops)
 
+## stablehlo.slice
+
+### Semantics
+
+Extracts a sub-tensor from the `operand` using the dimension indices from the
+operands.
+
+More formally, `result[i0, ..., ik, ..., iR-1] = operand[j0+s0, ..., jk+sk, ..., jR-1+sR-1]`
```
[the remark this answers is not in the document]
ghpvnist: Thanks! I've been wondering how to deal with the first element since stride does not apply to it.

docs/spec_draft.md
sdasgup3: Nice option!

>> Did you want me to add anything specific to the Semantics section to clarify empty tensors?
We do not have to write anything specific for 0 dimension size.

docs/spec_draft.md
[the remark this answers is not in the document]
burmako: I don't think the proposed formula is correct. Stride means skipping X elements at a time, not shifting everything by X elements.  Here's how the correct formula could look like:

```
More formally, `result[i0, ..., iR-1] = operand[j0, ..., jR-1]` where
`jd = start_indices[d] + id * strides[d]`.
```

A few additional notes about this formula in comparison with the wording in the pull request and proposals:
  * Instead of `i0, ..., ik, ..., iR-1`, I think we can say `i0, ..., iR-1`. All dimensions are treated the same by this op, so there's no need to call out `ik`.
  * I think that "where `jd = ...`" is easier to reach than when the `jd` formula is replicated inline.
  * Stuff like `jd <= limit_indices[d]` and `d < R` follows from constraints, so I don't think we should mention it here.
  * I think that using `d` here is a better fit than `k` because constraints use `d`.

Overall, after we're done with Data Movement and Reduction ops, we should review the notation of all ops and make sure it's consistent: #289. I don't think we need to do this now because it is likely that future ops will uncover further notions and potential inconsistencies.

docs/spec_draft.md
burmako: I think it's the other way around - C2 can be derived from C3 :)

docs/spec_draft.md
```diff
@@ -1243,6 +1244,78 @@ Numeric precision is implementation-defined.
...
+| `result` | tensor of any supported types |
+
+### Constraints
+
+  * (C1) `operand` and `result` have the same element type.
+  * (C2) `start_indices`, `limit_indices`, and `strides` have the same type.
+  * (C3) size(`start_indices`) = size(`limit_indices`) = size(`strides`) =
+  rank(`operand`).
+  * (C4) 0 $\le$ `start_indices[d]` $\le$ `limit_indices[d]` $\le$
+  size(`operand[d]`) for all dimension `d`.
```
burmako, line 1283: This is called `dim(operand, d)` in the spec for `stablehlo.pad`.

docs/spec_draft.md
burmako, line 1259: To be consistent with other constraints, how about "for all dimension `d`"?

stablehlo/dialect/StablehloOps.td
```diff
@@ -1471,6 +1471,24 @@ def StableHLO_SliceOp: StableHLO_Op<
...
+    Extracts a sub-tensor from the `operand` using the dimension indices from
+    the operands.
+
+    See:
+    https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md#stablehloslice
+
+    Example:
+    ```mlir
+    %0 = "stablehlo.slice" (%arg0) {
+      start_indices = dense<2> : tensor<i64>,
```
burmako, line 1485: This should be `tensor<1xi64>` and same for other attributes below.

docs/spec_draft.md
```diff
@@ -1243,6 +1244,78 @@ Numeric precision is implementation-defined.
...
+  * (C5) 0 $\lt$ `strides[d]` for all dimension `d`.
+
+### Examples
+
+```mlir
+// 1-dimensional slice
+
+// %operand: [0, 1, 2, 3, 4]
+%result = "stablehlo.slice" (%operand) {
+  start_indices = dense<2> : tensor<i64>,
```
burmako, line 1293: This should be `tensor<1xi64>` and same for other attributes below.

docs/spec_draft.md
```diff
@@ -1243,6 +1244,78 @@ Numeric precision is implementation-defined.
...
+  size(`operand[d]`) for all dimension `d`.
+  * (C5) 0 $\lt$ `strides[d]` for all dimension `d`.
+
+### Examples
+
+```mlir
+// 1-dimensional slice
+
+// %operand: [0, 1, 2, 3, 4]
+%result = "stablehlo.slice" (%operand) {
```
burmako, line 1292: Extraneous whitespace between the mnemonic and the open parenthesis. (Same below).

docs/spec_draft.md
burmako: "Though I think during implementation, we may do something along the lines of `while (start_indices[i] < limit_indices[i] ...) {}`". I don't think any explicit comparisons will be needed in the implementation of the interpreter, because the only place where these comparisons should be done is type inference (or constraints in the spec terminology). Also cf. my comment about the proposed formula for the semantics sections.

burmako (changes requested):

docs/spec_draft.md
ghpvnist: You're right, I was viewing `start_indices` as the current index. My intent was to initialize `curr_index` as `start_indices` and iterating `curr_index` to select elements until it breaks the condition: `while (curr_index < limit_indices)`.

stablehlo/dialect/StablehloOps.td
```diff
@@ -1471,6 +1471,24 @@ def StableHLO_SliceOp: StableHLO_Op<
       [NoSideEffect, SameOperandsAndResultElementType,
        AllTypesMatch<["start_indices", "limit_indices", "strides"]>,
        DeclareOpInterfaceMethods<InferTypeOpInterface>]> {
+  let summary = "Iota operator";
```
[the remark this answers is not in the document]
ghpvnist: Thanks for the catch!

stablehlo/dialect/StablehloOps.td
ghpvnist: You're right. Thanks.

docs/spec_draft.md
burmako: I don't think that we need even that. Because our formula expresses the op as a transformation of index spaces, we can say something like (the code below uses operator overloads which don't exist at the moment):

```
Tensor eval(SliceOp op, const Tensor &operand) {
  Tensor result(op.getType());
  for (auto resultIt = result.index_begin(); resultIt != result.index_end(); ++resultIt) {
    auto operandIndex = op.getStartIndices() + *resultIt * op.getStrides();
    result.set(*resultIt, operand.get(operandIndex));
  }
  return result;
}
```

Note how `curr_index < limit_indices` never comes up in this loop. That's because we know that `op.getType()` is correct, so we know that `operandIndex` will always be in bounds. I think this shows the power of decoupling constraints from execution.

burmako (changes requested):

docs/spec_draft.md
```diff
@@ -1259,6 +1260,79 @@ Numeric precision is implementation-defined.
...
+### Results
+
+| Name     | Type                          |
+|----------|-------------------------------|
+| `result` | tensor of any supported types |
+
+### Constraints
+
+  * (C1) `operand` and `result` have the same element type.
+  * (C2) `operand` and `result` have the same rank.
```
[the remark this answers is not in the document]
burmako: I see this comment was resolved but the code hasn't changed?

burmako (changes requested):
I think there's still one remaining unresolved comment.

burmako (approved):

PR #282, opened 2022-10-08
Update CHLO to use prefixed accessors

burmako:
Follows up on #203 to complete the migration of the StableHLO repository to kEmitAccessorPrefix_Prefixed.


GleasonK (approved):
@sdasgup3 this will likely cause some build failures when integrated into `mlir-hlo`.

PR #283, opened 2022-10-08
Improve Python bindings for CHLO and StableHLO

burmako:
1) Fixes chlo.py and stablehlo.py to properly export constructors for attributes and types.

2) Aligns Python accessor names with C++ accessor names. E.g. renames `ComparisonDirectionAttr.comparison_direction` to `value`. We don't promise source compatibility for Python bindings at the moment, so no compatibility shims are introduced.

3) Adds exhaustive tests for ChloModule.cpp and StablehloModule.cpp.

(As promised in the code review which replaced MLIR-HLO's CHLO with StableHLO's CHLO).

### DIFF
--- a/stablehlo/integrations/c/ChloAttributes.cpp
+++ b/stablehlo/integrations/c/ChloAttributes.cpp
@@ -22,20 +22,19 @@ limitations under the License.
 //===----------------------------------------------------------------------===//
 
 MlirAttribute chloComparisonDirectionAttrGet(MlirContext ctx,
-                                             MlirStringRef direction) {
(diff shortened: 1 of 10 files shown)


sdasgup3:
Should we tag this with MHLO so as to do simialr improvements: https://github.com/tensorflow/tensorflow/blob/3c46a67aaa938c643dc0f597e3f9f786fb4aa8f5/tensorflow/compiler/xla/mlir_hlo/include/mlir-hlo-c/Attributes.h#L210

sdasgup3 (approved):

burmako:
That's a good point! Tagged with "MHLO".

Issue #285, opened 2022-10-09
update interpreter of add, multiply, maximum, and minimum w.r.t supporting bool type


sdasgup3:
### Request description

Originated from https://github.com/openxla/stablehlo/pull/264#pullrequestreview-1131986897.

The spec of these ops are updated as part of https://github.com/openxla/stablehlo/pull/264. The issue is to update the interpreter.

burmako:
Let's also include `multiply` here, since it's about to land.

sdasgup3:
The multiply op will be taken care of by https://github.com/openxla/stablehlo/pull/240

PR #117, opened 2022-09-14
Avoid using DenseElementsAttr::getRawData

sdasgup3:
fixes https://github.com/openxla/stablehlo/issues/83

The PR implements  the underlying data storage of the tensor using an `mlir::HeapAsmResourceBlob` object represents a mutable blob of data. This provides two improvements:
1. No need for using `DenseElementsAttr::getRawData()` which cannot be relied  to provide contiguous representation of underlying data in cases like boolean and splats.
2.  An `mlir::HeapAsmResourceBlob` object, being mutable, simplifies how we store the underlying Tensor data.

PR #286, opened 2022-10-09
Add interpreter and tests for bitwise ops

sdasgup3:


stablehlo/reference/Element.cpp
```diff
@@ -187,6 +187,28 @@ Element Element::operator-(const Element &other) const {
...
+Element Element::operator^(const Element &other) const {
+  auto left = this->getIntegerValue();
+  auto right = other.getIntegerValue();
+  return Element(getType(), left ^ right);
+}
+
+Element Element::operator~() const {
+  return Element(getType(), ~this->getIntegerValue());
+}
+
```
burmako, line 211: Let's order alphabetically here and below:

```
$ python3
>>> sorted(['+', '-', '&', '|', '^', '~'])
['&', '+', '-', '^', '|', '~']
```

stablehlo/reference/Types.cpp
```diff
@@ -36,8 +36,10 @@ bool isSupportedSignedIntegerType(Type type) {
          type.isSignlessInteger(64);
 }
 
+bool isSupportedBooleanType(Type type) { return type.isSignlessInteger(1); }
+
 bool isSupportedIntegerType(Type type) {
-  return isSupportedUnsignedIntegerType(type) ||
+  return isSupportedBooleanType(type) || isSupportedUnsignedIntegerType(type) ||
```
burmako, line 42: I don't think we should be treating booleans as integers. 

On the plus side, this allows us to avoid changing `Element.h` from `std::variant<APInt, APFloat, std::pair<APFloat, APFloat>>` to `std::variant<bool, APInt, APFloat, std::pair<APFloat, APFloat>>` and updating `map` in `Element.cpp`.

On the minus side, this hides logical errors. E.g. a bunch of existing code e.g. overloaded operators in `Element` will silently accept booleans, and some of that code, e.g. `Element::operator-` will be silently incorrect.

Given that, I recommend that we model booleans as a separate family of types and as a separate variant in `Element.h`. Existing operators in `Element` can either be updated in this PR, or can `report_fatal_error` for now and wait until #285.

burmako (changes requested):

stablehlo/reference/Types.cpp
sdasgup3: done.

stablehlo/reference/Tensor.cpp
```diff
@@ -177,6 +178,12 @@ Element Tensor::get(ArrayRef<int64_t> index) const {
     }
   }
 
+  // Handle boolean type.
+  if (isSupportedBooleanType(elementType)) {
+    auto elementData = reinterpret_cast<const uint8_t *>(elementPtr);
+    return Element(elementType, *elementData == 0 ? false : true);
```
ghpvnist, line 184: Can be simplified to:
`return Element(elementType, *elementData? true : false);`

ghpvnist (approved):
Minor changes and lgtm.

ghpvnist:
Another note, I know we had a bug related to #117 for splats, but do we need to explicitly test for them? Though we are not implicitly testing for splat values, imo this is a similar case to not testing overflows in AddOp. What are your thoughts?

stablehlo/reference/Element.cpp
```diff
@@ -134,6 +147,15 @@ APInt Element::getIntegerValue() const {
   return std::get<APInt>(value_);
 }
 
+bool Element::getBooleanValue() const {
+  if (!isSupportedBooleanType(type_))
+    report_fatal_error(StringRef("Accessing value of a type different than "
+                                 "what is stored in Element object") +
+                       LLVM_PRETTY_FUNCTION);
```
burmako, line 154: I don't think that `LLVM_PRETTY_FUNCTION` is pulling its weight in the interpreter (not just in this PR but everywhere else too). Instead of a oneliner `report_fatal_error("foo")`, we get multiple lines of boilerplate, and it doesn't look like it's worth the cost.

If the idea is to make error messages easily traceable back to the corresponding functions, then we can simply make them unique, right? E.g. "Element is not a boolean" instead of "Accessing value of a type different than what is stored in Element object".

Same applies for other uses of `LLVM_PRETTY_FUNCTION`. E.g. "float & float is unsupported" instead of "Unsupported floating-point type", etc.

docs/status.md
```diff
@@ -35,7 +35,7 @@ one of the following tracking labels.
 | all_gather               |      no       |     yes*     |        no         |        no         |     no      |
```
burmako, line 35: Let's downgrade `add`, `maximum` and `minimum`  in `status.md` to `Interpreter: no`?

stablehlo/reference/Tensor.cpp
```diff
@@ -157,7 +157,8 @@ Element Tensor::get(ArrayRef<int64_t> index) const {
       auto elementData = reinterpret_cast<const int64_t *>(elementPtr);
       return Element(elementType, APInt(intTy.getWidth(), *elementData,
                                         intTy.isSignedInteger()));
-    } else if (elementType.isUnsignedInteger(4) ||
+    } else if (elementType.isSignlessInteger(1) ||
```
burmako, line 160: `isSignlessInteger(1)` is no longer needed here.

stablehlo/reference/Tensor.cpp
burmako: Let's be really picky about the underlying representation to avoid bool <=> int mishaps:

```
if (*elementData == 0) return Element(false);
if (*elementData == 1) return Element(true);
report_fatal_error("Unsupported boolean value");
```

stablehlo/reference/Tensor.cpp
```diff
@@ -410,8 +425,9 @@ Tensor makeTensor(DenseElementsAttr attr) {
                   HeapAsmResourceBlob::allocateAndCopy<int64_t>(intValues));
   }
 
-  // Handle unsigned integer types.
-  if (elemType.isUnsignedInteger(4) || elemType.isUnsignedInteger(8)) {
+  // Handle boolean & unsigned integer types.
+  if (elemType.isSignlessInteger(1) || elemType.isUnsignedInteger(4) ||
```
burmako, line 429: `isSignlessInteger(1)` seems to be no longer needed here either.

burmako (changes requested):

sdasgup3:
> Another note, I know we had a bug related to #117 for splats, but do we need to explicitly test for them? Though we are not implicitly testing for splat values, imo this is a similar case to not testing overflows in AddOp. What are your thoughts?

We have tests for testing the behavior of various ops. And then we can have tests to test the interpreter itself. IMO, the reason we did not test the corner cases for AddOp is mainly because they are anyway implementation defined.  So does not mater what output we check in the lit checks, some other implementation can well diverge: Hence no take away from those tests. 

Regarding testing for splats, it would be ideal if we  can write tests serving both the purposes: checking a behavior which needs splats as a tensor value. I hope in future we could get those tests. 

For now I added a few explicit tests https://github.com/openxla/stablehlo/pull/286/files#diff-bfe3f3f1eb10f8a56d681bf5d69e08696e30f989e2e2fae190ecc3bb2a3ad7bcR158

But I could just remove them, if we get those cases implicitly.

burmako (approved):

ghpvnist (approved):

PR #292, opened 2022-10-10
Handle bounds in the Transpose op shape function

smit-hinsu:
This makes it possible to propagate the input bounds while creating the Transpose op.

### DIFF
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -4138,7 +4138,7 @@ LogicalResult TransposeOp::reifyReturnTypeShapes(
 // Method for InferTypeOpInterface: infer the return type from the operand type
 // and the permutation.
 LogicalResult TransposeOp::inferReturnTypes(
-    MLIRContext* /*context*/, Optional<Location> loc, ValueRange operands,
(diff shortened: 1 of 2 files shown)


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -4168,13 +4168,31 @@ LogicalResult TransposeOp::inferReturnTypes(
                              " of [",
                              range, "] but got ", permutation);
 
+  llvm::SmallVector<int64_t, 4> inputBounds(rank, ShapedType::kDynamicSize);
+  bool hasBounds = false;
+  if (auto encoding =
+          rankedTy.getEncoding().dyn_cast_or_null<TypeExtensionsAttr>()) {
+    inputBounds = llvm::to_vector<4>(encoding.getBounds());
+    hasBounds = true;
+  }
```
zhouxin913, line 4177: It is possible that we put this part as an util function in the future. It looks good to check as-is now, as the code-sharing refactoring is on-going and this can be refined later if more ops will have some parts like this.

zhouxin913 (approved):

stablehlo/dialect/StablehloOps.cpp
smit-hinsu: sounds good.

PR #303, opened 2022-10-10
Update inference test to use stablehlo dialect instead of mhlo

smit-hinsu:
This seems to have escaped update during bootstrap and therefore it doesn't do the intended testing.


smit-hinsu:
Sent https://github.com/openxla/stablehlo/pull/304 instead as this as spurious commits.

PR #304, opened 2022-10-10
Update inference test to use stablehlo dialect instead of mhlo

smit-hinsu:
This prevents reduce and reduce_window shape functions' testing.


stablehlo/tests/infer_stablehlo.mlir
```diff
@@ -471,16 +471,16 @@ func.func @complex_sparsity(%arg0: tensor<10x10xf32, #CSR>, %arg1: tensor<10x10x
...
-    "mhlo.return"(%1) : (tensor<4xf32>) -> ()
+    %1 = "stablehlo.add"(%arg2, %arg3) : (tensor<4xf32>, tensor<4xf32>) -> tensor<4xf32>
+    "stablehlo.return"(%1) : (tensor<4xf32>) -> ()
 
   }) {dimensions = dense<[0]> : tensor<1xi64>} : (tensor<4x4xf32>, tensor<4xf32>) -> tensor<4xf32>
-  %2 = "mhlo_test.get_return_type_components"(%0)
+  %2 = "hlo_test_infer.get_return_type_components"(%0)
       : (tensor<4xf32>) -> tensor<4xindex>
-// CHECK: %1 = "mhlo_test.get_return_type_components"(%0) : (tensor<4xf32>) -> tensor<4xindex>
+// CHECK: %1 = "hlo_test_infer.get_return_type_components"(%0) : (tensor<4xf32>) -> tensor<4xindex>
```
burmako, line 483: Seems like the check needs to be updated to `hlo_test_infer.return_types`. This is a discrepancy from how `mhlo_test` does things, and I think this is how this went unnoticed in code review.

burmako (changes requested):

stablehlo/tests/infer_stablehlo.mlir
zhouxin913: I'm surprised that this even worked. The `mhlo_test` was introduced with oversight in https://github.com/openxla/stablehlo/pull/50. Let me think a little bit of this and reply later, as I believe this is not a blocking PR for our other tasks.

burmako: This worked because: 1) this file allows unregistered dialects, 2) because the CHECK mistakenly checks that the output IR is the same as input IR. 1 made sure that the test doesn't crash right away, 2 made sure that the test passes.

smit-hinsu: Actually, I mistakenly thought that this was introduced in the bootstrap but this was a later addition. so, this will likely require an actual CHECK to verify the shapes.

regarding return_types vs get_return_type_components, I think we use return_types for types with sparsity or bounds and get_return_type_components for other types.

zhouxin913: @burmako I see thanks!

@smit-hinsu "use return_types for types with sparsity or bounds and get_return_type_components for other types" 
Yes. Let's use "hlo_test_infer.get_return_type_components" (your current code) for this PR. And use `get_return_type` in the future when we add bounds for this op.
Thanks for catching this "mhlo" error! Approved.

zhouxin913 (approved):

burmako (approved):

stablehlo/tests/infer_stablehlo.mlir
burmako: @smit-hinsu It looks like tests are failing.

smit-hinsu: yes, was waiting on the clarification in updating the CHECK statements. fixed the tests.

Talking with Xin if we can disable allow unregistered ops and also having consistent style of CHECK statements. We will do that separately.

Issue #289, opened 2022-10-09
Tighten the formal notation that we're using in the spec


burmako:
After all Data Movement and Reduction ops are specced, we should review and tighten our formal notation. Some stuff that has come up in the recent pull requests:
  * Make sure we define all metasyntactic variables and functions in advance, like `i`, `j`, `d`, `k`, `R`, `dim(...)`, `rank(...)`, `size(...)`, etc.
  * Consider using Latex syntax like $i_{d}$ instead of `id`.
  * Settle on using $\le$ vs `<=` (the former is not copypasteable).
  * Consistent syntax for stuff like "for all foo in bar" and "for any...".
  * When to use `[i0, ..., ik, ..., iR-1]` vs `[i0, ..., iR-1]` (or maybe even `[i0, ...]`).
  * "size(`operand`)" vs "`size(operand)`"

sdasgup3:
For reverse op, we use "All dimensions in `dimensions` are unique."

In this context, we means that all the entries of `dimensions` (which logically specify dimensions) are unique.  IMO, this is a bit confusing with

```
For all dimensions `i` in `operand`
```
where we straightforwardly mean the dimensions of `operand`

Can we simplify the above by saying 

The elements of `dimensions` are unique.

sdasgup3:
We might need to define the following items which are currently in use. 

* dim(x: tensor, y:int) as the dimension size of yth dimension of x
* size(x: 1-D tensor) = dim(x, 0)

ghpvnist:
Unify the use of:
* 0 vs $0$
* = vs $=$ vs `=`

ghpvnist:
Unrelated to formal notations, but I figured it's better to address here while going through all the notations:
We currently use new lines to express dimensions in the examples, but sometimes we do not introduce new lines.
Decide at which nth dimension we should not add new lines. For example:
```
[
    [[1, 2], [3, 4]],
    [[5, 6], [7, 8]]
]
```
vs
```
[
    [[1, 2],
     [3, 4]],
    [[5, 6],
     [7, 8]]
]
```
vs
```
[
    [
        [1, 2],
        [3, 4]
    ],
    [
        [5, 6],
        [7, 8]
    ]
]
```

burmako:
How do we order constraints? Following the order in which underlying inputs/outputs are defined in ODS? What about constraints for the same inputs? Which ones go first?

burmako:
[...]

PR #307, opened 2022-10-11
Add spec for DotGeneralOp

### LINKED ISSUE #299: Add spec for DotGeneralOp
### Request description

To spec the stablehlo.dot_general

ghpvnist:
Regarding `status.md`:
* Verifier is set to `revisit` since the verifier and the XLA semantics do not account for `precision_config`.

closes #299


ghpvnist:
Some questions:
* I'm not sure whether the constraint that batching dimensions need to be less than contracting dimensions since this is allowed in XLA currently.
* Also not exactly sure what the `precision_config` does.
* Since verifier for `precision_config` is not present, we can have arbitrary number of precision_config values. How many precisions can the op specify?
* Currently, I've assigned type inference as `no` due to missing `precision_config` checks since it may affect type inference.

docs/spec_draft.md
```diff
@@ -543,6 +544,124 @@ produces an implementation-defined value.
...
+### Results
+
+| Name     | Type                                                       |
+|----------|------------------------------------------------------------|
+| `result` | tensor of integer, floating-point or complex element types |
+
+### Constraints
+
+  * (C1) `lhs`, `rhs` and `result` have the same element type.
+  * (C2) `lhs` and `rhs` have the same number of `batching_dimensions`,
```
sdasgup3, line 572: size(`lhs_batching_dimensions`) $=$ size(`rhs_batching_dimensions`)  and size(`lhs_contracting_dimensions`) $=$ size(`rhs_contracting_dimensions`)

docs/spec_draft.md
```diff
@@ -543,6 +544,124 @@ produces an implementation-defined value.
...
+  `contracting_dimensions`.
+  * (C3) `lhs_batching_dimensions` and `lhs_contracting_dimensions` values are
```
sdasgup3, line 574: Just like reverse op
All dimensions in `lhs_batching_dimensions`, `lhs_contracting_dimensions` are unique

docs/spec_draft.md
```diff
@@ -543,6 +544,124 @@ produces an implementation-defined value.
...
+  unique.
+  * (C4) `rhs_batching_dimensions` and `rhs_contracting_dimensions` values are
+  unique.
+  * (C5) `dim(lhs_batching_dimensions, i)`, `dim(rhs_batching_dimensions, i)`
```
sdasgup3, line 578: Can you elaborate what are we checking here?

`dim(lhs_batching_dimensions, i)` < `dim(lhs_contracting_dimensions, j)`

From the look it seems: The dimension size of lhs_batching_dimensions need to be less that the dimension sizes of contracting dimensions

docs/spec_draft.md
```diff
@@ -543,6 +544,124 @@ produces an implementation-defined value.
...
+  $\lt$ `dim(lhs_contracting_dimensions, j)`,
+  `dim(rhs_contracting_dimensions, j)` for all `i` and for all `j`.
+  * (C6) 0 $\le$ `lhs_batching_dimensions[d]`, `lhs_contracting_dimensions[d]`
```
sdasgup3, line 581: For C6-9:

 lhs_batching_dimensions is a 1-D tensor having one dimension (or axis) `0`, and so when we are saying  rhs_batching_dimensions[d] for all dimensions `d` it be ambiguous. How about 

0<= rhs_batching_dimensions[i] < rank(`lhs`) for all i $\in$ [0, size(`rhs_batching_dimensions`))

from https://github.com/openxla/stablehlo/issues/289#issuecomment-1273993605, we have plans to include size.

docs/spec_draft.md
```diff
@@ -543,6 +544,124 @@ produces an implementation-defined value.
...
+  $\lt$ rank(`lhs`).
+  * (C7) 0 $\le$ `rhs_batching_dimensions[d]`, `rhs_contracting_dimensions[d]`
+  $\lt$ rank(`rhs`) for all dimensions `d`.
+  * (C8) `dim(lhs, lhs_batching_dimensions[i])` =
```
sdasgup3, line 585: same here 

From C2 we already established that sizes of lhs and rhs batching dimensions are the same. 
`dim(lhs, lhs_batching_dimensions[i])` = `dim(rhs, rhs_batching_dimensions[i])`  for all  i $\in$ [0, size(`rhs_batching_dimensions`))

docs/spec_draft.md
```diff
@@ -543,6 +544,124 @@ produces an implementation-defined value.
...
+  `dim(rhs, rhs_batching_dimensions[i])` for all dimensions `i`.
+  * (C9) `dim(lhs, lhs_contracting_dimensions[i])` =
+  `dim(rhs, rhs_contracting_dimensions[i])` for all dimensions `i`.
+  * (C10) `dim(result, i)` is equal to (in order):
+    * `lhs_batching_dimensions[i]` for all 0 $\le$ `i` $\lt$
+    rank(`lhs_batching_dimensions`).
```
sdasgup3, line 591: Did you mean size? rank(lhs_batching_dimensions) is 1

sdasgup3:
>> the check that batching dimension numbers need to be less than contracting dimension numbers is not present in the verifier, but is in sync with XLA semantics

Did you mean the batching dimensions need to appear before the contracting dimensions?

sdasgup3:
>> Also not exactly sure what the precision_config does.

[This](https://github.com/google/jax/blob/9bb2c999d6b7eb574c23209a2ec543abad891f7f/jax/_src/lax/lax.py#L658) gives a good explanation.

stablehlo/dialect/StablehloOps.td
```diff
@@ -2048,9 +2048,24 @@ def StableHLO_DotGeneralOp: StableHLO_ShapedInterfaceOp<"dot_general",
...
+    matrix/matrix multiplication over dimensions specified in
+    `dot_dimension_numbers`.
 
-    See https://www.tensorflow.org/xla/operation_semantics#dotgeneral.
+    See:
+    https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md#stablehlodot_general
+
+    Example:
+    ```mlir
+    %0 = "stablehlo.dot_general"(%arg0, %arg1) {
```
burmako, line 2059: Let's use `%result`, `%lhs` and `%rhs` to be consistent with the Examples section above?

burmako (changes requested):
Haven't reviewed Constraints yet, because I understand that both Semantics and Constraints are WIP. Please re-request review when you get a chance to push the update to Semantics.

docs/spec_draft.md
```diff
@@ -543,6 +544,124 @@ produces an implementation-defined value.
...
+
+### Results
+
+| Name     | Type                                                       |
+|----------|------------------------------------------------------------|
+| `result` | tensor of integer, floating-point or complex element types |
+
+### Constraints
+
+  * (C1) `lhs`, `rhs` and `result` have the same element type.
```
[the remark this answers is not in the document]
sdasgup3: IMO, for StableHLO DotGeneralOp, we can consider the preferred element type to be same as the output element type.

According https://github.com/tensorflow/tensorflow/blob/9315da3fe49a8b36eadcc7e539b8692012320303/tensorflow/compiler/xla/service/shape_inference.cc#L768

should we write that the result type needs to be wider than any element type of the operand only for non floating-point type.

stablehlo/dialect/StablehloOps.td
ghpvnist: Sure. I had them named as such since the other ops in the file is following this format. I'll make another PR renaming the examples to match the spec.

docs/spec_draft.md
ghpvnist: This constraint should be removed.

docs/spec_draft.md
ghpvnist: Thanks for the catch!

docs/spec_draft.md
```diff
@@ -543,6 +544,124 @@ produces an implementation-defined value.
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.dot_general
+
+### Semantics
+
+Performs general dot products between vectors, vector/matrix and matrix/matrix
```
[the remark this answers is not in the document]
ghpvnist: Thanks, I've updated the description. Let me know if this explains better.

docs/spec_draft.md
```diff
@@ -543,6 +544,124 @@ produces an implementation-defined value.
...
+multiplication over dimensions specified in `dot_dimension_numbers`.
+
+### Operands
+
+| Name                    | Type                                                                                                                                                                                                 |
+|-------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `lhs`                   | tensor of integer, floating-point or complex element types                                                                                                                                           |
```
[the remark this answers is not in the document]
ghpvnist: Thanks, I've updated the PR. The booleans are treated as 1s and 0s for the purpose of the dot product, which makes sense.

docs/spec_draft.md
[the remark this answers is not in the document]
ghpvnist: I've updated the comment regarding some doubts I had, and it makes sense to use the result element type as the `preferred_element_type`, and thanks for catching this. XLA doc does not mention the use of `preferred_element_type` either. I'm wondering if this was left out deliberately since this goes into how the hardware computes it.

ghpvnist:
Update:

I spent a couple of days last week figuring out the execution semantics of this op and implemented the interpreter for DotGeneral #337 as a result. I spent majority of the time verifying whether the constraints I put were correct as they were different from XLA semantics. It turns out that the constraints are now the same as XLA's and I now have a better understanding of how this op works. 

The PR is now updated with a proposed formal notation for review.

ghpvnist:
Merging the interpreter PR #337 into this PR.

stablehlo/reference/Ops.cpp
```diff
@@ -76,6 +76,97 @@ Tensor eval(CosineOp op, const Tensor &operand) {
...
+  ArrayRef<int64_t> lhsShape = lhs.getType().getShape();
+  ArrayRef<int64_t> lhsBatchingDims =
+      op.getDotDimensionNumbers().getLhsBatchingDimensions();
+  ArrayRef<int64_t> rhsBatchingDims =
+      op.getDotDimensionNumbers().getRhsBatchingDimensions();
+  ArrayRef<int64_t> lhsContractingDims =
+      op.getDotDimensionNumbers().getLhsContractingDimensions();
+  ArrayRef<int64_t> rhsContractingDims =
+      op.getDotDimensionNumbers().getRhsContractingDimensions();
+  std::vector<int64_t> lhsNonContractingDims;
```
sdasgup3, line 90: rename it to lhsNonBatchingNonContractingDims? 
Same with rhsNonContractingDims?

stablehlo/reference/Ops.cpp
```diff
@@ -76,6 +76,97 @@ Tensor eval(CosineOp op, const Tensor &operand) {
...
+      lhsNonContractingDims.push_back(i);
+  std::vector<int64_t> rhsNonContractingDims;
+  for (int64_t i = 0; i < rhs.getType().getRank(); ++i)
+    if (std::find(rhsBatchingDims.begin(), rhsBatchingDims.end(), i) ==
+            rhsBatchingDims.end() &&
+        std::find(rhsContractingDims.begin(), rhsContractingDims.end(), i) ==
+            rhsContractingDims.end())
+      rhsNonContractingDims.push_back(i);
+  std::vector<int64_t> contractingDimSizes;
+  for (uint64_t i = 0; i < lhsContractingDims.size(); ++i)
```
sdasgup3, line 105: fuse the two loops to collect contractingDimSizes and compute   totalContractingSize

stablehlo/reference/Ops.cpp
```diff
@@ -76,6 +76,97 @@ Tensor eval(CosineOp op, const Tensor &operand) {
...
+    contractingDimSizes.push_back(lhsShape[lhsContractingDims[i]]);
+  auto totalContractingSize = 1;
+  for (uint64_t i = 0; i < contractingDimSizes.size(); ++i)
+    totalContractingSize *= contractingDimSizes[i];
+  for (auto resIt = result.index_begin(); resIt != result.index_end();
+       ++resIt) {
+    Type resElTy =
+        op->getResultTypes().front().dyn_cast<ShapedType>().getElementType();
+    Element sum;
```
sdasgup3, line 114: Can we extract the computation to get 0-initialized sum in a lambda `initializeAccumulator`? That will make the code within the current loop more clear..

stablehlo/reference/Ops.cpp
```diff
@@ -76,6 +76,97 @@ Tensor eval(CosineOp op, const Tensor &operand) {
...
+      FloatType flType =
+          resElTy.cast<ComplexType>().getElementType().cast<FloatType>();
+      bool roundingErr;
+      real.convert(flType.getFloatSemantics(), APFloat::rmNearestTiesToEven,
+                   &roundingErr);
+      imag.convert(flType.getFloatSemantics(), APFloat::rmNearestTiesToEven,
+                   &roundingErr);
+      sum = Element(resElTy, std::complex<APFloat>(real, imag));
+    }
+    std::vector<int64_t> lhsIdx(lhs.getType().getRank());
```
sdasgup3, line 141: Can you add some comment here (may be using a simple example) to explain what are we doing below?

sdasgup3: Something like for batching and non-batching-non-constracting dimensions the indices of result, lhs and rhs are the same.

sdasgup3:
> Here is a suggestion for formalism:
> 
> Let batching_dims = [0, 1, .., size(`lhs_batching_dims`)]
> 
> foreach dimension pair (c,c') ∈ (lhs_contracting_dimension, rhs_contracting_dimensions)
> 
> result[...,ip,...]=∑(c,c′)=(0,0)(dim(lhs,c),dim(rhs,c′))lhs(...,jc,...,jq,...)∗rhs(...,kc′,...,kr,...)
> 
> * If p is in `batching dimension`, j[`lhs_batching_dims`[p]] = ip and k[`rhs_batching_dims`[p]] = ip
> * jq = kr = ip Otherwise // For non batching and non contracting


@burmako 
From reviewing implementation and feed back from @ghpvnist, it seems the underlying algorithm formalized here is right. But the way it is formalized might  be confusing and need re-work. wdyt?

ghpvnist:
@sdasgup3 @burmako Here is a proposed reformulation. Wdyt?
Let:
* `batching_dims` = [0, 1, .., size(`lhs_batching_dims`)].
* LCD = `lhs_contracting_dimensions`.
* RCD = `rhs_contracting_dimensions`.

$$ `result`[...,\ ip,\ ...] = \sum_{d=0}^{size(LCD)} \sum_{(c,\ c')=(0,\ 0)}^{(dim(lhs,\ LCD[d]),\ dim(rhs,\ RCD[d]))} lhs(...,\ jc,\ ...,\ jq,\ ...)\ *\ rhs(...,\ kc',\ ...,\ kr,\ ...)$$

* If p is in `batching_dims`, j[`lhs_batching_dims`[p]] = ip and k[`rhs_batching_dims`[p]] = ip.
* jq = kr = ip otherwise (For non batching and non contracting dimensions).

sdasgup3:
> @sdasgup3 @burmako Here is a proposed reformulation. Wdyt? Let:

Looks good. 

After re-thinking it seems that we need to reword the last line with:  `jq = kr = ip otherwise` (same for the earlier proposed formulation https://github.com/openxla/stablehlo/pull/307#issuecomment-1276910301). No need to mention `For non batching and non contracting dimensions` as p in `result` cannot be a contracting dimension.

docs/spec_draft.md
```diff
@@ -663,6 +664,142 @@ produces an implementation-defined value.
...
+
+  * `DEFAULT`: Fastest mode, but least accurate.
+  * `HIGH`: Slower, but more accurate.
+  * `HIGHEST`: Slowest, but most accurate.
+
+### Inputs
+
+| Name                         | Type                                         |
+|------------------------------|----------------------------------------------|
+| `lhs`                        | tensor of any supported types                |
```
sdasgup3, line 700: type? and below

docs/spec_draft.md
```diff
@@ -663,6 +664,142 @@ produces an implementation-defined value.
...
+| `rhs`                        | tensor of any supported types                |
+| `lhs_batching_dimensions`    | 1-dimensional tensor constant of type `si64` |
+| `rhs_batching_dimensions`    | 1-dimensional tensor constant of type `si64` |
+| `lhs_contracting_dimensions` | 1-dimensional tensor constant of type `si64` |
+| `rhs_contracting_dimensions` | 1-dimensional tensor constant of type `si64` |
+| `precision_config`           | constant of type `si64`                      |
```
sdasgup3, line 706: How about 

constant of type `enum {DEFAULT, HIGHEST, HIGH}` ?

@burmako @ghpvnist 
IMO introducing enum does not need introduction. We wanted to convey that precision_config takes special constant values w/o dealing with how the enum case is [actually stored](https://github.com/llvm/llvm-project/blob/1f54641d33d57d85d925e72ffda9c76aef4e9601/mlir/include/mlir/IR/EnumAttr.td#L42).

sdasgup3: From https://github.com/openxla/stablehlo/blob/07578f02f7a6d1477e43a3c8f52db6c5ec33eb68/stablehlo/dialect/StablehloEnums.td#L46 it seems like an optional array, where we can have multiple constraints specified. Do you know what is the semantics if the the array.

In the constraints we can mention that the `precision config` is optional and what the default value is.

docs/spec_draft.md
```diff
@@ -663,6 +664,142 @@ produces an implementation-defined value.
...
+  $\in$ [0, size(`rhs_batching_dimensions`)).
+  * (C8) 0 $\le$ `rhs_contracting_dimensions[d]` $\lt$ rank(`rhs`) for all `i`
+  $\in$ [0, size(`rhs_contracting_dimensions`)).
+  * (C9) `dim(lhs, lhs_batching_dimensions[i])` =
+  `dim(rhs, rhs_batching_dimensions[i])` for all `i` $\in$ [0,
+  size(`lhs_batching_dimensions`)).
+  * (C10) `dim(lhs, lhs_contracting_dimensions[i])` =
+  `dim(rhs, rhs_contracting_dimensions[i])` for all `i` $\in$ [0,
+  size(`lhs_contracting_dimensions`)).
+  * (C11) `dim(result, i)` is equal to (in order):
```
sdasgup3, line 737: The i in `result` is different than  index `i` of lhs and rhs

How about:

The shape of `result` is given by:

{`lhs_batching_dimensions[i]` :  for all 0 $\le$ `i` $\lt$ size(`lhs_batching_dimensions`)} + {`dim(lhs, i)` : for all i $\in$ [0, rank(`lhs`)) where `i` $\notin$ `lhs_batching_dimensions` and `i` $\notin$ `lhs_contracting_dimensions`} + {`dim(rhs, i)` : for all i $\in$ [0, rank(`rhs`)) where `i` $\notin$ `rhs_batching_dimensions` and `i` $\notin$ `rhs_contracting_dimensions`}

ghpvnist: Thanks, this is much cleaner

docs/spec_draft.md
ghpvnist: Fair point. I have no objections. I'll make the changes once everyone is in agreement.

docs/spec_draft.md
```diff
@@ -664,6 +665,143 @@ produces an implementation-defined value.
...
+(0,\ 0)}^{(dim(lhs,\ LCD[d]),\ dim(rhs,\ RCD[d]))}
+lhs(...,\ jc,\ ...,\ jq,\ ...)\ *\ rhs(...,\ kc',\ ...,\ kr,\ ...)$$
+
+* If p is in `batching_dims`, j[`lhs_batching_dims`[p]] = ip and
+k[`rhs_batching_dims`[p]] = ip.
+* jq = kr = ip otherwise.
+
+`precision_config` controls the tradeoff between speed and accuracy for
+computations on accelerator backends. This can be one of the following:
+
```
sdasgup3, line 694: You can add the constraints on precision config following the  changes in 
https://github.com/openxla/stablehlo/pull/441

sdasgup3:
>> Type inference is set to no since precision_config is not accounted for.
missing checks for `precision_config` should only affect the verification, right? The type inference must be `yes`.

docs/spec_draft.md
```diff
@@ -941,6 +942,143 @@ produces an implementation-defined value.
...
+* LCD = `lhs_contracting_dimensions`.
+* RCD = `rhs_contracting_dimensions`.
+
+$$result[...,\ ip,\ ...] = \sum_{d\ =\ 0}^{size(LCD)} \sum_{(c,\ c')\ =\
+(0,\ 0)}^{(dim(lhs,\ LCD[d]),\ dim(rhs,\ RCD[d]))}
+lhs(...,\ jc,\ ...,\ jq,\ ...)\ *\ rhs(...,\ kc',\ ...,\ kr,\ ...)$$
+
+* If p is in `batching_dims`, j[`lhs_batching_dims`[p]] = ip and
+k[`rhs_batching_dims`[p]] = ip.
+* jq = kr = ip otherwise.
```
burmako, line 967: Proposed rewording:

```
Computes dot products between slices of `lhs` and slices of `rhs` and produces
`result`.

More formally, `result[result_index] = dot_product` where
  * `lhs_result_dimensions = [d for d in axes(lhs) and d not in lhs_batching_dimensions and d not in lhs_contracting_dimensions]`.
  * `rhs_result_dimensions = [d for d in axes(rhs) and d not in rhs_batching_dimensions and d not in rhs_contracting_dimensions]`.
  * `result_batching_index + result_lhs_index + result_rhs_index = result_index`
    where `size(result_batching_index) = size(lhs_batching_dimensions)`,
    `size(result_lhs_index) = size(lhs_result_dimensions)` and
    `size(result_rhs_index) = size(rhs_result_dimensions)`.
  * `transposed_lhs = transpose(lhs, lhs_batching_dimensions + lhs_result_dimensions + lhs_contracting_dimensions)`.
  * `transposed_lhs_slice = slice(result_batching_index + result_lhs_index + [:, ...:])`.
  * `reshaped_lhs_slice = reshape(transposed_lhs_slice, dims(lhs, lhs_contracting_dimensions))`.
  * `transposed_rhs = transpose(rhs, rhs_batching_dimensions + rhs_result_dimensions + rhs_contracting_dimensions)`.
  * `transposed_rhs_slice = slice(result_batching_index + result_rhs_index + [:, ...:])`.
  * `reshaped_rhs_slice = reshape(transposed_rhs_slice, dims(rhs, rhs_contracting_dimensions))`.
  * `dot_product = reduce(
    inputs=[multiply(reshaped_lhs_slice, reshaped_rhs_slice)], 
    init_values=[0],
```

Feedback:
  1. Proposed wording reformulates "general dot product" as "dot products between slices of lhs and slices of rhs", which more clearly summarizes the behavior of the op.
  2. In the last few weeks, we've been gravitating towards expressing ops via other ops if possible, and the proposed formalism goes along the same lines.

burmako (changes requested):

docs/spec_draft.md
ghpvnist: Looking back at previous PRs with the new enum phrasing, I'll change this accordingly.

docs/spec_draft.md
```diff
@@ -941,6 +942,143 @@ produces an implementation-defined value.
...
+
+`precision_config` controls the tradeoff between speed and accuracy for
+computations on accelerator backends. This can be one of the following:
+  * `DEFAULT`: Fastest mode, but least accurate (default value).
+  * `HIGH`: Slower, but more accurate.
+  * `HIGHEST`: Slowest, but most accurate.
```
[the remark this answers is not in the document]
ghpvnist: If we avoid `DEFAULT`, what about the enum under Inputs and examples?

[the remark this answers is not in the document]
burmako: Sorry for the confusion. What I meant is to propose removing "(default value)" from the first bullet. I agree that `DEFAULT` needs to stay, at least for now.

stablehlo/reference/Interpreter.cpp
```diff
@@ -87,6 +87,11 @@ llvm::Expected<SmallVector<Tensor>> eval(func::FuncOp func,
       Tensor runtimeOperand = fetchOperand(cosineOp.getOperand());
       Tensor runtimeResult = eval(cosineOp, runtimeOperand);
       populateResults({runtimeResult});
+    } else if (auto dotGeneralOp = dyn_cast<DotGeneralOp>(op)) {
```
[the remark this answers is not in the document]
ghpvnist: Sure, I'll reopen the ticket and migrate the updated interpreter changes out of this PR.

docs/spec_draft.md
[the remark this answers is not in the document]
ghpvnist: Fast vs slow ultimately depends on how much precision you want to have as the final result. If intermediate results are stored using more bits than its original element type (for example, storing f32 element on a f64 or storing f32 elements in two f16, etc.), then depending on how we implement or is implemented in hardware, more precision/calculations are needed to store the intermediate results. Slowest meaning it has the most precise result approximating the original number vs fastest has the least precise of other enum types at the cost of precision.

What about the following? Feel free to suggest something else:
```
  * `DEFAULT`: Fastest calculation, but least accurate approximation to the original number.
  * `HIGH`: Slower calculation, but more accurate approximation to the original number.
  * `HIGHEST`: Slowest calculation, but most accurate approximation to the original number.
```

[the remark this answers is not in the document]
ghpvnist: Here is a more detailed comment on this: https://github.com/google/jax/blob/9bb2c999d6b7eb574c23209a2ec543abad891f7f/jax/_src/lax/lax.py#L661-L679

docs/spec_draft.md
```diff
@@ -1928,6 +1929,123 @@ produces an implementation-defined value.
...
+
+| Name                         | Type                                                        |
+|------------------------------|-------------------------------------------------------------|
+| `lhs`                        | tensor of any supported type                                |
+| `rhs`                        | tensor of any supported type                                |
+| `lhs_batching_dimensions`    | 1-dimensional tensor constant of type `si64`                |
+| `rhs_batching_dimensions`    | 1-dimensional tensor constant of type `si64`                |
+| `lhs_contracting_dimensions` | 1-dimensional tensor constant of type `si64`                |
+| `rhs_contracting_dimensions` | 1-dimensional tensor constant of type `si64`                |
+| `precision_config`           | variadic number of enum of `DEFAULT`, `HIGH`, and `HIGHEST` |
```
ghpvnist, line 1977: I'm wondering whether we should use "or" here since it doesn't make sense for the two `precision_config` values to have different enum types (e.g. [DEFAULT, HIGHEST]) since this would result in opposing precision.
```
variadic number of enum of `DEFAULT`, `HIGH`, or `HIGHEST`
```

ghpvnist: Also noticed that `PACKED_NIBBLE` has been added recently:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/xla_data.proto#L788

docs/spec_draft.md
ghpvnist: This can be left for future work, but does this new formulation affect the current interpreter implementation since we want to have a reference implementation for the interpreter?

stablehlo/dialect/StablehloOps.td
```diff
@@ -2231,10 +2231,25 @@ def StableHLO_DotGeneralOp: StableHLO_ShapedInterfaceOp<"dot_general",
...
+    ```mlir
+    %result = "stablehlo.dot_general"(%lhs, %rhs) {
+      dot_dimension_numbers = #stablehlo.dot<
+        lhs_batching_dimensions = [0],
+        rhs_batching_dimensions = [0],
+        lhs_contracting_dimensions = [2],
+        rhs_contracting_dimensions = [1]
+      >,
+      precision_config = [#stablehlo<precision DEFAULT>, #stablehlo<precision DEFAULT>]
+    } : (tensor<2x2x2xi32>, tensor<2x2x2xi32>) -> tensor<2x2x2xi32>
```
burmako, line 2251: Let's update the .td file as well - both the first sentence and the example.

docs/spec_draft.md
```diff
@@ -1928,6 +1929,123 @@ produces an implementation-defined value.
...
+
+More formally, `result[result_index] = dot_product`, where:
+  * `lhs_result_dimensions = [d for d in axes(lhs) and d not in lhs_batching_dimensions and d not in lhs_contracting_dimensions]`.
+  * `rhs_result_dimensions = [d for d in axes(rhs) and d not in rhs_batching_dimensions and d not in rhs_contracting_dimensions]`.
+  * `result_batching_index + result_lhs_index + result_rhs_index = result_index`
+    where `size(result_batching_index) = size(lhs_batching_dimensions)`,
+    `size(result_lhs_index) = size(lhs_result_dimensions)` and
+    `size(result_rhs_index) = size(rhs_result_dimensions)`.
+  * `transposed_lhs = transpose(lhs, lhs_batching_dimensions + lhs_result_dimensions + lhs_contracting_dimensions)`.
+  * `transposed_lhs_slice = slice(result_batching_index + result_lhs_index + [:, ...:])`.
```
burmako, line 1947: Looks like I made a typo here and missed a comma: "[:, ..., :]". Same below.

docs/spec_draft.md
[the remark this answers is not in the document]
burmako: Reading this without the requisite background knowledge, as a StableHLO producer I wouldn't know what exactly to expect here and as a StableHLO consumer I wouldn't know how to implement this. However, given that numerical accuracy is TBD at the moment, I think it's fine to leave things in this state. I've opened a ticket to elaborate further: https://github.com/openxla/stablehlo/issues/755.

docs/spec_draft.md
burmako: Let's leave resolving this to when we get to the reference implementation. Generally, it is an open question of how the spec is going to be mapped on the interpreter.

docs/spec_draft.md
burmako: Nice find! Yes, indeed it's been added not so long ago. We have https://github.com/openxla/stablehlo/issues/742 to track potentially adding that to StableHLO.

burmako (changes requested):

burmako:
Oops sorry assigned the PR to Sandeep by mistake.

PR #314, opened 2022-10-11
Rename "array of type si64" to "tensor constant of type si64"

burmako:
I remembered that in the spec we already define "tensor constant", so let's use that definition.

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -1054,13 +1054,13 @@ More formally, `result[i0, ..., iR-1]` is equal to:
 
 ### Operands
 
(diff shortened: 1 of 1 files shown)


ghpvnist:
We use the word `n-dimensional array` in the Types section as well. Should these also be reworded?

sdasgup3 (approved):

burmako:
I think that it's fine to use "array" in the definitions of "tensor" and "shape" as a fundamental low-level primitive. However, once we have defined the high-level primitives, let's use them whenever possible.

ghpvnist (approved):

PR #317, opened 2022-10-12
Refactor spec to have consistent usage of "any supported element types"

sdasgup3:
refactor to have consistent usages of "any supported element types"

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -198,8 +198,8 @@ described below)
 ### Semantics
 
 Performs element-wise absolute value of `operand` tensor and produces a `result`
-tensor. For floating-point type, it implements the `abs` operation from the
-IEEE-754 specification.
+tensor. For floating-point element types, it implements the `abs` operation from
+the IEEE-754 specification.
 
 For n-bit signed integer, the absolute value of $-2^{n-1}$ is implementation-
 defined and one of the following:
@@ -210,13 +210,13 @@ defined and one of the following:
 ### Inputs
 
 | Name      | Type                                                      |
-|-----------|-----------------------------------------------------------|
+| --------- | --------------------------------------------------------- |
 | `operand` | tensor of signed integer, floating-point, or complex type |
 
 ### Outputs
 
 | Name     | Type                                                      |
-|----------|-----------------------------------------------------------|
+| -------- | --------------------------------------------------------- |
 | `result` | tensor of signed integer, floating-point, or complex type |
 
 ### Constraints
@@ -253,9 +253,9 @@ defined and one of the following:
 ### Semantics
 
 Performs element-wise addition of two tensors `lhs` and `rhs` and produces a
-`result` tensor. For integer type, if the element-wise sum has an
-unsigned/signed overflow, the result is implementation-defined and one of the
-following:
+`result` tensor. For integer element types, if the element-wise sum has an
+unsigned/signed overflow, the result is implementation-defined and one
+of the following:
 
   * mathematical result modulo $2^n$, where n is the bit width of the result,
   for unsigned overflow. For signed integer overflow, wraps the result around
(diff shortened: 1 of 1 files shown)


docs/spec_draft.md
```diff
@@ -438,16 +438,16 @@ tensor. More formally,
 
 ### Inputs
 
-| Name        | Type                                              |
-|-------------|---------------------------------------------------|
-| `inputs`    | variadic number of tensors of any supported types |
-| `dimension` | `si64`                                            |
+| Name        | Type                                                      |
+|-------------|-----------------------------------------------------------|
+| `inputs`    | variadic number of tensors of any supported element types |
```
burmako, line 443: Should we perhaps go the other way around and contract "element types" just "types"?

burmako, line 443: Also what do you think about the wording "of any supported type" (i.e. singular vs plural)? As I was reading TCP design docs today (e.g. [this one](https://docs.google.com/document/d/1x8kCUsYzFEIeWO38h32bHlmDrieeu9UJTxpKe9gUm_g/edit#heading=h.ey5ld3h9qkk2)), I noticed that they use the signular form, and I think it reads better - "tensor of supported types" might suggest that there may be multiple different types involved.

burmako (changes requested):

docs/spec_draft.md
sdasgup3: We also have Tensor as a supported type, hence the "element type"

burmako: The "Types" section says "Tensor types... Element type E is any one of the supported element types mentioned above", so "any supported type" effectively means "any supported element type".

sdasgup3: makes sense to me.

docs/spec_draft.md
```diff
@@ -1117,7 +1117,7 @@ More formally, `result[i0, ..., iR-1]` is equal to:
 
 | Name                | Type                                         |
 |---------------------|----------------------------------------------|
-| `operand`           | tensor of any supported types                |
+| `operand`           | tensor of any supported element types                |
```
ghpvnist, line 1120: Tables formats are misaligned here and below.

ghpvnist (approved):

docs/spec_draft.md
```diff
@@ -586,14 +586,14 @@ produces an implementation-defined value.
 
 | Name  | Type                                                       |
 |-------|------------------------------------------------------------|
-| `lhs` | tensor of integer, floating-point or complex element types |
-| `rhs` | tensor of integer, floating-point or complex element types |
+| `lhs` | tensor of integer, floating-point or complex type          |
+| `rhs` | tensor of integer, floating-point or complex type          |
```
burmako, line 590: This table looks misaligned.

docs/spec_draft.md
```diff
@@ -586,14 +586,14 @@ produces an implementation-defined value.
...
 
 ### Outputs
 
 | Name     | Type                                                       |
 |----------|------------------------------------------------------------|
-| `result` | tensor of integer, floating-point or complex element types |
+| `result` | tensor of integer, floating-point or complex type          |
```
burmako, line 596: This one too.

docs/spec_draft.md
```diff
@@ -198,8 +198,8 @@ described below)
 ### Semantics
 
 Performs element-wise absolute value of `operand` tensor and produces a `result`
-tensor. For floating-point element types, it implements the `abs` operation from
-the IEEE-754 specification.
+tensor. For floating-point type, it implements the `abs` operation from the
```
burmako, line 201: I think that the plural form (i.e. "types") reads better than the singular form (i.e. "type") outside of the tables.

In my mind, "tensor of signed integer, floating-point, or complex type" reads better with singular because for every given tensor there's just one element type.

To the contrary, I think that "For floating-point types, it implements the `abs` operation from the IEEE-754 specification" reads better with plural because it applies to multiple floating-point types at once (`bf16`, `f16`, etc). When I read "For floating-point type, it implements the `abs` operation", what I find asking myself is "for what type specifically?".

burmako (changes requested):

docs/spec_draft.md
sdasgup3: Thanks for the explanation. Got it!

burmako (approved):

PR #322, opened 2022-10-12
Move shape function of TransposeOp to shared file

zhouxin913:
The bounded dynamism parts of the shape function of TransposeOp can't be moved directly: `TypeInference.h/cpp` is designed to be shared for both StableHLO and MHLO, thus they can't involve anything about the custom Attributes like `TypeExtensionsAttr`. So this PR prepares the bounds for the `operand` before calling the new shared `inferTransposeOp()` and later add the resultBounds on top of the result Type. MHLO is expected to follow the same pattern to call this shared method as well.

A few things open to discussion:
1. Does the existing code https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/StablehloOps.cpp#L4189 mean that we can't have bounds and sparse tensor encoding at the same time?
2. Is there any specific reason we prefer `inferReturnTypes` over `inferReturnTypeComponents`:
`inferReturnTypes` has arg`SmallVectorImpl<Type>& inferredReturnTypes` while `inferReturnTypeComponents` has `SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes`: AFAIK, `ShapedTypeComponents` also supports attribute: https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Interfaces/InferTypeOpInterface.h#L101. If OK, I would consider use `inferReturnTypeComponents` instead, like other ops.

### DIFF
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -183,6 +183,26 @@ LogicalResult TypeExtensionsAttr::verifyEncoding(
       getBounds(), RankedTensorType::get(bounds, elementType), emitError);
 }
 
+llvm::SmallVector<int64_t> getBounds(Type type) {
+  llvm::SmallVector<int64_t> bounds;
+  auto rankedTy = type.dyn_cast<RankedTensorType>();
+  if (rankedTy)
+    if (auto encoding =
+            rankedTy.getEncoding().dyn_cast_or_null<TypeExtensionsAttr>())
+      bounds = llvm::to_vector<4>(encoding.getBounds());
+  return bounds;
+}
+
+void addBounds(MLIRContext* context, SmallVectorImpl<Type>& types,
+               llvm::SmallVector<int64_t> bounds) {
+  if (!bounds.empty()) {
+    auto type = types.pop_back_val().cast<RankedTensorType>();
+    Attribute encoding = TypeExtensionsAttr::get(context, bounds);
+    types.emplace_back(RankedTensorType::get(type.getShape(),
+                                             type.getElementType(), encoding));
+  }
+}
+
 //===----------------------------------------------------------------------===//
 // ReduceScatterOp
 //===----------------------------------------------------------------------===//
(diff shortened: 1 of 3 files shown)


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -183,6 +183,26 @@ LogicalResult TypeExtensionsAttr::verifyEncoding(
...
+void addBounds(MLIRContext* context, SmallVectorImpl<Type>& types,
+               llvm::SmallVector<int64_t> bounds) {
+  if (!bounds.empty()) {
+    auto type = types.pop_back_val().cast<RankedTensorType>();
+    Attribute encoding = TypeExtensionsAttr::get(context, bounds);
+    types.emplace_back(RankedTensorType::get(type.getShape(),
+                                             type.getElementType(), encoding));
+  }
+}
+
```
burmako, line 205: This is why we have `BoundedAttrInterface` and `BoundedDialectInterface::createBoundedAttr` (see example of their usage [here](https://github.com/openxla/stablehlo/blob/424819cc8939b9db1f7d02dbd76f8358a9e234ef/stablehlo/dialect/Base.h#L218-L289)). I think they should be sufficient to avoid passing the bounds around.

stablehlo/dialect/TypeInference.h
```diff
@@ -144,6 +144,13 @@ LogicalResult inferSortOp(
     Region& comparator,
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes);
 
+LogicalResult inferTransposeOp(MLIRContext* context, Optional<Location> loc,
+                               Value operand,
+                               llvm::SmallVector<int64_t>& inputBounds,
+                               DenseIntElementsAttr permutation,
+                               SmallVectorImpl<Type>& inferredReturnTypes,
+                               llvm::SmallVector<int64_t>& resultBounds);
```
burmako, line 152: I think we decided to use `inferReturnTypes` in the discussions back then because we thought that `inferReturnTypeComponents` doesn't support encodings. If it does, then let's use `inferReturnTypeComponents` for consistency with everything else.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -4135,65 +4155,20 @@ LogicalResult TransposeOp::reifyReturnTypeShapes(
...
-    if (hasBounds) {
-      resultBounds.push_back(inputBounds[dim]);
-    }
-  }
+  LogicalResult result = hlo::inferTransposeOp(
+      context, loc, adaptor.getOperand(), inputBounds, adaptor.getPermutation(),
+      inferredReturnTypes, resultBounds);
 
-  // If the input type doesn't have bounds, propagate the input type's encoding
-  // to handle sparse tensor encoding.
```
burmako, line 4190: "Does the existing code https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/StablehloOps.cpp#L4189 mean that we can't have bounds and sparse tensor encoding at the same time?". Yes, that's right.

burmako (changes requested):

stablehlo/dialect/TypeInference.h
smit-hinsu: This sounds good to me.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -948,6 +948,52 @@ LogicalResult inferSortOp(
...
+LogicalResult inferTransposeOp(MLIRContext* context, Optional<Location> loc,
+                               Value operand,
+                               llvm::SmallVector<int64_t>& operandBounds,
+                               DenseIntElementsAttr permutation,
+                               SmallVectorImpl<Type>& inferredReturnTypes,
+                               llvm::SmallVector<int64_t>& resultBounds) {
+  auto type = operand.getType();
+  auto rankedTy = type.dyn_cast<RankedTensorType>();
+  if (!rankedTy) {
+    auto shapedTy = type.dyn_cast<ShapedType>();
```
smit-hinsu, line 960: is this cast necessary? inferredReturnTypes is of type `Type`?

smit-hinsu (commented):
Looks good to me after we use the interface that Eugene mentioned.

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Thanks for this example! Change. 
Revert a lot of change so that comparing to HEAD is more easier to review.

stablehlo/dialect/TypeInference.h
zhouxin913: Ah `inferReturnTypeComponents` does not work: this assert block it:
[inferReturnTypeComponents](https://github.com/llvm/llvm-project/blob/main/mlir/lib/Interfaces/InferTypeOpInterface.cpp#L190)

stablehlo/dialect/StablehloOps.cpp
zhouxin913: What if we need both? I guess it is not possible until this TODO is done? https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/StablehloAttrs.td#L134

stablehlo/dialect/TypeInference.cpp
zhouxin913: Fixed, thanks!

stablehlo/dialect/StablehloOps.cpp
burmako, line 205: It doesn't look like we need these functions anymore?

stablehlo/dialect/StablehloOps.cpp
burmako: I agree. I don't think this will be possible until this TODO is resolved or we find some other way to combine the two.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -948,6 +948,63 @@ LogicalResult inferSortOp(
...
+  std::iota(range.begin(), range.end(), 0);
+  if (!std::is_permutation(range.begin(), range.end(), permutation.begin()))
+    return emitOptionalError(loc,
+                             "attribute permutation must be a permutation"
+                             " of [",
+                             range, "] but got ", permutation);
+
+  llvm::SmallVector<int64_t, 4> inputBounds(rank, ShapedType::kDynamicSize);
+  bool hasBounds = false;
+  BoundedDialectInterface* dialect = nullptr;
```
burmako, line 981: We'll really need to cut down on the boilerplate around getting and setting bounds, but let's wait until similar logic is added to other ops and then see what we can deduplicate.

burmako (changes requested):
There's one more minor change, and the pull request description needs to be updated to follow the changes to the code, but otherwise we're good to go.

stablehlo/dialect/TypeInference.cpp
zhouxin913: Sure makes sense.

stablehlo/tests/infer_stablehlo.mlir
```diff
@@ -644,3 +644,6 @@ func.func @transpose_with_bounds(%arg0: tensor<?x2x?x4xi32, #stablehlo.type_exte
   %1 = "hlo_test_infer.get_return_types"(%0) : (tensor<*xi32>) -> tensor<*xindex>
   func.return %1 : tensor<*xindex>
 }
+
+// -----
+
```
burmako, line 649: Extraneous change / merge artifact?

stablehlo/dialect/StablehloOps.cpp
burmako: Still there in the GitHub UI.

burmako (changes requested):
Also there's a typo in the PR description: "questino".

stablehlo/dialect/StablehloOps.cpp
zhouxin913: It turns out that I mess up with changes among branches, fixed.

burmako (approved):

PR #325, opened 2022-10-13
Add spec for IfOp

### LINKED ISSUE #311: Add spec for IfOp
### Request description

To spec stablehlo.if.

subhankarshah:
closes #311

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -696,7 +696,7 @@ IEEE-754 specification.
 
 ### Semantics
 
-Produces the output from executing exactly one branch from `true_branch` or
+Produces the output from executing exactly one function from `true_branch` or
 `false_branch` depending on the value of `pred`. Formally, if `pred` is `true`,
 output of `true_branch` is returned, else if pred is `false`, output of
 `false_branch` is returned.
@@ -711,15 +711,15 @@ output of `true_branch` is returned, else if pred is `false`, output of
 
 ### Outputs
 
-| Name     | Type                                             |
-|----------|--------------------------------------------------|
-| `result` | variadic number of tensors of any supported type |
+| Name      | Type                                             |
+|-----------|--------------------------------------------------|
+| `results` | variadic number of tensors of any supported type |
 
 ### Constraints
 
   * (C1) `true_branch` and `false_branch` have 0 inputs.
   * (C2) `true_branch` and `false_branch` have the same output types.
-  * (C3) For all `i`, `type(result[i]) = type(branches[0]).outputs[i]`.
+  * (C3) For all `i`, `type(results[i]) = type(true_branch).outputs[i]`.
 
 ### Examples
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -696,27 +696,30 @@ IEEE-754 specification.
 
 ### Semantics
 
-Returns the result of `true_branch` if `pred` is `true`, `false_branch` if
-`pred` is `false`.
+Produces the output from executing exactly one branch from `true_branch` or
+`false_branch` depending on the value of `pred`. Formally, if `pred` is `true`,
+output of `true_branch` is returned, else if pred is `false`, output of
+`false_branch` is returned.
 
 ### Inputs
 
 | Name           | Type                                       |
 |----------------|--------------------------------------------|
 | `pred`         | 1-dimensional tensor constant of type `i1` |
-| `true_branch`  | function of type `function`                |
-| `false_branch` | function of type `function`                |
+| `true_branch`  | `function`                                 |
+| `false_branch` | `function`                                 |
 
 ### Outputs
 
-| Name     | Type               |
-|----------|--------------------|
-| `result` | any supported type |
+| Name     | Type                                             |
+|----------|--------------------------------------------------|
+| `result` | variadic number of tensors of any supported type |
 
 ### Constraints
 
-  * (C1) `true_branch` and `false_branch` take 0 arguments.
-  * (C2) Return types of `true_branch` and `false_branch` are same.
+  * (C1) `true_branch` and `false_branch` have 0 inputs.
+  * (C2) `true_branch` and `false_branch` have the same output types.
+  * (C3) For all `i`, `type(result[i]) = type(branches[0]).outputs[i]`.
 
 ### Examples
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -171,6 +171,7 @@ described below)
    * [divide](#stablehlodivide)
    * [exponential](#stablehloexponential)
    * [floor](#stablehlofloor)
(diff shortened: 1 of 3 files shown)


docs/spec_draft.md
```diff
@@ -680,6 +681,48 @@ IEEE-754 specification.
...
+
+### Outputs
+
+| Name     | Type               |
+|----------|--------------------|
+| `result` | any supported type |
+
+### Constraints
+
+  * (C1) `true_branch` and `false_branch` take 0 arguments.
```
sdasgup3, line 707: At the top we are defining what a function is. IMO, we should say that the function can capture implicit arguments from the parent scope.

burmako: Later today, I was planning to send a PR that describes function semantics. Will ping it here when done.

subhankarshah: @sdasgup3 yes, I was thinking about that as well, as in the functions have 0 arguments but they can use whatever is in the scope of parent function.

docs/spec_draft.md
```diff
@@ -680,6 +681,48 @@ IEEE-754 specification.
...
+|----------------|--------------------------------------------|
+| `pred`         | 1-dimensional tensor constant of type `i1` |
+| `true_branch`  | function of type `function`                |
+| `false_branch` | function of type `function`                |
+
+### Outputs
+
+| Name     | Type               |
+|----------|--------------------|
+| `result` | any supported type |
```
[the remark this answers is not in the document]
subhankarshah: they can be anything, not necessarily a variadic number of tensors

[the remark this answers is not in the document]
burmako: "variadic number of tensors of any supported type" sounds like the right wording to me as well. We don't have ops which return individual element types, and "variadic" is not a type in our type system, so "variadic number of tensors" is as general as we can go.

docs/spec_draft.md
```diff
@@ -680,6 +681,48 @@ IEEE-754 specification.
...
+Returns the result of `true_branch` if `pred` is `true`, `false_branch` if
+`pred` is `false`.
+
+### Inputs
+
+| Name           | Type                                       |
+|----------------|--------------------------------------------|
+| `pred`         | 1-dimensional tensor constant of type `i1` |
+| `true_branch`  | function of type `function`                |
+| `false_branch` | function of type `function`                |
```
burmako, line 697: Since this column says "Type", I think this should be just "\`function\`".

docs/spec_draft.md
```diff
@@ -680,6 +681,48 @@ IEEE-754 specification.
...
+### Outputs
+
+| Name     | Type               |
+|----------|--------------------|
+| `result` | any supported type |
+
+### Constraints
+
+  * (C1) `true_branch` and `false_branch` take 0 arguments.
+  * (C2) Return types of `true_branch` and `false_branch` are same.
```
burmako, line 708: Let's reword the constraints based on the definition of "function type" from https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md#types and of "function" from https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md#programs, i.e.

```
  * (C1) `true_branch` and `false_branch` have 0 inputs.
  * (C2) `true_branch` and `false_branch` have the same output types.
```

docs/spec_draft.md
```diff
@@ -680,6 +681,48 @@ IEEE-754 specification.
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.if
+
+### Semantics
+
+Returns the result of `true_branch` if `pred` is `true`, `false_branch` if
+`pred` is `false`.
```
burmako, line 689: How about "Produces the output of"?
  * "Produces" follows the convention that other ops use and avoids ambiguity with `return`.
  * "the output of" matches how functions are defined in terms of inputs and outputs in the spec.

burmako, line 689: Also, let's call out that only one branch is executed - this is material to the semantics of the op. E.g. "Only one branch is executed depending on the value of `pred`".

stablehlo/dialect/StablehloOps.td
```diff
@@ -1142,15 +1142,18 @@ def StableHLO_IfOp: StableHLO_Op<"if", [
   let summary = "If operator";
```
burmako, line 1142: Let's also delete the comment above `def StableHLO_IfOp`. It is not relevant to StableHLO.

docs/spec_draft.md
burmako, line 708: This also needs "(C3) For all `i`, `type(results[i])` = `type(true_branch).outputs[i]`.

burmako (changes requested):

docs/spec_draft.md
subhankarshah: Produces the output from executing exactly one branch from `true_branch` or
`false_branch` depending on the value of `pred`. Formally, if `pred` is `true`,
output of `true_branch` is returned, else if pred is `false`, output of
`false_branch` is returned.

Split it into two lines, still using the word `return` in second line.

subhankarshah: Captured this in first line of two lines.

docs/spec_draft.md
```diff
@@ -691,6 +692,51 @@ IEEE-754 specification.
...
+
+| Name     | Type                                             |
+|----------|--------------------------------------------------|
+| `result` | variadic number of tensors of any supported type |
+
+### Constraints
+
+  * (C1) `true_branch` and `false_branch` have 0 inputs.
+  * (C2) `true_branch` and `false_branch` have the same output types.
+  * (C3) For all `i`, `type(result[i]) = type(branches[0]).outputs[i]`.
```
burmako, line 722: "type(true_branch)"

burmako (changes requested):
One minor change, and we're good to go.

docs/spec_draft.md
```diff
@@ -691,6 +692,51 @@ IEEE-754 specification.
...
+|----------------|--------------------------------------------|
+| `pred`         | 1-dimensional tensor constant of type `i1` |
+| `true_branch`  | `function`                                 |
+| `false_branch` | `function`                                 |
+
+### Outputs
+
+| Name     | Type                                             |
+|----------|--------------------------------------------------|
+| `result` | variadic number of tensors of any supported type |
```
burmako, line 716: As discussed in the CaseOp PR, this should be called "results". Nice catch!

burmako (changes requested):

burmako (approved):

ghpvnist (approved):

docs/spec_draft.md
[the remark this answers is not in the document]
burmako: Looks like this comment has already been addressed? The spec currently says "(C2) `true_branch` and `false_branch` have the same output types".

PR #331, opened 2022-10-14
Audit status.md and fix mismatches with reality

burmako:
When looking into https://github.com/openxla/stablehlo/pull/253, I realized that I have no idea how to reliably review it in its Markdown form.

I noticed a few values which didn't reflect reality, and I was concerned that there are more, and I'm going to miss them because I don't have a way to systematically review status.md.

I decided to convert status.md into spreadsheet form to automate some aspects of the review, and I'm very happy with the outcome. Now I feel confident that I haven't missed anything. There's no easy way to share the Google Sheets spreadsheet that I used because we don't yet have an OpenXLA-specific Google Drive, but this pull request at least shares the outcome.

Summary of data changes:
  1) "Specification" column didn't need any changes.
  2) "Verification" column was updated based on the rules provided in
     #253, plus I also accounted for open tickets for verification of
     some of the specced ops.
  3) "Type Inference" column needed about a dozen changes with no rhyme
     or reason to them - I went through all the ops, looked at their
     status in the ODS and in StablehloOps.cpp, made changes if needed.
  4) "Pretty Printing" column didn't need any changes.
  5) "Interpreter" column needed one update - "yes" for "multiply".

Summary of style changes:
  1) Rows are now sorted alphabetically (previously they were sorted
     almost alphabetically).
  2) Rows are now aligned left, so that we don't need to think whether
     to align "yes" one whitespace to the left or one whitespace to the
     right to keep it centered.
  3) "yes(need-revisit)" has been renamed to a more concise "revisit".


bhack:
Is there a way to programmatically fill a baseline of this table? 
So that only few things need to be manually annotated..

burmako:
"Is there a way to programmatically fill a baseline of this table?". Yeah, I was thinking the same thing last afternoon when I started working on this. There is some stuff that can be populated automatically, but a lot of it is manual:
  * "Specification" is automatic: https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md#index-of-ops.
  * "Verification"/"Type Inference" can be seeded automatically based on running purely textual scripts over StablehloOps.td and StablehloOps.cpp. The main difficulty is distinguishing between "yes*" and "revisit" - this requires human review, and I populated this information using GitHub PRs for StableHLO and internal PRs for MHLO.
  * "Prettyprinting" is almost automatic: all ops that have assemblyFormat or hasCustomAssemblyFormat + all Elementwise ops are "yes", the rest are "no". There are a few manual exceptions: `map` is "no", and a few ops with assembly format are "revisit".
  * "Interpreter" is automatic: https://github.com/openxla/stablehlo/blob/main/stablehlo/reference/Ops.h.

Luckily I don't think we'll need to do this again, otherwise we should probably automate this.

bhack:
>  "yes*" and "revisit" 

Yes, revisit was expected but It could be just a manual layer over yes or an extra annotated metadata in the source code.

> and a few ops with assembly format are "revisit".
Same

> Luckily I don't think we'll need to do this again, otherwise we should probably automate this.

Why? Are we not going to update this for every op related PR  and every when we will switch no to revisit or yes?
If yes I suppose it could be a constant source of misalignment

burmako:
"Why? Are we not going to update this for every op related PR and also when we will switch no to revisit or yes?". We are, but it seems that maintaining this needs human work anyway - either via updating status.md or via updating extra metadata in source code?

bhack:
> either via updating status.md or via updating extra metadata in source code?

But almost just for the "revist" status that it is like a  "TODO" in source code. But we could always regenerate the table instead of looking again manually in source code in batch for misalignment.

But we will see if the next drift will happen or not. My 2c is that generally it is better to maintain a single ground truth in the source code itself.

burmako:
"My 2c is that generally it is better to maintain a single ground truth in the source code itself and emit tables". Agreed. 

We could have three bits in TableGen definitions along the lines of `let verifierReviewed = 1`, `let typeInferenceReviewed = 1` and `let prettyPrintingReviewed = 1` - and I think that would be sufficient for generating status.md completely automatically.

That would be an incremental improvement over this PR, and these bits could even be populated automatically from status.md.

bhack:
Yes I think it is a nice WIP solution

docs/status.md
```diff
@@ -20,126 +20,129 @@ one of the following tracking labels.
...
+| map                      | no            | yes*         | yes*           | no              | no          |
+| maximum                  | yes           | yes          | yes            | yes             | yes         |
+| minimum                  | yes           | yes          | yes            | yes             | yes         |
+| multiply                 | yes           | yes          | yes            | yes             | yes         |
+| negate                   | yes           | yes          | yes            | yes             | yes         |
+| not                      | yes           | yes          | yes            | yes             | yes         |
+| optimization_barrier     | no            | revisit      | no             | yes             | no          |
+| or                       | yes           | yes          | yes            | yes             | yes         |
+| outfeed                  | no            | revisit      | no             | no              | no          |
+| pad                      | yes           | revisit      | yes            | no              | no          |
```
zhouxin913, line 104: What is the reason we change pad from yes -> revisit?

ghpvnist (approved):

zhouxin913 (approved):

PR #335, opened 2022-10-14
Rename the region accessors for ReduceOp and ReduceWindowOp to match the intent and hlo spec

sdasgup3:
The existing name `body` suits more the region name of whileOp

### DIFF
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -2950,14 +2950,14 @@ LogicalResult ReduceWindowOp::inferReturnTypeComponents(
       location, adaptor.getInputs(), adaptor.getInitValues(),
       adaptor.getWindowDimensions(), adaptor.getWindowStrides(),
       adaptor.getBaseDilations(), adaptor.getWindowDilations(),
-      adaptor.getPadding(), adaptor.getBody(), inferredReturnShapes);
+      adaptor.getPadding(), adaptor.getComputation(), inferredReturnShapes);
 }
 
 // Get the operation used for reduction applied to `result_index`th result. Its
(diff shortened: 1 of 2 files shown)


sdasgup3:
@GleasonK
I believe this change might  effect the compatibility guarantees.  Are we ready at this point in ascertaining those guarantees?

burmako:
Another question is how strongly we feel about renaming these specific fields right away.

There are a lot of opportunities for improving various names in StableHLO/MHLO, but we've been holding off so far to avoid complicating the ongoing migration from MHLO to StableHLO, with the only exception of renaming `operands` to `inputs` because `operands` will stop working in a few weeks.

Perhaps we could start collecting renaming ideas in a ticket similar to #289 (@atondwal has an entire document with ideas), so that at a later point we could evaluate them all in concern and then execute on all of them at once to avoid slow dripping breaking changes.

GleasonK:
> @GleasonK
> I think this change can have compatibility issues.  Are we ready at this point in ascertaining the compatibility guarantees?

We don't offer API compatibility guarantees so I don't think this should be an issue. Can confirm by checking out main, serializing a file, and try to deserialize in this branch. Happy to help with any of that if needed.

burmako (changes requested):
As mentioned in the comment above, I'd like to propose to table the renaming discussions until later when we have a chance to review all the names. This way we'll ensure that: 1) the opset has uniform naming style, 2) all C++/Python breaking changes related to naming are done at once instead of spread over weeks or months of breakages.

sdasgup3:
Superseded by https://github.com/openxla/stablehlo/issues/351

sdasgup3:
@GleasonK No action from you end.

Issue #327, opened 2022-10-13
Migrate all uses of -1 to kDynamicSize


burmako:
See https://discourse.llvm.org/t/rfc-unify-kdynamicsize-and-kdynamicstrideoroffset/64534 for the motivation. We're in a pretty good shape with respect to that, but cursory investigation reveals that we're still using -1 in a few places.

GleasonK:
Added some changes in #569

I see two remaining tasks after that PR is submitted:
1. Fix any Attributes that have printer/parser that relies on the value of `kDynamicSize`, such as gather.
2. Go through the existing error messages and check if any leak the value of `kDynamicSize`.

burmako:
Let's also change all references to `kDynamicSize` to `kDynamic`. In #577, I did that for references in code, but it looks like there are some references in comments which I missed.

burmako:
Implemented in https://github.com/openxla/stablehlo/pull/569.

PR #343, opened 2022-10-17
Handle bounds in the Slice op shape function

zhouxin913:
Also move common parts among this op and Transpose op as common method.

@smit-hinsu add as reviewer as well.


stablehlo/dialect/TypeInference.cpp
```diff
@@ -946,9 +958,14 @@ LogicalResult inferSliceOp(Optional<Location> location, Value operand,
...
   SmallVector<int64_t, 4> strideVals(strides.getValues<int64_t>());
 
+  llvm::SmallVector<int64_t> inputBounds;
+  BoundedDialectInterface* dialect = getBounds(rankedTy, inputBounds);
+
   SmallVector<int64_t, 4> shape;
   shape.reserve(rank);
+  SmallVector<int64_t> resultBounds;
   for (int64_t i = 0, e = rank; i != e; i++) {
+    if (!inputBounds.empty()) resultBounds.push_back(inputBounds[i]);
```
smit-hinsu, line 968: this seems like too conservative? In the test we added, we know that the 2nd dim has size at most 2 with [0, 4) with strides = 2?

Good test!

zhouxin913: Thanks for bring up this!
Yes I'm thinking of this today as well. IIUC, the bound should play the role similar to dim_size, please see my new fixed commit.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -946,9 +958,16 @@ LogicalResult inferSliceOp(Optional<Location> location, Value operand,
...
+  llvm::SmallVector<int64_t> inputBounds;
+  BoundedDialectInterface* dialect = getBounds(rankedTy, inputBounds);
+
   SmallVector<int64_t, 4> shape;
   shape.reserve(rank);
+  SmallVector<int64_t> resultBounds;
   for (int64_t i = 0, e = rank; i != e; i++) {
+    if (!inputBounds.empty())
+      resultBounds.push_back(
+          inferSliceDim(inputBounds[i], start[i], limit[i], strideVals[i]));
```
burmako, line 970: While we're here, let's fix `inferSliceDim` to work with `ShapedType::kDynamicSize` instead of `-1`. It is likely that `-1` will stop working in a few weeks.

burmako (changes requested):

stablehlo/dialect/TypeInference.cpp
```diff
@@ -975,8 +994,11 @@ LogicalResult inferSliceOp(Optional<Location> location, Value operand,
     shape.push_back(inferSliceDim(rankedTy.getDimSize(i), start[i], limit[i],
                                   strideVals[i]));
   }
-  inferredReturnTypes.assign(
-      {RankedTensorType::get(shape, rankedTy.getElementType())});
+  Attribute encoding = inputBounds.empty()
+                           ? rankedTy.getEncoding()
+                           : dialect->createBoundedAttr(resultBounds);
+  inferredReturnTypes.emplace_back(
```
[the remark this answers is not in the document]
zhouxin913: Done.

stablehlo/dialect/TypeInference.cpp
zhouxin913: Fixed. I do not touch any "-1" outside `inferSliceDim`, as I assume there is another issue tracking that already.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -902,6 +902,18 @@ LogicalResult inferReduceWindowOp(
...
+  if (auto encoding =
+          rankedTy.getEncoding().dyn_cast_or_null<BoundedAttrInterface>()) {
+    bounds.append(llvm::to_vector<4>(encoding.getBounds()));
+    BoundedDialectInterface* dialect =
+        cast<BoundedDialectInterface>(&encoding.getDialect());
+    return dialect;
+  }
+  return nullptr;
+}
+
```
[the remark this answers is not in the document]
zhouxin913: Thanks for the suggestion! I think for a while of these inferfaces and was not able to get a clean one.
I copy your comment and signatures unchanged except this:
In both method, should we **report_fatal_error** when not a HLO_BoundedAttrInterface? The original @smit-hinsu 's code just return empty and move on: I assume that the intent was adding a soft check for use cases that use bounds as the encoding. Will strict checks affect MHLO more than expected?

stablehlo/dialect/Base.cpp
```diff
@@ -146,5 +146,24 @@ LogicalResult verifyBounds(ArrayRef<int64_t> bounds, ShapedType type,
   return success();
 }
 
+ArrayRef<int64_t> encodingToBounds(Attribute encoding) {
+  if (!encoding) return {};
```
burmako, line 150: You don't need this line. If encoding is null, then `dyn_cast_or_null` will not crash but return null, which is what we want.

stablehlo/dialect/Base.cpp
```diff
@@ -146,5 +146,24 @@ LogicalResult verifyBounds(ArrayRef<int64_t> bounds, ShapedType type,
...
+}
+
+FailureOr<Attribute> boundsToEncoding(Attribute prototype,
+                                      ArrayRef<int64_t> bounds,
+                                      Optional<Location> loc) {
+  if (bounds.empty()) return prototype;
+  if (!prototype)
+    return emitOptionalError(loc,
+                             "Expect an prototype attribute to obtain the "
+                             "underlying dialect but got none");
```
burmako, line 163: Is it possible that there will be an expected situation where we have non-empty bounds and empty prototype? I don't think so, so I believe that this should be `report_fatal_error` and the return type will no longer need `FailureOr`. The function will also no longer need a `Location`.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -975,8 +980,12 @@ LogicalResult inferSliceOp(Optional<Location> location, Value operand,
     shape.push_back(inferSliceDim(rankedTy.getDimSize(i), start[i], limit[i],
                                   strideVals[i]));
   }
-  inferredReturnTypes.assign(
-      {RankedTensorType::get(shape, rankedTy.getElementType())});
+
+  auto encodingOrErr =
```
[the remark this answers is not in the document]
burmako: I think that with boundsToEncoding simplified as discussed above this might already be simple enough: `RankedTensorType::get(shape, elementType, boundsToEncoding(encoding, bounds))`. What do you think?

stablehlo/dialect/TypeInference.cpp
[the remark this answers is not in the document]
burmako: Good call about report_fatal_error not needed in `encodingToBounds` - I agree. Provided feedback about the rest in a separate comment.

stablehlo/dialect/TypeInference.cpp
burmako: There is: #327.

burmako (changes requested):

stablehlo/dialect/TypeInference.cpp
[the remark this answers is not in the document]
smit-hinsu: that is neat!

stablehlo/dialect/TypeInference.cpp
```diff
@@ -946,9 +958,16 @@ LogicalResult inferSliceOp(Optional<Location> location, Value operand,
   SmallVector<int64_t, 4> limit(limitIndices.getValues<int64_t>());
   SmallVector<int64_t, 4> strideVals(strides.getValues<int64_t>());
 
+  llvm::SmallVector<int64_t> inputBounds;
```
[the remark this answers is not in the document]
zhouxin913: I think this line is gone from an old commit? Resolved with no change made.

stablehlo/tests/infer_stablehlo.mlir
```diff
@@ -647,3 +647,13 @@ func.func @transpose_with_bounds(%arg0: tensor<?x2x?x4xi32, #stablehlo.type_exte
   %1 = "hlo_test_infer.get_return_types"(%0) : (tensor<*xi32>) -> tensor<*xindex>
   func.return %1 : tensor<*xindex>
 }
+
+// -----
+
+// CHECK-LABEL: func @slice_with_bounds
+func.func @slice_with_bounds(%arg0: tensor<3x?xi32, #stablehlo.type_extensions<bounds = [-1, 4]>>) -> tensor<*xindex> {
```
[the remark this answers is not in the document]
zhouxin913: Sure! Added.

stablehlo/dialect/Base.cpp
zhouxin913: Yeah this makes sense: report_fatal_error used in a totally unexpected case which should not happen at all. Fixed.

stablehlo/dialect/TypeInference.cpp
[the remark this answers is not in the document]
zhouxin913: Sure!

stablehlo/dialect/Base.cpp
zhouxin913: OK removed.

burmako (approved):

PR #344, opened 2022-10-17
Add spec for WhileOp

### LINKED ISSUE #326: Add specification for WhileOp
### Request description

To spec stablehlo.while.

subhankarshah:
closes #326

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -1814,7 +1814,7 @@ where `i[d] = j[permutation[d]]`.
 
 ### Semantics
 
-Produces the output from executing `body` function 0 or more times till the
+Produces the output from executing `body` function 0 or more times while the
 `cond` function outputs `true`. More formally, the semantics can be expressed
 using Python-like syntax as follows:
 
@@ -1825,7 +1825,7 @@ while cond(internal_state) == True:
 results = internal_state
 ```
 
-The behaviour of an infinite loop is undefined.
+The behaviour of an infinite loop is implementation-defined.
 
 ### Inputs
 
@@ -1856,15 +1856,15 @@ The behaviour of an infinite loop is undefined.
 // %input0: 0
 // %input1: 10
 %results:2 = "stablehlo.while"(%input0, %input1) ({
-^bb0(%argument0: tensor<i32>, %argument1: tensor<i32>):
-  %predicate = "stablehlo.compare"(%argument0, %argument1) {
-    comparison_direction = #stablehlo<comparison_direction LT>
-  } : (tensor<i32>, tensor<i32>) -> tensor<i1>
-  "stablehlo.return"(%predicate) : (tensor<i1>) -> ()
-},  {
-^bb0(%argument0: tensor<i32>, %argument1: tensor<i32>):
-  %output0 = "stablehlo.add"(%argument0, %constant0) : (tensor<i32>, tensor<i32>) -> tensor<i32>
-  "stablehlo.return"(%output0, %argument1) : (tensor<i32>, tensor<i32>) -> ()
+  ^bb0(%arg0: tensor<i32>, %arg1: tensor<i32>):
+    %0 = "stablehlo.compare"(%arg0, %arg1) {
+      comparison_direction = #stablehlo<comparison_direction LT>
+    } : (tensor<i32>, tensor<i32>) -> tensor<i1>
+    "stablehlo.return"(%0) : (tensor<i1>) -> ()
+}, {
+  ^bb0(%arg0: tensor<i32>, %arg1: tensor<i32>):
+    %0 = "stablehlo.add"(%arg0, %constant0) : (tensor<i32>, tensor<i32>) -> tensor<i32>
+    "stablehlo.return"(%0, %arg1) : (tensor<i32>, tensor<i32>) -> ()
 }) : (tensor<i32>, tensor<i32>) -> (tensor<i32>, tensor<i32>)
 // %results#0: 10
 // %results#1: 10
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -1837,8 +1837,9 @@ previous itertaion.
 
 ### Constraints
 
-  * (C1) For all `i`, `type(operands[i])` = `type(results[i])` = `type(cond).arguments[i]` = `type(body).arguments[i]` = `type(body).outputs[i]`
-  * (C2) `cond` outputs a `tensor<i1>`
+  * (C1) Size of `operands` is at least 1.
+  * (C2) For all `i`, `type(operands[i])` = `type(results[i])` = `type(cond).arguments[i]` = `type(body).arguments[i]` = `type(body).outputs[i]`
+  * (C3) `cond` outputs a `tensor<i1>`
 
 ### Examples
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -1813,13 +1813,19 @@ where `i[d] = j[permutation[d]]`.
 ## stablehlo.while
 
 ### Semantics
+
 Produces the output from executing `body` function 0 or more times till the
-`cond` function outputs `true`. More formally, `body` function is executed once
-every time the `cond` function returns `true` and all execution stops once
-`cond` function returns `false`. For the first iteration both `cond` and `body`
-arguments are initialzed with the vaules of arguments of the `while` operation,
-for subsequent iterations they are fed with the output of `body` from the
-previous itertaion.
+`cond` function outputs `true`. More formally, the semantics can be expressed
+using Python-like syntax as follows:
+
+```python
+internal_state = operands
+while cond(internal_state) == True:
+  internal_state = body(internal_state)
+results = internal_state
+```
+
+The behaviour of an infinite loop is undefined.
 
 ### Inputs
 
@@ -1837,33 +1843,31 @@ previous itertaion.
 
 ### Constraints
 
-  * (C1) Size of `operands` is at least 1.
-  * (C2) For all `i`, `type(operands[i])` = `type(results[i])` = `type(cond).arguments[i]` = `type(body).arguments[i]` = `type(body).outputs[i]`
-  * (C3) `cond` outputs a `tensor<i1>`
+  * (C1) `cond` has type `(T0, ..., TN-1) -> tensor<i1>`, where
+         `Ti` = `type(operands[i])`.
+  * (C2) `body` has type `(T0, ..., TN-1) -> (T0, ..., TN-1)`, where
+         `Ti` = `type(operands[i])`.
+  * (C3) For all `i`, `type(results[i])` = `type(operands[i])`.
 
 ### Examples
 
 ```mlir
 // %constant0: 1
-// %constant1: 2
 // %input0: 0
 // %input1: 10
-// %input2: 0
-%results:3 = "stablehlo.while"(%input0, %input1, %input2) ({
-^bb0(%argument0: tensor<i32>, %argument1: tensor<i32>, %argument2: tensor<i32>):
+%results:2 = "stablehlo.while"(%input0, %input1) ({
+^bb0(%argument0: tensor<i32>, %argument1: tensor<i32>):
   %predicate = "stablehlo.compare"(%argument0, %argument1) {
     comparison_direction = #stablehlo<comparison_direction LT>
   } : (tensor<i32>, tensor<i32>) -> tensor<i1>
   "stablehlo.return"(%predicate) : (tensor<i1>) -> ()
 },  {
-^bb0(%argument0: tensor<i32>, %argument1: tensor<i32>, %argument2: tensor<i32>):
+^bb0(%argument0: tensor<i32>, %argument1: tensor<i32>):
   %output0 = "stablehlo.add"(%argument0, %constant0) : (tensor<i32>, tensor<i32>) -> tensor<i32>
-  %output2 = "stablehlo.add"(%argument2, %constant1) : (tensor<i32>, tensor<i32>) -> tensor<i32>
-  "stablehlo.return"(%output0, %argument1, %output2) : (tensor<i32>, tensor<i32>, tensor<i32>) -> ()
-}) : (tensor<i32>, tensor<i32>, tensor<i32>) -> (tensor<i32>, tensor<i32>, tensor<i32>)
+  "stablehlo.return"(%output0, %argument1) : (tensor<i32>, tensor<i32>) -> ()
+}) : (tensor<i32>, tensor<i32>) -> (tensor<i32>, tensor<i32>)
 // %results#0: 10
 // %results#1: 10
-// %results#2: 20
 ```
 
 [Back to Ops](#index-of-ops)
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -1813,10 +1813,13 @@ where `i[d] = j[permutation[d]]`.
 ## stablehlo.while
 
 ### Semantics
-Produces the output from executing `body` function zero or more times till the
+Produces the output from executing `body` function 0 or more times till the
 `cond` function outputs `true`. More formally, `body` function is executed once
 every time the `cond` function returns `true` and all execution stops once
-`cond` function returns `false`.
+`cond` function returns `false`. For the first iteration both `cond` and `body`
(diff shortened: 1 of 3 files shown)


docs/spec_draft.md
```diff
@@ -1632,6 +1633,60 @@ where `i[d] = j[permutation[d]]`.
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.while
+
+### Semantics
+Produces the output from executing `body` function zero or more times till the
+`cond` function outputs `true`. More formally, `body` function is executed once
+every time the `cond` function returns `true` and all execution stops once
+`cond` function returns `false`.
```
sdasgup3, line 1642: We need to say something about how the data flow across the two region blocks. For example, the two regions arguments are initialed with the inputs of `stablehlo.while` in the first iteration. In the subsequent iterations the arguments are fed by the return value of the `body` region.

We can use https://mlir.llvm.org/docs/Dialects/SCFDialect/#scfwhile-mlirscfwhileop as  a reference.

subhankarshah: Yes, let me update the semantics

subhankarshah: Does this look better now?

docs/spec_draft.md
```diff
@@ -1808,6 +1809,64 @@ where `i[d] = j[permutation[d]]`.
...
+## stablehlo.while
+
+### Semantics
+Produces the output from executing `body` function 0 or more times till the
+`cond` function outputs `true`. More formally, `body` function is executed once
+every time the `cond` function returns `true` and all execution stops once
+`cond` function returns `false`. For the first iteration both `cond` and `body`
+arguments are initialzed with the vaules of arguments of the `while` operation,
+for subsequent iterations they are fed with the output of `body` from the
+previous itertaion.
```
burmako, line 1821: Consider expressing the formalization in pseudocode - I think that'll be much easier to read. For example: 

"""
More formally, the semantics can be expressed as follows using Python-like 
syntax:

```C++
internal_state = operands;
while (cond(internal_state) == true) {
  internal_state = body(internal_state);
}
results = internal_state;
```
"""

burmako, line 1821: When reframed as pseudocode, this description suggests the question of what happens if the loop ends up being infinite. Do we allow that? Do we disallow that? Do we call this undefined behavior?

docs/spec_draft.md
```diff
@@ -1808,6 +1809,64 @@ where `i[d] = j[permutation[d]]`.
...
+},  {
+^bb0(%argument0: tensor<i32>, %argument1: tensor<i32>, %argument2: tensor<i32>):
+  %output0 = "stablehlo.add"(%argument0, %constant0) : (tensor<i32>, tensor<i32>) -> tensor<i32>
+  %output2 = "stablehlo.add"(%argument2, %constant1) : (tensor<i32>, tensor<i32>) -> tensor<i32>
+  "stablehlo.return"(%output0, %argument1, %output2) : (tensor<i32>, tensor<i32>, tensor<i32>) -> ()
+}) : (tensor<i32>, tensor<i32>, tensor<i32>) -> (tensor<i32>, tensor<i32>, tensor<i32>)
+// %results#0: 10
+// %results#1: 10
+// %results#2: 20
+```
```
burmako, line 1866: Let's simplify the example, because we likely don't need three inputs and two constants to demonstrate how while loops work. 

E.g. how about a while loop that increments a `tensor<i32>` until it hits 10? Using one input and one constant, we'll demonstrate how data flows from operands to cond/body to body etc + implicit capture in cond.

burmako (changes requested):

joker-eph:
I have a chain of CL (not finished) that is changing mhlo.while to match better scf.while in how values are flowing between the condition and the body, which is a pre-requisite or the representation to work with RegionBranchInterface and in general with how we think of control-flow between regions in MLIR. Shall we align on where we want to land before documenting as "spec" what happens to be just the current implementation?

burmako:
I think that documenting the current implementation of `stablehlo.while` is valuable because this is a well-understood iteration of the opset - it is used in production via MHLO with a number of users across the community, and it was unanimously approved to be snapshotted as StableHLO. 

Of course, this doesn't preclude improvements beyond the snapshotted opset. I think it would be great to have an RFC to change `stablehlo.while` to explore alternative designs! If I recall correctly, one part of the motivation for redesigning `mhlo.while` was enabling better optimizations, and that sounds like a compelling reason. After the RFC gets accepted, the updated spec will supersede the current spec.

"Shall we align on where we want to land before documenting as "spec" what happens to be just the current implementation?". tl;dr. I think it that would be great to both align on a future design and to document the current design, and both these processes can proceed in parallel.

docs/spec_draft.md
```diff
@@ -1808,6 +1809,64 @@ where `i[d] = j[permutation[d]]`.
...
+
+### Outputs
+
+| Name      | Type                                             |
+|-----------|--------------------------------------------------|
+| `results` | variadic number of tensors of any supported type |
+
+### Constraints
+
+  * (C1) Size of `operands` is at least 1.
```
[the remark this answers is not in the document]
sdasgup3: It seems that HLO does accept 0 size of operands. Shouldn't we align with that?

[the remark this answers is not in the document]
burmako: Nice catch! Similarly, `if` and `case` cannot have 0 outputs in StableHLO, but I don't think that HLO prohibits that. What should we do about that?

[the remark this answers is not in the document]
joker-eph: > It seems that HLO does accept 0 size of operands. Shouldn't we align with that?

To me the only way to answer at such question is "what is the use case"? What are we trying to support? 
IMO our users does not include HLO : it should be about TF/JAX/PyTorch/... and what they may want to express), MHLO and HLO are rather "constraints" in that StableHLO has to ensure that before exposing anything to users it has to be able to be supported by MHLO downstream.

docs/spec_draft.md
```diff
@@ -1808,6 +1809,64 @@ where `i[d] = j[permutation[d]]`.
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.while
+
+### Semantics
+Produces the output from executing `body` function 0 or more times till the
+`cond` function outputs `true`. More formally, `body` function is executed once
```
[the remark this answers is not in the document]
subhankarshah: until cond is `false`   <==>  till cond is `true`
I was also a bit unsure.
Maybe we skip both in favour of > as long as cond is true

docs/spec_draft.md
subhankarshah: Thanks this looks good. Concise and expressive.

docs/spec_draft.md
subhankarshah: Sure

docs/spec_draft.md
subhankarshah: the interpreter suggests that, It is allowed and results in undefined behaviour

docs/spec_draft.md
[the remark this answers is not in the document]
subhankarshah: I cant think of any use cases, as an empty operand list would mean, it does not return/modify anything. It can either be an empty while loop that ends immediately or an infinite loop, either of which is not useful. We can add this to the verifier and add a constraint or let this be as is. Let me know your thoughts or if there is a use case that might be possible for such a construct.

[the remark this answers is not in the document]
burmako: So far, we don't have side-effecting ops in the spec, so it doesn't matter, but in the future we will, e.g. distribution ops. With distribution ops like send/recv, it's not inconceivable to have a loop body that doesn't take inputs but still does useful work. Same for if/case. Let's open a ticket to get to the bottom of this.

docs/spec_draft.md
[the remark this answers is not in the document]
burmako: In my mind, "until" and "till" are synonyms, which is why I proposed a rewording. Okay, how about this: "while the `cond` function outputs `true`"?

docs/spec_draft.md
```diff
@@ -1808,6 +1809,68 @@ where `i[d] = j[permutation[d]]`.
...
+    comparison_direction = #stablehlo<comparison_direction LT>
+  } : (tensor<i32>, tensor<i32>) -> tensor<i1>
+  "stablehlo.return"(%predicate) : (tensor<i1>) -> ()
+},  {
+^bb0(%argument0: tensor<i32>, %argument1: tensor<i32>):
+  %output0 = "stablehlo.add"(%argument0, %constant0) : (tensor<i32>, tensor<i32>) -> tensor<i32>
+  "stablehlo.return"(%output0, %argument1) : (tensor<i32>, tensor<i32>) -> ()
+}) : (tensor<i32>, tensor<i32>) -> (tensor<i32>, tensor<i32>)
+// %results#0: 10
+// %results#1: 10
```
burmako, line 1869: Proposed reformatting:

```
  %results:2 = "stablehlo.while"(%input0, %input1) ({
    ^bb0(%arg0: tensor<i32>, %arg1: tensor<i32>):
      %0 = "stablehlo.compare"(%arg0, %arg1) {
        comparison_direction = #stablehlo<comparison_direction LT>
      } : (tensor<i32>, tensor<i32>) -> tensor<i1>
      "stablehlo.return"(%0) : (tensor<i1>) -> ()
  }, {
    ^bb0(%arg0: tensor<i32>, %arg1: tensor<i32>):
      %0 = "stablehlo.add"(%arg0, %constant0) : (tensor<i32>, tensor<i32>) -> tensor<i32>
      "stablehlo.return"(%0, %arg1) : (tensor<i32>, tensor<i32>) -> ()
  }) : (tensor<i32>, tensor<i32>) -> (tensor<i32>, tensor<i32>)
```

Feedback:
  1. As far as aligning code goes, let's follow the code example in sort.
  2. `%argN` is an idiomatic way of saying "%argumentN" in MLIR.
  3. I think it's fine to shorten variable names to `%N` when auxiliary ops are involved. Naming inputs and outputs of the primary op is useful because this allows to easily map the spec on them, but everything else - probably not necessary.

burmako (changes requested):

docs/spec_draft.md
```diff
@@ -1808,6 +1809,68 @@ where `i[d] = j[permutation[d]]`.
...
+using Python-like syntax as follows:
+
+```python
+internal_state = operands
+while cond(internal_state) == True:
+  internal_state = body(internal_state)
+results = internal_state
+```
+
+The behaviour of an infinite loop is undefined.
```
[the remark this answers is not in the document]
joker-eph: Termination has profound effects and I think this aspects requires careful consideration and as such should be well motivated.

Same for UB vs Implementation-defined: can't swap each other freely and the spec should define the terminology and the intent behind it.

docs/spec_draft.md
```diff
@@ -1808,6 +1809,68 @@ where `i[d] = j[permutation[d]]`.
...
+| `body`     | `function`                                       |
+
+### Outputs
+
+| Name      | Type                                             |
+|-----------|--------------------------------------------------|
+| `results` | variadic number of tensors of any supported type |
+
+### Constraints
+
```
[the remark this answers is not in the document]
subhankarshah: #365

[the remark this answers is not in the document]
subhankarshah: `if` and `case` can have time delay loops within them (if time delay is a use case), specifically with while since 0 outputs mean 0 inputs to while and 0 inputs to condition and body as well, a legitimate use case that is not an empty or an infinite loop is hard to imagine.

docs/spec_draft.md
```diff
@@ -1808,6 +1809,68 @@ where `i[d] = j[permutation[d]]`.
...
+using Python-like syntax as follows:
+
+```python
+internal_state = operands
+while cond(internal_state) == True:
+  internal_state = body(internal_state)
+results = internal_state
+```
+
+The behaviour of an infinite loop is implementation-defined.
```
burmako, line 1827: I agree with Mehdi that "undefined behavior" and "implementation-defined behavior" mean two different things, and undefined behavior is not a good fit here.

From https://en.wikipedia.org/wiki/Undefined_behavior: "In computer programming, undefined behavior (UB) is the result of executing a program whose behavior is prescribed to be unpredictable, in the language specification to which the computer code adheres. This is different from unspecified behavior, for which the language specification does not prescribe a result, and implementation-defined behavior that defers to the documentation of another component of the platform (such as the ABI or the translator documentation)".

"The behaviour of an infinite loop is implementation-defined" seems like a good fit for the spec draft, given that it: 1) calls out a corner case, 2) specifies its behavior in a way that's consistent with existing implementations without precluding further refinements. I think it would be okay to merge this PR and leave the rest to #383, but let's wait for Mehdi's input before doing that.

@joker-eph Do you have objections to the current wording? Or maybe advice on how to define this better for this op and/or at the scope of the entire spec?

burmako (approved):

docs/spec_draft.md
joker-eph: I don't understand why you think that UB isn't a good fit here?

burmako: As you said in the nearby comment thread, termination requires careful consideration. Saying that infinite loops are UB is one way to provide a definitive ruling on this, but I think there are other reasonable alternatives if we want to go deeper into things like memory / thread model.

Correct me if I'm wrong, but in my mind "implementation-defined" is the most non-committal way to describe this behavior if we don't want to definitively decide on this at the moment, hence the proposed wording.

joker-eph: Actually no: undefined behavior is the least committal way of describe anything: it basically forbids this in the input. It's like saying "it isn't permitted".
"Implementation-defined" means that StableHLO supports it and no generic code can assume it won't be present.

burmako: Looks like we might be using different words to describe similar things. What I meant is that calling this implementation-defined seems to be the least amount of guarantees that we can make to consumers at the time. 

If in the future, as part of working on #383 and #384, we'll want to turn non-termination into UB (and I think we likely will), then we'll be strengthening these guarantees. This seems easier to do than go the other way around - call this UB right now, and if upon closer inspection we decide that non-termination is not UB, then relax the guarantees.

joker-eph: > Looks like we might be using different words to describe similar things.

Not clear to me: we can't be using "implementation defined" and "undefined behavior" for the same thing...

> What I meant is that calling this implementation-defined seems to be the least amount of guarantees that we can make to consumers at the time.

I am not sure I understand this sentence, maybe because "consumers" isn't clear to me here?

> This seems easier to do than go the other way around - call this UB right now, and if upon closer inspection we decide that non-termination is not UB, then relax the guarantees.

I have the complete opposite view on this. You can't make something UB later, but you can define it later.

burmako: "we can't be using "implementation defined" and "undefined behavior" for the same thing". I think you and I both mean the same thing for "implementation defined" and "undefined behavior", but we might have a difference in opinions on what "the least committal" means in this case. Let's discuss this in a higher-bandwidth setting and follow up with a resolution here afterwards.

burmako: Please change to "The behaviour of an infinite loop is TBD" as discussed above. After that, we can merge.

burmako (changes requested):

subhankarshah:
@burmako The approval was reversed. Please approve again, I have resolved the comment.

burmako (approved):

docs/status.md
```diff
@@ -144,5 +144,5 @@ one of the following tracking labels.
 | unary_einsum             | no            | revisit      | no             | no              | no          |
 | uniform_dequantize       | no            | yes*         | yes*           | yes             | no          |
 | uniform_quantize         | no            | yes*         | infeasible     | yes             | no          |
-| while                    | no            | yes*         | yes*           | revisit         | no          |
+| while                    | yes           | revisit      | yes            | revisit         | no          |
```
sdasgup3, line 147: Approved otherwise!

Do let me know why the verification is a `revisit` for while/If/Case?

sdasgup3 (approved):

docs/status.md
subhankarshah: Checkout #365 
revisit is to finalise behaviour on 0 operands

PR #275, opened 2022-10-07
spec for GatherOp

sdasgup3:
fixes #245

PR #353, opened 2022-10-17
Add spec for ReduceOp

### LINKED ISSUE #300: Add spec for ReduceOp
### Request description

To spec stablehlo.reduce.

ghpvnist:
closes #300

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -1701,34 +1701,48 @@ and produces a `result` tensor.
 
 ### Semantics
 
-Applies a `computation` function to `inputs` and `init_values` along the
-`dimensions` and produces a `result` tensor.
+Applies a `body` function to `inputs` and `init_values` along the `dimensions`
+and produces a `result` tensor.
+
+The `body` function has to be associative, and it produces an
+implementation-defined behavior otherwise. Similarly, if `init_values` do not
+form identity under the `body` function, it produces an implementation-defined
+behavior.
 
 ### Inputs
 
 | Name          | Type                                             |
 |---------------|--------------------------------------------------|
 | `inputs`      | variadic number of tensors of any supported type |
-| `init_values` | variadic number of tensors of any supported type |
+| `init_values` | variadic number of scalars of any supported type |
 | `dimensions`  | 1-dimensional tensor constant of type `si64`     |
-| `computation` | `function`                                       |
+| `body`        | `function`                                       |
 
 ### Outputs
 
-| Name     | Type                         |
-|----------|------------------------------|
-| `result` | tensor of any supported type |
+| Name      | Type                                             |
+|-----------|--------------------------------------------------|
+| `results` | variadic number of tensors of any supported type |
 
 ### Constraints
 
-  * (C1) `inputs`, `init_values` and `result` have the same element type.
-  * (C2) `inputs` and `init_values` have the same type.
-  * (C3) `inputs` have N tensors where N >= 1.
-  * (C4) 0 $\lt$ `dimensions[d]` $\lt$ rank(`inputs[d]`) for all dimension `d`.
-  * (C5) `computation` inputs have the same type as `inputs`, and output has the
-  same type as `result`.
-  * (C6) `init_values` have to form an identity under the `computation`.
-  * (C7) `dim(result, i) = dim(inputs[0], j)` for all `j` $\notin$ `dimensions`.
+  * (C1) `inputs`, `init_values` have the same element type.
+  * (C2) All `inputs` have the same shape.
+  * (C3) type(`inputs[k]`) $=$ type(`init_values[k]`) for all `k` $\in$ [0, N).
+  * (C4) size(`inputs`) $=$ size(`init_values`) $=$ size(`result`) $=$ N where
+  N >= 1.
+  * (C5) `init_values` have to form an identity under the `body`.
+  * (C6) 0 $\lt$ `dimensions[d]` $\lt$ rank(`inputs[d]`) for all dimension `d`.
+  * (C7) All dimensions in `dimensions` are unique.
+  * (C8) The full function type of `body` is
+  `(T0,..., TN-1, T0, ..., TN-1) -> (T0, ..., TN-1)`, where the first and second
+  set of input types `T0,..., TN-1` corresponds to the element types of `inputs`
+  and `init_values` respectively. The output types `(T0,..., TN-1)` corresponds
+  to the element types of `results`.
+  * (C9) rank(`results[k]`) = rank(`inputs[k]`) - size(`dimensions`) for all `k`
+  $\in$ [0, N).
+  * (C10) dim(`results[k], d`) $=$ dim(`inputs[k], d`) if `d` $\notin$
+  `dimensions`.
 
 ### Examples
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -1701,20 +1701,38 @@ and produces a `result` tensor.
 
 ### Semantics
 
-Applies a `body` function to `inputs` and `init_values` along the `dimensions`
-and produces a `result` tensor.
+Performs a reduction function `body` over the `inputs` and `init_values` along
+the `dimensions` and produces a `result` tensor.
 
-The `body` function has to be associative, and it produces an
-implementation-defined behavior otherwise. Similarly, if `init_values` do not
-form identity under the `body` function, it produces an implementation-defined
-behavior.
+The order of reductions is implementation-defined, which means that `body` and
+`init_values` must form a monoid to guarantee that the operation produces the
+same results for all inputs on all implementations.
+
+However, this condition doesn't hold for many popular reductions. E.g.
+floating-point addition for `body` and zero for `init_values` don't actually
+form a monoid because floating-point addition is not associative. What this
+means for numeric precision is implementation-defined.
+
+More formally, `results[:][result_index] = reduce(input_slices)` where:
+  * `input_slices` = `inputs[:][ri0, ..., :, ..., riR-1]`, where `ri` are
+    individual elements in `result_index` and `:` are inserted at `dimensions`.
+  * `reduce(input_slices)` = `eval(schedule)` for some binary tree `schedule`
+    where:
+    * `eval(node)` = `body(eval(node.left), eval(node.right))`.
+    * `eval(leaf)` = `leaf.value`.
+  * `schedule` is an implementation-defined full binary tree whose in-order
+    traversal consists of:
+    * `input_slices[:][index]` values, for all `index` in the index space
+      of `input_slices`, in the ascending lexicographic order of `index`.
+    * Interspersed with an implementation-defined amount of `init_values`
+      at implementation-defined positions.
 
 ### Inputs
 
 | Name          | Type                                             |
 |---------------|--------------------------------------------------|
 | `inputs`      | variadic number of tensors of any supported type |
-| `init_values` | variadic number of scalars of any supported type |
+| `init_values` | 0-dimensional tensor of any supported type       |
 | `dimensions`  | 1-dimensional tensor constant of type `si64`     |
 | `body`        | `function`                                       |
 
@@ -1726,35 +1744,31 @@ behavior.
 
 ### Constraints
 
-  * (C1) `inputs`, `init_values` have the same element type.
-  * (C2) All `inputs` have the same shape.
-  * (C3) type(`inputs[k]`) $=$ type(`init_values[k]`) for all `k` $\in$ [0, N).
-  * (C4) size(`inputs`) $=$ size(`init_values`) $=$ size(`result`) $=$ N where
+  * (C1) All `inputs` have the same shape.
+  * (C2) element_type(`inputs[k]`) $=$ element_type(`init_values[k]`) $=$
+  element_type(`results[k]`) for all `k` $\in$ [0, N).
+  * (C3) size(`inputs`) $=$ size(`init_values`) $=$ size(`results`) $=$ N where
   N >= 1.
-  * (C5) `init_values` have to form an identity under the `body`.
-  * (C6) 0 $\lt$ `dimensions[d]` $\lt$ rank(`inputs[d]`) for all dimension `d`.
-  * (C7) All dimensions in `dimensions` are unique.
-  * (C8) The full function type of `body` is
-  `(T0,..., TN-1, T0, ..., TN-1) -> (T0, ..., TN-1)`, where the first and second
-  set of input types `T0,..., TN-1` corresponds to the element types of `inputs`
-  and `init_values` respectively. The output types `(T0,..., TN-1)` corresponds
-  to the element types of `results`.
-  * (C9) rank(`results[k]`) = rank(`inputs[k]`) - size(`dimensions`) for all `k`
-  $\in$ [0, N).
-  * (C10) dim(`results[k], d`) $=$ dim(`inputs[k], d`) if `d` $\notin$
-  `dimensions`.
+  * (C4) 0 $\le$ `dimensions[d]` $\lt$ rank(`inputs[0][d]`) for all dimension
+  `d`.
+  * (C5) All dimensions in `dimensions` are unique.
+  * (C6) The full function type of `body` is
+  `(T0, ..., TN-1, T0, ..., TN-1) -> (T0, ..., TN-1)`, where the first and
+  second set of input types `T0,..., TN-1` corresponds to the element types of `inputs` and `init_values` respectively. The output types `(T0, ..., TN-1)` corresponds to the element types of `results`.
+  * (C7) shape(`results[k]`) $=$ shape(`inputs[k]`) except that the dimension
+  sizes of `inputs[k]` corresponding to `dimensions` are not included.
 
 ### Examples
 
 ```mlir
-// %inputs0 = [[0, 1, 2, 3, 4, 5]]
-// %init_values = 0
-%result = "stablehlo.reduce"(%inputs0, %init_values) ({
+// %input = [[0, 1, 2, 3, 4, 5]]
+// %init_value = 0
+%result = "stablehlo.reduce"(%input, %init_value) ({
   ^bb0(%arg0: tensor<i32>, %arg1: tensor<i32>):
     %0 = "stablehlo.add"(%arg0, %arg1) : (tensor<i32>, tensor<i32>) -> tensor<i32>
     "stablehlo.return"(%0) : (tensor<i32>) -> ()
 }) {
-  dimensions = dense<[1]> : tensor<1xi64>
+  dimensions = dense<1> : tensor<1xi64>
 } : (tensor<1x6xi32>, tensor<i32>) -> tensor<1xi32>
 // %result = [15]
 ```
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -1701,8 +1701,8 @@ and produces a `result` tensor.
 
 ### Semantics
 
-Performs a reduction function `body` over the `inputs` and `init_values` along
-the `dimensions` and produces a `result` tensor.
+Applies a function `body` to `inputs` and `init_values` along the `dimensions`
+and produces a `result` tensor.
 
 The order of reductions is implementation-defined, which means that `body` and
 `init_values` must form a monoid to guarantee that the operation produces the
@@ -1729,12 +1729,12 @@ More formally, `results[:][result_index] = reduce(input_slices)` where:
 
 ### Inputs
 
-| Name          | Type                                             |
-|---------------|--------------------------------------------------|
-| `inputs`      | variadic number of tensors of any supported type |
-| `init_values` | 0-dimensional tensor of any supported type       |
-| `dimensions`  | 1-dimensional tensor constant of type `si64`     |
-| `body`        | `function`                                       |
+| Name          | Type                                                           |
+|---------------|----------------------------------------------------------------|
+| `inputs`      | variadic number of tensors of any supported type               |
+| `init_values` | variadic number of 0-dimensional tensors of any supported type |
+| `dimensions`  | 1-dimensional tensor constant of type `si64`                   |
+| `body`        | `function`                                                     |
 
 ### Outputs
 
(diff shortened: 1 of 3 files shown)


docs/spec_draft.md
```diff
@@ -1266,6 +1267,56 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+### Constraints
+
+  * (C1) `inputs`, `init_values` and `result` have the same element type.
+  * (C2) `inputs` and `init_values` have the same type.
+  * (C3) `inputs` have N tensors where N >= 1.
+  * (C4) 0 $\lt$ `dimensions[d]` $\lt$ rank(`inputs[d]`) for all dimension `d`.
+  * (C5) `computation` inputs have the same type as `inputs`, and output has the
+  same type as `result`.
+  * (C6) `init_values` have to form an identity under the `computation`.
+  * (C7) `dim(result, i) = dim(inputs[0], j)` for all `j` $\notin$ `dimensions`.
```
sdasgup3, line 1301: rank(results[k]) = inputs[k] - size(dimensions) for all k $\in$ [0, N)

docs/spec_draft.md
```diff
@@ -1266,6 +1267,56 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+
+| Name     | Type                         |
+|----------|------------------------------|
+| `result` | tensor of any supported type |
+
+### Constraints
+
+  * (C1) `inputs`, `init_values` and `result` have the same element type.
+  * (C2) `inputs` and `init_values` have the same type.
+  * (C3) `inputs` have N tensors where N >= 1.
```
sdasgup3, line 1296: size(`inputs`) $=$ size(`init_values`) $=$ size(`result`) $=$ N and N $\ge$ 1

docs/spec_draft.md
```diff
@@ -1266,6 +1267,56 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+### Outputs
+
+| Name     | Type                         |
+|----------|------------------------------|
+| `result` | tensor of any supported type |
+
+### Constraints
+
+  * (C1) `inputs`, `init_values` and `result` have the same element type.
+  * (C2) `inputs` and `init_values` have the same type.
```
sdasgup3, line 1295: type(inputs[k]) = type(init_values[k]) for all k $\in$ [0, N)

docs/spec_draft.md
```diff
@@ -1266,6 +1267,56 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+| `dimensions`  | 1-dimensional tensor constant of type `si64`     |
+| `computation` | `function`                                       |
+
+### Outputs
+
+| Name     | Type                         |
+|----------|------------------------------|
+| `result` | tensor of any supported type |
+
+### Constraints
```
sdasgup3, line 1292: All `inputs` have the same shape.

docs/spec_draft.md
sdasgup3, line 1301: All the dimensions in `dimensions` are unique.

docs/spec_draft.md
sdasgup3, line 1292:
```
The full function type of `body` is `(T0,..., TN-1, T0, ..., TN-1) -> (T0, ..., TN-1)`,
         where the first and second set of input types `T0,..., TN-1`
         corresponds to the element types of `inputs` and `init_values`
         respectively. The output types `(T0,..., TN-1)` corresponds to the
         element types of `results`.
```

sdasgup3, line 1292: Ca we say something about the shape of any result[k] ? Something along the lines of 
https://github.com/tensorflow/tensorflow/blob/ab758b2c95382eeba943c741e7f8f3d9ba1b4f2b/tensorflow/compiler/xla/service/shape_inference.cc#L2304

stablehlo/dialect/StablehloOps.td
```diff
@@ -1341,9 +1341,9 @@ def StableHLO_ReduceOp: StableHLO_ShapedInterfaceOp<"reduce", [
 
   let hasCustomAssemblyFormat = 1;
 
-  // TODO(hinsu): Verify that the attached body arguments and results are
+  // TODO(hinsu): Verify that the attached computation arguments and results are
   // compatible with reduce op's operands.
-  let regions = (region SizedRegion<1>:$body);
+  let regions = (region SizedRegion<1>:$computation);
```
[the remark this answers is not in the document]
ghpvnist: Does the spec continue to use the former name (body) or use the new name (computation), and fix the ODS in the future?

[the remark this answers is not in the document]
burmako: The spec should be using current names (at the moment, in this case that's `body`). When updating the names in the future, we'll update both the spec and the ODS at the same time.

docs/spec_draft.md
```diff
@@ -1266,6 +1267,56 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+
+  * (C1) `inputs`, `init_values` and `result` have the same element type.
+  * (C2) `inputs` and `init_values` have the same type.
+  * (C3) `inputs` have N tensors where N >= 1.
+  * (C4) 0 $\lt$ `dimensions[d]` $\lt$ rank(`inputs[d]`) for all dimension `d`.
+  * (C5) `computation` inputs have the same type as `inputs`, and output has the
+  same type as `result`.
+  * (C6) `init_values` have to form an identity under the `computation`.
```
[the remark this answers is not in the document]
ghpvnist: I believe so. From what I understand, every reduce iteration takes previous value and does computation with current value. Since the first iteration does not have a previous value, the init_value should provide an identity value.

The example in XLA doc shows two ways of computing reduction of 1-D array [10, 11, 12, 13]:
```
f(10, f(11, f(12, f(init_value, 13)))
but there are also many other possibilities, e.g.
f(init_value, f(f(10, f(init_value, 11)), f(f(init_value, 12), f(init_value, 13))))
```

Here, if the reduction is a summation and init_value was not 0, and say 1, the result of the two variations will be different.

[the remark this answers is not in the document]
sdasgup3: You are right.
I was mostly concerned about verification of this particular constraint. It seems that we should write something like

It is implementation defined behavior if the `initial value` has a effect on the `computation`

ghpvnist:
Status for verification is set to `revisit` since we currently do not have any verification to assert that `init_values` form an identity for `body` reduce function.

docs/spec_draft.md
```diff
@@ -1267,6 +1268,70 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+
+  * (C1) `inputs`, `init_values` have the same element type.
+  * (C2) All `inputs` have the same shape.
+  * (C3) type(`inputs[k]`) $=$ type(`init_values[k]`) for all `k` $\in$ [0, N).
+  * (C4) size(`inputs`) $=$ size(`init_values`) $=$ size(`result`) $=$ N where
+  N >= 1.
+  * (C5) `init_values` have to form an identity under the `body`.
+  * (C6) 0 $\lt$ `dimensions[d]` $\lt$ rank(`inputs[d]`) for all dimension `d`.
+  * (C7) All dimensions in `dimensions` are unique.
+  * (C8) The full function type of `body` is
```
sdasgup3, line 1308: Even though I suggested the constraint C8, but it seems I was not entirely correct. Following up on https://github.com/openxla/stablehlo/issues/369, there is a bit of additional information that we might need to convey. 

How about we say 
`(T0,..., TN-1, T'0, ..., T'N-1) -> (T0, ..., TN-1)`, where
 * Tk corresponds to the element type of inputs[k] or `results[k]` ignoring floating-point precision.
 * T'k corresponds to the element type of `init_values[k]`    ignoring floating-point precision.

burmako: `SameElementTypeIgnoringFpPrecision` seems to be a frequent occurrence in shape_inference.cc (e.g. it's used for concat, pad, binary, map, etc), but personally I've always treated it as a quirk of client HLO rather than something that we should follow in MHLO or StableHLO. What are the pros and cons of adding this feature to the spec? What is the motivation on the HLO side?

sdasgup3: >> What is the motivation on the HLO side?

This allows HLO verifier to accomodate mixed precision in HLO instruction, that is the operands, output, and parameters/output of the called computations can have different precisions (BF16 and F32).

There exists [passes](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/bfloat16_normalization.h) to add converters (f32->bf16 and reverse) so as to normalize the mixed precision version of the program.

If the implementation does not support bf16 or mixed precision, then the converters are folded away using [link](https://github.com/tensorflow/tensorflow/blob/b2ad6ba8c77da03fb10a894699a0c1a850a4322f/tensorflow/compiler/xla/service/bfloat16_conversion_folding.h#L25).

>> What are the pros and cons of adding this feature to the spec?

### Pros
The StableHLO spec will be aligned verbatim with the verifier. 

### Cons
To me it seems like a details tied to the implementation and would increase the complexity of the spec.
 
My take:
Even if we do not add this detail (which, more specifically, is a relaxation to type check), user can fully rely on the spec only to write  "type verified" program. After thinking more about it, I am aligned with your though process.

docs/spec_draft.md
```diff
@@ -1267,6 +1268,70 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+
+| Name      | Type                                             |
+|-----------|--------------------------------------------------|
+| `results` | variadic number of tensors of any supported type |
+
+### Constraints
+
+  * (C1) `inputs`, `init_values` have the same element type.
+  * (C2) All `inputs` have the same shape.
+  * (C3) type(`inputs[k]`) $=$ type(`init_values[k]`) for all `k` $\in$ [0, N).
```
burmako, line 1302:
```
* (C3) element_type(`inputs[k]`) $=$ element_type(`init_values[k]`) $=$
element_type(`results[k]`) for all `k` $\in$ [0, N).
```

Feedback:
  1. Only element types are equal, types are not never equal because inputs cannot be scalars.
  2. We also need to call out element types of results because they are never called out otherwise.

docs/spec_draft.md
```diff
@@ -1267,6 +1268,70 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+  * (C4) size(`inputs`) $=$ size(`init_values`) $=$ size(`result`) $=$ N where
+  N >= 1.
+  * (C5) `init_values` have to form an identity under the `body`.
+  * (C6) 0 $\lt$ `dimensions[d]` $\lt$ rank(`inputs[d]`) for all dimension `d`.
+  * (C7) All dimensions in `dimensions` are unique.
+  * (C8) The full function type of `body` is
+  `(T0,..., TN-1, T0, ..., TN-1) -> (T0, ..., TN-1)`, where the first and second
+  set of input types `T0,..., TN-1` corresponds to the element types of `inputs`
+  and `init_values` respectively. The output types `(T0,..., TN-1)` corresponds
+  to the element types of `results`.
```
burmako, line 1312:
```
* (C8) `body` has type `(tensor<E0>, ..., tensor<EN-1>, tensor<E0>, ..., tensor<EN-1>) -> (tensor<E0>, ..., tensor<EN-1>)`
where `Ek = element_type(inputs[k])`.
```

Feedback:
  1. I don't think constraints should have an opinion on whether the first or the second half of parameters map on inputs or init_values, unless this matters for verification / type inference. This should be covered in semantics.
  2. `Ek = element_type(inputs[k])` is sufficient to describe all the correspondence relationships.

docs/spec_draft.md
```diff
@@ -1267,6 +1268,70 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+  * (C9) rank(`results[k]`) = rank(`inputs[k]`) - size(`dimensions`) for all `k`
+  $\in$ [0, N).
+  * (C10) dim(`results[k], d`) $=$ dim(`inputs[k], d`) if `d` $\notin$
+  `dimensions`.
```
burmako, line 1316: As it's currently written, C10 is incorrect because the `d` in `dim(inputs[k], d)` actually needs to increase as we move past `d`s in `dimensions`. 

The correct formulation would be pretty laborious to write, so I propose that we replace C9 and C10 with a slightly informal formulation, similarly to what we did in #275 in a similar situation:

```
* (C9) `shape(results[k]) = shape(inputs[k])` except that the dimension sizes
of `inputs[k]` corresponding to `dimensions` are not included.
```

burmako (changes requested):

docs/spec_draft.md
burmako: "Given that MHLO doesn't support mixed precision". Actually I was wrong about this, and #369 goes into details of where exactly mixed precision is currently supported in MHLO (and in StableHLO).

Since this is a cross-cutting concern, i.e. it applies to a bunch of ops - not just `reduce` and not limited to the ops that #369 is talking about - I propose that we don't address it in this PR, but instead write the spec as if mixed precision is not a thing and then address all the ops at once.

docs/spec_draft.md
```diff
@@ -1267,6 +1268,70 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+
+### Semantics
+
+Applies a `body` function to `inputs` and `init_values` along the `dimensions`
+and produces a `result` tensor.
+
+The `body` function has to be associative, and it produces an
+implementation-defined behavior otherwise. Similarly, if `init_values` do not
+form identity under the `body` function, it produces an implementation-defined
+behavior.
```
burmako, line 1281: And here's the formalization (I'm surprised at how compact it turned out):

```
More formally, `results[k][result_index] = reduce(input_slices)[k]` for all `k`
$\in$ [0, N), where:
  * `input_slices[k]` = `inputs[k][ri0, ..., :, ..., riR-1]` for all `k` 
    $\in$ [0, N), where `ri` are individual elements in `result_index` and `:` 
    are inserted at `dimensions`.
  * `reduce(input_slices)` = `eval(schedule)` for some binary tree `schedule` 
    where:
      * `eval(node)` = `body(eval(node.left), eval(node.right))`.
      * `eval(leaf)` = `leaf.value`.
  * `schedule` is an implementation-defined full binary tree whose in-order
    traversal consists of:
      * `input_slices[:][index]` values, for all `index` in the index space
        of `input_slices`, in the ascending lexicographic order of `index`.
      * Interspersed with an implementation-defined amount of `init_values`
        at implementation-defined positions.
```

pifon2a:
Hi, I have a question about verification of `mhlo.reduce` and whether the behavior should be changed in stablehlo or maybe in stablehlo and mhlo.

```
    %0:3 = mhlo.reduce (%arg0 init: %arg3),
                                     (%arg1 init: %arg4), 
                                     (%arg2 init: %arg5) across dimensions = [0, 1] 
         : (tensor<?x?xf32>, tensor<?x4xi32>, tensor<?x5xi32>, 
            tensor<f32>, tensor<i32>, tensor<i32>
          ) -> (tensor<f32>, tensor<i32>, tensor<i32>)
     reducer(%arg6: tensor<f32>, %arg9: tensor<f32>) 
                 (%arg7: tensor<i32>, %arg10: tensor<i32>) 
                 (%arg8: tensor<i32>, %arg11: tensor<i32>)  {
      %1 = mhlo.add %arg6, %arg9 : tensor<f32>
      %2 = mhlo.add %arg7, %arg10 : tensor<i32>
      %3 = mhlo.add %arg8, %arg11 : tensor<i32>
      mhlo.return %1, %2, %3 : tensor<f32>, tensor<i32>, tensor<i32>
    }
 ```

This is at the moment an IR that passes verification. Notice, that `%arg1` has `?x?` shape, `%arg1` has `?x4` and `%arg2` has `?x5` shape. Verification checks that shape of `%arg0` is "compatible" with `?x4` and `?x5` using this code in `stablehlo/dialect/TypeInference.cpp`:

```
if (failed(mlir::verifyCompatibleShape(inputArgTypes[rankedInputIdx],
                                             inputArgTypes[inputIdx]))) {
        return emitOptionalError(
            location, "expects all inputs to have compatible shapes. Shape at",
            " input-index ", inputIdx,
      }
```

i.e. it matches the first ranked input and tries to compare the rest of the ranked input shapes to it.

Would it be possible/feasible to insert casts to reshape all inputs to the common static shape? Example, if we have `tensor<?x?xf32>`, `tensor<?x10xf32>` and `tensor<4x?xf32` inputs, then we can cast all of them to `tensor<4x10xf32>`.

File: docs/spec_draft.md (hunk 11 above)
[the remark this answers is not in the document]
Comment (ghpvnist) [reply docs/spec_draft.md:1281]: Thanks for the suggestion.

File: docs/spec_draft.md (hunk 10 above)
Comment (ghpvnist) [reply docs/spec_draft.md:1316]: This is a neat way to put it, thanks.

**Comment by sdasgup3** [comment]
@pifon2a Thanks for pointing out. We will address this a part of  https://github.com/openxla/stablehlo/issues/394.

**Comment by ghpvnist** [comment]
> Hi, I have a question about verification of `mhlo.reduce` and whether the behavior should be changed in stablehlo or maybe in stablehlo and mhlo.

Hi @pifon2a, yes you are right, and this should be changed in both stablehlo and mhlo.

> Would it be possible/feasible to insert casts to reshape all inputs to the common static shape? Example, if we have `tensor<?x?xf32>`, `tensor<?x10xf32>` and `tensor<4x?xf32` inputs, then we can cast all of them to `tensor<4x10xf32>`.

I believe this is okay since we cannot verify whether the dynamic shapes are compatible (it would be a runtime error).

File: docs/spec_draft.md (hunk 11 above)
Comment (ghpvnist) [reply docs/spec_draft.md]: > `schedule` is an implementation-defined full binary tree whose in-order
    traversal consists of:

Though the actual traversal order is implementation-defined (and whether interpreter should compute in parallel), shouldn't we say reverse in-order traversal since we want to compute the bottom-most nodes first if we want it to be computed in parallel:
reduce sum: `[1, 2, 3, 4]`
`f(f(f(f(i, 1), 2), 3), 4)`
```
                   f(f(f(f(i, 1), 2), 3), 4)
                      /                   \
      f(f(f(i, 1), 2), 3)                  4
         /                \
  f(f(i, 1), 2)            3
     /    \
f(i, 1)    2
  /  \ 
 i    1 
```
vs
`f(f(f(i, 1), f(i, 2)), f(f(i, 3), f(i, 4)))`
```
    f(f(f(i, 1), f(i, 2)), f(f(i, 3), f(i, 4)))
           /                            \
   f(f(i, 1), f(i, 2))        f(f(i, 3), f(i, 4))
     /        \                       /        \
 f(i, 1)   f(i, 2)            f(i, 3)        f(i, 4)
  /  \        /  \                /  \         /  \
 i    1      i    2              i    3       i    4
```

Comment (sdasgup3) [reply docs/spec_draft.md]: This is what I understood from the formalism:
1. `input_slices = ...`. The order of elements in `input_slices`, which depends on how `result_index` are visited, is not mentioned. 
2. Given a particular order of elements in `input_slices`, there could be multiple binary trees following the same order via inorder traversal. It's implementation defined which one of those are picked. In the example above, both the trees are fine to be created by two different implementations as they have the same in-order traversals.  

4. Given a particular tree, the order in which the nodes (including non-leaf nodes) are evaluated is related more to how someone wants to implement it: which does not matter in the formulation. Maybe a non-leaf node can be evaluated partially before moving on to the some other and can be resumed later once its dependencies are met or evaluation  follows a recursive top down approach where the leafs will be evaluated first (as you pointed out). 

I think the reason for (1) is that the order does not matter. Now (2) is based on that order. My question is why we need the full binary tree schedule to follow the order in (1) via in-order traversal. Why not any just full binary tree with the elements of `input_slice`?

Comment (burmako) [reply docs/spec_draft.md]: "It's implementation defined which one of those are picked. In the example above, both the trees are fine to be created by two different implementations as they have the same in-order traversals". Yes, that's what I had in mind in the proposed formulation. The in-order traversal called out above is only used to constrain the order in which the leaves are laid out in these potential trees and is unrelated the traversal used to evaluate a particular tree.

"My question is why we need the full binary tree schedule to follow the order in (1) via in-order traversal. Why not any just full binary tree with the elements of input_slice?". I think that's because we don't know if the reduction operation is commutative or not. So it's fine to do `(a op b) op (c op d)`, `((a op b) op c) op d)` etc. But it's not fine to do e.g. `(b op a) op (c op d)`.

File: docs/spec_draft.md [hunk 12]
```diff
@@ -1268,6 +1269,89 @@ More formally, `result[i0, ..., iR-1]` is equal to:
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.reduce
+
+### Semantics
+
+Performs a reduction function `body` over the `inputs` and `init_values` along
```
burmako, line 1276: I don't think it's super idiomatic to say "perform a function over inputs". "Apply a function to inputs" is one of the more idiomatic options.

docs/spec_draft.md
```diff
@@ -1268,6 +1269,89 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+      of `input_slices`, in the ascending lexicographic order of `index`.
+    * Interspersed with an implementation-defined amount of `init_values`
+      at implementation-defined positions.
+
+### Inputs
+
+| Name          | Type                                             |
+|---------------|--------------------------------------------------|
+| `inputs`      | variadic number of tensors of any supported type |
+| `init_values` | 0-dimensional tensor of any supported type       |
```
burmako, line 1307: I think this should be "variadic number of 0-dimensional tensors of any supported type".

docs/status.md
```diff
@@ -107,7 +107,7 @@ one of the following tracking labels.
 | real                     | no            | yes*         | yes*           | yes             | no          |
 | real_dynamic_slice       | no            | revisit      | no             | yes             | no          |
 | recv                     | no            | revisit      | no             | no              | no          |
-| reduce                   | no            | yes*         | yes*           | revisit         | no          |
+| reduce                   | yes           | yes          | yes            | revisit         | no          |
```
burmako, line 110: I was wrong to initially recommend to flip verification to "yes". While we're figuring out #369, I think it should be "revisit".

docs/spec_draft.md
```diff
@@ -1268,6 +1269,89 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+  * (C3) size(`inputs`) $=$ size(`init_values`) $=$ size(`results`) $=$ N where
+  N >= 1.
+  * (C4) 0 $\le$ `dimensions[d]` $\lt$ rank(`inputs[0][d]`) for all dimension
+  `d`.
+  * (C5) All dimensions in `dimensions` are unique.
+  * (C6) The full function type of `body` is
+  `(T0, ..., TN-1, T0, ..., TN-1) -> (T0, ..., TN-1)`, where the first and
+  second set of input types `T0,..., TN-1` corresponds to the element types of
+  `inputs` and `init_values` respectively. The output types `(T0, ..., TN-1)`
+  corresponds to the element types of `results`.
```
burmako, line 1331: C6 is no longer needed now that we have C8. Given that, let's overwrite C6 with C8, so that the constraints are ordered in the same way as inputs and outputs are, i.e. C5: dimensions, C6: body, C7: results.

burmako (changes requested):

burmako (approved):

PR #354, opened 2022-10-17
Add spec for ReduceWindowOp

### LINKED ISSUE #301: Add spec for ReduceWindowOp
### Request description

To spec the stablehlo.reduce_window.

sdasgup3:
fixes #301 

Verification is "revisit" because of #369.


docs/spec_draft.md
```diff
@@ -1267,6 +1268,137 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+
+<p align="center">
+  <img src="spec_draft/reduce_window.png" />
+</p>
+
+Conceptually, the kth dimension of `inputs[k]` is dilated by implicitly
+placing `base_dilations[k] - 1` holes between each of the entries in that
+dimension.
+
+`padding[k][0]` and `padding[k][1]` specify the amount of
```
ghpvnist, line 1289: This should be updated to match 1-dimensional `padding`.

docs/spec_draft.md
```diff
@@ -1267,6 +1268,137 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+    * Default value of 1 for each dimension of inputs[k] for any k $\in$ [0, N).
+    * size(`base_dilations`) $=$ rank(`inputs[k]`) for any k $\in$ [0, N).
+    * `base_dilations[i]` $\gt 0$ for all i $\in$ [0, size(base_dilations))
+
+  * (C7) On `window_dilations` (optional)
+    * Default value of 1 for each dimension of inputs[k] for any k $\in$ [0, N).
+    * size(`window_dilations`) $=$ rank(`inputs[k]`) for any k $\in$ [0, N).
+    * `window_dilations[i]` $\gt 0$ for all i $\in$ [0, size(window_dilations))
+
+  * (C8) On `padding`(optional)
```
ghpvnist, line 1355: Can you add a constraint or point out that negative padding should not result in `negative` dimension output or that it would result in an empty result? See `PadOp`'s type inference related to negative padding:
https://github.com/openxla/stablehlo/blob/8b49e34d73752d10d94d929a5952f4bbf3eb7f1d/stablehlo/dialect/TypeInference.cpp#L861-L865

docs/spec_draft.md
```diff
@@ -1267,6 +1268,137 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+         N $\ge$ 1
+
+  * (C2) All `init_values` values have to form an identity under reduction
+         `body`. The behavior of the operation is undefined otherwise.
+
+  * (C3) All `inputs` have the same shape.
+
+  * (C4) On `window_dimensions`
+    * size(`window_dimensions`) $=$ rank(`inputs[k]`) for any k $\in$ [0, N).
+    * `window_dimensions[i]` $\gt 0$ for all i $\in$ [0, size(window_dimensions))
```
ghpvnist, line 1338: [0, size(`window_dimensions`))

docs/spec_draft.md
```diff
@@ -1267,6 +1268,137 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+
+  * (C5) On `window_strides` (optional)
+    * Default value of 1 for each dimension of inputs[k] for any k $\in$ [0, N).
+    * size(`window_strides`) $=$ rank(`inputs[k]`) for any k $\in$ [0, N).
+    * `window_strides[i]` $\gt 0$ for all i $\in$ [0, size(window_strides))
```
ghpvnist, line 1343: [0, size(`window_strides`))

docs/spec_draft.md
```diff
@@ -1267,6 +1268,137 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+
+  * (C6) On `base_dilations` (optional)
+    * Default value of 1 for each dimension of inputs[k] for any k $\in$ [0, N).
+    * size(`base_dilations`) $=$ rank(`inputs[k]`) for any k $\in$ [0, N).
+    * `base_dilations[i]` $\gt 0$ for all i $\in$ [0, size(base_dilations))
```
ghpvnist, line 1348: [0, size(`base_dilations`))

docs/spec_draft.md
```diff
@@ -1267,6 +1268,137 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+    * size(`window_dilations`) $=$ rank(`inputs[k]`) for any k $\in$ [0, N).
+    * `window_dilations[i]` $\gt 0$ for all i $\in$ [0, size(window_dilations))
+
+  * (C8) On `padding`(optional)
+    * Default value of (0,0) for each dimension of inputs[k] for any k $\in$ [0, N).
+    * size(`padding`) $=$ rank(`inputs[k]`) for any k $\in$ [0, N).
+    * dim(`padding[k]`, 0) $=$ N and dim(`padding[k]`, 1) = 2 for
+      any k $\in$ [0, N).
+
+  * (C9) The `body` has type `(T0,..., TN-1, T0, ..., TN-1) ->
```
ghpvnist, line 1361: `(T0, ..., TN-1, T0, ..., TN-1)` Similarly below.

ghpvnist (changes requested):

docs/spec_draft.md
```diff
@@ -1267,6 +1268,137 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+
+| Name                | Type                                             | Constraints |
+|---------------------|--------------------------------------------------|-------------|
+| `inputs`            | variadic number of tensors of any supported type | (C1), (C3)  |
+| `init_values`       | variadic number of scalars of any supported type | (C1), (C2)  |
+| `window_dimensions` | 1-dimensional tensor constant type `si64`        | (C4)        |
+| `window_strides`    | 1-dimensional tensor constant type `si64`        | (C5)        |
+| `base_dilations`    | 1-dimensional tensor constant type `si64`        | (C6)        |
+| `window_dilations`  | 1-dimensional tensor constant type `si64`        | (C7)        |
+| `padding`           | 2-dimensional tensor constant type `si64`        | (C8)        |
```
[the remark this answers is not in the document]
ghpvnist: I see that default value should be 0-equivalent.

docs/spec_draft.md
```diff
@@ -1267,6 +1268,137 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+    * size(`base_dilations`) $=$ rank(`inputs[k]`) for any k $\in$ [0, N).
+    * `base_dilations[i]` $\gt 0$ for all i $\in$ [0, size(base_dilations))
+
+  * (C7) On `window_dilations` (optional)
+    * Default value of 1 for each dimension of inputs[k] for any k $\in$ [0, N).
+    * size(`window_dilations`) $=$ rank(`inputs[k]`) for any k $\in$ [0, N).
+    * `window_dilations[i]` $\gt 0$ for all i $\in$ [0, size(window_dilations))
+
+  * (C8) On `padding`(optional)
+    * Default value of (0,0) for each dimension of inputs[k] for any k $\in$ [0, N).
```
ghpvnist, line 1356: Why is the default value specifically a two value tuple (0, 0)? Perhaps just 0-equivalent value? (e.g. false for boolean and (0.0, 0.0) for complex)

ghpvnist (changes requested):

docs/spec_draft.md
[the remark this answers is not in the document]
sdasgup3: The `I64ElementsAttr` can be multi-dimensional as well. It is basically a `DenseIntElementsAttr`.
Also, we have a check https://github.com/openxla/stablehlo/blob/72cbcdf530620c321aac46506447b8eb1f3a69d8/stablehlo/dialect/TypeInference.cpp#L116

docs/spec_draft.md
sdasgup3: Please see the comment above.

docs/spec_draft.md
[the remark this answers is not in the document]
ghpvnist: Ah I didn't realize they could be multi-dimensional. Thanks for the pointers you can ignore my comment.

docs/spec_draft.md
```diff
@@ -1267,6 +1268,137 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+  * (C1) size(`inputs`) $=$ size(`init_values`) $=$ size(`results`) $=$ N and
+         N $\ge$ 1
+
+  * (C2) All `init_values` values have to form an identity under reduction
+         `body`. The behavior of the operation is undefined otherwise.
+
+  * (C3) All `inputs` have the same shape.
+
+  * (C4) On `window_dimensions`
+    * size(`window_dimensions`) $=$ rank(`inputs[k]`) for any k $\in$ [0, N).
```
ghpvnist, line 1335: Can `window_dimensions` be larger than the shape of `inputs[k]`?

docs/spec_draft.md
```diff
@@ -1267,6 +1268,137 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+    * `window_dimensions[i]` $\gt 0$ for all i $\in$ [0, size(window_dimensions))
+
+  * (C5) On `window_strides` (optional)
```
ghpvnist, line 1339: Do you think we should allow `window_strides` to be potentially larger than the shape of `inputs[k]`? I think leaving this constraint out is fine but just wanted to hear your thoughts.

docs/spec_draft.md
```diff
@@ -1267,6 +1268,132 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+    * Default value of 1 for each dimension of inputs[k] for any k $\in$ [0, N).
+    * size(`window_strides`) $=$ rank(`inputs[k]`) for any k $\in$ [0, N).
+    * `window_strides[i]` $\gt 0$ for all i $\in$ [0, size(`window_strides`))
+
+  * (C6) On `base_dilations` (optional)
+    * Default value of 1 for each dimension of inputs[k] for any k $\in$ [0, N).
+    * size(`base_dilations`) $=$ rank(`inputs[k]`) for any k $\in$ [0, N).
+    * `base_dilations[i]` $\gt 0$ for all i $\in$ [0, size(`base_dilations`))
+
+  * (C7) On `window_dilations` (optional)
```
ghpvnist, line 1349: These are all related questions, but should we restrict the `window_dilations` to be up to shape of `inputs[k]` or leave it since a combination of `window_dimensions` and `window_strides` could affect the resulting window to be out of bounds anyway?

docs/spec_draft.md
```diff
@@ -1267,6 +1268,137 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+### Semantics
+
+Performs a reduction function `body` over all elements in each window of a
+sequence of N `inputs` tensors and produces a sequence of N `results` tensor.
+
+The following diagram shows how elements in `results[k]` are computed from
+`operand[k]` using a concrete example.
+
+<p align="center">
+  <img src="spec_draft/reduce_window.png" />
```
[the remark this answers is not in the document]
sdasgup3: I have updated the image to properly bold out the inputs and added relevant colors to the result computations. I have not yet added the comprehensive inputs args as they are all mentioned below the relevant image blocks. IMO, the facts the inputs are not properly bolded might have caused the confusion, Please let me know how it looks now.

docs/spec_draft.md
```diff
@@ -1267,6 +1268,132 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+### Semantics
+
+Performs a reduction function `body` over all elements in each window of a
+sequence of N `inputs` tensors and produces a sequence of N `results` tensor.
+
+The following diagram shows how elements in `results[k]` are computed from
+`operand[k]` using a concrete example.
+
+<p align="center">
+  <img src="spec_draft/reduce_window.svg" />
```
ghpvnist, line 1282: Can we use a darker shade of green? It's a bit hard to see the colored text.

docs/spec_draft.md
```diff
@@ -1267,6 +1268,132 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+    * `window_dimensions[i]` $\gt 0$ for all i $\in$
+      [0, size(`window_dimensions`))
+
+  * (C5) On `window_strides` (optional)
+    * Default value of 1 for each dimension of inputs[k] for any k $\in$ [0, N).
+    * size(`window_strides`) $=$ rank(`inputs[k]`) for any k $\in$ [0, N).
+    * `window_strides[i]` $\gt 0$ for all i $\in$ [0, size(`window_strides`))
+
+  * (C6) On `base_dilations` (optional)
+    * Default value of 1 for each dimension of inputs[k] for any k $\in$ [0, N).
```
ghpvnist, line 1345: Need backtick: `inputs[k]`

ghpvnist (changes requested):

docs/spec_draft.md
sdasgup3: Yes, that is possible in which case the corresponding window will be discarded.

docs/spec_draft.md
sdasgup3: Thanks for bringing this up. 
Like `window_dimensions`, the strides for a dimension d can be more than dim(inputs[k], d) in which case the corresponding window will be discarded for reduction.

docs/spec_draft.md
sdasgup3: Yes you are right. The dilated window could span beyond the shape of inputs[k] and if that happened the window is discarded for reduction computation.

docs/spec_draft.md
sdasgup3: The padding tuple (a, b) for a particular dimension d represents the a number of padding values to be added  at the low-end (next to index 0) and b number of padding values to be added at the high-end (next to the
highest index) of dimension `d` respectively. For this op, the semantics is that the padding should not affect the result of computation: So either we can skip them or use int_value as the padding value.

sdasgup3: The default value (0,0) means no padding added at either ends.

docs/spec_draft.md
sdasgup3: For this op, if the negative padding results in negative dimension size, then the corresponding dimension size of the result is negative.
This is taken care of in C9
`strided_bound[i] = (padded_dialated_input[i] == 0 || dialated_window[i] > padded_dialated_input[i]) ? 0 : floor((padded_dialated_input[i] - dialated_window[i]) / window_strides[i]) + 1`.

docs/spec_draft.md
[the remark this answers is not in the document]
sdasgup3: Yes the default padding value is same as init_value. Please note that here the default value means the no padding added to either side of dimension.

https://github.com/openxla/stablehlo/pull/354#discussion_r1003826337

docs/spec_draft.md
ghpvnist: Ah I think there was a confusion regarding this. First, I didn't realize that padding could be 2-dimensional, so now I understand what (0, 0) was meant for. However, for this specific case, I was referring to what the default of the padded tensor would be. For example, if we pad a 2x2 matrix with (1, 1), what is the filled value to the cells that was added as a result of the padding?
```
// operand
[-1, -1]
[-1, -1]

// padded operand
[?, ?]
[-1, -1]
[-1, -1]
[?, ?]
```
What is the value of `?` is this the same as `init_values[k]`? This could be related to the actual implementation, but since if we default this to 0-equivalent, doing a reduction operation like max on a 4x1 window of [0, -1, -1, 0] where 0 is the "default" padded value, then would returning a 0 be a side-effect or intentional?

sdasgup3: >> This could be related to the actual implementation
Yes. If the implementation decides on **not** to skip those values and rather include them in computation, then `init_value` should be the choice. IOW, having `init_value` for padding and dilation holes will have the same effect as skipping those.

ghpvnist: Makes sense, as you mentioned, for product reduction, since the `init_value` would most likely be 1, the dilation holes having the value 1 would have the same effect as skipping it.

docs/spec_draft.md
```diff
@@ -1267,6 +1268,145 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+### Semantics
+
+Performs a reduction function `body` over all elements in each window of a
+sequence of N `inputs` tensors and produces a sequence of N `results` tensor.
+
+The following diagram shows how elements in `results[k]` are computed from
+`operand[k]` using a concrete example.
+
+<p align="center">
+  <img src="spec_draft/reduce_window.svg" />
```
ghpvnist, line 1282: One last change, similar to how we specify `input[k]`, can you do the same for `input_value = 0` -> `input_values[k] = 0` on the right and also in the 4 calculations at the bottom?

ghpvnist (approved):

docs/spec_draft.md
sdasgup3: Thanks for the catch! done

ghpvnist (approved):

burmako (changes requested):

docs/spec_draft.md
```diff
@@ -1268,6 +1269,145 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+    where:
+    * `eval(node)` = `body(eval(node.left), eval(node.right))`.
+    * `eval(leaf)` = `leaf.value`.
+
+  * `schedule` is an implementation-defined full binary tree whose in-order
+    traversal consists of:
+    * `input_slices[:][index]` values, for all `index` in the index space
+      of `input_slices`, in the ascending lexicographic order of `index`.
+    * Interspersed with an implementation-defined amount of `init_values`
+      at implementation-defined positions.
```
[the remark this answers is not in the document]
burmako: Actually, there's no need to duplicate the definition of `reduce`. The proposed simplification can be simplified even further:

```
More formally, `results[:][result_index] = reduce(windows, init_values, axes(inputs[:]), body)` where:
  * `padded_inputs = pad(inputs[:], init_values[:], padding[:, 0], padding[:, 1], base_dilations)`.
  * `windows = slice(padded_inputs[:], window_start, window_start + window_dimensions, window_dilations)`.
```

This also means that we can drop the "The order of reductions is implementation-defined" paragraph.

burmako (changes requested):

docs/spec_draft.md
```diff
@@ -1268,6 +1269,145 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
+    %0 = stablehlo.add %arg0, %arg1 : tensor<ui8>
+    "stablehlo.return"(%0) : (tensor<ui8>) -> ()
+}) {
+  window_dimensions = dense<[2, 1]> : tensor<2xi64>,
+  window_strides = dense<[4, 1]> : tensor<2xi64>,
+  base_dilations = dense<[2, 1]> : tensor<2xi64>,
+  window_dilations = dense<[3, 1]> : tensor<2xi64>,
+  padding = dense<[[2, 1], [0, 0]]> : tensor<2x2xi64>
+} : (tensor<3x2xui8>, tensor<ui8>) -> tensor<2x2xui8>
+// %result = [[0, 0], [3, 4]]
```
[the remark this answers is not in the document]
sdasgup3: Yup. I meant init_value to be 0.

docs/spec_draft.md
```diff
@@ -3406,6 +3407,98 @@ More formally, `results[:][result_index] = reduce(input_slices)` where:
...
+<img align="center" src="spec_draft/reduce_window.svg" />
+
+The kth dimension of `inputs[k]` is dilated by placing `base_dilations[k] - 1`
+holes between each of the entries in that dimension.  `padding[k][0]` and
+`padding[k][1]` specify the amount of padding added at the low-end (next to
+index 0) and the high-end (next to the highest index) of dimension `k`
+respectively. The amount of padding can be negative, where the absolute value of
+negative padding indicates the number of elements to remove from the specified
+dimension. Dilation of `inputs[k]` occurs logically before padding, so in the
+case of negative padding, elements are removed from the dilated `inputs[k]`.
```
burmako, line 3429: Let's drop this paragraph because: 1) we have an example on the picture above, 2) its contents most duplicate the spec for `pad` which is referenced in the formalism.

burmako (approved):
LGTM except for a minor comment.

PR #360, opened 2022-10-18
Add type inference for stablehlo.convolution

zhouxin913:
This PR  includes:
1. Delete current `verify()` and merge to new `inferReturnTypeComponents()`
2. Prepare the shape function ready to be merged to `TypeInference.cpp` in coming PR. 
3. Double checks verification against the XLA [`shape_inference.cpp`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/shape_inference.cc): All covered.

It does not include:
1. Move to `TypeInference.cpp`: it will make this PR extremely hard to review. Track in https://github.com/openxla/stablehlo/issues/361
2. Adding new logic: the only change is add assert `num_dims = num_spatial_dims + 2` same as XLA, see D1 below.

## Differences with [XLA](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/shape_inference.cc):
- D1.  This PR aligns StableHLO with XLA: revert the support for unknown dimensions inherited from MHLO 

Existing MHLO (and StableHLO) supports unknown dims, but XLA does not: [code](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/shape_inference.cc#L1698) like this:
```
const int num_dims = num_spatial_dims + 2;
if (lhs.rank() != num_dims) { // error out
```
but now in StableHLO the input rank can be larger than num_spatial_dims + 2: see [test](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/ops_stablehlo.mlir#L3731). Note that the "2" means "batch_dim + feature_dim". How to infer the extra dims? The existing [inferConvolutionOpReturnShape](https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/StablehloOps.cpp#L1593) inits every dims directly from the "actually return type" and then do the inferences. 
After discussion, this PR changes StableHLO to that do not support unknown dimensions at all. The unknown dimension in MHLO is only used for CustomCallOp when `hlo.custom_call` is translated back into `mhlo.convolution`, and should be maintained seperately. StableHLO does not need in involve this.

The test is `@conv2d_generic` in `verify_conv.mlir` which is a negative test (which is a positive test in MHLO).

- D2. XLA has [layout check](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/shape_inference.cc#L1710)
```
TF_DCHECK_OK(ShapeUtil::ValidateShapeWithOptionalLayout(lhs));
TF_DCHECK_OK(ShapeUtil::ValidateShapeWithOptionalLayout(rhs));
```
I assume StableHLO have not and do not need it.

- D3. @sdasgup3 Are you aware that why StableHLO/MHLO have checks in both [verify_conv.mlir](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/verify_conv.mlir) and [stablehloOp.mlir](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/ops_stablehlo.mlir) at the same time? 
Sandeep's reply: merge them to 'verify_conv.mlir` and will check them with Spec later.

- D4. XLA has error "Dynamic output feature dim on convolution kernel is not supported" https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/shape_inference.cc#L1886
But StableHLO supports it by design in [test](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/verify_conv.mlir#L746).


sdasgup3:
>> D3. @sdasgup3 Are you aware that why StableHLO/MHLO have checks in both [verify_conv.mlir](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/verify_conv.mlir) and [stablehloOp.mlir](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/ops_stablehlo.mlir) at the same time? I believe the most of them in stablehloOp.mlir can be removed (except the "unknown dims" mentioned above and "quantization" ones). But looks like you maintained them in MHLO, what do you think of it?

The tests in stablehloOps.mlir are already existing and  the new tests are added in verify_conv.mlir after I worked on the verifier for it. AFAIR, I checked that the tests in both the files are exclusive. 

So, for now, you move all of them to the dedicated verify_conv.mlir file. This week I will be working on speccing the convolution op. As part of it I can double check if any any test is redundant and hence to be removed.

stablehlo/dialect/StablehloOps.td
```diff
@@ -1868,7 +1868,8 @@ def StableHLO_CollectivePermuteOp: StableHLO_Op<"collective_permute",
   let hasVerifier = 1;
 }
 
-def StableHLO_ConvolutionOp : StableHLO_Op<"convolution", [NoSideEffect]> {
+def StableHLO_ConvolutionOp : StableHLO_Op<"convolution", 
+    [NoSideEffect, InferTensorType]> {
```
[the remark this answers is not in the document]
zhouxin913: Resolved by rebase and force push: as there's not so many comments on old commits yet.

sdasgup3:
> In this PR I change the initiation part by setting them to "-1". Another possible better approach is loop and fill them as according to the input. Is this a better (and correct) way ?

The extra dimensions (than spatial, feature, and batch) that we have in the input, kernel and output are known as "unknown dimensions". As per some documentation:
```
 the dimensions in ConvDimensionNumbersAttr need not be "dense".
For example, we could have spatial dimensions [0,2], batch dimension [3], and
feature dimension [4].  In this case, dimension [1] is "unknown". 
```

In order to infer the return type of the conv op, in the presence of unknown dimensions, we need to check if we have a mapping between the unknown dimensions in input/kernel and the output. We cannot fill in the dim sizes of "unknown dimensions" with -1 as they do have static sizes as in the [example](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/ops_stablehlo.mlir#L3708) .

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -1419,67 +1419,76 @@ namespace {
...
+                             "and output spatial-dimensions, but got ",
+                             spatialDimNum, ", ",
+                             kernelSpatialDimensions.size(), ", and ",
+                             outputSpatialDimensions.size(), " resp.");
 
   // P2.
-  SmallVector<int64_t> inputDnums(inputSpatialDimensions.size() + 2);
-  inputDnums[0] = op.getDimensionNumbers().getInputBatchDimension();
-  inputDnums[1] = op.getDimensionNumbers().getInputFeatureDimension();
+  SmallVector<int64_t> inputDnums(spatialDimNum + 2);
```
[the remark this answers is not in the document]
sdasgup3: Same for below windowDimNums and OutputDimNums?

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -1589,42 +1623,41 @@ LogicalResult verifyConvolutionAttributes(ConvolutionOp op) {
...
+    int64_t outputFeatureDimension, ArrayRef<int64_t> outputSpatialDimensions,
+    int64_t batchGroupCount, const ArrayRef<hlo::WindowDimension> window) {
+  // We keep the 'unknown' dimensions as it is in the
   // output-shape. To do that we initilize the output dimensions with the shape
   // of the return-type and updates only the spatial + non-spatial dimensions.
   // Precondition 2 ensures that size of output-shape == size of input-shape.
-  SmallVector<int64_t> outputDimensions =
-      to_vector(op.getResult().getType().cast<ShapedType>().getShape());
+  auto lhsType = lhs.getType().cast<RankedTensorType>();
+  SmallVector<int64_t> outputDimensions(lhsType.getShape().size(),
```
sdasgup3, line 1636: Does not seem right as the unknown dimension may not mean dynamic?

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -1589,42 +1623,41 @@ LogicalResult verifyConvolutionAttributes(ConvolutionOp op) {
 //  1. Input args to ConvolutionOp 'op' are RankedTypes.
 //  2. rank-of(input-type) == rank-of(output-type)
 SmallVector<int64_t> inferConvolutionOpReturnShape(
-    ConvolutionOp op, const ArrayRef<hlo::WindowDimension> window) {
-  // We keep the 'unknown' dimensions (cl/415132294) as it is in the
+    Value lhs, Value rhs, int64_t inputBatchDimension,
```
sdasgup3, line 1626: Is there a reason there args are not put one per line?

docs/status.md
```diff
@@ -58,7 +58,7 @@ one of the following tracking labels.
 | concatenate              | yes           | yes          | yes            | yes             | no          |
 | constant                 | yes           | yes          | yes            | yes             | yes         |
 | convert                  | no            | yes*         | infeasible     | yes             | no          |
-| convolution              | no            | revisit      | no             | revisit         | no          |
+| convolution              | no            | yes*         | yes*           | revisit         | no          |
```
sdasgup3, line 61: One point here:
Currently, with client HLO not supporting unknown dimensions and StableHLO supporting it, we cannot say a `yes*` here.

stablehlo/dialect/StablehloOps.cpp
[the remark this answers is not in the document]
zhouxin913: Sure!

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Fixed.

stablehlo/dialect/StablehloOps.cpp
zhouxin913: As discussed offline, we do not plan to support unknown dims in StableHLO, thus these init will all be covered by known dims. Closed this comment.

docs/status.md
zhouxin913: Change to revisit for both.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -4834,37 +4913,26 @@ char nonSpatialDimToString(NonSpatialDim dim) {
...
-          numDims =
-              *std::max_element(spatialDims.begin(), spatialDims.end()) + 1;
-        }
-        for (const auto& dim : nonSpatialDims) {
-          numDims = std::max(numDims, dim.first + 1);
-        }
-
-        llvm::SmallVector<int64_t> dims(numDims, kUnknownDim);
+      [&p](ArrayRef<int64_t> spatial_dims,
+           ArrayRef<std::pair<int64_t, NonSpatialDim>> non_spatial_dims) {
```
sdasgup3, line 4918: naming convention of spatial_dims and non_spatial_dims. Shouldn't they be camel Case.

zhouxin913: Fixed. thanks!

docs/status.md
zhouxin913: Now we decide that StableHLO does not support dynamism either. So I decide to change back to yes*. Please let me know if you have other ideas.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -5025,7 +5025,7 @@ ParseResult parseConvolutionDimensions(AsmParser& parser,
                 llvm::SmallDenseMap<NonSpatialDim, int64_t, 4,
                                     DenseMapInfoNonSpatialDim>>;
 
-  // Note that the allowed_non_spatial_dims is a set (as opposed to unordered
+  // Note that the allowed_non_spatialDims is a set (as opposed to unordered
```
sdasgup3, line 5028: `allowedNonSpatialDims`?

sdasgup3 (approved):

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Fixed thanks!

GleasonK:
Wanted to leave a note here that this change has compatibility implications. "Remove support for" is a buzz-phrase that generally indicates backward incompatibility.

That said - This can be considered an exception, given that it is an unused feature and somewhat of a bug fix given that the StableHLO spec had no intention to inherit this behavior from MHLO.

If there is any disagreement, we can still consider adding backward compatibility for this change.

PR #363, opened 2022-10-19
Clarify the semantics of StableHLO's sort operation.

bchetioui:
The comparator must satisfy the same requirements as C++'s Compare concept, i.e. two elements `v1` and `v2` are considered to be equal by the comparator iff `comparator(v1, v2) = false && comparator(v2, v1) == false`. This precision is important to exclude comparators such as LE or GE, for which setting is_stable to true does not produce a stable sort.


sdasgup3:
Thanks for the suggestion! That makes sense to me.

burmako (approved):

Issue #32, opened 2022-08-22
We need a public stable python API


joker-eph:
Just like #30 and #31, our Python layer relies on ODS and as such as some caveats. We should instead manually write a facade with a nice builder API for StableHLO.

Issue #31, opened 2022-08-22
We need a public stable C API for StableHLO


joker-eph:
C-level APIs is the interoperability mechanism by excellence when one seeks stability. While MLIR has the ability to auto-generate some APIs for dialect, it'll suffer from the same caveats as #30 (changes to TableGen will break it).

We should likely define a stable C API layer to create a StableHLO graph.

Issue #30, opened 2022-08-22
We need a public C++ API for StableHLO


joker-eph:
At the moment, other than the ODS auto-generated builder, there is no API to create a StableHLO graph from C++. This means that a change in MLIR can change the C++ API.

Also there is no "graph" container defined yet.

GleasonK:
Eugene and I are currently looking into compatibility guarantees in (#4). This includes API guarantees which should unblock this ticket along with its related tickets (#31, #32).

GleasonK:
No progress on this issue in the past week. Same rationale as last week, stable APIs are lower priority than a stable serialization format. Will work with Eugene on prioritizing this issue and coming up with a rough timeline.

jpienaar:
This and the Python API issue both came up this week in > 3 different venues.

lsy323:
Looks like there isn't C++ API that is similar to HLO builder API in xla_builder(https://github.com/openxla/xla/blob/main/xla/client/xla_builder.h). Having a similar C++ builder API should make the migration from HLO to StableHLO easier for frameworks currently emits HLO.

It seems StableHLO Python API similar to xla_builder is already there. StableHLO Python API seems like to be code-gen'ed, so I cannot find them in the source code. But here is an example usage in in JAX (https://github.com/google/jax/blob/34d5a6259f150064103fce241ff7d5b54e5ed8b0/jax/_src/prng.py#L1125)

cc @JackCaoG @miladm

janpfeifer:
Also, this would be a nice path for an upgrade for those still using `xla_builder.h`.  Notice that building a graph using `xla_builder.h`, there is a path to convert that to StableHLO (it used to work a few months back) -- so in a way `xla_builder.h` is already a StableHLO C++ API of sorts, but it's convoluted, and I'm not sure which ops and functionality are missing in `xla_builder.h`.

GodRishUniverse:
Interested in knowing if there is any update to this?

GleasonK:
We have made some effort in this space in `stablehlo/integrations/cpp/builder/`:
https://github.com/openxla/stablehlo/blob/main/stablehlo/integrations/cpp/builder/StablehloBuilderTest.cpp

That said we don't have any API compatibility guarantees on these. Mostly best-effort. These were developed to try and abstract away some of the MLIR oddities from folks who aren't as familiar with MLIR. Also worth noting that since the addition of these builders the ergonomics of the upstream generated c++ builders has improved as well.

PR #367, opened 2022-10-19
Add spec for CholeskyOp

### LINKED ISSUE #356: Add spec for CholeskyOp

ghpvnist:
closes #356

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -498,8 +498,8 @@ lower-triangular (if `lower` is `true`) or upper-triangular (if `lower` is
 upper triangle or strict lower triangle correspondingly, are
 implementation-defined.
 
-If there exists `i` where the input matrix is not an Hermitian
-positive-definite matrix, then the result is implementation-defined.
+If there exists `i` where the input matrix is not an Hermitian positive-definite
+matrix, then the behavior is undefined.
 
 ### Inputs
 
@@ -519,7 +519,6 @@ positive-definite matrix, then the result is implementation-defined.
   * (C1) `a` and `result` have the same type.
   * (C2) rank(`a`) >= 2.
   * (C3) dim(`a`, -2) = dim(`a`, -1).
-  * (C4) `lower` is optional with default value `false`.
 
 ### Examples
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -166,6 +166,7 @@ described below)
    * [broadcast_in_dim](#stablehlobroadcast_in_dim)
    * [case](#stablehlocase)
    * [ceil](#stablehloceil)
+   * [cholesky](#stablehlocholesky)
    * [concatenate](#stablehloconcatenate)
    * [constant](#stablehloconstant)
    * [cosine](#stablehlocosine)
@@ -481,6 +482,60 @@ IEEE-754 specification.
 ```
 
 &nbsp;[More Examples](../stablehlo/tests/interpret_ceil.mlir)
+
+[Back to Ops](#index-of-ops)
+
+## stablehlo.cholesky
+
+### Semantics
+
+Computes the Cholesky decomposition of a batch of symmetric (Hermitian) positive
+definite matrices.
+
+If `lower` is `true`, it computes the lower-triangular matrices `l` such that
+$a = l \cdot l^{T}$. If `lower` is `false`, it computdr the upper-triangular
+matrices `u` such that $a = u^{T} \cdot u$.
+
+If the rank of `a` is greater than 2, `a` is treated as a batch of matrices,
+where all except the minor 2 dimensions are batch dimensions.
+
+If `a` is not symmetric (Hermitian) positive definite, the result is
+implementation-defined.
+
+### Inputs
+
+| Name    | Type                                       |
+|---------|--------------------------------------------|
+| `a`     | tensor of floating-point or complex type   |
+| `lower` | 0-dimensional tensor constant of type `i1` |
+
+### Outputs
+
+| Name     | Type                                     |
+|----------|------------------------------------------|
+| `result` | tensor of floating-point or complex type |
+
+### Constraints
+
+  * (C1) `a` and `result` have the same element type.
+  * (C2) rank(`a`) > 2.
+
+### Examples
+
+```mlir
+// %a: [
+//      [1, 2, 3],
+//      [2, 20, 26],
+//      [3, 26, 70]
+//     ]
+// %lower: true
+%result = "stablehlo.cholesky"(%a, %lower) : (tensor<3x3xi32>) -> tensor<3x3xi32>
+// %result: [
+//           [1, 0, 0],
+//           [2, 4, 0],
+//           [3, 5, 6]
+//          ]
+```
 
 [Back to Ops](#index-of-ops)
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -489,18 +489,17 @@ IEEE-754 specification.
 
 ### Semantics
 
-Computes the Cholesky decomposition of a batch of symmetric (Hermitian) positive
-definite matrices.
+Computes the Cholesky decomposition of a batch of matrices.
 
-If `lower` is `true`, it computes the lower-triangular matrices `l` such that
-$a = l \cdot l^{T}$. If `lower` is `false`, it computes the upper-triangular
-matrices `u` such that $a = u^{T} \cdot u$.
-
-If the rank of `a` is greater than 2, all except the minor 2 dimensions of `a`
-are treated as batch dimensions.
-
-If `a` is not symmetric (Hermitian) positive definite, the result is
+More formally, for all `i`, `result[i0, ..., iR-3, :, :]` is a Cholesky
+decomposition of `a[i0, ..., iR-3, :, :]`, in the form of either of a
+lower-triangular (if `lower` is `true`) or upper-triangular (if `lower` is
+`false`) matrix. The output values in the opposite triangle, i.e. the strict
+upper triangle or strict lower triangle correspondingly, are
 implementation-defined.
+
+If there exists `i` where the input matrix is not an Hermitian
+positive-definite matrix, then the result is implementation-defined.
 
 ### Inputs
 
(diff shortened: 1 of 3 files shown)


docs/spec_draft.md
```diff
@@ -483,6 +484,59 @@ IEEE-754 specification.
...
+### Outputs
+
+| Name     | Type                                     |
+|----------|------------------------------------------|
+| `result` | tensor of floating-point or complex type |
+
+### Constraints
+
+  * (C1) `a` and `result` have the same element type.
+  * (C2) rank(`a`) > 2.
```
sdasgup3, line 520: Add this one as well

The two minor dimensions of 'a' must have equal size, or
`dim(a, R-1) == dim(a, R-2), R = rank(a)`. If not, then  the result is
implementation-defined.

docs/spec_draft.md
```diff
@@ -483,6 +484,59 @@ IEEE-754 specification.
...
+definite matrices.
+
+If `lower` is `true`, it computes the lower-triangular matrices `l` such that
+$a = l \cdot l^{T}$. If `lower` is `false`, it computes the upper-triangular
+matrices `u` such that $a = u^{T} \cdot u$.
+
+If the rank of `a` is greater than 2, all except the minor 2 dimensions of `a`
+are treated as batch dimensions.
+
+If `a` is not symmetric (Hermitian) positive definite, the result is
```
sdasgup3, line 501: drop this as it is already covered in the constraints.

docs/spec_draft.md
```diff
@@ -483,6 +484,59 @@ IEEE-754 specification.
...
 
+## stablehlo.cholesky
+
+### Semantics
+
+Computes the Cholesky decomposition of a batch of symmetric (Hermitian) positive
+definite matrices.
+
+If `lower` is `true`, it computes the lower-triangular matrices `l` such that
+$a = l \cdot l^{T}$. If `lower` is `false`, it computes the upper-triangular
```
sdasgup3, line 495: In the HLO spec, it it says 
$a = l \cdot l^{T}$
with `l` being a matrix. I believe the assumption is that `a` is also a matrix. 

We can define the similar formula with `l` being a tensor with R (= rank) >=2 using the StableHLO versions of the `transpose` and `dot_general` ops.

$l^{T}$ could defined as `stabhlehlo.transpose` on `l` with `permutations = [0, 1, ..., R-1, R-2]`.

$l \cdot l^{T}$, is defined as `stablehlo.dot_general` with
 - `lhs_batching_dims = rhs_batching_dims = {i : 0 <= i < R-2}`
 -  `lhs_contracting_dims = [R-1]`
 -  `rhs_contracting_dims = [R-2]`

ghpvnist: Yes, that is a correct intuition to derive `a` from `result`.

docs/spec_draft.md
```diff
@@ -483,6 +484,59 @@ IEEE-754 specification.
...
+matrices `u` such that $a = u^{T} \cdot u$.
+
+If the rank of `a` is greater than 2, all except the minor 2 dimensions of `a`
+are treated as batch dimensions.
+
+If `a` is not symmetric (Hermitian) positive definite, the result is
+implementation-defined.
```
burmako, line 502: How about replacing the Semantics section as follows:

```
Computes the Cholesky decomposition of a batch of matrices.

More formally, for all `i`, `result[i0, ..., iR-3, :, :]` is a Cholesky 
decomposition of `a[i0, ..., iR-3, :, :]`, in the form of either of a 
lower-triangular (if `lower` is `true`) or upper-triangular (if `lower` is
`false`) matrix. The output values in the opposite triangle, i.e. the strict
upper triangle or strict lower triangle correspondingly, are 
implementation-defined.

If there exists `i` where the input matrix is not an Hermitian
positive-definite matrix, then the result is implementation-defined.
```

Feedback:
  1. Cholesky decomposition already implies Hermitian, positive-definite matrices, so there's no need to repeat that in the first sentence. As discussed in earlier PRs, we can assume the knowledge of high school / college maths in this spec.
  2. "If the rank of `a` is greater than 2, all except the minor 2 dimensions of `a` are treated as batch dimensions" is okay for building intuition, but it's kind of already there thanks to the first sentence.
  3. In addition to intuition, we need formal notation, to be consistent with how we specced other ops.
  4. The current version of the spec seems to have missed the "Output data is returned in the same triangle; the values in the other triangle are implementation-defined and may be anything" part of https://www.tensorflow.org/xla/operation_semantics#cholesky.

docs/spec_draft.md
burmako: Agreed about `(C3) dim(a, R-1) == dim(a, R-2), where R = rank(a)`. 

As far as "If not, then the result is implementation-defined" goes, I don't think that's right. If not, then the op fails verification, and I don't think that implementations can have opinions about this.

burmako, line 520: "(C2) `rank(a) >= 2`". 2 is also a valid rank - in that case, there won't be any batch dimensions, but that's fine.

docs/spec_draft.md
```diff
@@ -483,6 +484,59 @@ IEEE-754 specification.
...
+
+### Outputs
+
+| Name     | Type                                     |
+|----------|------------------------------------------|
+| `result` | tensor of floating-point or complex type |
+
+### Constraints
+
+  * (C1) `a` and `result` have the same element type.
```
burmako, line 519: "...have the same type". The shapes are also the same.

docs/spec_draft.md
```diff
@@ -483,6 +484,59 @@ IEEE-754 specification.
...
+
+### Examples
+
+```mlir
+// %a: [
+//      [1.0, 2.0, 3.0],
+//      [2.0, 20.0, 26.0],
+//      [3.0, 26.0, 70.0]
+//     ]
+%result = "stablehlo.cholesky"(%a) {lower=true} : (tensor<3x3xf32>) -> tensor<3x3xf32>
```
burmako, line 530: `{ lower = true }` is the canonical MLIR syntax.

burmako (changes requested):

docs/spec_draft.md
burmako: Actually, how about we use the notation from FFT, and reframe this as "* (C3) `dim(a, -2) = dim(a, -1)`"?

ghpvnist: Thanks for the catch!

ghpvnist: Done.

docs/spec_draft.md
ghpvnist: Thanks for clearing this up.

docs/spec_draft.md
```diff
@@ -483,6 +484,61 @@ IEEE-754 specification.
...
+
+If there exists `i` where the input matrix is not an Hermitian
+positive-definite matrix, then the result is implementation-defined.
+
+### Inputs
+
+| Name    | Type                                       |
+|---------|--------------------------------------------|
+| `a`     | tensor of floating-point or complex type   |
+| `lower` | 0-dimensional tensor constant of type `i1` |
```
sdasgup3, line 508: `lower` being a optional attribute, how about calling it in constants section  like
* (C3) `lower` is optional with default value `false`.

sdasgup3 (approved):

docs/spec_draft.md
ghpvnist: This is the XLA's implementation: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/cholesky_expander.cc

For a non-Hermitian matrix:
`a = [[1, 2], [3, 4]]`
The result is:
`result = [[1, 2], [0, -nan]]`

We are getting some deterministic interpreted calculation of this algorithm (Specifically Cholesky–Banachiewicz as noted in the comments to the link above). Though the result is not defined/interpretable for a non-Hermitian positive definite matrix, it is still giving an implementation defined result. Wdyt? @burmako

burmako: "it is still giving an implementation defined result". We had an actionable conversation with Mehdi about the meaning of "implementation-defined behavior" and "undefined behavior" in https://github.com/openxla/stablehlo/pull/344#discussion_r1002609416.

In a nutshell, based on this conversation, I think that we should prefer UB over implementation-defined behavior (I'll refer to it as IB in this comment, but this is not a standard acronym) whenever we can. 

UB means that we are disallowing certain classes of programs, whereas IB means that we are allowing these programs but aren't taking a stance on their behavior. IB is very hard or impossible to walk back - once there are StableHLO programs in the wild, we cannot do much about them. As a result, if both UB and IB are an option, and we're not sure which one to pick, we should pick UB.

Now, if some behavior is legitimately allowed in StableHLO programs but may be different across various implementations, then UB is not an option, and IB might be an appropriate solution. E.g. back then, we decided that precision for many floating-point ops is implementation-defined - that's an appropriate way of using IB (UB doesn't even apply here, so it's not an option). Similarly, overflow for many integer ops is implementation-defined - that's also IB, because this is something that a lot of existing programs rely on, so we cannot make that UB.

As a result, here's the recommended rewording of the subsequent paragraph.

```
If there exists `i` where the input matrix is not an Hermitian
positive-definite matrix, then the behavior is undefined.
```

docs/spec_draft.md
```diff
@@ -484,6 +485,62 @@ IEEE-754 specification.
...
+| Name     | Type                                     |
+|----------|------------------------------------------|
+| `result` | tensor of floating-point or complex type |
+
+### Constraints
+
+  * (C1) `a` and `result` have the same type.
+  * (C2) rank(`a`) >= 2.
+  * (C3) dim(`a`, -2) = dim(`a`, -1).
+  * (C4) `lower` is optional with default value `false`.
```
burmako, line 522: I think we had a PR review in September where we decided to not bring up default values in the spec, but I cannot recall when or where exactly that was.

To recount the rationale from back then, I think that default values are an implementation detail of the StableHLO API, so I don't think the spec should be prescriptive about then, especially since we don't yet know how the API is going to look like (see #30, #31 and #32 which we deprioritized for the time being).

I'm sure that we'll figure this out later, and maybe default values will make a resurgence in the spec, but for now let's not include them.

burmako (changes requested):

burmako (approved):

PR #374, opened 2022-10-20
Migrate all uses of -1 to kDynamicSize

### LINKED ISSUE #327: Migrate all uses of -1 to kDynamicSize

ghpvnist:
The process for replacing uses of -1 was:

1. Find all occurrences of `-1` (and `- 1`).
2. Read the context of the code to determine whether the use of `-1` (or `- 1`) is appropriate.
3. Replace `-1` with `ShapedType::kDynamicSize` where appropriate.

closes #327

### DIFF
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -4797,7 +4797,7 @@ ParseResult parseConvolutionDimensions(AsmParser& parser,
     }
 
     llvm::SmallDenseMap<int64_t, int64_t> spatialDimsMap;
-    constexpr int64_t kInvalidDimension = -1;
+    constexpr int64_t kInvalidDimension = ShapedType::kDynamicSize;
     // Keep track of the maximum spatial dimension parsed as we expect to see
     // all the dimensions from 0 to maximum dimension parsed.
     int64_t maxParsedSpatialDim = kInvalidDimension;
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -898,7 +898,7 @@ LogicalResult inferReduceOp(
   uint64_t numInputs = inputs.size();
 
   // Check for unranked tensors in input operands.
-  int64_t rankedInputIdx = -1;
+  int64_t rankedInputIdx = ShapedType::kDynamicSize;
   for (uint64_t inputIdx = 0; inputIdx < numInputs; ++inputIdx) {
     if (inputArgTypes[inputIdx].hasRank()) {
       rankedInputIdx = inputIdx;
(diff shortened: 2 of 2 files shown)


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -4797,7 +4797,7 @@ ParseResult parseConvolutionDimensions(AsmParser& parser,
     }
 
     llvm::SmallDenseMap<int64_t, int64_t> spatialDimsMap;
-    constexpr int64_t kInvalidDimension = -1;
```
burmako, line 4800: It looks like in all the instances covered in this PR, `-1` is used as a legit sentinel value for a specific algorithm (axis number here and input number below). I think that it makes all of them out of scope for the `kDynamicSize` change.

The motivation behind #327 is to fix all the situations where `-1` is used as a dimension size to indicate that the size is dynamic. In those situations, `kDynamicSize` should be used instead to make code robust (because `kDynamicSize` is going to change to a different value soon, and all the uses of `-1` in that capacity will become incorrect).

In other situations (like the ones in this PR), I think we should continue using `-1`. E.g. I would be surprised if something called `kInvalidDimension` was equal to `kDynamicSize`, because the former is a dimension number and the latter is a dimension size.

What do you think?

burmako (changes requested):

stablehlo/dialect/StablehloOps.cpp
ghpvnist: Understood now. The other notable places that use `-1` come from td files (some comments use them, which we will most likely overwrite when we get to those ops, at least in StablehloOps.td, and a few other similar occurrences in ChloOps.td) and mlir files (where the `?` dimension size is hard coded as -1).

Then based on my understanding, I should be changing the mlir files containing the -1 dimension size, but I'm not sure how I can turn this into kDynamicSize. I might need a bit more context to resolving this issue. Or should I be looking somewhere else entirely?

burmako, line 4800: As discussed in the team meeting, can you also check for `< 0`. Also, we'll sync separately on what to do with MLIR files. It appears that the following two changes will be sufficient: 1) prettyprinter for TypeExtensions, 2) shape inference testing infra to prettyprint -1 as ?.

burmako (changes requested):

burmako:
We discussed with @ghpvnist that I will be continuing this work and will make the changes called out in the comments above. I've assigned the ticket to myself and will be closing this PR for now.

Issue #154, opened 2022-09-19
Unify verifier and type inference errors


zhouxin913:
### Request description

In verifiers and shape functions of StableHLO / MHLO, different ops have inconsistent error messages. Consider wrap them with macros and/or utils functions.

zhouxin913:
Unassigned as low priority.

burmako:
To provide more context: both #63 and this ticket highlight an important area for future work - improving error messages in various parts of StableHLO, with the focus on consistency with existing infrastructure (upstream verifier and parser errors).

We definitely want to address that eventually, but right now there are bigger gaps in the implementation of the dialect - namely, completeness of verifiers & shape functions. As a result, we've unassigned #63 and this tickets for now, but we'll be getting back to them in the future.

zhouxin913:
This bug is partially address by https://github.com/openxla/stablehlo/issues/270, which unifies the code of how error emits with static `emitOptionalError()`.
But there's will more to do: address the message strings themselves.

PR #375, opened 2022-10-21
Handle bounds in the PadOp shape function

zhouxin913:
Add bounds support for `PadOp`
Why change the shape function interface from `inferReturnTypeComponents` to `inferReturnTypes`? Appending `inferredReturnShapes.emplace_back(..., encoding_attr)` inside `inferReturnTypeComponents` will trigger error in [upstream](https://github.com/llvm/llvm-project/blob/main/mlir/lib/Interfaces/InferTypeOpInterface.cpp#L190)
```
assert(shapeAndType.getAttribute() == nullptr && "attribute not supported");
```

This PR follow the verification of in dim loop similar to https://github.com/openxla/stablehlo/pull/378/files.

Also add @smit-hinsu as reviewer.

### DIFF
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -3363,15 +3363,15 @@ LogicalResult SetDimensionSizeOp::inferReturnTypes(
 // PadOp
 //===----------------------------------------------------------------------===//
 
-LogicalResult PadOp::inferReturnTypeComponents(
-    MLIRContext*, Optional<Location> location, ValueShapeRange operands,
+LogicalResult PadOp::inferReturnTypes(
+    MLIRContext*, Optional<Location> location, ValueRange operands,
     DictionaryAttr attributes, RegionRange regions,
-    SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
+    SmallVectorImpl<Type>& inferredReturnTypes) {
   PadOp::Adaptor adaptor(operands, attributes, regions);
   return hlo::inferPadOp(location, adaptor.getOperand(),
                          adaptor.getPaddingValue(), adaptor.getEdgePaddingLow(),
                          adaptor.getEdgePaddingHigh(),
-                         adaptor.getInteriorPadding(), inferredReturnShapes);
+                         adaptor.getInteriorPadding(), inferredReturnTypes);
 }
 
 LogicalResult PadOp::reifyReturnTypeShapes(
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -414,6 +414,17 @@ LogicalResult verifyReducerShape(
   return success();
 }
 
+// Returns output dimension size for slice result for the given arguments.
+// Returns ShapedType::kDynamicSize if arguments are illegal.
+static int64_t inferSliceDim(int64_t inputDim, int64_t start, int64_t end,
+                             int64_t stride) {
+  if (inputDim == ShapedType::kDynamicSize || start < 0 || start > end ||
+      end > inputDim || stride == 0)
+    return ShapedType::kDynamicSize;
+
+  return llvm::divideCeil(end - start, stride);
+}
+
 //===----------------------------------------------------------------------===//
 // Shape functions for ops.
 //===----------------------------------------------------------------------===//
@@ -878,9 +889,15 @@ LogicalResult inferPadOp(Optional<Location> location, Value operand,
 
       if (!inputBounds.empty())
         resultBounds.push_back(ShapedType::kDynamicSize);
-    } else  // dynamic dim with bounds
-      resultBounds.push_back(inferPadDim(inputBounds[i], paddingLowVal,
-                                         paddingHighVal, paddingInteriorVal));
+    } else {  // dynamic dim with bounds
+      int64_t resultBound = inferPadDim(inputBounds[i], paddingLowVal,
+                                        paddingHighVal, paddingInteriorVal);
+      if (resultBound < 0 && resultBound != ShapedType::kDynamicSize)
+        return emitOptionalError(
+            location, "Padding result in negative size for bound of dimension ",
+            i);
+      resultBounds.push_back(resultBound);
+    }
   }
   inferredReturnTypes.push_back(RankedTensorType::get(
       resultShape, inputType.getElementType(),
@@ -1124,26 +1141,20 @@ LogicalResult inferSliceOp(Optional<Location> location, Value operand,
   SmallVector<int64_t, 4> shape;
   shape.reserve(rank);
   SmallVector<int64_t> resultBounds;
-
-  auto inferSliceDim = [&](int64_t inputDim, int64_t start, int64_t end,
-                           int64_t stride) {
-    return inputDim == ShapedType::kDynamicSize
-               ? ShapedType::kDynamicSize
-               : static_cast<int64_t>(llvm::divideCeil(end - start, stride));
-  };
-
   for (int64_t i = 0, e = rank; i != e; i++) {
+    if (!inputBounds.empty())
+      resultBounds.push_back(
+          inferSliceDim(inputBounds[i], start[i], limit[i], strideVals[i]));
     if (hlo::isDynamicDimSize(rankedTy.getDimSize(i))) {
       shape.push_back(ShapedType::kDynamicSize);
-      if (inputBounds.empty()) continue;  // dynamic dim without bounds
+      continue;
     }
     // P3.
     if (start[i] < 0)
       return emitOptionalError(location, "negative start index ", start[i],
                                " in dimension ", i);
     // P4.
-    if (!hlo::isDynamicDimSize(rankedTy.getDimSize(i)) &&
-        limit[i] > rankedTy.getDimSize(i))
+    if (limit[i] > rankedTy.getDimSize(i))
       return emitOptionalError(location, "limit index ", limit[i],
                                " is larger than dimension size ",
                                rankedTy.getDimSize(i), " in dimension ", i);
@@ -1157,14 +1168,8 @@ LogicalResult inferSliceOp(Optional<Location> location, Value operand,
       return emitOptionalError(location, "stride must be positive but got ",
                                strideVals[i], " in dimension ", i);
 
-    if (!hlo::isDynamicDimSize(rankedTy.getDimSize(i))) {  // static dim
-      shape.push_back(inferSliceDim(rankedTy.getDimSize(i), start[i], limit[i],
-                                    strideVals[i]));
-      if (!inputBounds.empty())
-        resultBounds.push_back(ShapedType::kDynamicSize);
-    } else  // dynamic dim with bounds
-      resultBounds.push_back(
-          inferSliceDim(inputBounds[i], start[i], limit[i], strideVals[i]));
+    shape.push_back(inferSliceDim(rankedTy.getDimSize(i), start[i], limit[i],
+                                  strideVals[i]));
   }
 
   inferredReturnTypes.push_back(RankedTensorType::get(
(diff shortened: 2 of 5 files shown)


stablehlo/dialect/TypeInference.cpp
```diff
@@ -1114,20 +1124,26 @@ LogicalResult inferSliceOp(Optional<Location> location, Value operand,
...
-    if (!inputBounds.empty())
-      resultBounds.push_back(
-          inferSliceDim(inputBounds[i], start[i], limit[i], strideVals[i]));
     if (hlo::isDynamicDimSize(rankedTy.getDimSize(i))) {
       shape.push_back(ShapedType::kDynamicSize);
-      continue;
+      if (inputBounds.empty()) continue;  // dynamic dim without bounds
     }
     // P3.
     if (start[i] < 0)
```
smit-hinsu, line 1141: Good find! It is always better to verify these attributes even if they are not used?

Also, it seems slice op change is completely unrelated to pad inference.  If that is correct, let's spin up another PR to keep the changes focused?

https://google.github.io/eng-practices/review/developer/small-cls.html

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -3363,15 +3363,15 @@ LogicalResult SetDimensionSizeOp::inferReturnTypes(
 // PadOp
 //===----------------------------------------------------------------------===//
 
-LogicalResult PadOp::inferReturnTypeComponents(
-    MLIRContext*, Optional<Location> location, ValueShapeRange operands,
+LogicalResult PadOp::inferReturnTypes(
```
smit-hinsu, line 3366: If the assert is the only issue, the correct way would be to remove that upstream and just use the attribute in building RankedTensorType. It wasn't used so far and therefore the assert was added. 

I am okay to make progress in parallel or simply wait on that to be fixed. Not sure how long it takes for StableHLO LLVM upgrade.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -866,15 +867,24 @@ LogicalResult inferPadOp(
...
+                                           paddingHighVal, paddingInteriorVal);
+      if (expectedOutput < 0)
+        return emitOptionalError(
+            location, "Padding result in negative size for dimension ", i);
+      resultShape.push_back(expectedOutput);
+
+      if (!inputBounds.empty())
+        resultBounds.push_back(ShapedType::kDynamicSize);
+    } else  // dynamic dim with bounds
+      resultBounds.push_back(inferPadDim(inputBounds[i], paddingLowVal,
```
smit-hinsu, line 882: add check for negative on bounds as well?

stablehlo/dialect/TypeInference.cpp
```diff
@@ -866,15 +867,24 @@ LogicalResult inferPadOp(
...
-    if (expectedOutput < 0)
-      return emitOptionalError(
-          location, "Padding result in negative size for dimension ", i);
-    resultShape.push_back(expectedOutput);
+
+    if (!hlo::isDynamicDimSize(inputShape[i])) {  // static dim
+      int64_t expectedOutput = inferPadDim(inputShape[i], paddingLowVal,
+                                           paddingHighVal, paddingInteriorVal);
+      if (expectedOutput < 0)
+        return emitOptionalError(
```
smit-hinsu, line 875: curious if there is some way or plan to share this verifications with op verifiers with some helpers?

smit-hinsu (commented):
Thanks!

This PR looks good to me but have related question/comments.

stablehlo/dialect/StablehloOps.cpp
burmako: Addressing this upstream is a good thing to do, but to me it looks like a lower priority than making progress with shape functions, given that they are needed both for the TF/XLA bridge and JAX, whereas inferReturnTypes vs inferReturnTypeComponents doesn't seem to make much difference at this point.

zhouxin913: Like Eugene states, we do not have strong preference to use which interface and there's no reason to be blocked by that in any way. So let's move on with this. And only consider fix the upstream when we have a stronger or higher priority motivation.

stablehlo/dialect/TypeInference.cpp
zhouxin913: Small PRs: totally agreed. Will send the other one soon later.
"verify these attributes even if they are not used": No, I do not intent to restrict the verifications in any way, with the concern that it may break MHLO further. I mean, for "unbounded dynamic" case, we did not check the attrs, and I keep it as-is in this PR (simply continue the loop). But for bounded dynamic case, if the attrs are used to calculate the result_bound, then I suggest to check them, like we check them for "static dim case.

stablehlo/dialect/TypeInference.cpp
zhouxin913: Sure! I was thinking of this once, and glad you point it out! Added.

stablehlo/dialect/TypeInference.cpp
zhouxin913: Not quite follow what you meant here? Please leave comments if the sentences above do not answer your question:
1. For error message level sharing: it is tracked in https://github.com/openxla/stablehlo/issues/154 but we do not have cycles to work on it yet
2. This `expectedOutput < 0` is very specific op logic so I guess no way to do it in ODS?
3. `Op.verify()` are expected to be removed and merged to shape functions (thus shared with MHLO)

stablehlo/dialect/TypeInference.cpp
smit-hinsu: Thanks! yes, the other PR has checks in the front and seems safer.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -851,10 +852,21 @@ LogicalResult inferPadOp(
...
+  auto inferPadDim = [&](int64_t inputDim, int64_t paddingLowVal,
+                         int64_t paddingHighVal, int64_t paddingInteriorVal) {
+    return inputDim == ShapedType::kDynamicSize
+               ? ShapedType::kDynamicSize
+               : inputDim + paddingLowVal + paddingHighVal +
+                     std::max<int64_t>(inputDim - 1, 0LL) * paddingInteriorVal;
+  };
+
   for (int i = 0, e = inputShape.size(); i < e; i++) {
     if (hlo::isDynamicDimSize(inputShape[i])) {
```
smit-hinsu, line 867: similar to the Slice op, move this logic after we check that paddingInteriorVal is not negative. That way, even if input dim is dynamic and bounds are missing, we report an error if paddingInteriorVal is negative.

smit-hinsu (approved):
Looks good once these comments are resolved!

stablehlo/dialect/TypeInference.cpp
zhouxin913: Oh, the link you posted is of a different op `DynamicPadOp` right? that op will be taken care of in the future. 
This PR is about `PadOp` and I believe there's no `PadOp::verify()` any more (deleted a while ago).

stablehlo/dialect/TypeInference.cpp
```diff
@@ -866,15 +878,30 @@ LogicalResult inferPadOp(
...
+        resultBounds.push_back(ShapedType::kDynamicSize);
+    } else {  // dynamic dim with bounds
+      int64_t resultBound = inferPadDim(inputBounds[i], paddingLowVal,
+                                        paddingHighVal, paddingInteriorVal);
+      if (resultBound < 0 && resultBound != ShapedType::kDynamicSize)
+        return emitOptionalError(
+            location, "Padding result in negative size for bound of dimension ",
+            i);
+      resultBounds.push_back(resultBound);
+    }
```
burmako, line 900: I think it's a great idea to share logic between the cases which compute results for static dimension sizes and for bounds of dynamic dimension sizes. However, I think that the amount of code in this PR to implement this is fairly excessive.

Here's the code that we want to share:

```C++
int64_t expectedOutput =
    inputShape[i] + paddingLowVal + paddingHighVal +
    std::max<int64_t>(inputShape[i] - 1, 0LL) * paddingInteriorVal;
if (expectedOutput < 0)
  return emitOptionalError(
      location, "Padding result in negative size for dimension ", i);
resultShape.push_back(expectedOutput);
```

And here's how it can be rewritten, without needing additional lambdas and ifs:

```C++
// Assuming that `bool isDynamicDim = hlo::isDynamicDimSize(inputShape[i]);` is declared above.
int64_t operandSizeOrBound = isDynamicDim ? inputBounds[i]: inputShape[i];
int64_t resultSizeOrBound =
    operandSizeOrBound + paddingLowVal + paddingHighVal +
    std::max<int64_t>(inputShape[i] - 1, 0LL) * paddingInteriorVal;
if (resultSizeOrBound < 0) {
  auto what = isDynamicDim ? "bound" : "size";
  return emitOptionalError(
      location, "Padding result in negative ", what, " for dimension ", i);
}
(isDynamicDim ? resultBounds : resultShape).push_back(expectedOutput);
```

burmako (changes requested):

stablehlo/dialect/TypeInference.cpp
zhouxin913: The new commit is a lot simplified by the comment above and the long discussion result from https://github.com/openxla/stablehlo/pull/378.  The keys are:
1. Reuse code for `isStaticDim` case and `isStaticBound` case
2. Init the result_shape and result_bounds with (len, ShapedType::kDynamicSize) which avoid a lot of "appending"s in complicated different cases.

stablehlo/dialect/TypeInference.cpp
zhouxin913: Sure, and the new code move verification in proper places: move as much up as possible.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -866,15 +878,30 @@ LogicalResult inferPadOp(
...
+        return emitOptionalError(
+            location, "Padding result in negative size for dimension ", i);
+      resultShape.push_back(expectedOutput);
+
+      if (!inputBounds.empty())
+        resultBounds.push_back(ShapedType::kDynamicSize);
+    } else {  // dynamic dim with bounds
+      int64_t resultBound = inferPadDim(inputBounds[i], paddingLowVal,
+                                        paddingHighVal, paddingInteriorVal);
+      if (resultBound < 0 && resultBound != ShapedType::kDynamicSize)
```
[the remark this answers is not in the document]
zhouxin913: This is indeed an issue: the old commit can't tell whether the -1 comes from an originally -1 or a calculated -1. Note that for bounded dynamic case, we still can have -1 for some dynamic dims. But NVM, the new commit resolves this issue by new constructed code.

smit-hinsu (approved):

stablehlo/tests/infer_stablehlo.mlir
```diff
@@ -657,3 +656,30 @@ func.func @slice_with_bounds(%arg0: tensor<3x?x?xi32, #stablehlo.type_extensions
...
 }
+
+// -----
+
+// CHECK-LABEL: @pad_with_bounds
+func.func @pad_with_bounds(%arg0: tensor<3x?x?xf16, #stablehlo.type_extensions<bounds = [-1, 3, -1]>>, %arg1: tensor<f16>) -> tensor<*xindex> {
+  %0 = "stablehlo.pad"(%arg0, %arg1) {
+    edge_padding_high = dense<[0, 0, 0]> : tensor<3xi64>,
+    edge_padding_low = dense<[2, 2, 0]> : tensor<3xi64>,
+    interior_padding = dense<[1, 1, 1]> : tensor<3xi64>
```
burmako, line 667: Let's reorder attributes, so that they come in the order they are declared in the ODS: low, high and then interior.

burmako (approved):

stablehlo/tests/infer_stablehlo.mlir
zhouxin913: Fixed!

Issue #369, opened 2022-10-19
Unifying the element type inference for ops involving reduction operation


sdasgup3:
### Request description

We found a few disparities in inferring the return element type of operations involving reduction.

Lets first see the constraints we have on reducer functions type (from https://github.com/openxla/stablehlo/blob/8ec92006166d0602d0dca32e0267d169c2078e0d/stablehlo/dialect/TypeInference.cpp#L273 or 
https://github.com/tensorflow/tensorflow/blob/b362eaeb53cc21aa38f3eb1551500ced4fca0c97/tensorflow/compiler/xla/service/shape_inference.cc#L62, the former is inspired from the later) 

Consider typical reduce-* op syntax:

```
op(I(i), V(j)):
  block(BI(i), BV(j)):
      ... some computation ...
  return(R(i))

I(i)  : i-th input of op
V(j)  : j-th init-value of op
BI(i) : i-th input of reducer-function
BV(j) : j-th init-value of reducer-function
R(i)  : i-th return-type
```

The constraints that we verify are:
 1. Check that BI(i) and R(i) have same shape and element-type.
 2. Check that BV(j) and R(i) have same shape and element-type, ignoring fp-precision.
 3. Check that V(j) and R(i) have same shape and element-type, ignoring fp-precision.
 4. Check that I(i) and BV(j) have same element-type, ignoring fp-precision.
 
Now the return type of the followings ops are inferred as follows:

| op | Return element type inference |
|-----|------------------------------|
| reduce | Same as element type of R(i)  [link](https://github.com/tensorflow/tensorflow/blob/b362eaeb53cc21aa38f3eb1551500ced4fca0c97/tensorflow/compiler/xla/service/shape_inference.cc#L2319).  | 
| reduce_window | Same as element type of V(i) [link](https://github.com/tensorflow/tensorflow/blob/b362eaeb53cc21aa38f3eb1551500ced4fca0c97/tensorflow/compiler/xla/service/shape_inference.cc#L2381) |
| select_and_scatter | Same as element type of I(i) ignoring fp precision [link](https://github.com/tensorflow/tensorflow/blob/b362eaeb53cc21aa38f3eb1551500ced4fca0c97/tensorflow/compiler/xla/service/shape_inference.cc#L2431)

From 1-4, it seems like for all the ops. we can say the element type of return is "element type of I(i) ignoring fp precision ".

The proposal here is the StableHLO spec to follow the same behavior as in HLO.

[...]

Issue #212, opened 2022-09-28
Add type inference for stablehlo.select_and_scatter.


zhouxin913:
### Request description

Follow `docs/type_inference.md`.

zhouxin913:
Still WIP and expect to finish EOW.

zhouxin913:
Delayed and expect to complete in coming 2 weeks.

PR #376, opened 2022-10-21
Add spec for SelectAndScatterOp

### LINKED ISSUE #302: Add spec for SelectAndScatterOp
### Request description

To spec stablehlo.select_and_scatter

sdasgup3:
fixes #302 

Verification is "revisit" because of #369.
Type inference is `no` based on https://github.com/openxla/stablehlo/issues/212


docs/spec_draft.md
```diff
@@ -1443,6 +1444,138 @@ specification. Numeric precision is implementation-defined.
...
+selected locations whose shape matches `source`. These are then scattered to the
+output which is initialized with `init_value`.  Multiple scattered elements
+which land in the same output location are combined using the `scatter`
+function.
+
+The following diagram shows how elements in `results` are computed from
+`operand` and `source` using a concrete example.
+
+<p align="center">
+  <img src="spec_draft/select_and_scatter.png" />
```
ghpvnist, line 1462: Need to update the image's `padding` to be 1-dimensional.

ghpvnist, line 1462: result shape 5x2, but the image shape is 4x2?

ghpvnist, line 1462: Maximum of blue window is 6, and maximum of pink window is 8, so shouldn't the output result be (i.e. no summing of 6+8):
[5, 6]
[7, 8]

ghpvnist (changes requested):

docs/spec_draft.md
```diff
@@ -1443,6 +1444,138 @@ specification. Numeric precision is implementation-defined.
...
+### Inputs
+
+| Name                | Type                                      | Constraints      |
+|---------------------|-------------------------------------------|------------------|
+| `operand`           | tensor of any supported type              | (C1-C10)         |
+| `source`            | tensor of any supported type              | (C1-C3)          |
+| `init_value`        | scalar of any supported type              | (C4), (C9)       |
+| `window_dimensions` | 1-dimensional tensor constant type `si64` | (C1), (C3), (C5) |
+| `window_strides`    | 1-dimensional tensor constant type `si64` | (C6)             |
+| `padding`           | 2-dimensional tensor constant type `si64` | (C3), (C7)       |
```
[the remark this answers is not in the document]
ghpvnist: You can ignore this comment.

docs/spec_draft.md
```diff
@@ -1443,6 +1444,138 @@ specification. Numeric precision is implementation-defined.
...
+
+Formally,
+ * Apply [stablehlo.reduce_window](#stablehloreduce_window) with the following
+   parameters to get `selected_values`.
+   * `inputs` $=$ `{ operand }`,
+   * `init_values` = {identity value for `select`}.
+   * `window_dimensions`, `window_strides`, and `padding` are used as is.
+   * `body` $=$ `select`, to select an element from each window by applying it
+     across each window.
+ * Let `selected_indices = {index-of(value, operand: value in selected_values}`,
```
ghpvnist, line 1474: Typo: Missing closing parenthesis near index-of

docs/spec_draft.md
```diff
@@ -1443,6 +1444,138 @@ specification. Numeric precision is implementation-defined.
...
+
+### Outputs
+
+| Name     | Type                         | Constraints |
+|----------|------------------------------|-------------|
+| `result` | tensor of any supported type | (C9-C10)    |
+
+### Constraints
+
+  * (C1) rank(`source`) $=$ rank(`operand`) $=$ size(`window_dimensions`).
```
ghpvnist, line 1509: Nit:
```
rank(`operand`) $=$ rank(`source`) $=$ size(`window_dimensions`).
```

docs/spec_draft.md
```diff
@@ -1443,6 +1444,138 @@ specification. Numeric precision is implementation-defined.
...
+
+  * (C2) `operand` and `source` have the same element type ignoring floating-point
```
ghpvnist, line 1511: Do we need the "ignoring floating-point precision" part as part of the constraint?

ghpvnist (changes requested):

docs/spec_draft.md
```diff
@@ -1443,6 +1444,138 @@ specification. Numeric precision is implementation-defined.
...
+        element type same as that of 'operand'. It is called with the property
+        that the first parameter's index vector is lexicographically less than
+        the second parameter's index vector. The select function returns true if
+        the first parameter is selected and returns false if the second
+        parameter is selected and the function must hold transitivity
+        (i.e., if select(a, b) and select(b, c) are true, then select(a, c) is
+        also true) so that the selected element does not depend on the order of
+        the elements traversed for a given window.
+
+  * (C9) The `scatter` has type `(T, T) -> T, where the first and second input
```
ghpvnist, line 1545: Is the order of arguments for `scatter` enforced (aka `operand` first and `init_value` second)?

docs/spec_draft.md
```diff
@@ -1443,6 +1444,138 @@ specification. Numeric precision is implementation-defined.
...
+    * Default value of 1 for each dimension of `operand`.
+    * size(`window_strides`) $=$ rank(`operand`).
+    * `window_strides[i]` $\gt 0$ for all i $\in$ [0, size(window_strides)).
+
+  * (C7) On `padding`(optional)
+    * Default value of (0,0) for each dimension of `operand`.
+    * size(`padding`) $=$ rank(`operand`).
+    * dim(`padding`, 0) $=$ N and dim(`padding`, 1) = 2.
+
+  * (C8) `select` has type `(T, T) -> i1` such that T is a 0-D tensor with
```
ghpvnist, line 1535: Could we maybe simplify this to refer to stablehlo.select which doesn't yet exist since it seems to be defining what the select operation should be doing.

ghpvnist (changes requested):

docs/spec_draft.md
```diff
@@ -1443,6 +1444,138 @@ specification. Numeric precision is implementation-defined.
...
+    * size(`window_dimensions`) $=$ rank(`operand`).
+    * `window_dimensions[i]` $\gt 0$ for all i $\in$ [0, size(window_dimensions)).
+
+  * (C6) On `window_strides` (optional)
+    * Default value of 1 for each dimension of `operand`.
+    * size(`window_strides`) $=$ rank(`operand`).
+    * `window_strides[i]` $\gt 0$ for all i $\in$ [0, size(window_strides)).
+
+  * (C7) On `padding`(optional)
+    * Default value of (0,0) for each dimension of `operand`.
```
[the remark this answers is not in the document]
sdasgup3: same as https://github.com/openxla/stablehlo/pull/354#discussion_r1003826337

docs/spec_draft.md
```diff
@@ -1443,6 +1444,138 @@ specification. Numeric precision is implementation-defined.
...
+        (i.e., if select(a, b) and select(b, c) are true, then select(a, c) is
+        also true) so that the selected element does not depend on the order of
+        the elements traversed for a given window.
+
+  * (C9) The `scatter` has type `(T, T) -> T, where the first and second input
+         type `T` corresponds to the element type of `operand` and `init_value`
+         respectively. The output type `T` corresponds to the element types of
+         `result`.
+
+  * (C10) type(`operand`) $=$ type(`result`).
```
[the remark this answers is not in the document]
sdasgup3: Just like SortOp, I am using `type(..)`  macro. Also added an [action item](https://github.com/openxla/stablehlo/issues/289#issuecomment-1292822627) for improving notations.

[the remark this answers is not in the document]
sdasgup3: The constraints maintain the order in which the inputs.outputs are mentioned. That way `result` comes the last. Please let me know if I am missing your point.

docs/spec_draft.md
sdasgup3: Thanks! There was a bug in the image where the overlapping blue and purple windows  should have a common max location. The chosen values does not represent that.

Here is what is intended in the image:
|Window | selection location in source or result for max | source value to be scatted to that location |
|--|--|--|
|green | 2,0 | source[0,0] = 5|
|gold | 3,0 | source[1,0] = 7|
|blue | 2,1 | source[0,1] = 6|
|purple | 2,1 | source[1,1] = 8|

docs/spec_draft.md
sdasgup3: Thanks for the catch. No it is not needed.

docs/spec_draft.md
sdasgup3: Nice point. Fixed it as pointed out in  [ReduceOp::comment](https://github.com/openxla/stablehlo/pull/353#discussion_r1002758771).

docs/spec_draft.md
sdasgup3: My idea is to re-use `stablehlo.reduce_window` and that the `select` function needs to be used as the `reduce_window::body`. But the problem was  `reduce_window::body` constraints with type `(T, T) -> T` are different that that of `select` with type ``(T, T) -> i1`. Rewrote the formulation to fit them in. 

Reusing `reduce_window` can effectively simplify the operation. Let me know if you have follow up feedback.

stablehlo/dialect/StablehloOps.td
```diff
@@ -2304,7 +2304,27 @@ def StableHLO_SelectAndScatterOp: StableHLO_Op<"select_and_scatter",
     Multiple scattered elements which land in the same output location are
     combined using the `scatter` function.
 
-    See https://www.tensorflow.org/xla/operation_semantics#selectandscatter.
+    See:
```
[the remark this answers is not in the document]
sdasgup3: I tried but it seems the hard without loosing information.

docs/spec_draft.md
```diff
@@ -1443,6 +1444,141 @@ specification. Numeric precision is implementation-defined.
...
+
+  * (C2) `operand` and `source` have the same element type.
+
+  * (C3) `dim(source, i) = strided_bound[i]` for any i $\in$ axis(`source`)
+    * `strided_bound[i] = (padded_operand[i] == 0 || window_dimensions[i] > padded_operand[i]) ? 0 : floor((padded_operand[i] - window_dimensions[i]) / window_strides[i]) + 1`.
+    * `padded_operand[i] = padding[i][0] + dim(operand, i) + padding[i][1]`.
+
+  * (C4) On `window_dimensions`
+    * size(`window_dimensions`) $=$ rank(`operand`).
+    * `window_dimensions[i]` $\gt 0$ for all i $\in$ [0, size(window_dimensions)).
```
ghpvnist, line 1537: Typo: size(`window_dimensions`)

docs/spec_draft.md
```diff
@@ -1443,6 +1444,141 @@ specification. Numeric precision is implementation-defined.
...
+| Name                | Type                                       | Constraints      |
+|---------------------|--------------------------------------------|------------------|
+| `operand`           | tensor of any supported type               | (C1-C9)          |
+| `source`            | tensor of any supported type               | (C1-C3)          |
+| `init_value`        | 0-dimensional tensor of any supported type | (C4)             |
+| `window_dimensions` | 1-dimensional tensor constant type `si64`  | (C1), (C3), (C4) |
+| `window_strides`    | 1-dimensional tensor constant type `si64`  | (C3), (C5)       |
+| `padding`           | 2-dimensional tensor constant type `si64`  | (C3), (C6)       |
+| `select`            | `function`                                 | (C7)             |
+| `scatter`           | `function`                                 | (C4), (C8)       |
```
ghpvnist, line 1517: Same here.

ghpvnist (changes requested):

docs/spec_draft.md
```diff
@@ -1443,6 +1444,144 @@ specification. Numeric precision is implementation-defined.
...
+|----------|------------------------------|-------------|
+| `result` | tensor of any supported type | (C9)        |
+
+### Constraints
+
+  * (C1) rank(`operand`) $=$ rank(`source`) $=$ size(`window_dimensions`).
+
+  * (C2) `operand` and `source` have the same element type.
+
+  * (C3) `dim(source, i) = strided_bound[i]` for any i $\in$ axis(`source`)
```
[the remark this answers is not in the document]
sdasgup3: Let us leave that for now to avoid making a fix on  https://github.com/openxla/stablehlo/issues/289#issuecomment-1295351422

docs/spec_draft.md
```diff
@@ -1443,6 +1444,142 @@ specification. Numeric precision is implementation-defined.
...
+    * Default value of 1 for each dimension of `operand`.
+    * size(`window_strides`) $=$ rank(`operand`).
+    * `window_strides[i]` $\gt 0$ for all i $\in$ [0, size(window_strides)).
+
+  * (C6) On `padding`(optional)
+    * Default value of (0,0) for each dimension of `operand`.
+    * size(`padding`) $=$ rank(`operand`).
+    * dim(`padding`, 0) $=$ N and dim(`padding`, 1) = 2.
+
+  * (C7) `select` has type `(tensor<E>, tensor<E>) -> tensor<E>` where
```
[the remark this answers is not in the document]
sdasgup3: Thanks!

docs/spec_draft.md
```diff
@@ -1443,6 +1444,141 @@ specification. Numeric precision is implementation-defined.
...
+     * `source_values[index]` values, for all `index` in the index space
+      of `source_values`, in the ascending lexicographic order of `index`.
+      * Interspersed with an implementation-defined amount of `init_values`
+      at implementation-defined positions.
+
+### Inputs
+
+| Name                | Type                                       | Constraints      |
+|---------------------|--------------------------------------------|------------------|
+| `operand`           | tensor of any supported type               | (C1-C9)          |
```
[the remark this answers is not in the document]
sdasgup3: Done as per every mention
https://github.com/openxla/stablehlo/pull/275#discussion_r996253467

stablehlo/dialect/StablehloOps.td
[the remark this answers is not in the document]
sdasgup3: makes sense I will include that in https://github.com/openxla/stablehlo/blob/main/docs/spec_checklist.md in a followup PR.

ghpvnist (approved):

docs/spec_draft.md
```diff
@@ -1444,6 +1445,144 @@ specification. Numeric precision is implementation-defined.
...
+    * Default value of (0,0) for each dimension of `operand`.
+    * dim(`padding`, 0) $=$ rank(`operand`) and dim(`padding`, 1) = 2.
+
+  * (C7) `select` has type `(tensor<E>, tensor<E>) -> tensor<E>` where
+         `E = element_type(operand)`.
+
+  * (C8) The `scatter` has type `(tensor<E>, tensor<E>) -> tensor<E>` where
+         `E = element_type(operand)`.
+
+  * (C9) type(`operand`) $=$ type(`result`).
```
burmako, line 1558: Also, I don't remember what PR it was, but I think we decided to not make default values part of the spec and treat them as part of the API. Given that, let's drop "(optional)" and "Default value...".

docs/spec_draft.md
```diff
@@ -1444,6 +1445,144 @@ specification. Numeric precision is implementation-defined.
...
+    * `window_strides[i]` $\gt 0$ for all i $\in$ [0, size(window_strides)).
+
+  * (C6) On `padding`(optional)
+    * Default value of (0,0) for each dimension of `operand`.
+    * dim(`padding`, 0) $=$ rank(`operand`) and dim(`padding`, 1) = 2.
+
+  * (C7) `select` has type `(tensor<E>, tensor<E>) -> tensor<E>` where
+         `E = element_type(operand)`.
+
+  * (C8) The `scatter` has type `(tensor<E>, tensor<E>) -> tensor<E>` where
```
burmako, line 1555: Drop "The" at the beginning of the bullet.

docs/spec_draft.md
```diff
@@ -1444,6 +1445,144 @@ specification. Numeric precision is implementation-defined.
...
+
+### Constraints
+
+  * (C1) rank(`operand`) $=$ rank(`source`) $=$ size(`window_dimensions`).
+
+  * (C2) `operand` and `source` have the same element type.
+
+  * (C3) `dim(source, i) = strided_bound[i]` for any i $\in$ axis(`source`)
+    * `strided_bound[i] = (padded_operand[i] == 0 || window_dimensions[i] > padded_operand[i]) ? 0 : floor((padded_operand[i] - window_dimensions[i]) / window_strides[i]) + 1`.
+    * `padded_operand[i] = padding[i][0] + dim(operand, i) + padding[i][1]`.
```
burmako, line 1537: Same feedback as in https://github.com/openxla/stablehlo/pull/354/files#r1030970487. This will also allow to subsume the `rank(source)` part of C1.

burmako (changes requested):

docs/spec_draft.md
```diff
@@ -1444,6 +1445,144 @@ specification. Numeric precision is implementation-defined.
...
+
+The order of reductions is implementation-defined, which means that `scatter`
+and `init_value` must form a monoid to guarantee that the operation produces
+the same results for all inputs on all implementations.
+
+More formally,
+ * Apply [stablehlo.reduce_window](#stablehloreduce_window) with the following
+   parameters to get `selected_values` as output.
+   * `inputs` $=$ { `operand` },
+   * `init_values` $=$ { identity value for `select` },
```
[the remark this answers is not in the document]
sdasgup3: I can see the problem with the formulation and this is not what  I intend to do.

Given,

1. In the [hlo_evaluator](https://github.com/tensorflow/tensorflow/blob/73d23daf5cce6a06653b9c65992b4a4717644765/tensorflow/compiler/xla/hlo/evaluator/hlo_evaluator_typed_visitor.h#L1778), the padding values are skipped and for the very first selection iteration a value is compared with itself.  

2. Per the the HLO semantics,  the `init_value` parameter is to be used as the initial value for the scatter operation. , not the select function.

I am tempted to skip the padding values. IMO, `reduceWindowOp` cannot be used as is, but we might have to borrow parts of reduceWindow's formulation to do scatter of the non-padded values.

docs/spec_draft.md
```diff
@@ -1444,6 +1445,144 @@ specification. Numeric precision is implementation-defined.
...
+   * `window_dimensions`, `window_strides`, and `padding` are used as is,
+   * `body` $=$ `reduce_with_select` defined as
+
+     ```c++
+     reduce_with_select(tensor<E> arg0, tensor<E> arg1) -> tensor<E> {
+      return select(arg0, arg1) ? arg0 : arg1;
+     }
+     ```
+     where `E = element_type(operand)`. `select` is transitive, else the
+     behavior of the operation is undefined.
```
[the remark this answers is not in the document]
sdasgup3: After re-reading the following from the spec
```
 the function must hold transitivity (i.e., if select(a, b) and select(b, c) are true, then select(a, c) is also true) so that the selected element does not depend on the order of the elements traversed for a given window.
```

it seems that if the function is not transitive,  then the selection result will become order dependent which anyway is implementation defined.  

Having said that implementation defined seems like a apt treatment. (we need not to say so explicitly as mentioned by your comment)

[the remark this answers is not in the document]
burmako: Do we have users for such implementation-defined behavior, i.e. do people rely on that? If not, it would be the best to call this undefined behavior to rule it out completely.

For `reduce`, we couldn't really call that undefined behavior that because several practically-important reducers don't form a monoid with practically-important init_values. But I'd love to do that too if we could.

docs/spec_draft.md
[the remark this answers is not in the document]
burmako: It'd be good to avoid duplicating these nontrivial formalisms, and towards that end we could handwave this similarly to `all_reduce`:

```
 * `selected_values = reduce_window_without_init(...)` with the following inputs:
   * `inputs` $=$ [ `operand` ].
   * `window_dimensions`, `window_strides`, and `padding` which are used as is.
   * `base_dilations` $=$ `windows_dilations` $=$ `[1, ..., 1]`.
   * `body` defined as:
     ```C++
     (tensor<E> arg0, tensor<E> arg1) -> tensor<E> {
      return select(arg0, arg1) ? arg0 : arg1;
     }
     ```
     where `E = element_type(operand)`.
  * where `reduce_window_without_init` works exactly like `reduce_window`, 
    except that the `schedule` of the underlying `reduce` doesn't include 
    init values.
```

burmako (changes requested):

sdasgup3:
>> Do we have users for such implementation-defined behavior, i.e. do people rely on that? If not, it would be the best to call this undefined behavior to rule it out completely.

We have the following ticket to explore further:

https://github.com/openxla/stablehlo/issues/714

burmako (approved):

PR #377, opened 2022-10-21
Add spec for BatchNormInferenceOp

### LINKED ISSUE #347: Add spec for BatchNormInferenceOp
### Request description

To spec for stablehlo.batch_norm_inference

subhankarshah:
closes #347

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -163,6 +163,7 @@ described below)
    * [abs](#stablehloabs)
    * [add](#stablehloadd)
    * [and](#stablehloand)
+   * [batch_norm_inference](#stablehlobatch_norm_inference)
    * [broadcast_in_dim](#stablehlobroadcast_in_dim)
    * [case](#stablehlocase)
    * [ceil](#stablehloceil)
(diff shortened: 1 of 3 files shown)


docs/spec_draft.md
```diff
@@ -343,6 +344,102 @@ logical operation.
...
+  %upsized_offset = func.call @upsize(%offset) : (tensor<2xf32>) -> tensor<2x2x2xf32>
+  %upsized_mean = func.call @upsize(%mean) : (tensor<2xf32>) -> tensor<2x2x2xf32>
+  %upsized_variance = func.call @upsize(%variance) : (tensor<2xf32>) -> tensor<2x2x2xf32>
+  %upsized_epsilon = "mhlo.broadcast_in_dim"(%epsilon){
+    broadcast_dimensions = dense<[0,1,2]> : tensor<3xi64>
+  } : (tensor<f32>) -> tensor<2x2x2xf32>
+
+  %mean_removed = "mhlo.subtract"(%operand, %upsized_mean) : (tensor<2x2x2xf32>, tensor<2x2x2xf32>) -> tensor<2x2x2xf32>
+  %pre_sqrt = "mhlo.add"(%variance, %upsized_epsilon) : (tensor<2x2x2xf32>, tensor<2x2x2xf32>) -> tensor<2x2x2xf32>
+  %inv_sqrt = "mhlo.rsqrt"(%pre_sqrt) : (tensor<2x2x2xf32>) -> tensor<2x2x2xf32>
```
sdasgup3, line 385: mhlo -> stablehlo

docs/spec_draft.md
```diff
@@ -343,6 +344,102 @@ logical operation.
...
+func.func @batch_norm_inference(
+    %operand: tensor<2x2x2xf32>,
+    %scale: tensor<2xf32>,
+    %offset: tensor<2xf32>,
+    %mean: tensor<2xf32>,
+    %variance: tensor<2xf32>,
+    %epsilon: tensor<f32>,
+    %feature_index: tensor<i64>
+    ) -> tensor<2x2x2xf32> {
+  // Assuming %feature_index = 1
```
ghpvnist, line 374: `feature_index` isn't used anywhere with the value `1`. Can you also explain how this applies with non-1 `feature_index` value?

ghpvnist (changes requested):

docs/spec_draft.md
```diff
@@ -343,6 +344,102 @@ logical operation.
...
+| `result` | tensor of floating-point type |
+
+### Constraints
+
+  * (C1) `feature_index` $\lt$ rank(`operand`).
+  * (C2) size(`scale`) $=$ `dim(operand, feature_index)`.
+  * (C3) size(`offset`) $=$ `dim(operand, feature_index)`.
+  * (C4) size(`mean`) $=$ `dim(operand, feature_index)`.
+  * (C5) size(`variance`) $=$ `dim(operand, feature_index)`.
+
```
sdasgup3, line 419: type(`operand`) = type(`result`)

docs/spec_draft.md
```diff
@@ -343,6 +344,102 @@ logical operation.
...
+```mlir
+func.func @upsize(%0: tensor<2xf32>) -> tensor<2x2x2xf32> {
+  %1 = mhlo.reshape %0 : (tensor<2xf32>) -> tensor<1x2x1xf32>
+  %2 = "mhlo.broadcast_in_dim"(%1){
+    broadcast_dimensions = dense<[0,1,2]> : tensor<3xi64>
+  } : (tensor<1x2x1xf32>) -> tensor<2x2x2xf32>
+  func.return %2 : tensor<2x2x2xf32>
+}
+
+func.func @batch_norm_inference(
```
sdasgup3, line 365: Can we check if https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/batchnorm_expander.cc gives the same behavior? 

Ignore if already checked.

docs/spec_draft.md
```diff
@@ -343,6 +344,102 @@ logical operation.
...
+
+### Semantics
+
+Normalizes the `operand` tensor across batch and spatial dimensions, using
+`mean` and `variance` provided in input, for each feature in the `feature_index`
+dimension and produces a `result` tensor. More formally, the semantics can be
+expressed using MLIR syntax as follows:
+
+```mlir
+func.func @upsize(%0: tensor<2xf32>) -> tensor<2x2x2xf32> {
```
sdasgup3, line 357: After looking into the equivalent stablehlo code, which demonstrates how this op can be implemented using other ops, I am not sure if this helps in improving the readability of its specification. Moreover, there could be other (maybe inefficient) ways this can be implemented.

Maybe we should, 

1. include the overview of the underlying computation involved in the op  as mentioned in https://www.tensorflow.org/xla/operation_semantics#batchnormtraining. I agree that it not super simple either, but helps the reader to understand how the various inputs like mean, epsilon etc, are connected. 

2. Add comments to the implementation so as to expose its connection with (1), provided that connection  exists AND easy to point out using comments **.  

** I have not gone through the  implementation fully.  Hence not sure. Will do that later and comment further.

docs/spec_draft.md
```diff
@@ -343,6 +344,67 @@ logical operation.
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.batch_norm_inference
+
+### Semantics
+
+Normalizes the `operand` tensor across batch and spatial dimensions, using
+`mean` and `variance` provided in input, for each feature in the `feature_index`
```
sdasgup3, line 352: can we skip `provided in input`. That is because it seems we are qualifying some as input and others (`feature_index`) not.

docs/spec_draft.md
```diff
@@ -343,6 +344,67 @@ logical operation.
...
+  * (C5) size(`mean`) $=$ `dim(operand, feature_index)`.
+  * (C6) size(`variance`) $=$ `dim(operand, feature_index)`.
+  * (C7) `operand` and `result` have the same type.
+
+### Examples
+
+```mlir
+// %operand: [
+//            [[1.0, 2.0], [3.0, 4.0]],
+//            [[5.0, 6.0], [7.0, 8.0]],
```
sdasgup3, line 390: comma at the end

stablehlo/dialect/StablehloOps.td
```diff
@@ -1582,9 +1582,20 @@ def StableHLO_BatchNormInferenceOp : StableHLO_Op<"batch_norm_inference",
     InferTensorType]> {
   let summary = "Batch Normalization for Inference";
   let description = [{
-    Normalizes an array across batch and spatial dimensions.
+    Normalizes the `operand` tensor across batch and spatial dimensions, using
+    `mean` and `variance` provided in input, for each feature in the
+    `feature_index` dimension and produces a `result` tensor.
+
+    See:
+    https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md#stablehlobatch_norm_inference
```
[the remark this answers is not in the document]
subhankarshah: sorry, I did not get this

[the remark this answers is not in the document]
sdasgup3: From this diff it seems that the link is not aligned well. After looking at your branch it looks good. SO nvm.

sdasgup3 (approved):

docs/spec_draft.md
```diff
@@ -343,6 +344,67 @@ logical operation.
...
+### Outputs
+
+| Name     | Type                          |
+|----------|-------------------------------|
+| `result` | tensor of floating-point type |
+
+### Constraints
+
+  * (C1) 0 $\le$ `feature_index` $\lt$ rank(`operand`).
+  * (C2) 0 $\lt$ rank(`operand`).
```
sdasgup3, line 378: C2 redundant as it can be derived from C1.

docs/spec_draft.md
```diff
@@ -349,11 +349,10 @@ logical operation.
...
 
 Normalizes the `operand` tensor across batch and spatial dimensions, for each
-feature in the `feature_index` dimension and produces a `result` tensor. It uses
-`mean` and `variance` provided in the input instead of computing them. Refer the
-`Batch Normalizing Transform` algorithm from the
-[Batch Normalization](https://arxiv.org/abs/1502.03167) paper for the detailed
-algorithm.
+feature in the `feature_index` dimension and produces a `result` tensor. The
+algorithm for `Batch Norm Inference` is same as 
+[Batch Norm Training](#stablehlobatch_norm_training) except, it uses `mean`
```
sdasgup3, line 354: replace ") except, it..." by "), except it"

sdasgup3 (approved):

docs/spec_draft.md
```diff
@@ -343,6 +344,66 @@ logical operation.
...
 
+## stablehlo.batch_norm_inference
+
+### Semantics
+
+Normalizes the `operand` tensor across batch and spatial dimensions, for each
+feature in the `feature_index` dimension and produces a `result` tensor. The
+algorithm for `Batch Norm Inference` is same as 
+[Batch Norm Training](#stablehlobatch_norm_training), except it uses `mean`
+and `variance` provided in the input instead of computing them.
```
burmako, line 355: I took some time to compare: 1) the algorithm in the paper, 2) the previous notation that used MLIR, 3) the current formulation - and think what would be the best for the spec.

Re: 1)/3), the algorithm in the spec is algebraically the same as what's used by [BatchNormExpander](https://github.com/tensorflow/tensorflow/blob/8e611d89cacc31c6ee3f2ae94cc34fc76402fd8e/tensorflow/compiler/xla/service/batchnorm_expander.cc) and [UnfuseBatchNormInference](https://github.com/tensorflow/mlir-hlo/blob/8ce6d6bea758ef9dac37ee67f2180d71b334d9fe/lib/Dialect/mhlo/transforms/unfuse_batch_norm.cc#L101). The HLO-based and MHLO-based passes perform computations a bit differently, perhaps accidentally or perhaps to improve numerical accuracy, but algebraically it's the same thing.

What I find somewhat lacking in Algorithm 1 in the spec, as far as speccing `batch_norm_inference` goes, is that it's not immediately clear how various inputs map on the notation in the paper. It's not super hard to figure out that gamma stands for scale, beta stands for offset, etc, but that's not immediately evident either. Also `feature_index` is something that doesn't come up in the algorithm at all.

Re: 2), I think it correlates very well with the inputs and explains how all of them are used, but it was pretty verbose. Also, MLIR notation isn't really designed for parameterization, so it wasn't clear how to generalize it from a single example with concrete shapes and concrete value of feature_index.

Instead, I propose that we use Python-based pseudocode to express the algorithm - I think the pseudocode has the strong sides of 1), 2) and 3) without having their downsides. Here's how the Semantics section could look like:

============
Normalizes the `operand` tensor across all dimensions except for the 
`feature_index` dimension and produces a `result` tensor.

```Python
def batch_norm_inference(operand, scale, offset, mean, variance, epsilon, feature_index):
  # Broadcast inputs to shape(operand)
  scale_bcast = broadcast_in_dim(scale, [feature_index], shape(operand))
  offset_bcast = broadcast_in_dim(offset, [feature_index], shape(operand))
  mean_bcast = broadcast_in_dim(mean, [feature_index], shape(operand))
  variance_bcast = broadcast_in_dim(variance, [feature_index], shape(operand))
  epsilon_bcast = broadcast_in_dim(constant(epsilon), [], shape(operand))

  # Perform normalization using the provided `mean` and `variance` instead of 
  # computing them like `batch_norm_training` does.
  centered_operand = subtract(operand, mean_bcast)
  stddev = sqrt(add(variance_bcast, epsilon_bcast))
  normalized_operand = divide(centered_operand, stddev)
  return add(multiply(scale_bcast, normalized_operand), offset_bcast)
```
============

burmako (changes requested):

docs/spec_draft.md
```diff
@@ -348,6 +349,83 @@ logical operation.
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.batch_norm_inference
+
+### Semantics
+
+Normalizes the `operand` tensor across batch and spatial dimensions, for each
```
burmako, line 356: Since we don't define what batch and spatial dimensions are, let's reformulate in a more precise fashion:

```
Normalizes the `operand` tensor across all dimensions except for the 
`feature_index` dimension and produces a `result` tensor.
```

docs/spec_draft.md
```diff
@@ -348,6 +349,83 @@ logical operation.
...
+feature in the `feature_index` dimension and produces a `result` tensor. More
+formally, the semantics can be expressed using Python-like syntax as follows:
```
burmako, line 358: Proposed rewording to: 1) emphasize the relationship with existing ops, 2) to talk about numerical accuracy in the same way that we're doing this with other ops:

```
More formally, this operation can be expressed via a decomposition to existing
StableHLO operations as defined below. Numeric precision is 
implementation-defined.
```

burmako (changes requested):

burmako (approved):

ghpvnist (approved):

PR #378, opened 2022-10-21
Refine the inferred bounds of SliceOp

zhouxin913:
This PR refines `SliceOp`: the existing logic of bounds is kind of tricky: https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/TypeInference.cpp#L1120: it is using `inputBounds[i], start[i], limit[i], strideVals[i]` but jump over all the verification for them in later lines. This looks danganous and refactored in this PR. Now the logic for for dim `i` the loop is:
1. Do checks for each attr[i]
2. If the dim is dynamic and NO bounds, set result_dim=-1, continue loop;
3. For static dim, calculate result_dim; set -1 for result_bound if need
4. For dynamic dim with bounds, set -1 for result_dim; calculate result_bound if need.

Also add @smit-hinsu as reviewer.

NOTE: the (1) above may have **MHLO impact**, so need fix MHLO tests accordingly if some "positive" tests error out: they are invalid actually.

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -1101,15 +1101,30 @@ LogicalResult inferSliceOp(Optional<Location> location, Value operand,
   SmallVector<int64_t, 4> strideVals(strides.getValues<int64_t>());
 
   ArrayRef<int64_t> inputBounds = encodingToBounds(rankedTy.getEncoding());
-  SmallVector<int64_t> shape;
-  shape.reserve(rank);
-  SmallVector<int64_t> resultBounds;
+  SmallVector<int64_t> shape(rank, ShapedType::kDynamicSize);
+  SmallVector<int64_t> resultBounds(inputBounds.size(),
+                                    ShapedType::kDynamicSize);
 
   for (int64_t i = 0, e = rank; i != e; i++) {
     // P3.
     if (start[i] < 0)
       return emitOptionalError(location, "negative start index ", start[i],
                                " in dimension ", i);
+
+    // P4.
+    bool isStaticDim = !hlo::isDynamicDimSize(rankedTy.getDimSize(i));
+    bool isStaticBound =
+        !inputBounds.empty() && !hlo::isDynamicDimSize(inputBounds[i]);
+    if (isStaticDim || isStaticBound) {
+      int64_t operandSizeOrBound =
+          isStaticDim ? rankedTy.getDimSize(i) : inputBounds[i];
+      StringRef sizeOrBound = isStaticDim ? "size" : "bound";
+      if (limit[i] > operandSizeOrBound)
+        return emitOptionalError(location, "limit index ", limit[i],
+                                 " is larger than dimension ", sizeOrBound, " ",
+                                 operandSizeOrBound, " in dimension ", i);
+    }
+
     // P5.
     if (start[i] > limit[i])
       return emitOptionalError(location, "start index ", start[i],
@@ -1120,29 +1135,8 @@ LogicalResult inferSliceOp(Optional<Location> location, Value operand,
       return emitOptionalError(location, "stride must be positive but got ",
                                strideVals[i], " in dimension ", i);
 
-    bool isDynamicDim = hlo::isDynamicDimSize(rankedTy.getDimSize(i));
-    if (isDynamicDim) {
-      shape.push_back(ShapedType::kDynamicSize);
-      if (inputBounds.empty()) continue;
-    }
-
-    int64_t operandSizeOrBound =
-        isDynamicDim ? inputBounds[i] : rankedTy.getDimSize(i);
-    StringRef sizeOrBound = isDynamicDim ? "bound" : "size";
-
-    // P4.
-    if (operandSizeOrBound != ShapedType::kDynamicSize &&
-        limit[i] > operandSizeOrBound)
-      return emitOptionalError(location, "limit index ", limit[i],
-                               " is larger than dimension ", sizeOrBound, " ",
-                               operandSizeOrBound, " in dimension ", i);
-
-    int64_t resultSizeOrBound = static_cast<int64_t>(
+    shape[i] = static_cast<int64_t>(
         llvm::divideCeil(limit[i] - start[i], strideVals[i]));
-    (isDynamicDim ? resultBounds : shape).push_back(resultSizeOrBound);
-
-    if (!isDynamicDim && !inputBounds.empty())  // use bound=-1 for static dim
-      resultBounds.push_back(ShapedType::kDynamicSize);
   }
 
   inferredReturnTypes.push_back(RankedTensorType::get(
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -463,6 +463,7 @@ LogicalResult inferBatchNormTrainingOp(
   return success();
 }
 
+// Used by IfOp and CaseOp
 LogicalResult inferConditionalOp(Optional<Location> location,
                                  RegionRange branches,
                                  SmallVectorImpl<Type>& inferredReturnTypes) {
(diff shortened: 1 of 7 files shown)


stablehlo/dialect/TypeInference.cpp
```diff
@@ -1114,20 +1103,26 @@ LogicalResult inferSliceOp(Optional<Location> location, Value operand,
...
+  };
+
   for (int64_t i = 0, e = rank; i != e; i++) {
-    if (!inputBounds.empty())
-      resultBounds.push_back(
-          inferSliceDim(inputBounds[i], start[i], limit[i], strideVals[i]));
     if (hlo::isDynamicDimSize(rankedTy.getDimSize(i))) {
       shape.push_back(ShapedType::kDynamicSize);
-      continue;
+      if (inputBounds.empty()) continue;  // dynamic dim without bounds
```
smit-hinsu, line 1117: Could we move this at L1144? (have similar comment in the other PR.) I spoke too soon there that this PR logic seems good.

That way we check that start is valid even if we are not planning to use it in the computation.

Might be related to your plans to share verification between shape functions and op verify so it is okay if we just add TODO now.

smit-hinsu (approved):
Thanks!

stablehlo/dialect/TypeInference.cpp
zhouxin913: Please note that the "op verifier" thing has nothing to do with the topic we discussed here: the `PadOp::verify()` was totally removed a while ago.

"Could we move this at L1144",
"That way we check that start is valid even if we are not planning to use it in the computation." 
Let me explain this situation a little bit more: In existing StableHLO/MHLO code we have
```
for each dim:
  if (hlo::isDynamicDimSize(rankedTy.getDimSize(i))) {
      shape.push_back(ShapedType::kDynamicSize);
      continue;
  // Checks for attributes
  // calculate static result_dim
```
which actually tolerate any arbitrary attr values for a "?" dim: it does not check anything. 

Plan A (my current commit): keep tolerate invalid attr values and do "continue;" for unbounded case.
Plan B (Smit suggests, I like it but with concerns): Move the `// Checks for attributes` part firstly whatever.
@burmako What do you think of this (The other PR has exactly same issue) ? Like Smit says, we can "move this line at L1144", but in a (maybe small) risk that MHLO could have "positive tests with invalid attrs in '?' dim" ? Which may need more fixing in integration?

smit-hinsu: ah, now I see that we rely on just shape inference for these ops. sorry for the confusion. These are already de-duped.

But, does that mean that the following op won't marked illegal?

"stablehlo.slice"(%arg0) {start_indices = dense<**[-1, 0]**> : tensor<2xi64>, limit_indices = dense<[2, 4]> : tensor<2xi64>, strides = dense<[1, 2]> : tensor<2xi64>} : (tensor<?x?xi32>) -> tensor<?x?xi32>
  %1 = "hlo_test_infer.get_return_type_components"(%0)


If that is the case, then Plan B becomes more important and it is better we do that sooner rather than later. I also doubt that we will hit many illegal IR as downstream users of these dialects will be checking that anyway. Just my hunch so it might still be better to verify upfront.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1114,20 +1103,26 @@ LogicalResult inferSliceOp(Optional<Location> location, Value operand,
...
     }
     // P3.
     if (start[i] < 0)
       return emitOptionalError(location, "negative start index ", start[i],
                                " in dimension ", i);
     // P4.
-    if (limit[i] > rankedTy.getDimSize(i))
+    if (!hlo::isDynamicDimSize(rankedTy.getDimSize(i)) &&
```
smit-hinsu, line 1124: if it is dynamic but limit > bound, then also there should be an error?

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1141,8 +1136,14 @@ LogicalResult inferSliceOp(Optional<Location> location, Value operand,
...
 
-    shape.push_back(inferSliceDim(rankedTy.getDimSize(i), start[i], limit[i],
-                                  strideVals[i]));
+    if (!hlo::isDynamicDimSize(rankedTy.getDimSize(i))) {  // static dim
+      shape.push_back(inferSliceDim(rankedTy.getDimSize(i), start[i], limit[i],
+                                    strideVals[i]));
+      if (!inputBounds.empty())
+        resultBounds.push_back(ShapedType::kDynamicSize);
+    } else  // dynamic dim with bounds
+      resultBounds.push_back(
```
smit-hinsu, line 1145: let's add a test demonstrating the issue this PR is fixing?

stablehlo/dialect/TypeInference.cpp
burmako: I support Plan B - both here and for PadOp.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1141,8 +1136,14 @@ LogicalResult inferSliceOp(Optional<Location> location, Value operand,
...
+          inferSliceDim(inputBounds[i], start[i], limit[i], strideVals[i]));
```
burmako, line 1146: inferSliceDim is used only once, so let's inline it into the callsite, so that readers don't have to jump back and forth between the definition and the callsite. Then we can combine both static sizes and dynamic bounds in a single snippet of code in a very straightforward way:

```C++
// Assuming that `bool isDynamicDim = hlo::isDynamicDimSize(inputShape[i]);` is declared above.
int64_t operandSizeOrBound = isDynamicDim ? inputBounds[i] : rankedTy.getDimSize(i);
int64_t resultSizeOrBound = static_cast<int64_t>(llvm::divideCeil(limit[i] - start[i], strideVals[i]));
(isDynamicDim ? resultBounds : shape).push_back(resultSizeOrBound);
```

burmako (changes requested):

stablehlo/dialect/TypeInference.cpp
zhouxin913: As everyone is agreed (great!) Let's make the verification is correct order. And add a note in description that this maybe have MHLO impact.

stablehlo/dialect/TypeInference.cpp
zhouxin913: Done.

stablehlo/dialect/TypeInference.cpp
zhouxin913: Thanks for the demo! Your suggested is used except the last line "(isDynamicDim ? resultBounds : shape)", we have 3 cases instead of 2:
1. unbounded dynamic
2. bounded dynamic
3. static
So a "?:" is not quite useful here. So I propose my new code in new commit. Please let me know what you think of it.

stablehlo/dialect/TypeInference.cpp
zhouxin913: Good point! Added code and test.

smit-hinsu (approved):

stablehlo/dialect/TypeInference.cpp
burmako: Here's the full code snippet that I had in mind. I liked your original idea to `continue` early for the unbounded dynamic case and based the proposal on that:

```C++
  for (int64_t i = 0, e = rank; i != e; i++) {
    // P3.
    if (start[i] < 0)
      return emitOptionalError(location, "negative start index ", start[i],
                               " in dimension ", i);
    // P5.
    if (start[i] > limit[i])
      return emitOptionalError(location, "start index ", start[i],
                               " is larger than limit index ", limit[i],
                               " in dimension ", i);
    // P6.
    if (strideVals[i] <= 0)
      return emitOptionalError(location, "stride must be positive but got ",
                               strideVals[i], " in dimension ", i);

    bool isDynamicDim = hlo::isDynamicDimSize(inputShape[i]);
    if (isDynamicDim) {
      shape.push_back(ShapedType::kDynamicSize);
      if (inputBounds.empty()) continue;
    }

    int64_t operandSizeOrBound = isDynamicDim ? inputBounds[i] : rankedTy.getDimSize(i);
    StringRef sizeOrBound = isDynamicDim ? "bound" : "size";

    // P4.
    if (limit[i] > operandSizeOrBound)
      return emitOptionalError(location, "limit index ", limit[i],
                               " is larger than dimension ", sizeOrBound,
                               rankedTy.getDimSize(i), " in dimension ", i);

    int64_t resultSizeOrBound = static_cast<int64_t>(
        llvm::divideCeil(limit[i] - start[i], strideVals[i]));
    (isDynamicDim ? resultBounds : shape).push_back(resultSizeOrBound);
  }
```

What I like about this proposal is that it allows to write P4 and the final computation in an almost generic way - only 1 conditional is needed. For comparison, with the current PR, we need 3 `isDynamicDimSize` checks and 3 `!inputBounds.empty` checks.

zhouxin913: Thanks a lot for the code snippet above! and used with 2 changes:
1. in P4, `limit[i] > operandSizeOrBound` -> `operandSizeOrBound != ShapedType::kDynamicSize &&
        limit[i] > operandSizeOrBound`. Because in bounded dynamic case, the `inputBounds[i]` (`operandSizeOrBound`) could be -1 even for a dynamic dim. We should skip P4 check in such case.
2. in the end, I add 2 new lines: 
```
if (!isDynamicDim && !inputBounds.empty())  // use bound=-1 for static dim
  resultBounds.push_back(ShapedType::kDynamicSize);
``` 
for the static dim when dealing with bounded dynamism.

For (2), an alternative way is to initializing the `resultBounds`/`shape` with (size_k, -1). But I do not see it worth an big change like that.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1114,23 +1103,31 @@ LogicalResult inferSliceOp(Optional<Location> location, Value operand,
...
   shape.reserve(rank);
   SmallVector<int64_t> resultBounds;
+
+  auto inferSliceDim = [&](int64_t inputDim, int64_t start, int64_t end,
+                           int64_t stride) {
+    return inputDim == ShapedType::kDynamicSize
+               ? ShapedType::kDynamicSize
+               : static_cast<int64_t>(llvm::divideCeil(end - start, stride));
+  };
+
```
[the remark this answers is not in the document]
zhouxin913: Removed.

stablehlo/dialect/TypeInference.cpp
burmako: You're right, there's a possibility which I missed in the code snippet in my previous comment - fully dynamic dimension (? for size, ? for bound). 

This prompted me to write down all the possibilities that can happen, which can be found in the table below, given that `Z = llvm::divideCeil(limit - start, stride)`, and I think I found some bugs in the current implementation:

|    | Input Shape | Input Bound | Action             | Output Shape | Output Bound |
|----|-------------|-------------|--------------------|--------------|--------------|
| C1 | X           | N/A         | Check `X <= limit` | Z            | N/A          |
| C2 | X           | ?           | Check `X <= limit` | Z            | N/A          |
| C3 | X           | Y           | Impossible         | Impossible   | Impossible   |
| C4 | ?           | N/A         | No checks          | Z            | N/A          |
| C5 | ?           | ?           | No checks          | Z            | N/A          |
| C6 | ?           | Y           | Check `Y <= limit` | Z            | N/A          |

I think that the only thing that changes if we vary different possibilities between input shapes and input bounds is P4 - if either the dim size or the dim bound are static, then we need to check whether they are `<= limit`. But the output should always be the same - fully static shape without bounds (because `start`, `limit` and `strides` unambiguously specify a static dim size):

```C++
  ArrayRef<int64_t> inputBounds = encodingToBounds(rankedTy.getEncoding());
  SmallVector<int64_t> shape(rank, ShapedType::kDynamicSize);
  SmallVector<int64_t> resultBounds(inputBounds.size(),
                                    ShapedType::kDynamicSize);

  for (int64_t i = 0, e = rank; i != e; i++) {
    // P3.
    if (start[i] < 0)
      return emitOptionalError(location, "negative start index ", start[i],
                               " in dimension ", i);

    // P4.
    bool isStaticDim = !hlo::isDynamicDimSize(rankedTy.getDimSize(i));
    bool isStaticBound =
        !inputBounds.empty() && !hlo::isDynamicDimSize(inputBounds[i]);
    if (isStaticDim || isStaticBound) {
      int64_t operandSizeOrBound =
          isStaticDim ? rankedTy.getDimSize(i) : inputBounds[i];
      StringRef sizeOrBound = isStaticDim ? "size" : "bound";
      if (limit[i] > operandSizeOrBound)
        return emitOptionalError(location, "limit index ", limit[i],
                                 " is larger than dimension ", sizeOrBound, " ",
                                 operandSizeOrBound, " in dimension ", i);
    }

    // P5.
    if (start[i] > limit[i])
      return emitOptionalError(location, "start index ", start[i],
                               " is larger than limit index ", limit[i],
                               " in dimension ", i);
    // P6.
    if (strideVals[i] <= 0)
      return emitOptionalError(location, "stride must be positive but got ",
                               strideVals[i], " in dimension ", i);

    shape[i] = static_cast<int64_t>(
        llvm::divideCeil(limit[i] - start[i], strideVals[i]));
  }
```

I've created a pull request to your branch that updates the logic and the tests accordingly: https://github.com/zhouxin913/stablehlo/pull/1.

burmako (changes requested):
See discussion in the unresolved thread.

stablehlo/dialect/TypeInference.cpp
zhouxin913: The table makes it extremely clear! and I agree the big conclusions we have here are:
1)  The new logic we summarize in the table will not affect the old users in any way (who do not use bounds at all) 
2) for some ops/cases, the bounds in input will lead to bounds in output. But for some other ops like this SliceOp, it may replace the bounds with static dims. This will be expected to be case by case for every op in the future. 

https://github.com/zhouxin913/stablehlo/pull/1 is approved and merged, which update this PR with a new commit.

burmako (approved):

PR #389, opened 2022-10-25
Handle bounds in the ConcatenateOp shape function

zhouxin913:
`ConcatenateOp` has variadic inputs, so `inferMostSpecificType` need to be used to infer multiple dims (size, bound) from multiple input tensors, which set the rules of how to infer the return bounds for binary ops like `AddOp`.
However, `ConcatenateOp` is more complicated:
1. For the dim that will be concatenated, we need a new set of different rules to infer the dimension sizes and bounds.
2. For the rest N-1 dimension, the existing rules used by binary ops can be re-used.
More details for (1) from the new code comment:
```
// Inference rules to concat dimensions with bounds (lhs/rhs are commutative):
//       Dim of lhs     Dim of rhs      Infer
//  c0:  X              Y               X+Y
//  c1:  X              ?               ?
//  c2:  X              ?, B            ?, X+B
//  c3:  ?              ?               ?
//  c4:  ?              ?, B            ?
//  c5:  ?, B1          ?, B2           ?, B1+B2
```
So, in the PR, `inferMostSpecificType` become large but refactored to be more easy to follow:
1. `inferMergedDimAndBound()`: The old kernel part of how to handle each dim for binary ops is moved to this separate util function. NO logic change but rewrite with more readability.
2. `inferConcatenatedDimAndBound()`: The new rules are implemented by this new dedicated method.
3. `inferMostSpecificType()` is still the public entry point: with a new argument `concatenateDims` which indicates which dims will be concatenated instead of merged.

I would expect that in the future more categories of inference rules will be introduced for interface `inferMostSpecificType()`, and it makes sense to me to refactor further when that happens.
I prefer not to split the (1) and (2) to two different PRs, as (1) is more meaningful when staying with (2) (3). 

Add reviewer @smit-hinsu .

### DIFF
--- a/stablehlo/dialect/Base.cpp
+++ b/stablehlo/dialect/Base.cpp
@@ -183,9 +183,10 @@ Attribute boundsToEncoding(Attribute prototype, ArrayRef<int64_t> bounds) {
 //  c3:  ?              ?               ?
 //  c4:  ?              ?, B            ?
 //  c5:  ?, B           ?, C            ?, B+C
-static std::pair<int64_t, int64_t> inferConcatenatedDimAndBound(
-    int64_t leftSize, int64_t rightSize, int64_t leftBound,
-    int64_t rightBound) {
+std::pair<int64_t, int64_t> inferConcatenatedDimAndBound(int64_t leftSize,
+                                                         int64_t rightSize,
+                                                         int64_t leftBound,
+                                                         int64_t rightBound) {
   bool isLeftStaticDim = !isDynamicDimSize(leftSize);
   bool isRightStaticDim = !isDynamicDimSize(rightSize);
   int64_t size = ShapedType::kDynamicSize;
@@ -200,7 +201,7 @@ static std::pair<int64_t, int64_t> inferConcatenatedDimAndBound(
         !isDynamicDimSize(rightSizeOrBound))
       bound = leftSizeOrBound + rightSizeOrBound;
   }
-  return std::make_pair(size, bound);
+  return {size, bound};
 }
 
 // Inference rules to merge dimensions with bounds (lhs/rhs are commutative):
@@ -244,10 +245,10 @@ FailureOr<std::pair<int64_t, int64_t>> inferMergedDimAndBound(
   return std::make_pair(size, bound);
 }
 
-LogicalResult inferMostSpecificType(Optional<Location> location,
-                                    TypeRange inputTypes,
-                                    SmallVectorImpl<Type>& inferredReturnTypes,
-                                    int64_t concatenateDim) {
+// TODO(zhouxin) Refactor to better handle errors and return single type
+LogicalResult inferMostSpecificType(
+    Optional<Location> location, TypeRange inputTypes,
+    SmallVectorImpl<Type>& inferredReturnTypes) {
   SmallVector<RankedTensorType> rankedTypes;
   for (auto inputType : inputTypes)
     if (auto rankedType = inputType.dyn_cast<RankedTensorType>())
@@ -258,7 +259,7 @@ LogicalResult inferMostSpecificType(Optional<Location> location,
   }
 
   auto rank = rankedTypes[0].getRank();
-  SmallVector<int64_t> inferredSizes(rankedTypes[0].getShape());
+  SmallVector<int64_t> inferredSizes(rank, ShapedType::kDynamicSize);
   SmallVector<int64_t> inferredBounds(rank, ShapedType::kDynamicSize);
   bool anyInputHaveBounds = false;
 
@@ -267,30 +268,18 @@ LogicalResult inferMostSpecificType(Optional<Location> location,
     ArrayRef<int64_t> bounds = encodingToBounds(rankedType.getEncoding());
     if (!bounds.empty()) anyInputHaveBounds = true;
 
-    // Init the inferredSizes[dim] / inferredBounds[dim] with rankedTypes[0]
-    if (it.index() == 0) {
-      if (!bounds.empty()) inferredBounds = to_vector(bounds);
-      continue;
-    }
-
-    // Infer each dim for current rankedType (from inputTypes[1:end])
     for (int dim = 0; dim < rank; ++dim) {
       std::pair<int64_t, int64_t> inferredDimAndBound;
-
       int64_t leftSize = inferredSizes[dim];
       int64_t rightSize = rankedType.getShape()[dim];
       int64_t leftBound = inferredBounds[dim];
       int64_t rightBound =
           bounds.empty() ? ShapedType::kDynamicSize : bounds[dim];
-      if (dim == concatenateDim) {
-        inferredDimAndBound = inferConcatenatedDimAndBound(
-            leftSize, rightSize, leftBound, rightBound);
-      } else {
-        auto inferredDimAndBoundOrErr = inferMergedDimAndBound(
-            location, dim, leftSize, rightSize, leftBound, rightBound);
-        if (failed(inferredDimAndBoundOrErr)) return failure();
-        inferredDimAndBound = *inferredDimAndBoundOrErr;
-      }
+
+      auto inferredDimAndBoundOrErr = inferMergedDimAndBound(
+          location, dim, leftSize, rightSize, leftBound, rightBound);
+      if (failed(inferredDimAndBoundOrErr)) return failure();
+      inferredDimAndBound = *inferredDimAndBoundOrErr;
       inferredSizes[dim] = inferredDimAndBound.first;
       inferredBounds[dim] = inferredDimAndBound.second;
     }
@@ -304,7 +293,6 @@ LogicalResult inferMostSpecificType(Optional<Location> location,
           // there are no bounds at all in inputs, thus sparsity attributes will
           // be included in the return type
           anyInputHaveBounds ? inferredBounds : llvm::ArrayRef<int64_t>({}))));
-
   return success();
 }
 
--- a/stablehlo/dialect/Base.cpp
+++ b/stablehlo/dialect/Base.cpp
@@ -176,5 +176,151 @@ Attribute boundsToEncoding(Attribute prototype, ArrayRef<int64_t> bounds) {
   return dialect->createBoundedAttr(bounds);
 }
 
+// Inference rules to concat dimensions with bounds (lhs/rhs are commutative):
+//       Dim of lhs     Dim of rhs      Infer
+//  c0:  X              Y               X+Y
+//  c1:  X              ?               ?
+//  c2:  X              ?, B            ?, X+B
+//  c3:  ?              ?               ?
+//  c4:  ?              ?, B            ?
+//  c5:  ?, B1          ?, B2           ?, B1+B2
+static std::pair<int64_t, int64_t> inferConcatenatedDimAndBound(
+    int64_t leftSize, int64_t rightSize, int64_t leftBound,
+    int64_t rightBound) {
+  bool isLeftStaticDim = leftSize != ShapedType::kDynamicSize;
+  bool isRightStaticDim = rightSize != ShapedType::kDynamicSize;
+  if (isLeftStaticDim && isRightStaticDim)
+    return std::make_pair(leftSize + rightSize, ShapedType::kDynamicSize);
+
+  if (isRightStaticDim) {  // Simplify code by set static to left if any
+    std::swap(leftSize, rightSize);
+    std::swap(leftBound, rightBound);
+    std::swap(isLeftStaticDim, isRightStaticDim);
+  }
+
+  bool isLeftStaticBound = leftBound != ShapedType::kDynamicSize;
+  bool isRightStaticBound = rightBound != ShapedType::kDynamicSize;
+  if (isLeftStaticDim) {
+    if (isRightStaticBound)
+      return std::make_pair(ShapedType::kDynamicSize, leftSize + rightBound);
+    else
+      return std::make_pair(ShapedType::kDynamicSize, ShapedType::kDynamicSize);
+  }
+  if (isLeftStaticBound && isRightStaticBound)
+    return std::make_pair(ShapedType::kDynamicSize, leftBound + rightBound);
+  return std::make_pair(ShapedType::kDynamicSize, ShapedType::kDynamicSize);
+}
+
+// Inference rules to merge dimensions with bounds (lhs/rhs are commutative):
+//       Dim of lhs     Dim of rhs      Infer
+//  c0:  X              X               X
+//  c1:  X              ?               X
+//  c2:  X              ?, B(>X)        X
+//  c3:  X              ?, B(<X)        Will error out by compatible checks
+//  c4:  ?              ?               ?
+//  c5:  ?              ?, B            ?, B
+//  c6:  ?, B           ?, B            ?, B
+//  c7:  ?, B1          ?, B2           ?, min(B1, B2)
+FailureOr<std::pair<int64_t, int64_t>> inferMergedDimAndBound(
+    Optional<Location> location, int64_t dim, int64_t leftSize,
+    int64_t rightSize, int64_t leftBound, int64_t rightBound) {
+  bool isLeftStaticDim = leftSize != ShapedType::kDynamicSize;
+  bool isRightStaticDim = rightSize != ShapedType::kDynamicSize;
+  if (isLeftStaticDim && isRightStaticDim && leftSize != rightSize)
+    return emitOptionalError(location, "Mismatch dimension size ", leftSize,
+                             " and ", rightSize, " in dimension ", dim);
+  if (isLeftStaticDim || isRightStaticDim)
+    return isLeftStaticDim
+               ? std::make_pair(leftSize, ShapedType::kDynamicSize)
+               : std::make_pair(rightSize, ShapedType::kDynamicSize);
+
+  bool isLeftStaticBound = leftBound != ShapedType::kDynamicSize;
+  bool isRightStaticBound = rightBound != ShapedType::kDynamicSize;
+  if (isLeftStaticBound && isRightStaticBound)
+    return std::make_pair(ShapedType::kDynamicSize,
+                          std::min(leftBound, rightBound));
+  if (isLeftStaticBound)
+    return std::make_pair(ShapedType::kDynamicSize, leftBound);
+  if (isRightStaticBound)
+    return std::make_pair(ShapedType::kDynamicSize, rightBound);
+  return std::make_pair(ShapedType::kDynamicSize, ShapedType::kDynamicSize);
+}
+
+LogicalResult inferMostSpecificType(
+    Optional<Location> location, ValueTypeRange<ValueRange> inputTypes,
+    SmallVectorImpl<Type>& inferredReturnTypes,
+    const SmallVector<int64_t>& concatenateDims) {
+  SmallVector<RankedTensorType> rankedTypes;
+  for (auto inputType : inputTypes)
+    if (auto rankedType = inputType.dyn_cast<RankedTensorType>())
+      rankedTypes.push_back(rankedType);
+  if (rankedTypes.empty()) {
+    inferredReturnTypes.push_back(inputTypes[0]);
+    return success();
+  }
+
+  auto rank = rankedTypes[0].getRank();
+  BoundedDialectInterface* dialect = nullptr;
+  SmallVector<int64_t> inferredDimSizes(rank, ShapedType::kDynamicSize);
+  SmallVector<int64_t> inferredBounds(rank, ShapedType::kDynamicSize);
+
+  for (const auto& it : llvm::enumerate(rankedTypes)) {
+    RankedTensorType rankedType = it.value();
+    ArrayRef<int64_t> bounds;
+    if (auto boundedAttr =
+            rankedType.getEncoding().dyn_cast_or_null<BoundedAttrInterface>()) {
+      dialect = cast<BoundedDialectInterface>(&boundedAttr.getDialect());
+      bounds = boundedAttr.getBounds();
+    } else if (rankedType.getEncoding()) {
+      // TODO(zhouxin) infer sparsity encoding after b/238903065 is fixed.
+      inferredReturnTypes.push_back(inputTypes[0]);
+      return success();
+    }
+
+    // Init the inferredDimSizes[dim] / inferredBounds[dim] with rankedTypes[0]
+    if (it.index() == 0) {
+      for (int dim = 0; dim < rank; ++dim) {
+        inferredDimSizes[dim] = rankedType.getDimSize(dim);
+        inferredBounds[dim] =
+            bounds.empty() ? ShapedType::kDynamicSize : bounds[dim];
+      }
+      continue;
+    }
+
+    // Infer each dim for current rankedType (from inputTypes[1:end])
+    for (int dim = 0; dim < rank; ++dim) {
+      std::pair<int64_t, int64_t> inferredDimAndBound;
+
+      int64_t leftSize = inferredDimSizes[dim];
+      int64_t rightSize = rankedType.getShape()[dim];
+      int64_t leftBound = inferredBounds[dim];
+      int64_t rightBound =
+          bounds.empty() ? ShapedType::kDynamicSize : bounds[dim];
+      if (std::find(concatenateDims.begin(), concatenateDims.end(), dim) !=
+          concatenateDims.end()) {
+        inferredDimAndBound = inferConcatenatedDimAndBound(
+            leftSize, rightSize, leftBound, rightBound);
+      } else {
+        auto inferredDimAndBoundOrErr = inferMergedDimAndBound(
+            location, dim, leftSize, rightSize, leftBound, rightBound);
+        if (failed(inferredDimAndBoundOrErr)) return failure();
+        inferredDimAndBound = *inferredDimAndBoundOrErr;
+      }
+      inferredDimSizes[dim] = inferredDimAndBound.first;
+      inferredBounds[dim] = inferredDimAndBound.second;
+    }
+  }
+
+  Attribute encoding = nullptr;
+  if (llvm::any_of(inferredBounds,
+                   [](auto el) { return el != ShapedType::kDynamicSize; })) {
+    encoding = dialect->createBoundedAttr(inferredBounds);
+  }
+  inferredReturnTypes.push_back(RankedTensorType::get(
+      inferredDimSizes, rankedTypes[0].getElementType(), encoding));
+
+  return success();
+}
+
 }  // namespace hlo
 }  // namespace mlir
--- a/stablehlo/dialect/Base.cpp
+++ b/stablehlo/dialect/Base.cpp
@@ -189,26 +189,19 @@ static std::pair<int64_t, int64_t> inferConcatenatedDimAndBound(
     int64_t rightBound) {
   bool isLeftStaticDim = leftSize != ShapedType::kDynamicSize;
   bool isRightStaticDim = rightSize != ShapedType::kDynamicSize;
-  if (isLeftStaticDim && isRightStaticDim)
-    return std::make_pair(leftSize + rightSize, ShapedType::kDynamicSize);
+  int64_t size = ShapedType::kDynamicSize;
+  int64_t bound = ShapedType::kDynamicSize;
 
-  if (isRightStaticDim) {  // Simplify code by set static to left if any
-    std::swap(leftSize, rightSize);
-    std::swap(leftBound, rightBound);
-    std::swap(isLeftStaticDim, isRightStaticDim);
+  if (isLeftStaticDim && isRightStaticDim) {
+    size = leftSize + rightSize;
+  } else {
+    int64_t leftSizeOrBound = isLeftStaticDim ? leftSize : leftBound;
+    int64_t rightSizeOrBound = isRightStaticDim ? rightSize : rightBound;
+    if (leftSizeOrBound != ShapedType::kDynamicSize &&
+        rightSizeOrBound != ShapedType::kDynamicSize)
+      bound = leftSizeOrBound + rightSizeOrBound;
   }
-
-  bool isLeftStaticBound = leftBound != ShapedType::kDynamicSize;
-  bool isRightStaticBound = rightBound != ShapedType::kDynamicSize;
-  if (isLeftStaticDim) {
-    if (isRightStaticBound)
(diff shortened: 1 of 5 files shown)


stablehlo/dialect/Base.cpp
```diff
@@ -176,5 +176,151 @@ Attribute boundsToEncoding(Attribute prototype, ArrayRef<int64_t> bounds) {
...
+//  c0:  X              Y               X+Y
+//  c1:  X              ?               ?
+//  c2:  X              ?, B            ?, X+B
+//  c3:  ?              ?               ?
+//  c4:  ?              ?, B            ?
+//  c5:  ?, B1          ?, B2           ?, B1+B2
+static std::pair<int64_t, int64_t> inferConcatenatedDimAndBound(
+    int64_t leftSize, int64_t rightSize, int64_t leftBound,
+    int64_t rightBound) {
+  bool isLeftStaticDim = leftSize != ShapedType::kDynamicSize;
```
smit-hinsu, line 190: how about something like,

size = ShapedType::kDynamicSize;
bound = ShapedType::kDynamicSize;

```
if (isLeftStaticDim && isRightStaticDim) {
  size = leftSize + rightSize;
} else {
  leftSizeOrBound = isLeftStaticDim ? leftSize : leftBound;
  rightSizeOrBound = isRightStaticDim ? rightSize : rightBound;
  if (leftSizeOrBound != ShapedType::kDynamicSize && 
      rightSizeOrBound != ShapedType::kDynamicSize)
    bound = leftSizeOrBound + rightSizeOrBound;
}
```

stablehlo/dialect/Base.h
```diff
@@ -51,6 +51,16 @@ bool isCompatibleForHloTypeInference(Type tp1, Type tp2);
...
 
+// Infer single most specific return type from inputTypes with support for
+// bounds. (Size, bound) of each dimension of the return type will be inferred
+// from corresponding dimensions of every inputType by merging them. If
+// `concatenateDims` is set, then the dimensions in `concatenateDims` will be
+// concatenated into return type instead.
+LogicalResult inferMostSpecificType(
+    Optional<Location> location, ValueTypeRange<ValueRange> inputTypes,
+    SmallVectorImpl<Type> &inferredReturnTypes,
+    const SmallVector<int64_t> &concatenateDims = {});
```
smit-hinsu, line 62: we will have at most one concat dims so better to take an optional.

stablehlo/dialect/Base.cpp
```diff
@@ -176,5 +176,151 @@ Attribute boundsToEncoding(Attribute prototype, ArrayRef<int64_t> bounds) {
...
+      dialect = cast<BoundedDialectInterface>(&boundedAttr.getDialect());
+      bounds = boundedAttr.getBounds();
+    } else if (rankedType.getEncoding()) {
+      // TODO(zhouxin) infer sparsity encoding after b/238903065 is fixed.
+      inferredReturnTypes.push_back(inputTypes[0]);
+      return success();
+    }
+
+    // Init the inferredDimSizes[dim] / inferredBounds[dim] with rankedTypes[0]
+    if (it.index() == 0) {
```
smit-hinsu, line 281: I think we don't really need to have this special case and the rest of the logic will take care of this as well?

stablehlo/dialect/Base.cpp
```diff
@@ -176,5 +176,151 @@ Attribute boundsToEncoding(Attribute prototype, ArrayRef<int64_t> bounds) {
...
+  SmallVector<int64_t> inferredBounds(rank, ShapedType::kDynamicSize);
+
+  for (const auto& it : llvm::enumerate(rankedTypes)) {
+    RankedTensorType rankedType = it.value();
+    ArrayRef<int64_t> bounds;
+    if (auto boundedAttr =
+            rankedType.getEncoding().dyn_cast_or_null<BoundedAttrInterface>()) {
+      dialect = cast<BoundedDialectInterface>(&boundedAttr.getDialect());
+      bounds = boundedAttr.getBounds();
+    } else if (rankedType.getEncoding()) {
```
smit-hinsu, line 274: nit: use a variable for encoding.

stablehlo/dialect/Base.cpp
```diff
@@ -176,5 +176,151 @@ Attribute boundsToEncoding(Attribute prototype, ArrayRef<int64_t> bounds) {
...
+        if (failed(inferredDimAndBoundOrErr)) return failure();
+        inferredDimAndBound = *inferredDimAndBoundOrErr;
+      }
+      inferredDimSizes[dim] = inferredDimAndBound.first;
+      inferredBounds[dim] = inferredDimAndBound.second;
+    }
+  }
+
+  Attribute encoding = nullptr;
+  if (llvm::any_of(inferredBounds,
```
smit-hinsu, line 315: we could have a helper function to get encoding attribute from bounds? it will be a tiny helper but will improve the code readability.

stablehlo/tests/infer_stablehlo.mlir
```diff
@@ -693,3 +693,108 @@ func.func @pad_with_negative_inferred_bounds(%arg0: tensor<3x?x?xf16, #stablehlo
   %1 = "hlo_test_infer.get_return_types"(%0) : (tensor<*xf16>) -> tensor<*xindex>
   func.return %1 : tensor<*xindex>
 }
+
+// -----
+
+// CHECK-LABEL: @concat_bounds_c0
```
smit-hinsu, line 699: it will be easier to follow if we put all c{0...5} in a single test with multiple dims?

Add one test with unranked as well?

smit-hinsu (commented):
Thanks!

stablehlo/dialect/Base.cpp
zhouxin913: Nice code! Replaced and thanks.

stablehlo/tests/infer_stablehlo.mlir
zhouxin913: "with multiple dims" This is possible for `AddOp` but not for `ConcatenateOp`: each test can only contain one "concat" dim thus there's no way merge them into one big tensor to include every cases.

Unranked case is not related to bounds, right? There is a test here: https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/ops_stablehlo.mlir#L872

stablehlo/dialect/Base.h
zhouxin913: Yes changed to from `const SmallVector<int64_t> &concatenateDims = {}` to `int64_t concatenateDim =-1`

stablehlo/dialect/Base.cpp
zhouxin913: We can initialize inferredDimSizes[i] = **-1** then deal with it again all input types, as -1 is an identical element for AppOp with rule:
```
     Dim of lhs     Dim of rhs      Infer
c1:  X              ?               X
```
But for ConcatenateOp, **0** should be the identical element but -1 is not:
```
     Dim of lhs     Dim of rhs      Infer
c1:  X              ?               ? (instead of X)
```
Thus I init `inferredDimSizes\inferredBounds` with rankedTypes[0], and loop from rankedTypes[1~end]

stablehlo/dialect/Base.cpp
zhouxin913: New commit is much refined with our existing `encodingToBounds() / boundsToEncoding()` utils.

stablehlo/dialect/Base.cpp
zhouxin913: New commit is much refined with our existing `encodingToBounds() / boundsToEncoding()` utils.

stablehlo/dialect/Base.cpp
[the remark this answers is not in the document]
zhouxin913: Nice advise! There is already an
```
inline static bool isDynamicDimSize(int64_t val) {
  return val == ShapedType::kDynamicSize;
}
``` in `TypeInference.h`. Let me try to reuse it~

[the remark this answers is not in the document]
Comment (GleasonK) [reply stablehlo/dialect/Base.cpp:190]: Nice! Maybe rename it to `isDynamicSize` because we are using it on both bounds and dims by the looks of it?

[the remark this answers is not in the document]
Comment (burmako) [reply stablehlo/dialect/Base.cpp:190]: Good idea! Also, maybe we could have both `isStaticSize` and `isDynamicSize`?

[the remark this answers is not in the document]
Comment (zhouxin913) [reply stablehlo/dialect/Base.cpp]: Added a TODO for it: as it happens in a lot of places and could be done in a follow up PR. Thanks!

File: stablehlo/dialect/Base.h [hunk 7]
```diff
@@ -51,6 +57,16 @@ bool isCompatibleForHloTypeInference(Type tp1, Type tp2);
 // sparsity.
 bool isCompatibleForHloTypeInference(TypeRange tp1, TypeRange tp2);
 
+// Infer single most specific return type from inputTypes with support for
+// bounds. (Size, bound) of each dimension of the return type will be inferred
+// from corresponding dimensions of every inputType by merging them. If
```
smit-hinsu, line 62: In case of concatenate dim, add the requirement that all types must be ranked? Also, add invariance check for that as well.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -574,60 +574,16 @@ LogicalResult inferConcatenateOp(Optional<Location> location, ValueRange inputs,
...
-      // If a dimension is not dynamic, the output shape should match.
-      if (ShapedType::isDynamic(outShape[it.index()])) {
-        outShape[it.index()] = it.value();
-      }
-    }
-  }
-
-  outShape[dimension] = 0;
-
   for (auto operand : inputs.getTypes()) {
```
smit-hinsu, line 577: isn't this conservative compared to binary op?

In case of inputs tensor<2x?xf32> and tensor<*xf32> with dim = 1, we know that the result type for legal cases would be tensor<2x?xf32>.

stablehlo/dialect/Base.h
```diff
@@ -51,6 +57,16 @@ bool isCompatibleForHloTypeInference(Type tp1, Type tp2);
...
+// `concatenateDim` is set other than default -1, then that dimensions will be
+// concatenated into return type instead.
+LogicalResult inferMostSpecificType(Optional<Location> location,
```
smit-hinsu, line 65: I have mixed emotions of using the same helper for both binary and concatenation, this helps avoid duplication but also a bit convoluted. Like, concat does all checks and handles unranked before this but check for bounds happen here.

stablehlo/dialect/Base.cpp
```diff
@@ -176,5 +175,137 @@ Attribute boundsToEncoding(Attribute prototype, ArrayRef<int64_t> bounds) {
...
+  }
+
+  inferredReturnTypes.push_back(RankedTensorType::get(
+      inferredDimSizes, rankedTypes[0].getElementType(),
+      boundsToEncoding(
+          rankedTypes[0].getEncoding(),
+          // Empty array as argument is an indicator to boundsToEncoding() that
+          // there are no bounds at all in inputs, thus sparsity attributes will
+          // be included in the return type
+          anyInputHaveBounds ? inferredBounds : llvm::ArrayRef<int64_t>({}))));
```
smit-hinsu, line 305: Seems like we didn't propagate sparse encoding for binary ops earlier? What should be the logic there in case of different sparse encoding for lhs and rhs? Could we add a test there? Same for concatenate op.

stablehlo/tests/infer_stablehlo.mlir
smit-hinsu: sorry, I was mistaken on combining the tests. 

yes, it is not related to bounds but we need some test for shape inference as well. Also, let's keep all concatenate ops together. There is `func.func @concat` earlier in this file.

optional: Instead of c0 to c5, consider describing in name or comment about what that particular test is testing. may be like static size with static bound.

stablehlo/dialect/Base.cpp
smit-hinsu: ah, right. 

this can be done outside of the loop with

auto inferredDimSizes = llvm::to_vector<4>(rankedType.getShape())

Similarly for bounds.

smit-hinsu (approved):

burmako (changes requested):

stablehlo/dialect/Base.cpp
zhouxin913: makes sense! Fixed.

stablehlo/tests/infer_stablehlo.mlir
zhouxin913: Sure reviewers. Added the comment of c0~5 rules in test as well and also refer it to `Base.cpp`.

"we need some test for shape inference as well" yes we do. The `concat_1D_unranked` test case is for unranked case. I think you are looking for a test with `hlo_test_infer.get_return_types` and CHECK `type0 = tensor<*x>`. Now I need think about whether we need this (all ops missing this) in later PRs.

"keep all concatenate ops together." It is a good idea. But now every bounded tests are put under spliter:
//===------------===//
// Bounded Dynamism
//===------------===//
So let leave it to future work on how to scatter the tests into proper different files.

stablehlo/dialect/Base.h
zhouxin913: in `inferMostSpecificType()` if all the types are unranked, it will return `inputTypes[0]`.  So far I think it's enough even for concat dim. The callers can pass in ranked or unranked as well with no worries.

stablehlo/dialect/Base.h
```diff
@@ -42,6 +42,12 @@ limitations under the License.
 namespace mlir {
 namespace hlo {
 
+// TODO(zhouxin) change to a better name as it's used by both of size and bound
```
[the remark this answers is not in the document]
zhouxin913: Added.

stablehlo/dialect/Base.cpp
```diff
@@ -176,5 +175,137 @@ Attribute boundsToEncoding(Attribute prototype, ArrayRef<int64_t> bounds) {
   return dialect->createBoundedAttr(bounds);
 }
 
+// Inference rules to concat dimensions with bounds (lhs/rhs are commutative):
```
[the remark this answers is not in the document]
zhouxin913: Good advise! Fixed.

stablehlo/dialect/Base.cpp
```diff
@@ -176,5 +175,137 @@ Attribute boundsToEncoding(Attribute prototype, ArrayRef<int64_t> bounds) {
...
+      bound = leftSizeOrBound + rightSizeOrBound;
+  }
+  return std::make_pair(size, bound);
+}
+
+// Inference rules to merge dimensions with bounds (lhs/rhs are commutative):
+//       Dim of lhs     Dim of rhs      Infer
+//  c0:  X              X               X
+//  c1:  X              ?               X
+//  c2:  X              ?, B(>X)        X
```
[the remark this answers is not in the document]
zhouxin913: Fixed.

stablehlo/dialect/Base.cpp
```diff
@@ -176,5 +175,137 @@ Attribute boundsToEncoding(Attribute prototype, ArrayRef<int64_t> bounds) {
...
+//  c3:  X              ?, B(<X)        Will error out by compatible checks
+//  c4:  ?              ?               ?
+//  c5:  ?              ?, B            ?, B
+//  c6:  ?, B           ?, B            ?, B
+//  c7:  ?, B1          ?, B2           ?, min(B1, B2)
```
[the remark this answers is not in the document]
zhouxin913: Right. Add comment fixed. No code change needed.

stablehlo/dialect/Base.cpp
```diff
@@ -176,5 +175,137 @@ Attribute boundsToEncoding(Attribute prototype, ArrayRef<int64_t> bounds) {
...
+  bool isRightStaticBound = !isDynamicDimSize(rightBound);
+  if (isLeftStaticBound && isRightStaticBound)
+    return std::make_pair(ShapedType::kDynamicSize,
+                          std::min(leftBound, rightBound));
+  if (isLeftStaticBound)
+    return std::make_pair(ShapedType::kDynamicSize, leftBound);
+  if (isRightStaticBound)
+    return std::make_pair(ShapedType::kDynamicSize, rightBound);
+  return std::make_pair(ShapedType::kDynamicSize, ShapedType::kDynamicSize);
+}
```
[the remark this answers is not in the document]
zhouxin913: Thanks for the advise, and used, especially the 2 error messages here. They are useful for the completeness of this standalone function, tho both of them have not show up in the tests yet: now Binary ops checks the correctness of operands outside already.

stablehlo/dialect/Base.cpp
```diff
@@ -176,5 +175,137 @@ Attribute boundsToEncoding(Attribute prototype, ArrayRef<int64_t> bounds) {
...
+    if (auto rankedType = inputType.dyn_cast<RankedTensorType>())
+      rankedTypes.push_back(rankedType);
+  if (rankedTypes.empty()) {
+    inferredReturnTypes.push_back(inputTypes[0]);
+    return success();
+  }
+
+  auto rank = rankedTypes[0].getRank();
+  // BoundedDialectInterface* dialect = nullptr;
+  SmallVector<int64_t> inferredDimSizes(rank, ShapedType::kDynamicSize);
```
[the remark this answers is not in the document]
zhouxin913: Yeah makes senses. Fixed.

stablehlo/dialect/Base.cpp
```diff
@@ -176,5 +175,137 @@ Attribute boundsToEncoding(Attribute prototype, ArrayRef<int64_t> bounds) {
...
+  }
+  return std::make_pair(size, bound);
+}
+
+// Inference rules to merge dimensions with bounds (lhs/rhs are commutative):
+//       Dim of lhs     Dim of rhs      Infer
+//  c0:  X              X               X
+//  c1:  X              ?               X
+//  c2:  X              ?, B(>X)        X
+//  c3:  X              ?, B(<X)        Will error out by compatible checks
```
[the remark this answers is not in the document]
zhouxin913: Fixed.

stablehlo/dialect/TypeInference.cpp
zhouxin913: You example is very good: inputs: 
<2x?xf32>, <*xf32> with concat dim 1 should infer <2x?xf32>
This can be addressed by ingoreing all the unranked inputs and only the the ranked ones.
But soon I realize the above claim is wrong: for example in case of
<2x?xf32>, <*xf32> with concat dim 0 should infer <?x?xf32> instead of <2x?xf32>
My solution to all these are: treat unranked input type as a fully dynamic type <?x...?x>. Please see the new commit.f

stablehlo/dialect/Base.h
zhouxin913: Thanks for the discussions! After deep thinking, I estimate that benefits is much less then the shortage to use a common function for both cases. So, I split it in the new commit. Let me answer the feedback first:
0. The `FailureOr<Type> mergeTypes()` interface above looks a clean kill. But in fact the concat op need both util methods `inferConcatenatedDimAndBound` for the concat_dim and `inferMergedDimAndBound` for the rest N-1 dims. Thus we have to pass in 2 functions instead of 1.
2 & 3: Due to my schedule, I put them into future "TODO(zhouxin) Refactor to better handle errors and return single type" in  `Base.cpp`
4. Fixed.

The biggest change in the newest commit is the full support of "mixed of unranked and ranked cases". I think this behavior issue is of top importance of this PR. See this [In the other comment above ](https://github.com/openxla/stablehlo/pull/389#discussion_r1008943148)

stablehlo/dialect/Base.cpp
zhouxin913: We do, and used the rankTypes[0].encoding() for the return encoding. 
Now for both in Binary and Concat op, we both keep using `rankTypes[0].encoding()`. But, I think this is an open question: let's track in https://github.com/openxla/stablehlo/issues/407

stablehlo/dialect/Base.cpp
```diff
@@ -176,5 +175,137 @@ Attribute boundsToEncoding(Attribute prototype, ArrayRef<int64_t> bounds) {
...
+  if (isLeftStaticDim && isRightStaticDim) {
+    size = leftSize + rightSize;
+  } else {
+    int64_t leftSizeOrBound = isLeftStaticDim ? leftSize : leftBound;
+    int64_t rightSizeOrBound = isRightStaticDim ? rightSize : rightBound;
+    if (!isDynamicDimSize(leftSizeOrBound) &&
+        !isDynamicDimSize(rightSizeOrBound))
+      bound = leftSizeOrBound + rightSizeOrBound;
+  }
+  return std::make_pair(size, bound);
```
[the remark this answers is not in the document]
zhouxin913: Thanks! it works partially. When the return type is `FailureOr<std::pair<int64_t, int64_t>>` it can't automatically convert. Let me figure out this in the new added TODOs about refactoring the error handling.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -574,60 +574,16 @@ LogicalResult inferConcatenateOp(Optional<Location> location, ValueRange inputs,
   auto firstType = inputs[0].getType().cast<ShapedType>();
   auto outElement = firstType.getElementType();
```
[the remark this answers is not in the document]
zhouxin913: Changed. Code are much refactored.

stablehlo/dialect/Base.cpp
```diff
@@ -176,5 +175,137 @@ Attribute boundsToEncoding(Attribute prototype, ArrayRef<int64_t> bounds) {
...
+  SmallVector<RankedTensorType> rankedTypes;
+  for (auto inputType : inputTypes)
+    if (auto rankedType = inputType.dyn_cast<RankedTensorType>())
+      rankedTypes.push_back(rankedType);
+  if (rankedTypes.empty()) {
+    inferredReturnTypes.push_back(inputTypes[0]);
+    return success();
+  }
+
+  auto rank = rankedTypes[0].getRank();
```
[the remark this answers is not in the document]
zhouxin913: Code are much refactored and this become out-of-date.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -530,10 +530,10 @@ LogicalResult inferCaseOp(Optional<Location> location, RegionRange branches,
 LogicalResult inferConcatenateOp(Optional<Location> location, ValueRange inputs,
                                  int64_t dimension,
                                  SmallVectorImpl<Type>& inferredReturnTypes) {
-  RankedTensorType firstRankedType;
-  int firstRankedIndex = -1;
   if (dimension < 0)
     return emitOptionalError(location, "dimension ", dimension, " is negative");
+  RankedTensorType firstRankedType;
+  int firstRankedIndex = -1;
   for (uint64_t i = 0; i < inputs.size(); i++) {
```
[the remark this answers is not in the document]
zhouxin913: Code are much refactored and this become out-of-date.

stablehlo/tests/infer_stablehlo.mlir
```diff
@@ -693,3 +693,150 @@ func.func @pad_with_negative_inferred_bounds(%arg0: tensor<3x?x?xf16, #stablehlo
...
+
+// Alse see this in Base.cpp
+// Inference rules to concat dimensions with bounds (lhs/rhs are commutative):
+//       Dim of lhs     Dim of rhs      Infer
+//  c0:  X              Y               X+Y
+//  c1:  X              ?               ?
+//  c2:  X              ?, B            ?, X+B
+//  c3:  ?              ?               ?
+//  c4:  ?              ?, B            ?
+//  c5:  ?, B           ?, C            ?, B+C
```
burmako, line 707: Add a newline here to separate comments and tests.

burmako, line 707: Also, let's simplify ranked tests by only using one dimension, i.e. instead of `5x1xi32` and `dimension = 1`, let's use `1xi32` and `dimension = 0` etc.

stablehlo/tests/infer_stablehlo.mlir
```diff
@@ -693,3 +693,150 @@ func.func @pad_with_negative_inferred_bounds(%arg0: tensor<3x?x?xf16, #stablehlo
...
+func.func @concat_bounds_unranked_c1(
+  %arg0: tensor<*xi32>, 
+  %arg1: tensor<5x?xi32, #stablehlo.type_extensions<bounds = [-1, 4]>>)  -> tensor<*xindex> {
+  %result = "stablehlo.concatenate"(%arg0, %arg1) { dimension = 1 : i64 } : (
+    tensor<*xi32>, 
+    tensor<5x?xi32, #stablehlo.type_extensions<bounds = [-1, 4]>>) -> tensor<5x?xi32>
+  // CHECK: types0 = tensor<5x?xi32>
+  %1 = "hlo_test_infer.get_return_types"(%result) : (tensor<5x?xi32>) -> tensor<*xindex> 
+  func.return %1 : tensor<*xindex>
+}
```
burmako, line 842: Typo: no newline at the end of the file.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -571,62 +571,64 @@ LogicalResult inferConcatenateOp(Optional<Location> location, ValueRange inputs,
...
-    auto shapedTy = type.cast<ShapedType>();
-    if (!shapedTy.hasRank()) {
-      continue;
-    }
-
-    for (const auto& it : llvm::enumerate(shapedTy.getShape())) {
-      // If a dimension is not dynamic, the output shape should match.
-      if (ShapedType::isDynamic(outShape[it.index()])) {
-        outShape[it.index()] = it.value();
+  // Infer the most specific (size, bound) of all dimenstions of the return type
```
burmako, line 580: Typo: "dimensions".

stablehlo/dialect/TypeInference.cpp
```diff
@@ -571,62 +571,64 @@ LogicalResult inferConcatenateOp(Optional<Location> location, ValueRange inputs,
...
+  auto rank = firstRankedType.getRank();
+  SmallVector<int64_t> inferredSizes(rank, ShapedType::kDynamicSize);
+  SmallVector<int64_t> inferredBounds(rank, ShapedType::kDynamicSize);
+  // Note: for the concatenate dimension, 0 should be the identical element:
```
burmako, line 584: Typo: "identity element".

stablehlo/dialect/TypeInference.cpp
```diff
@@ -571,62 +571,64 @@ LogicalResult inferConcatenateOp(Optional<Location> location, ValueRange inputs,
...
+  // c1: (<5x?xf32>, <*xf32>) with concat dim 1 should infer <5x?xf32>
+  // Instead, they should be replaced with dynamic tensors: tensor<?x...?x>
+  for (const auto& it : llvm::enumerate(inputs.getTypes())) {
+    RankedTensorType rankedType = it.value().dyn_cast<RankedTensorType>();
+    SmallVector<int64_t> bounds;
+    if (rankedType)
+      bounds = to_vector(encodingToBounds(rankedType.getEncoding()));
+    if (!bounds.empty()) anyInputHaveBounds = true;
+
+    // Infer each dim for current rankedType (from inputTypes[1:end])
```
burmako, line 600: I don't think that "from inputTypes[1:end]" applies anymore.

stablehlo/dialect/Base.cpp
```diff
@@ -176,5 +175,126 @@ Attribute boundsToEncoding(Attribute prototype, ArrayRef<int64_t> bounds) {
...
+  } else {
+    if (isLeftStaticBound && isRightStaticBound)
+      bound = std::min(leftBound, rightBound);
+    else
+      bound = isLeftStaticBound ? leftBound : rightBound;
+  }
+  return std::make_pair(size, bound);
+}
+
+// TODO(zhouxin) Refactor to better handle errors and return single type
```
burmako, line 248: Let's open a ticket about this TODO, so that we don't forget what it means, perhaps with a reference to this PR.

stablehlo/dialect/Base.h
burmako: To be honest, I liked the idea of sharing code between binary and concat, because this way we make sure that subtle logic like `<5x?xf32> op <*xf32>` is handled correctly. However given the schedule restrictions, I'm fine with merging as is and leaving follow-up to a ticket.

burmako (approved):

stablehlo/dialect/TypeInference.cpp
zhouxin913: Done.

stablehlo/dialect/Base.cpp
zhouxin913: Sure opened https://github.com/openxla/stablehlo/issues/408

stablehlo/dialect/TypeInference.cpp
zhouxin913: Done.

stablehlo/dialect/TypeInference.cpp
zhouxin913: Done.

stablehlo/tests/infer_stablehlo.mlir
zhouxin913: Done.

stablehlo/tests/infer_stablehlo.mlir
zhouxin913: That is done on purpose for this part of code in `inferConcatenateOp()`:
```
if (dim == dimension) {
  inferredDimAndBound = inferConcatenatedDimAndBound(
      leftSize, rightSize, leftBound, rightBound);
} else {
  auto inferredDimAndBoundOrErr = inferMergedDimAndBound(
      location, dim, leftSize, rightSize, leftBound, rightBound);
  if (failed(inferredDimAndBoundOrErr)) return failure();
  inferredDimAndBound = *inferredDimAndBoundOrErr;
}
```
Removing that `5` dim will weaken the tests for no covering the `else` part above.

zhouxin913: Sounds good.

stablehlo/dialect/Base.h
zhouxin913: Let's follow up in https://github.com/openxla/stablehlo/issues/409. Thanks for reviewing!

stablehlo/tests/infer_stablehlo.mlir
burmako: Got it, let's keep the tests as is then.

Issue #391, opened 2022-10-25
Converge type inference for MHLO and StableHLO ReduceOp


burmako:
https://github.com/tensorflow/mlir-hlo/commit/68901a9e9d5f52969b4bbb1b4f20f7246ff3a118 has made divergent changes to type inference for MHLO and StableHLO ReduceOp, and we need to converge them back.

PR #392, opened 2022-10-26
Remove ReduceOp::build() to sync with MLIR-HLO

zhouxin913:
fix https://github.com/openxla/stablehlo/issues/391
Please see the detailed link in the related bug.

### DIFF
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -2759,19 +2759,6 @@ static TensorType getReduceResultType(Type operandTy,
   return RankedTensorType::get(shape, elementTy);
 }
 
-void ReduceOp::build(OpBuilder& builder, OperationState& state,
-                     ValueRange operands, ValueRange initValues,
-                     DenseIntElementsAttr dimensions) {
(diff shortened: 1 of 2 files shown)


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2759,19 +2759,6 @@ static TensorType getReduceResultType(Type operandTy,
...
-  SmallVector<Type, 1> resultTy;
-  resultTy.reserve(operands.size());
-
-  for (Value operand : operands) {
-    resultTy.push_back(
-        getReduceResultType(operand.getType(), dimensions, &builder));
-  }
-  build(builder, state, resultTy, operands, initValues, dimensions);
-}
-
```
burmako, line 2774: Need to remove the builder as well from the ODS.

burmako (changes requested):

stablehlo/dialect/StablehloOps.cpp
zhouxin913: I thought it was about another builder, but it turns out it's not. Removed, thanks!

burmako (approved):

PR #393, opened 2022-10-26
Integrate LLVM at llvm/llvm-project@12204429f2f9

subhankarshah:


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2759,19 +2759,6 @@ static TensorType getReduceResultType(Type operandTy,
...
-  SmallVector<Type, 1> resultTy;
-  resultTy.reserve(operands.size());
-
-  for (Value operand : operands) {
-    resultTy.push_back(
-        getReduceResultType(operand.getType(), dimensions, &builder));
-  }
-  build(builder, state, resultTy, operands, initValues, dimensions);
-}
-
```
burmako, line 2774: I've just merged #392, so if you rebase, you shouldn't need these diffs anymore.

burmako (changes requested):

stablehlo/dialect/StablehloOps.cpp
subhankarshah: Done

burmako (approved):

PR #397, opened 2022-10-27
Fix error handling and add testing for printDenseI64Array

GleasonK:
The code `DenseI64ArrayAttr::parse(parser, Type{}).dyn_cast<DenseI64ArrayAttr>();` is problematic since if parsing fails, the null attribute will crash on `dyn_cast`.

Move the `dyn_cast` to after error handling. Add test case.

### DIFF
--- a/stablehlo/dialect/AssemblyFormat.cpp
+++ b/stablehlo/dialect/AssemblyFormat.cpp
@@ -17,6 +17,7 @@ limitations under the License.
 
 #include "llvm/Support/ErrorHandling.h"
 #include "llvm/Support/Regex.h"
+#include "mlir/IR/BuiltinAttributes.h"
 #include "stablehlo/dialect/Base.h"
 
 namespace mlir {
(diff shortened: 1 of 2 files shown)


stablehlo/dialect/AssemblyFormat.cpp
```diff
@@ -286,12 +286,12 @@ void printDenseI64Array(OpAsmPrinter& p, Operation* op,
 
 ParseResult parseDenseI64Array(OpAsmParser& parser,
                                DenseIntElementsAttr& attr) {
-  DenseI64ArrayAttr arrayAttr =
-      DenseI64ArrayAttr::parse(parser, Type{}).dyn_cast<DenseI64ArrayAttr>();
```
burmako, line 290: Let's perhaps use dyn_cast_or_null instead?

burmako (changes requested):

stablehlo/dialect/AssemblyFormat.cpp
GleasonK: Ah - that's how that function is used. Fixed.

Issue #351, opened 2022-10-17
Allow uniform naming across various StableHLO artifacts


sdasgup3:
### Request description

The need for such a ticket initiated in https://github.com/openxla/stablehlo/pull/335#issuecomment-1279538623.

Let us use this ticket to collect opportunities to apply uniform naming across various StableHLO artifacts like the entities in the ODS.

From https://github.com/openxla/stablehlo/pull/335#pullrequestreview-1143225138

```
... ensure that: 1) the opset has uniform naming style, 2) all C++/Python breaking changes related to naming are done at once instead of spread over weeks or months of breakages.
```

ghpvnist:
Some ops have inconsistent naming with our standard e.g. using "output" instead of "result" in the ODS. The spec, ODS example, and pretty-print currently defaults to "result" even if the ODS name is "output". These ops are (in no particular order):
* Constant
* Iota
* DynamicIota
* CreateToken
* BatchNormTraining
* DynamicBroadcastInDim
* DynamicReshape
* RngBitGenerator
* ReducePrecision

burmako:
Looks like `log_plus_one` is a misnomer. Unlike `exponential_minus_one`, which computes `exp(x) - 1`, this op computes `log(1 + x)`, so I think that it should be renamed to `log_one_plus`.

burmako:
Some more ideas for standardizing names:
  1. input vs operand (output vs result already called out above).
  2. Consistent use of singulars and plurals (e.g. TupleOp::val should be vals, etc).
  3. IotaOp::iota_dimension (and multiple similar cases) vs ReverseOp::dimensions.
  4. Specifying the "result" name explicitly vs relying on TableGen to provide name it implicitly.
  5. Same thing as above but for derived unary and binary ops.
  6. IsFiniteOp:x/y - can we use standard names?
  7. CompareOp::compare_type - why not comparison_type, given that this is how the enum is called?
  8. CholeskyOp::a - can we use standard name?
  9. DotGeneralOp::dot_dimension_numbers vs ConvolutionOp::dimension_numbers.
  10. CompareOp::comparison_direction/FftOp::fft_type/etc vs just direction/type/etc.
  11. Compare names with HLO and see if that exposes more inconsistencies. Would be good to align with HLO if possible.
  12. Compare op names and op mnemonics. E.g. why SinOp has the "sine" mnemonic? Can we have op names and op mnemonics be the same (modulo case).

ghpvnist:
[...]

PR #398, opened 2022-10-27
Add spec for BatchNormTrainingOp

### LINKED ISSUE #346: Add spec for BatchNormTrainingOp
### Request description

To spec stablehlo.batch_norm_training

subhankarshah:
closes #346


docs/spec_draft.md
```diff
@@ -343,6 +344,67 @@ logical operation.
...
+// %operand: [
+//            [[1.0, 2.0], [3.0, 4.0]],
+//            [[3.0, 4.0], [1.0, 2.0]]
+//           ]
+// %scale: [1.0, 1.0]
+// %offset: [1.0, 1.0]
+%results:3 = "mhlo.batch_norm_training"(%operand, %scale, %offset) {
+  epsilon = 0.0 : f32,
+  feature_index = 1 : i64
+} : (tensor<2x2x2xf32>, tensor<2xf32>, tensor<2xf32>, tensor<2xf32>, tensor<2xf32>) -> tensor<2x2x2xf32>
```
sdasgup3, line 397: fix the arguments.

docs/spec_draft.md
```diff
@@ -343,6 +344,67 @@ logical operation.
...
+| Name         | Type                                        |
+|--------------|---------------------------------------------|
+| `output`     | tensor of floating-point type               |
+| `batch_mean` | 1-dimensional tensor of floating-point type |
+| `batch_var`  | 1-dimensional tensor of floating-point type |
+
+### Constraints
+
+  * (C1) 0 $\le$ `feature_index` $\lt$ rank(`operand`).
+  * (C2) 0 $\lt$ rank(`operand`).
```
sdasgup3, line 378: C2 redundant as it can be derived from C1.

stablehlo/dialect/StablehloOps.td
```diff
@@ -1607,9 +1607,20 @@ def StableHLO_BatchNormTrainingOp : StableHLO_Op<"batch_norm_training",
...
+    See:
+    https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md#stablehlobatch_norm_training
 
-    See https://www.tensorflow.org/xla/operation_semantics#batchnormtraining
+    Example:
+    ```mlir
+    %results:3 = "mhlo.batch_norm_training"(%operand, %scale, %offset) {
+      epsilon = 0.0 : f32,
+      feature_index = 1 : i64
+    } : (tensor<2x2x2xf32>, tensor<2xf32>, tensor<2xf32>, tensor<2xf32>, tensor<2xf32>) -> tensor<2x2x2xf32>
```
sdasgup3, line 1622: fix arguments.

docs/spec_draft.md
```diff
@@ -343,6 +344,66 @@ logical operation.
...
 
+## stablehlo.batch_norm_training
+
+### Semantics
+
+Computes mean and variance across batch and spatial dimensions and normalizes
+the `operand` tensor, for each feature in the `feature_index` dimension and
+produces `output`, `batch_mean` and `batch_var` tensors. Refer
+[Batch Normalization Paper](https://arxiv.org/abs/1502.03167) for the detailed
+algorithm.
```
subhankarshah, line 355: image

How about saying this Batch Norm Transform from the paper

subhankarshah: @sdasgup3 reworded the semantics

sdasgup3 (approved):

sdasgup3:
As per our conversation, this op will be referred to by the BatchNorm Inference highlighting the changes needed.

docs/spec_draft.md
burmako: As discussed offline, the current plan is to reformulate the Semantics section similarly to `batch_norm_inference`: https://github.com/openxla/stablehlo/pull/377#discussion_r1013499398. I'll wait for the reformulation before reviewing the section.

docs/spec_draft.md
```diff
@@ -343,6 +344,67 @@ logical operation.
...
+
+| Name         | Type                                        |
+|--------------|---------------------------------------------|
+| `output`     | tensor of floating-point type               |
+| `batch_mean` | 1-dimensional tensor of floating-point type |
+| `batch_var`  | 1-dimensional tensor of floating-point type |
+
+### Constraints
+
+  * (C1) 0 $\le$ `feature_index` $\lt$ rank(`operand`).
```
burmako, line 378: Similarly to https://github.com/openxla/stablehlo/pull/377#discussion_r1013442113, let's add another constraint here that says:

```
  * (C2) `operand`, `scale`, `offset`, `mean`, variance` and `result` have the 
    same element type.
```

docs/status.md
```diff
@@ -42,7 +42,7 @@ one of the following tracking labels.
 | atan2                    | no            | yes*         | yes*           | yes             | no          |
 | batch_norm_grad          | no            | yes*         | yes*           | no              | no          |
 | batch_norm_inference     | no            | yes*         | yes*           | no              | no          |
-| batch_norm_training      | no            | yes*         | yes*           | no              | no          |
+| batch_norm_training      | yes           | yes          | yes            | no              | no          |
```
burmako, line 45: Similarly to https://github.com/openxla/stablehlo/pull/377#discussion_r1013445041, let's change verification to `revisit` while we're figuring out #369.

docs/spec_draft.md
```diff
@@ -343,6 +344,67 @@ logical operation.
...
+%results:3 = "stablehlo.batch_norm_training"(%operand, %scale, %offset) {
+  epsilon = 0.0 : f32,
+  feature_index = 2 : i64
+} : (tensor<2x2x2xf32>, tensor<2xf32>, tensor<2xf32>) -> (tensor<2x2x2xf32>, tensor<2xf32>, tensor<2xf32>)
+// %results#0: [
+//              [[0, 0], [2, 2]],
+//              [[2, 2], [0, 0]]
+//             ]
+// %results#1: [2, 3]
+// %results#2: [1, 1]
```
burmako, line 403: Similarly to https://github.com/openxla/stablehlo/pull/377#discussion_r1013464149, since we're using floating-point notation above, let's use it for the example output as well.

burmako, line 403: Let's use the same example across `batch_norm_inference` and `batch_norm_training` to emphasize how closely they are related. More concretely, let's take the example for `batch_norm_training` and copy/paste its operand, scale, offset, output, batch_var and batch_mean into `batch_norm_inference`.

burmako (changes requested):

docs/spec_draft.md
subhankarshah: The outputs of the interpreter were without decimal point so I removed to be precise, will add it again.

docs/spec_draft.md
```diff
@@ -348,6 +349,93 @@ logical operation.
...
+  # Compute `mean` and `variance` (as 1-dimensional tensors)
+  mean = compute_mean(operand, feature_index)
+  variance = compute_variance(operand, feature_index)
+
+  # Perform normalization with computed `mean` and `variance` instead of
+  # depending on `mean` and `variance` provided in the input like
+  # `batch_norm_inference`
+  return batch_norm_inference(operand, scale, offset, mean,
+                              variance, epsilon, feature_index)
+```
```
burmako, line 383: Proposed refactoring of the code:

```Python
def compute_mean(operand, feature_index):
  (sum,) = reduce(
      inputs=[operand],
      init_values=[0],
      dimensions=[i for i in range(rank(operand)) if i != feature_index],
      body=lambda x, y: add(x, y))
  divisor = constant(num_elements(operand) / dim(operand, feature_index))
  divisor_bcast = broadcast_in_dim(mean, [], shape(sum))
  return divide(sum, divisor_bcast)

def compute_variance(operand, feature_index):
  mean = compute_mean(operand, feature_index)
  mean_bcast = broadcast_in_dim(mean, [feature_index], shape(operand))
  centered_operand = subtract(operand, mean_bcast)
  return compute_mean(mul(centered_operand, centered_operand), feature_index)

def batch_norm_training(operand, scale, offset, epsilon, feature_index):
  mean = compute_mean(operand, feature_index)
  variance = compute_variance(operand, feature_index)
  return batch_norm_inference(operand, scale, offset, mean, variance, epsilon,
                              feature_index)
```

Feedback:
  1. For `reduce`, follow the order of inputs in the ODS.
  2. Respect varargs for `inputs`, `init_values` and `results` of `reduce`.
  3. Change `x + y` to `add(x, y)`.
  4. Inline `centered_operand_squared` to avoid introducing a new name.
  5. Remove comments in `batch_norm_training` - they don't seem to provide additional context, unlike with `batch_norm_inference`.
  6. Add broadcasts where needed.
  7. Reformat using `pyformat`.

burmako (changes requested):

docs/spec_draft.md
subhankarshah: 1. I had seed the order, but I wanted to follow the order in which they appear in MLIR syntax, which is why I used named arguments (will change)
2. I had noticed that as well but I thought single values are treated as array of 1 element
3. I had thought of that as well but I thought people would confuse it with stablehlo add, come to think about it that we want stablehlo add
4. sure
5. sure
6. I thought division by scalar is automatically broadcasted
7. sure

Thanks for the feedback!

burmako: 1a. I think that named arguments are a nice touch because they help understand what each argument means - that's less obvious without looking into the `reduce` spec than e.g. with `add` or `mul`.
1b/2. Let's use the ordering and the types from the spec - I'd say that MLIR implementation details are just implementation details as far as the spec goes.
6. Generally, StableHLO doesn't allow implicit broadcasting. The only two instances that support them are described in #442, and I think I'd like to get rid of them in the future.

docs/spec_draft.md
```diff
@@ -348,6 +349,93 @@ logical operation.
...
+
+Computes mean and variance across all dimensions except for the `feature_index`
+dimension and normalizes the `operand` tensor producing `output`, `batch_mean`
+and `batch_var` tensors. More formally, this operation can be expressed as a
+decomposition to existing StableHLO operations using Python-like syntax as
+follows:
+
+```python
+def compute_mean(operand, feature_index):
+  (sum,) = reduce(
```
subhankarshah, line 364: I had first written sum then changed it to aggregate because it is reserved in python, also syntax highlighting makes it blue

docs/spec_draft.md
```diff
@@ -348,6 +349,93 @@ logical operation.
...
+      inputs=[operand],
+      init_values=[0],
+      dimensions=[i for i in range(rank(operand)) if i != feature_index],
+      body=lambda x, y: add(x, y))
+  divisor = constant(num_elements(operand) / dim(operand, feature_index))
+  divisor_bcast = broadcast_in_dim(divisor, [], shape(sum))
+  return divide(sum, divisor_bcast)
+
+def compute_variance(operand, mean, feature_index):
```
subhankarshah, line 373: I have taken mean as a parameter in compute_variance, it is ok for this case, but if we want to convey that compute_variance is pure and does not depend on externally computed mean then we can remove it.

docs/spec_draft.md
```diff
@@ -348,6 +349,93 @@ logical operation.
...
+  mean_bcast = broadcast_in_dim(mean, [feature_index], shape(operand))
+  centered_operand = subtract(operand, mean_bcast)
```
subhankarshah, line 375: would it be better if we wrote `stablehlo.broadcast_in_dim` and `stablehlo.subtract` etc
this was a suggestion from @sdasgup3

docs/spec_draft.md
subhankarshah: thanks

docs/spec_draft.md
burmako: Let's keep the old version of variance to emphasize that it can be computed from just `operand` and `feature_index`.

docs/spec_draft.md
burmako: I think that in the context of a pseudocode in the StableHLO spec it's fine to not qualify these functions with `stablehlo.`. Unless there are naming conflicts with meta-functions like `shape` or `num_elements`, which we don't have at the moment.

docs/spec_draft.md
burmako: Let's keep `sum` for the sake of clarity. Calling variables `sum` is not Pythonic, but we only promised "Python-like syntax" above. Also, when Markdown is rendered, the highlighting is gone, which is nice.

burmako (changes requested):

burmako (approved):

docs/spec_draft.md
```diff
@@ -348,6 +349,94 @@ logical operation.
...
+### Examples
+
+```mlir
+// %operand: [
+//            [[1.0, 2.0], [3.0, 4.0]],
+//            [[3.0, 4.0], [1.0, 2.0]]
+//           ]
+// %scale: [1.0, 1.0]
+// %offset: [1.0, 1.0]
+%results:3 = "stablehlo.batch_norm_training"(%operand, %scale, %offset) {
```
burmako, line 426: Nice catch from Sandeep in #351! This should be called `%output:3`, same in the example in the .td file.

burmako (changes requested):

docs/spec_draft.md
subhankarshah: Why though? The individual tensor results are named and not a single result named $output / $outputs
```
let results = (outs
      RankedTensorOf<[HLO_Float]>:$output,
      1DTensorOf<[HLO_Float]>:$batch_mean,
      1DTensorOf<[HLO_Float]>:$batch_var);
```

subhankarshah: Can we unpack the results of an operation in MLIR like -
```%output, %batch_mean, %batch_var = "stablehlo.batct_norm_training"(... .......){}```

Comment (subhankarshah) [reply docs/spec_draft.md:426]: Whatever the final decision is, you may LGTM this pr with the comment, I will make the change and merge.

Comment (burmako) [reply docs/spec_draft.md:426]: You're right - that's my oversight. I like the idea of saying `%output, %batch_mean, %batch_var`, then we can refer to the individual results by name in the comment below.

**Review by burmako — APPROVED** [review]

File: docs/spec_draft.md (hunk 12 above)
Comment (sdasgup3) [reply docs/spec_draft.md:426]: LGTM from my side with the fix for the above.

**Review by sdasgup3 — APPROVED** [review]

File: docs/status.md [hunk 13]
```diff
@@ -52,7 +52,7 @@ one of the following tracking labels.
 | atan2                    | no            | yes*         | yes*           | yes             | no          |
 | batch_norm_grad          | no            | yes*         | yes*           | no              | no          |
 | batch_norm_inference     | yes           | revisit      | yes            | no              | no          |
-| batch_norm_training      | no            | yes*         | yes*           | no              | no          |
+| batch_norm_training      | no            | revisit      | yes            | no              | no          |
 | bitcast_convert          | no            | yes*         | infeasible     | yes             | no          |
```
subhankarshah, line 56: marked revisit like in `batch_norm_inference`

docs/spec_draft.md
```diff
@@ -414,6 +415,95 @@ Numeric precision is implementation-defined.
...
+
+```mlir
+// %operand: [
+//            [[1.0, 2.0], [3.0, 4.0]],
+//            [[3.0, 4.0], [1.0, 2.0]]
+//           ]
+// %scale: [1.0, 1.0]
+// %offset: [1.0, 1.0]
+%output, %batch_mean, %batch_var =
+"stablehlo.batch_norm_training"(%operand, %scale, %offset) {
```
subhankarshah, line 493: updated ODS with the same unpacking format

docs/status.md
subhankarshah: @burmako

docs/spec_draft.md
subhankarshah: @burmako

docs/status.md
subhankarshah: #369

subhankarshah:
@burmako few minor changes for review.

docs/spec_draft.md
burmako: Let's keep variable names and the op on the same line? I don't think we split them across multiple lines anywhere in the spec.

docs/status.md
burmako: Nice catch! Indeed, this should be `revisit`.

burmako (changes requested):

burmako (approved):

Issue #400, opened 2022-10-28
Split verifier for ReduceOp and SortOp


zhouxin913:
### Request description

Aligned with design change in (P3) https://github.com/openxla/stablehlo/blob/main/docs/type_inference.md in PR: (TBD)

burmako:
Superseded by https://github.com/openxla/stablehlo/issues/623.

Issue #394, opened 2022-10-26
Verification of ReduceOp is not strict enough.


sdasgup3:
### Request description

```
func.func @variadic_reduce_add(%arg0: tensor<?x?xf32>, %arg1: tensor<?x4xi32>, 
    %arg2: tensor<?x5xi32>, %arg3: tensor<f32>,
    %arg4: tensor<i32>, %arg5: tensor<i32>
    ) -> (tensor<f32>, tensor<i32>, tensor<i32>) {
  %0:3 = mhlo.reduce(%arg0 init: %arg3),
                    (%arg1 init: %arg4),
                    (%arg2 init: %arg5) across dimensions = [0, 1] {someattr}
   : (tensor<?x?xf32>, tensor<?x4xi32>, tensor<?x5xi32>,
      tensor<f32>, tensor<i32>, tensor<i32>
   ) -> (tensor<f32>, tensor<i32>, tensor<i32>)
     reducer(%arg6: tensor<f32>, %arg9: tensor<f32>) 
            (%arg7: tensor<i32>, %arg10: tensor<i32>)
            (%arg8: tensor<i32>, %arg11: tensor<i32>)  {
      %1 = mhlo.add %arg6, %arg9 : tensor<f32>
      %2 = mhlo.add %arg7, %arg10 : tensor<i32>
      %3 = mhlo.add %arg8, %arg11 : tensor<i32>
      mhlo.return %1, %2, %3 : tensor<f32>, tensor<i32>, tensor<i32>
    }
    return %0#0, %0#1, %0#2 : tensor<f32>, tensor<i32>, tensor<i32>
  }
```

https://source.corp.google.com/piper///depot/google3/third_party/stablehlo/stablehlo/dialect/TypeInference.cpp;rcl=482658831;l=914

input shapes `tensor<?x4xi32>`, `tensor<?x5xi32>` are incompatible, but verifier allows it, because it checks the compatibility with the first argument that has type `tensor<?x?xf32>`.

I think it should infer the "most static" shape first and then try to check whether the arguments are compatible with it. It would be nice to enforce arguments to have the same shape by inserting casts to this inferred type.

sdasgup3:
>> It would be nice to enforce arguments to have the same shape by inserting casts to this inferred type.
While re-thinking on that:
StableHLO framework, by itself, will not be adding any extra casts to enforce those constraints. However, "inserting casts" seems like a nice alternative to adding verification checks.

ghpvnist:
[...]

PR #401, opened 2022-10-28
Split verifiers and shape functions for ops with regions

zhouxin913:
This is a pure refactor PR which fixes https://github.com/openxla/stablehlo/issues/400 (deprecated) and https://github.com/openxla/stablehlo/issues/623 (new)

As in the https://github.com/openxla/stablehlo/issues/623 , "we only need this for ops with regions", a full list of ops with region extract from https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/StablehloOps.td includes **11 ops** in total:

| Op | what's done |
| --- | --- |
| AllReduceOp | No change (already split) |
| CaseOp |  No change (region is indispensable for type inference)  |
| IfOp |  No change (region is indispensable for type inference)  |
| MapOp |  No change (region is indispensable for type inference)  |
| ReduceOp |  Split |
| ReduceScatterOp |  No change (Type Inference implementation on hold see https://github.com/openxla/stablehlo/issues/725) |
| ReduceWindowOp |  Split |
| ScatterOp | No change (already split) |
| SelectAndScatterOp |  No change (already split) |
| SortOp |  Split |
| WhileOp |  Split |

The ideal split is that verifiers contain as almost all verifications, and the shape functions are simple as possible, but note: 
1. `IfOp/CaseOp/MapOp`: We need info from region(s) to infer the return type, so an init function without regions is always invalid and should not exist. No change for them in this PR.
2. As both verifier & shape function need verification of the inputs/attrs, so we need put them in a separate utils functions.
`ReduceOp`: introduce new util `verifyReduceOpInputsAndInferShape()`
`ReduceWindow`: introduce new util `verifyReduceWindowOpInputsAndInferWindow()`
In each op, verifier does (1) call this new util function (2) verify region
shape function:  (1) call this new util function (2) generate inferred type from the intermediate result from (1)
3. Besides, the verification logic needs further fix see https://github.com/openxla/stablehlo/issues/394, but this is out of scope of this PR.

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -1070,8 +1070,7 @@ LogicalResult inferOutfeedOp(MLIRContext* context, Optional<Location> location,
 }
 
 LogicalResult inferReduceOp(
-    Optional<Location>, ValueRange inputs,
-    DenseIntElementsAttr dimensions,
+    Optional<Location>, ValueRange inputs, DenseIntElementsAttr dimensions,
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
   for (auto input : inputs) {
     Type elementTy = getElementTypeOrSelf(input.getType());
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -1069,16 +1069,38 @@ LogicalResult inferOutfeedOp(MLIRContext* context, Optional<Location> location,
   return success();
 }
 
+LogicalResult inferReduceOp(
+    Optional<Location> location, ValueRange inputs, ValueRange initValues,
+    DenseIntElementsAttr dimensions, Region& body,
+    SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
+  for (auto input : inputs) {
+    Type elementTy = getElementTypeOrSelf(input.getType());
+    auto rankedTy = input.getType().dyn_cast<RankedTensorType>();
+    if (!rankedTy) {
+      inferredReturnShapes.emplace_back(elementTy);
+      continue;
+    }
+
+    int64_t rank = rankedTy.getRank();
+    llvm::SmallVector<bool, 4> dimsMask(rank, false);
+    for (int64_t dim : dimensions.getValues<int64_t>()) dimsMask[dim] = true;
+    SmallVector<int64_t, 4> shape;
+    for (int64_t i = 0; i < rank; ++i)
+      if (!dimsMask[i]) shape.push_back(rankedTy.getDimSize(i));
+    inferredReturnShapes.emplace_back(shape, elementTy);
+  }
+  return success();
+}
+
 // We intend to verify the following properties
 //  P1. Verify all `inputs` need to have compatible shapes.
 //  P2. Verify that
 //      1. the dimensions of reduce-op are in-bounds for the given shape.
 //      2. the dimension-attribute have no duplicate entries.
 //  P3. Verify the inner block defining the reducer function.
-LogicalResult inferReduceOp(
-    Optional<Location> location, ValueRange inputs, ValueRange initValues,
-    DenseIntElementsAttr dimensions, Region& body,
-    SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
+LogicalResult verifyReduceOp(Optional<Location> location, ValueRange inputs,
+                             ValueRange initValues,
+                             DenseIntElementsAttr dimensions, Region& body) {
   SmallVector<TensorType> inputArgTypes{llvm::map_range(
       inputs.getTypes(),
       [](Type t) -> TensorType { return t.cast<TensorType>(); })};
@@ -1144,16 +1166,6 @@ LogicalResult inferReduceOp(
   if (failed(verifyReducerShape(location, block, inputArgTypes, initValueTypes,
                                 numInputs, newDimensions, allInputsUnranked)))
     return failure();
-
-  for (uint64_t inputIdx = 0; inputIdx < numInputs; ++inputIdx) {
-    TensorType inputType = inputArgTypes[inputIdx];
-    Type elementType = inputType.getElementType();
-    if (inputType.hasRank())
-      inferredReturnShapes.emplace_back(newDimensions, elementType);
-    else
-      inferredReturnShapes.emplace_back(elementType);
-  }
-
   return success();
 }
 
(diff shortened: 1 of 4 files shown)


GleasonK:
Is the purpose of unused arguments just to make the integration easier? There's a good chance that this change breaks the integ because clang-tidy will (should?) fail on unused arguments. If we did want to go the unused route, would suggest the [`[[maybe_unused]]`](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused) attribute.

Personal preference is to remove unused variables. We own both clients of the shared infrastructure so I don't see too big of a reason to prefer stability.

stablehlo/dialect/TypeInference.h
```diff
@@ -163,6 +167,9 @@ LogicalResult inferSortOp(
     Region& comparator,
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes);
 
+LogicalResult verifySortOp(Optional<Location> location, ValueRange inputs,
+                           uint64_t dimension, Region& comparator);
+
```
burmako, line 172: How about ordering these functions alphabetically, i.e. first all `inferFooOp` functions and then all `verifyFooOp` functions?

stablehlo/dialect/TypeInference.h
```diff
@@ -163,6 +167,9 @@ LogicalResult inferSortOp(
...
 LogicalResult inferTransposeOp(Optional<Location> loc, Value operand,
                                DenseIntElementsAttr permutation,
                                SmallVectorImpl<Type>& inferredReturnTypes);
```
burmako, line 175: Also, as discussed earlier today, can you update the PR to include all ops with regions which currently have type inference implemented for them?

burmako (changes requested):

stablehlo/dialect/TypeInference.h
zhouxin913: Thinking about same thing, resolved in next commit.

stablehlo/dialect/TypeInference.h
zhouxin913: Sure, I also put the full list of ops with region in https://github.com/openxla/stablehlo/issues/623#issuecomment-1341682168

zhouxin913:
> Is the purpose of unused arguments just to make the integration easier? There's a good chance that this change breaks the integ because clang-tidy will (should?) fail on unused arguments. If we did want to go the unused route, would suggest the [`[[maybe_unused]]`](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused) attribute.
> 
> Personal preference is to remove unused variables. We own both clients of the shared infrastructure so I don't see too big of a reason to prefer stability.

Thanks for the reminder! @GleasonK  Yes I decide to go without necessary args. If in some case the an arg is not used, I will go with arg  type but without arg name: to avoid clang-tidy error in both StableHLO/MHLO.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1069,186 +1069,62 @@ LogicalResult inferOutfeedOp(MLIRContext* context, Optional<Location> location,
...
-  for (uint64_t inputIdx = 0; inputIdx < numInputs; ++inputIdx) {
-    TensorType inputType = inputArgTypes[inputIdx];
-    Type elementType = inputType.getElementType();
-    if (inputType.hasRank())
-      inferredReturnShapes.emplace_back(newDimensions, elementType);
-    else
-      inferredReturnShapes.emplace_back(elementType);
+    int64_t rank = rankedTy.getRank();
+    llvm::SmallVector<bool, 4> dimsMask(rank, false);
+    for (int64_t dim : dimensions.getValues<int64_t>()) dimsMask[dim] = true;
```
smit-hinsu, line 1085: Do we verify dimensions attribute before reaching here to make sure that this doesn't lead to crash because of invalid arguments? (I haven't read this change thoroughly)

zhouxin913: Great question.
The current commit was (for `Reduce` and `ReduceWindow`): 1. put all verifications in verifier. 2. infer shape assuming everything is valid in this shape function.
After a second thought now, I think this is probably not OK when the shape functions is called alone (for example in the `create<ReduceOp>` case which this PR is trying to address. Thus, I have no choice but to do the necessary validation again here: put them in an util function and called both by verifier and shape function). 
Will update in a next commit, then make this comment as resolved.

zhouxin913: Updated.
ReduceOp: introduce new util `verifyReduceOpInputsAndInferShape()`
ReduceWindow: introduce new util `verifyReduceWindowOpInputsAndInferWindow()`
In each op, both the verifier & shape functions calls this util function.

burmako (approved):
Excellent PR description! I agree with both the analysis and the implementation.

What are our next steps for CaseOp, IfOp and MapOp? You've made a convincing case that regions are indispensable for type inference of these ops, but in that case what do we do with the autogenerated builders? These questions don't have to be answered in this PR, but we should have a ticket about this, and I think that this ticket should have high priority.

zhouxin913:
> Excellent PR description! I agree with both the analysis and the implementation.
> 
> What are our next steps for CaseOp, IfOp and MapOp? You've made a convincing case that regions are indispensable for type inference of these ops, but in that case what do we do with the autogenerated builders? These questions don't have to be answered in this PR, but we should have a ticket about this, and I think that this ticket should have high priority.

Thanks for the quick review!
Opened https://github.com/openxla/stablehlo/issues/738 and let us think and track in this new bug: whether/how to fix this.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1551,9 +1506,155 @@ LogicalResult inferTriangularSolveOp(
...
+  SmallVector<TensorType> inputArgTypes{llvm::map_range(
+      inputs.getTypes(),
+      [](Type t) -> TensorType { return t.cast<TensorType>(); })};
+  SmallVector<TensorType> initValueTypes{llvm::map_range(
+      initValues.getTypes(),
+      [](Type t) -> TensorType { return t.cast<TensorType>(); })};
+
+  // P1. & P2.
+  SmallVector<int64_t> newDimensions;
+  if (failed(verifyReduceOpInputsAndInferShape(
```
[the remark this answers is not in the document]
zhouxin913: Let me defense with these:
1. As far as digging to code, `verify()` runs before `inferReturnTypeComponents()`, I have shared you with a doc of my founds.
2. This PR aims to address another tricky case that:  the shape function is called **alone** without the verifier: thus we have no choice but to duplicate these verification logic. See https://github.com/openxla/stablehlo/issues/623 

Thanks for pointing out the final part in the comment! Could you open a bug for it and elaborate a bit? I feel it's an optimization, and could do it when we have more cycles. Please correct me otherwise.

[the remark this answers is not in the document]
smit-hinsu: > 1. As far as digging to code, `verify()` runs before `inferReturnTypeComponents()`, I have shared you with a doc of my founds.
you are right. I didn't realize that InferTypeOpInterface sets verifyWithRegions to `1`. 

That means that we need to move all the verification to verifyRegions by using `hasRegionVerifier`. By the time it reaches verifyRegions, inference has been verified so we don't have to do those checks again.

PR #404, opened 2022-10-29
Add type inference for stablehlo.scatter

zhouxin913:
This RP does not touch verify() or verifications: https://github.com/openxla/stablehlo/pull/402 decides that we handle shape function as simple as possible without verifications, thus the verify() will be handled separately.

`inferReturnTypes()` interface is used as ScatterOp is in the list of bounded dynamism plan and this interface is the only choose we have: see detail in description of https://github.com/openxla/stablehlo/pull/375


zhouxin913:
Hi, this PR was old and now fully rebased and updated with no major changes. Please review, thanks!

Assigned to reviewer as discussed.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1246,8 +1246,22 @@ LogicalResult inferReduceWindowOp(
...
+  for (auto input : inputs) {
+    Type elementTy = getElementTypeOrSelf(input.getType());
+    auto rankedTy = input.getType().dyn_cast<RankedTensorType>();
+    if (!rankedTy) {
+      inferredReturnTypes.emplace_back(UnrankedTensorType::get(elementTy));
+      continue;
+    }
+
+    inferredReturnTypes.emplace_back(input.getType());
+  }
```
burmako, line 1264: Also this loop can be replaced by `llvm::append_range(inferredReturnTypes, inputs.getTypes());`.

docs/status.md
```diff
@@ -131,7 +131,7 @@ one of the following tracking labels.
 | round_nearest_afz        | yes           | yes          | yes            | yes             | no          |
 | round_nearest_even       | yes           | yes          | yes            | yes             | no          |
 | rsqrt                    | yes           | yes          | yes            | yes             | no          |
-| scatter                  | yes           | revisit      | no             | no              | no          |
+| scatter                  | yes           | revisit      | yes*           | no              | no          |
```
burmako, line 134: "Type Inference" should be set to "yes" because we now have a spec for `scatter`, and this implementation satisfies it.

burmako (changes requested):

stablehlo/dialect/TypeInference.cpp
zhouxin913: Nice! Fixed.

burmako (approved):

PR #405, opened 2022-10-29
Add type inference for stablehlo.selectandscatter

zhouxin913:
Like https://github.com/openxla/stablehlo/pull/404 This RP does not touch verify() or verifications: https://github.com/openxla/stablehlo/pull/402 decides that we handle shape function as simple as possible without verifications, thus the verify() will be handled separately.

inferReturnTypes() interface is used as SelectAndScatterOp is in the list of bounded dynamism plan and this interface is the only choose we have: see detail in description of https://github.com/openxla/stablehlo/pull/375


zhouxin913:
Rebased and update with no major changes.

Assigned to reviewer.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1281,6 +1281,17 @@ LogicalResult inferSelectOp(
...
 }
 
+LogicalResult inferSelectAndScatterOp(
+    Value operend, SmallVectorImpl<Type>& inferredReturnTypes) {
+  Type elementTy = getElementTypeOrSelf(operend.getType());
+  auto rankedTy = operend.getType().dyn_cast<RankedTensorType>();
+  if (rankedTy)
+    inferredReturnTypes.emplace_back(operend.getType());
+  else
+    inferredReturnTypes.emplace_back(UnrankedTensorType::get(elementTy));
```
burmako, line 1291: Same feedback as in #404. This function can be simplified to `inferredReturnTypes.push_back(operand.getType())`.

stablehlo/dialect/TypeInference.h
```diff
@@ -187,6 +187,9 @@ LogicalResult inferSelectOp(
     Optional<Location> location, Value pred, Value onTrue, Value onFalse,
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes);
 
+LogicalResult inferSelectAndScatterOp(
+    Value operend, SmallVectorImpl<Type>& inferredReturnTypes);
```
burmako, line 191: Typo: "operand".

docs/status.md
```diff
@@ -133,7 +133,7 @@ one of the following tracking labels.
 | rsqrt                    | yes           | yes          | yes            | yes             | no          |
 | scatter                  | yes           | revisit      | no             | no              | no          |
 | select                   | yes           | yes          | yes            | yes             | no          |
-| select_and_scatter       | no            | revisit      | no             | no              | no          |
+| select_and_scatter       | no            | revisit      | yes*           | no              | no          |
```
burmako, line 136: "Type Inference" should be set to "yes" because we now have a spec for `select_and_scatter`, and this implementation satisfies it.

burmako (changes requested):

stablehlo/dialect/TypeInference.h
zhouxin913: Fixed.

burmako (approved):

PR #410, opened 2022-10-31
Add spec for TriangularSolveOp

### LINKED ISSUE #358: Add spec for TriangularSolveOp

ghpvnist:
Verification is set to revisit due to #534
closes #358

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -2701,7 +2701,7 @@ not accessed.
 `transpose_a` can be one of the following:
   * `TRANSPOSE_INVALID`: Illegal matrix; transpose is invalid.
   * `NO_TRANSPOSE`: Perform operation as-is.
-  * `TRANSPOSE`: Perform operation on transpose of `a`
+  * `TRANSPOSE`: Perform operation on transpose of `a`.
   * `ADJOINT`: Perform operation on conjugate transpose of `a`.
 
 ### Inputs
@@ -2723,11 +2723,11 @@ not accessed.
 
 ### Constraints
 
-  * (C1) `a` and `b` have the same type.
-  * (C2) `a` and `b` have the same rank and rank $\ge$ 2.
-  * (C3) dim(`a`, -2) = dim(`a`, -1) = dim(`b`, -2) = dim(`b`, -1).
+  * (C1) `a` and `b` have the same type and rank $\ge$ 2.
+  * (C2) dim(`a`, -2) = dim(`a`, -1).
+  * (C3) `a` and `b` have same shape except `R-1` and `R-2` dimensions.
   * (C4) Suppose `a` has shape `[..., M, M]`, then `b` has shape `[..., M, K]`
-  if `left_side` is true and `[..., K, M]` otherwise.
+    if `left_side` is true and `[..., K, M]` otherwise.
 
 ### Examples
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -2684,26 +2684,24 @@ where `i[d] = j[permutation[d]]`.
 
 ### Semantics
 
-Solves systems of linear equations of a batch of matrices with lower or upper
-triangular coefficients by forward- or back-substitution.
+Solves batches of systems of linear equations with lower or upper triangular
+coefficient matrices.
 
 More formally, given `a` and `b`, `result[i0, ..., iR-3, :, :]` is the solution
 to `op(a[i0, ..., iR-3, :, :]) * x = b[i0, ..., iR-3, :, :]` when `left_side` is
 `true` or `x * op(a[i0, ..., iR-3, :, :]) = b[i0, ..., iR-3, :, :]` when
-`left_side` is `false`, solving for the variable `x` where `op(a)` is either
-`op(a) = a`, or `op(a) = Transpose(a)`, or `op(a) = Conj(Transpose(a))`.
+`left_side` is `false`, solving for the variable `x` where `op(a)` is determined
+by `transpose_a`, which can be one of the following:
+  * `NO_TRANSPOSE`: Perform operation using `a` as-is.
+  * `TRANSPOSE`: Perform operation on transpose of `a`.
+  * `ADJOINT`: Perform operation on conjugate transpose of `a`.
 
 Input data is read only from the lower triangle of `a`, if `lower` is `true` or
 upper triangle of `a`, otherwise. Output data is returned in the same triangle;
 the values in the other triangle are implementation-defined.
 
-If `unit_diagonal` is true, the diagonal elements of `a` are assumed to be 1 and
-not accessed.
-
-`transpose_a` can be one of the following:
-  * `NO_TRANSPOSE`: Perform operation using `a` as-is.
-  * `TRANSPOSE`: Perform operation on transpose of `a`.
-  * `ADJOINT`: Perform operation on conjugate transpose of `a`.
+If `unit_diagonal` is true, then the implementation can assume that the diagonal
+elements of `a` are equal to 1, otherwise the behavior is undefined.
 
 ### Inputs
 
@@ -2724,25 +2722,27 @@ not accessed.
 
 ### Constraints
 
-  * (C1) `a` and `b` have the same element type and rank $\ge$ 2.
-  * (C2) dim(`a`, -2) = dim(`a`, -1).
-  * (C3) dim(`a`, `i`) $=$ dim(`b`, `i`) for all `i` $\in$ [0, R-3]. If
-    `left_size = true`, dim(`b`, -2) $=$ dim(`a`, -1) and dim(`b`, -1) $=$
-    dim(`a`, -1) otherwise.
-  * (C4) `b` and `result` have the same type.
+  * (C1) `a` and `b` have the same element type
+  * (C2) rank(`a`) $=$ rank(`b`) $\ge$ 2.
+  * (C3) dim(`a`, -2) = dim(`a`, -1).
+  * (C4) The relationship between `shape(a)` and `shape(b)` is as follows:
+    * For all `i` $\in$ [0, R-3], dim(`a`, `i`) $=$ dim(`b`, `i`).
+    * `dim(a, R-2)` = `dim(b, left_side ? R-2 : R-1)`.
+    * `dim(a, R-1)` is the same as discussed in C2.
+  * (C5) `b` and `result` have the same type.
 
 ### Examples
 
 ```mlir
 // %a = [
-//       [1, 0, 0],
-//       [2, 4, 0],
-//       [3, 5, 6]
+//       [1.0, 0.0, 0.0],
+//       [2.0, 4.0, 0.0],
+//       [3.0, 5.0, 6.0]
 //      ]
 // %b = [
-//       [2, 0, 0],
-//       [4, 8, 0],
-//       [6, 10, 12]
+//       [2.0, 0.0, 0.0],
+//       [4.0, 8.0, 0.0],
+//       [6.0, 10.0, 12.0]
 //      ]
 %result = "stablehlo.triangular_solve"(%a, %b) {
   left_side = true,
@@ -2751,9 +2751,9 @@ not accessed.
   transpose_a = #stablehlo<transpose NO_TRANSPOSE>
 } : (tensor<3x3xf32>, tensor<3x3xf32>) -> tensor<3x3xf32>
 // %result: [
-//           [2, 0, 0],
-//           [0, 2, 0],
-//           [0, 0, 2]
+//           [2.0, 0.0, 0.0],
+//           [0.0, 2.0, 0.0],
+//           [0.0, 0.0, 2.0]
 //          ]
 ```
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -2695,12 +2695,12 @@ solving for the variable `x` given `a` and `b`, where `op(a)` is either
 output values in the opposite triangle, i.e. the strict upper triangle or strict
 lower triangle correspondingly, are implementation-defined.
 
-If `unit_diagonal` is true, the diamonal elements of `a` are assumed to be 1 and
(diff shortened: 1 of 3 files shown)


docs/spec_draft.md
```diff
@@ -1810,6 +1811,83 @@ where `i[d] = j[permutation[d]]`.
...
+| Name     | Type                                     |
+|----------|------------------------------------------|
+| `result` | tensor of floating-point or complex type |
+
+### Constraints
+
+  * (C1) `a` and `b` have the same type and rank $\ge$ 2.
+  * (C2) dim(`a`, -2) = dim(`a`, -1).
+  * (C3) `a` and `b` have same shape except `R-1` and `R-2` dimensions.
+  * (C4) Suppose `a` has shape `[..., M, M]`, then `b` has shape `[..., M, K]`
```
sdasgup3, line 1860: * dim(`a`, i) $=$ dim(`b`, i) if i $\in$ [0, R-3] and dim(`a`, i) $=$ dim(`b`, R - `left_side` ? 2 : 1 ) otheriwse.

docs/spec_draft.md
```diff
@@ -1810,6 +1811,83 @@ where `i[d] = j[permutation[d]]`.
...
+solving for the variable `x` given `a` and `b`, where `op(a)` is either
+`op(a) = a`, or `op(a) = Transpose(a)`, or `op(a) = Conj(Transpose(a))`. The
+output values in the opposite triangle, i.e. the strict upper triangle or strict
+lower triangle correspondingly, are implementation-defined.
+
+If `unit_diagonal` is true, the diamonal elements of `a` are assumed to be 1 and
+not accessed.
+
+`transpose_a` can be one of the following:
+  * `TRANSPOSE_INVALID`: Illegal matrix; transpose is invalid.
```
sdasgup3, line 1833: This option lead to verification error both in MHLO and HLO?
https://github.com/openxla/stablehlo/blob/ca7ef4f80152e9d5314e4c9750590d018553e91c/stablehlo/dialect/TypeInference.cpp#L1354

We should add that in Constraints.
Next, if that is not a valid value can we know why we have this option in the first place.

docs/spec_draft.md
```diff
@@ -1810,6 +1811,83 @@ where `i[d] = j[permutation[d]]`.
...
+| `unit_diagonal` | constant of type `i1`                                                   |
+| `transpose_a`   | enum of `TRANSPOSE_INVALID`, `NO_TRANSPOSE`, `TRANSPOSE`, and `ADJOINT` |
+
+### Outputs
+
+| Name     | Type                                     |
+|----------|------------------------------------------|
+| `result` | tensor of floating-point or complex type |
+
+### Constraints
```
sdasgup3, line 1855: The type inference of result type is missing:
* `result` and `b` have the same type.

docs/spec_draft.md
```diff
@@ -1810,6 +1811,83 @@ where `i[d] = j[permutation[d]]`.
...
+
+More formally, for all `i`, `result[i0, ..., iR-3, :, :]` is the solution to
+`op(a[i0, ..., iR-3, :, :]) * x = b[i0, ..., iR-3, :, :]` (`left_side` is true)
+`x * op(a[i0, ..., iR-3, :, :]) = b[i0, ..., iR-3, :, :]` (`left_side` is false)
+solving for the variable `x` given `a` and `b`, where `op(a)` is either
+`op(a) = a`, or `op(a) = Transpose(a)`, or `op(a) = Conj(Transpose(a))`. The
+output values in the opposite triangle, i.e. the strict upper triangle or strict
+lower triangle correspondingly, are implementation-defined.
+
+If `unit_diagonal` is true, the diamonal elements of `a` are assumed to be 1 and
```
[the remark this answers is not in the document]
ghpvnist: Thanks!

docs/spec_draft.md
ghpvnist: Thanks for the catch, I'll merge this constraint with (C1).

docs/spec_draft.md
ghpvnist: I updated the constraints. Please take a look.

docs/spec_draft.md
ghpvnist: The reason we have this check is that proto in HLO has enum value of 0 as default. As for why NO_TRANSPOSE is not the default of 0, I do not know. To sync with HLO, this enum value seems necessary. @burmako Thoughts?
https://github.com/tensorflow/tensorflow/blob/910db383e2d5a649eda3ddf3e474852e3a6fddbb/tensorflow/compiler/xla/xla_data.proto#L670-L675

sdasgup3: Let us drop the TRANSPOSE_INVALID.

docs/spec_draft.md
```diff
@@ -1854,11 +1856,12 @@ not accessed.
 
 ### Constraints
 
-  * (C1) `a` and `b` have the same type and rank $\ge$ 2.
+  * (C1) `a`, `b`, and `result` have the same element type and rank $\ge$ 2.
```
sdasgup3, line 1859: (C1) and (C4) has overlapping information. 
(C4) imples that `b` and `result` have the same element type and rank >=2.

I would prefer
(C1) `a` and  `b` have the same element type and rank $\ge$ 2.
(C4)  `b` and `result` have the same type.

docs/spec_draft.md
```diff
@@ -1810,6 +1811,86 @@ where `i[d] = j[permutation[d]]`.
...
+
+| Name     | Type                                     |
+|----------|------------------------------------------|
+| `result` | tensor of floating-point or complex type |
+
+### Constraints
+
+  * (C1) `a`, `b`, and `result` have the same element type and rank $\ge$ 2.
+  * (C2) dim(`a`, -2) = dim(`a`, -1).
+  * (C3) dim(`a`, `i`) $=$ dim(`b`, `i`) for all `i` $\in$ [0, R-3]. If
```
sdasgup3, line 1861: I preferred the following better but might be less readable.
 dim(`a`, i) $=$ dim(`b`, i) if i $\in$ [0, R-3] and dim(`a`, i) $=$ dim(`b`, R - `left_side` ? 2 : 1 ) otheriwse.

Let us keep your edit.

sdasgup3 (approved):

sdasgup3:
Feel free to asign to Eugene once you are done with the changes.

docs/spec_draft.md
```diff
@@ -2279,6 +2280,85 @@ where `i[d] = j[permutation[d]]`.
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.triangular_solve
+
+### Semantics
+
+Solves systems of linear equations of a batch of matrices with lower or upper
+triangular coefficients by forward- or back-substitution.
```
burmako, line 2288: Proposed rewording:

```
Solves batches of systems of linear equations with lower or upper triangular
coefficient matrices.
```

Feedback:
  1. "by forward- or back-substitution". Is this an essential part of semantics? If not, let's drop this and let implementations decide how they want to solve these systems of linear equations.
  2. I don't think it's "systems of equations of a batch of ...". I think it's "batches of systems of equations of ...".
  3. "coefficient matrix" is [a term of art](https://en.wikipedia.org/wiki/Coefficient_matrix), so let's keep it unchanged.

docs/spec_draft.md
```diff
@@ -2279,6 +2280,85 @@ where `i[d] = j[permutation[d]]`.
...
+`true` or `x * op(a[i0, ..., iR-3, :, :]) = b[i0, ..., iR-3, :, :]` when
+`left_side` is `false`, solving for the variable `x` where `op(a)` is either
+`op(a) = a`, or `op(a) = Transpose(a)`, or `op(a) = Conj(Transpose(a))`.
+
+Input data is read only from the lower triangle of `a`, if `lower` is `true` or
+upper triangle of `a`, otherwise. Output data is returned in the same triangle;
+the values in the other triangle are implementation-defined.
+
+If `unit_diagonal` is true, the diagonal elements of `a` are assumed to be 1 and
+not accessed.
```
burmako, line 2301: Proposed reformulation:

```
If `unit_diagonal` is true, then the implementation can assume that the diagonal
elements of `a` are equal to 1, otherwise the behavior is undefined.
```

Feedback:
  1) "the implementation can assume" is consistent with other similar ops.
  2) "and not accessed" is dropped to let the implementations decide on their 
     own what they want to do with this knowledge.
  3) "otherwise the behavior is undefined" needs to be said explicitly.

docs/spec_draft.md
```diff
@@ -2279,6 +2280,85 @@ where `i[d] = j[permutation[d]]`.
...
+### Semantics
+
+Solves systems of linear equations of a batch of matrices with lower or upper
+triangular coefficients by forward- or back-substitution.
+
+More formally, given `a` and `b`, `result[i0, ..., iR-3, :, :]` is the solution
+to `op(a[i0, ..., iR-3, :, :]) * x = b[i0, ..., iR-3, :, :]` when `left_side` is
+`true` or `x * op(a[i0, ..., iR-3, :, :]) = b[i0, ..., iR-3, :, :]` when
+`left_side` is `false`, solving for the variable `x` where `op(a)` is either
+`op(a) = a`, or `op(a) = Transpose(a)`, or `op(a) = Conj(Transpose(a))`.
```
burmako, line 2294: How about "where `op(a)` is determined by `transpose_a` as described below"  to avoid duplication and introducing meta-functions (e.g. `Transpose` which is inconsistent with `transpose`)?

docs/status.md
```diff
@@ -139,7 +139,7 @@ one of the following tracking labels.
 | torch_index_select       | no            | revisit      | no             | no              | no          |
 | trace                    | no            | revisit      | no             | yes             | no          |
 | transpose                | yes           | yes          | yes            | yes             | yes         |
-| triangular_solve         | no            | yes*         | yes*           | no              | no          |
+| triangular_solve         | yes           | yes          | yes            | no              | no          |
```
burmako, line 142: Let's open a ticket to drop `TRANSPOSE_INVALID` and change Verification to "revisit" until the ticket is fixed.

docs/spec_draft.md
```diff
@@ -2279,6 +2280,85 @@ where `i[d] = j[permutation[d]]`.
...
+
+### Outputs
+
+| Name     | Type                                     |
+|----------|------------------------------------------|
+| `result` | tensor of floating-point or complex type |
+
+### Constraints
+
+  * (C1) `a` and `b` have the same element type and rank $\ge$ 2.
```
burmako, line 2327: I don't think it's conventional to merge element type and rank into a single constraint. Let's split them apart like for e.g. `cholesky`:

```
  * (C1) `a` and `b` have the same element type.
  * (C2) `rank(a) = rank(b)` $\ge$ 2.
```

docs/spec_draft.md
burmako: How about the following reformulation:

```
  * (C3) The relationship between `shape(a)` and `shape(b)` is as follows:
    * For all `i` $\in$ [0, R-3], dim(`a`, `i`) $=$ dim(`b`, `i`).
    * `dim(a, R-2)` = `dim(b, left_side ? R-2 : R-1)`.
    * `dim(a, R-1)` is the same as discussed in C2.
```

The idea is to emphasize that C3 exhaustively covers all dimension sizes of `a`. The current formulation makes it hard because all the cases are mixed together in the same paragraph together with an if. Also, `dim(a, ...)` is on the left-hand side of equations in the first sentence, and on the right-hand side in the second sentence.

docs/spec_draft.md
```diff
@@ -2279,6 +2280,85 @@ where `i[d] = j[permutation[d]]`.
...
+// %a = [
+//       [1, 0, 0],
+//       [2, 4, 0],
+//       [3, 5, 6]
+//      ]
+// %b = [
+//       [2, 0, 0],
+//       [4, 8, 0],
+//       [6, 10, 12]
+//      ]
```
burmako, line 2346: How about saying `[1.0, 0.0, 0.0]` etc, so that the commented values parse as floats by MLIR? Same for the result.

burmako (changes requested):

docs/status.md
ghpvnist: Filed an issue here: #534

docs/spec_draft.md
ghpvnist: Makes sense. I moved the description for `transpose_a` for this to flow better.

docs/spec_draft.md
```diff
@@ -2512,6 +2513,85 @@ where `i[d] = j[permutation[d]]`.
...
+| `result` | tensor of floating-point or complex type |
+
+### Constraints
+
+  * (C1) `a` and `b` have the same element type
+  * (C2) rank(`a`) $=$ rank(`b`) $\ge$ 2.
+  * (C3) dim(`a`, -2) = dim(`a`, -1).
+  * (C4) The relationship between `shape(a)` and `shape(b)` is as follows:
+    * For all `i` $\in$ [0, R-3], dim(`a`, `i`) $=$ dim(`b`, `i`).
+    * `dim(a, R-2)` = `dim(b, left_side ? R-2 : R-1)`.
```
subhankarshah, line 2563: `dim(a, R-1)` = `dim(a, R-2)` = `dim(b, left_side ? R-2 : R-1)`.

@ghpvnist how about we remove C2 and the third sub point and combine them like this?

docs/spec_draft.md
subhankarshah: `dim(a, R-1)` = `dim(a, R-2)` = `dim(b, left_side ? R-2 : R-1)`.

How about we remove C2 and the third sub point and combine them like this?

subhankarshah: Keeping the rest of C3 and the first sub-point as is ofcourse

ghpvnist: Sure thing.

burmako (approved):

PR #411, opened 2022-11-01
Allow inlining StableHLO ops

burmako:
This hasn't been discussed much, but as part of defining the
relationship between StableHLO and MHLO, it looks reasonable to avoid
doing transformations on StableHLO and leave that to MHLO.

This way, we don't have to duplicate stuff like constant folding,
canonicalization, etc etc.

Inlining is one of the transformations that's in a gray area because
this is something that we get for free from upstream, so I don't think
that the duplication argument applies here.

### DIFF
--- a/stablehlo/tests/transform_stablehlo.mlir
+++ b/stablehlo/tests/transform_stablehlo.mlir
@@ -1,0 +1,12 @@
+// RUN: stablehlo-opt --inline %s | FileCheck %s
+
+func.func @main(%arg0: tensor<f32>) -> tensor<f32> {
+  %0 = func.call @callee(%arg0): (tensor<f32>) -> tensor<f32>
+  func.return %0 : tensor<f32>
+}
+
+// CHECK-NOT: func.func private @callee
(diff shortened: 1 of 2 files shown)


stablehlo/tests/transform_stablehlo.mlir
```diff
@@ -0,0 +1,12 @@
+// RUN: stablehlo-opt --inline %s | FileCheck %s
+
+func.func @main(%arg0: tensor<f32>) -> tensor<f32> {
+  %0 = func.call @callee(%arg0): (tensor<f32>) -> tensor<f32>
```
GleasonK, line 4: Maybe add a positive aspect to this test case, something to the effect of:

```
CHECK:      @main
CHECK-NEXT: stablehlo.abs
```

GleasonK (approved):
Code looks good. Could you add a note to the PR with the following before merge:

1. A bit more info on the use case / why this is needed.
2. What makes this transformation different than others that we want it in our shallow dialect which doesn't have transformations (currenlty)?

Also I'll take a look at the whitespace issue now, not sure how we keep merging whitespace..

PR #413, opened 2022-11-01
Add spec for RngOp

### LINKED ISSUE #412: Add spec for RngOp

ghpvnist:
closes #412


docs/spec_draft.md
```diff
@@ -1406,6 +1407,76 @@ and produces a `result` tensor. More formally,
...
+
+Constructs an output of a given shape with random numbers generated given
+`rng_distribution`.
+
+`rng_distribution` is one of the following:
+  * `UNIFORM`: the uniform distribution over the interval `[a,b)`. The
+    parameters and output element type have to be a boolean type, an integral
+    type or a floating point types, and the types have to be consistent.
+  * `NORMAL`: the normal distribution with parameters $\mu$ = `a` and $\sigma$ =
+    `b`. The parameters and output shape have to have a floating point
```
sdasgup3, line 1423: The parameter type and shape are already clarified in Inputs/Outputs and Constraints. SPO we can skip this "The parameters ...scalar valued."

docs/spec_draft.md
```diff
@@ -1406,6 +1407,76 @@ and produces a `result` tensor. More formally,
 ```
 
 [Back to Ops](#index-of-ops)
+
+## stablehlo.rng
+
+### Semantics
+
+Constructs an output of a given shape with random numbers generated given
```
sdasgup3, line 1415: Constructs a `result` tensor of a given shape `shape` with random numbers generated using `rng_distribution` algorithm.  If `rng_distribution` $=$ `UNIFORM`, then the random numbers are generated following the uniform distribution over the interval  `[a,b)`. If `rng_distribution` $=$ `NORMAL`, then the random numbers are generated following the  normal distribution with mean = `a` and standard deviation = `b`.

docs/spec_draft.md
```diff
@@ -1406,6 +1407,76 @@ and produces a `result` tensor. More formally,
...
+`rng_distribution`.
+
+`rng_distribution` is one of the following:
```
sdasgup3, line 1418: With the explanation above, we do not need any of the following (including the formalism).

stablehlo/dialect/StablehloOps.td
```diff
@@ -2727,27 +2727,17 @@ def StableHLO_OptimizationBarrierOp : StableHLO_Op<"optimization_barrier",
...
 def StableHLO_RngOp : StableHLO_Op<"rng", [InferTensorTypeWithReify, AllElementTypesMatch<["a", "b", "result"]>]> {
-  let summary = "RNG with uniform distribution.";
+  let summary = "RNG operator with uniform or normal distribution.";
   let description = [{
-    Constructs an output of a given shape with random numbers generated
-    following the given `rng_distribution` with two parameters:
-      `UNIFORM`: the uniform distribution over the interval `[a,b)`. The parameters
-                 and output element type have to be a boolean type, an integral type or a
-                 floating point types, and the types have to be consistent.
+    Constructs an output of a given shape with random numbers generated given
```
sdasgup3, line 2732: update the description as per the spec.

docs/spec_draft.md
```diff
@@ -1406,6 +1407,76 @@ and produces a `result` tensor. More formally,
...
+  * (C3) If `rng_distribution = NORMAL`, `a` and `b` have floating-point type.
+
+### Examples
+
+```mlir
+// %a = 0
+// %b = 2
+// %shape = [3, 3]
+%result = "stablehlo.rng"(%a, %b, %shape) {
+  rng_distribution = #stablehlo<rng_distribution NORMAL>
```
sdasgup3, line 1469: all the lement types should be float as per the (C3) and the verification check.

stablehlo/dialect/StablehloOps.td
```diff
@@ -2727,27 +2727,17 @@ def StableHLO_OptimizationBarrierOp : StableHLO_Op<"optimization_barrier",
...
-
-                See https://www.tensorflow.org/xla/operation_semantics#rngnormal.
+    See:
+    https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md#stablehlorng
 
     Example:
-
     ```mlir
-    %1 = stablehlo.rng %arg0, %arg1, %0, distribution = NORMAL : (tensor<f32>, tensor<f32>, tensor<3xi64>) -> tensor<2x3x5xf32>
+    %result = stablehlo.rng %a, %b, %shape, distribution = NORMAL : (tensor<i32>, tensor<i32>, tensor<2xi32>) -> tensor<3x3xi32>
```
sdasgup3, line 2740: update the elementtypes.

docs/spec_draft.md
```diff
@@ -1406,6 +1407,61 @@ and produces a `result` tensor. More formally,
...
+implementation-defined.
+
+### Inputs
+
+| Name               | Type                                                 |
+|--------------------|------------------------------------------------------|
+| `a`                | constant of integer, boolean, or floating-point type |
+| `b`                | constant of integer, boolean, or floating-point type |
+| `shape`            | 1-dimensional tensor of integer type                 |
+| `rng_distribution` | constant of type `enum {UNIFORM, NORMAL}`            |
```
sdasgup3, line 1432: ` enum of UNIFORM and NORMAL`?

sdasgup3 (approved):

docs/spec_draft.md
```diff
@@ -1406,6 +1407,61 @@ and produces a `result` tensor. More formally,
...
+| Name     | Type                                               |
+|----------|----------------------------------------------------|
+| `result` | tensor of integer, boolean, or floating-point type |
+
+### Constraints
+
+  * (C1) `a`, `b`, and `result` have the same element type.
+  * (C2) dim(`result`, `i`) = `shape[i]` for all `i` $\in$ [0, size(`shape`)).
+  * (C3) If `rng_distribution = NORMAL`, `a`, `b`, and `result` have the same
+    floating-point element type.
```
burmako, line 1445: How about reordering C2 and C3, so that the related constraints C1 and C3 are located next to each other?

docs/spec_draft.md
```diff
@@ -1406,6 +1407,61 @@ and produces a `result` tensor. More formally,
...
+If `rng_distribution = UNIFORM` and `b` $\le$ `a`, the result is
+implementation-defined.
+
+### Inputs
+
+| Name               | Type                                                 |
+|--------------------|------------------------------------------------------|
+| `a`                | constant of integer, boolean, or floating-point type |
+| `b`                | constant of integer, boolean, or floating-point type |
+| `shape`            | 1-dimensional tensor of integer type                 |
```
burmako, line 1431: Let's call this "1-dimensional tensor constant of type `si64`" for now, because otherwise it is unclear what C2 means, and also update the `shape` in examples to say `i64` instead of `i32`. I'll sort this out later in the dynamism RFC.

docs/spec_draft.md
```diff
@@ -1406,6 +1407,61 @@ and produces a `result` tensor. More formally,
...
+// %b = 2
+// %shape = [3, 3]
+%result = "stablehlo.rng"(%a, %b, %shape) {
+  rng_distribution = #stablehlo<rng_distribution NORMAL>
+} : (tensor<f32>, tensor<f32>, tensor<2xi32>) -> tensor<3x3xf32>
+// %result: [
+//           [1, 0, 1],
+//           [1, 1, 1],
+//           [0, 0, 0]
+//          ]
```
burmako, line 1460: The result doesn't seem to have been updated to use `f32`.

burmako (changes requested):

docs/spec_draft.md
```diff
@@ -1406,6 +1407,61 @@ and produces a `result` tensor. More formally,
...
+## stablehlo.rng
+
+### Semantics
+
+Constructs a `result` tensor of a given shape `shape` with random numbers
+generated using the `rng_distribution` algorithm. If `rng_distribution` $=$
+`UNIFORM`, then the random numbers are generated following the uniform
+distribution over the interval [`a`, `b`). If `rng_distribution` $=$ `NORMAL`,
+then the random numbers are generated following the normal distribution with
+mean = `a` and standard deviation = `b`.
```
burmako, line 1420: Also what happens if `b` is negative in the `NORMAL` case?

docs/spec_draft.md
```diff
@@ -1406,6 +1407,61 @@ and produces a `result` tensor. More formally,
...
+
+If `rng_distribution = UNIFORM` and `b` $\le$ `a`, the result is
+implementation-defined.
```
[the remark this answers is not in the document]
ghpvnist: By implementation defined, I mean that if some random number generator has this implementation:
```
rng(a, b, shape):
  result = [...]
  a, b = if a > b ? b , a :  a, b
  if a == b:
    result[:] = None
  ...
  
```
it will be able to handle the a >= b case. If other algorithms returns failure or handles it differently, that could also happen.

> I.e. is there a legitimate use for UNIFORM in conjunction with a  b?

I don't think so.

docs/spec_draft.md
ghpvnist: Thanks for catching this, I meant to use UNIFORM with `i32`.

docs/spec_draft.md
ghpvnist: Good point. I think it should be implementation defined as well, since according to the cpp [doc](https://en.cppreference.com/w/cpp/numeric/random/normal_distribution), even though the std is negative, the normal distribution still outputs a number. In the case of negative std, it will "flip" the curve horizontally, but since the distribution curve is symmetrical, I don't know if this is of concern. Very interesting.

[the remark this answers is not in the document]
ghpvnist: This is what HLO is doing:

https://github.com/tensorflow/tensorflow/blob/2662c525af2f99a731290657e4b494fee29feb64/tensorflow/compiler/xla/hlo/evaluator/hlo_evaluator.cc#L801-L811

which uses some predetermined config seed (default 0) or a monotonic seed per evaluator using `std::random_device`.

docs/spec_draft.md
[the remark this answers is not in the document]
burmako: "I don't think so". In that case, let's say:

```
If `rng_distribution = UNIFORM` and `a` $\ge$ `b`, the behavior is undefined.
Same for `rng_distribution = NORMAL` and negative `b`.
```

Undefined behavior is the tightest form of restrictions we can put on runtime semantics, so let's do this if we have a choice (which we seem we have in this situation).

docs/spec_draft.md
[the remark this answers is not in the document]
burmako: Are there any guarantees that we can provide the producers about the behavior of this op? E.g. can an implementation return the same values over and over upon successive calls?

burmako (changes requested):
Left further feedback as standalone comments.

docs/spec_draft.md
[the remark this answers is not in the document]
ghpvnist: There are no guarantees we can provide. The interpreter will provide nondeterministic values upon successive calls. I will clarify this in the spec.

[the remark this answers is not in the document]
ghpvnist: Actually, I don't think there is a need for any clarification. Since we don't provide a parameter to initialize the seed, the default should be to provide random values. Otherwise, the op is not so random :)

docs/spec_draft.md
```diff
@@ -1917,6 +1918,62 @@ and produces a `result` tensor. More formally,
...
+`result` tensor of a given shape `shape`. If `rng_distribution` $=$ `UNIFORM`,
+then the random numbers are generated following the uniform distribution over
+the interval [`a`, `b`). If `rng_distribution` $=$ `NORMAL`, then the random
+numbers are generated following the normal distribution with mean = `a` and
+standard deviation = `b`.
+
+If `rng_distribution = UNIFORM` and `a` $\ge$ `b`, the behavior is undefined.
+Similarly, if `rng_distribution = NORMAL` and `b` $\le$ 0, the behavior is
+undefined.
+
```
burmako, line 1936: Proposed rewording:

```
Generates random numbers using the `rng_distribution` algorithm and produces a
`result` tensor of a given shape `shape`. 

If `rng_distribution` $=$ `UNIFORM`, then the random numbers are generated 
following the uniform distribution over the interval [`a`, `b`). If `a` $\ge$ 
`b`, the behavior is undefined.

If `rng_distribution` $=$ `NORMAL`, then the random numbers are generated 
following the normal distribution with mean = `a` and standard deviation = `b`.
If `b` $\lt$ 0, the behavior is undefined.

The exact way how random numbers are generated is implementation-defined.
For example, they may or may not be deterministic, and they may or may not
use hidden state.
```

Feedback:
  1. Upon second look, I'm realizing that it might be better to group all UNIFORM-specific prose and all NORMAL-specific prose together in dedicated paragraphs.
  2. Let's call out the lack of any guarantees for the generated random numbers.

docs/spec_draft.md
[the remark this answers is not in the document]
burmako: I think it would be good to explicitly call out the extent of guarantees that we can provide. Proposed some wording in a nearby comment.

burmako (changes requested):

docs/spec_draft.md
ghpvnist: Looks good to me.

burmako (approved):

PR #415, opened 2022-11-01
Add spec for RngBitGeneratorOp

### LINKED ISSUE #414: Add spec for RngBitGeneratorOp

ghpvnist:
Raised issue #643 to track which types are supported for the inputs/outputs.
Verification is set to `revisit` due to #486.
closes #414

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -3127,17 +3127,11 @@ deterministic between implementations.
 
 ### Constraints
 
-  * (C1) shape(`initial_state`) $=$ shape(`output_state`).
-  * (C2) For `initial_state`, `output_state`, and `output`:
-    * If `rng_algorithm = DEFAULT`:
-      * `initial_state`, `output_state`, and `output` have backend-specific type
-        requirements.
-    * If `rng_algorithm = THREE_FRY`:
-      * type(`initial_state`) $=$ type(`output_state`) $=$ `tensor<2xui64>`.
-      * element_type(`output`) $\in$ {`ui32`, `ui64`}.
-    * If `rng_algorithm = PHILOX`:
-      * type(`initial_state`) $=$ type(`output_state`) $=$ `tensor<3xui64>`.
-      * element_type(`output`) $\in$ {`ui32`, `ui64`}.
+  * (C1) type(`initial_state`) $=$ type(`output_state`).
+  * (C2) size(`initial_state`) depends on `rng_algorithm`:
+    * `DEFAULT`: implementation-defined.
+    * `THREE_FRY`: `2`.
+    * `PHILOX`: `2` or `3`.
 
 ### Examples
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -3097,16 +3097,18 @@ hidden state.
 
 ### Semantics
 
-Returns an `output` filled with uniform random data and an updated
-`output_state` using the `rng_algorithm` given `initial_state`. The output is
-guaranteed to be deterministic from the initial state, but it is not guaranteed
-to be deterministic between backends and different compiler versions.
+Returns an `output` filled with uniform random bits and an updated output state
+`output_state` given an initial state `initial_state` using the pseudorandom
+number generator algorithm `rng_algorithm`. The output is guaranteed to be
+deterministic function of `initial_state`, but it is not guaranteed to be
+deterministic between backends and different compiler versions.
 
 `initial_state` is the initial state of the current random number generation.
 The initial state, required shape, and valid values are dependent on the
 algorithm used.
 
-`rng_algorithm` is one of the following:
+`rng_algorithm` is one of the following, different variants of the algorithm is
+implementation-defined:
   * `DEFAULT`: Backend specific algorithm.
   * `THREE_FRY`: Threefry counter-based PRNG algorithm.*
   * `PHILOX`: Philox algorithm to generate random numbers in parallel.*
@@ -3114,17 +3116,12 @@ algorithm used.
 \* See: [Salmon et al. SC 2011. Parallel random numbers: as easy as 1, 2, 3.
 ](http://www.thesalmons.org/john/random123/papers/random123sc11.pdf)
 
-More formally, given the function `rng_algorithm` which takes an
-`initial_state` as input, the `output` with uniform random data is generated
-given `initial_state`:
-  * `output_state, output = rng(initial_state)`
-
 ### Inputs
 
-| Name            | Type                                                 |
-|-----------------|------------------------------------------------------|
-| `initial_state` | tensor of integer or floating-point type             |
-| `rng_algorithm` | constant of type `enum {DEFAULT, THREE_FRY, PHILOX}` |
+| Name            | Type                                         |
+|-----------------|----------------------------------------------|
+| `initial_state` | tensor of integer or floating-point type     |
+| `rng_algorithm` | enum of `DEFAULT`, `THREE_FRY`, and `PHILOX` |
 
 ### Outputs
 
@@ -3135,22 +3132,31 @@ given `initial_state`:
 
 ### Constraints
 
-  * (C1) Shape of `initial_state` vary depending on the `rng_algorithm`:
-    * If `rng_algorithm = DEFAULT`, `initial_state` has backend-specific
-      requirements.
-    * If `rng_algorithm = THREE_FRY`, `initial_state` is 2-dimensional.
-    * If `rng_algorithm = PHILOX`, `initial_state` is 3-dimensional.
-  * (C2) `output_state` has the same shape as `initial_state`.
+  * (C1) shape(`initial_state`) $=$ shape(`output_state`).
+  * (C2) The relationship between `initial_state`, `output_state`, and `output`
+      varies:
+    * If `rng_algorithm = DEFAULT`:
+      * `initial_state`, `output_state`, and `output` have backend-specific
+        requirements.
+    * If `rng_algorithm = THREE_FRY`:
+      * type(`initial_state`) $=$ type(`output_state`) $=$ `2xui64`.
+      * element_type(`output`) $\in$ {`ui32`, `ui64`}.
+    * If `rng_algorithm = PHILOX`:
+      * type(`initial_state`) $=$ type(`output_state`) $=$ `3xui64`.
+      * element_type(`output`) $\in$ {`ui32`, `ui64`}.
 
 ### Examples
 
 ```mlir
-// %initial_state: [1, 2, 3]
+// %initial_state: [1, 2]
 %output_state, %output = "stablehlo.rng_bit_generator"(%initial_state) {
-  rng_algorithm = #stablehlo.rng_algorithm<DEFAULT>
-} : (tensor<3xi32>) -> (tensor<3xi32>, tensor<2x2xf64>)
-// %output_state: []
-// %output: [[], []]
+  rng_algorithm = #stablehlo.rng_algorithm<THREE_FRY>
+} : (tensor<2xui64>) -> (tensor<2xui64>, tensor<2x2xui64>)
+// %output_state: [1, 6]
+// %output: [
+//           [9236835810183407956, 16087790271692313299],
+//           [18212823393184779219, 2658481902456610144]
+//          ]
 ```
 
 [Back to Ops](#index-of-ops)
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -3103,10 +3103,6 @@ number generator algorithm `rng_algorithm`. The output is guaranteed to be
 deterministic function of `initial_state`, but it is not guaranteed to be
 deterministic between backends and different compiler versions.
 
-`initial_state` is the initial state of the current random number generation.
-The initial state, supported element type, and valid values are dependent on the
-algorithm used.
-
 `rng_algorithm` is one of the following, different variants of the algorithm is
 implementation-defined:
   * `DEFAULT`: Backend specific algorithm.
@@ -3138,10 +3134,10 @@ implementation-defined:
       * `initial_state`, `output_state`, and `output` have backend-specific type
         requirements.
     * If `rng_algorithm = THREE_FRY`:
-      * type(`initial_state`) $=$ type(`output_state`) $=$ `2xui64`.
+      * type(`initial_state`) $=$ type(`output_state`) $=$ `tensor<2xui64>`.
       * element_type(`output`) $\in$ {`ui32`, `ui64`}.
     * If `rng_algorithm = PHILOX`:
-      * type(`initial_state`) $=$ type(`output_state`) $=$ `3xui64`.
+      * type(`initial_state`) $=$ type(`output_state`) $=$ `tensor<3xui64>`.
       * element_type(`output`) $\in$ {`ui32`, `ui64`}.
 
 ### Examples
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -382,6 +382,7 @@ syntax.
    * [reshape](#stablehloreshape)
    * [reverse](#stablehloreverse)
    * [rng](#stablehlorng)
+   * [rng_bit_generator](#stablehlorng_bit_generator)
    * [round_nearest_afz](#stablehloround_nearest_afz)
    * [round_nearest_even](#stablehloround_nearest_even)
    * [rsqrt](#stablehlorsqrt)
@@ -3088,6 +3089,60 @@ hidden state.
 //           [1, 1, 1],
 //           [0, 0, 0]
 //          ]
+```
+
+[Back to Ops](#index-of-ops)
+
+## stablehlo.rng_bit_generator
+
+### Semantics
+
+Returns an `output` filled with uniform random data and an updated
+`output_state` using the `rng_algorithm` given `initial_state`. The output is
+guaranteed to be deterministic from the initial state, but it is not guaranteed
+to be deterministic between backends and different compiler versions.
+
+`initial_state` is the initial state of the current random number generation.
+The initial state, required shape, and valid values are dependent on the
+algorithm used.
+
+`rng_algorithm` is one of the following:
+  * `DEFAULT`: Backend specific algorithm.
+  * `THREE_FRY`: ThreeFry counter-based PRNG algorithm.
+  * `PHILOX`: Philox algorithm to generate random numbers in parallel.
+
+### Inputs
+
+| Name            | Type                                                 |
+|-----------------|------------------------------------------------------|
+| `initial_state` | tensor of integer or floating-point type             |
+| `rng_algorithm` | constant of type `enum {DEFAULT, THREE_FRY, PHILOX}` |
+
+### Outputs
+
+| Name           | Type                                     |
+|----------------|------------------------------------------|
+| `output_state` | tensor of integer or floating-point type |
+| `output`       | tensor of integer or floating-point type |
+
+### Constraints
+
+  * (C1) Shape of `initial_state` vary depending on the `rng_algorithm`:
+    * If `rng_algorithm = DEFAULT`, `initial_state` has backend-specific
+      requirements.
+    * If `rng_algorithm = THREE_FRY`, `initial_state` is 2-dimensional.
+    * If `rng_algorithm = PHILOX`, `initial_state` is 3-dimensional.
+  * (C2) `output_state` has the same shape as `initial_state`.
+
+### Examples
+
+```mlir
+// %initial_state: [1, 2, 3]
+%output_state, %output = "stablehlo.rng_bit_generator"(%initial_state) {
+  rng_algorithm = #stablehlo.rng_algorithm<DEFAULT>
+} : (tensor<3xi32>) -> (tensor<3xi32>, tensor<2x2xf64>)
+// %output_state: []
+// %output: [[], []]
 ```
 
 [Back to Ops](#index-of-ops)
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -3104,7 +3104,7 @@ deterministic function of `initial_state`, but it is not guaranteed to be
 deterministic between implementations.
 
 `rng_algorithm` is one of the following:
-  * `DEFAULT`: Backend specific algorithm.
+  * `DEFAULT`: Implementation-defined algorithm.
   * `THREE_FRY`: Implementation-defined variant of the Threefry algorithm.*
   * `PHILOX`: Implementation-defined variant of the Philox algorithm.*
 
@@ -3115,14 +3115,14 @@ deterministic between implementations.
 
 | Name            | Type                                         |
 |-----------------|----------------------------------------------|
-| `initial_state` | tensor of integer or floating-point type     |
+| `initial_state` | 1-dimensional tensor of type `ui64`          |
 | `rng_algorithm` | enum of `DEFAULT`, `THREE_FRY`, and `PHILOX` |
 
 ### Outputs
 
 | Name           | Type                                     |
 |----------------|------------------------------------------|
-| `output_state` | tensor of integer or floating-point type |
+| `output_state` | 1-dimensional tensor of type `ui64`      |
 | `output`       | tensor of integer or floating-point type |
 
 ### Constraints
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -3101,13 +3101,12 @@ Returns an `output` filled with uniform random bits and an updated output state
 `output_state` given an initial state `initial_state` using the pseudorandom
 number generator algorithm `rng_algorithm`. The output is guaranteed to be
 deterministic function of `initial_state`, but it is not guaranteed to be
-deterministic between backends and different compiler versions.
+deterministic between implementations.
 
-`rng_algorithm` is one of the following, different variants of the algorithm is
-implementation-defined:
+`rng_algorithm` is one of the following:
   * `DEFAULT`: Backend specific algorithm.
-  * `THREE_FRY`: Threefry counter-based PRNG algorithm.*
-  * `PHILOX`: Philox algorithm to generate random numbers in parallel.*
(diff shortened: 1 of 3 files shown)


docs/spec_draft.md
```diff
@@ -1406,6 +1407,69 @@ and produces a `result` tensor. More formally,
...
+
+`rng_algorithm` is one of the following:
+  * `DEFAULT`: Backend specific algorithm.
+  * `THREE_FRY`: Threefry counter-based PRNG algorithm.*
+  * `PHILOX`: Philox algorithm to generate random numbers in parallel.*
+
+\* See: [Salmon et al. SC 2011. Parallel random numbers: as easy as 1, 2, 3.
+](http://www.thesalmons.org/john/random123/papers/random123sc11.pdf)
+
+More formally, given the function `rng_algorithm` which takes an
```
sdasgup3, line 1432: we can skip this formalism, as the function of the op is straight-forward.

docs/spec_draft.md
```diff
@@ -1406,6 +1407,69 @@ and produces a `result` tensor. More formally,
...
+`initial_state` as input, the `output` with uniform random data is generated
+given `initial_state`:
+  * `output_state, output = rng(initial_state)`
+
+### Inputs
+
+| Name            | Type                                                 |
+|-----------------|------------------------------------------------------|
+| `initial_state` | tensor of integer or floating-point type             |
+| `rng_algorithm` | constant of type `enum {DEFAULT, THREE_FRY, PHILOX}` |
```
sdasgup3, line 1442: enum of `DEFAULT` `THREE_FRY`, and `PHILOX`

docs/spec_draft.md
```diff
@@ -1406,6 +1407,69 @@ and produces a `result` tensor. More formally,
...
+
+### Semantics
+
+Returns an `output` filled with uniform random data and an updated
+`output_state` using the `rng_algorithm` given `initial_state`. The output is
+guaranteed to be deterministic from the initial state, but it is not guaranteed
+to be deterministic between backends and different compiler versions.
+
+`initial_state` is the initial state of the current random number generation.
+The initial state, required shape, and valid values are dependent on the
```
sdasgup3, line 1421: I think by `required shape` they mean the shape of the initial state. Because the output shape is provided by the user only.
Having said that we can remove the entire sentence*** because we already talked about the  shape dependence on algorithm in the constants section. 

*** Before removing please also look into the following:

`valid values are dependent on the algorithm`: 
Can we know what do they mean by "valid values". IOW, what qualifies value to be valid Vs  invalid ?

docs/spec_draft.md
```diff
@@ -1406,6 +1407,69 @@ and produces a `result` tensor. More formally,
...
+|----------------|------------------------------------------|
+| `output_state` | tensor of integer or floating-point type |
+| `output`       | tensor of integer or floating-point type |
+
+### Constraints
+
+  * (C1) Shape of `initial_state` vary depending on the `rng_algorithm`:
+    * If `rng_algorithm = DEFAULT`, `initial_state` has backend-specific
+      requirements.
+    * If `rng_algorithm = THREE_FRY`, `initial_state` is 2-dimensional.
```
sdasgup3, line 1456: In the hlo spec we have the element-type also specified u64[2]. But I cannot see the any verification check in MHLO. Can we find the relevance of that element-type?

docs/spec_draft.md
```diff
@@ -1406,6 +1407,69 @@ and produces a `result` tensor. More formally,
...
+
+### Examples
+
+```mlir
+// %initial_state: [1, 2, 3]
+%output_state, %output = "stablehlo.rng_bit_generator"(%initial_state) {
+  rng_algorithm = #stablehlo.rng_algorithm<DEFAULT>
+} : (tensor<3xi32>) -> (tensor<3xi32>, tensor<2x2xf64>)
+// %output_state: []
+// %output: [[], []]
```
sdasgup3, line 1468: Just curious if there is ever a any value generated for this particular output shape and `initial_state` ? If yes, then we should use some values instead.

stablehlo/dialect/StablehloOps.td
```diff
@@ -2770,27 +2770,26 @@ def StableHLO_RngOp : StableHLO_Op<"rng", [InferTensorTypeWithReify, AllElementT
 def StableHLO_RngBitGeneratorOp : StableHLO_Op<"rng_bit_generator", [Pure]> {
   let summary = "Uniform random number generator operator";
   let description = [{
-    Returns an output with a given shape filled with uniform random bits using
-    the specified algorithm (or backend default) and returns an updated state
-    (with the same shape as initial state) and the generated random data.
+    Returns an `output` filled with uniform random data and an updated
```
sdasgup3, line 2773: update the description as p[er the spec.

stablehlo/dialect/StablehloOps.td
```diff
@@ -2770,27 +2770,26 @@ def StableHLO_RngOp : StableHLO_Op<"rng", [InferTensorTypeWithReify, AllElementT
...
     ```mlir
-    %0, %1 = stablehlo.rng_bit_generator %arg0, algorithm = PHILOX : (tensor<3xui64>) -> (tensor<3xui64>, tensor<2x2xui32>)
+    %output_state, %output = stablehlo.rng_bit_generator %initial_state, algorithm = DEFAULT : (tensor<3xi32>) -> (tensor<3xi32>, tensor<2x2xf64>)
     ```
   }];
   let arguments = (ins
-    StableHLO_RngAlgorithmAttr:$rng_algorithm,
-    HLO_IntOrFpTensor:$initial_state
+    HLO_IntOrFpTensor:$initial_state,
+    StableHLO_RngAlgorithmAttr:$rng_algorithm
```
sdasgup3, line 2786: Is there any specific reason to reorder those?

ghpvnist: The file's convention uses operands first and attributes last and pretty print uses this order as well, hence this change.

docs/spec_draft.md
ghpvnist: The element type requirement comes from the two algorithms. Indeed, we don't have verification check, so status for verification should be `revisit`. Thanks for pointing this out.

docs/spec_draft.md
ghpvnist: By valid values, it means the element type since THREE_FRY and PHILOX algorithms requires ui64 `initial_state` and outputs ui64 `output_state` and outputs either u32 or u64 `output` (aka, no other precision or element types are allowed implicitly in the verifier).

There is no explicit check for this since the DEFAULT prng algorithm may support *uniform* (it's hard to produce uniform floating-points since many bit representations can represent NaNs, for example) floating-point prng, which seems to be a very niche [special case](https://electronics.stackexchange.com/questions/586009/how-is-a-floating-point-prng-designed) though.

I think the use case of this op is to generate uniformly random bits, not random numbers.

docs/spec_draft.md
ghpvnist: Thanks, fixed now.

docs/spec_draft.md
sdasgup3: Yes, also have a ticket to track it.

stablehlo/dialect/StablehloOps.td
sdasgup3: Maes sense then.

docs/spec_draft.md
```diff
@@ -1718,6 +1719,75 @@ and produces a `result` tensor. More formally,
...
+
+  * (C1) shape(`initial_state`) $=$ shape(`output_state`).
+  * (C2) The relationship between `initial_state`, `output_state`, and `output`
+      varies:
+    * If `rng_algorithm = DEFAULT`:
+      * `initial_state`, `output_state`, and `output` have backend-specific
+        requirements.
+    * If `rng_algorithm = THREE_FRY`:
+      * type(`initial_state`) $=$ type(`output_state`) $=$ `2xui64`.
+      * element_type(`output`) $\in$ {`ui32`, `ui64`}.
```
sdasgup3, line 1770: I know we have already gone through this, but still have a few questions:
Is the second constraint of the `output` on element-type ever mentioned in the paper. I know that you did exhaustive experiments to find the behavior of HLO. The thing that I am afraid about is that what if the element-type(`output`), just like the shape('output`), is also dependent on the algorithm. 

From HLO spec
```
It and the required shape and valid values are dependent on the algorithm used. 
```
it does not say that the element-type is algo dependent, but OTOH we does not say what it is.

Can we investigate if this is just an implementation detail and not related to algorithm? If that is the case, then skip any constraint  on  the `output`.

docs/spec_draft.md
```diff
@@ -1718,6 +1719,75 @@ and produces a `result` tensor. More formally,
...
+
+| Name           | Type                                     |
+|----------------|------------------------------------------|
+| `output_state` | tensor of integer or floating-point type |
+| `output`       | tensor of integer or floating-point type |
+
+### Constraints
+
+  * (C1) shape(`initial_state`) $=$ shape(`output_state`).
+  * (C2) The relationship between `initial_state`, `output_state`, and `output`
```
sdasgup3, line 1763: * On `initial_state`, `output_state`, and `output` based on `rng_algorithm`

docs/spec_draft.md
```diff
@@ -1718,6 +1719,75 @@ and produces a `result` tensor. More formally,
...
+      varies:
+    * If `rng_algorithm = DEFAULT`:
+      * `initial_state`, `output_state`, and `output` have backend-specific
```
sdasgup3, line 1766: ... backend specific type requirements.

docs/spec_draft.md
```diff
@@ -1718,6 +1719,75 @@ and produces a `result` tensor. More formally,
...
+### Semantics
+
+Returns an `output` filled with uniform random bits and an updated output state
+`output_state` given an initial state `initial_state` using the pseudorandom
+number generator algorithm `rng_algorithm`. The output is guaranteed to be
+deterministic function of `initial_state`, but it is not guaranteed to be
+deterministic between backends and different compiler versions.
+
+`initial_state` is the initial state of the current random number generation.
+The initial state, required shape, and valid values are dependent on the
```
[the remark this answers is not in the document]
ghpvnist: Correct me if I'm wrong, "required shape" is referring to element type in HLO semantics. So how about:
```
The initial state, supported element type, and valid values are dependent on the
algorithm used.
```

docs/spec_draft.md
ghpvnist: The output should be either `ui32` or `ui64` based on the variant of the algorithm used. As mentioned above, it is generating random bits not random numbers, so 32-bit algorithm would generate random 32-bits (similarly for 64 bits). As for whether 32-input can generate 64-bit output is not possible, since that is not how the algorithm is designed. It is possible, however, to use a 64-bit algorithm to generate a 32-bit value (and throwing away the other half), which is implementation-defined variant of the algorithm as stated in the spec.

ghpvnist: However, DEFAULT algorithm can do whatever it supports.

docs/spec_draft.md
```diff
@@ -1917,6 +1918,74 @@ and produces a `result` tensor. More formally,
...
+
+### Constraints
+
+  * (C1) shape(`initial_state`) $=$ shape(`output_state`).
+  * (C2) For `initial_state`, `output_state`, and `output`:
+    * If `rng_algorithm = DEFAULT`:
+      * `initial_state`, `output_state`, and `output` have backend-specific type
+        requirements.
+    * If `rng_algorithm = THREE_FRY`:
+      * type(`initial_state`) $=$ type(`output_state`) $=$ `2xui64`.
```
sdasgup3, line 1967: type(`initial_state`) $=$ type(`output_state`) $=$ `tensor<2xui64>`. Same at other places.

docs/spec_draft.md
```diff
@@ -1917,6 +1918,74 @@ and produces a `result` tensor. More formally,
...
+### Semantics
+
+Returns an `output` filled with uniform random bits and an updated output state
+`output_state` given an initial state `initial_state` using the pseudorandom
+number generator algorithm `rng_algorithm`. The output is guaranteed to be
+deterministic function of `initial_state`, but it is not guaranteed to be
+deterministic between backends and different compiler versions.
+
+`initial_state` is the initial state of the current random number generation.
+The initial state, supported element type, and valid values are dependent on the
```
sdasgup3, line 1933: The element type  already refers to the valid values.
IMO we can just remove this sentence as the constraints already covers this aspect.

docs/spec_draft.md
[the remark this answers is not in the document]
sdasgup3: We can remove this as this is covered by the constraints section. element type defines the values ..

docs/spec_draft.md
sdasgup3: Thanks for explaining.

sdasgup3:
LGTM except for the few leftover comments.

sdasgup3 (approved):

docs/spec_draft.md
ghpvnist: Added #486

docs/spec_draft.md
```diff
@@ -1917,6 +1918,70 @@ and produces a `result` tensor. More formally,
...
+| Name            | Type                                         |
+|-----------------|----------------------------------------------|
+| `initial_state` | tensor of integer or floating-point type     |
+| `rng_algorithm` | enum of `DEFAULT`, `THREE_FRY`, and `PHILOX` |
+
+### Outputs
+
+| Name           | Type                                     |
+|----------------|------------------------------------------|
+| `output_state` | tensor of integer or floating-point type |
```
burmako, line 1952: As far as I can see from [lax.py](https://github.com/google/jax/blob/40e81c3a8655709c9aa554b52e1a85842786e034/jax/_src/lax/lax.py#L4373-L4380) and the investigation in this PR, both `initial_state` and `output_state` are "1-dimensional tensor of `ui64`"?

docs/spec_draft.md
```diff
@@ -1917,6 +1918,70 @@ and produces a `result` tensor. More formally,
...
+  * (C2) For `initial_state`, `output_state`, and `output`:
+    * If `rng_algorithm = DEFAULT`:
+      * `initial_state`, `output_state`, and `output` have backend-specific type
+        requirements.
+    * If `rng_algorithm = THREE_FRY`:
+      * type(`initial_state`) $=$ type(`output_state`) $=$ `tensor<2xui64>`.
+      * element_type(`output`) $\in$ {`ui32`, `ui64`}.
+    * If `rng_algorithm = PHILOX`:
+      * type(`initial_state`) $=$ type(`output_state`) $=$ `tensor<3xui64>`.
+      * element_type(`output`) $\in$ {`ui32`, `ui64`}.
```
burmako, line 1967: Proposed rewording:

```
  * (C1) `type(initial_state) = type(output_state) = tensor<2xui64>`.
  <the end, no more constraints>
```

Feedback:
  1) Per [hlo_verifier.cc](http://google3/third_party/tensorflow/compiler/xla/service/hlo_verifier.cc;l=928-929;rcl=487909736), I think we can make a stronger claim that `type(initial_state) = type(output_state)`.
  2) Furthermore, according to [lax.py](https://github.com/google/jax/blob/40e81c3a8655709c9aa554b52e1a85842786e034/jax/_src/lax/lax.py#L4373-L4380), both `initial_state` and `output_state` are always `tensor<2xui64>`?
  3) Output's element_type can be covered in Inputs/Outputs.

docs/spec_draft.md
```diff
@@ -1917,6 +1918,70 @@ and produces a `result` tensor. More formally,
...
+
+### Examples
+
+```mlir
+// %initial_state: [1, 2]
+%output_state, %output = "stablehlo.rng_bit_generator"(%initial_state) {
+  rng_algorithm = #stablehlo.rng_algorithm<THREE_FRY>
```
burmako, line 1974: `#stablehlo<rng_algorithm THREE_FRY>`

stablehlo/dialect/StablehloOps.td
```diff
@@ -2788,18 +2788,18 @@ def StableHLO_RngOp : StableHLO_Op<"rng", [InferTensorTypeWithReify, AllElementT
...
 
-    See https://www.tensorflow.org/xla/operation_semantics#rngbitgenerator.
+    See:
+    https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md#stablehlorng_bit_generator
 
     Example:
-
     ```mlir
-    %0, %1 = stablehlo.rng_bit_generator %arg0, algorithm = PHILOX : (tensor<3xui64>) -> (tensor<3xui64>, tensor<2x2xui32>)
+    %output_state, %output = stablehlo.rng_bit_generator %initial_state, algorithm = DEFAULT : (tensor<2xui64>) -> (tensor<2xui64>, tensor<2x2xui64>)
```
burmako, line 2802: Let's use the same example as above.

burmako (changes requested):
I found this to be a fairly underdocumented op. I have a few questions, and provided some probably incorrect feedback to start the discussion.

docs/spec_draft.md
```diff
@@ -1917,6 +1918,70 @@ and produces a `result` tensor. More formally,
...
+`output_state` given an initial state `initial_state` using the pseudorandom
+number generator algorithm `rng_algorithm`. The output is guaranteed to be
+deterministic function of `initial_state`, but it is not guaranteed to be
+deterministic between backends and different compiler versions.
+
+`rng_algorithm` is one of the following, different variants of the algorithm is
+implementation-defined:
+  * `DEFAULT`: Backend specific algorithm.
+  * `THREE_FRY`: Threefry counter-based PRNG algorithm.*
+  * `PHILOX`: Philox algorithm to generate random numbers in parallel.*
```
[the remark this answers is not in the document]
ghpvnist: 2. From what I could tell by the wording in the comment, DEFAULT can be _any_ backend specific prng algorithm, but I'm not 100% sure since common hardwares might already be implementing some variant of Threefry or Philox as the default/common practice or something else entirely:
https://github.com/tensorflow/tensorflow/blob/c82413e988b1dfe12b2ae9c668cd75f3de3fd778/tensorflow/compiler/xla/xla_data.proto#L652-L657
4. I believe Threefry and Philox are both general "algorithms" that work just like different variants of e.g. AES-256 and AES-512 (Here, the "algorithm" is AES). In Threefry, it refers to many "variants" of the "Threefry" algorithm using how many inputs, different number of bits, and # of rounds to compare speed across different hardwares. XLA decided to use a variant Threefry-2-32-20, which uses 2 32-bit input with 20 rounds of shifting, that is not mentioned in the paper. Hence, I'm more inclined to believe that the variant chosen is implementation-defined and `initial_state` follows from the implementer's choice of the variant.

Just fyi, Philox variant used in XLA is 4-32-10. Bit truncation/concat is done to generate other length bits.

docs/spec_draft.md
ghpvnist: I separated into these constraints since I do not know what `DEFAULT` is capable of doing. It can possibly have different shape and element type.

docs/spec_draft.md
```diff
@@ -1917,6 +1918,70 @@ and produces a `result` tensor. More formally,
...
+|-----------------|----------------------------------------------|
+| `initial_state` | tensor of integer or floating-point type     |
+| `rng_algorithm` | enum of `DEFAULT`, `THREE_FRY`, and `PHILOX` |
+
+### Outputs
+
+| Name           | Type                                     |
+|----------------|------------------------------------------|
+| `output_state` | tensor of integer or floating-point type |
+| `output`       | tensor of integer or floating-point type |
```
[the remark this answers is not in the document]
ghpvnist: The floating-point originally came from the ODS, assuming that maybe the DEFAULT would take in floating-point element types since prng of a floating point number is rare but still a valid use case. As for whether we should allow `ui32` or `ui64` vs unsigned integer type, this is an open question. How about restricting the types as much as possible for now to `ui32` or `ui64` and open an issue to track this?

docs/spec_draft.md
ghpvnist: `initial_state` and `output_state` for DEFAULT algorithm may be able to take n-dimensional tensor of element types other than `ui64`. Similar to my comment above, I propose that we restrict to 1-dimensional tensor of `ui64` and opening an issue to track this in the same ticket.

docs/spec_draft.md
ghpvnist: Since DEFAULT is open ended, it makes sense to have the current proposed. However, we could use your proposed rewording and note this in the same issue.

docs/spec_draft.md
ghpvnist: Created issue #643

docs/spec_draft.md
```diff
@@ -2527,6 +2528,68 @@ hidden state.
...
+### Semantics
+
+Returns an `output` filled with uniform random bits and an updated output state
+`output_state` given an initial state `initial_state` using the pseudorandom
+number generator algorithm `rng_algorithm`. The output is guaranteed to be
+deterministic function of `initial_state`, but it is not guaranteed to be
+deterministic between implementations.
+
+`rng_algorithm` is one of the following:
+  * `DEFAULT`: Backend specific algorithm.
```
burmako, line 2542: How about "Implementation-defined algorithm", to be in line with how we're calling similar things elsewhere in the spec.

docs/spec_draft.md
burmako: Let's restrict to "1-dimensional tensor of type `u64`". From what I understand, it looks like the significant flexibility of `DEFAULT` might be useful privately to XLA, but the public interface is still `tensor<2xui64>`.

docs/spec_draft.md
burmako: Let's bias towards the simplest formulation (i.e the one proposed above) and sort this out in the ticket.

burmako (changes requested):

docs/spec_draft.md
ghpvnist: The proposed rewording is missing constraint for `output`, and PHILOX algorithm requires `tensor<3xui64>` whereas THREE_FRY is `tensor<2xui64>`. How about:
```
  * (C1) If `rng_algorithm` is:
    * `THREE_FRY`: `type(initial_state) = type(output_state) = tensor<2xui64>`.
    * `PHILOX`: `type(initial_state) = type(output_state) = tensor<3xui64>`.
  * (C2) element_type(`output`) $\in$ {`ui32`, `ui64`}.
```

burmako: PHILOX (at least on XLA CPU) is fine with both `2xui64` and `3xui64` (`2xui64` is important because this is what JAX produces). Perhaps we can use the following format to simplify speccing this:

```
  * (C1) `type(initial_state) = type(output_state)`.
  * (C2) `size(initial_state)` depends on `rng_algorithm`:
    * `DEFAULT`: implementation-defined.
    * `THREE_FRY`: `2`.
    * `PHILOX`: `2` or `3`.
```

For the element type, where is the restriction coming from? [xla_builder.cc](https://github.com/tensorflow/tensorflow/blob/8b7373a66dfcb8ccf994ea0b1d720b34a4d32cbf/tensorflow/compiler/xla/client/xla_builder.cc#L2566-L2601) happily admits all floating-point and integer types. [hlo_verifier.cc](https://github.com/tensorflow/tensorflow/blob/8b7373a66dfcb8ccf994ea0b1d720b34a4d32cbf/tensorflow/compiler/xla/service/hlo_verifier.cc#L920-L938) doesn't have an opinion. I would imagine that we don't need to cover this in Constraints?

burmako (changes requested):

docs/spec_draft.md
ghpvnist: Thanks for catching this; I wasn't aware that PHILOX could take `2xui64` in [JAX](https://github.com/google/jax/blob/40e81c3a8655709c9aa554b52e1a85842786e034/jax/_src/lax/lax.py#L4373-L4380) because [XLA doc](https://www.tensorflow.org/xla/operation_semantics#rngbitgenerator) showed otherwise.

Since StableHLO supports 4-bit integers while [xla_builder.cc](https://github.com/tensorflow/tensorflow/blob/8b7373a66dfcb8ccf994ea0b1d720b34a4d32cbf/tensorflow/compiler/xla/client/xla_builder.cc#L2566-L2601) does not, and [hlo_verifier.cc](https://github.com/tensorflow/tensorflow/blob/8b7373a66dfcb8ccf994ea0b1d720b34a4d32cbf/tensorflow/compiler/xla/service/hlo_verifier.cc#L920-L938) also does not have verification for supported element types, I believe we have to call out exactly which element types are supported regardless, either in Inputs section or in Constraints. I'm leaning towards the Constraints, which may look like below. Suggestions?

```
  * (C1) type(`initial_state`) $=$ type(`output_state`).
  * (C2) size(`initial_state`) depends on `rng_algorithm`:
    * `DEFAULT`: implementation-defined.
    * `THREE_FRY`: `2`.
    * `PHILOX`: `2` or `3`.
  * (C3) element_type(`initial_state`) depends on `rng_algorithm`:
    * `DEFAULT`: implementation-defined.
    * `THREE_FRY`: `ui8`, `ui16`, `ui32`, `ui64`, `si8`, `si16`, `si32`, `si64`,
      `bf16`, `f16`, `f32`, or `f64`.
    * `PHILOX`: `ui8`, `ui16`, `ui32`, `ui64`, `si8`, `si16`, `si32`, `si64`,
      `bf16`, `f16`, `f32`, or `f64`.
```

burmako: The situation with 4-bit integers is not unique to this op. E.g. `stablehlo.add` supports them, while `kAdd` doesn't. I wouldn't add special treatment for si4/ui4 here in this specific op. Opened #661 to track the general question.

burmako (changes requested):
See the comment above regarding 4-bit integers.

burmako (approved):

Issue #223, opened 2022-09-29
Spec parity with StableHLO spec and HLO semantics for floating-point arithmetic ops


sdasgup3:
### Request description

The Specification for most of the floating-point arithmetic ops are based on the IEEE 754 standard. https://github.com/openxla/stablehlo/issues/119 demands checking StableHLO spec against HLO specification (which is what the current StableHLO ODS, verifiers, and type inferencers are based on) for parity. For floating-point arithmetic operations, that means that we need to check if the HLO semantics meets all the exception behaviors mentioned in the standard. Unfortunately we are not doing that at the moment for the following reasons:
1. Time consuming to follow the standard for every behavior of a particular op, write tests and run some form of executor of HLO (an interpreter is we are using now).
2.  Maybe we should check the behavior on hardware platforms running HLO: something  we are not doing yet.
  
My proposal is  to addressed this aspect of parity check later.

sdasgup3:
Also for stablehlo.compare when the compare_type = TOTALORDER, we need to double check if the behavior of HLO semantics matches with "5.10 Details of totalOrder predicate".

burmako:
Another item that came up today is the behavior of shift ops when shifting beyond the width of the integer type. Here's what MHLO does in that case: https://github.com/tensorflow/mlir-hlo/commit/ea5251b9e5783c8cd4298571d4b2863e74dcdb28.

Issue #180, opened 2022-09-22
Clarify float->int rounding behavior in convert


sdasgup3:
### Request description

See https://github.com/iree-org/iree/issues/6160, is the current expectation that an MHLO consumer can arbitrarily decide rounding behavior? (and so backends/devices may differ)

ghpvnist:
Unassigning issue for now since the issue has not been worked on for some time. This should be addressed along with #969.

PR #421, opened 2022-11-01
Add spec for ConvertOp

### LINKED ISSUE #418: Add spec for ConvertOp
### Request description

To add spec for `stablehlo.convert`

ghpvnist:
closes #418

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -1833,6 +1833,17 @@ overflow, the result is implementation-defined and one of the following:
   * saturation to $2^{n-1} - 1$ (or $-2^{n-1}$) for signed overflow and
     saturation to $2^n - 1$ (or $0$) for unsigned overflow.
 
+For conversions involving floating-point-to-floating-point or
+complex-to-complex, if the source value can be exactly represented in the
+destination type, the result value is that exact representation. If the source
+value is between two adjacent destination values, the result is
+implementation-defined. Otherwise, the behavior is undefined.
+
+For conversions involving floating-point-to-complex or
+complex-to-floating-point, the imaginary value is zero or truncated,
+respectively. The conversion of inner representation of complex element type
+follows the floating-point-to-floating-point conversion.
+
 For conversions involving floating-point-to-integer or complex-to-integer,
 the fractional part is truncated. The behavior is undefined if the truncated
 value cannot be represented in the destination type.
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -1833,31 +1833,36 @@ overflow, the result is implementation-defined and one of the following:
   * saturation to $2^{n-1} - 1$ (or $-2^{n-1}$) for signed overflow and
     saturation to $2^n - 1$ (or $0$) for unsigned overflow.
 
-For conversions involving floating-point-to-floating-point or
-complex-to-complex, if the source value can be exactly represented in the
-destination type, the result value is that exact representation. If the source
-value is between two adjacent destination values, the result is
+For conversions involving **floating-point-to-floating-point** or
+**integer-to-floating-point**, if the source value can be exactly represented in
+the destination type, the result value is that exact representation. If the
+source value is between two adjacent destination values, the result is
 implementation-defined. Otherwise, the behavior is undefined.
 
-For conversions involving floating-point-to-complex or
-complex-to-floating-point, the imaginary value is zero or truncated,
-respectively. The conversion of inner representation of complex element type
-follows the floating-point-to-floating-point conversion.
+Conversion involving **complex-to-complex** follows the same behavior of
+**floating-point-to-floating-point** conversions for converting the inner
+representation.
 
-For conversions involving floating-point-to-integer or complex-to-integer,
-the fractional part is truncated. The behavior is undefined if the truncated
-value cannot be represented in the destination type.
+For conversions involving **floating-point-to-complex** or
+**complex-to-floating-point**, the destination imaginary value is zeroed or the
+source imaginary value is ignored, respectively. The conversion of inner
+representation of complex element type follows the
+**floating-point-to-floating-point** conversion.
 
-For conversions involving integer-to-floating-point or integer-to-complex,
-the destination value is exact if possible and for conversions involving
-rounding, the result is rounded using any normalizing rounding-direction
-attributes `round` from the IEEE-754 specification (e.g. `roundTiesToEven`). If
-the value being converted is outside of the range of values destination type can
-represent, the result is undefined.
+Conversions involving **integer-to-complex** follows the same behavior as
+**integer-to-floating-point** conversion while converting the source integer to
+destination real part. The destination imaginary part is zeroed.
 
-For boolean-to-any-supported-type conversions, the value `false` is converted to
-zero-equivalent, and the value `true` is converted to one-equivalent and
-vice-versa.
+For conversions involving **floating-point-to-integer**, the fractional part is
+truncated. The behavior is undefined if the truncated value cannot be
+represented in the destination type. Conversions involving
+**complex-to-integer** follows the same behavior while converting the source
+real part to destination integer. The source imaginary part is ignored.
+
+For **boolean-to-any-supported-type** conversions, the value `false` is
+converted to zero, and the value `true` is converted to one. For
+**any-supported-type-to-boolean** conversions, a zero value is converted to
+`false` and any non-zero value is converted to `true`.
 
 ### Inputs
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -1825,7 +1825,7 @@ Performs an element-wise conversion of values similar to an element-wise
 `static_cast` in C++ from one element type to another on `operand` tensor and
 produces a `result` tensor.
 
-For conversions involving integer-to-integer, if there is an unsigned/signed
+For conversions involving **integer-to-integer**, if there is an unsigned/signed
 overflow, the result is implementation-defined and one of the following:
   * mathematical result modulo $2^n$, where n is the bit width of the result,
     for unsigned overflow. For signed integer overflow, wraps the result around
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -1822,46 +1822,180 @@ Produces an `output` tensor from a constant `value`.
 ### Semantics
 
 Performs an element-wise conversion of values from one element type to another
-on `operand` tensor and produces a `result` tensor.
+on `operand` tensor and produces a `result` tensor. The result is rounded using
+any normalizing rounding-direction attributes `round` from the IEEE-754
+specification (e.g. `roundTiesToEven`).
 
-More formally, conversion from element `A` of `operand` to element `B` of
-another type is defined as the following:
-  * Conversion of same element type with different precision:
-    * The result is rounded using any normalizing rounding-direction attributes
-      from the IEEE-754 specification (e.g. `roundTiesToEven`).
-  * Conversion from integer to:
-    * signed integer: signExtend or truncate.
-    * unsigned integer: zero-extend or truncate.
-    * boolean: `B = A != 0 ? true : false`.
-    * floating-point*: `B = (float) A`.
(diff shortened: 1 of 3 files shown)


docs/spec_draft.md
```diff
@@ -582,6 +583,39 @@ Produces a `result` tensor from a constant `value`.
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.convert
+
+### Semantics
+
+Performs an element-wise conversion of values from one element type to another
+on `operand` tensor.
```
burmako, line 591: As discussed offline, this needs to go into details of what it means to convert from one element type to another.

burmako (changes requested):

docs/spec_draft.md
```diff
@@ -582,6 +583,78 @@ Produces a `result` tensor from a constant `value`.
...
+More formally, conversion from element `A` of `operand` to element `B` of
+another type is defined as the following:
+  * Conversion of same element type with different precision:
+    * The result is rounded using any normalizing rounding-direction attributes
+      from the IEEE-754 specification (e.g. `roundTiesToEven`).
+  * Conversion from integer to:
+    * signed integer: signExtend or truncate.
+    * unsigned integer: zero-extend or truncate.
+    * boolean: `B = A != 0 ? true : false`.
+    * floating-point*: `B = (float) A`.
```
sdasgup3, line 602: As per the standard `The roundTiesToEven rounding-direction attribute shall be the default rounding-direction attribute for results in binary formats.` Is this the behavior of HLO as well? If yes then we can write

* floating point
  * `A` is representable in `T'`, B = round(`A`), where `round` implements the `roundToIntegralTiesToEven`.
  * `A` is not representable in `T'`, B  = `+/-infinity`

docs/spec_draft.md
```diff
@@ -582,6 +583,78 @@ Produces a `result` tensor from a constant `value`.
...
+    * complex*: `B = ((float) A, 0.0)`.
```
sdasgup3, line 603: Can we check that the same consideration with floating point target applies here?

docs/spec_draft.md
```diff
@@ -582,6 +583,78 @@ Produces a `result` tensor from a constant `value`.
...
+  * Conversion from boolean to:
+    * signed integer: `B = A ? 1 : 0`.
+    * unsigned integer: `B = A ? 1 : 0`.
+    * floating-point: `B = A ? 1.0 : 0.0`.
+    * complex: `B = A ? (1.0, 0.0) : (0.0, 0.0)`.
+  * Conversion from floating-point to:
+    * signed integer*: `B = round(A)`.
```
sdasgup3, line 610: Same as integer source and integer target, we need to consider scenarios when the `A` in representable in `T'` or not.

docs/spec_draft.md
```diff
@@ -582,6 +583,78 @@ Produces a `result` tensor from a constant `value`.
...
+    * unsigned integer*: `B = round(A)`.
+    * boolean: `B = A != 0.0 ? true : false`.
+    * floating-point: `B = round(A)` with upcast or downcast.
```
sdasgup3, line 613: Instead of upcast and downcast, try using rounding to convert a representable value in `T` to another representable value `B` in `T'`

docs/spec_draft.md
sdasgup3: >>  Is this the behavior of HLO as well? 

One way to get the default rounding mode in HLO: find two values x and y such that y > x and one is odd and the other is even.And then try to convert a midway value.
   
https://onlinegdb.com/0ekW_Yc-F

docs/spec_draft.md
```diff
@@ -582,6 +583,78 @@ Produces a `result` tensor from a constant `value`.
...
 [Back to Ops](#index-of-ops)
 
+## stablehlo.convert
+
+### Semantics
+
+Performs an element-wise conversion of values from one element type to another
+on `operand` tensor and produces a `result` tensor.
+
+More formally, conversion from element `A` of `operand` to element `B` of
```
sdasgup3, line 593: Also can we have a table like the following to detail all the conversions

| `operand` type, `T` | `result` type `T'` | `B` |
|--|--|--|

and merging multiple cells related to column `result` type like https://jsfiddle.net/7h89y55r/

ghpvnist:
The updated semantics is inspired by these documents:
* https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/literal.cc#L1552-L1611
* https://eel.is/c++draft/conv
* https://eel.is/c++draft/conv.rank

ghpvnist:
Note that for fp-to-fp or complex-to-complex case, I say the rounding is implementation defined vs int-to-fp or int-to-complex, I say rounding is one of ieee spec rounding modes. Let me know which you think is more suitable.

docs/spec_draft.md
```diff
@@ -1420,6 +1421,74 @@ Produces an `output` tensor from a constant `value`.
...
+destination type, the result value is that exact representation. If the source
+value is between two adjacent destination values, the result is
+implementation-defined. Otherwise, the behavior is undefined.
+
+For conversions involving floating-point-to-complex or
+complex-to-floating-point, the imaginary value is zero or truncated,
+respectively. The conversion of inner representation of complex element type
+follows the floating-point-to-floating-point conversion.
+
+For conversions involving floating-point-to-integer or complex-to-integer,
```
sdasgup3, line 1451: Reword:
For conversions involving **floating-point-to-integer**,
the fractional part is truncated. The behavior is undefined if the truncated
value cannot be represented in the destination type. Conversions involving **complex-to-integer** follows the same behavior while converting the source real part to destination integer. Additionally, the source imaginary part is ignored.

docs/spec_draft.md
```diff
@@ -1420,6 +1421,74 @@ Produces an `output` tensor from a constant `value`.
...
+  * saturation to $2^{n-1} - 1$ (or $-2^{n-1}$) for signed overflow and
+    saturation to $2^n - 1$ (or $0$) for unsigned overflow.
+
+For conversions involving floating-point-to-floating-point or
+complex-to-complex, if the source value can be exactly represented in the
+destination type, the result value is that exact representation. If the source
+value is between two adjacent destination values, the result is
+implementation-defined. Otherwise, the behavior is undefined.
+
+For conversions involving floating-point-to-complex or
```
sdasgup3, line 1446: delete

sdasgup3 (approved):

ghpvnist:
Linking issue regarding float -> int rounding behavior #180

docs/spec_draft.md
```diff
@@ -1420,6 +1421,79 @@ Produces an `output` tensor from a constant `value`.
...
+    for unsigned overflow. For signed integer overflow, wraps the result around
+    the representable range $[-2^{n-1},\ 2^{n-1} - 1]$.
+  * saturation to $2^{n-1} - 1$ (or $-2^{n-1}$) for signed overflow and
+    saturation to $2^n - 1$ (or $0$) for unsigned overflow.
+
+For conversions involving **floating-point-to-floating-point** or
+**integer-to-floating-point**, if the source value can be exactly represented in
+the destination type, the result value is that exact representation. If the
+source value is between two adjacent destination values, the result is
+implementation-defined. Otherwise, the behavior is undefined.
```
burmako, line 1444: How about we formulate this less ambitiously for now and leave figuring out the details (which way the rounding goes, is the "otherwise" case IB or UB) to #223?

```
For conversions involving **floating-point-to-floating-point** or
**integer-to-floating-point**, if the source value can be exactly represented in
the destination type, the result value is that exact representation. Otherwise,
the behavior is TBD.
```

docs/spec_draft.md
```diff
@@ -1420,6 +1421,79 @@ Produces an `output` tensor from a constant `value`.
...
+
+Conversion involving **complex-to-complex** follows the same behavior of
+**floating-point-to-floating-point** conversions for converting the inner
+representation.
```
burmako, line 1448: "for converting real and imaginary parts".

docs/spec_draft.md
```diff
@@ -1420,6 +1421,79 @@ Produces an `output` tensor from a constant `value`.
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.convert
+
+### Semantics
+
+Performs an element-wise conversion of values similar to an element-wise
+`static_cast` in C++ from one element type to another on `operand` tensor and
+produces a `result` tensor.
```
burmako, line 1430: Let's drop the mention of `static_cast` from C++ to keep the specification self-contained.

```
Performs an element-wise conversion from one element type to another on 
`operand` tensor and produces a `result` tensor.
```

docs/spec_draft.md
```diff
@@ -1420,6 +1421,79 @@ Produces an `output` tensor from a constant `value`.
...
+
+Conversion involving **complex-to-complex** follows the same behavior of
+**floating-point-to-floating-point** conversions for converting the inner
+representation.
+
+For conversions involving **floating-point-to-complex** or
+**complex-to-floating-point**, the destination imaginary value is zeroed or the
+source imaginary value is ignored, respectively. The conversion of inner
+representation of complex element type follows the
+**floating-point-to-floating-point** conversion.
```
burmako, line 1454: "The conversion of the real part...".

docs/spec_draft.md
```diff
@@ -1420,6 +1421,79 @@ Produces an `output` tensor from a constant `value`.
...
+
+Conversions involving **integer-to-complex** follows the same behavior as
+**integer-to-floating-point** conversion while converting the source integer to
+destination real part. The destination imaginary part is zeroed.
+
+For conversions involving **floating-point-to-integer**, the fractional part is
+truncated. The behavior is undefined if the truncated value cannot be
+represented in the destination type. Conversions involving
```
burmako, line 1462: Similarly to above, I would like to propose say "If the truncated value cannot be represented in the destination type, the behavior is TBD" and leave figuring out the rest to #223.

docs/spec_draft.md
```diff
@@ -1420,6 +1421,79 @@ Produces an `output` tensor from a constant `value`.
...
+**complex-to-integer** follows the same behavior while converting the source
+real part to destination integer. The source imaginary part is ignored.
+
+For **boolean-to-any-supported-type** conversions, the value `false` is
+converted to zero, and the value `true` is converted to one. For
+**any-supported-type-to-boolean** conversions, a zero value is converted to
+`false` and any non-zero value is converted to `true`.
```
burmako, line 1469: Unfortunately, I don't have the time right now to run comprehensive experiments on my own, so I wanted to check in. When writing the spec, have you got a chance to check each supported configuration of operand/result types via `run_hlo_module`?

burmako (changes requested):

docs/spec_draft.md
ghpvnist: In that case, we should refrain from closing #180 prematurely and link this issue to #223 then.

docs/spec_draft.md
ghpvnist: I've done most of the testing for different configurations but definitely not exhaustive (17^2 combinations!), and specifically for booleans, I've confirmed that boolean to any type and any type to boolean behaves as noted in the spec. I think we should also leave this to #223.

docs/spec_draft.md
```diff
@@ -396,7 +396,11 @@ syntax.
    * [complex](#stablehlocomplex)
    * [concatenate](#stablehloconcatenate)
    * [constant](#stablehloconstant)
+<<<<<<< HEAD
    * [convolution](#stablehloconvolution)
+=======
+   * [convert](#stablehloconvert)
+>>>>>>> 07ecb6d (Add spec for ConvertOp)
```
burmako, line 403: Please fix conflict markers.

docs/spec_draft.md
burmako: Agreed, let's leave this to #223.

docs/spec_draft.md
burmako: Sounds good.

burmako (changes requested):

burmako (approved):

PR #429, opened 2022-11-02
Add spec for CompareOp

### LINKED ISSUE #419: Add spec for CompareOp
### Request description

To add spec for `stablehlo.compare`

subhankarshah:
closes #419 
Ticket for verification #571

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -859,7 +859,7 @@ The values of `comparison_direction` have the following semantics:
   * `LT`: `lhs` $\lt$ `rhs`.
 
 The values of `compare_type` have the following semantics:
-  * `FLOAT`: If element-type is floating-type, standard floating-point
+  * `FLOAT`: If element-type is floating-type, ordered floating-point
              comparison of `lhs` and `rhs` is performed as specified in
              IEEE-754, if element-type is complex-type lexicographic comparison
              on (real, imaginary) pairs is performed with corner cases TBD.
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -861,17 +861,19 @@ formally, `result[i0, ..., iR-1]` =
   * `LT`: `lhs` $\lt$ `rhs`.
 
 `compare_type` can be one of the following:
-  * `FLOAT`: Floating type comparison of `lhs` and `rhs`.
-  * `TOTALORDER`: Total ordering floating type comparison of `lhs` and `rhs`.
+  * `FLOAT`: Standard floating-point comparison of `lhs` and `rhs` as specified
+             in IEEE-754.
+  * `TOTALORDER`: `totalOrder` floating-point comparison of `lhs` and `rhs` as
+                  specified in IEEE-754.
   * `SIGNED`: Signed comparison of `lhs` and `rhs`.
   * `UNSIGNED`: Unsigned comparison of `lhs` and `rhs`.
 
 ### Inputs
 
-| Name                   | Type                                                   |
-|------------------------|--------------------------------------------------------|
-| `lhs`                  | tensor of any supported type                           |
-| `rhs`                  | tensor of any supported type                           |
+| Name                   | Type                                                    |
+|------------------------|---------------------------------------------------------|
+| `lhs`                  | tensor of any supported type                            |
+| `rhs`                  | tensor of any supported type                            |
 | `comparison_direction` | enum of `EQ`, `NE`, `GE`, `GT`, `LE`, and `LT`          |
 | `compare_type`         | enum of `FLOAT`, `TOTALORDER`, `SIGNED`, and `UNSIGNED` |
 
@@ -885,6 +887,12 @@ formally, `result[i0, ..., iR-1]` =
 
   * (C1) `lhs` and `rhs` have the same element type.
   * (C2) `lhs`, `rhs`, and `result` have the same shape.
+  * (C3) Let `E` be the `lhs` element type:
+    * If `E` is complex type, `comparison_direction` $\in$ `{EQ, NE}`,
+      `compare_type` = `FLOAT`.
+    * If `E` is signed integer type, `compare_type` = `SIGNED`.
+    * If `E` is unsigned integer or boolean type, `compare_type` = `UNSIGNED`.
+    * If `E` is floating-type, `compare_type` $\in$ `{FLOAT, TOTALORDER}`.
 
 ### Examples
 
@@ -892,6 +900,7 @@ formally, `result[i0, ..., iR-1]` =
 // %lhs: [1.0, 3,0]
 // %rhs: [1.1, 2.9]
 %result = "stablehlo.compare"(%lhs, %rhs) {
+  compare_type = #stablehlo<compare_type FLOAT>,
   comparison_direction = #stablehlo<comparison_direction LT>
 } : (tensor<2xf32>, tensor<2xf32>) -> tensor<2xi1>
 // %result: [true, false]
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -889,7 +889,8 @@ The values of `compare_type` have the following semantics:
   * (C2) `lhs`, `rhs`, and `result` have the same shape.
   * (C3) Given `E` is the `lhs` element type, the following are legal values of
          `compare_type`:
-    * If `E` is complex type, `compare_type` = `FLOAT`.
+    * If `E` is complex type, `compare_type` = `FLOAT` and
+      `comparison_direction` $\in$ `{EQ, NE}`.
     * If `E` is signed integer type, `compare_type` = `SIGNED`.
     * If `E` is unsigned integer or boolean type, `compare_type` = `UNSIGNED`.
     * If `E` is floating-type, `compare_type` $\in$ `{FLOAT, TOTALORDER}`.
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -848,11 +848,9 @@ operations correspond to [stablehlo.minimum](#stablehlominimum) and
 ### Semantics
 
 Performs element-wise comparison of `lhs` and `rhs` tensors according to
-`comparison_direction` and `compare_type`, and produces a `result` tensor. More
-formally, `result[i0, ..., iR-1]` =
-`comparison_direction lhs[i0, ..., iR-1], rhs[i0, ..., iR-1]`.
+`comparison_direction` and `compare_type`, and produces a `result` tensor.
 
-`comparison_direction` can be one of the following:
+The values of `comparison_direction` have the following semantics:
   * `EQ`: `lhs` $=$ `rhs`.
   * `NE`: `lhs` $\ne$ `rhs`.
   * `GE`: `lhs` $\ge$ `rhs`.
@@ -860,9 +858,11 @@ formally, `result[i0, ..., iR-1]` =
   * `LE`: `lhs` $\le$ `rhs`.
   * `LT`: `lhs` $\lt$ `rhs`.
 
-`compare_type` can be one of the following:
-  * `FLOAT`: Standard floating-point comparison of `lhs` and `rhs` as specified
-             in IEEE-754.
+The values of `compare_type` have the following semantics:
+  * `FLOAT`: If element-type is floating-type, standard floating-point
+             comparison of `lhs` and `rhs` is performed as specified in
+             IEEE-754, if element-type is complex-type lexicographic comparison
+             on (real, imaginary) pairs is performed with corner cases TBD.
   * `TOTALORDER`: `totalOrder` floating-point comparison of `lhs` and `rhs` as
                   specified in IEEE-754.
   * `SIGNED`: Signed comparison of `lhs` and `rhs`.
@@ -887,9 +887,9 @@ formally, `result[i0, ..., iR-1]` =
 
   * (C1) `lhs` and `rhs` have the same element type.
   * (C2) `lhs`, `rhs`, and `result` have the same shape.
-  * (C3) Let `E` be the `lhs` element type:
-    * If `E` is complex type, `comparison_direction` $\in$ `{EQ, NE}`,
-      `compare_type` = `FLOAT`.
+  * (C3) Given `E` is the `lhs` element type, the following are legal values of
+         `compare_type`:
+    * If `E` is complex type, `compare_type` = `FLOAT`.
     * If `E` is signed integer type, `compare_type` = `SIGNED`.
     * If `E` is unsigned integer or boolean type, `compare_type` = `UNSIGNED`.
     * If `E` is floating-type, `compare_type` $\in$ `{FLOAT, TOTALORDER}`.
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -862,8 +862,8 @@ The values of `compare_type` have the following semantics:
   * `FLOAT`: If element-type is floating-type, ordered floating-point comparison
              of `lhs` and `rhs` is performed for all values of
              `comparison_direction` except for `NE` for which unordered
-             comparison is performed. If element-type is complex-type, only `EQ`
-             and `NE` comparison of (real, imag) pairs is supported.
+             comparison is performed. If element-type is complex-type,
+             lexicographic comparison of (real, imag) pairs is performed.
   * `TOTALORDER`: `totalOrder` floating-point comparison of `lhs` and `rhs` as
                   specified in IEEE-754.
   * `SIGNED`: Signed comparison of `lhs` and `rhs`.
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -187,6 +187,7 @@ described below)
    * [ceil](#stablehloceil)
    * [cholesky](#stablehlocholesky)
    * [clamp](#stablehloclamp)
+   * [compare](#stablehlocompare)
    * [complex](#stablehlocomplex)
    * [concatenate](#stablehloconcatenate)
    * [constant](#stablehloconstant)
@@ -838,6 +839,67 @@ operations correspond to [stablehlo.minimum](#stablehlominimum) and
 // %max: [10, 15, 20]
 %result = "stablehlo.clamp"(%min, %operand, %max) : (tensor<3xi32>, tensor<3xi32>, tensor<3xi32>) -> tensor<3xi32>
 // %result: [5, 13, 20]
+```
+
+[Back to Ops](#index-of-ops)
+
+## stablehlo.compare
+
+### Semantics
+
+Performs element-wise comparison of `lhs` and `rhs` tensors based on
+`comparison_direction` and produces a `result` tensor. More formally,
+`result[i0, ..., iR-1]` =
+`compare_function(lhs[i0, ..., iR-1], rhs[i0, ..., iR-1])`.
+
+### Inputs
+
+| Name                   | Type                          |
+|------------------------|-------------------------------|
+| `lhs`                  | tensor of any supported type  |
+| `rhs`                  | tensor of any supported type  |
+| `comparison_direction` | type of comparison to perform |
+
+### Outputs
+
+| Name     | Type                |
+|----------|---------------------|
+| `result` | tensor of type `i1` |
+
+### Constraints
+
+  * (C1) `lhs` and `rhs` have the same type.
+  * (C2) `lhs` and `result` have the same shape.
+
+### Examples
+
+```mlir
+// compare booleans
+
+// %lhs: [true, false]
+// %rhs: [false, true]
+%result = "stablehlo.compare"(%lhs, %rhs) {
+  comparison_direction = #stablehlo<comparison_direction LT>
+} : (tensor<2xi1>, tensor<2xi1>) -> tensor<2xi1>
+// %result: [false, true]
+
+// compare integers
+
+// %lhs: [1, 3]
+// %rhs: [2, 2]
+%result = "stablehlo.compare"(%lhs, %rhs) {
+  comparison_direction = #stablehlo<comparison_direction LT>
+} : (tensor<2xi32>, tensor<2xi32>) -> tensor<2xi1>
+// %result: [true, false]
+
+// compare floats
+
+// %lhs: [1.0, 3,0]
+// %rhs: [1.1, 2.9]
+%result = "stablehlo.compare"(%lhs, %rhs) {
+  comparison_direction = #stablehlo<comparison_direction LT>
+} : (tensor<2xf32>, tensor<2xf32>) -> tensor<2xi1>
+// %result: [true, false]
 ```
 
 [Back to Ops](#index-of-ops)
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -850,7 +850,10 @@ operations correspond to [stablehlo.minimum](#stablehlominimum) and
 Performs element-wise comparison of `lhs` and `rhs` tensors according to
 `comparison_direction` and `compare_type`, and produces a `result` tensor.
 
-The values of `comparison_direction` have the following semantics:
+The values of `comparison_direction` and `compare_type` have the following
+semantics:
+
+For integer and boolean element types:
   * `EQ`: `lhs` $=$ `rhs`.
   * `NE`: `lhs` $\ne$ `rhs`.
   * `GE`: `lhs` $\ge$ `rhs`.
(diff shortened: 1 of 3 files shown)


stablehlo/dialect/StablehloOps.td
```diff
@@ -1423,19 +1423,17 @@ def StableHLO_CompareOp: StableHLO_Op<"compare", [Pure, SameOperandsElementType,
...
-
     ```mlir
-    %0 = stablehlo.compare LT, %arg0, %arg1 : (tensor<2xi32>, tensor<2xi32>) -> tensor<2xi1>
-    %1 = stablehlo.compare LT, %arg0, %arg1, TOTALORDER : (tensor<2xi32>, tensor<2xi32>) -> tensor<2xi1>
+    %result = "stablehlo.compare"(%lhs, %rhs) {
+      comparison_direction = #stablehlo<comparison_direction LT>
+    } : (tensor<2xf32>, tensor<2xf32>) -> tensor<2xi1>
     ```
   }];
   let arguments = (ins
```
subhankarshah, line 1439: in line 1445 below `let results = (outs HLO_PredTensor);`
result has not alias $result

docs/spec_draft.md
```diff
@@ -485,6 +486,67 @@ IEEE-754 specification.
...
+`result[i0, ..., iR-1]` =
+`compare_function(lhs[i0, ..., iR-1], rhs[i0, ..., iR-1])`.
+
+### Inputs
+
+| Name                   | Type                          |
+|------------------------|-------------------------------|
+| `lhs`                  | tensor of any supported type  |
+| `rhs`                  | tensor of any supported type  |
+| `comparison_direction` | type of comparison to perform |
```
subhankarshah, line 504: type of `comparison_direction` is an enum
I have skipped an optional attribute `comparison_type`

burmako: All the values of `comparison_direction` need to be documented in the Semantics section. Similarly for `comparison_type` - if the spec skips it, how will the users know what it means?

burmako (changes requested):

docs/spec_draft.md
subhankarshah: I just needed some pointers here, I will add an enum section in the spec and specify both `comparison_direction` and `comparison_type` enums.

subhankarshah: Thanks

stablehlo/dialect/StablehloOps.td
subhankarshah: Should I go ahead and add that?

docs/spec_draft.md
```diff
@@ -485,6 +486,67 @@ IEEE-754 specification.
...
 [Back to Ops](#index-of-ops)
 
+## stablehlo.compare
+
+### Semantics
+
+Performs element-wise comparison of `lhs` and `rhs` tensors based on
+`comparison_direction` and produces a `result` tensor. More formally,
+`result[i0, ..., iR-1]` =
+`compare_function(lhs[i0, ..., iR-1], rhs[i0, ..., iR-1])`.
```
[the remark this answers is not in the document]
subhankarshah: Will define it in the spec. Thanks

stablehlo/dialect/StablehloOps.td
```diff
@@ -1423,19 +1423,17 @@ def StableHLO_CompareOp: StableHLO_Op<"compare", [Pure, SameOperandsElementType,
...
+    https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md#stablehlocompare
 
     Example:
-
     ```mlir
-    %0 = stablehlo.compare LT, %arg0, %arg1 : (tensor<2xi32>, tensor<2xi32>) -> tensor<2xi1>
-    %1 = stablehlo.compare LT, %arg0, %arg1, TOTALORDER : (tensor<2xi32>, tensor<2xi32>) -> tensor<2xi1>
+    %result = "stablehlo.compare"(%lhs, %rhs) {
+      comparison_direction = #stablehlo<comparison_direction LT>
+    } : (tensor<2xf32>, tensor<2xf32>) -> tensor<2xi1>
```
GleasonK, line 1436: Use prettyprint.

```
%result = stablehlo.compare LT, %lhs, %rhs : (tensor<2xf32>, tensor<2xf32>) -> tensor<2xi1>
```

stablehlo/dialect/StablehloOps.td
GleasonK: I'm in favor of single-output ops using `$result` but not sure if we should proactively add it without a good use case. I've hit some integration snags adding `$result` with the [`getResult`](https://mlir.llvm.org/doxygen/classmlir_1_1Operation.html#a56f58b55c803b3313da7b4a04a3d542d) function, now that we have prefixed accessors. The old function gets shadowed on the op.

Come to think of it - I'm not sure why this method gets shadowed since it has a different signature, I'll add that to my list of things to investigate eventually.. After adding `$result`, it should be that `getResult() == getResult(0)`.

burmako: "I'm not sure why this method gets shadowed since it has a different signature". There's already `getResult()` in the [`OneResult`](https://github.com/llvm/llvm-project/blob/f62d8a1a5044df7b8d72033d056375b4ab256012/mlir/include/mlir/IR/OpDefinition.h#L610) trait upstream.

GleasonK: got it, so no functional benefit to adding `$result` here, unless we needed to use it in pretty printing which isn't the case with `compare`. Probably best to not add it then

docs/spec_draft.md
ghpvnist: For the sake of completeness, here are the enums for `comparison_direction`:
https://github.com/openxla/stablehlo/blob/46e99f875a1b2ca7bbf17187b388904f8803e3e7/stablehlo/dialect/StablehloEnums.td#L103-L108

Here are the enums for `comparison_type`, and I suggest writing it in similar fashion to [this](https://github.com/openxla/stablehlo/pull/410/files#diff-c4355252a2f05bfb0090a2b8f6af3a2e83b4c89504fc408fe1cb52c0bd2e8f14R1832-R1836):
https://github.com/openxla/stablehlo/blob/46e99f875a1b2ca7bbf17187b388904f8803e3e7/stablehlo/dialect/StablehloEnums.td#L127-L131

docs/spec_draft.md
```diff
@@ -485,6 +486,67 @@ IEEE-754 specification.
...
+|------------------------|-------------------------------|
+| `lhs`                  | tensor of any supported type  |
+| `rhs`                  | tensor of any supported type  |
+| `comparison_direction` | type of comparison to perform |
+
+### Outputs
+
+| Name     | Type                |
+|----------|---------------------|
+| `result` | tensor of type `i1` |
```
ghpvnist, line 510: tensor of boolean type

docs/spec_draft.md
```diff
@@ -485,6 +486,67 @@ IEEE-754 specification.
...
+
+### Constraints
+
+  * (C1) `lhs` and `rhs` have the same type.
```
ghpvnist, line 514: Here is a proposal for separating the constraints more clearly (element type of result is already in Outputs):
```
* (C1) `lhs` and `rhs` have the same element type.
* (C2) `lhs`, `rhs`, and `result` have the same shape.
```

ghpvnist (changes requested):

docs/spec_draft.md
```diff
@@ -485,6 +486,62 @@ IEEE-754 specification.
...
+  * `SIGNED`: Signed comparison of `lhs` and `rhs`
+  * `UNSIGNED`: Unsigned comparison of `lhs` and `rhs`
+
+### Inputs
+
+| Name                   | Type                                                |
+|------------------------|-----------------------------------------------------|
+| `lhs`                  | tensor of any supported type                        |
+| `rhs`                  | tensor of any supported type                        |
+| `comparison_direction` | enum of `EQ`, `NE`, `GE`, `GT`, `LE`, `LT`          |
```
ghpvnist, line 518: Nit:
`LE`, and `LT`
`SIGNED`, and `UNSIGNED`

docs/spec_draft.md
```diff
@@ -485,6 +486,62 @@ IEEE-754 specification.
...
+
+### Semantics
+
+Performs element-wise comparison of `lhs` and `rhs` tensors according to
+`comparison_direction` and `compare_type`, and produces a `result` tensor. More
+formally, `result[i0, ..., iR-1]` =
+`comparison_direction lhs[i0, ..., iR-1], rhs[i0, ..., iR-1]`.
+
+`comparison_direction` can be one of the following:
+  * `EQ`: `lhs` $=$ `rhs`
```
ghpvnist, line 499: Let's put consistent punctuations in all the bullet points (.)

ghpvnist (approved):
Minor changes and LGTM

docs/spec_draft.md
```diff
@@ -485,6 +486,71 @@ IEEE-754 specification.
...
+  * `NE`: `lhs` $\ne$ `rhs`.
+  * `GE`: `lhs` $\ge$ `rhs`.
+  * `GT`: `lhs` $\gt$ `rhs`.
+  * `LE`: `lhs` $\le$ `rhs`.
+  * `LT`: `lhs` $\lt$ `rhs`.
+
+`compare_type` can be one of the following:
+  * `FLOAT`: Standard floating-point comparison of `lhs` and `rhs` as specified
+             in IEEE-754.
+  * `TOTALORDER`: `totalOrder` floating-point comparison of `lhs` and `rhs` as
```
sdasgup3, line 509: Added https://github.com/openxla/stablehlo/issues/223#issuecomment-1312078643
to make sure that the claim is valid.

docs/spec_draft.md
```diff
@@ -485,6 +486,71 @@ IEEE-754 specification.
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.compare
+
+### Semantics
+
+Performs element-wise comparison of `lhs` and `rhs` tensors according to
+`comparison_direction` and `compare_type`, and produces a `result` tensor. More
+formally, `result[i0, ..., iR-1]` =
```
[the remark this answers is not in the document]
subhankarshah: agree

docs/spec_draft.md
```diff
@@ -485,6 +486,71 @@ IEEE-754 specification.
...
+`comparison_direction` can be one of the following:
+  * `EQ`: `lhs` $=$ `rhs`.
+  * `NE`: `lhs` $\ne$ `rhs`.
+  * `GE`: `lhs` $\ge$ `rhs`.
+  * `GT`: `lhs` $\gt$ `rhs`.
+  * `LE`: `lhs` $\le$ `rhs`.
+  * `LT`: `lhs` $\lt$ `rhs`.
+
+`compare_type` can be one of the following:
+  * `FLOAT`: Standard floating-point comparison of `lhs` and `rhs` as specified
```
sdasgup3, line 507: Just for clarity:
`FLOAT`: Ordered floating point comparison of .... 

means that if any one of the operand is a NaN it return would be false. This goes with the spec
```
Four mutually exclusive relations are possible: less than, equal, greater than, and unordered. The last case
arises when at least one operand is NaN.
```
That is if any of the operand is NaN, then the result of compare is `unordered`, which is not equal to other options. 

Also the behavior of the evaluator:
```
func.func @main() -> tensor<i1> {
  %0 = mhlo.constant dense<0x7fc00000> : tensor<f32>
  %1 = mhlo.constant dense<7.5> : tensor<f32>
  %2 = "mhlo.compare"(%0, %1) {compare_type = #mhlo<comparison_type FLOAT>, comparison_direction = #mhlo<comparison_direction XXX>}   :(tensor<f32>, tensor<f32>) -> tensor<i1>
  func.return %2 : tensor<i1>
}

the program will be evaluated to false irrespectove of the value of XXX.
```

docs/spec_draft.md
```diff
@@ -485,6 +486,71 @@ IEEE-754 specification.
...
+## stablehlo.compare
+
+### Semantics
+
+Performs element-wise comparison of `lhs` and `rhs` tensors according to
+`comparison_direction` and `compare_type`, and produces a `result` tensor. More
+formally, `result[i0, ..., iR-1]` =
+`comparison_direction lhs[i0, ..., iR-1], rhs[i0, ..., iR-1]`.
+
+`comparison_direction` can be one of the following:
```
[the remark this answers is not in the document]
subhankarshah: Followed the convention as outlined in [this pr](https://github.com/openxla/stablehlo/pull/410/files#diff-c4355252a2f05bfb0090a2b8f6af3a2e83b4c89504fc408fe1cb52c0bd2e8f14R2303)

docs/spec_draft.md
```diff
@@ -485,6 +486,71 @@ IEEE-754 specification.
...
+### Constraints
+
+  * (C1) `lhs` and `rhs` have the same element type.
+  * (C2) `lhs`, `rhs`, and `result` have the same shape.
+  * (C3) Let `E` be the `lhs` element type:
+    * If `E` is complex type, `comparison_direction` $\in$ `{EQ, NE}`,
+      `compare_type` = `FLOAT`.
+    * If `E` is signed integer type, `compare_type` = `SIGNED`.
+    * If `E` is unsigned integer or boolean type, `compare_type` = `UNSIGNED`.
+    * If `E` is floating-type, `compare_type` $\in$ `{FLOAT, TOTALORDER}`.
```
sdasgup3, line 538: From https://github.com/tensorflow/tensorflow/blob/4e0e42544237ce612d5ec3e4dc6a6a8ab8e58df2/tensorflow/compiler/xla/comparison_util.cc#L240 it seems that the TOTALORDER is never being uses:

When I tried the following on the hlo evaluator:
```
func.func @main() -> tensor<i1> {
  %0 = mhlo.constant dense<-0.0> : tensor<f32>
  %1 = mhlo.constant dense<0.0> : tensor<f32>
  %2 = "mhlo.compare"(%0, %1) {compare_type = #mhlo<comparison_type TOTALORDER>, comparison_direction = #mhlo<comparison_direction EQ>}   :(tensor<f32>, tensor<f32>) -> tensor<i1>
  func.return %2 : tensor<i1>
}
```
the result is same as compare_type == FLOAT.

Did you have the same observation about TOTALORDER being unused?

docs/spec_draft.md
```diff
@@ -485,6 +486,71 @@ IEEE-754 specification.
...
+| Name     | Type                   |
+|----------|------------------------|
+| `result` | tensor of boolean type |
+
+### Constraints
+
+  * (C1) `lhs` and `rhs` have the same element type.
+  * (C2) `lhs`, `rhs`, and `result` have the same shape.
+  * (C3) Let `E` be the `lhs` element type:
+    * If `E` is complex type, `comparison_direction` $\in$ `{EQ, NE}`,
```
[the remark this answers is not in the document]
subhankarshah: #547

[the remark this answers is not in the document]
subhankarshah: We don't support it yet, I have added a ticket to add support for complex types.

docs/spec_draft.md
subhankarshah: XLA uses TOTALORDER, there are examples of comparison that result in different outputs with FLOAT and TOTALORDER

docs/spec_draft.md
subhankarshah: Thanks

subhankarshah:
@sdasgup3 Should we open a ticket for implementing GT,LT etc for complex types etc, as of now only EQ and NE is supported for complex types.

sdasgup3:
>> Should we open a ticket for implementing GT,LT etc for complex types etc, as of now only EQ and NE is supported for complex types.

Given that they are unimplemented in HLO and Jax does not rely on the support of GT/LT compare primitives  for complex types on HLO, we should not have to support them in StableHLO. Currently, Jax is lowering these ops to tlexicographic float comparison on real and imag parts. Same can be done to target StableHLO.

docs/status.md
```diff
@@ -62,7 +62,7 @@ one of the following tracking labels.
 | cholesky                 | yes           | yes          | yes            | yes             | no          |
 | clamp                    | no            | yes*         | yes*           | yes             | no          |
 | collective_permute       | no            | revisit      | revisit        | no              | no          |
-| compare                  | no            | yes*         | yes*           | yes             | no          |
+| compare                  | yes           | revisit      | yes            | yes             | no          |
```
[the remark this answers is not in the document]
subhankarshah: if we dont want complex support then the verification is correct.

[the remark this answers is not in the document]
subhankarshah: #571
Opened ticket for fixing verification.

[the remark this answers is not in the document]
subhankarshah: Thanks for pointing out.

docs/spec_draft.md
```diff
@@ -703,6 +704,72 @@ matrix, then the behavior is undefined.
...
+| Name     | Type                   |
+|----------|------------------------|
+| `result` | tensor of boolean type |
+
+### Constraints
+
+  * (C1) `lhs` and `rhs` have the same element type.
+  * (C2) `lhs`, `rhs`, and `result` have the same shape.
+  * (C3) Given `E` is the `lhs` element type, the following are legal values of
+         `compare_type`:
```
sdasgup3, line 752: and `comparison_direction`:

docs/spec_draft.md
```diff
@@ -703,6 +704,72 @@ matrix, then the behavior is undefined.
...
+    * If `E` is complex type, `compare_type` = `FLOAT` and
+      `comparison_direction` $\in$ `{EQ, NE}`.
+    * If `E` is signed integer type, `compare_type` = `SIGNED`.
```
sdasgup3, line 755: with all supported values of `comparison_direction`.

here and below

sdasgup3 (approved):
LGTM with a few changes.

docs/spec_draft.md
```diff
@@ -703,6 +704,72 @@ matrix, then the behavior is undefined.
...
+  * `NE`: `lhs` $\ne$ `rhs`.
+  * `GE`: `lhs` $\ge$ `rhs`.
+  * `GT`: `lhs` $\gt$ `rhs`.
+  * `LE`: `lhs` $\le$ `rhs`.
+  * `LT`: `lhs` $\lt$ `rhs`.
+
+The values of `compare_type` have the following semantics:
+  * `FLOAT`: If element-type is floating-type, ordered floating-point
+             comparison of `lhs` and `rhs` is performed as specified in
+             IEEE-754, if element-type is complex-type lexicographic comparison
```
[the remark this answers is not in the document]
subhankarshah: complex comparison is only for EQ and NE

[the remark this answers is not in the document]
subhankarshah: removed lexicographic comparison part and reworded.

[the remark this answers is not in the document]
sdasgup3: The fact that only `EQ` and `NE` is supported for complex type is already covered in constraints. 
Also, we cannot remove the lexicographic part as it provides information about how the comparison should happen, irrespective of what comparison direction we chose. 

The suggested pros covers both the aspects.

sdasgup3 (approved):

docs/spec_draft.md
```diff
@@ -703,6 +704,76 @@ matrix, then the behavior is undefined.
...
+|----------|------------------------|
+| `result` | tensor of boolean type |
+
+### Constraints
+
+  * (C1) `lhs` and `rhs` have the same element type.
+  * (C2) `lhs`, `rhs`, and `result` have the same shape.
+  * (C3) Given `E` is the `lhs` element type, the following are legal values of
+         `compare_type` and `comparison_direction`:
+    * If `E` is complex type, `compare_type` = `FLOAT` and
```
burmako, line 754: Why can't `compare_type` be `TOTALORDER` for complex tensors?

docs/spec_draft.md
```diff
@@ -703,6 +704,76 @@ matrix, then the behavior is undefined.
...
+The values of `compare_type` have the following semantics:
+  * `FLOAT`: If element-type is floating-type, ordered floating-point comparison
+             of `lhs` and `rhs` is performed for all values of
+             `comparison_direction` except for `NE` for which unordered
+             comparison is performed. If element-type is complex-type,
+             lexicographic comparison of (real, imag) pairs is performed.
+  * `TOTALORDER`: `totalOrder` floating-point comparison of `lhs` and `rhs` as
+                  specified in IEEE-754.
+  * `SIGNED`: Signed comparison of `lhs` and `rhs`.
+  * `UNSIGNED`: Unsigned comparison of `lhs` and `rhs`.
```
burmako, line 731: Proposed rewording:

```
The values of `comparison_direction` and `compare_type` have the following 
semantics:

For integer and boolean element types:
  * `EQ`: `lhs` $=$ `rhs`.
  * `NE`: `lhs` $\ne$ `rhs`.
  * `GE`: `lhs` $\ge$ `rhs`.
  * `GT`: `lhs` $\gt$ `rhs`.
  * `LE`: `lhs` $\le$ `rhs`.
  * `LT`: `lhs` $\lt$ `rhs`.

For floating-point element types and `compare_type = FLOAT`, the op implements
the following IEEE-754 operations:
  * `EQ`: `compareQuietEqual`.
  * `NE`: `compareQuietNotEqual`.
  * `GE`: `compareQuietGreaterEqual`.
  * `GT`: `compareQuietGreater`.
  * `LE`: `compareQuietLessEqual`.
  * `LT`: `compareQuietLess`.

For floating-point element types and `compare_type = TOTALORDER`, the op
uses the combination of `totalOrder` and `compareQuietEqual` operations from 
IEEE-754.

For complex element types, lexicographic comparison of `(real, imag)` pairs is 
performed using the provided `comparison_direction` and `compare_type`.
```

Feedback:
  1. Need to go into detail about what `comparison_direction` values mean for floats.
  2. Sounds like we can safely ignore `compare_type` for non-floats.
  3. I think it's better to structure semantics discussion by element type rather than by attribute values.
  4. Not exactly sure how `TOTALORDER` is supposed to work for `EQ`, `NE`, `GE` and `LE`. `totalOrder` from IEEE-754 doesn't have the "or equal" flavor. I made a guess of how it could work, would be good to verify it.

docs/spec_draft.md
burmako: From what I can see, [elemental_ir_emitter.cc](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/elemental_ir_emitter.cc) completely ignores `compare_type`?

burmako (changes requested):

docs/spec_draft.md
```diff
@@ -703,6 +704,76 @@ matrix, then the behavior is undefined.
...
+
+### Inputs
+
+| Name                   | Type                                                    |
+|------------------------|---------------------------------------------------------|
+| `lhs`                  | tensor of any supported type                            |
+| `rhs`                  | tensor of any supported type                            |
+| `comparison_direction` | enum of `EQ`, `NE`, `GE`, `GT`, `LE`, and `LT`          |
+| `compare_type`         | enum of `FLOAT`, `TOTALORDER`, `SIGNED`, and `UNSIGNED` |
```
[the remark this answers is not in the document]
subhankarshah: This code pointer says its deprecated, it is not possible to compare signed integers as unsigned or vice versa.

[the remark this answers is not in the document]
subhankarshah: So we can go with `DEFAULT` and `TOTALORDER`, I think that is a good idea.

docs/spec_draft.md
subhankarshah: I am not sure of the reason, but it cant 😅
The default `FLOAT` is only allowed for complex types

docs/spec_draft.md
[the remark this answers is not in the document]
subhankarshah: If we go with this, then C3 would look something along the lines of
If element type `E` of `lhs` is `FLOAT` compare_type can be specified as `TOTAL_ORDER`.
Or we might even skip C3 if we have outlined it in the semantics.

In future we can consider updating the op definition itself to have just TOTAL_ORDER comparison_types and cover rest by default.

docs/spec_draft.md
burmako: You're right! I missed [this part](https://github.com/tensorflow/tensorflow/blob/82e83ea97011c58945c8aebab969800e0d4aebf1/tensorflow/compiler/xla/service/hlo_verifier.cc#L2418-L2442) of hlo_verifier.cc. Thank you for finding out and documenting this.

burmako (approved):

PR #434, opened 2022-11-02
Add spec for BitcastConvertOp

### LINKED ISSUE #417: Add spec for BitcastConvertOp
### Request description

To add spec for `stablehlo.bitcast_convert`

ghpvnist:
closes #417

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -165,6 +165,7 @@ described below)
    * [and](#stablehloand)
    * [batch_norm_inference](#stablehlobatch_norm_inference)
    * [batch_norm_training](#stablehlobatch_norm_training)
+   * [bitcast_convert](#stablehlobitcast_convert)
    * [broadcast_in_dim](#stablehlobroadcast_in_dim)
    * [case](#stablehlocase)
    * [ceil](#stablehloceil)
@@ -504,6 +505,57 @@ Numeric precision is implementation-defined.
 //          ]
 // %batch_mean: [2.0, 3.0]
 // %batch_var: [1.0, 1.0]
+```
+
+[Back to Ops](#index-of-ops)
+
+## stablehlo.bitcast_convert
+
+### Semantics
+
+Performs an element-wise bitcast operation from `operand` tensor and produces
+a `result` tensor where the bits are reinterpreted with the new element type.
+Bitcast is implemented as a low-level cast, so machines with different
+floating-point representations will give different results (e.g. endianesss).
+Conversions between different bitwidths are not element-wise and creates/deletes
+a dimension if new element type requires less/more bits, respectively.
+
+More formally, let `E` and `E'` be the `operand` and `result` element type,
+respectively:
+  * If `numBits(E) > numBits(E')`,
+    `result[i0, ..., iR-1, :] = operand[j0, ..., jR-1]` where
+    `R = rank(operand)`, and `dim(result, -1) = numBits(E)/numBits(E')`.
+  * If `numBits(E') > numBits(E)`,
+    `result[i0, ..., iR-2] = operand[j0, ..., jR-2, :]` where
+    `R = rank(operand)`, and `dim(operand, -1) = numBits(E')/numBits(E)`.
+
+### Inputs
+
+| Name      | Type                         |
+|-----------|------------------------------|
+| `operand` | tensor of any supported type |
+
+### Outputs
+
+| Name     | Type                         |
+|----------|------------------------------|
+| `result` | tensor of any supported type |
+
+### Constraints
+
+  * (C1) `operand` and `result` have the same shape except for the last
+    dimension.
+  * (C2) Cannot convert between real and complex type.
+
+### Examples
+
+```mlir
+// %operand: [0,0, 1.0]
+%result = "stablehlo.bitcast_convert"(%operand) : (tensor<2xf32>) -> tensor<2x4xi8>
+// %result: [
+//           [0, 0, 0, 0],
+//           [0, 0, -128, 63] // little-endian representation of floating-point 1
+//          ]
 ```
 
 [Back to Ops](#index-of-ops)
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -513,22 +513,23 @@ Numeric precision is implementation-defined.
 
 ### Semantics
 
-Performs an element-wise bitcast operation on `operand` tensor and produces a
-`result` tensor where the bits are reinterpreted with the new element type.
-Conversions between different bitwidths are not element-wise and creates/deletes
-a dimension if new element type requires less/more bits, respectively.
+Performs a bitcast operation on `operand` tensor and produces a `result` tensor
+where the bits of the entire `operand` tensor are reinterpreted using the
+type of the `result` tensor.
 
 Let `E` and `E'` be the `operand` and `result` element type respectively, and
 `R = rank(operand)`:
-  * If `numBits(E)` $=$ `numBits(E')`,
-     `bitRepr(result[i0, ..., iR-1]) = bitRepr(operand[i0, ..., iR-1])`.
-  * If `numBits(E)` $\gt$ `numBits(E')`,
-    `bitRepr(result[i0, ..., iR-1, :]) = bitRepr(operand[i0, ..., iR-1])`.
-  * If `numBits(E)` $\lt$ `numBits(E')`,
-    `bitRepr(result[i0, ..., iR-2]) = bitRepr(operand[i0, ..., iR-2, :])`.
+  * If `num_bits(E')` $=$ `num_bits(E)`,
+    `bits(result[i0, ..., iR-1]) = bits(operand[i0, ..., iR-1])`.
+  * If `num_bits(E')` $\lt$ `num_bits(E)`,
+    `bits(result[i0, ..., iR-1, :]) = bits(operand[i0, ..., iR-1])`.
+  * If `num_bits(E')` $\gt$ `num_bits(E)`,
+    `bits(result[i0, ..., iR-2]) = bits(operand[i0, ..., iR-2, :])`.
 
-Bitcast is implemented as a low-level cast, so machines with different
-floating-point representations (e.g. endianesss) will give different results.
+The behavior of `bits` is implementation-defined because the exact
+representation of tensors is implementation-defined, and the exact
+representation of floating-point types is implementation-defined (e.g. IEEE-754
+allows implementations to deviate on endianness).
 
 ### Inputs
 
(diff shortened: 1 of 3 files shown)


docs/spec_draft.md
```diff
@@ -345,6 +346,57 @@ logical operation.
...
+| Name     | Type                         |
+|----------|------------------------------|
+| `result` | tensor of any supported type |
+
+### Constraints
+
+  * (C1) `operand` and `result` have the same shape except for the last
+    dimension.
+  * (C2) Cannot convert between real and complex type.
+
```
subhankarshah, line 386: max(bit_width(E),bit_width(E')) % min(bit_width(E),bit_width(E')) == 0
bit-widths should be perfectly divisible.

subhankarshah: If numBits(E') > numBits(E) then dim(operand, -1) = numBits(E') / numBits(E)

ghpvnist: > max(numBits(E), numBits(E')) % min(numBits(E), numBits(E')) == 0
> numBits should be perfectly divisible.

Since the current supported types are powers of 2, it will be perfectly divisible by default.

> If numBits(E') > numBits(E) then dim(operand, -1) = numBits(E') / numBits(E)

Thanks for the suggestion. I moved the entire shape inference to the constraints section.

docs/spec_draft.md
```diff
@@ -345,6 +346,56 @@ logical operation.
...
 [Back to Ops](#index-of-ops)
 
+## stablehlo.bitcast_convert
+
+### Semantics
+
+Performs an element-wise bitcast operation from `operand` tensor and produces
+a `result` tensor where the bits are reinterpreted with the new element type.
+Bitcast is implemented as a low-level cast, so machines with different
+floating-point representations will give different results (e.g. endianesss).
```
subhankarshah, line 356: floating-point representation(endianesss)

docs/spec_draft.md
```diff
@@ -345,6 +346,56 @@ logical operation.
...
+
+| Name     | Type                         |
+|----------|------------------------------|
+| `result` | tensor of any supported type |
+
+### Constraints
+
+  * (C1) `operand` and `result` have the same shape except for the last
+    dimension.
+  * (C2) Cannot convert between real and complex type.
```
subhankarshah, line 376: Conversion between real and complex types is not permitted.

docs/spec_draft.md
subhankarshah: I think this belongs to the formal semantics, you can make 3 cases (add one case with equality)

subhankarshah: Let `E` and `E'` be the `operand` and `result` element type and `R = rank(operand)` respectively:
    * If `numBits(E) == numBits(E')`,
       `bitRepr(result[i0, ..., iR-1]) = bitRepr(operand[i0, ..., iR-1])`
        `shape(result) = shape(operand)`
    * If `numBits(E) > numBits(E')`,
      `bitRepr(result[i0, ..., iR-1, :]) = bitRepr(operand[i0, ..., iR-1])`
      `rank(result) = R+1` and `dim(result, -1) = numBits(E)/numBits(E')`
    * If `numBits(E') > numBits(E)`,
      `bitRepr(result[i0, ..., iR-2]) = bitRepr(operand[i0, ..., iR-2, :])`
      `rank(result) = R-1` and `dim(operand, -1) = numBits(E)/numBits(E')`

subhankarshah: If you put this in semantics you can remove C1

subhankarshah: Let me know what you think

docs/spec_draft.md
subhankarshah: Bitcast is implemented as a low-level cast, so machines with different
floating-point representation(e.g. endianesss) will give different results.

This goes after formal semantics

docs/spec_draft.md
```diff
@@ -345,6 +346,56 @@ logical operation.
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.bitcast_convert
+
+### Semantics
+
+Performs an element-wise bitcast operation from `operand` tensor and produces
```
subhankarshah, line 353: operation on

docs/spec_draft.md
subhankarshah: Note I have removed j0 ... and changed them to i0 ... because they are same

subhankarshah (changes requested):
Left some comments.

stablehlo/dialect/StablehloOps.td
```diff
@@ -1644,18 +1644,15 @@ def StableHLO_BitcastConvertOp : StableHLO_ShapedInterfaceOp<"bitcast_convert",
...
 
-    See https://www.tensorflow.org/xla/operation_semantics#bitcastconverttype.
+    See:
+    https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md#stablehlobitcast_convert
 
     Example:
-
     ```mlir
-    %0 = stablehlo.bitcast_convert %arg0 : (tensor<2xi32>) -> tensor<2xf32>
+    %result = stablehlo.bitcast_convert %operand : (tensor<2xf32>) -> tensor<2x4xi8>
```
GleasonK, line 1655: Nice, this is a way better example.

docs/spec_draft.md
ghpvnist: I've updated the semantics and constraints. Please take a look and let me know any suggestions.

stablehlo/dialect/StablehloOps.td
ghpvnist: I think so too!

docs/spec_draft.md
```diff
@@ -345,6 +346,66 @@ logical operation.
...
+## stablehlo.bitcast_convert
+
+### Semantics
+
+Performs an element-wise bitcast operation on `operand` tensor and produces a
+`result` tensor where the bits are reinterpreted with the new element type.
+Conversions between different bitwidths are not element-wise and creates/deletes
+a dimension if new element type requires less/more bits, respectively.
+
+Bitcast is implemented as a low-level cast, so machines with different
```
subhankarshah, line 358: This can be after formal semantics.

docs/spec_draft.md
```diff
@@ -345,6 +346,66 @@ logical operation.
...
+floating-point representations (e.g. endianesss) will give different results.
+
+Let `E` and `E'` be the `operand` and `result` element type respectively, and
+`R = rank(operand)`:
+  * If `numBits(E) == numBits(E')`,
+     `bitRepr(result[i0, ..., iR-1]) = bitRepr(operand[i0, ..., iR-1])`.
+  * If `numBits(E) > numBits(E')`,
+    `bitRepr(result[i0, ..., iR-1, :]) = bitRepr(operand[i0, ..., iR-1])`.
+  * If `numBits(E') > numBits(E)`,
```
subhankarshah, line 367: `numBits(E) < numBits(E')`
reverse the order to keep in consistent with conditions above

subhankarshah (approved):
LGTM, left two minor comments

docs/spec_draft.md
```diff
@@ -430,6 +431,66 @@ Numeric precision is implementation-defined.
...
+  * (C1) `operand` and `result` have the same shape except for the last
+    dimension. Let `E` and `E'` be the `operand` and `result` element type,
+    respectively and `R = rank(operand)`:
+    * If `numBits(E)` $=$ `numBits(E')`, shape(`operand`) $=$ shape(`result`).
+    * If `numBits(E)` $\gt$ `numBits(E')`, dim(`result`, `i`) $=$
+      dim(`operand`, `i`) for all `i` $\in$ [0, `R`-1), and
+      `dim(result, -1) = numBits(E)/numBits(E')`.
+    * If `numBits(E')` $\gt$ `numBits(E)`, dim(`result`, `i`) $=$
+      dim(`operand`, `i`) for all `i` $\in$ [0, `R`-1), and
+      `dim(operand, -1) = numBits(E')/numBits(E)`.
```
burmako, line 478: Proposed rewording:

```
  * (C1) Let `E` and `E'` be the `operand` and `result` element type,
    respectively and `R = rank(operand)`:
    * If `numBits(E')` $=$ `numBits(E)`, shape(`result`) $=$ shape(`operand`).
    * If `numBits(E')` $\lt$ `numBits(E)`:
      * `rank(result) = R+1`.
      * dim(`result`, `i`) $=$ dim(`operand`, `i`) for all `i` $\in$ [0, `R`-1].
      * `dim(result, R) = numBits(E)/numBits(E')`.
    * If `numBits(E')` $\gt$ `numBits(E)`:
      * `rank(result) = R-1`.
      * dim(`result`, `i`) $=$ dim(`operand`, `i`) for all `i` $\in$ [0, `R`-1).
      * `dim(operand, R-1) = numBits(E')/numBits(E)`.
```

Feedback:
  1) Dropped "`operand` and `result` have the same shape except for the last dimension" because sometimes they have exactly the same shape.
  2) Same as in the formalism, reordered formulae so that result is consistently on the lhs and operand is consistently on the rhs.
  3) Introduced another level of bullets to make it individual cases stand out among prose with ifs and foralls.
  4) Used absolute axes everywhere, i.e. `R-1` instead of `-1`, again for consistency reasons. This uncovered a bug in the second bullet.

docs/spec_draft.md
```diff
@@ -430,6 +431,66 @@ Numeric precision is implementation-defined.
...
+  * (C2) Conversion between real and complex types is not permitted.
```
burmako, line 479: "Conversion between complex and non-complex types is not permitted".

burmako (changes requested):

docs/spec_draft.md
```diff
@@ -508,6 +509,68 @@ Numeric precision is implementation-defined.
...
+    `bits(result[i0, ..., iR-1]) = bits(operand[i0, ..., iR-1])`.
+  * If `num_bits(E')` $\lt$ `num_bits(E)`,
+    `bits(result[i0, ..., iR-1, :]) = bits(operand[i0, ..., iR-1])`.
+  * If `num_bits(E')` $\gt$ `num_bits(E)`,
+    `bits(result[i0, ..., iR-2]) = bits(operand[i0, ..., iR-2, :])`.
+
+The behavior of `bits` is implementation-defined because the exact
+representation of tensors is implementation-defined, and the exact
+representation of floating-point types is implementation-defined (e.g. IEEE-754
+allows implementations to deviate on endianness).
```
burmako, line 532: Resurrecting a question from one of the resolved conversations: "Is it fair to call the behavior completely implementation-defined? Or can we guarantee something for integer to integer casts? What about booleans?".

burmako (changes requested):

docs/spec_draft.md
ghpvnist: For conversions itself, we cannot guarantee integer to integer casts either since they are also subject to endianness. The behavior of `bits` meta function is defined for booleans since endianness is over multiple bytes, but boolean is one byte. However, this does not mean that say, float to bool or bool to floats, are defined since the floating representation is dependent on endianness. All casts (except for casts of one type to itself like bool to bool or i32 to i32) are therefore implementation-defined. We can either specify that `bits` is defined for booleans but the rest is implementation defined or just say that everything is implementation-defined.

burmako: Sounds good. Let's just say that everything is implementation-defined.

burmako (approved):

PR #438, opened 2022-11-03
Remove explicit references to kEmitAccessors_Prefixed

tpopp:
This is the default and the option will be removed by https://reviews.llvm.org/D136727


burmako (approved):
Thank you, Tres!

PR #446, opened 2022-11-04
Update status.md to reflect conventional use

burmako:
Updates the wording that explain in details what various labels means, consistent with how we've been using them in the last two months.

### DIFF
--- a/docs/status.md
+++ b/docs/status.md
@@ -17,16 +17,26 @@ particular aspect, as mentioned in the corresponding column, is tracked using
 one of the following tracking labels.
 
  - Generic labels
-    - **yes**: complete
-    - **no**: not complete yet, but part of [the roadmap](https://github.com/openxla/stablehlo#roadmap).
+    - **yes**: there is a comprehensive implementation.
+    - **no**: there is no implementation, but working on that is part of
+      [the roadmap](https://github.com/openxla/stablehlo#roadmap).
+      Note that Verifier can never be labeled as "no" because the ODS already
+      implements some verification.
  - Customized labels for Verifier and Type Inference
-    - **yes**: in sync with [StableHLO semantics](https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md).
-    - **yes\***: in sync with [XLA semantics](https://www.tensorflow.org/xla/operation_semantics).
-    - **revisit**: there is an implementation in between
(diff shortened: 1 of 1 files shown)


docs/status.md
```diff
@@ -17,16 +17,26 @@ particular aspect, as mentioned in the corresponding column, is tracked using
...
-      [the ODS](https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/StablehloOps.td)
-      and [StablehloOps.cpp](https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/StablehloOps.cpp)
-      but it needs to be revisited to determine its status.
-    - **infeasible**: infeasible to implement by design.
+    - **yes**: there is an implementation, and it's in sync with
+      [StableHLO semantics](https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md).
+    - **yes\***: there is an implementation, and it's in sync with
+      [XLA semantics](https://www.tensorflow.org/xla/operation_semantics).
+      Since XLA semantics is oftentimes underdocumented, we are using
+      [hlo_verifier.cc](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/hlo_verifier.cc)
```
sdasgup3, line 31: Thanks for highlighting  that the hlo_vefirier and the shape_inference to superseded the HLO spec in case of conflict. We often run into this.

[no action]

docs/status.md
```diff
@@ -17,16 +17,26 @@ particular aspect, as mentioned in the corresponding column, is tracked using
...
+      and [shape_inference.cc](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/shape_inference.cc)
+      as the reference.
+    - **revisit**: there is an implementation, but it doesn't fall under "yes"
+      or "yes\*" - either because we haven't audited it yet, or because we have
+      and found issues.
```
sdasgup3, line 36: One question to clarify: about ` we have and found issues.`

As we already found issues, there must be a ticket for it. Now we might have two options 

"yes*": At this point the semantics is same as HLO.
"revisit": There might be other issue and need need deeper investigation.

sdasgup3 (approved):

docs/status.md
burmako: "Now we might have two options: "yes*": At this point the semantics is same as HLO...". That cannot be the case by definition because in the earlier part of the sentence we say "there is an implementation, but it doesn't fall under "yes" or "yes\*"".

PR #439, opened 2022-11-03
Update spec checklist to account for prettyprinting conventions

burmako:
This is something that we've been following for some time now, and I figured it'd be good to document explicitly and have a chance to discuss.

PR #453, opened 2022-11-05
(abs..constant) Align examples in spec/ODS with spec_checklist.md

burmako:
In #439, we documented the conventions for examples in spec/ODS, and I went to check how closely the previous work on the spec follows these conventions.

I discovered quite a few divergencies, so I started going through the spec fixing things. This PR covers 10 out of the currently specced 43 ops.

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -241,20 +241,9 @@ defined and one of the following:
 ### Examples
 
 ```mlir
-// integers
 // %operand: [-2, 0, 2]
 %result = "stablehlo.abs"(%operand) : (tensor<3xi32>) -> tensor<3xi32>
 // %result: [2, 0, 2]
-
-// floats
-// %operand: [-2.2, 0.0, 2.2]
-%result = "stablehlo.abs"(%operand) : (tensor<3xf32>) -> tensor<3xf32>
-// %result = [2.2, 0.0, 2.2]
-
-// complex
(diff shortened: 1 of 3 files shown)


docs/spec_draft.md
```diff
@@ -312,9 +301,9 @@ the IEEE-754 specification. For boolean element type, the behavior is same as
 
 ### Semantics
 
-Performs element-wise bitwise AND of two tensors `lhs` and `rhs` of integer
-types and produces a `result` tensor. For boolean tensors, it computes the
-logical operation.
+Performs element-wise bitwise or logical AND of two tensors `lhs` and `rhs` and
```
sdasgup3, line 304: Can we do the same changes for or/xor/not? 

These are one still having multiple examples:
``` 
stablehlo.divide
stablehlo.or
stablehlo.xor
stablehlo.not
stablehlo.exponential
stablehlo.iota
stablehlo.log
stablehlo.logistic
stablehlo.negate
stablehlo.reverse
stablehlo.slice
stablehlo.sort
stablehlo.sqrt
```

sdasgup3 (approved):
LGTM with a question.

docs/spec_draft.md
burmako: In this PR, I wanted to limit myself to just abs..constant, because I didn't have the time to go through all the ops (we've already have plenty of ops :)). Let's track under #454.

PR #456, opened 2022-11-07
Fix the LINK_LIBS of StablehloTypeInference

yaochengji:
Found it in an [onnx-mlir pr](https://github.com/onnx/onnx-mlir/pull/1829)


GleasonK (approved):
Thanks for catching this!

`TypeInference.cpp` uses `stablehlo/dialect/Base.h` and several functions from `Base.h`, this dependency is definitely needed in the cmake file.

I'm guessing this slipped through out build system because `StablehloOps`, which is the only use of `StablehloTypeInference`, also depends on `StablehloBase`.

Issue #444, opened 2022-11-04
Warning spew


silvasean:
### What happened?

```
/usr/local/google/home/silvasean/pg/torch-mlir/torch-mlir/externals/mlir-hlo/stablehlo/stablehlo/dialect/ChloOps.cpp:318:42: warning: extra ';' outside of a function is incompatible with C++98 [-Wc++98-compat-extra-semi]
BROADCAST_BINARY_OP_DEFS(BroadcastZetaOp);
                                         ^
```
And various other similar warnings

### Steps to reproduce your issue

Build with Clang with default warning settings (I think; or add -Wc++98-compat-extra-semi if necessary).

### Version information

mlir-hlo @ 36238f16441cd1a884af988d4400d2ebb0c75bbc

GleasonK:
Curious what your build warning flags are? 

We currently use `-Wall` in CI, and build will pass with `-Wall -Wextra`, but not `-Weverything` largely due to C++98 incompats. I'm surprised we care about C++98 compat flags, given the amount of c++98 incompatible features we use, to show a few:

```
/usr/local/google/home/gleasonk/Coding/llvm-project/llvm/include/llvm/Support/ErrorHandling.h:121:73: error: 'nullptr' is incompatible with C++98 [-Werror,-Wc++98-compat]
llvm_unreachable_internal(const char *msg = nullptr, const char *file = nullptr,
                                                                        ^
...
/usr/local/google/home/gleasonk/Coding/llvm-project/llvm/include/llvm/Support/PointerLikeTypeTraits.h:75:10: error: 'constexpr' specifier is incompatible with C++98 [-Werror,-Wc++98-compat]
  static constexpr int NumLowBitsAvailable = 2;
         ^
```

Happy to include the `-Wc++98-compat-extra-semi` in build, but trying to figure out where to draw the line on the definition of a "warning free build".

silvasean:
Oh, fascinating! This is part of the default LLVM build flags.

LLVM_ENABLE_PEDANTIC (default ON) leads to this piece of code which appears to normalize warning about the extra semicolons across GCC and clang:
https://github.com/llvm/llvm-project/blob/c37f29c49ea237eaf65dd137bffc1f3316f82951/llvm/cmake/modules/HandleLLVMOptions.cmake#L712

Maybe the takeaway here is to build with -pedantic and use -Wc++98-compat-extra-semi to smooth across GCC/Clang differences like LLVM does?

[...]

PR #457, opened 2022-11-07
Build StableHLO as a part of LLVM in CI

GleasonK:
A suggestion made in #444, build StableHLO as a part of LLVM in CI. Many clients do this so we should make sure the build is clean.

Other minor cleanup: Remove matrix, we only run tests on ubuntu currently. No need for a matrix, no need to build for anything except host.


GleasonK:
cc: @silvasean

burmako:
As discussed offline, we're already building StableHLO as a part of LLVM when running CI for Python bindings. Can we make use of that instead?

burmako (changes requested):
See the comment above.

build_tools/github_actions/ci_build_stablehlo_python_api.sh
```diff
@@ -36,7 +36,9 @@ cmake -GNinja \
   -DCMAKE_CXX_COMPILER=clang++ \
   -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
   -DCMAKE_C_COMPILER=clang \
-  -DCMAKE_C_COMPILER_LAUNCHER=ccache
+  -DCMAKE_C_COMPILER_LAUNCHER=ccache \
+  -DLLVM_ENABLE_WARNINGS=ON \
+  -DLLVM_ENABLE_WERROR=ON
```
burmako, line 41: Let's also enable `-DLLVM_ENABLE_PEDANTIC=ON`, following up on #455? Other than that, LGTM.

burmako (approved):

build_tools/github_actions/ci_build_stablehlo_python_api.sh
GleasonK: It should be automatically enabled, as the LLVM default is on:
https://llvm.org/docs/CMake.html

This build should mimic the upstream build more closely with a `WERROR` flag added. Might make more sense to remove the `ENABLE_WARNINGS` flag, since that defaults to `On` as well

burmako: I don't have a strong opinion, but I think I have a slight preference towards explicitly enabling both `LLVM_ENABLE_WARNINGS` and `LLVM_ENABLE_PEDANTIC` here to make the intent clear and to make sure that changes to the defaults of these flags don't silently affect our CI (although the latter is probably quite unlikely, so not a strong argument).

GleasonK: Sounds good, added pedantic.

CMakeLists.txt
```diff
@@ -38,6 +38,7 @@ endif()
 # Options and settings
 #-------------------------------------------------------------------------------
 option(STABLEHLO_ENABLE_BINDINGS_PYTHON "Enables StableHLO Python bindings" OFF)
+option(STABLEHLO_STRICT_LINT "Build StableHLO with strict warnings and warnings as errors" OFF)
```
burmako, line 41: How about `STABLEHLO_ENABLE_STRICT_BUILD`?
  1) To be consistent with another `STABLEHLO_ENABLE_*` option above.
  2) I think `BUILD` is a better fit thant `LINT` because even the description of the option says "lint".

burmako (changes requested):

burmako (approved):

Issue #442, opened 2022-11-04
Consider migrating SelectOp and ClampOp to explicit broadcasting


burmako:
StableHLO doesn't generally support implicit broadcasting, except for the case of SelectOp's `pred` as well as ClampOp's `min` and `max`. At least for SelectOp, both JAX and TensorFlow depend on this behavior, so we cannot simply remove it, but I think that in the future we should consider deprecating it for consistency reasons.

burmako:
Looks like [MapOp](https://github.com/tensorflow/tensorflow/blob/1d34295f45965c7090039773dea44c25ac00ec00/tensorflow/compiler/xla/service/shape_inference.cc#L1172-L1181) also supports a limited form of broadcasting in HLO.

PR #460, opened 2022-11-07
Add spec for ClampOp

### LINKED ISSUE #452: Add spec for ClampOp

subhankarshah:
closes #452

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -789,7 +789,10 @@ matrix, then the behavior is undefined.
 
 Clamps every element of the `operand` tensor between a minimum and maximum
 value and produces a `result` tensor. More formally, `result[i0, ..., iR-1]` =
-`minium(maximum(operand[i0, ..., iR-1], min[0, ..., iR-1]), max[i0, ..., iR-1])`.
+`minimum(maximum(operand[i0, ..., iR-1], min[i0, ..., iR-1]), max[i0, ..., iR-1])`,
+where `minimum` and `maximum` operations correspond to
+[stablehlo.minimum](#stablehlominimum) and
+[stablehlo.maximum](#stablehlomaximum).
 
 ### Inputs
 
@@ -807,7 +810,7 @@ value and produces a `result` tensor. More formally, `result[i0, ..., iR-1]` =
 
 ### Constraints
 
-  * (C1) `min`, `operand`, `max` and `result` have the same type.
+  * (C1) `min`, `operand`, `max`, and `result` have the same type.
 
 ### Examples
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -174,6 +174,7 @@ described below)
    * [case](#stablehlocase)
    * [ceil](#stablehloceil)
    * [cholesky](#stablehlocholesky)
+   * [clamp](#stablehloclamp)
    * [complex](#stablehlocomplex)
    * [concatenate](#stablehloconcatenate)
    * [constant](#stablehloconstant)
@@ -778,6 +779,44 @@ matrix, then the behavior is undefined.
 //           [2.0, 4.0, 0.0],
 //           [3.0, 5.0, 6.0]
 //          ]
+```
+
+[Back to Ops](#index-of-ops)
+
+## stablehlo.clamp
+
+### Semantics
+
+Clamps every element of the `operand` tensor between a minimum and maximum
+value and produces a `result` tensor. More formally, `result[i0, ..., iR-1]` =
+`minium(maximum(operand[i0, ..., iR-1], min[0, ..., iR-1]), max[i0, ..., iR-1])`.
+
+### Inputs
+
+| Name      | Type                         |
+|-----------|------------------------------|
+| `min`     | tensor of any supported type |
+| `operand` | tensor of any supported type |
+| `max`     | tensor of any supported type |
+
+### Outputs
+
+| Name     | Type                         |
+|----------|------------------------------|
+| `result` | tensor of any supported type |
+
+### Constraints
+
+  * (C1) `min`, `operand`, `max` and `result` have the same type.
+
+### Examples
+
+```mlir
+// %min: [5, 10, 15]
+// %operand: [3, 13, 23]
+// %max: [10, 15, 20]
+%result = "stablehlo.clamp"(%min, %operand, %max) : (tensor<3xi32>, tensor<3xi32>, tensor<3xi32>) -> tensor<3xi32>
+// %result: [5, 13, 20]
 ```
 
 [Back to Ops](#index-of-ops)
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -789,9 +789,10 @@ matrix, then the behavior is undefined.
 
 Clamps every element of the `operand` tensor between a minimum and maximum
 value and produces a `result` tensor. More formally, `result[i0, ..., iR-1]` =
-`minimum(maximum(operand[i0, ..., iR-1], min[i0, ..., iR-1]), max[i0, ..., iR-1])`,
-where `minimum` and `maximum` operations correspond to
-[stablehlo.minimum](#stablehlominimum) and
+`minimum(maximum(operand[i0, ..., iR-1], min_val), max_val)`,
+where `min_val = rank(min) == 0 ? min : min[i0, ..., iR-1]`,
+`max_val = rank(max) == 0 ? max : max[i0, ..., iR-1]`, `minimum` and `maximum`
+operations correspond to [stablehlo.minimum](#stablehlominimum) and
 [stablehlo.maximum](#stablehlomaximum).
 
 ### Inputs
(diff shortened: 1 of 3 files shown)


burmako:
Let's keep reviewers limited to just Gunhyun for now, so that it's clear whose responsibility it is to review the PR (similarly to how having a single assignee makes it clear whose responsibility it is to move things forward). When the PR is ready for my review, please add me back as a reviewer (and assign the PR to me, as usual).

ghpvnist (changes requested):

ghpvnist (approved):

docs/spec_draft.md
```diff
@@ -627,6 +628,47 @@ matrix, then the behavior is undefined.
...
+
+### Outputs
+
+| Name     | Type                         |
+|----------|------------------------------|
+| `result` | tensor of any supported type |
+
+### Constraints
+
+  * (C1) `min`, `operand`, `max`, and `result` have the same type.
```
burmako, line 658: Just like `select`, this op currently supports a restricted form of broadcasting. Let's investigate if it's essential, and if it is, let's support it in the spec, similarly to `select`.

docs/spec_draft.md
```diff
@@ -627,6 +628,47 @@ matrix, then the behavior is undefined.
...
+## stablehlo.clamp
+
+### Semantics
+
+Clamps every element of the `operand` tensor between a minimum and maximum
+value and produces a `result` tensor. More formally, `result[i0, ..., iR-1]` =
+`minimum(maximum(operand[i0, ..., iR-1], min[i0, ..., iR-1]), max[i0, ..., iR-1])`,
+where `minimum` and `maximum` operations correspond to
+[stablehlo.minimum](#stablehlominimum) and
+[stablehlo.maximum](#stablehlomaximum).
```
burmako, line 640: Do we need `clamp` in the StableHLO opset (as compared with e.g. CHLO) if it's just a composition of `minimum` and `maximum`? Or there are some floating-point shenanigans involved, and it isn't?

docs/status.md
```diff
@@ -60,7 +60,7 @@ one of the following tracking labels.
 | cbrt                     | no            | yes*         | yes*           | yes             | no          |
 | ceil                     | yes           | yes          | yes            | yes             | yes         |
 | cholesky                 | yes           | yes          | yes            | yes             | no          |
-| clamp                    | no            | yes*         | yes*           | yes             | no          |
+| clamp                    | yes           | yes          | yes            | yes             | no          |
```
burmako, line 63: Verification should be "revisit" because [kClamp supports mixed precision](https://github.com/tensorflow/tensorflow/blob/20c6943d3cd7e07da162f7778a0af5d3776274b4/tensorflow/compiler/xla/service/shape_inference.cc#L3301-L3306). We'll need to sort this out as part of #369.

burmako (changes requested):

docs/spec_draft.md
subhankarshah: Were you referring to issues like mixed precision when talking about floating-point shenanigans?

docs/status.md
subhankarshah: Sure

docs/spec_draft.md
subhankarshah: #442

subhankarshah: TF uses scalar max and min (the example is form XLA operation semantics)
```
let operand: s32[3] = {-1, 5, 9};
let min: s32 = 0;
let max: s32 = 6;

Clamp(min, operand, max) = s32[3]{0, 5, 6};
```

subhankarshah: Jax also supports boradcasting max and min

subhankarshah: [Jax code pointer](https://github.com/google/jax/blob/0da02dd41c3b4bb5fb685f1a6d421db2a324c2d5/jax/_src/lax/lax.py#L2997)

burmako: Thank you for looking into this! Then we'll need to update the formalism and the constraints, similarly to `select`, to account for this.

docs/spec_draft.md
burmako: Sorry - I should've been more precise. I meant something different from mixed precision.

Let's take another example first. One might think that `stablehlo.minimum` can be replaced by a combination of `stablehlo.compare` and `stablehlo.select`. However, as illustrated by [elemental_ir_emitter.cc](https://github.com/tensorflow/tensorflow/blob/49959d2f7d5e497edf90d6cdebc9afa3247234b7/tensorflow/compiler/xla/service/llvm_ir/llvm_util.cc#L108-L120), it's not that easy:
  * First, because of the floating-point shenanigans that I was referring to, we need not two but four ops to represent minimum in full generality because we need to properly account for NaNs.
  * Furthermore, depending on compiler flags, sometimes we don't want to properly account for NaNs and want two ops instead.

As a result, I don't think it would be a good idea to not have `minimum` (or `maximum`) ops in StableHLO, because without these dedicated ops, we won't be able to reliably support a wide range of producers (there are producers who care about fastmath, and there are those who don't).

Now, what about `stablehlo.clamp`? From what I can see, [elemental_ir_emitter.cc](https://github.com/tensorflow/tensorflow/blob/49959d2f7d5e497edf90d6cdebc9afa3247234b7/tensorflow/compiler/xla/service/elemental_ir_emitter.cc#L1906-L1927) seems to suggest that it's completely equivalent to a combination of `minimum` and `maximum`. If that's indeed the case, do we need `clamp` as a dedicated op in the StableHLO opset? Or maybe it should be moved to CHLO?

burmako (changes requested):

docs/spec_draft.md
subhankarshah: 1. Minimum/maximum do not support complex datatypes(but that is a different topic of discussing altogether) and implicit broadcasting, but clamp does support implicit broadcasting. 
2. We can still lower it to other ops in stablehlo (broadcast, min and max) and decide to move it to chlo, in this regard what should be our guiding principles - If an op can be completely and correctly lowered to other ops in stablehlo we are free to move it to chlo or are there other nitty-gritties to consider?

docs/spec_draft.md
subhankarshah: Sure, done.

docs/spec_draft.md
sdasgup3: I think we should have the a clamp operation in StableHLO.

From the [above link](https://github.com/tensorflow/tensorflow/blob/49959d2f7d5e497edf90d6cdebc9afa3247234b7/tensorflow/compiler/xla/service/elemental_ir_emitter.cc#L1906-L1927) , the clamp on integral types seems straightforward, but for the floating point type,  the lowering is based on  `EmitFloatMin` and `EmitFloatMax` which poses the same problem, as you mentioned above, preventing us from getting rid of of max/min.

```
 if (primitive_util::IsFloatingPointType(prim_type)) {
    return EmitFloatMin(max_value, EmitFloatMax(min_value, arg_value, ""), "");
```

docs/spec_draft.md
```diff
@@ -627,6 +628,51 @@ matrix, then the behavior is undefined.
...
 
+## stablehlo.clamp
+
+### Semantics
+
+Clamps every element of the `operand` tensor between a minimum and maximum
+value and produces a `result` tensor. More formally, `result[i0, ..., iR-1]` =
+`minimum(maximum(operand[i0, ..., iR-1], min_val), max_val)`,
+where `min_val = rank(min) == 0 ? min : min[i0, ..., iR-1]`,
+`max_val = rank(max) == 0 ? max : max[i0, ..., iR-1]`, `minimum` and `maximum`
```
burmako, line 639: I like the `_val` convention. Let's change the spec for `select` to also use it, i.e. to say `pred_val` instead of `predicate`.

docs/spec_draft.md
burmako: 1) Good catch! Given that we are considering removing support for complex tensors from minimum/maximum, we should remove it from clamp as well using the same rationale. Let's update #560 accordingly? cc @sdasgup3 to confirm.

2) "in this regard what should be our guiding principles - If an op can be completely and correctly lowered to other ops in stablehlo we are free to move it to chlo". Mostly, yes, except that I would replace "are free to" with "it's a good idea", because we need an RFC for changes like that.

3) @sdasgup3 I'm not sure I follow. Sounds like your example actually supports the theory that `clamp` is trivially decomposable to `minimum` and `maximum`?

burmako (approved):
LGTM with a minor comment. Although let's conclude the ongoing discussion about how essential the `clamp` op is for the opset before merging this PR.

docs/spec_draft.md
sdasgup3: 1. Will get back asap.

3. Ah! I got this wrong.  Yes, Stablehlo Max/Min is already doing all the heavy lifting in dealing with all the "floating-point shenanigans". With that the clamp implementation, based on max/min, seems trivially decomposable. 

From the guidelines you mentioned above  it seems like an ideal candidate for that.

sdasgup3: relevant discussion https://github.com/openxla/stablehlo/pull/495#discussion_r1028776747 on how should be approach the transition of an op to CHLO.

docs/spec_draft.md
subhankarshah: Sure, thanks.

docs/spec_draft.md
burmako: 1) Left a comment on #560 to not forget about complex types in `clamp`.
2-3) Opened #582 to continue the discussion of potentially moving `clamp` to CHLO.

burmako (approved):

PR #461, opened 2022-11-07
add spec for after-all

### LINKED ISSUE #443: Add spec for AfterAllOp
### Request description

To add spec for `stablehlo.after_all`

sdasgup3:
fixes #443 

The shape inference is not implemented : Hence `no` in its status.


docs/spec_draft.md
```diff
@@ -38,8 +41,8 @@ immutable n-dimensional arrays and are referred to in the document as
   `si64` and are non-negative (dimension sizes equal to zero are allowed,
   and their meaning is described below). Ranks equal to zero are also allowed,
   and their meaning is also described below.
-  * **Element type** `E` is any one of the supported element types mentioned
-  above.
+  * **Element type** `E` is any one of the supported element types, except
+  the token type, mentioned above.
```
burmako, line 45: Why is token type defined in the section dedicated to element types, when it's not an element type? Let's define it e.g. right above the "Function types" section? Then this change can be reverted.

docs/spec_draft.md
```diff
@@ -281,16 +285,16 @@ the IEEE-754 specification. For boolean element type, the behavior is same as
...
 ### Inputs
 
-| Name  | Type                         |
-|-------|------------------------------|
-| `lhs` | tensor of any supported type |
-| `rhs` | tensor of any supported type |
+| Name  | Type                                                        |
+|-------|-------------------------------------------------------------|
+| `lhs` | tensor of integer, boolean, floating-point, or complex type |
+| `rhs` | tensor of integer, boolean, floating-point, or complex type |
```
burmako, line 291: Given that above we're saying that token types cannot be element types, we don't need this change (and a couple dozen of similar changes below). This is similar to how we know that "any supported type" in "tensor of any supported type" cannot refer to function types.

docs/spec_draft.md
```diff
@@ -305,7 +309,32 @@ the IEEE-754 specification. For boolean element type, the behavior is same as
 // %result: [[6, 8], [10, 12]]
 ```
 
-&nbsp;[More Examples](../stablehlo/tests/interpret_add.mlir)
+# stablehlo.after_all
+
+### Semantics
```
burmako, line 314: Typo: no newline between the section header and the section body.

docs/spec_draft.md
```diff
@@ -305,7 +309,32 @@ the IEEE-754 specification. For boolean element type, the behavior is same as
...
+| `inputs` | variadic number of `token` |
+
+### Outputs
+
+| Name     | Type  |
+|----------|-------|
+| `result` | token |
+
+### Constraints
+  * size(`inputs`) $\ge$ 0.
```
burmako, line 331: How would a size of inputs be less than zero? I think we can drop the entire Constraints section.

docs/spec_draft.md
```diff
@@ -26,6 +26,9 @@ Following are the supported element types in StableHLO:
     (represents a pair of `f64`). Exact representation of complex types
     (e.g. whether the real part or the imaginary part comes first in memory)
     is implementation-defined.
+  * **Token Types** Values of this type are used for ordering side-effecting
+    operations using data dependencies which are robust against compiler
+    optimizations.
```
burmako, line 31: Proposed rewording:

```
  * **Token type** Values of this type are used for imposing order on execution
    of side-effecting operations using data dependencies.
```

Feedback:
  1) How about "Token type", i.e. singular form, since there's only one token type? Same for "Boolean type" above.
  2) The conventional capitalization is "Token type", i.e. capital "Token" and lowercase "type".
  3) Arguably, the operations are already ordered when they are put together in
  functions. What is ordered is execution of these operations.
  4) Let's drop "which are robust against compiler optimizations". Not all transformations are optimizations. And use cases of tokens aren't limited to just transformations (e.g. there are also analyses).

docs/spec_draft.md
```diff
@@ -305,7 +309,32 @@ the IEEE-754 specification. For boolean element type, the behavior is same as
 // %result: [[6, 8], [10, 12]]
 ```
 
-&nbsp;[More Examples](../stablehlo/tests/interpret_add.mlir)
+# stablehlo.after_all
+
+### Semantics
+Ensures that the operations producing the `inputs` are ordered before any
+operation that depend on `result`.
```
burmako, line 316: Proposed rewording:

```
Ensures that the operations producing the `inputs` are executed before any
operations that depend on `result`. Execution of this operation does nothing,
it only exists to establish data dependencies from `result` to `inputs`.
```

Feedback:
  1) "are ordered" => "are executed" given the discussion above on the definition of the token type.
  2) "any operation that depend" doesn't compute grammatically, so I proposed one possible fix.
  3) Explicitly calling out that execution of this operation does nothing.

burmako (changes requested):

sdasgup3:
As an additional change, added `token` as supported types of If/Case/Whlie Ops.

stablehlo/dialect/StablehloOps.td
```diff
@@ -1117,17 +1117,16 @@ def StableHLO_AfterAllOp : StableHLO_Op<"after_all", [Pure]> {
   let summary = "AfterAll operator";
 
   let description = [{
-    AfterAll takes a variadic number of tokens and produces a single token.
-    Tokens are primitive types which can be threaded between side-effecting
-    operations to enforce ordering. AfterAll can be used as a join of tokens
-    for ordering a operation after a set operations.
+    Ensures that the operations producing the `inputs` are ordered before any
+    operation that depend on `result`.
```
burmako, line 1121: Looks like you might've forgotten to update the description?

docs/spec_draft.md
```diff
@@ -585,9 +617,9 @@ returned.
 
 ### Outputs
 
-| Name      | Type                                             |
-|-----------|--------------------------------------------------|
-| `results` | variadic number of tensors of any supported type |
+| Name      | Type                                                      |
+|-----------|-----------------------------------------------------------|
+| `results` | variadic number of tensors of token or any supported type |
```
burmako, line 622: "variadic number of tensors of any supported type or tokens"? `HLO_TensorOrToken` means `AnyTypeOf<[HLO_Tensor, HLO_Token]>`.

burmako (changes requested):

stablehlo/dialect/StablehloOps.td
sdasgup3: Yup. fixed!

docs/spec_draft.md
```diff
@@ -304,6 +308,34 @@ the IEEE-754 specification. For boolean element type, the behavior is same as
...
+
+| Name     | Type                       |
+|----------|----------------------------|
+| `inputs` | variadic number of `token` |
+
+### Outputs
+
+| Name     | Type  |
+|----------|-------|
+| `result` | token |
```
burmako, line 329: You probably meant to wrap "token" in backticks?

stablehlo/dialect/StablehloOps.td
```diff
@@ -1117,17 +1117,18 @@ def StableHLO_AfterAllOp : StableHLO_Op<"after_all", [Pure]> {
...
 
   let description = [{
-    AfterAll takes a variadic number of tokens and produces a single token.
-    Tokens are primitive types which can be threaded between side-effecting
-    operations to enforce ordering. AfterAll can be used as a join of tokens
-    for ordering a operation after a set operations.
+    Ensures that the operations producing the `inputs` are executed before any
+    operations that depend on `result`. Execution of this operation does
+    nothing, it only exists to establish data dependencies from `result` to
+    `inputs`.
```
burmako, line 1123: Let's use just one sentence to follow the checklist?

burmako (approved):
LGTM with two minor comments.

burmako (approved):

PR #482, opened 2022-11-08
Add spec for ReducePrecisionOp

### LINKED ISSUE #451: Add spec for ReducePrecisionOp

ghpvnist:
closes #451

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -3534,11 +3534,9 @@ More formally, `results[:][j0, ..., jR-1] = reduce(input_slices)` where:
 ### Semantics
 
 Reduces floating-point precision element-wise by modeling the effect of
-converting floating-point value to a lower precision format using
-`exponent_bits` and `mantissa_bits` and back to the original format on
-`operand` tensor and produces a `result` tensor. `exponent_bits` and
-`mantissa_bits` can be specified arbitrarily, but not all bit sizes may be
-supported on all hardware implementations.
+converting a floating-point value to a lower precision format using
+`exponent_bits` and `mantissa_bits` and back to the original precision format of
+`operand` tensor and produces a `result` tensor.
 
 ### Inputs
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -431,6 +431,7 @@ syntax.
    * [real](#stablehloreal)
    * [recv](#stablehlorecv)
    * [reduce](#stablehloreduce)
+   * [reduce_precision](#stablehloreduce_precision)
    * [reduce_scatter](#stablehloreduce_scatter)
    * [reduce_window](#stablehloreduce_window)
    * [remainder](#stablehloremainder)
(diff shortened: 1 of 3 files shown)


docs/spec_draft.md
```diff
@@ -1779,6 +1780,50 @@ More formally, `results[:][result_index] = reduce(input_slices)` where:
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.reduce_precision
+
+### Semantics
+
+Reduces floating-point precision element-wise by modeling the effect of
+converting floating-point value to a lower precision format using
+`exponent_bits` and `mantissa_bits` and back to the original format on
```
sdasgup3, line 1789: on -> of

docs/spec_draft.md
```diff
@@ -1779,6 +1780,50 @@ More formally, `results[:][result_index] = reduce(input_slices)` where:
...
+`operand` tensor and produces a `result` tensor. `exponent_bits` and
+`mantissa_bits` can be specified arbitrarily, but not all bit sizes may be
```
sdasgup3, line 1791: IMO this line is meant to be provided just for the XLA compiler and  the spec should not care about  the support status of the implementation. One problem would be the spec will become dependent on the evolution of the implementation. 

Note that this is different from implementation defined which says diff implementation can take different stance.

docs/spec_draft.md
```diff
@@ -1779,6 +1780,48 @@ More formally, `results[:][result_index] = reduce(input_slices)` where:
...
 
+## stablehlo.reduce_precision
+
+### Semantics
+
+Reduces floating-point precision element-wise by modeling the effect of
+converting a floating-point value to a lower precision format using
+`exponent_bits` and `mantissa_bits` and back to the original precision format of
+`operand` tensor and produces a `result` tensor.
+
```
sdasgup3, line 1791: The text from the operational semantics 
`The number of exponent or mantissa bits may exceed the corresponding value for type T; the corresponding portion of the conversion is then simply a no-op.` seems relevant and should be mentioned.

For example, the checks [check:`dest_mantissa_bits < src_mantissa_bits`](https://github.com/tensorflow/tensorflow/blob/179de2937afc79bb1b0dad7c35927ab17780e1c0/tensorflow/compiler/xla/hlo/evaluator/hlo_evaluator_typed_visitor.h#L2116) and [check: `dest_exponent_bits < src_exponent_bits`](https://github.com/tensorflow/tensorflow/blob/179de2937afc79bb1b0dad7c35927ab17780e1c0/tensorflow/compiler/xla/hlo/evaluator/hlo_evaluator_typed_visitor.h#L2138) covered specific cases. When the number of exponent or mantissa bits exceeds the corresponding values of `operand`, the corresponding portion of checks are skipped.

docs/spec_draft.md
```diff
@@ -1779,6 +1780,48 @@ More formally, `results[:][result_index] = reduce(input_slices)` where:
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.reduce_precision
+
+### Semantics
```
sdasgup3, line 1785: As we do not have a precedent in IEE spec, we need to define all the aspects of the op.  

```
 The input values are rounded to the nearest value representable with the given number of mantissa bits ,using "ties to even" semantics and any values that exceed the range specified by the number of exponent bits are clamped to positive or negative infinity 
```

docs/spec_draft.md
ghpvnist: Yes, I agree that adding additional constraints could solve this. For now, I've updated the semantics to mention that it's a no-op in that case.

docs/spec_draft.md
ghpvnist: Should NaNs be mentioned?

sdasgup3: From the [computations](https://github.com/tensorflow/tensorflow/blob/83ccd51d51f47251507405f79f6fe00874d0be2f/tensorflow/compiler/xla/hlo/evaluator/hlo_evaluator_typed_visitor.h#L2096) of the HLO evaluator it seems that we will return back NaN. 

Having said that if we can capture these computations in the spec then we do not have to mention it. Otherwise, we should mention the case of NaN.

docs/spec_draft.md
burmako: Proposed reformulation:

```
Performs element-wise conversion of `operand` to another floating-point type
that uses `exponent_bits` and `mantissa_bits` and back to the original
floating-point type and produces a `result` tensor.

More formally:

  * The mantissa bits of the original value are updated to round the original 
    value to the nearest value representable with `mantissa_bits` using 
    `roundToIntegralTiesToEven` semantics.
  * Then, if `mantissa_bits` are smaller than the number of mantissa bits of
    the original value, the mantissa bits are truncated to `mantissa_bits`.
  * Then, if the exponent bits of the intermediate result don't fit into the 
    range provided by `exponent_bits`, the intermediate result overflows to 
    infinity using the original sign or underflows to zero using the 
    original sign.
```

Feedback:
  1. Replaced "model" with "convert". Maybe it's just me, but the word "model" always looked weird to me, as if we're doing something that's not real.
  2. Dropped the "reduce precision" part because we won't be reducing precision if exponent and mantissa bits are >= the number of exponent and mantissa bits of the original element type.
  3. Formalized the semantics into a sequence of steps using the link provided by Sandeep.
  4. I don't think we need to add other constraints. It looks like semantics can be described just fine without them.

docs/spec_draft.md
burmako: I don't think we have to explicitly call out NaNs. They seem to follow the proposed formalism below.

burmako (changes requested):

burmako (approved):

Issue #492, opened 2022-11-09
Tighten verification check for ComplexOp


ghpvnist:
### What happened?

There is no verification check that `lhs` and `rhs` need to have the same floating-point element type.

### Steps to reproduce your issue

This StableHLO program passes stablehlo-opt:
```
module @main {
  func.func @main() -> tensor<complex<f32>> {
    %0 = stablehlo.constant dense<1.0> : tensor<f32>
    %1 = stablehlo.constant dense<2.0> : tensor<f64>
    %2 = "stblehlo.complex"(%0, %1) : (tensor<f32>, tensor<f64>) -> tensor<complex<f32>>
    return %2 : tensor<complex<f32>>
  }
}
```

burmako:
```
$ build/bin/stablehlo-opt ~/example.mlir
~/example.mlir:5:10: error: 'stablehlo.complex' op requires the same element type for all operands
    %2 = "stablehlo.complex"(%0, %1) : (tensor<f32>, tensor<f64>) -> tensor<complex<f32>>
         ^
```

ComplexOp's has `SameOperandsElementType` in its trait list, so this is already implemented: https://github.com/openxla/stablehlo/blob/283702a1378d4319b0ac551c3b4502a8e22bc386/stablehlo/dialect/StablehloOps.td#L717-L718.

PR #493, opened 2022-11-09
Add spec for ComplexOp

### LINKED ISSUE #468: Add spec for ComplexOp

ghpvnist:
closes #468

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -632,10 +632,9 @@ imaginary values, `lhs` and `rhs`, and produces a `result` tensor.
 
 ### Constraints
 
-  * (C1) `lhs` and `rhs` have the same floating-point element type.
-  * (C2) shape(`lhs`) $=$ shape(`rhs`) $=$ shape(`result`).
-  * (C3) element_type(`lhs`) $=$ underlying floating-point element type in
-    element_type(`result`).
+  * (C1) `lhs` and `rhs` have the same type.
+  * (C2) shape(`result`) $=$ shape(`lhs`).
+  * (C3) element_type(`result`) = complex_type(element_type(`lhs`)).
 
 ### Examples
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -168,6 +168,7 @@ described below)
    * [case](#stablehlocase)
    * [ceil](#stablehloceil)
    * [cholesky](#stablehlocholesky)
+   * [complex](#stablehlocomplex)
    * [concatenate](#stablehloconcatenate)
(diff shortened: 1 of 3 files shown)


ghpvnist:
Verification is `revisit` since the verifier currently allows `lhs` and `rhs` to have different floating-point element type. Issue raised in #492

docs/spec_draft.md
```diff
@@ -609,6 +610,44 @@ matrix, then the behavior is undefined.
...
+
+### Outputs
+
+| Name     | Type                   |
+|----------|------------------------|
+| `result` | tensor of complex type |
+
+### Constraints
+
+  * (C1) `lhs` and `rhs` have the same floating-point element type.
```
sdasgup3, line 635: How about element_type(`lhs`) $=$ element_type(`rhs`). As we already mentioned that the element type is floating point in the table.

sdasgup3 (approved):

docs/status.md
```diff
@@ -63,7 +63,7 @@ one of the following tracking labels.
 | clamp                    | no            | yes*         | yes*           | yes             | no          |
 | collective_permute       | no            | revisit      | revisit        | no              | no          |
 | compare                  | no            | yes*         | yes*           | yes             | no          |
-| complex                  | no            | yes*         | yes*           | yes             | no          |
+| complex                  | yes           | revisit      | yes            | yes             | no          |
```
burmako, line 66: Given the discussion in #492, I think that Verification should be "yes".

burmako (changes requested):

docs/spec_draft.md
```diff
@@ -609,6 +610,44 @@ matrix, then the behavior is undefined.
...
+  * (C2) shape(`lhs`) $=$ shape(`rhs`) $=$ shape(`result`).
+  * (C3) element_type(`lhs`) $=$ underlying floating-point element type in
+    element_type(`result`).
```
burmako, line 638: Also, not sure if it's just me, but it feels more natural to express C2 and C3 `with element_type(result)` on the left-hand side of the equation, i.e.:

```
  * (C2) `shape(result) = shape(lhs)`.
  * (C3) `element_type(result) = complex_type(element_type(lhs))`.
```

This way, it's also easier to implement type inference for this op based on the Constraints section. It's not hard to invert `element_type(lhs) = real_type(element_type(result))` - but it feels awkward. Given how many times I wrote "feels", this is not a blocking comment, and we can leave figuring this out to #289.

burmako (changes requested):

burmako (approved):

PR #494, opened 2022-11-09
Add spec for SignOp

### LINKED ISSUE #480: Add spec for SignOp

ghpvnist:
closes #480

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -221,6 +221,7 @@ syntax.
    * [shift_left](#stablehloshift_left)
    * [shift_right_arithmetic](#stablehloshift_right_arithmetic)
    * [shift_right_logical](#stablehloshift_right_logical)
+   * [sign](#stablehlosign)
    * [sine](#stablehlosine)
    * [slice](#stablehloslice)
    * [sort](#stablehlosort)
@@ -3121,6 +3122,46 @@ number of bits and produces a `result` tensor.
 // %rhs: [1, 2, 3, 2, 1, 3]
 %result = "stablehlo.shift_right_logical"(%lhs, %rhs): (tensor<6xi8>, tensor<6xi8>) -> tensor<6xi8>
 // %result: [127, 32, 27, 1, 1, 0]
+```
+
+[Back to Ops](#index-of-ops)
+
+## stablehlo.sign
+
+### Semantics
+
+Returns the sign of the `operand` element-wise and produces a `result` tensor.
+For each element `e`:
+```
+sign(e) = -1  if e < 0
+        = -0  if e = -0
+        = NaN if e = NaN
+        = +0  if e = +0
+        = 1   if e > 0
+```
+
+### Inputs
+
+| Name      | Type                                                      |
+|-----------|-----------------------------------------------------------|
+| `operand` | tensor of signed integer, floating-point, or complex type |
+
+### Outputs
+
+| Name     | Type                                                      |
+|----------|-----------------------------------------------------------|
+| `result` | tensor of signed integer, floating-point, or complex type |
+
+### Constraints
+
+  * (C1) `operand` and `result` have the same type.
+
+### Examples
+
+```mlir
+// %operand: [-10.0, -0.0, 0x7FFFFFFF, 0.0, 10.0]
+%result = "stablehlo.sign"(%operand) : (tensor<5xf32>) -> tensor<5xf32>
+// %result: [-1.0, -0.0, 0x7FFFFFFF, 0.0, 1.0]
 ```
 
 [Back to Ops](#index-of-ops)
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -3130,7 +3130,8 @@ number of bits and produces a `result` tensor.
 
 ### Semantics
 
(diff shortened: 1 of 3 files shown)


docs/spec_draft.md
```diff
@@ -1996,6 +1997,46 @@ where `predicate = rank(pred) == 0 ? pred : pred[i0, ..., iR-1]`.
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.sign
+
+### Semantics
+
+Returns the sign of the `operand` element-wise and produces a `result` tensor.
+For each element `e`:
```
subhankarshah, line 2005: can me make it `x`, e would appear as Euler's number in exponential op

subhankarshah: Also if we can refer something like IEEE-754 then we can remove this

subhankarshah: Or if we can deem this to be standard common knowledge / implementation everywhere (this I am not sure of)

subhankarshah (approved):
LGTM with one minor comment.

docs/spec_draft.md
ghpvnist: > can me make it x, e would appear as Euler's number in exponential op

Sure, will do.

> Also if we can refer something like IEEE-754 then we can remove this

No such operation exist in IEEE-754.

> Or if we can deem this to be standard common knowledge / implementation everywhere (this I am not sure of)

I'll keep this thread open for the next reviewer.

stablehlo/dialect/StablehloOps.td
```diff
@@ -570,23 +570,15 @@ def StableHLO_SignOp: StableHLO_UnaryElementwiseOp<"sign",
...
-            = +0  : x = +0
-            = 1   : x > 0
-    ```
+    Returns the sign of the `operand` element-wise and produces a `result`
+    tensor.
 
-    See
-    https://www.tensorflow.org/xla/operation_semantics#element-wise_unary_functions.
+    See:
+    https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md#stablehlorsign
```
burmako, line 577: Typo: "#stablehlosign".

burmako (changes requested):

docs/spec_draft.md
```diff
@@ -1996,6 +1997,46 @@ where `predicate = rank(pred) == 0 ? pred : pred[i0, ..., iR-1]`.
...
+
+Returns the sign of the `operand` element-wise and produces a `result` tensor.
+For each element `x`:
+```
+sign(x) = -1  if x < 0
+        = -0  if x = -0
+        = NaN if x = NaN
+        = +0  if x = +0
+        = 1   if x > 0
+```
```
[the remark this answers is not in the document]
sdasgup3: I agree on keeping the complex support for this op.

docs/spec_draft.md
```diff
@@ -2313,6 +2314,67 @@ where `predicate = rank(pred) == 0 ? pred : pred[i0, ..., iR-1]`.
...
+    if stablehlo.compare(lhs=x, rhs=0.0, comparison_direction=EQ, compare_type=FLOAT):
+      return 0.0
+    if stablehlo.compare(lhs=x, rhs=0.0, comparison_direction=GT, compare_type=FLOAT):
+      return 1.0
+else if element_type(x) is complex:
+  if x.real is NaN or x.imag is NaN:
+    return NaN
+  else:
+    return x / stablehlo.abs(x)
+```
```
burmako, line 2349: Proposed rewording:

```
Returns the sign of the `operand` element-wise and produces a `result` tensor.
More formally, for each element `x`, the semantics can be expressed using 
Python-like syntax as follows:
```

```Python
def sign(x):
  if is_integer(x):
    if compare(x, 0, LT, SIGNED): return -1
    if compare(x, 0, EQ, SIGNED): return 0
    if compare(x, 0, GT, SIGNED): return 1
  elif is_float(x):
    if x is NaN:
      return NaN
    else:
      if compare(x, 0.0, LT, FLOAT): return -1.0
      if compare(x, -0.0, EQ, FLOAT): return -0.0
      if compare(x, +0.0, EQ, FLOAT): return +0.0
      if compare(x, 0.0, GT, FLOAT): return 1.0
  elif is_complex(x):
    if x.real is NaN or x.imag is NaN:
      return NaN
    else:
      return divide(x, abs(x))
```

Feedback:
  1. "using stablehlo.compare as the comparator" sounds like a detail that best belongs in the formalism.
  2. Let's use Python syntax - I made a few changes to the code to accommodate that: a) added a "def", b) changed "if element_type(x) is foo" to just "is_foo(x)", 3) changed "else if" to "elif".
  3. I've also compressed the code by dropping "stablehlo.", removing kwarg syntax (since constant names unambiguously point to which parameter is which) and moving most returns to the same line as ifs to simplify eyeballing.

burmako (changes requested):

sdasgup3 (approved):
The current PR with the proposed changes looks good to me.

burmako (approved):

burmako:
Merging without waiting for CI - the only CI-able change in the PR is documentation within the .td file.

PR #495, opened 2022-11-09
Add spec for CbrtOp

### LINKED ISSUE #466: Add spec for CbrtOp

ghpvnist:
closes #466

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -184,6 +184,7 @@ described below)
    * [batch_norm_training](#stablehlobatch_norm_training)
    * [broadcast_in_dim](#stablehlobroadcast_in_dim)
    * [case](#stablehlocase)
+   * [cbrt](#stablehlocbrt)
    * [ceil](#stablehloceil)
    * [cholesky](#stablehlocholesky)
    * [clamp](#stablehloclamp)
@@ -702,6 +703,40 @@ returned.
   "stablehlo.return"(%result_branch1) : (tensor<i32>) -> ()
 }) : (tensor<i32>) -> tensor<i32>
 // %result: 11
+```
+
+[Back to Ops](#index-of-ops)
+
+## stablehlo.cbrt
+
+### Semantics
+
+Performs element-wise cubic root operation on `operand` tensor and produces a
+`result` tensor, implementing the `rootn(x, 3)` operation from the IEEE-754
+specification.
+
+### Inputs
+
+| Name      | Type                          |
+|-----------|-------------------------------|
+| `operand` | tensor of floating-point type |
+
+### Outputs
+
+| Name     | Type                          |
+|----------|-------------------------------|
+| `result` | tensor of floating-point type |
+
+### Constraints
+
+  * (C1) `operand` and `result` have the same type.
+
+### Examples
+
+```mlir
+// %operand: [0.0, 1.0, 8.0, 27.0]
+%result = "stablehlo.cbrt"(%operand) : (tensor<4xf32>) -> tensor<4xf32>
+// %result: [0.0, 1.0, 2.0, 3.0]
 ```
 
 [Back to Ops](#index-of-ops)
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -713,7 +713,7 @@ returned.
 
 Performs element-wise cubic root operation on `operand` tensor and produces a
 `result` tensor, implementing the `rootn(x, 3)` operation from the IEEE-754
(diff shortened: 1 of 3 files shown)


docs/status.md
```diff
@@ -57,7 +57,7 @@ one of the following tracking labels.
 | broadcast                | no            | yes*         | yes*           | yes             | no          |
 | broadcast_in_dim         | yes           | yes          | infeasible     | yes             | no          |
 | case                     | yes           | revisit      | yes            | no              | no          |
-| cbrt                     | no            | yes*         | yes*           | yes             | no          |
+| cbrt                     | yes           | yes          | yes            | yes             | no          |
```
subhankarshah, line 60: Does is fail for complex types on during compile time

subhankarshah (approved):
LGTM with one comment

docs/status.md
subhankarshah: If so open a ticket

ghpvnist: Here, the input and output type is explicitly `HLO_FpTensor`, so it will fail during compile time.
https://github.com/openxla/stablehlo/blob/e9cacdccbb73b8b03b4de6e8be110eec94dd3d31/stablehlo/dialect/StablehloOps.td#L218-L219

burmako: [shape_inference.cc](https://github.com/tensorflow/tensorflow/blob/49959d2f7d5e497edf90d6cdebc9afa3247234b7/tensorflow/compiler/xla/service/shape_inference.cc#L265-L274) allows both floating-point and complex inputs/outputs. Similar ops - `rsqrt` and `sqrt` - also allow complex inputs/outputs. It looks like this is a bug in `cbrt`, and we should have a ticket to allow complex tensors as well? (And Verification should be "no"?).

docs/spec_draft.md
```diff
@@ -518,6 +519,40 @@ returned.
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.cbrt
+
+### Semantics
+
+Performs element-wise cubic root operation on `operand` tensor and produces a
+`result` tensor, implementing the `rootn(x, 3)` operation from the IEEE-754
+specification.
```
burmako, line 528: Do we want to say something about numeric precision, similarly to what we did for `rsqrt`?

burmako (changes requested):

docs/spec_draft.md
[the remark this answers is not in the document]
sdasgup3: Thanks for the pointers! I strongly feel that we should drop this op.

At thin point we should add a ticket to track the drop this op from StableHLO and its addition to https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/ChloOps.td.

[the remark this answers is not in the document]
burmako: "and close this PR". I think we should keep this PR because for better or for worse `cbrt` is already part of the StableHLO opset, so we'll need to keep it around for a while for compatibility reasons. At this point, let's limit ourselves to opening a ticket along the lines of "Consider moving CbrtOp to CHLO" and referencing this discussion there.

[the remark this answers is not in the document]
ghpvnist: I opened #586 to continue the discussion of potentially moving `cbrt` to CHLO.

docs/status.md
ghpvnist: Thanks, I've created an issue #587 to track this.

docs/status.md
```diff
@@ -57,7 +57,7 @@ one of the following tracking labels.
 | broadcast                | no            | yes*         | yes*           | yes             | no          |
 | broadcast_in_dim         | yes           | yes          | infeasible     | yes             | no          |
 | case                     | yes           | revisit      | yes            | no              | no          |
-| cbrt                     | yes           | yes          | yes            | yes             | no          |
+| cbrt                     | yes           | no           | yes            | yes             | no          |
```
burmako, line 60: Verification should be "revisit".

docs/spec_draft.md
```diff
@@ -713,19 +713,20 @@ returned.
 
 Performs element-wise cubic root operation on `operand` tensor and produces a
 `result` tensor, implementing the `rootn(x, 3)` operation from the IEEE-754
-specification.
+specification. For complex element types, it computes a complex cube root, with
```
burmako, line 716: Typo: "cubic root".

burmako (changes requested):

burmako (approved):

PR #496, opened 2022-11-09
Add spec for ShiftLeftOp, ShiftRightArithmeticOp and ShiftRightLogicalOp

### LINKED ISSUE #477: Add spec for ShiftLeftOp

### LINKED ISSUE #478: Add spec for ShiftRightArithmeticOp

### LINKED ISSUE #479: Add spec for ShiftRightLogicalOp

subhankarshah:
closes #477
closes #479
closes #478

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -197,6 +197,9 @@ described below)
    * [rng](#stablehlorng)
    * [rsqrt](#stablehlorsqrt)
    * [select](#stablehloselect)
+   * [shift_left](#stablehloshift_left)
+   * [shift_right_arithmetic](#stablehloshift_right_arithmetic)
+   * [shift_right_logical](#stablehloshift_right_logical)
    * [sine](#stablehlosine)
    * [slice](#stablehloslice)
    * [sort](#stablehlosort)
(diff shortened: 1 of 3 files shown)


docs/spec_draft.md
```diff
@@ -2146,6 +2149,111 @@ where `predicate = rank(pred) == 0 ? pred : pred[i0, ..., iR-1]`.
...
+
+[Back to Ops](#index-of-ops)
+
+## stablehlo.shift_right_arithmetic
+
+### Semantics
+
+Performs element-wise arithmetic right-shift operation on the `lhs` tensor by
+`rhs` number of bits and produces a `result` tensor.
+
```
ghpvnist, line 2193: It might be handy to have a bit-representational example to distinguish the difference between arithmetic and logical right shift. Wdyt @burmako?

docs/status.md
```diff
@@ -136,9 +136,9 @@ one of the following tracking labels.
 | select_and_scatter       | no            | revisit      | no             | no              | no          |
 | send                     | no            | revisit      | no             | no              | no          |
 | set_dimension_size       | no            | yes*         | yes*           | yes             | no          |
-| shift_left               | no            | yes*         | yes*           | yes             | no          |
-| shift_right_arithmetic   | no            | yes*         | yes*           | yes             | no          |
-| shift_right_logical      | no            | yes*         | yes*           | yes             | no          |
+| shift_left               | yes           | yes          | yes            | yes             | no          |
```
ghpvnist, line 139: Verification for all three would need to be `revisit` here since it does not currently check that only integers are allowed.

ghpvnist (approved):
Minor comments, LGTM.

docs/status.md
subhankarshah: Sure

subhankarshah:
#512 ticket for revisiting verification

docs/spec_draft.md
```diff
@@ -2146,6 +2149,111 @@ where `predicate = rank(pred) == 0 ? pred : pred[i0, ..., iR-1]`.
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.shift_left
+
+### Semantics
+
+Performs element-wise left-shift operation on the `lhs` tensor by `rhs` number
```
ghpvnist, line 2156: What happens in case sign bit overflows? e.g. 1<-0 or 0<-1 for the sign bit? It could be obvious for some, but could also be worth pointing it out that bits are shifted regardless of the signedness. Here is the proposal for the wording and the following ops for consistency. Wdyt?:
```
Performs element-wise signless bit-shift operation on the `lhs` tensor by `rhs`
number of bits to the left and produces a `result` tensor.
```

docs/spec_draft.md
```diff
@@ -2146,6 +2149,111 @@ where `predicate = rank(pred) == 0 ? pred : pred[i0, ..., iR-1]`.
...
+// %result: [-2, -8, 24, 0, -128, 0]
+```
+
+[Back to Ops](#index-of-ops)
+
+## stablehlo.shift_right_arithmetic
+
+### Semantics
+
+Performs element-wise arithmetic right-shift operation on the `lhs` tensor by
```
ghpvnist, line 2191: Similarly:
```
Performs element-wise arithmetic bit-shift operation on the `lhs` tensor by
`rhs` number of bits to the right and produces a `result` tensor.
```

docs/spec_draft.md
```diff
@@ -2146,6 +2149,111 @@ where `predicate = rank(pred) == 0 ? pred : pred[i0, ..., iR-1]`.
...
+// %result: [-1, -32, -5, 1, 1, 0]
+```
+
+[Back to Ops](#index-of-ops)
+
+## stablehlo.shift_right_logical
+
+### Semantics
+
+Performs element-wise logical right-shift operation on the `lhs` tensor by `rhs`
```
ghpvnist, line 2226: Similarly:
```
Performs element-wise logical bit-shift operation on the `lhs` tensor by `rhs`
number of bits to the right and produces a `result` tensor.
```

ghpvnist (changes requested):

docs/spec_draft.md
subhankarshah: I have covered the case in examples. Since left-shift operation is a standard common operation in most languages wrote it as is.

docs/spec_draft.md
subhankarshah: I would not prefer changing `right-shift` -> `bit-shift` because the operation is widely known as `logical right-shift`.

docs/spec_draft.md
subhankarshah: I would not prefer changing `right-shift` -> `bit-shift` because the operation is widely known as `arithmetic right-shift`.

subhankarshah: mentioning `right` or `left` is important IMO

subhankarshah:
@ghpvnist I agree formal semantics with bit representation would make things very clear, and I had half a mind to write that even before your comment. The other comments about reword might not be an issue if we decide to have the bit representations as formal semantics.

ghpvnist (approved):
LGTM with open comment to decide whether bit representation example would be worth having.

docs/spec_draft.md
burmako: Let's figure this out as part of #289 and then apply uniformly to all examples in the spec. Meanwhile, I propose we leave these examples as is.

docs/spec_draft.md
burmako: Agreed about keeping "left shift", "arithmetic right shift" and "logical right shift" without explanations, similarly how we are doing this for well-known floating-point functions like `sin` or `cos`.

docs/spec_draft.md
```diff
@@ -2146,6 +2149,111 @@ where `predicate = rank(pred) == 0 ? pred : pred[i0, ..., iR-1]`.
...
+
+### Constraints
+
+  * (C1) `lhs`, `rhs`, and `result` have the same type.
+
+### Examples
+
+```mlir
+// %lhs: [-1, -128, -36, 5, 3, 7]
+// %rhs: [1, 2, 2, 2, 1, 3]
```
burmako, line 2250: `shift_right_arithmetic` is using `%rhs: [1, 2, 3, 2, 1, 3]`. Let's do the same here? The result would be `[127, 32, 27, 1, 1, 0]`.

burmako (changes requested):

burmako (approved):

PR #502, opened 2022-11-10
Add spec for Expm1Op

### LINKED ISSUE #469: Add spec for Expm1Op

ghpvnist:
closes #469

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -175,6 +175,7 @@ described below)
    * [count_leading_zeros](#stablehlocount_leading_zeros)
    * [divide](#stablehlodivide)
    * [exponential](#stablehloexponential)
+   * [exponential_minus_one](#stablehloexponential_minus_one)
    * [fft](#stablehlofft)
    * [floor](#stablehlofloor)
    * [gather](#stablehlogather)
@@ -936,6 +937,42 @@ implementation-defined.
 // %operand: (1.0, 2.0)
 %result = "stablehlo.exponential"(%operand) : (tensor<complex<f32>>) -> tensor<complex<f32>>
 // %result: (-1.13120438, 2.47172667)
+```
+
+[Back to Ops](#index-of-ops)
+
+## stablehlo.exponential_minus_one
+
+### Semantics
+
+Performs element-wise exponential minus one operation on `operand` tensor and
+produces a `result` tensor. For floating-point element types, it implements the
+`expm1` operation from the IEEE-754 specification. For complex element types, it
+computes a complex exponential minus one, with corner cases TBD. Numeric
+precision is implementation-defined.
+
+### Inputs
+
+| Name      | Type                                     |
+|-----------|------------------------------------------|
+| `operand` | tensor of floating-point or complex type |
+
+### Outputs
+
+| Name     | Type                                     |
+|----------|------------------------------------------|
+| `result` | tensor of floating-point or complex type |
+
+### Constraints
+
+  * (C1) `operand` and `result` have the same type.
+
+### Examples
+
+```mlir
+// %operand: [0.0, 1.0]
+%result = "stablehlo.exponential_minus_one"(%operand) : (tensor<2xf32>) -> tensor<2xf32>
+// %result: [0.0, 1.71828187]
 ```
 
 [Back to Ops](#index-of-ops)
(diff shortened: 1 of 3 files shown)


docs/spec_draft.md
```diff
@@ -850,6 +851,42 @@ implementation-defined.
...
 [Back to Ops](#index-of-ops)
 
+## stablehlo.exponential_minus_one
+
+### Semantics
+
+Performs element-wise exponential minus one operation on `operand` tensor and
+produces a `result` tensor. For floating-point element types, it implements the
+`expm1` operation from the IEEE-754 specification. For complex element types, it
+computes a complex exponential minus one, with corner cases TBD. Numeric
```
subhankarshah, line 861: The PowOp is similar, I did not mention anything about corner cases, not sure if it is required or mention of IEEE spec although precision being implementation defined is correct to mention. Complex exponential is well defined mathematically.

subhankarshah: Let me know your thoughts, PowOp is similar in some regards

docs/spec_draft.md
```diff
@@ -850,6 +851,42 @@ implementation-defined.
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.exponential_minus_one
+
+### Semantics
+
+Performs element-wise exponential minus one operation on `operand` tensor and
```
subhankarshah, line 858: "exponential operation" is well defined? Or we can write
```Performs element-wise exponentiation of Euler's number `e` by the `operand` tensor followed by a decrement of 1 and produces a `result` tensor.```

Comment (subhankarshah) [reply docs/spec_draft.md:858]: Given we explain the operation, we might(or might not) choose to omit the stuff about IEEE-754 and corner cases.

File: docs/status.md [hunk 3]
```diff
@@ -88,7 +88,7 @@ one of the following tracking labels.
 | dynamic_update_slice     | no            | revisit      | no             | yes             | no          |
 | einsum                   | no            | revisit      | no             | no              | no          |
 | exponential              | yes           | yes          | yes            | yes             | no          |
-| exponential_minus_one    | no            | yes*         | yes*           | yes             | no          |
+| exponential_minus_one    | yes           | yes          | yes            | yes             | no          |
```
subhankarshah, line 91: Did you check that integer operand gives compilation errors and not unimplemented error during runtime?

subhankarshah (changes requested):
Added a few comments

docs/spec_draft.md
subhankarshah: I am thinking why should we link IEEE spec for simple ops such as these, but we have done that for other ops such as addition etc, so we can refer the spec for consistency.

ghpvnist: > Let me know your thoughts, PowOp is similar in some regards

Complex exponential is well defined, but still need to mention it here since it is not defined in IEEE-754, and we have been following this convention for other similar elemental ops thus written here for consistency.

> I am thinking why should we link IEEE spec for simple ops such as these

We are aiming to follow a standard for ops that require special handling such as numeric precision, error handling, corner cases, etc., and IEEE-754 is a well-established one that has defined most, if not all, of these behaviors.

docs/status.md
ghpvnist: Thanks for pointing it out. Filed an issue #509

docs/spec_draft.md
ghpvnist: Exponential operations is well defined and also in stablehlo.exponential. I like the current way since it is more consistent with stablehlo.exponential's semantics and is also concise. Wdyt about keeping the semantics as-is?

docs/status.md
```diff
@@ -88,7 +88,7 @@ one of the following tracking labels.
 | dynamic_update_slice     | no            | revisit      | no             | yes             | no          |
 | einsum                   | no            | revisit      | no             | no              | no          |
 | exponential              | yes           | yes          | yes            | yes             | no          |
-| exponential_minus_one    | no            | yes*         | yes*           | yes             | no          |
+| exponential_minus_one    | yes           | revisit      | yes            | yes             | no          |
```
subhankarshah, line 91: Can you mention the verification ticket for revisit

subhankarshah (approved):
LGTM

docs/status.md
ghpvnist: Sure, it's mentioned in a previous [comment](https://github.com/openxla/stablehlo/pull/502#discussion_r1019756530), but I'll mention it as a standalone comment for visibility.

burmako (approved):

PR #505, opened 2022-11-10
Add spec for PowOp

### LINKED ISSUE #473: Add spec for PowOp

subhankarshah:
closes #473

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -2221,22 +2221,36 @@ and produces a `result` tensor.
 Performs element-wise exponentiation of `lhs` tensor by `rhs` tensor and
 produces a `result` tensor.
 
+For integer element types, if the exponentiation has an unsigned/signed
+overflow, the result is implementation-defined and one of the following:
+
+  * mathematical result modulo $2^n$, where n is the bit width of the result,
+  for unsigned overflow. For signed integer overflow, wraps the result around
+  the representable range $[-2^{n-1},\ \ 2^{n-1} - 1]$.
+  * saturation to $2^{n-1} - 1$ (or $-2^{n-1}$) for signed overflow and
+  saturation to $2^n - 1$ (or $0$) for unsigned overflow.
+
+For floating-point element types, it implements the `pow` operation from the
+IEEE-754 specification. For complex element types, it computes complex
+exponentiation, with corner cases TBD. Numeric precision is
+implementation-defined.
+
 ### Inputs
 
-| Name  | Type                                              |
-|-------|---------------------------------------------------|
-| `lhs` | tensor of integer, floating-point or complex type |
-| `rhs` | tensor of integer, floating-point or complex type |
+| Name  | Type                                               |
+|-------|----------------------------------------------------|
+| `lhs` | tensor of integer, floating-point, or complex type |
+| `rhs` | tensor of integer, floating-point, or complex type |
 
 ### Outputs
 
-| Name     | Type                                              |
-|----------|---------------------------------------------------|
-| `result` | tensor of integer, floating-point or complex type |
+| Name     | Type                                               |
+|----------|----------------------------------------------------|
+| `result` | tensor of integer, floating-point, or complex type |
 
 ### Constraints
 
-  * (C1) `lhs`, `rhs` and, `result` have the same type.
+  * (C1) `lhs`, `rhs`, and `result` have the same type.
 
 ### Examples
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -2230,6 +2230,12 @@ overflow, the result is implementation-defined and one of the following:
   * saturation to $2^{n-1} - 1$ (or $-2^{n-1}$) for signed overflow and
   saturation to $2^n - 1$ (or $0$) for unsigned overflow.
 
+For an integer, `x`, raised to a negative power, `y`, the behaviour is as
+follows:
+  * If `abs(x)` $\gt$ 1, then result is 0.
+  * If `abs(x)` $=$ 1, then result is equivalet to `x^abs(y)`.
+  * If `abs(x)` $=$ 0, then behaviour is implementation defined.
+
 For floating-point element types, it implements the `pow` operation from the
 IEEE-754 specification. For complex element types, it computes complex
 exponentiation, with corner cases TBD. Numeric precision is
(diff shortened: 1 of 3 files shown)


docs/spec_draft.md
```diff
@@ -1791,6 +1792,41 @@ and produces a `result` tensor.
...
+
+### Outputs
+
+| Name     | Type                                              |
+|----------|---------------------------------------------------|
+| `result` | tensor of integer, floating-point or complex type |
+
+### Constraints
+
+  * (C1) `lhs`, `rhs` and, `result` have the same type.
```
ghpvnist, line 1817:
```
* (C1) `lhs`, `rhs`, and `result` have the same type.
```

docs/spec_draft.md
```diff
@@ -1791,6 +1792,41 @@ and produces a `result` tensor.
...
+
+### Examples
+
+```mlir
+// %lhs: [-2.0, -0.0, -36.0, 5.0, 3.0, 10000.0]
+// %rhs: [2.0, 2.0, 1.1, 2.0, -1.0, 10.0]
+%result = "mhlo.power"(%lhs, %rhs) : (tensor<6xf32>, tensor<6xf32>) -> tensor<6xf32>
```
ghpvnist, line 1824:
```
%result = "stablehlo.power"(%lhs, %rhs) : (tensor<6xf32>, tensor<6xf32>) -> tensor<6xf32>
```

docs/spec_draft.md
```diff
@@ -1791,6 +1792,41 @@ and produces a `result` tensor.
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.power
+
+### Semantics
+
+Performs element-wise exponentiation of `lhs` tensor by `rhs` tensor and
+produces a `result` tensor.
```
ghpvnist, line 1800: 1. Can you refer to the IEEE-754 semantics?
2. What is the semantics of a complex power?
3. What happens to integer overflow?

docs/spec_draft.md
```diff
@@ -1791,6 +1792,41 @@ and produces a `result` tensor.
...
+
+### Inputs
+
+| Name  | Type                                              |
+|-------|---------------------------------------------------|
+| `lhs` | tensor of integer, floating-point or complex type |
```
ghpvnist, line 1806: Nit: Oxford comma here and below.

docs/status.md
```diff
@@ -113,7 +113,7 @@ one of the following tracking labels.
 | outfeed                  | no            | revisit      | no             | no              | no          |
 | pad                      | yes           | yes          | yes            | yes             | no          |
 | popcnt                   | yes           | yes          | yes            | yes             | no          |
-| power                    | no            | yes*         | yes*           | yes             | no          |
+| power                    | yes           | yes          | yes            | yes             | no          |
```
ghpvnist, line 116: power is also one of those StableHLO_BinaryElementwiseOp where it does not have complete verification check for supported types. This should be mark `revisit` with a link to an issue addressing this.

ghpvnist (changes requested):

docs/status.md
subhankarshah: I checked it fails for i1 during compilation

subhankarshah: Other types are allowed AFAIR

ghpvnist: On the StableHLO side?

subhankarshah: stablehlo-opt allows i1, need to add a ticket, thanks

subhankarshah: #513

docs/spec_draft.md
subhankarshah: thanks

ghpvnist (approved):

docs/spec_draft.md
```diff
@@ -1791,6 +1792,55 @@ and produces a `result` tensor.
...
+
+Performs element-wise exponentiation of `lhs` tensor by `rhs` tensor and
+produces a `result` tensor.
+
+For integer element types, if the exponentiation has an unsigned/signed
+overflow, the result is implementation-defined and one of the following:
+
+  * mathematical result modulo $2^n$, where n is the bit width of the result,
+  for unsigned overflow. For signed integer overflow, wraps the result around
+  the representable range $[-2^{n-1},\ \ 2^{n-1} - 1]$.
```
ghpvnist, line 1807: Double space?
```
the representable range $[-2^{n-1},\ 2^{n-1} - 1]$.
```

docs/spec_draft.md
```diff
@@ -1791,6 +1792,55 @@ and produces a `result` tensor.
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.power
+
+### Semantics
```
ghpvnist, line 1797: Have you considered what happens with negative `lhs` and negative `rhs`?

subhankarshah: The behaviour is standard mathematical behaviour and illegal expressions result in NANs. Let me know if you feel we need to write something in this regard.

burmako: That's fine for floating-point numbers (and for complex numbers too, for which we consistently say "corner cases TBD" in the spec). What about integers though? NaN won't cut it for integers.

subhankarshah: negative power for integers greater than 1 results in 0
negative integers in `lhs` work as expected because fractional powers are not permitted due to C1
0 raised to negative power causes runtime exception.

subhankarshah: divide by zero resulting in exception is expected behaviour and everything else also seems expected behaviour.

burmako: "negative power for integers greater than 1 results in 0". This was not expected behavior for me. Does this mean that this computation is performed in floating point and then rounded towards negative infinity or something?

"divide by zero resulting in exception is expected behaviour". This is inconsistent with https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md#errors which says "error conditions are [...] possible, e.g. through integer overflows, out-of-bounds accesses, etc. Unless explicitly called out, all these errors result in implementation-defined behavior".

Overall, it looks like we need to explicitly document the behavior of raising integers to negative powers with three distinct cases: 1) less than one, 2) one, 3) greater than one.

burmako (approved):

Issue #485, opened 2022-11-08
Move MHLO machinery for printing array refs into AssemblyFormat


GleasonK:
### Request description

Found in #403. Move printing machinery for array refs to `AssemblyFormat.h`. Currently we have `printDenseI64Array`, which works for `I64ElementsAttr` and `BroadcastDimsAttr` but not for `def StableHLO_Dim : ArrayRefParameter<"int64_t", "Dimension">;`

As a follow-up change the HLODim printing to use this new functionality as suggested in #403.

burmako:
Once done, let's use the new assembly format in `OutputOperandAlias`. See #403 for context.

burmako:
The migrate from MHLO part is done in https://github.com/openxla/stablehlo/pull/546, which has just been merged. As discussed, there's still the matter of sharing this machinery with MHLO via `AssemblyFormat.h`.

GleasonK:
Fixed in https://github.com/tensorflow/mlir-hlo/commit/c1f1538a2cc3f7508842403f6f6f13ce52d8834f

PR #546, opened 2022-11-14
Add parser for StableHlo_Dim

atondwal:
Pulling in patch from tensorflow/third_party/stablehlo, which is needed for
stablehlo-to-mhlo roundtrip littests


stablehlo/dialect/StablehloAttrs.td
```diff
@@ -20,7 +20,10 @@ limitations under the License.
 include "mlir/IR/OpBase.td"
 include "mlir/IR/TensorEncoding.td"
 
-def StableHLO_Dim : ArrayRefParameter<"int64_t", "Dimension">;
+def StableHLO_Dim : ArrayRefParameter<"int64_t", "Dimension"> {
+  let parser = "mlir::stablehlo::parseIntArray($_parser)";
+  let printer = "mlir::stablehlo::printIntArray($_printer, $_self)";
+}
```
burmako, line 26: Can you also migrate `OutputOperandAlias` to use the same assembly format that's used in MHLO?

burmako (changes requested):

stablehlo/dialect/StablehloAttrs.td
GleasonK: +1, was thinking the same

GleasonK (approved):
LGTM, I have plenty of cleanup to do in terms of migrating communal code from `StablehloOp.h --> AssemblyFormat.h`, and I can add this to the list.

I would still consider #485 open until this code is shared. But that work is a little more involved (submit to stablehlo, integrate, move mhlo to shared code) and I'll be doing related things so happy to do that later.

burmako (approved):

Issue #514, opened 2022-11-14
Dedicated markdown doc for Reference


bhack:
### Request description

Can we have a dedicated markdown doc for Reference?
There are some info in the interpreter markdown but I think having a distinct document for  reference, as it is also an indipendent folder in the repo, it could also help to clarify the interpreter domain vs the reference impl.

sdasgup3:
Thanks for the request. 

Currently `stableho/docs` is the one stop-shop for all the documentation of the repo. IMO, making the existing doc more 
informative, by identifying and filing the gaps, would be a better approach than departing from the norm.

Just to understand the requirement even better: 
>> clarify the interpreter domain vs the reference impl.
From the [Data Model Section](https://github.com/openxla/stablehlo/blob/main/docs/interpreter.md#data-model) of the existing doc, the domain of reference implementation can be inferred to be StableHLO programs. But it can be made clearer or/and maybe I am missing something else. Can you please elaborate on it further?

bhack:
I think part it is the naming as we have an `interpreter.md` but interpreter is also a c++ file in reference folder. 

Other then the naming, probably if we want to keep the same single file we could separate the section for data model, interpreter registration (`interpreter.cc`?) and the implementation of the execution semantic of the op.

What do you think?

sdasgup3:
Dear @bhack 
Thanks for the explanation and suggestions!
 
Let us go with renaming the doc to `reference.md` and thereby avoiding any unintentional connection of it with implementation filenames.

PR #549, opened 2022-11-15
renaming interpreter.md

sdasgup3:
fixes https://github.com/openxla/stablehlo/issues/514

cc: @bhack


bhack:
> The op-level eval as mentioned in (2)

What is `(2)`?

sdasgup3:
>> What is (2)?
Thanks for the catch. Fixed it in the doc.

bhack:
I will separate more 1 and 2 in two section but if the text is too short it is ok.

sdasgup3:
Let us keep the bullets as is. 
Fixed the indentation.

Issue #560, opened 2022-11-16
Removing complex type supported for Min/Max ops


sdasgup3:
### Request description

From https://github.com/openxla/stablehlo/pull/187#discussion_r979437990, we were having parity with Jax on the support for complex type for Min/Max Ops. As HLO does not support those types, Jax lowers them to to first MHLO by rewriting a set of [lexicographic comparisons](https://github.com/google/jax/blob/7c85ca38f45e3ae0d8a80b2d33cf908de9c6564f/jax/_src/lax/lax.py#L2230-L2238), and then the MHLO trivially gets converted to HLO. 

With this ticket, we decide to keep the type support of StableHLO min/max same as that of HLO. Jax needs to do similar lowering, as mentioned before, to use StableHLO directly.

sdasgup3:
Involves: 
1. Updating the spec.
2. Adding verification checks to avoid complex types. 
3. Fixing the complex type support in the interpreter.

burmako:
Should we also remove support for complex tensors from ClampOp and SignOp (see linked PRs with discussions above)?

burmako:
Deprioritizing this issue to focus on completing the StableHLO spec by the end of the year. Planning to pick it up again in early 2023.

burmako:
Re: SignOp, we have decided to keep complex support there: https://github.com/openxla/stablehlo/pull/494#discussion_r1028678291.

sdasgup3:
Related issue: https://github.com/openxla/stablehlo/issues/997

PR #561, opened 2022-11-16
remove complex type support from Min/Max Ops

sdasgup3:
fixes https://github.com/openxla/stablehlo/issues/560

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -1559,22 +1559,21 @@ function, with corner cases TBD. Numeric precision is implementation-defined.
 
 Performs element-wise max operation on tensors `lhs` and `rhs` and produces a
 `result` tensor. For floating-point element types, it implements the `maximum`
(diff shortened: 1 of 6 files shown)


burmako (changes requested):
Thank you for your contribution! Given that this pull request changes the status quo that exists for MHLO/HLO, we'll need to turn it into an RFC. Our RFC process is still work in progress (#196), so let's play it by ear in the meanwhile.

Can you update the description of the PR to provide a detailed motivation? This will be essential for reviewing the change in more detail than usual, warranted by the divergence from MHLO/HLO.

Something like: 1) Background (describe the status quo, including what producers and what consumers support complex min/max), 2) Proposal (basically a summary of this PR), 3) Compatibility implications (please coordinate with @GleasonK on this).

burmako:
Closing this pull request for now, with the intent to revive it once #560 is picked up again.

PR #566, opened 2022-11-17
Migrate MHLO changes for printing dimensions

GleasonK:
- Migrate unknown dimension printing `printDimensionSizes`.
- Move `parseIntArray` to `AssemblyFormat.h` to be shared.

This unblocks work to make a few other attributes declarative. Also opens the door to make some printing infrastructure communal between MHLO/StableHLO. Will look into removing `parseDims` today.

### DIFF
--- a/stablehlo/dialect/AssemblyFormat.cpp
+++ b/stablehlo/dialect/AssemblyFormat.cpp
@@ -313,7 +313,8 @@ void printDynamicShape(AsmPrinter& p, llvm::ArrayRef<int64_t> shape) {
   p << ']';
 }
 
-ParseResult parseDynamicShape(AsmParser& parser, FailureOr<SmallVector<int64_t>> & shapeResult) {
+ParseResult parseDynamicShape(AsmParser& parser,
+                              FailureOr<SmallVector<int64_t>>& shapeResult) {
   SmallVector<int64_t> shape;
   auto parseElt = [&]() -> ParseResult {
     if (!parser.parseOptionalQuestion()) {
@@ -322,7 +323,8 @@ ParseResult parseDynamicShape(AsmParser& parser, FailureOr<SmallVector<int64_t>>
     }
     return parser.parseInteger(shape.emplace_back());
   };
-  auto res = parser.parseCommaSeparatedList(AsmParser::Delimiter::Square, parseElt);
+  auto res =
+      parser.parseCommaSeparatedList(AsmParser::Delimiter::Square, parseElt);
   if (succeeded(res)) {
     shapeResult = shape;
   } else {
@@ -403,12 +405,13 @@ void printHloDim(AsmPrinter& printer, ArrayRef<int64_t> ints) {
   printer << ']';
 }
 
-ParseResult parseIntArray(AsmParser& parser, FailureOr<SmallVector<int64_t>> & ints) {
+ParseResult parseIntArray(AsmParser& parser,
+                          FailureOr<SmallVector<int64_t>>& ints) {
   ints = parseHloDim(parser);
   return success(/*isSuccess=*/succeeded(ints));
 }
 
-ParseResult parseIntArray(AsmParser& parser, SmallVector<int64_t> & ints) {
+ParseResult parseIntArray(AsmParser& parser, SmallVector<int64_t>& ints) {
   auto intsOrFail = parseHloDim(parser);
   if (failed(intsOrFail)) {
     return failure();
@@ -420,8 +423,6 @@ ParseResult parseIntArray(AsmParser& parser, SmallVector<int64_t> & ints) {
 void printIntArray(AsmPrinter& printer, ArrayRef<int64_t> ints) {
   printHloDim(printer, ints);
 }
-
-
 
 }  // namespace hlo
 }  // namespace mlir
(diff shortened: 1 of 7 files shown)


stablehlo/dialect/StablehloAttrs.td
```diff
@@ -21,8 +21,8 @@ include "mlir/IR/OpBase.td"
 include "mlir/IR/TensorEncoding.td"
 
 def StableHLO_Dim : ArrayRefParameter<"int64_t", "Dimension"> {
```
burmako, line 23: In MHLO, we recently renamed this to `MHLO_Dims`. Since we're here, how about we rename this to `StableHLO_DIms`? This is a plural thing after all.

stablehlo/dialect/AssemblyFormat.h
```diff
@@ -172,6 +172,17 @@ void printDenseI64Array(OpAsmPrinter& p, Operation* op,
 
 ParseResult parseDenseI64Array(OpAsmParser& parser, DenseIntElementsAttr& attr);
 
+// DynamicShape - Print an array of ints. Dynamic dimensions printed as `?`.
```
burmako, line 175: Proposed rewording:

```
DimensionSizes - Print an array of dimension sizes. Dynamic sizes printed as `?`
```

Feedback
  1. `ArrayRef<int64_t>` doesn't have to be dynamic, so I think `DimensionSizes` is a better fitting name.
  1. Strictly speaking, dimensions (i.e. axes) and dimension sizes are two different things.

stablehlo/dialect/AssemblyFormat.h
```diff
@@ -198,6 +209,24 @@ void printCustomCallTarget(AsmPrinter& p, Operation*, StringAttr target);
 
 ParseResult parseCustomCallTarget(AsmParser& parser, StringAttr& target);
 
+// HloDim - Print an ArrayRef of integrer values with brackets.
```
burmako, line 212: Proposed rewording:

```
IntArray - Print an array of ints with brackets.
Unlike DimensionSizes, doesn't have special handling for kDynamicSize.
```

Feedback:
  1. I think that using "dim" in this name is confusing with respect to `DimensionSizes`. Let's just say `IntArray`?
  2. "array of ints" to be consistent with above.

stablehlo/dialect/AssemblyFormat.h
```diff
@@ -198,6 +209,24 @@ void printCustomCallTarget(AsmPrinter& p, Operation*, StringAttr target);
...
+void printHloDim(AsmPrinter& printer, ArrayRef<int64_t> ints);
+
+// IntArray - Calls parseHloDim, with an interface that works for
+// custom directives.
+ParseResult parseIntArray(AsmParser& parser, SmallVector<int64_t>& ints);
+
+ParseResult parseIntArray(AsmParser& parser,
+                          FailureOr<SmallVector<int64_t>>& ints);
+
+void printIntArray(AsmPrinter& printer, ArrayRef<int64_t> ints);
```
burmako, line 228: Is this used anywhere at the moment?

stablehlo/dialect/AssemblyFormat.cpp
```diff
@@ -300,6 +301,38 @@ ParseResult parseDenseI64Array(OpAsmParser& parser,
...
 }
 
+void printDynamicShape(AsmPrinter& p, llvm::ArrayRef<int64_t> shape) {
+  auto printIntOrQuestion = [&](int64_t value) {
+    if (ShapedType::isDynamic(value))
+      p << '?';
+    else
+      p << value;
+  };
+  p << " [";
```
burmako, line 311: Why does this function print a leading whitespace and the functions below don't?

stablehlo/dialect/AssemblyFormat.h
burmako, line 228: The functions above are ordered print/parse/print/parse/.... Here the order seems to have switched to parse/print/parse/print. Let's make things consistent?

stablehlo/dialect/AssemblyFormat.cpp
```diff
@@ -300,6 +301,38 @@ ParseResult parseDenseI64Array(OpAsmParser& parser,
...
+  llvm::interleaveComma(shape, p, printIntOrQuestion);
+  p << ']';
+}
+
+ParseResult parseDynamicShape(AsmParser& parser,
+                              FailureOr<SmallVector<int64_t>>& shapeResult) {
+  SmallVector<int64_t> shape;
+  auto parseElt = [&]() -> ParseResult {
```
burmako, line 319: `parseIntOrQuestion`?

burmako (changes requested):

stablehlo/dialect/AssemblyFormat.h
GleasonK: This is the one used by the custom directive

stablehlo/dialect/AssemblyFormat.h
GleasonK: As discussed in separate thread, would like this to use DimensionSizes, but need to make sure that is always used as dimension sizes first. Will look into that in followup PR.

GleasonK: Accepted proposed rewording.

stablehlo/dialect/AssemblyFormat.cpp
GleasonK: `parseDimensionSizes` - Needs to match for custom declaration.

stablehlo/dialect/AssemblyFormat.cpp
GleasonK: I'll add this to my list to investigate. This may be an upstream bug. There is no space automatically added between literals and custom directives when specifying declarative assembly _for attributes_. There is a space for ops.

For now, I'll keep the custom declarations consistent and add an explicit space to the tablegen declaration.

burmako (approved):

Issue #498, opened 2022-11-10
Missing verification of `replica_groups` for AllGather/AllReduce/AllToAll/CollectivePermute/ReduceScatter


sdasgup3:
### Request description

Seems that we have the [verification routine](https://github.com/openxla/stablehlo/blob/283702a1378d4319b0ac551c3b4502a8e22bc386/stablehlo/dialect/StablehloOps.h#L110), but not applied to the above ops. Need to evaluate its application for each op.

sdasgup3:
Even though collectivePermute does not have `replica_groups` as an explicit parameter. But the `source_target_pairs` have somewhat similar constraints to what we need to check for `replica_groups`.

burmako:
Also added ReduceScatterOp to the title, since we have five collective ops in StableHLO.

sdasgup3:
As per the dicsusiion in https://github.com/openxla/stablehlo/pull/503#discussion_r1035095336, the verifiier should expect the `replica_groups` to be non-empty.

sdasgup3:
closed my mistake. There are still some pending items to be taken care of.

PR #564, opened 2022-11-17
Add spec for ReduceScatterOp

sdasgup3:
fixes #524 

A few points
  - Type Inference is marked `infeasible` as the return type of the op depends upon the [shard_count](https://github.com/tensorflow/tensorflow/blob/20c6943d3cd7e07da162f7778a0af5d3776274b4/tensorflow/compiler/xla/service/hlo_verifier.cc#L452) which [depends on the result type](https://github.com/tensorflow/tensorflow/blob/20c6943d3cd7e07da162f7778a0af5d3776274b4/tensorflow/compiler/xla/service/hlo_verifier.cc#L426). Note that the `shard_count` is a parameter in HLO spec, and not present in MHLO. During export of MHLO to HLO, this param is [derived ](https://github.com/tensorflow/tensorflow/blob/a1acd6a6466f58ed4197b7beaa2f3e0b6fcfc32a/tensorflow/compiler/xla/translate/mhlo_to_hlo/mlir_hlo_to_hlo.cc#L766) using result type.
  - Note that we still do not have the [check](https://github.com/tensorflow/tensorflow/blob/060ac3b5de3192f9746ba587854a73e7cc83c4fd/tensorflow/compiler/xla/service/hlo_verifier.cc#L515) for shard_count == subgroup_size. The subgroup_size depends on [module configuration](https://github.com/tensorflow/tensorflow/blob/20c6943d3cd7e07da162f7778a0af5d3776274b4/tensorflow/compiler/xla/service/hlo_verifier.cc#L95) which manages the settings and values which affect the compiled executable outside of the HLO code itself and I am not sure if that info is available in StableHLO IR.

**update**
 - With https://github.com/openxla/stablehlo/issues/650 and from [code reference](https://github.com/tensorflow/tensorflow/blob/fe8b73b03298b9ca01de22c9fd430ee2ca040821/tensorflow/compiler/xla/service/hlo_verifier.cc#L512), the type inference should be made feasible. At this point there is no type inference implementation, hence marked `no`.
 - The verifier should be `revisit` based on https://github.com/openxla/stablehlo/issues/652

PR #568, opened 2022-11-17
Add spec for CollectivePermuteOp

### LINKED ISSUE #523: Add spec for CollectivePermuteOp
### Request description

To spec `stablehlo.collective_permute`

sdasgup3:
fixes #523

A few points

1. adds verification checks for ColllectivePermute w.r.t. https://github.com/openxla/stablehlo/issues/498 
2. We do not have the upper bound check for replica ids in `source_target_pairs` as in https://github.com/tensorflow/tensorflow/blob/54926b8320184c2f07066d84b798e52ba6a74675/tensorflow/compiler/xla/service/hlo_verifier.cc#L714 because those checks [depends on](https://github.com/tensorflow/tensorflow/blob/54926b8320184c2f07066d84b798e52ba6a74675/tensorflow/compiler/xla/service/hlo_verifier.cc#L694) HloModuleConfig which not exposed in StableHLO. 

upd:
1. verification is `revisit` based on https://github.com/openxla/stablehlo/issues/498 (check empty replicas) and (2) above

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -1565,7 +1565,6 @@ The operation splits the StableHLO grid into `process_groups` as follows:
 Afterwards, `result@process` is given by:
 
  * `operand@process_groups[i, 0]`, if there exists an `i` such that `process_groups[i, 1] = process`.
- * `[0, ..., 0]`, otherwise.
  * `broadcast_in_dim(0, [], shape(result))`, otherwise.
 
 ### Inputs
@@ -1587,8 +1586,8 @@ Afterwards, `result@process` is given by:
   * (C1) dim(`source_target_pairs`, 1) $=$ 2.
   * (C2) All values in `source_target_pairs[:, 0]` are unique.
   * (C3) All values in `source_target_pairs[:, 1]` are unique.
-  * (C4) $0 \ge$ source_target_pairs[i][0], source_target_pairs[i][1] $\lt N$,
-         where $N$ is given by
+  * (C4) $0 \le$ source_target_pairs[i][0], source_target_pairs[i][1] $\lt N$,
+         where $N$ depends on the process grouping strategy:
     * If `cross_replica`, `num_replicas`.
     * If `cross_partition`, `num_partitions`.
   * (C5) type(`result`) $=$ type(`operand`).
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -1555,9 +1555,9 @@ The operation splits the StableHLO grid into `process_groups` as follows:
   * `channel_id > 0`,
     `cross_partition(replica_groups)`.
 
-Afterwards, `result@receiver` is given by
- * `operand@process_groups[i][0]`, if there exist an `i` such that  `process_groups[i][1] = receiver`.
- * `result@receiver = T, otherwise, # where T is a zero valued tensor constant with type(T) = type(operand)`, otherwise.
+Afterwards, `result@target` is given by
+ * `operand@process_groups[i][0]`, if there exist an `i` such that  `process_groups[i][1] = target`.
+ * `result@target = T, otherwise, # where T is a zero valued tensor constant with type(T) = type(operand)`, otherwise.
 
 ### Inputs
 
@@ -1576,8 +1576,8 @@ Afterwards, `result@receiver` is given by
 ### Constraints
 
   * (C1) dim(`source_target_pairs`, 1) $=$ 2.
-  * (C2) All memmbers of { `source_target_pairs`[i][0] : i $\in$ [0, dim(`source_target_pairs`, 0)) } are unique.
-  * (C3) All memmbers of { `source_target_pairs`[i][1] : i $\in$ [0, dim(`source_target_pairs`, 0)) } are unique.
+  * (C2) All members of { `source_target_pairs`[i][0] : i $\in$ [0, dim(`source_target_pairs`, 0)) } are unique.
+  * (C3) All members of { `source_target_pairs`[i][1] : i $\in$ [0, dim(`source_target_pairs`, 0)) } are unique.
   * (C4) $0 \gt$ source_target_pairs[i][0], source_target_pairs[i][1] $\lt N$,
          where $N$ is given by
     * `num_replicas`, if `cross_replica`.
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -645,7 +645,9 @@ Afterwards, within each `process_group`:
 // %operand@(1, 0): [[5.0, 6.0], [7.0, 8.0]]
 %result = "stablehlo.all_gather"(%operand) {
   all_gather_dim = 1 : i64,
-  replica_groups = dense<[[0, 1]]> : tensor<1x2xi64>
+  replica_groups = dense<[[0, 1]]> : tensor<1x2xi64>,
+  // channel_id = 0
+  channel_handle = #stablehlo.channel_handle<handle = 0, type = 0>
   // use_global_device_ids = false
 } : (tensor<2x2xf32>) -> tensor<2x4xf32>
 // %result@(0, 0): [[1.0, 2.0, 5.0, 6.0], [3.0, 4.0, 7.0, 8.0]]
@@ -729,7 +731,10 @@ Afterwards, within each `process_group`:
     %0 = "stablehlo.add"(%arg0, %arg1) : (tensor<f32>, tensor<f32>) -> tensor<f32>
     "stablehlo.return"(%0) : (tensor<f32>) -> ()
 }) {
-  replica_groups = dense<[[0, 1]]> : tensor<1x2xi64>
+  replica_groups = dense<[[0, 1]]> : tensor<1x2xi64>,
+  // channel_id = 0
+  channel_handle = #stablehlo.channel_handle<handle = 0, type = 0>
+  // use_global_device_ids = false
 } : (tensor<4xf32>) -> tensor<4xf32>
 // %result@(0, 0): [6.0, 8.0, 10.0, 12.0]
 // %result@(1, 0): [6.0, 8.0, 10.0, 12.0]
@@ -1546,18 +1551,22 @@ operations correspond to [stablehlo.minimum](#stablehlominimum) and
 
 ### Semantics
 
-Within each process group in the StableHLO grid, sends and receives data between
-processes in the group.
+Within each process group in the StableHLO grid, sends the value of the
+`operand` tensor from the source process to the target process and produces a
+`result` tensor.
 
 The operation splits the StableHLO grid into `process_groups` as follows:
+
   * `channel_id <= 0`,
     `cross_replica(replica_groups)`.
   * `channel_id > 0`,
     `cross_partition(replica_groups)`.
 
-Afterwards, `result@target` is given by
- * `operand@process_groups[i][0]`, if there exist an `i` such that  `process_groups[i][1] = target`.
- * `result@target = T, otherwise, # where T is a zero valued tensor constant with type(T) = type(operand)`, otherwise.
+Afterwards, `result@process` is given by:
+
+ * `operand@process_groups[i, 0]`, if there exists an `i` such that `process_groups[i, 1] = process`.
+ * `[0, ..., 0]`, otherwise.
+ * `broadcast_in_dim(0, [], shape(result))`, otherwise.
 
 ### Inputs
 
@@ -1576,12 +1585,12 @@ Afterwards, `result@target` is given by
 ### Constraints
 
   * (C1) dim(`source_target_pairs`, 1) $=$ 2.
-  * (C2) All members of { `source_target_pairs`[i][0] : i $\in$ [0, dim(`source_target_pairs`, 0)) } are unique.
-  * (C3) All members of { `source_target_pairs`[i][1] : i $\in$ [0, dim(`source_target_pairs`, 0)) } are unique.
-  * (C4) $0 \gt$ source_target_pairs[i][0], source_target_pairs[i][1] $\lt N$,
+  * (C2) All values in `source_target_pairs[:, 0]` are unique.
+  * (C3) All values in `source_target_pairs[:, 1]` are unique.
+  * (C4) $0 \ge$ source_target_pairs[i][0], source_target_pairs[i][1] $\lt N$,
          where $N$ is given by
-    * `num_replicas`, if `cross_replica`.
-    * `num_partitions`, if `cross_partition`.
+    * If `cross_replica`, `num_replicas`.
+    * If `cross_partition`, `num_partitions`.
   * (C5) type(`result`) $=$ type(`operand`).
 
 ### Examples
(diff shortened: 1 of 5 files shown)


docs/spec_draft.md
```diff
@@ -777,6 +787,137 @@ matrix, then the behavior is undefined.
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.collective_permute
+
+### Semantics
+
+Sends and receives data between source and target replica pairs mentioned in
```
subhankarshah, line 794: rebase with master and simplify.

subhankarshah (changes requested):
Rebase with master.

docs/spec_draft.md
```diff
@@ -1526,6 +1527,64 @@ operations correspond to [stablehlo.minimum](#stablehlominimum) and
...
+processes in the group.
+
+The operation splits the StableHLO grid into `process_groups` as follows:
+  * `channel_id <= 0`,
+    `cross_replica(replica_groups)`.
+  * `channel_id > 0`,
+    `cross_partition(replica_groups)`.
+
+Afterwards, `result@receiver` is given by
+ * `operand@process_groups[i][0]`, if there exist an `i` such that  `process_groups[i][1] = receiver`.
```
subhankarshah, line 1544: instead of using receiver we can use `source` and `target` because that is the terminology of the attribute `source_target_pairs`.

subhankarshah: also can use `source_target_pairs` instead of `process_groups`

docs/spec_draft.md
```diff
@@ -1526,6 +1527,64 @@ operations correspond to [stablehlo.minimum](#stablehlominimum) and
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.collective_permute
+
+### Semantics
+
+Within each process group in the StableHLO grid, sends and receives data between
```
subhankarshah, line 1534: Within each process group in the StableHLO grid, sends data from `source` process to `target` process for all `source_target_pairs`.

subhankarshah: Ok, I now see why we are using the same terminology. But `process_groups` give the impression of bi-directional data transfer.

docs/spec_draft.md
```diff
@@ -1526,6 +1527,64 @@ operations correspond to [stablehlo.minimum](#stablehlominimum) and
...
+processes in the group.
+
+The operation splits the StableHLO grid into `process_groups` as follows:
```
subhankarshah, line 1537: The operation splits the StableHLO grid into `source_target_pairs ` as follows:
* `channel_id <= 0`,
    `source_target_pairs` = `cross_replica(replica_groups)`.
* `channel_id > 0`,
    `source_target_pairs` = `cross_replica(replica_groups)`.

docs/spec_draft.md
```diff
@@ -1526,6 +1527,64 @@ operations correspond to [stablehlo.minimum](#stablehlominimum) and
...
+
+| Name     | Type                         |
+|----------|------------------------------|
+| `result` | tensor of any supported type |
+
+### Constraints
+
+  * (C1) dim(`source_target_pairs`, 1) $=$ 2.
+  * (C2) All memmbers of { `source_target_pairs`[i][0] : i $\in$ [0, dim(`source_target_pairs`, 0)) } are unique.
+  * (C3) All memmbers of { `source_target_pairs`[i][1] : i $\in$ [0, dim(`source_target_pairs`, 0)) } are unique.
```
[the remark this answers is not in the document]
subhankarshah: Reword suggestion.
Every `source` $=$ `source_target_pair[0]` for all `source_target_pair` $in$ `source_target_pairs` is unique.

subhankarshah (changes requested):
@sdasgup3 I have added some comments, please let me know WDYT.

docs/spec_draft.md
sdasgup3: >> instead of using receiver we can use source and target because that is the terminology of the attribute source_target_pairs.

Done.

>> also can use source_target_pairs instead of `process_groups`

`process_groups` is the groups of process pairs formed depending on the group forming strategy `cross_replica` or `cross_partition`. We cannot use the `source_target_pairs` directly as it has replica  or partition ids, but  not the process ids which is what we need to  for the communication.

docs/spec_draft.md
sdasgup3: We are consistently using `process_groups` to denote the result from various group forming strategy. [example](https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md#semantics-3)

docs/spec_draft.md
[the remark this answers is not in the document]
sdasgup3: Nice suggestion. Let us keep it open for the opinion from next reviewer. 

Mine rationals was to be in sync with what we have for other ops like: `All dimensions in `dimensions` are unique.`

docs/spec_draft.md
subhankarshah: sounds fair

docs/spec_draft.md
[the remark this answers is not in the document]
subhankarshah: fair.

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -1250,6 +1250,16 @@ func.func @collective_permute_duplicate_sources(%arg0: tensor<128x32xf32>) -> te
 
 // -----
 
+func.func @collective_permute_duplicate_sources(%arg0: tensor<128x32xf32>) -> tensor<128x32xf32> {
```
subhankarshah, line 1253: name is misleading, how about, permute_invalid_sources

subhankarshah: i see that the other test names are also misleading

docs/spec_draft.md
```diff
@@ -1528,6 +1529,64 @@ operations correspond to [stablehlo.minimum](#stablehlominimum) and
...
+| `result` | tensor of any supported type |
+
+### Constraints
+
+  * (C1) dim(`source_target_pairs`, 1) $=$ 2.
+  * (C2) All members of { `source_target_pairs`[i][0] : i $\in$ [0, dim(`source_target_pairs`, 0)) } are unique.
+  * (C3) All members of { `source_target_pairs`[i][1] : i $\in$ [0, dim(`source_target_pairs`, 0)) } are unique.
+  * (C4) $0 \gt$ source_target_pairs[i][0], source_target_pairs[i][1] $\lt N$,
+         where $N$ is given by
+    * `num_replicas`, if `cross_replica`.
```
subhankarshah, line 1570: I think the second one is used more frequently -
*  `num_replicas`, if `cross_replica`.
*  if `cross_replica`, `num_replicas`.

subhankarshah (approved):
LGTM with minor comments

docs/spec_draft.md
sdasgup3: Thanks! Let us keep the suggestion open for the next reviewer.

docs/spec_draft.md
```diff
@@ -1529,6 +1530,64 @@ operations correspond to [stablehlo.minimum](#stablehlominimum) and
...
+
+| Name     | Type                         |
+|----------|------------------------------|
+| `result` | tensor of any supported type |
+
+### Constraints
+
+  * (C1) dim(`source_target_pairs`, 1) $=$ 2.
+  * (C2) All members of { `source_target_pairs`[i][0] : i $\in$ [0, dim(`source_target_pairs`, 0)) } are unique.
+  * (C3) All members of { `source_target_pairs`[i][1] : i $\in$ [0, dim(`source_target_pairs`, 0)) } are unique.
```
burmako, line 1568: How about "All values in `source_target_pairs[:, 0]` are unique" and the same for C3?

docs/spec_draft.md
burmako: I vote for the second one, similarly to what we wrote in C3 of AllGatherOp.

docs/spec_draft.md
```diff
@@ -1529,6 +1530,64 @@ operations correspond to [stablehlo.minimum](#stablehlominimum) and
...
+  * (C4) $0 \gt$ source_target_pairs[i][0], source_target_pairs[i][1] $\lt N$,
```
burmako, line 1569: This should be $\ge$ not $\gt$.

docs/spec_draft.md
```diff
@@ -1529,6 +1530,64 @@ operations correspond to [stablehlo.minimum](#stablehlominimum) and
...
+  * (C5) type(`result`) $=$ type(`operand`).
+
+### Examples
+
+```mlir
+// num_replicas: 2
+// num_partitions: 1
+// %operand@(0, 0): [[1, 2], [5, 6]]
+%result = "stablehlo.collective_permute"(%operand) {
+  source_target_pairs = dense<[[0, 1]]> : tensor<2x2xi64>,
```
burmako, line 1582: Let's also explicitly mention `channel_id` given that we don't have a notion of optional inputs in the spec.

```
// channel_id = 0
channel_handle = #stablehlo.channel_handle<handle = 0, type = 0>
```

burmako, line 1582: Same comment for AllGatherOp and AllReduceOp whose examples also don't feature `channel_id`, but they should. Sorry - my oversight when reviewing those PRs.

docs/spec_draft.md
```diff
@@ -1529,6 +1530,64 @@ operations correspond to [stablehlo.minimum](#stablehlominimum) and
...
+    * `num_replicas`, if `cross_replica`.
+    * `num_partitions`, if `cross_partition`.
+  * (C5) type(`result`) $=$ type(`operand`).
+
+### Examples
+
+```mlir
+// num_replicas: 2
+// num_partitions: 1
+// %operand@(0, 0): [[1, 2], [5, 6]]
```
burmako, line 1580: We also need to provide some value for `%operand@(1, 0)`.

docs/spec_draft.md
burmako: How about:

```
Within each process group in the StableHLO grid, sends the value of the 
`operand` tensor from the source process to the target process and produces a
`result` tensor.
```

Feedback:
  1. Good point about bidirectional data transfer. That was confusing to me as well.
  2. I think it's okay to use the terminology "source process" and "target process", emphasizing the relationship between these processes and values in `source_target_pairs`, even if these values aren't directly process ids.

docs/spec_draft.md
```diff
@@ -1529,6 +1530,64 @@ operations correspond to [stablehlo.minimum](#stablehlominimum) and
...
+
+The operation splits the StableHLO grid into `process_groups` as follows:
+  * `channel_id <= 0`,
+    `cross_replica(replica_groups)`.
+  * `channel_id > 0`,
+    `cross_partition(replica_groups)`.
+
+Afterwards, `result@target` is given by
+ * `operand@process_groups[i][0]`, if there exist an `i` such that  `process_groups[i][1] = target`.
+ * `result@target = T, otherwise, # where T is a zero valued tensor constant with type(T) = type(operand)`, otherwise.
```
burmako, line 1548: Proposed rewording:

```
Afterwards, `result@process` is given by:
 * `operand@process_groups[i, 0]`, if there exists an `i` such that `process_groups[i, 1] = process`.
 * `[0, ..., 0]`, otherwise.
```

Feedback:
  1. Let's use the `result@process` name, given that not every process is necessarily a target.
  2. Using `process_groups[i, 0]` notation rather than `[i][0]` because `process_groups` is a 2d tensor (so only one level of indexing is needed).
  3. Let's simplify the second bullet. If the proposed notation doesn't sound strict enough because `result` can be multi-dimensional, we can use the `broadcast_in_dim(0, [], shape(result))` notation.

burmako (changes requested):

docs/spec_draft.md
```diff
@@ -1529,6 +1530,64 @@ operations correspond to [stablehlo.minimum](#stablehlominimum) and
...
 [Back to Ops](#index-of-ops)
 
+## stablehlo.collective_permute
+
+### Semantics
+
+Within each process group in the StableHLO grid, sends and receives data between
+processes in the group.
+
+The operation splits the StableHLO grid into `process_groups` as follows:
```
burmako, line 1540: Add newline to follow #729.

docs/spec_draft.md
```diff
@@ -1529,6 +1530,64 @@ operations correspond to [stablehlo.minimum](#stablehlominimum) and
...
+  * `channel_id <= 0`,
+    `cross_replica(replica_groups)`.
+  * `channel_id > 0`,
+    `cross_partition(replica_groups)`.
+
+Afterwards, `result@target` is given by
```
burmako, line 1546: Add newline to follow #729.

burmako (changes requested):

docs/spec_draft.md
sdasgup3: Thanks!

I would prefer `broadcast`.

docs/spec_draft.md
sdasgup3: added the same for `all_gather` and `all_reduce` in this PR and for `reduce_scatter` in https://github.com/openxla/stablehlo/pull/564

docs/spec_draft.md
```diff
@@ -1541,6 +1547,71 @@ operations correspond to [stablehlo.minimum](#stablehlominimum) and
...
+|----------|------------------------------|
+| `result` | tensor of any supported type |
+
+### Constraints
+
+  * (C1) dim(`source_target_pairs`, 1) $=$ 2.
+  * (C2) All values in `source_target_pairs[:, 0]` are unique.
+  * (C3) All values in `source_target_pairs[:, 1]` are unique.
+  * (C4) $0 \ge$ source_target_pairs[i][0], source_target_pairs[i][1] $\lt N$,
+         where $N$ is given by
```
burmako, line 1591: "depends on the process grouping strategy" to be consistent with AllGatherOp. Otherwise, it's unclear what "if cross_replica" means.

docs/spec_draft.md
```diff
@@ -1541,6 +1547,71 @@ operations correspond to [stablehlo.minimum](#stablehlominimum) and
...
+| Name     | Type                         |
+|----------|------------------------------|
+| `result` | tensor of any supported type |
+
+### Constraints
+
+  * (C1) dim(`source_target_pairs`, 1) $=$ 2.
+  * (C2) All values in `source_target_pairs[:, 0]` are unique.
+  * (C3) All values in `source_target_pairs[:, 1]` are unique.
+  * (C4) $0 \ge$ source_target_pairs[i][0], source_target_pairs[i][1] $\lt N$,
```
burmako, line 1590: Hold on, shouldn't this be `\le`?

docs/spec_draft.md
```diff
@@ -1541,6 +1547,71 @@ operations correspond to [stablehlo.minimum](#stablehlominimum) and
...
+
+  * `channel_id <= 0`,
+    `cross_replica(replica_groups)`.
+  * `channel_id > 0`,
+    `cross_partition(replica_groups)`.
+
+Afterwards, `result@process` is given by:
+
+ * `operand@process_groups[i, 0]`, if there exists an `i` such that `process_groups[i, 1] = process`.
+ * `[0, ..., 0]`, otherwise.
```
burmako, line 1568: This bullet is not needed?

burmako (changes requested):
A few minor changes, and we're good to go.

docs/spec_draft.md
sdasgup3: my bad! right.

burmako (approved):

PR #569, opened 2022-11-17
Change -1 to ? printing when possible

GleasonK:
Prepare for dynamic dim size to be changed from -1 to INTMIN.

This is a first step at removing the reliance on upstream `kDynamicSize`. There is more refactoring that can be done, but this change will at least make it so that some of our error messages will display `?` instead of the value, and our infer tests will continue to display `-1` regardless of what the underlying dynamic size is.

### DIFF
--- a/stablehlo/dialect/AssemblyFormat.cpp
+++ b/stablehlo/dialect/AssemblyFormat.cpp
@@ -14,6 +14,7 @@ limitations under the License.
 ==============================================================================*/
 
 #include "stablehlo/dialect/AssemblyFormat.h"
+
 #include <cstdint>
 #include <string>
 
@@ -327,7 +328,8 @@ FailureOr<SmallVector<int64_t>> parseDimensionSizes(AsmParser& parser) {
     }
     return parser.parseInteger(dims.emplace_back());
   };
-  if (failed(parser.parseCommaSeparatedList(AsmParser::Delimiter::Square, parseElt))) {
+  if (failed(parser.parseCommaSeparatedList(AsmParser::Delimiter::Square,
+                                            parseElt))) {
     return failure();
   }
   return dims;
--- a/stablehlo/dialect/AssemblyFormat.cpp
+++ b/stablehlo/dialect/AssemblyFormat.cpp
@@ -305,43 +305,43 @@ ParseResult parseDenseI64Array(OpAsmParser& parser,
   return success();
 }
 
-std::string dimensionToString(int64_t dim) {
+std::string dimSizeToString(int64_t dim) {
   if (hlo::isDynamicDimSize(dim)) {
     return "?";
   }
   return std::to_string(dim);
 }
 
-void printDimensionSizes(AsmPrinter& p, llvm::ArrayRef<int64_t> dims) {
+void printDimSizes(AsmPrinter& p, llvm::ArrayRef<int64_t> dimSizes) {
   p << '[';
-  llvm::interleaveComma(dims, p,
-                        [&p](int64_t dim) { p << dimensionToString(dim); });
+  llvm::interleaveComma(
+      dimSizes, p, [&p](int64_t dimSize) { p << dimSizeToString(dimSize); });
   p << ']';
 }
 
-FailureOr<SmallVector<int64_t>> parseDimensionSizes(AsmParser& parser) {
-  SmallVector<int64_t> dims;
+FailureOr<SmallVector<int64_t>> parseDimSizes(AsmParser& parser) {
+  SmallVector<int64_t> dimSizes;
   auto parseElt = [&]() -> ParseResult {
     if (!parser.parseOptionalQuestion()) {
-      dims.push_back(ShapedType::kDynamic);
+      dimSizes.push_back(ShapedType::kDynamic);
       return success();
     }
-    return parser.parseInteger(dims.emplace_back());
+    return parser.parseInteger(dimSizes.emplace_back());
   };
   if (failed(parser.parseCommaSeparatedList(AsmParser::Delimiter::Square,
                                             parseElt))) {
     return failure();
   }
-  return dims;
+  return dimSizes;
 }
 
-ParseResult parseDimensionSizes(AsmParser& parser,
-                                FailureOr<SmallVector<int64_t>>& dims) {
-  auto failOrDims = parseDimensionSizes(parser);
-  if (failed(failOrDims)) {
+ParseResult parseDimSizes(AsmParser& parser,
+                          FailureOr<SmallVector<int64_t>>& dimSizes) {
+  auto failOrDimSizes = parseDimSizes(parser);
+  if (failed(failOrDimSizes)) {
     return failure();
   }
-  dims = std::move(*failOrDims);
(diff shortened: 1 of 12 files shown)


stablehlo/integrations/python/tests/stablehlo.py
```diff
@@ -20,7 +20,6 @@
 from mlir import ir
 from mlir.dialects import stablehlo
 
-
```
burmako, line 23: Typo: need two lines separating imports from functions and functions from each other.

stablehlo/integrations/python/tests/stablehlo.py
```diff
@@ -196,6 +195,8 @@ def test_token_type():
 
 @run
 def test_type_extensions():
-  attr = stablehlo.TypeExtensions.get(bounds=[128, -1])
+
```
burmako, line 198: Typo: extraneous newline.

stablehlo/tests/TestUtils.cpp
```diff
@@ -98,8 +109,9 @@ struct InferReturnTypeComponentsPattern : public RewritePattern {
     auto *newOp = rewriter.create(state);
     for (const auto &it : llvm::enumerate(components)) {
       if (it.value().hasRank()) {
-        newOp->setAttr((StringRef("dims") + Twine(it.index())).str(),
-                       rewriter.getI64ArrayAttr(it.value().getDims()));
+        newOp->setAttr(
+            (StringRef("dims") + Twine(it.index())).str(),
+            rewriter.getI64ArrayAttr(fixDynamicDims(it.value().getDims())));
```
burmako, line 114: I think we should switch from -1 to ? here as well. How about `rewriter.getStringAttr(dimSizesToString(it.value().getDims()))`?

stablehlo/dialect/AssemblyFormat.h
```diff
@@ -218,7 +222,6 @@ ParseResult parseCustomCallTarget(AsmParser& parser, StringAttr& target);
 //     [1, 2]
 void printIntArray(AsmPrinter& printer, ArrayRef<int64_t> ints);
```
burmako, line 223: `printIntArray` is unused - let's delete?

stablehlo/dialect/AssemblyFormat.h
```diff
@@ -218,7 +222,6 @@ ParseResult parseCustomCallTarget(AsmParser& parser, StringAttr& target);
...
 
-FailureOr<SmallVector<int64_t>> parseIntArray(AsmParser& parser);
 ParseResult parseIntArray(AsmParser& parser, SmallVector<int64_t>& ints);
```
burmako, line 225: `parseIntArray` is only used in `parseDims` within StablehloOps.cpp. Given that, I think we should delete `parseIntArray` from `AssemblyFormat.h`.

As far as what happens with the implementation of `parseIntArray`, I think that we have two options:
  1) Inline the implementation into `parseDims`.
  2) Delete `parseDims` and replace its usages with `parseDimSizes`.

I think that 1) is more consistent with this comment in #327: https://github.com/openxla/stablehlo/issues/327#issuecomment-1319353977, i.e. I'm supportive of deferring the cleanup of `parseDims` to future PRs. I think that we should delete `parseIntArray` right away though.

burmako (changes requested):

stablehlo/tests/TestUtils.cpp
GleasonK: We could, but that would print as `dim0 = "[0, ?]"` which is also a bit inconsistent with quotes.

The other alternative is registering this op to control printing, or creating a new attr type for dim size that has a custom printer/parser to use for this type. I don't think we should register test ops unless we make a whole new test dialect, and we currently use ArrayRefParam for dim sizes which also feels correct. So I could go either way for this one between -1 and the string above.

stablehlo/dialect/AssemblyFormat.h
GleasonK: There was no fallout from (2) in StableHLO so let's make that change now. If there is any fallout from MHLO (not sure there will be, looks like a lot of the uses of parseDim should only specify positive dim numbers), we can inline into a parseDims implementation there and clean up later.

stablehlo/dialect/AssemblyFormat.cpp
```diff
@@ -301,35 +305,43 @@ ParseResult parseDenseI64Array(OpAsmParser& parser,
...
 }
 
-void printDimensionSizes(AsmPrinter& p, llvm::ArrayRef<int64_t> shape) {
-  auto printIntOrQuestion = [&](int64_t value) {
-    if (ShapedType::isDynamic(value))
-      p << '?';
-    else
-      p << value;
-  };
+std::string dimSizeToString(int64_t dim) {
```
burmako, line 308: dimSize

stablehlo/dialect/AssemblyFormat.h
```diff
@@ -172,16 +173,19 @@ void printDenseI64Array(OpAsmPrinter& p, Operation* op,
...
 
-// DimensionSizes - Print an array of ints. Dynamic dimensions printed as `?`.
+// DimSizes - Print an array of ints. Dynamic dimensions printed as `?`.
 //
 //   Generic:
 //     [1, -1]
 //   Custom:
 //     [1, ?]
-void printDimensionSizes(AsmPrinter& p, llvm::ArrayRef<int64_t> shape);
+std::string dimSizeToString(int64_t dim);
```
burmako, line 182: dimSize

stablehlo/dialect/AssemblyFormat.h
```diff
@@ -172,16 +173,19 @@ void printDenseI64Array(OpAsmPrinter& p, Operation* op,
...
 
-ParseResult parseDimensionSizes(AsmParser& parser,
-                                FailureOr<SmallVector<int64_t>>& shape);
+void printDimSizes(AsmPrinter& p, llvm::ArrayRef<int64_t> dims);
```
burmako, line 184: dimSizes

stablehlo/dialect/AssemblyFormat.h
```diff
@@ -172,16 +173,19 @@ void printDenseI64Array(OpAsmPrinter& p, Operation* op,
...
+
+FailureOr<SmallVector<int64_t>> parseDimSizes(AsmParser& parser);
+ParseResult parseDimSizes(AsmParser& parser,
+                          FailureOr<SmallVector<int64_t>>& dims);
```
burmako, line 188: dimSizes

stablehlo/tests/TestUtils.cpp
burmako: "I don't think we should register test ops unless we make a whole new test dialect". Agreed.

Between keeping -1 and accepting quotes, I think quotes are preferable, because the MLIR community has really been trying to get away from -1, and the mission has basically succeeded.

burmako (changes requested):
Thank you for thoroughly addressing the naming comments! I found just a couple more `dims`.

burmako (approved):

PR #575, opened 2022-11-21
Add spec for OptimizationBarrierOp

sdasgup3:


docs/spec_draft.md
```diff
@@ -1940,6 +1941,40 @@ produces a `result` tensor. For boolean tensors, it computes the logical NOT.
...
 
+
+## stablehlo.optimization_barrier
+
+### Semantics
+
+Blocks any optimization pass from moving computations across the barrier. It
+ensures that the operations producing the `operand` are executed before any
+operations that depend on `result`.
+
```
subhankarshah, line 1952: Ensures that the operations that produce the `operand` are executed before any
operations that depend on the `result`, preventing any optimisation pass from
moving computations across the barrier.

docs/spec_draft.md
```diff
@@ -1940,6 +1941,40 @@ produces a `result` tensor. For boolean tensors, it computes the logical NOT.
...
+| Name     | Type                                                       |
+|----------|------------------------------------------------------------|
+| `result` | variadic number of tensors of any supported type or tokens |
+
+### Constraints
+
+  * (C1) size(`operand`) $=$ size(`result`).
+  * (C2) type(`operand`) $=$ type(`result`).
+
```
subhankarshah, line 1969: Why cant we write, operand = result

subhankarshah (changes requested):
Added a couple of minor comments.

docs/spec_draft.md
sdasgup3: `size` or `type` denote some of the static properties of these operands. `operand = result` might be confusing for someone to think that they have same values

docs/spec_draft.md
sdasgup3: Thanks for the suggestion! I was also trying to merge the two sentence like above but following is what bothers me:

It gives the impression that ensuring `operations that produce the operand are executed before any operations that depend on the result` is sufficient to `preventing any optimization pass from ....`. 

IMO this op is also a directive to the compiler  to  prevent any optimization which needs moving computations around the barriers even if those computations are not producers or consumers of the `optimization_barrier` op.  

Comparing with  [AfterAll](https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md#stablehloafter_all) which has similar wordings but, IMO, it can allow reordering computations across the op if those computations are not producers or consumers of the `afterAll` op. 

how about using
```
Ensures that the operations that produce the `operand` are executed before any
operations that depend on the `result` and prevents any optimization pass from
moving computations across the barrier.
```

sdasgup3: **upd** after some exploration: 

`AfterAll`, with input /output being tokens, can only be used to introduce data dependency between side-effecting operations (infeed/outfeed/send//recv). Where as `optimization_barrier` can be used to introduce the same between any (sideeffecting or not) operations. Other than that the two operations are the same. 

IMO, the line `preventing any optimization pass from moving computations across the barrier.` specifically means the optimizations like CSE that involves non-side-efecting operations and hence the same line in not present in AfterAll case.

Also, I believe both the operations can move operations across them of they are not the consumers or producers of the op.

Having said that I am tempted to adopt your suggestion with a slight modification:
```
Ensures that the operations that produce the `operand` are executed before any
operations that depend on the `result`, preventing any optimisation pass from
moving computations, which are producers or consumers of the `optimization_barrier` operation, across the barrier.
```

subhankarshah: `Also, I believe both the operations allow the opt passes to move operations across them if those operations are not the consumers or producers of the op. And I think introducing this detail whould complicate the spec.`

In this case `prevents any optimisation pass from moving computations across the barrier` is no longer true for either op.

docs/spec_draft.md
subhankarshah: we can just say `for all i` and remove the $\in$ part.

subhankarshah (approved):
LGTM, with few comments, feel free to make changes that you feel best and keep them open for the next reviewer.

docs/spec_draft.md
sdasgup3: I agree that this part of the sentence is "loose" in its intent.  Maybe refactoring it to the following will help.
... 
prevents any optimization pass from moving computations, which are either the producers or consumers of the optimization_barrier op, across the barrier. 
```

Let us seek opinion from the next reviewer.

File: stablehlo/dialect/TypeInference.h [hunk 3]
```diff
@@ -133,6 +133,9 @@ LogicalResult inferPadOp(Optional<Location> location, Value operand,
                          DenseIntElementsAttr interiorPadding,
                          SmallVectorImpl<Type>& inferredReturnTypes);
 
+LogicalResult inferOptimizationBarrierOp(
+    ValueRange inputs, SmallVectorImpl<Type>& inferredReturnTypes);
```
burmako, line 137: Probably should be called `operand`, to match the ODS?

docs/spec_draft.md
```diff
@@ -2555,6 +2556,42 @@ produces a `result` tensor. For boolean tensors, it computes the logical NOT.
...
+### Constraints
+
+  * (C1) size(`operand`) $=$ size(`result`).
+  * (C2) type(`operand[i]`) $=$ type(`result[i]`) for all i.
+
+### Examples
+
+```mlir
+%result:2 = "stablehlo.optimization_barrier"(%arg0, %arg1) : (tensor<1xf32>, tensor<1xi32>) -> (tensor<1xf32>, tensor<1xi32>)
```
burmako, line 2590:
```
// %operand0: 0.0
// %operand1: 1.0
%result0, %result1 = "stablehlo.optimization_barrier"(%operand0, %operand1) : (tensor<f32>, tensor<f32>) -> (tensor<f32>, tensor<f32>)
// %result0: 0.0
// %result1: 1.0
```

burmako: Or feel free to keep using `%result` - I don't have a preference either way.

stablehlo/dialect/StablehloOps.td
```diff
@@ -2744,25 +2744,25 @@ def StableHLO_TorchIndexSelectOp : StableHLO_Op<"torch_index_select", [Pure]> {
...
+    pass from moving computations across the barrier.
 
-    Ensures that all inputs are evaluated before any operators that depend on the barrier's outputs.
     See
-    https://www.tensorflow.org/xla/operation_semantics#optimizationbarrier
+    https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md#stablehlooptimization_barrier
+
+    Example:
+    ```mlir
+    %result:2 = stablehlo.optimization_barrier %arg0, %arg1 : tensor<1xf32>, tensor<1xi32>
```
burmako, line 2764: Let's sync with the syntax from the Examples section.

burmako (changes requested):

burmako (approved):

Issue #484, opened 2022-11-08
Formalize syntax and evaluation of StableHLO programs


burmako:
We've made significant progress with [the spec](https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md) in the last two months, and it looks like soon we will have all of the core ops specced after the existing HLO semantics. There's a bunch of ops in the PR queue right now, and about a dozen more distribution and elementwise ops to go, but finishing all that looks like a matter of several weeks at this point thanks to the great work done by the team:

| Category      | Ops             |
|:--------------|:----------------|
| Control Flow  | case, if, while |
| Data Movement | broadcast_in_dim, concatenate, gather, pad, reshape, reverse, scatter, slice, sort, transpose |
| Distribution  | after_all, all_gather, all_reduce, all_to_all, collective_permute, infeed, outfeed, recv, reduce_scatter, replica_id, send |
| Elementwise   | abs, add, and, atan2, bitcast_convert, cbrt, ceil, clamp, compare, complex, convert, cosine, count_leading_zeros, divide, exponential, exponential_minus_one, floor, imag, is_finite, log, log_plus_one, logistic, map, maximum, minimum, multiply, negate, not, or, popcnt, power, real, reduce_precision, remainder, round_nearest_afz, round_nearest_even, rsqrt, select, shift_left, shift_right_arithmetic, shift_right_logical, sign, sine, sqrt, subtract, tanh, xor |
| Extensibility | custom_call, get_tuple_element, tuple |
| Miscellaneous | batch_norm_grad, batch_norm_inference, batch_norm_training, cholesky, constant, fft, iota, optimization_barrier, return, rng, rng_bit_generator, triangular_solve |
| Reduction     | convolution, dot_general, reduce, reduce_window, select_and_scatter |

This will make it possible to do something cool that wasn't possible before - express RFCs as diffs to the spec - meaning that the work on evolution beyond what's currently in HLO would become more precise:

| Category      | Ops             |
|:--------------|:----------------|
| Dynamism      | compute_reshape_shape, cstr_reshapable, dynamic_broadcast_in_dim, dynamic_conv, dynamic_gather, dynamic_iota, dynamic_pad, dynamic_reshape, dynamic_slice, dynamic_update_slice, get_dimension_size, real_dynamic_slice, set_dimension_size |
| Quantization  | uniform_dequantize, uniform_quantize + &lt;new element types&gt; |
| Sparsity      | &lt;new tensor types&gt; |

[...]

PR #576, opened 2022-11-21
Add spec for ReplicaIdOp

### LINKED ISSUE #529: Add spec for ReplicaIdOp
### Request description

To spec `stablehlo.replica_id`

sdasgup3:
closes #529

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -2866,20 +2866,13 @@ implementation-defined value.
 
 ### Semantics
 
-Returns the unique id of the replica in the interval [0, N), where N is the
-number of replicas. Since all the replicas are running the same program, a
-`stablehlo.replica_id` call in the program will return a different value on each
-replica.
+Produces `replica_id` of the current process.
 
 ### Outputs
 
-| Name     | Type                  |
-|----------|-----------------------|
-| `result` | tensor of type `ui32` |
-
-### Constraints
-
-  * (C1) rank(`result`) $=$ 0.
+| Name     | Type                                |
+|----------|-------------------------------------|
+| `result` | 0-dimensional tensor of type `ui32` |
 
 ### Examples
 
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -377,6 +377,7 @@ syntax.
    * [real](#stablehloreal)
    * [reduce](#stablehloreduce)
    * [remainder](#stablehloremainder)
+   * [replica_id](#stablehloreplica_id)
    * [reshape](#stablehloreshape)
    * [reverse](#stablehloreverse)
    * [rng](#stablehlorng)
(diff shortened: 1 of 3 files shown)


docs/spec_draft.md
```diff
@@ -2215,6 +2216,33 @@ implementation-defined value.
...
+Returns the unique id of the replica in the interval [0, N), where N is the
+number of replicas. Since all the replicas are running the same program, a
+`stablehlo.replica_id` call in the program will return a different value on each
+replica.
+
+### Outputs
+
+| Name     | Type                  |
+|----------|-----------------------|
+| `result` | tensor of type `ui32` |
```
subhankarshah, line 2232: we can add the rank zero part here.
a 0  dimensional tensor
or a tensor constant (not sure it should be called a constant)

docs/spec_draft.md
```diff
@@ -2215,6 +2216,33 @@ implementation-defined value.
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.replica_id
+
+### Semantics
+
+Returns the unique id of the replica in the interval [0, N), where N is the
```
subhankarshah, line 2223: can we link the parallel execution semantics for users to understand what a replica is

docs/spec_draft.md
```diff
@@ -2215,6 +2216,33 @@ implementation-defined value.
...
+number of replicas. Since all the replicas are running the same program, a
+`stablehlo.replica_id` call in the program will return a different value on each
+replica.
+
```
subhankarshah, line 2227: Returns `replica_id` of a replica[#replicaexplanation_link]. Since all replicas run the same program, `stablehlo.replica_id` produces a different value on each replica.

We can omit the second sentence because across partitions replica_id might be same, also all of that should be clear from the parallel execution semantics.

docs/spec_draft.md
subhankarshah: We can then remove the constraints section altogether.

subhankarshah (changes requested):
Added few comments

docs/spec_draft.md
burmako: I would propose "Produces \`replica_id\` of the current process", because `replica_id` is already well-defined above, so there's no need to go into details of what a replica is or what is the range of what this op returns.

Agreed about explicitly referring to `replica_id` and about dropping the second sentence, given the discussion in #634. 

Re: linking to an explanation of what a replica_id is, I would say that this is not needed. E.g. we also don't explain what a token type, we just use it as needed.

docs/spec_draft.md
burmako: +1

docs/spec_draft.md
```diff
@@ -2215,6 +2216,33 @@ implementation-defined value.
...
+| `result` | tensor of type `ui32` |
+
+### Constraints
+
+  * (C1) rank(`result`) $=$ 0.
+
+### Examples
+
+```mlir
+%result = stablehlo.replica_id : tensor<ui32>
```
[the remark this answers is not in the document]
burmako: Good question! I would imagine that the op will return `0`, but I also see that [__xla_cpu_runtime_ReplicaId](https://github.com/tensorflow/tensorflow/blob/f511aa077646c1028c819090064945e6f109ff94/tensorflow/compiler/xla/service/cpu/cpu_runtime.cc#L932) currently crashes when I try to run an HLO program that only includes a replica_id op. 

Added this question to #484 as something to figure out. As far as this spec goes, this op can just say that it returns `replica_id`.

burmako (changes requested):

docs/spec_draft.md
sdasgup3: IMO we can skip that given that we have a section dedicated to it in the beginning. This is the same treatment we did to token.

burmako (changes requested):

docs/spec_draft.md
```diff
@@ -2861,6 +2862,26 @@ implementation-defined value.
...
+### Outputs
+
+| Name     | Type                                |
+|----------|-------------------------------------|
+| `result` | 0-dimensional tensor of type `ui32` |
+
+### Examples
+
+```mlir
+%result = stablehlo.replica_id : tensor<ui32>
```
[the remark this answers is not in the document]
sdasgup3: missed it. Thanks!

burmako (approved):

subhankarshah (approved):
LGTM

PR #578, opened 2022-11-22
Fix typos in step names of GitHub workflows

burmako:
Unless I'm missing something, these look like typos.


bhack (approved):

GleasonK (approved):
LGTM. These steps used to say the OS running, but since we only test Linux currently, I deleted that part. It looks like that cleanup missed some deletions. Thanks for fixing.

PR #503, opened 2022-11-10
Add spec for AllGatherOp

sdasgup3:
fixes #462 

Address the followings:
1. Adds verification checks for AllGather w.r.t. https://github.com/openxla/stablehlo/issues/498
2. fixes https://github.com/openxla/stablehlo/issues/491

A few points
  - Type Inference is marked `infeasible` as the return type of the op depends upon the [shard_count](https://github.com/tensorflow/tensorflow/blob/20c6943d3cd7e07da162f7778a0af5d3776274b4/tensorflow/compiler/xla/service/hlo_verifier.cc#L452) which [depends on the result type](https://github.com/tensorflow/tensorflow/blob/20c6943d3cd7e07da162f7778a0af5d3776274b4/tensorflow/compiler/xla/service/hlo_verifier.cc#L426). Note that the `shard_count` is a parameter in HLO spec, where as in MHLO it is [derived](https://github.com/tensorflow/tensorflow/blob/a1acd6a6466f58ed4197b7beaa2f3e0b6fcfc32a/tensorflow/compiler/xla/translate/mhlo_to_hlo/mlir_hlo_to_hlo.cc#L766) using result type before exporting to HLO.
  - With (1) and (2), the Verifier is a `yes`.  Note that we still do not have the [check](https://github.com/tensorflow/tensorflow/blob/20c6943d3cd7e07da162f7778a0af5d3776274b4/tensorflow/compiler/xla/service/hlo_verifier.cc#L436) for  `shard_count == subgroup_size`. The `subgroup_size` depends on [module configuration](https://github.com/tensorflow/tensorflow/blob/20c6943d3cd7e07da162f7778a0af5d3776274b4/tensorflow/compiler/xla/service/hlo_verifier.cc#L95)  which manages the settings and values which affect the compiled executable outside of the HLO code itself and I am not sure if that info is available in StableHLO IR.

upd:
 - With https://github.com/openxla/stablehlo/issues/650, the type inference should be made feasible. Marked it `revisit` until that is fixed. 
 - The verifier should be `revisit` based on https://github.com/openxla/stablehlo/issues/652

PR #634, opened 2022-11-26
Add spec for parallel execution semantics

burmako:
I think that recent PRs with specs for collective ops (e.g. #503) would benefit from a more formal definition of parallel execution semantics of StableHLO programs.

That will provide additional clarity which is needed at the moment (see discussions in the PRs) and reduce the duplication between the specs of individual collective ops.

The description of execution semantics isn't yet as formal as I'd like, but it can already benefit specs for collective ops. I'll leave the rest to follow-up work within #484.

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -213,7 +213,7 @@ lists of replica ids - and computes a Cartesian product of `replica_groups` by
 `replica_ids`. More formally:
 
 ```Python
-def cross_replica(replica_groups: List[List[ui32]]) -> List[List[Tuple[ui32, ui32]]]:
+def cross_replica(replica_groups: List[List[ReplicaId]]) -> List[List[ProcessId]]:
   for replica_group in replica_groups:
     for partition_id in partition_ids:
       process_group = []
@@ -233,7 +233,7 @@ a list of lists of partition ids - and computes a Cartesian product of
 elements and cover all `partition_ids`. More formally:
 
 ```Python
-def cross_partition(partition_groups: List[List[ui32]]) -> List[List[Tuple[ui32, ui32]]]:
+def cross_partition(partition_groups: List[List[PartitionId]]) -> List[List[ProcessId]]:
   for partition_group in partition_groups:
     for replica_id in replica_ids:
       process_group = []
@@ -254,7 +254,7 @@ computes Cartesian products of each `replica_group` by `partition_ids`.
 More formally:
 
 ```Python
-def cross_replica_and_partition(replica_groups: List[List[ui32]]) -> List[List[Tuple[ui32, ui32]]]:
+def cross_replica_and_partition(replica_groups: List[List[ReplicaId]]) -> List[List[ProcessId]]:
   for replica_group in replica_groups:
     process_group = []
     for partition_id in partition_ids:
@@ -274,7 +274,7 @@ process ids. `flattened_id_groups` must have unique elements and cover all
 `process_ids`. More formally:
 
 ```Python
-def flattened_ids(flattened_id_groups: List[List[ui32]]) -> List[List[Tuple[ui32, ui32]]]:
+def flattened_ids(flattened_id_groups: List[List[ui32]]) -> List[List[ProcessId]]:
   for flattened_id_group in flattened_id_groups:
     process_group = []
     for flattened_id in flattened_id_group:
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -102,26 +102,187 @@ never produced as outputs.
 
 ## Programs
 
-StableHLO programs consist of functions. Each function has inputs and outputs
-of supported types and a list of ops in static single-assignment (SSA) form
-which is terminated by a return op which produces the outputs of the function.
-StableHLO ops take inputs and produce outputs.
+**StableHLO programs** consist of **StableHLO functions**. Each function has
+inputs and outputs of supported types and a list of ops in static
+single-assignment (SSA) form which is terminated by a `return` op which produces
+the outputs of the function.
 
-```mlir
-ml_program.func @example_func(%arg: tensor<2x3xf32>) -> tensor<2x3xf32> {
- %0 = "stablehlo.floor"(%arg) : (tensor<2x3xf32>) -> tensor<2x3xf32>
- %1 = "stablehlo.ceil"(%arg) : (tensor<2x3xf32>) -> tensor<2x3xf32>
- %2 = "stablehlo.add"(%0, %1) : (tensor<2x3xf32>, tensor<2x3xf32>) -> tensor<2x3xf32>
- ml_program.return %2 : tensor<2x3xf32>
+Here is an example of a program that consists of a function `@main` which takes
+three inputs (`%image`, `%weights` and `%bias`) and produces one output (`%4`).
+Below we describe how this program can be executed.
+
+stablehlo.func @main(
+  %image: tensor<28x28xf32>,
+  %weights: tensor<784x10xf32>,
+  %bias: tensor<1x10xf32>
+) -> tensor<1x10xf32> {
+  %0 = "stablehlo.reshape"(%image) : (tensor<28x28xf32>) -> tensor<1x784xf32>
+  %1 = "stablehlo.dot"(%0, %weights) : (tensor<1x784xf32>, tensor<784x10xf32>) -> tensor<1x10xf32>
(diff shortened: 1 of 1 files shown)


docs/spec_draft.md
```diff
@@ -102,26 +102,191 @@ never produced as outputs.
...
+1) a StableHLO program, 2) operation statuses (not executed yet,
+already executed), and 3) intermediate values that the process is working on.
+The process starts with input values to the `main` function, progresses through
+the graph of ops updating operation statuses and intermediate values and
+finishes with output values. Further formalization is TBD.
+
+### Parallel execution
+
+StableHLO programs can be executed in parallel, organized into a 2D grid of
+`num_replicas` by `num_partitions` which both have type `ui32`.
```
fschlimb, line 152: I still find the terminology confusing. When I think of the term "replica" I have something in mind which is a copy or a clone; here they seem to denote "siblings", they do *not* replicate the data. Actually, in my mind the right term for these would be "partition" because the divide the data in parts (e.g. they partition the data). What's called a "partition" here is actually more like a module or sub-program. 

Probably part of the confusion stems from different background:  I know it from data-partitioning (e.g. like what METIS does); the term "partition" might be used to denote parts of programs elsewhere.

docs/spec_draft.md
```diff
@@ -102,26 +102,191 @@ never produced as outputs.
...
+`process_id = (replica_id, partition_id)`, where
+`replica_id ∊ replica_ids = [0, ..., num_replicas-1]` and
+`partition_id ∊ partition_ids = [0, ..., num_partitions-1]` which both have
+type `ui32`. The size of the grid is known statically for every program, and
+the position within the grid is known statically for every process.
+
+Within the grid, the programs can all be the same (in the "Single Program,
+Multiple Data" style), can all be different (in the "Multiple Program, Multiple
+Data" style) or something in between. Each program has access to its position
+within the grid via the `replica_id` and `partition_id` ops.
```
fschlimb, line 165: > Within the grid, the programs can all be the same...
> ...
> Each program has access to its position within the grid via the `replica_id` and `partition_id` ops.

The use of "program" is a bit confusing. Would this sbe clearer?
"
Within the grid, different processes can execute programs which are all the same...
...
Each *process* has access to its position....
"

docs/spec_draft.md
```diff
@@ -102,26 +102,191 @@ never produced as outputs.
...
+the processes in the StableHLO grid into **StableHLO process groups** and
+execute a joint computation within each process group, independently from other
+process groups.
+
+For all processes within each process group, collective ops introduce a
+synchronization barrier, i.e. the execution of a collective op in all these
+processes doesn't start until the execution of all these processes reaches
+the op. Further formalization, e.g. elaborating on what it means for execution
+to reach an op and what happens to a process group if it doesn't happen, is TBD.
+
```
fschlimb, line 195: Why is this synchronizing the start of the collective operation and not at its end? Why define execution semantics? I'd leave it to something simple like
"
Collective operations constitute potentially blocking synchronization points among processes within a process group.
"

docs/spec_draft.md
```diff
@@ -102,26 +102,191 @@ never produced as outputs.
...
+StableHLO programs can be executed in parallel, organized into a 2D grid of
+`num_replicas` by `num_partitions` which both have type `ui32`.
+
+In the **StableHLO grid**, `num_replicas * num_partitions` of StableHLO
+processes are executing at the same time. Each process has a unique
+`process_id = (replica_id, partition_id)`, where
+`replica_id ∊ replica_ids = [0, ..., num_replicas-1]` and
+`partition_id ∊ partition_ids = [0, ..., num_partitions-1]` which both have
+type `ui32`. The size of the grid is known statically for every program, and
```
sdasgup3, line 159: I am not sure why we use "..... for every program"
Isn't the size is the same for each participating programs. 
Or do you mean the sub-grid  which is going to use the same program exhibiting SPMD style parallelism.

docs/spec_draft.md
burmako: Eventually, I'd still like to formalize the semantics to the degree that the producers and consumers find reasonable. The current state of HLO semantics is underdocumented, which makes it hard to reason about it, and I think it's worth investing in more clarity.

This doesn't mean that we are planning to force standardization where implementation-defined behavior makes sense. Overall, the idea is to start with a baseline spec for HLO behavior and then incrementally refine that together as a community.

Given that, I agree about leaving off any kind of details of synchronization for now. I've updated the text along the lines of what you proposed.

sdasgup3 (approved):

docs/spec_draft.md
```diff
@@ -102,26 +102,192 @@ never produced as outputs.
...
-implementation-defined, as long as ops are evaluated before their uses. Possible
-execution orders of the above example program are `%0` → `%1` → `%2` → `return`
-or `%1` → `%0` → `%2` → `return`.
+## Execution
+
+### Sequential execution
+
+A StableHLO program is executed by providing input values to the `main` function
+and computing output values. Output values of a function are computed by
+executing the graph of ops rooted in the corresponding `return` op.
```
subhankarshah, line 135: Does this mean if the function does not return anything, even if it is invoked nothing would execute?

docs/spec_draft.md
```diff
@@ -102,26 +102,192 @@ never produced as outputs.
...
+
+The execution order is implementation-defined, as long as ops are executed
```
subhankarshah, line 137: Reword suggestion -
The execution follows topological ordering, but the exact order is implementation-defined.

docs/spec_draft.md
subhankarshah: We can explain the idea of partition and replica or draw a parallel, if we are not providing a formal definition.

docs/spec_draft.md
```diff
@@ -102,26 +102,192 @@ never produced as outputs.
...
+`partition_ids`. `replica_groups` must have unique elements and cover all
+`replica_ids`. More formally:
+
+```Python
+def cross_replica(replica_groups: List[List[ReplicaId]]) -> List[List[ProcessId]]:
+  for replica_group in replica_groups:
+    for partition_id in partition_ids:
+      process_group = []
+      for replica_id in replica_group:
+        process_group.append((replica_id, partition_id))
```
subhankarshah, line 222:
```python
process_group = [(replica_id, partition_id) for replica_id in replica_group]
```

docs/spec_draft.md
```diff
@@ -102,26 +102,192 @@ never produced as outputs.
...
+      yield process_group
```
subhankarshah, line 223: I am sorry, I did not understand the significance of `yield` here.

docs/spec_draft.md
```diff
@@ -102,26 +102,192 @@ never produced as outputs.
...
+      for replica_id in replica_group:
+        process_group.append((replica_id, partition_id))
+    yield process_group
+```
+
+For example, for `replica_groups = [[0, 1], [2, 3]]` and `num_partitions = 2`,
+`cross_replica_and_partition` will produce
+`[[(0, 0), (1, 0), (0, 1), (1, 1)], [(2, 0), (3, 0), (2, 1), (3, 1)]]`.
+
+4) **flattened_ids**. This strategy takes `flattened_id_groups` - a list of
```
subhankarshah, line 271: Do we need to specify this (possibly)allows both cross-replica and cross-partition communication.

docs/spec_draft.md
subhankarshah: If we add small sections for explaining replica (with replica_id etc) and partition (with partition_id etc) we can link to them from other parts of the spec.

docs/spec_draft.md
burmako: Thanks! That's a good suggestion, but it only applies well for the first two strategies - for the last two I think the code reads better with for loops. I'll keep the existing style for consistency between all strategies.

docs/spec_draft.md
burmako: I used it as a way to cut down on boilerplate - instead of defining `process_groups = []` at the top of the function, appending to it here and then returning it at the bottom of the function, we can just yield.

docs/spec_draft.md
burmako: Given the overall spirit of minimizing what we specify, I think we shouldn't call this out here.

By the way, your comment gave me motivation to review the descriptions of the previous three strategies, and I found a mistake. In "both cross-replica and cross-partition communications will be happening within each process group", "will" should really be "may".

burmako:
Merging given an approval for this PR and the absence of requests for changes. If I misread the room, and some of this feedback was blocking, please let me know, and I'll address in a follow-up PR.

Issue #639, opened 2022-11-28
Verify layout in InfeedOp


sdasgup3:
Missing the following check:

The size of layout entry for each operand should match the rank of the operand.

https://github.com/tensorflow/tensorflow/blob/094de67f996da201f89ceb33416ce0449b271c78/tensorflow/compiler/xla/translate/mhlo_to_hlo/mlir_hlo_to_hlo.cc#L2316

https://github.com/openxla/stablehlo/issues/629 is related as that will decide if we should support layout in StableHLO and how to model them.

sdasgup3:
FYI:
The following checks are already available in the verifier, but need to be revisited based on any revision of layout's implementation.

  *  size(`layout`) $=$ size(`results`) - 1.
  *  size(`layout[i]`) $=$ rank(`results[i]`) for all i $\in$ [0,size(`layout`)).

burmako:
Closing in favor of #629. In the corresponding spec PR, we have decided to not spec layouts for InfeedOp, so I imagine that before we fix layout verification for InfeedOp, we'll need to rationalize it (via #629) in the first place.

PR #580, opened 2022-11-22
Add spec for SendOp and RecvOp

sdasgup3:
fixes #527 

verification is revisit on https://github.com/openxla/stablehlo/issues/579 and https://github.com/openxla/stablehlo/issues/667

PR #640, opened 2022-11-28
Add spec for InfeedOp and OutfeedOp

### LINKED ISSUE #525: Add spec for InfeedOp/OutfeedOp
### Request description

To spec `stablehlo.infeed` and stablehlo.outfeed

sdasgup3:
fixes #525 

verification of infeed op is revisit based on https://github.com/openxla/stablehlo/issues/639

Note that for outfeed op we are using the  output name  as `results` based on the implicit name [introduced by the tablegen file](https://github.com/openxla/stablehlo/blob/dbf9964c6ce980e1976041700807bd94e6adf849/stablehlo/dialect/StablehloOps.td#L1030 ) even though it should be explicitly called out as `result` (singular). This will be addressed in https://github.com/openxla/stablehlo/issues/351#issuecomment-1326914173, item 4.

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -379,6 +379,7 @@ syntax.
    * [get_tuple_element](#stablehloget_tuple_element)
    * [if](#stablehloif)
    * [imag](#stablehloimag)
+   * [infeed](#stablehloinfeed)
    * [iota](#stablehloiota)
    * [is_finite](#stablehlois_finite)
    * [log](#stablehlolog)
@@ -392,6 +393,7 @@ syntax.
    * [not](#stablehlonot)
    * [optimization_barrier](#stablehlooptimization_barrier)
    * [or](#stablehloor)
+   * [outfeed](#stablehlooutfeed)
    * [pad](#stablehlopad)
    * [popcnt](#stablehlopopcnt)
    * [power](#stablehlopower)
@@ -2324,6 +2326,48 @@ More formally, for each element `x`: `imag(x) = is_complex(x) ? x.imag : 0.0`.
 
 [Back to Ops](#index-of-ops)
 
+
+## stablehlo.infeed
+
+### Semantics
+
+Reads `results` from the infeed queue of the device. Multiple
+`stablehlo.infeed` operations are allowed in a computation, but there must be a
+total order among them.
+
+The infeed configuration `infeed_config` includes target-dependent metadata
+needed for the backend compiler (e.g., infeed buffer address).
+
+### Inputs
+
+| Name            | Type                                                                                     |
+|-----------------|------------------------------------------------------------------------------------------|
+| `token`         | `token`                                                                                  |
+| `infeed_config` | string                                                                                   |
+| `layout`        | array of array of type `integer` [todo](https://github.com/openxla/stablehlo/issues/629) |
+
+### Outputs
+
+| Name      | Type                                                       |
+|-----------|------------------------------------------------------------|
+| `results` | variadic number of tensors of any supported type or tokens |
+
+### Constraints
+
+  * (C1) size(`results`) $\ge$ 1.
+  * (C2) type(`results`[-1]) $=$ `token`.
+  * (C3) size(`layout`) $=$ size(`results`) - 1.
+  * (C4) size(`layout[i]`) $=$ rank(`results[i]`) for all i $\in$ [0,
+         size(`layout`)).
+
+### Examples
+
+```mlir
+%results:2 = "stablehlo.infeed"(%token) { infeed_config = "", layout = [[2, 0, 1] } : (!stablehlo.token) -> (tensor<3x3x3xi32>, !stablehlo.token)
+```
+
+[Back to Ops](#index-of-ops)
+
 ## stablehlo.iota
 
 ### Semantics
@@ -2891,6 +2935,38 @@ operation.
 // %rhs: [[false, true], [false, true]]
 %result = "stablehlo.or"(%lhs, %rhs) : (tensor<2x2xi1>, tensor<2x2xi1>) -> tensor<2x2xi1>
 // %result: [[false, true], [true, true]]
+```
+
+[Back to Ops](#index-of-ops)
+
+## stablehlo.outfeed
+
+### Semantics
+
+Writes `inputs` to the outfeed queue of the device. The input `token` is used
+for ordering side-effecting operations.
+
+The outfeed configuration `outfeed_config` includes target-dependent metadata
+needed for the backend compiler.
+
+### Inputs
+
+| Name             | Type                                             |
+|------------------|--------------------------------------------------|
+| `inputs`         | variadic number of tensors of any supported type |
+| `token`          | `token`                                          |
+| `outfeed_config` | string                                           |
+
+### Outputs
+
+| Name      | Type    |
+|-----------|---------|
+| `results` | `token` |
+
+### Examples
+
+```mlir
+%results = "stablehlo.outfeed"(%arg0, %arg1) {outfeed_config = ""} : (tensor<3x3x3xi32>, !stablehlo.token) -> !stablehlo.token
 ```
 
 [Back to Ops](#index-of-ops)
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -26,6 +26,9 @@ Following are the supported element types in StableHLO:
     (represents a pair of `f64`). Exact representation of complex types
     (e.g. whether the real part or the imaginary part comes first in memory)
     is implementation-defined.
+  * **String type**  represent a sequence of bytes. Exact representation of
+    string type (e.g. null terminated or not, encoding etc.) is
+    implementation-defined.
 
 **Tensor types** are the cornerstone of the StableHLO type system. They model
 immutable n-dimensional arrays and are referred to in the document as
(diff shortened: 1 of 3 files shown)


docs/spec_draft.md
```diff
@@ -1576,6 +1578,48 @@ More formally, for each element `x`: `imag(x) = is_complex(x) ? x.imag : 0.0`.
...
+
+## stablehlo.infeed
+
+### Semantics
+
+Reads `results` from the infeed queue of the device. Multiple
+`stablehlo.infeed` operations are allowed in a computation, but there must be a
+total order among them.
+
+The infeed configuration `infeed_config` includes target-dependent metadata
```
subhankarshah, line 1590: `infeed_config` is an arbitrary sting which may include target-dependent metadata.

docs/spec_draft.md
```diff
@@ -1576,6 +1578,48 @@ More formally, for each element `x`: `imag(x) = is_complex(x) ? x.imag : 0.0`.
...
 [Back to Ops](#index-of-ops)
 
+
+## stablehlo.infeed
+
+### Semantics
+
+Reads `results` from the infeed queue of the device. Multiple
+`stablehlo.infeed` operations are allowed in a computation, but there must be a
+total order among them.
```
subhankarshah, line 1588: Reword suggestion for the second line - 
There must be a total order among all the `infeed` operations in a computation.

Since this looks like a constraint, we can move it to constraints section as well. WDYT?

docs/spec_draft.md
```diff
@@ -2108,6 +2152,38 @@ operation.
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.outfeed
+
+### Semantics
+
+Writes `inputs` to the outfeed queue of the device. The input `token` is used
+for ordering side-effecting operations.
```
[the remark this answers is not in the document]
sdasgup3: Removed "The input token .." as it is nothing new and already mentioned in "Token Types" section above.

Let me explore the rest.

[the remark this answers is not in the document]
sdasgup3: Yes, the outfeed ops must also obey the total order. Updated the text

docs/spec_draft.md
sdasgup3: Two reason we should not view it as constraints:

1. All the properties are have mentioned in the constraints section are such that they can be verified at the local (op level). This property is not local. 
2. To me this property is like a recommendation for program correctness and termination and is common across other parallel programming domains. 

I put it as a recommendation only in the text.

docs/spec_draft.md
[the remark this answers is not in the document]
subhankarshah: What I meant was we can just write 1 line in both places.

`token` is used for ordering side-effecting operations so they have a total order among them.
or ever shorter
`token` is used for ordering side-effecting operations.

We can choose to remove the following line.
Not having a total order among the infeed operations in a program might lead to
data races or even deadlocks.

docs/spec_draft.md
subhankarshah: Just keeping this unresolved for the next reviewer. I am not sure saying `infeed_config` arbitrary string would be right, but I would still go with this or maybe just remove the word arbitrary.

`infeed_config` is a sting which may contain target-dependent metadata.

docs/spec_draft.md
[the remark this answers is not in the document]
subhankarshah: keeping this open.

subhankarshah (approved):
LGTM with minor suggestions.

docs/spec_draft.md
```diff
@@ -2311,6 +2313,49 @@ More formally, for each element `x`: `imag(x) = is_complex(x) ? x.imag : 0.0`.
 
 [Back to Ops](#index-of-ops)
 
+
```
burmako, line 2316: Similarly to https://github.com/openxla/stablehlo/pull/580/files#r1037781796, let's introduce another section under "Parallel execution":

```
### Streaming communication

Every StableHLO process has access to two streaming interfaces:
  * **Infeed** that can be read from.
  * **Outfeed** that can be written to.

Unlike channels, which are used to communicate between processes and therefore
have processes at both of their ends, infeeds and outfeeds have their other
end implementation-defined.

Further formalization, e.g. how streaming communication influences execution
order and what kind of synchronization is introduced by it, is TBD.
```

docs/spec_draft.md
```diff
@@ -2311,6 +2313,49 @@ More formally, for each element `x`: `imag(x) = is_complex(x) ? x.imag : 0.0`.
...
+### Semantics
+
+Reads `results` from the infeed queue of the device.
+
+Not having a total order among the infeed operations in a program might lead to
+data races or even deadlocks.
+
+The infeed configuration string `infeed_config` includes target-dependent
+metadata.
```
burmako, line 2327: Similarly to https://github.com/openxla/stablehlo/pull/580/files#r1037788421, proposed rewording:

```
Reads data from the infeed and produces `results`.

Semantics of `infeed_config` is implementation-defined.

`results` consist of payload values which come first and a token which comes
last. The operation produces a token to reify the side effect of this operation 
as a value that other operations can take a data dependency on.
```

docs/spec_draft.md
```diff
@@ -2882,6 +2927,40 @@ operation.
...
+
+### Semantics
+
+Writes `inputs` to the outfeed queue of the device.
+
+Not having a total order among the outfeed operations in a program might lead to
+data races or even deadlocks.
+
+The outfeed configuration string  `outfeed_config` includes target-dependent
+metadata.
```
burmako, line 2940: Similarly https://github.com/openxla/stablehlo/pull/580/files#r1037795981, proposed rewording:

```
Writes `inputs` to the outfeed and produces `result`.

Semantics of `outfeed_config` is implementation-defined.

The operation takes a token and produces a token to reify its side effects
as a value that other operations can take a data dependency on.
```

burmako (changes requested):

docs/spec_draft.md
```diff
@@ -2311,6 +2313,49 @@ More formally, for each element `x`: `imag(x) = is_complex(x) ? x.imag : 0.0`.
...
+
+The infeed configuration string `infeed_config` includes target-dependent
+metadata.
+
+### Inputs
+
+| Name            | Type                                                                                     |
+|-----------------|------------------------------------------------------------------------------------------|
+| `token`         | `token`                                                                                  |
+| `infeed_config` | string                                                                                   |
```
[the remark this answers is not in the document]
sdasgup3: Added the sections.
PTAL.
@subhankarshah

docs/spec_draft.md
```diff
@@ -2311,6 +2313,49 @@ More formally, for each element `x`: `imag(x) = is_complex(x) ? x.imag : 0.0`.
...
+| `layout`        | array of array of type `integer` [todo](https://github.com/openxla/stablehlo/issues/629) |
```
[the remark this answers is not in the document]
sdasgup3: added markers to highlight the relevant ticket.

docs/spec_draft.md
```diff
@@ -26,6 +26,9 @@ Following are the supported element types in StableHLO:
     (represents a pair of `f64`). Exact representation of complex types
     (e.g. whether the real part or the imaginary part comes first in memory)
     is implementation-defined.
+  * **String type**  represent a sequence of bytes. Exact representation of
+    string type (e.g. null terminated or not, encoding etc.) is
+    implementation-defined.
```
burmako, line 31: Proposed rewording:

```
**String type** represent a sequence of bytes and is referred to in the document 
as `string`. Exact representation of string type (e.g. null terminated or not,
encoding etc.) is implementation-defined.

Strings types are not first class, i.e. StableHLO doesn't support values of
string types. Some StableHLO ops can take strings as inputs, but they are never
produced as outputs.
```

Feedback:
  1. Let's move out of element types and all the way to the very bottom of the section.
  2. Also let's update the third bullet in function types to say something like "Input types and output types are one of `tensor`, `token` or `tuple`.".

docs/spec_draft.md
```diff
@@ -339,6 +355,9 @@ syntax.
   floating-point constants of `f32` or `f64` types, e.g. `(12.34, 56.78)`,
   where the first constant is the real part, and the second constant is the
   imaginary part.
+  * **String constants** String constants are represented as a sequence of
+  bytes enclosed in double quotation mark symbols, e.g. "foo123?" (in ASCII
+  encoding) or "\18\A3" (in hex encoding).
```
burmako, line 360: Let's move string constants to the bottom of the list to reflect the proposed order in the Types section.

docs/spec_draft.md
```diff
@@ -2324,6 +2345,49 @@ More formally, for each element `x`: `imag(x) = is_complex(x) ? x.imag : 0.0`.
...
+  * (C1) size(`results`) $\ge$ 1.
+  * (C2) type(`results`[-1]) $=$ `token`.
+  * -- [Verify layout in InfeedOp](https://github.com/openxla/stablehlo/issues/639) --
+
+### Examples
+
+```mlir
+%results:2 = "stablehlo.infeed"(%token) {
+  infeed_config = "",
+  layout = [[2, 0, 1]]
```
burmako, line 2385: Drop layout, since we dropped it from Inputs?

burmako (changes requested):
A few minor changes, and we're ready to go.

burmako (approved):

burmako:
Merging without waiting for CI - the only CI-able change in the PR is documentation within the .td file.

PR #510, opened 2022-11-11
Add spec for AllReduceOp

sdasgup3:
fixes #504  

Address the followings:
1.  Adds verification checks for AllGather w.r.t. https://github.com/openxla/stablehlo/issues/498
2. Fixes https://github.com/openxla/stablehlo/issues/508

PR #659, opened 2022-11-30
Add headers to parallel execution semantics

subhankarshah:
Added headers to the process_group paragraphs in parallel execution semantics, so they can be linked from other parts of the spec.

### DIFF
--- a/docs/spec_draft.md
+++ b/docs/spec_draft.md
@@ -207,11 +207,12 @@ which the grid is split into process groups are shared between these ops and are
 described in this section. More formally, StableHLO supports the following
 four strategies.
 
-1) **cross_replica** (as in "only cross-replica communications will be happening
-within each process group"). This strategy takes `replica_groups` - a list of
-lists of replica ids - and computes a Cartesian product of `replica_groups` by
-`partition_ids`. `replica_groups` must have unique elements and cover all
-`replica_ids`. More formally:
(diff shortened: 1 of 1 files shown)


subhankarshah:
As a reader I would love to be able to jump to these and refer them again when I am reading the spec.

docs/spec_draft.md
```diff
@@ -268,11 +272,12 @@ For example, for `replica_groups = [[0, 1], [2, 3]]` and `num_partitions = 2`,
...
 
-4) **flattened_ids**. This strategy takes `flattened_id_groups` - a list of
-lists of "flattened" process ids in the
-`replica_id * num_partitions + partition_id` format - and turns them into
-process ids. `flattened_id_groups` must have unique elements and cover all
-`process_ids`. More formally:
+#### flattened_ids
+
+This strategy takes `flattened_id_groups` - a list of lists of "flattened"
+process ids in the form of `replica_id * num_partitions + partition_id` - and
```
subhankarshah, line 278: minor reword here.

burmako (approved):
I think it's a good idea to have individual sections for these strategies. Not yet sold on hyperlinking to them from other ops (left a comment in https://github.com/openxla/stablehlo/pull/510#discussion_r1036519786 to that end), but that's not a blocker for this PR.

PR #668, opened 2022-12-02
Fix header guards

GleasonK:


GleasonK:
Found types using `grep -riL "define STABLEHLO" --include="*.h" .` which lists header files that don't have include guards that start with `STABLEHLO`.

Eventually we should use clang-tidy to find things like this.

sdasgup3 (approved):

PR #691, opened 2022-12-04
Implement type inference for ReturnOp

burmako:
Cannot write a FileCheck-based test because ReturnOp doesn't have result values.


ghpvnist (changes requested):
status.md should also be updated

ghpvnist (approved):

PR #705, opened 2022-12-06
Fix broken HEAD

burmako:
PR #672 was based off HEAD which didn't yet have PR #510, and the former PR happened to make an incompatible change to an API (verifyReducerShape) used by the latter PR.

This PR fixes the incompatibility and fixes HEAD.


burmako:
Landing without approvals, because it's pretty late in Pacific Time, and we need to unbreak HEAD as soon as possible.

Issue #709, opened 2022-12-07
Preparation change for MLIR to only produce prefixed accessor.


sdasgup3:
### Request description

Corresponding StableHLO change: https://github.com/tensorflow/mlir-hlo/commit/1ee8f36bc17baea924ac8606dd546a114cb2aeca

PR #710, opened 2022-12-07
Preparation change for MLIR to only produce prefixed accessor. 

sdasgup3:
fixes https://github.com/openxla/stablehlo/issues/709


sdasgup3:
error in merged. need to fix.

PR #713, opened 2022-12-07
Add type inference for OutfeedOp

### LINKED ISSUE #699: Add type inference for OutfeedOp
### Request description

To add type inference for [stablehlo.outfeed](https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md#stablehlooutfeed)

ghpvnist:
In addition, it updates the semantics of OutfeedOp and SendOp to explicitly mention that the 'result' is a token type.
Also removes adaptors from CreateTokenOp and AfterAllOp that are not used.
closes #699

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -1055,6 +1055,13 @@ LogicalResult inferOptimizationBarrierOp(
   return success();
 }
 
+LogicalResult inferOutfeedOp(Optional<Location> location, ValueRange inputs,
+                             Value token, StringAttr outfeed_config,
+                             SmallVectorImpl<Type>& inferredReturnTypes) {
+  inferredReturnTypes.emplace_back(token.getType());
+  return success();
+}
+
 // We intend to verify the following properties
 //  P1. Verify all `inputs` need to have compatible shapes.
 //  P2. Verify that
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -1055,10 +1055,9 @@ LogicalResult inferOptimizationBarrierOp(
   return success();
 }
 
-LogicalResult inferOutfeedOp(Optional<Location> location, ValueRange inputs,
-                             Value token, StringAttr outfeed_config,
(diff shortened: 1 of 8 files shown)


stablehlo/dialect/TypeInference.cpp
```diff
@@ -986,6 +987,13 @@ LogicalResult inferOptimizationBarrierOp(
   return success();
 }
 
+LogicalResult inferOutfeedOp(Optional<Location> location, ValueRange inputs,
+                             Value token, StringAttr outfeed_config,
+                             SmallVectorImpl<Type>& inferredReturnTypes) {
+  inferredReturnTypes.emplace_back(token.getType());
```
ghpvnist, line 993: Since this function does not use `inputs` and `outfeed_config` should these be removed or included in case future changes require them?

burmako: Yes, let's remove for now.

burmako (changes requested):

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1055,6 +1055,12 @@ LogicalResult inferOptimizationBarrierOp(
   return success();
 }
 
+LogicalResult inferOutfeedOp(Optional<Location> location, Value token,
+                             SmallVectorImpl<Type>& inferredReturnTypes) {
+  inferredReturnTypes.push_back(token.getType());
```
burmako, line 1060: Let's drop `token` as well and just return `TokenType::get` to make things a bit more robust. Sorry - didn't notice this during the first round of reviews.

burmako (changes requested):

stablehlo/dialect/TypeInference.cpp
ghpvnist: Sure, but I'd instead need to pass `MLIRContext`. Though, using this will make this similar to the other PRs recently merged.

ghpvnist:
I realized that AfterAllOp, CreateTokenOp, and OutfeedOp all don't need an adaptor initialized, so I've removed them in this PR as well.

PR #717, opened 2022-12-07
Implement VHLO Bytecode Interface

### LINKED ISSUE #675: Implement bytecode interface for VHLO
### Request description

VHLO (Versioned StableHLO, #278) needs custom bytecode for all types and attributes, including forked upstream types (#674).

GleasonK:
Closes #675.

Favoring code duplication given the nature of the dialects, that they will diverge over time as any differences are introduced. StableHLO serialization only needs to support reading the latest version of the dialect. VHLO needs to support reading all versions of attributes and types. Also the only good way to share this code is via heavily templated methods.

### DIFF
--- a/stablehlo/dialect/CMakeLists.txt
+++ b/stablehlo/dialect/CMakeLists.txt
@@ -164,6 +164,7 @@ add_dependencies(mlir-headers VhloOpsIncGen)
 
 add_mlir_dialect_library(VhloOps
   PARTIAL_SOURCES_INTENDED
+  VhloBytecode.cpp
   VhloOps.cpp
   Version.cpp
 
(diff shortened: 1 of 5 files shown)


stablehlo/dialect/CMakeLists.txt
```diff
@@ -172,7 +172,18 @@ add_mlir_dialect_library(VhloOps
 
   LINK_LIBS PUBLIC
   StablehloAssemblyFormat
+  VhloBytecode
   MLIRIR
   MLIRSupport
   MLIRQuantDialect
 )
+
+add_mlir_dialect_library(VhloBytecode
```
burmako, line 181: Let's do the same thing as ChloBytecode.cpp and StablehloBytecode.cpp and merge this target into VhloOps?

burmako (approved):
LGTM except for a few minor comments.

stablehlo/dialect/CMakeLists.txt
GleasonK: Makes sense to keep these together. `VhloBytecode.cpp` depends on `VhloOps.h` and `VhloOps.cpp` depends on `VhloBytecode.h`.

PR #719, opened 2022-12-08
Fix typo for AllReduceOp columns

ghpvnist:
According to #510, the type inference column was accidentally set to `yes` instead of the verification column.

### DIFF
--- a/docs/status.md
+++ b/docs/status.md
@@ -46,7 +46,7 @@ one of the following tracking labels.
 | add                      | yes           | yes          | yes            | yes             | yes         |
 | after_all                | yes           | yes          | yes            | yes             | no          |
 | all_gather               | yes           | revisit      | no             | no              | no          |
-| all_reduce               | yes           | revisit      | yes            | no              | no          |
+| all_reduce               | yes           | yes          | revisit        | no              | no          |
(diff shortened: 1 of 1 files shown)


docs/status.md
```diff
@@ -46,7 +46,7 @@ one of the following tracking labels.
 | add                      | yes           | yes          | yes            | yes             | yes         |
 | after_all                | yes           | yes          | yes            | yes             | no          |
 | all_gather               | yes           | revisit      | no             | no              | no          |
-| all_reduce               | yes           | revisit      | yes            | no              | no          |
+| all_reduce               | yes           | yes          | revisit        | no              | no          |
```
sdasgup3, line 49: https://github.com/openxla/stablehlo/blob/01bbf2ae747f0ab95347d368368b44fefd34fb08/stablehlo/dialect/StablehloOps.cpp#L294

The verification is `revisit` based on https://github.com/openxla/stablehlo/issues/498

burmako: +1

burmako (changes requested):

docs/status.md
ghpvnist: Makes sense, thanks for the pointing to the related line @sdasgup3

PR #711, opened 2022-12-07
Add type inference for CreateTokenOp

ghpvnist:
closes #698

PR #721, opened 2022-12-08
Sever the dependency from TypeInference.h to StablehloOps.h

burmako:
I missed something important when reviewing #711.

We cannot have a dependency from TypeInference.h to StablehloOps.h. It should be the other way around - both StablehloOps.h and hlo_ops.h depending on TypeInference.h

This PR severs the dependency by introducing an interface that abstracts away creating token types.

### DIFF
--- a/stablehlo/dialect/Base.h
+++ b/stablehlo/dialect/Base.h
@@ -125,11 +125,11 @@ class HloDialectInterface : public DialectInterface::Base<HloDialectInterface> {
  public:
   HloDialectInterface(Dialect *dialect) : Base(dialect) {}
 
-  // Creates a TokenType type, specific this dialect.
+  // Creates a TokenType type, specific to this dialect.
   // See docs for the particular type in the corresponding dialect.
   virtual Type createTokenType() const = 0;
 
-  // Creates a TypeExtensions attribute, specific this dialect.
+  // Creates a TypeExtensions attribute, specific to this dialect.
   // See docs for the particular attribute in the corresponding dialect.
   virtual Attribute createTypeExtensions(ArrayRef<int64_t> bounds) const = 0;
 };
--- a/stablehlo/dialect/Base.h
+++ b/stablehlo/dialect/Base.h
@@ -118,14 +118,20 @@ ArrayRef<int64_t> encodingToBounds(Attribute encoding);
 // the underlying dialect that knows how to create these attributes.
 Attribute boundsToEncoding(Attribute prototype, ArrayRef<int64_t> bounds);
 
-// This interface is used for HLO dialects that have accompanying
-// BoundedAttrInterface attributes which can carry bounds for dimension sizes
-// of accompanying shaped types.
-class BoundedDialectInterface
-    : public DialectInterface::Base<BoundedDialectInterface> {
+// This interface is implemented by both StableHLO and MHLO dialects
+// and is used as the foundation for sharing verification, type inference and
+// prettyprinting logic between them.
+class HloDialectInterface : public DialectInterface::Base<HloDialectInterface> {
  public:
-  BoundedDialectInterface(Dialect *dialect) : Base(dialect) {}
-  virtual Attribute createBoundedAttr(ArrayRef<int64_t> bounds) const = 0;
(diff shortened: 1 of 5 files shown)


stablehlo/dialect/TypeInference.cpp
```diff
@@ -692,10 +692,10 @@ LogicalResult inferConcatenateOp(Optional<Location> location, ValueRange inputs,
   return success();
 }
 
-LogicalResult inferCreateTokenOp(MLIRContext* context,
-                                 Optional<Location> location,
+LogicalResult inferCreateTokenOp(Dialect* dialect, Optional<Location> location,
                                  SmallVectorImpl<Type>& inferredReturnTypes) {
-  inferredReturnTypes.push_back(stablehlo::TokenType::get(context));
+  auto hloDialect = cast<HloDialectInterface>(dialect);
```
GleasonK, line 697: Not sure if there's much of an advantage, but could use `dialect->getRegisteredInterface<HloDialectInterface>()` I believe. Looks like it was made for retrieving dialect interfaces, not sure if its outdated or preferred.

https://mlir.llvm.org/doxygen/classmlir_1_1Dialect.html#a46a6a74902ee28426cebc490d04ee67b

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -4499,10 +4502,14 @@ struct HLOInlinerInterface : public DialectInlinerInterface {
   }
 };
 
-struct HLOBoundedDialectInterface : public hlo::BoundedDialectInterface {
```
GleasonK, line 4502: (no action required)
Just did a quick search - looks like these were the only instances of the `HLO` in all caps naming convention, good to clean this up. Similarly for MHLO codebase these two structs are the only using this variable casing.

stablehlo/dialect/TypeInference.h
```diff
@@ -19,8 +19,11 @@ limitations under the License.
 
 #include "mlir/IR/BuiltinAttributes.h"
 #include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/Dialect.h"
 #include "mlir/IR/Location.h"
+#include "mlir/IR/Region.h"
 #include "mlir/IR/Types.h"
+#include "mlir/IR/Value.h"
```
GleasonK, line 26: Were Region and Value missing from this file? Changes are only for `Dialect` by the looks of it.

GleasonK (approved):
I build each target independently on the current main, but nothing broke. I'm guessing it's because the only reliance on `StablehloOps` is `TokenType` which is entirely defined in the header file used.

Also confirmed that there is nothing in CMakeLists that needs to be cleaned up. LGTM.

stablehlo/dialect/Base.h
```diff
@@ -118,14 +118,20 @@ ArrayRef<int64_t> encodingToBounds(Attribute encoding);
...
+// This interface is implemented by both StableHLO and MHLO dialects
+// and is used as the foundation for sharing verification, type inference and
+// prettyprinting logic between them.
+class HloDialectInterface : public DialectInterface::Base<HloDialectInterface> {
  public:
-  BoundedDialectInterface(Dialect *dialect) : Base(dialect) {}
-  virtual Attribute createBoundedAttr(ArrayRef<int64_t> bounds) const = 0;
+  HloDialectInterface(Dialect *dialect) : Base(dialect) {}
+
+  // Creates a TokenType type, specific this dialect.
```
sdasgup3, line 128: nit: specific to 
and below

sdasgup3 (approved):

stablehlo/dialect/TypeInference.h
burmako: Good point! I've reverted the change.

stablehlo/dialect/TypeInference.cpp
burmako: Looks like `cast` [calls](https://github.com/llvm/llvm-project/blob/b4028fbc1a8853c34517f44ef779d517a44bd71d/mlir/include/mlir/IR/Dialect.h#L324-L339) `getRegisteredInterface`. I think let's stick to `cast` which is uniform with other forms of casting that we have in the codebase.

PR #729, opened 2022-12-09
Change some MD syntax for compatibility with Python Markdown

scottamain:
This has no effect on the appearance on GitHub but fixes some things for rendering with other Markdown processors, particularly Python Markdown, which we will probably use for the OpenXLA website. (I'll squash the two changes when merging.)


docs/spec_draft.md
```diff
@@ -664,16 +671,17 @@ The operation splits the StableHLO grid into process groups as follows:
     `flattened_ids(replica_groups)`.
 
 Afterwards, within each `process_group`:
+
   * `operands@receiver = [operand@sender for sender in process_group]` for all
     `receiver` in `process_group`.
-  * ```
+  *  &nbsp;
```
GleasonK, line 677: Is this necessary? The current layout has the code block on the same line as the bullet point, the new one has an empty line then the code block (on GH md at least)

GleasonK: Same question for all `&nbsp;` introduced in this file.

GleasonK (commented):
This looks great! I have it on my task list to look into a markdown linter. If you have any suggestions for one compatible with python markdown let me know. I'm leaning towards `markdownlint`since it seems to be the standard. I believe it can catch most of the changes you made in this doc. I'll make sure whatever solution we choose catches things like spaces after headers, header consistency, using image markdown instead of image html, etc.

Will leave the approval to Eugene, but LGTM.

docs/spec_draft.md
scottamain: Unfortunately, yes. Python Markdown requires a whitespace after the bullet (so adding the entity makes it explicit, so editors don't remove the trailing space) but PyMd also requires code blocks start on a new line. I couldn't find any extensions to make it work otherwise (but I will send a feature request for the Superfences extension to allow it because I don't see the logic to disallow it). However, I think we should just work-around this shortcoming be avoiding the pattern altogether, because it's grammatically a little weird to start a bullet with a code block; it might be nice to have a phrase that leads into it.

scottamain:
Thanks Kevin. A lint tool would be great, and ideally one that is both runnable in the GitHub CI but also pluggable into popular IDEs so we all have a way to verify before uploading. I don't have any suggestions, but I think going with the more popular tool is a safe bet. Although, IMO, CommonMark is probably the superior parser out-of-the-box, PyMd has a lot of extensions that allow it to behave nearly the same way (these fences inside a bullet is the only difference I've seen that's still a nuisance). (FTR: Using PyMd for the website, via MkDocs, is a trade-off in order to get the other website build/design tools that I consider superior to the alternative static-site generators that do use CommonMark, such as Hugo.)

docs/spec_draft.md
scottamain: Actually I figured out a facepalm-inducing fix for this so the code block appears on the first line: nbsp is a "non-breaking space" that gets parsed by Markdown as a span of text (it gets treated the same as any other character strings), so I replaced it with the unicode for "space" and now the parser ignores it and collapses the code block onto the first line. 👍 

It's still worth considering an introductory phrase for these things, along some other grammatical styles for this document. But now that the visual effect is gone, we can do that as a low-priority cleanup later.

sdasgup3: I will consider adding one. In the meanwhile thanks for preventing that to become a bottleneck.

sdasgup3 (commented):
I think we had  a [reason](https://github.com/openxla/stablehlo/pull/275#discussion_r998913873) to name the directories in a certain way (similar to the name of the spec file). Can you please let us know your rationale?

scottamain:
Hi @sdasgup3 I think that reasoning still applies—there is a directory with the same name as the markdown file. It now resides in the images/ path so that, in the event that N files also need images, you don't end up with N new directories in this root directory, which gets messy and confusing. They're instead tucked away in the shared images dir.

docs/spec_draft.md
GleasonK: > It's still worth considering an introductory phrase for these things

Still on board with this. At this point we're relying on parser whitespace inconsistencies. Clever trick to get it to work, nonetheless! Marking as resolved.

GleasonK (approved):
Looks like the GH Actions piece of this change has been removed from this PR, making these changes pretty non-controversial. I agree with the images directory renaming. It is a similar directory structure to what we have in [115](https://github.com/openxla/stablehlo/pull/115/files).  I'll refactor the images directory in that CL to have an appropriately named subdirectory.

burmako (approved):

sdasgup3 (approved):

PR #673, opened 2022-12-02
Add support for PartitionIdOp

subhankarshah:
closes #632

PR #736, opened 2022-12-09
Add versions to attributes, enums, and types

### LINKED ISSUE #676: Implement versioned types and attributes for VHLO
### Request description

VHLO (Versioned StableHLO, https://github.com/openxla/stablehlo/pull/278) currently only has versioned ops. Before it is production ready, it will also need support for versioned Attributes and Types.

GleasonK:
- Moved `TokenType` to ODS in VHLO to add consistent implementation for versioned interface. We can consider moving StableHLO/MHLO TokenType decls to ODS as well.
- Added a `Versioned(Op|Attr|Type)Interface` declaration to version ops/attrs/types.
- Moved `Vhlo(Enums|Types|Attrs).h.inc` to bottom of `VhloOps.cpp` since the version interfaces had a dependency on `VhloDialect::getCurrentVersion()`.
- Updated legality checks to include checking attr and type versions satisfy the target version.
- Added test point for legal types/attrs, currently only `OutputOperandAlias`, added in `0.4.0` is the only attribute that can be tested for illegality.

Currently, there are only `V1` attributes and types. This means we do not need to worry about type or attribute converters in the current StableHLO release. To avoid introducing technical debt for the sake of testing type/attr conversion, I think it is best we submit conversion code after we have a use case for it.

I have confirmed that Type and Attribute conversions are possible, and have prototyped a few machineries for these conversions. See: https://github.com/openxla/stablehlo/commit/334b88ba190fc8127e76e60f2072a7b0f5f2ef2b

Closes #676.


stablehlo/dialect/VhloBase.td
```diff
@@ -34,6 +34,49 @@ def VersionedInterface : OpInterface<"VersionedInterface"> {
...
 }
 
+def VHLO_VersionedAttrInterface : AttrInterface<"VersionedAttrInterface"> {
+  let cppNamespace = "::mlir::vhlo";
+  let methods = [
+    InterfaceMethod<
+      "Returns the minimum version an attribute is supported in.",
+      "mlir::vhlo::Version", "getMinVersion">,
+    InterfaceMethod<
+      "Returns the maximum verison an attribute is supported in.",
```
burmako, line 44: Typo: "version". (Here, below and above).

stablehlo/dialect/VhloAttrs.td
```diff
@@ -22,6 +22,25 @@ include "stablehlo/dialect/VhloEnums.td"
 
 include "mlir/IR/AttrTypeBase.td"
 
+//===----------------------------------------------------------------------===//
+// Attribute Versioning
+//===----------------------------------------------------------------------===//
+
+class VHLO_AttrDef<string name,
+                   string minVersion = "0.3.0",
+                   string maxVersion = "current">
```
burmako, line 31: What do you think about removing default values for minVersion and maxVersion and explicitly specifying them everywhere? This way each entity's minimum and maximum version will be easily discoverable without reliance on code search. 

Also, for all ops, attributes and types that will be added after this PR, the minVersion default will be wrong, which might lead to confusion that we've already observed in https://github.com/openxla/stablehlo/pull/673#discussion_r1043710870. If removing default values for minVersion and maxVersion sounds like too much, perhaps let's remove the default value for minVersion?

stablehlo/dialect/VhloBase.td
```diff
@@ -34,6 +34,49 @@ def VersionedInterface : OpInterface<"VersionedInterface"> {
   ];
 }
 
+def VHLO_VersionedAttrInterface : AttrInterface<"VersionedAttrInterface"> {
+  let cppNamespace = "::mlir::vhlo";
+  let methods = [
+    InterfaceMethod<
+      "Returns the minimum version an attribute is supported in.",
```
burmako, line 41: How about expanding to "the minimum version of the VHLO dialect"? (Here, below and above).

stablehlo/dialect/VhloAttrs.td
```diff
@@ -22,6 +22,25 @@ include "stablehlo/dialect/VhloEnums.td"
...
+  : AttrDef<VHLO_Dialect, name, [VHLO_VersionedAttrInterface]> {
+  let extraClassDeclaration = [{
+    mlir::vhlo::Version getMinVersion() {
+      return *mlir::vhlo::Version::fromString("}] #  minVersion # [{");
```
burmako, line 35: This is something that I missed when reviewing the previous pull request, but what happens if someone accidentally specifies a version that doesn't parse. Can we `llvm::report_fatal_error` instead of dereferencing an empty optional?

stablehlo/dialect/VhloBase.td
burmako, line 44: "maximum version (inclusive)"?

stablehlo/tests/vhlo_to_version_downgrade_invalid.mlir
```diff
@@ -21,6 +21,18 @@ func.func @custom_call_v2_with_output_operand_alises(%arg0 : tensor<f32>) -> ten
 
 // -----
 
+func.func @op_with_output_operand_alises(%arg0 : tensor<f32>) -> () {
+  // expected-error @+1 {{failed to legalize operation 'vhlo.add' that was explicitly marked illegal}}
+  stablehlo.add %arg0, %arg0 {
+    op_alias_attr = #stablehlo.output_operand_alias<output_tuple_indices = [],
+                                                    operand_index = 0,
+                                                    operand_tuple_indices = []>
```
burmako, line 29: I think I understand why you're doing this (it would be good to test the newly added attribute, but testing stablehlo.custom_call with an output_operand_alias won't achieve much because the test will fail due to a different reason), but I don't think that testing this attribute in isolation is necessary. 

"What's not covered?" in the compatibility RFC says "Unspecced features: We may make backward incompatible changes to features which haven't been specced" and using unregistered attributes counts as an unspecced feature.

stablehlo/transforms/Passes.td
```diff
@@ -28,7 +28,7 @@ def VhloLegalizeToStablehloPass : Pass<"vhlo-legalize-to-stablehlo", "ModuleOp">
 def VhloToVersionPass : Pass<"vhlo-to-version"> {
   let summary = "Convert between versions of VHLO.";
   let options = [
-    Option<"targetVersion", "target", "std::string", "",
+    Option<"targetVersionClopt", "target", "std::string", "",
```
burmako, line 31: "targetVersionOption"? I don't think using an abbreviation here is worth it, since there's a straightforward conventional name which is just one character longer.

stablehlo/dialect/Version.h
```diff
@@ -59,6 +60,14 @@ class Version {
 };
 
 mlir::Diagnostic& operator<<(mlir::Diagnostic& diag, const Version& version);
+llvm::raw_ostream& operator<<(llvm::raw_ostream& os, const Version& version);
+
+template <typename VersionedInterface>
+bool isLegalVersionForTarget(VersionedInterface& interface,
+                             Version const& target) {
```
burmako, line 67: `const Version&` as the more conventional option in our codebase? (Here and everywhere else - this is something that I seem to have missed when reviewing the previous PR).

stablehlo/transforms/VhloToVersion.cpp
```diff
@@ -77,6 +78,69 @@ FailureOr<Version> validateTargetVersion(llvm::StringRef versionRef,
   return targetVersion;
 }
 
+template <typename AttrOrType>
+bool isDialect(AttrOrType at, llvm::StringRef dialectName) {
+  return (at.getDialect().getNamespace() == dialectName);
+}
+
+LogicalResult isLegalAttribute(Operation* op, NamedAttribute const& attr,
```
burmako, line 86: Let's remove the unused "op" parameter. (Same below).

stablehlo/transforms/VhloToVersion.cpp
```diff
@@ -77,6 +78,69 @@ FailureOr<Version> validateTargetVersion(llvm::StringRef versionRef,
...
+  if (typeInterface && isLegalVersionForTarget(typeInterface, targetVersion)) {
+    return success();
+  }
+
+  LLVM_DEBUG(llvm::dbgs() << "failed to legalize type " << type
+                          << " to version " << targetVersion << '\n');
+  return failure();
+}
+
+bool isLegalOpInTargetVersion(Operation* op, Version const& targetVersion) {
```
burmako, line 120: "isLegalOp" to be consistent with the two functions above? (Same for "isLegalVersionForTarget" => "isLegalVersion").

stablehlo/dialect/Version.h
```diff
@@ -59,6 +60,14 @@ class Version {
 };
 
 mlir::Diagnostic& operator<<(mlir::Diagnostic& diag, const Version& version);
+llvm::raw_ostream& operator<<(llvm::raw_ostream& os, const Version& version);
+
+template <typename VersionedInterface>
+bool isLegalVersionForTarget(VersionedInterface& interface,
```
burmako, line 66: Given that this function is only used in VhloToVersion.cpp, let's move it to that file.

stablehlo/transforms/VhloToVersion.cpp
```diff
@@ -77,6 +78,69 @@ FailureOr<Version> validateTargetVersion(llvm::StringRef versionRef,
...
+  if (attrInterface && isLegalVersionForTarget(attrInterface, targetVersion)) {
+    return success();
+  }
+
+  LLVM_DEBUG(llvm::dbgs() << "failed to legalize attribute " << attr.getName()
+                          << " to version " << targetVersion);
+  return failure();
+}
+
+LogicalResult isLegalType(Operation* op, Type const& type,
```
burmako, line 103: I don't think it's idiomatic to use `const` or `&` with Type. The former is because https://mlir.llvm.org/docs/Rationale/UsageOfConst/, and the latter is because Type is lightweight enough to be passed by value.

burmako (changes requested):

stablehlo/dialect/VhloBase.td
GleasonK: Fixed. Was also the same typo in `STABLEHLO_CUSTOM_CALL_API_VERISON_UNSPECIFIED`, fixed that too.

stablehlo/dialect/VhloAttrs.td
GleasonK: > If removing default values for minVersion and maxVersion sounds like too much, perhaps let's remove the default value for minVersion

I was thinking this. Lets do this in a follow up change, will be easier to review in isolation. Created  #779

stablehlo/transforms/VhloToVersion.cpp
GleasonK: Ah, didn't realize that went stale. Used to use `op->emitError` but this is called during legalization so the intermediate state may be illegal but end result it fine, so we don't want to emit error.

stablehlo/dialect/Version.h
GleasonK: Fixed everywhere. Previous company had the opposite coding standard, to avoid the contingency in the rule: "const applies to the thing left of it. ~If there is nothing on the left then it applies to the thing right of it.~"

Fixed this and all other params to use `const ClassName&`.

stablehlo/transforms/VhloToVersion.cpp
GleasonK: Hadn't seen that doc before, thanks! Fixed.

stablehlo/tests/vhlo_to_version_downgrade_invalid.mlir
GleasonK: > using unregistered attributes counts as an unspecced feature

I could see unregistered attributes being a silent compatibility-killer. If anything changes about an unregistered attr, the file won't deserialize. With that, there were two possible approaches:

1. Enforce a disallowance of unregistered attributes. This felt too strict since it seems like there _have_ been use cases for unregistered attrs in the past.
2. Don't worry about the attrs, so long as their values are representable and valid in VHLO. 

I went with approach 2. If we wanted to go with approach 1, I would want to make sure we are explicit that that is an unpsecced feature.

> won't achieve much because the test will fail due to a different reason

Outside of VHLO downgrade errors, this does not fail. Aside from the unregistered operation at the bottom of this file, the file if fully round-trippable through stablehlo-opt.

GleasonK: In the meantime, I added a comment to the test to explain that rationale.

stablehlo/dialect/VhloAttrs.td
GleasonK: Done. Should be fairly uncommon since this only applies to the hardcoded versions in TD files, but is a negligible performance hit for a better user experience. Generated code now reads: `"invalid version <version> in <(op|type|attr)-name>"`

```
  mlir::vhlo::Version getMinVersion() {
    auto version = mlir::vhlo::Version::fromString("0.3.0");
    if (failed(version)) llvm_unreachable("invalid version 0.3.0 in ArgResultAlias");
    return *version;
  }
```

GleasonK:
Rebased on main to fix merge conflict.

stablehlo/transforms/VhloToVersion.cpp
```diff
@@ -77,18 +78,87 @@ FailureOr<Version> validateTargetVersion(llvm::StringRef versionRef,
   return targetVersion;
 }
 
+template <typename AttrOrType>
+bool isDialect(AttrOrType at, llvm::StringRef dialectName) {
```
burmako, line 82: `isFromDialect`? During the previous review, the name looked off to me, but I couldn't come up with a better idea. Now I think I have.

stablehlo/tests/vhlo_to_version_downgrade_invalid.mlir
burmako: "I could see unregistered attributes being a silent compatibility-killer". Let's open a ticket to get to the bottom of this. I agree that it's too early to disallow unregistered attributes in VHLO while we have stuff like `mhlo.sharding` and about a handful of other public attributes unspecced, but could we perhaps have an allowlist for now and ban everything that's not there?

"In the meantime, I added a comment to the test to explain that rationale". Understood. After our discussion on a somewhat similar MLIR-HLO PR recently, I have another concern though. Semantically, having an `output_operand_alias` attribute on an `add` doesn't make sense, which doesn't make this a great test in my opinion. Perhaps we can drop this test for now, and once we have that allowlist (in a follow-up PR), we could introduce a more natural negative, e.g. a unit attribute whose name is not on the list?

burmako (changes requested):

stablehlo/tests/vhlo_to_version_downgrade_invalid.mlir
GleasonK: "but could we perhaps have an allowlist for now and ban everything that's not there"
Opened #784. We would need to version this allowlist as well. The alternative is supporting unregistered attrs so long as their values are supported by VHLO, which feels easier to document, use, and maintain.

"Perhaps we can drop this test for now, and once we have that allowlist (in a follow-up PR), we could introduce a more natural negative"
I'd propose the opposite, VHLO needs testing of unregistered attributes, and I'm not sure that allow-list is the proper path forward just yet. I'll change this to an unregistered attr of a custom_call since that makes a little more sense, but this test point has little to do with op semantics, nor the implementation defined semantics of unregistered attributes. This TP is more akin to [mhlo_bytecode_customizations.mlir](https://github.com/tensorflow/mlir-hlo/blob/984b124ecf96d3c0d1d447ffb70c1baca3d9868c/tests/Dialect/mhlo/mhlo_bytecode_customizations.mlir) - nonsense for the sake of testing something else entirely. In this case, we are testing that serialization versioning prevents the legalization of VHLO for a version that would not understand how to deserialize it. If this were to sneak through, even though it is nonsense and may be dropped, it would break deserialization entirely.

Per our discussion on the MLIR-HLO PR, I realized via test cases like this that we are able to add unregistered attributes into `func.func` attributes, which completely violates compatibility currently, re-enforcing the need to fork modularity ops, and have TPs like this for VHLO. While the need is much less for MHLO->StableHLO, I noted that it is _very_ important to test in VHLO, given its use case, and since this _is_ valid (but meaningless) MLIR.

GleasonK: Heh - turns out that this attr does sneak through when used as an unregistered attribute of `custom_call` currently. Will push a fix for that shortly.

stablehlo/tests/vhlo_to_version_downgrade_invalid.mlir
```diff
@@ -21,6 +21,22 @@ func.func @custom_call_v2_with_output_operand_alises(%arg0 : tensor<f32>) -> ten
 
 // -----
 
+/// Unregistered attributes are permitted so long as their value is
+// representable and valid in VHLO in the target version.
+func.func @custom_call_v2_with_output_operand_alises_unregistered(%arg0 : tensor<f32>) -> tensor<f32> {
```
burmako, line 26: Typo: "alises".

stablehlo/tests/vhlo_to_version_downgrade_invalid.mlir
```diff
@@ -21,6 +21,22 @@ func.func @custom_call_v2_with_output_operand_alises(%arg0 : tensor<f32>) -> ten
 
 // -----
 
+/// Unregistered attributes are permitted so long as their value is
```
burmako, line 24: Typo: extraneous `/`.

stablehlo/transforms/VhloToVersion.cpp
burmako: This doesn't look addressed. Forgot to push changes?

stablehlo/tests/vhlo_to_version_downgrade_invalid.mlir
burmako: Sounds good.

burmako (changes requested):

PR #686, opened 2022-12-03
Add type inference for DynamicUpdateSliceOp

ghpvnist:
This PR deletes current `verify` and merge with the new `inferReturnTypeComponents()` function.
Added additional verifications to cover (C3), (C4), (C6) from the spec.
Added additional tests to cover all constraints.
closes #656

PR #757, opened 2022-12-13
Support dynamic shapes in DynamicUpdateSliceOp's type inference

burmako:
https://github.com/openxla/stablehlo/pull/686 implemented type inference for DynamicUpdateSliceOp using the currently statically-shaped spec. To support the previous dynamically-shaped uses of DynamicUpdateSliceOp, this PR adds support for dynamic shapes as well.

### DIFF
--- a/stablehlo/dialect/Base.h
+++ b/stablehlo/dialect/Base.h
@@ -50,6 +50,10 @@ inline static bool isDynamicDimSize(int64_t val) {
   return ShapedType::isDynamic(val);
 }
 
+inline static bool isStaticDimSize(int64_t val) {
+  return !isDynamicDimSize(val);
+}
+
 // Returns true if the given types are the same for the purposes of HLO type
 // inference, accounting for special properties of quantization and sparsity.
 bool isCompatibleForHloTypeInference(Type tp1, Type tp2);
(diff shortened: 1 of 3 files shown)


stablehlo/dialect/Base.h
```diff
@@ -50,6 +50,10 @@ inline static bool isDynamicDimSize(int64_t val) {
   return ShapedType::isDynamic(val);
 }
 
+inline static bool isStaticDimSize(int64_t val) {
```
ghpvnist, line 53: Consider removing the TODO comment in line 48.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -989,17 +990,31 @@ LogicalResult inferDynamicUpdateSliceOp(
...
+        if (updateDim < 0)
+          return emitOptionalError(
+              location, "expects size at dimension ", index,
+              " of update to be non-negative. Got: ", updateDim, ".");
+      }
+    }
 
-  inferredReturnShapes.emplace_back(operandType.getShape(),
-                                    operandType.getElementType());
+  // (C1)
```
ghpvnist, line 1011: Can we move (C1) to the top?

ghpvnist (approved):
Minor comments, but otherwise lgtm.

stablehlo/dialect/TypeInference.cpp
burmako: It's tempting, but we probably shouldn't assume that the code upstream will clean up `inferredReturnShapes` if we end up failing verification. Let's keep C1 after all the checks, not the other way around.

PR #758, opened 2022-12-13
Minor updates to the CustomCallOp spec

burmako:
1) Change api_version from enum to integer constant (since we're calling it implementation-defined anyway).

2) Change the example from using `%inputs0` to the more conventional `%input0`. (Also fix the outfeed example to follow the same convention as well).


subhankarshah (approved):
LGTM

PR #759, opened 2022-12-13
Minor bookkeeping for recently approved RFCs

burmako:
Added "Status", "Initial version", "Last updated" metadata. Also added an FP8 RFC which redirects to the OpenXLA RFC to keep track of it.


GleasonK (approved):
LGTM

PR #760, opened 2022-12-13
Fix crashing builders for CaseOp/IfOp/MapOp

zhouxin913:
Shape function can be called directly from autogenerated `build()` function, which may not guarantee the added region(s) in `odsState.regions` to be non-empty. Need check it here to avoid a crash for the ops that need regions in type inference, like `IfOp/CaseOp/MapOp`.

An alternative way is to consider disable the autogenerated `build()` at all: so far we have no strong evidence that this is a better way and have not find a solution to this approach. While, This PR avoids surprised crash in dev of these ops and gives the error instead.

Note: in the tests of `ops_stablehlo.mlir`, a ODS-valid op must have at least 1 block in each region: so that this error will not trigger in the tests.

Doc is updated separately in https://github.com/openxla/stablehlo/pull/763


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2878,6 +2878,12 @@ LogicalResult MapOp::inferReturnTypeComponents(
     MLIRContext*, Optional<Location> location, ValueShapeRange operands,
     DictionaryAttr attributes, RegionRange regions,
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
+  // This function can be called directly from auto-gen builder(), and does not
+  // guarantee its arg `odsState.regions` to be non-empty. Need check it here to
+  // avoid `adaptor.getComputation()` crash.
+  if (regions.empty())
+    return emitOptionalError(location, "expect non-empty computation region");
```
burmako, line 2885: I think that we need a different kind of safeguard - the one that checks that the individual region is not `empty()`. The resultless builder does add regions, e.g.:

```C++
void MapOp::build(::mlir::OpBuilder &odsBuilder, ::mlir::OperationState &odsState, ::mlir::ValueRange inputs, ::mlir::DenseIntElementsAttr dimensions) {
  odsState.addOperands(inputs);
  odsState.addAttribute(getDimensionsAttrName(odsState.name), dimensions);
  (void)odsState.addRegion();

  ::llvm::SmallVector<::mlir::Type, 2> inferredReturnTypes;
  if (::mlir::succeeded(MapOp::inferReturnTypes(odsBuilder.getContext(),
                odsState.location, odsState.operands,
                odsState.attributes.getDictionary(odsState.getContext()),
                odsState.regions, inferredReturnTypes)))
    odsState.addTypes(inferredReturnTypes);
  else
    ::llvm::report_fatal_error("Failed to infer result type(s).");
}
```

What the resultless builder doesn't is populating these regions with operations, and that's something that we'll need to check for.

burmako, line 2885: Also, can we do these checks in `inferMapOp` and `inferConditionalOp`, so that MHLO can benefit from them as well?

burmako (changes requested):

stablehlo/dialect/StablehloOps.cpp
zhouxin913: Oh thanks, fixed.

zhouxin913: Fixed.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -627,6 +627,18 @@ LogicalResult verifyReduceWindowOpInputsAndInferWindow(
   return success();
 }
 
+// Shape function can be called directly from autogenerated `build()` function,
+// which may not guarantee the added region(s) in `odsState.regions` to be
+// non-empty. Need check it here to avoid a crash for the ops that need regions
+// in type inference, like `IfOp/CaseOp/MapOp`.
```
burmako, line 633: "like `IfOp/CaseOp/MapOp`" => "i.e.`IfOp/CaseOp/MapOp` to indicate that this is an exhaustive list.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -627,6 +627,18 @@ LogicalResult verifyReduceWindowOpInputsAndInferWindow(
...
+// An alternative way is to consider disable the autogenerated `build()` at all.
```
burmako, line 634: Let's drop this sentence. We've made a decision to not go this way, so there's no need to talk about it in the code to avoid confusing the readers.

burmako (changes requested):

stablehlo/dialect/TypeInference.cpp
zhouxin913: Sure.

burmako (approved):

Issue #762, opened 2022-12-13
Update type inference doc about regions


zhouxin913:
There has been several design changes about the type inference for ops with regions. The doc is out of date and need update.

PR #763, opened 2022-12-13
Update type inference doc about regions

zhouxin913:
Fix https://github.com/openxla/stablehlo/issues/762

1. Clarify the updated split rules among verifiers and shape function
2. Clarify how to write / maintain the tests.

Also add @sdasgup3 as CC who is in interested.


docs/type_inference.md
```diff
@@ -26,12 +26,21 @@ Both
...
 
-may have verification code to check operands/attributes/results. An ideal split would be that: let the verifiers check the operands/attributes, then let shape functions to calculate inferred result types and check the compatibility against the real result types. However, in reality this split has a few problems:
+may have verification code to check operands/attributes/results. An initial split would be that: let the verifiers check the operands/attributes, then let shape functions only calculate inferred result types and check the compatibility against the real result types. However, in reality this split has a few problems:
 
-1. Duplicated code: for example in verifiers we do some processing on the operands then verify some intermediate results, then in shape functions these intermediate results are useful to infer the final results. These intermediate results have to be calculated twice.
-2. Maintenance burden: as verifications of an op are contained in two different methods.
+- The shape function can be called by the auto-gen `builder()` alone, without calling the verifier first. So the related inputs must be verified in shape function as well.
```
burmako, line 31: "the autogenerated `build()` functions"

docs/type_inference.md
```diff
@@ -26,12 +26,21 @@ Both
...
+- Duplicated code: for example in verifiers we do some processing on the operands then verify some intermediate results, then in shape functions these intermediate results are useful to infer the final results. These intermediate results have to be calculated twice.
+- Maintenance burden: as verifications of an op are contained in two different methods.
+
+The solution is as follows:
+
+1. **For most ops w/o regions** (like PadOp):
```
burmako, line 37: Let's write "\`PadOp\`" for consistency with the second bullet.

burmako, line 37: "without regions". Let's expand the abbreviation here since it doesn't save that much typing.

docs/type_inference.md
```diff
@@ -26,12 +26,21 @@ Both
...
+Put all the verification code into the shape functions, and discard verifiers totally.
+
+2. **For ops w/ regions** (like `ReduceOp/IfOp`, a full list is [here](https://github.com/openxla/stablehlo/pull/401)): the `builder()` of the op can be created before all the components are in place, for [example](https://github.com/tensorflow/mlir-hlo/blob/master/mhlo/transforms/mhlo_canonicalize_reduction/mhlo_canonicalize_reduction.cc#L221), the `ReduceOp` is created without regions and soon the shape functions are used.
```
burmako, line 40: "with regions".

burmako, line 40: "the autogenerated builders don't take regions as parameters, so if these builders involve type inference, then the shape function will be called with empty regions".

burmako, line 40: Use a permalink here: https://github.com/tensorflow/mlir-hlo/blob/129ae36971a9e3e110d8b91b91a150942d13ff81/mhlo/transforms/mhlo_canonicalize_reduction/mhlo_canonicalize_reduction.cc#L221.

docs/type_inference.md
```diff
@@ -26,12 +26,21 @@ Both
...
+    1. If the regions are not needed for type inference (like `ReduceOp`), put the region related verification logic in verifiers. Duplicate some code if it is inevitable.
+    2. If the regions are needed for type inference (`IfOp/CaseOp/MapOp`), the shape function must verify the regions are not empty explicitly, even though the ODS may already guarantee its existence in the Op definition.
```
burmako, line 42: "If the regions are needed for type inference (`IfOp/CaseOp/MapOp`), then additionally the shape function...".

docs/type_inference.md
```diff
@@ -43,4 +52,14 @@ But stay careful about the missing pieces: for example, if the op contains the t
 
 **Where do we put tests for verifiers and type inference?**
 
-Verifications are kept in [ops\_stablehlo.mlir](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/ops_stablehlo.mlir) and [infer\_stablehlo.mlir](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/infer_stablehlo.mlir). If an op is complicated and could contain a lot of tests, consider adding a separate test file named `verify_<op_name>.mlir` or` verify_<your_topic>.mlir` within the same folder.
+[ops\_stablehlo.mlir](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/ops_stablehlo.mlir) contains the positive cases of ops, and (at least) 1 negative test for every verification error. It is also able to check the inferred return type is **compatible** (not the same!) as the real result type.
+
+[infer\_stablehlo.mlir](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/infer_stablehlo.mlir) verifies the existence of the shape function of an op by line with `hlo_test_infer.get_return_type_components"(%x):...` and the inferred type matches exactly as expected. One positive test per op in general.
+
+### What to do
```
burmako, line 59: Let's add a newline between the header and the text.

docs/type_inference.md
```diff
@@ -43,4 +52,14 @@ But stay careful about the missing pieces: for example, if the op contains the t
...
+When implement or revisit the verifier and/or shape function of an op:
```
burmako, line 60: "When implementing or revisiting...".

burmako (changes requested):

docs/type_inference.md
```diff
@@ -26,12 +26,21 @@ Both
...
+- Duplicated code: for example in verifiers we do some processing on the operands then verify some intermediate results, then in shape functions these intermediate results are useful to infer the final results. These intermediate results have to be calculated twice.
+- Maintenance burden: as verifications of an op are contained in two different methods.
+
+The solution is as follows:
+
+1. **For most ops w/o regions** (like PadOp):
+Put all the verification code into the shape functions, and discard verifiers totally.
+
+2. **For ops w/ regions** (like `ReduceOp/IfOp`, a full list is [here](https://github.com/openxla/stablehlo/pull/401)): the `builder()` of the op can be created before all the components are in place, for [example](https://github.com/tensorflow/mlir-hlo/blob/master/mhlo/transforms/mhlo_canonicalize_reduction/mhlo_canonicalize_reduction.cc#L221), the `ReduceOp` is created without regions and soon the shape functions are used.
+    1. If the regions are not needed for type inference (like `ReduceOp`), put the region related verification logic in verifiers. Duplicate some code if it is inevitable.
```
smit-hinsu, line 41: why not suggest use of verifyRegions so that verification that is part of type inference doesn't need to happen again? by the time verifyRegions is invoked, type inference would have been done already. This should help avoid most of the duplication.

docs/type_inference.md
zhouxin913: Thanks!

docs/type_inference.md
zhouxin913: Hi Smit, thanks for the suggestions: `verifyRegions()` seems not used in StableHLO / MHLO yet, but it's a good option. For the `ReduceOp`, we could consider put the region-related verification (link below)  into a new `verifyRegions()` method, but as you may notice, it relies on a lot of intermediate result from other parts of code above it. It it not clear enough for me for the moment to split them further.

As the crash that https://github.com/openxla/stablehlo/pull/760 and this PR trying to address is an P0 issue, I will submit after get approval. But we may continue the discussion in a new bug if necessary. 

region-related verification in `ReduceOp`: https://github.com/tensorflow/mlir-hlo/blame/master/stablehlo/stablehlo/dialect/TypeInference.cpp#L1551

zhouxin913: Mark as resolved for now, and still open to discussion in the future.

burmako (approved):

Issue #674, opened 2022-12-02
Add forked types and attributes for VHLO


GleasonK:
### Request description

VHLO (Versioned StableHLO, #278) is the portable serialization dialect that will be used for providing compatible StableHLO artifacts across versions. An important aspect of this is to control the serialization of all upstream types used, as such VHLO should only use forks of all upstream types used in the dialect.

PR #765, opened 2022-12-13
Use AnyType and AnyAttr for all types and attributes in VHLO

### LINKED ISSUE #734: Consider migrating VHLO to use generic datatypes (AnyType, AnyAttr)
### Request description

VHLO should be a layout representation, as such it should use very generic typing. This also allows us to separate op versioning from type versioning. I.e. if Tensor changes. all ops using tensor don't need to be updated as well, the tensor types just need to be upgrade/downgrade-able.

GleasonK:
VHLO is intended to be a representation of the IR format, not an exact typed dialect, as such use AnyType/AnyAttr everywhere. This change makes use of AnyType for all types.

The definition of AnyType can be updated to only include VHLO types, `AnyVhloType`, but this isn't necessary and should follow the work regarding forked types and attrs (#674).

Could not use `AnyAttr` for `DefaultValued` fields since the constructor of the attribute is required to accept the default value, so there are some remnants of `DefaultValuedStringAttr`. I also did not update `Shape_WitnessType` since that type needs to be forked. That update will come in the forking types PR.

Closes #734.

Note this PR is built on top of versioned attributes/types (#736) and should be reviewed/submitted after. Can [use this diffbase](https://github.com/openxla/stablehlo/pull/765/files/cee7cd6ca7c96525239d51e199a079348054d221..0c0eb99b94bbd1d5883c1fbaeb83617564aa96d9) or wait until other is merged to review.


stablehlo/dialect/VhloOps.td
```diff
@@ -65,246 +65,246 @@ class VHLO_Op<string mnemonic, string minVersion = "0.3.0", string maxVersion =
...
   );
   let assemblyFormat = "$results attr-dict (`:` type($results)^)?";
 }
 
 // Communication op definitions.
 def VHLO_InfeedOpV1 : VHLO_Op<"infeed"> {
   let arguments = (ins
-    VHLO_Token:$token,
+    VHLO_AnyType:$token,
     DefaultValuedStrAttr<StrAttr, "">:$infeed_config,
```
burmako, line 289: Can we change default valued attrs to `OptionalAttr<VHLO_AnyAttr>`? (Here and below).

stablehlo/dialect/VhloOps.td
```diff
@@ -65,246 +65,246 @@ class VHLO_Op<string mnemonic, string minVersion = "0.3.0", string maxVersion =
...
     OptionalAttr<ArrayAttr>:$layout
```
burmako, line 290: `OptionalAttr<VHLO_AnyAttr>`.

stablehlo/dialect/VhloOps.td
```diff
@@ -65,246 +65,246 @@ class VHLO_Op<string mnemonic, string minVersion = "0.3.0", string maxVersion =
...
   let results = (outs VHLO_Token);
 }
 
 def VHLO_SendOpV1 : VHLO_Op<"send"> {
   let arguments = (ins
-    Variadic<VHLO_Tensor>:$inputs,
-    VHLO_Token:$token,
+    Variadic<VHLO_AnyType>:$inputs,
+    VHLO_AnyType:$token,
     VHLO_ChannelHandle:$channel_handle,
```
burmako, line 308: `VHLO_AnyAttr`.

stablehlo/dialect/VhloOps.td
```diff
@@ -336,86 +336,86 @@ def VHLO_AfterAllOpV1 : VHLO_Op<"after_all"> {
...
   let arguments = (ins
-    VHLO_Tensor:$operand,
-    I64Attr:$all_gather_dim,
-    I64ElementsAttr:$replica_groups,
-    OptionalAttr<VHLO_ChannelHandle>:$channel_handle,
+    VHLO_AnyType:$operand,
+    VHLO_AnyAttr:$all_gather_dim,
+    VHLO_AnyAttr:$replica_groups,
+    OptionalAttr<VHLO_AnyAttr>:$channel_handle,
     UnitAttr:$use_global_device_ids
```
burmako, line 373: `VHLO_AnyAttr`. (Here and for all other occurrences of `use_global_device_ids`).

stablehlo/dialect/VhloOps.td
```diff
@@ -424,194 +424,194 @@ def VHLO_ReduceOpV1 : VHLO_Op<"reduce", "0.3.0", "current", [SameVariadicOperand
...
   );
-  let results = (outs VHLO_Tensor);
+  let results = (outs VHLO_AnyType);
 }
 
 def VHLO_CustomCallOpV1 : VHLO_Op<"custom_call", "0.3.0", "0.3.0"> {
   let arguments = (ins
-    Variadic<VHLO_TensorOrTokenOrTuple>:$inputs,
+    Variadic<VHLO_AnyType>:$inputs,
     StrAttr:$call_target_name,
```
burmako, line 615: `VHLO_AnyAttr`.

stablehlo/dialect/VhloOps.td
```diff
@@ -620,247 +620,247 @@ def VHLO_CustomCallOpV1 : VHLO_Op<"custom_call", "0.3.0", "0.3.0"> {
...
   let arguments = (ins
-    Variadic<VHLO_Tensor>:$inputs,
-    Variadic<VHLO_Tensor>:$init_values,
-    I64ElementsAttr:$window_dimensions,
+    Variadic<VHLO_AnyType>:$inputs,
+    Variadic<VHLO_AnyType>:$init_values,
+    VHLO_AnyAttr:$window_dimensions,
     // If strides or dilations attributes are missing then the default value is
     // one for each of the operand dimensions. Similarly, padding values are zero
     // for both low and high in each of the dimensions, if not specified.
```
burmako, line 841: Drop the comment?

stablehlo/dialect/VhloOps.td
```diff
@@ -869,95 +869,95 @@ def VHLO_OptimizationBarrierOpV1 : VHLO_Op<"optimization_barrier"> {
...
 
 def VHLO_ComputeReshapeShapeOpV1 : VHLO_Op<"compute_reshape_shape"> {
-  let arguments = (ins Index:$num_elements, 1DTensorOf<[AnyInteger, Index]>:$dynamic_shape);
-  let results = (outs 1DTensorOf<[AnyInteger, Index]>:$result);
+  let arguments = (ins VHLO_AnyType:$num_elements, VHLO_AnyType:$dynamic_shape);
+  let results = (outs VHLO_AnyType:$result);
 }
 
 def VHLO_CstrReshapableOpV1 : VHLO_Op<"cstr_reshapable"> {
   let results = (outs Shape_WitnessType:$result);
```
burmako, line 959: Not sure I follow the reasoning for why not change this to `VHLO_AnyType`.

stablehlo/dialect/VhloOps.td
```diff
@@ -336,86 +336,86 @@ def VHLO_AfterAllOpV1 : VHLO_Op<"after_all"> {
 }
 
 def VHLO_IfOpV1 : VHLO_Op<"if"> {
-  let arguments = (ins VHLO_PredTensor:$pred);
+  let arguments = (ins VHLO_AnyType:$pred);
   let regions = (region SizedRegion<1>:$true_branch,
```
burmako, line 340: `AnyRegion`? (Here and below).

burmako (changes requested):

stablehlo/dialect/VhloOps.td
GleasonK: `UnitAttr` is a strange one because it is defined by omission or presence.. Removing causes tests to fail that "use_global_device_ids" is unset.

stablehlo/dialect/VhloOps.td
GleasonK: Agree. I was working on this in parallel to forking types and had already forked builtin tensor.

I think it would be good after we have all this forked to update the definition of `VHLO_AnyType` to be any of the _VHLO_ Types, as opposed to any type at all.

stablehlo/dialect/VhloOps.td
GleasonK: Should it be optional or just `VHLO_AnyAttr`? Default valued means it will always have a value?

GleasonK:
Using AnyType/AnyAttr makes downgrade code a little messier. This change is still worth making though.

stablehlo/dialect/VhloOps.td
burmako: Sounds like we no longer need `include "mlir/Dialect/Shape/IR/ShapeBase.td"` in the updated version of VhloOps.td?

stablehlo/dialect/VhloOps.td
burmako: `OptionalAttr<VHLO_AnyAttr>` perhaps?

stablehlo/dialect/VhloOps.td
GleasonK: That's exactly what I had in mind. Will make those changes

stablehlo/dialect/VhloOps.td
GleasonK: removed.

stablehlo/dialect/VhloOps.td
GleasonK: I'm a bit surprised that worked but it worked. `UnitAttr` is not covered by `AnyAttr`, but is covered by `OptionalAttr<AnyAttr>`. I suppose that fits its definition. Resolved

GleasonK:
Addressed feedback. Rebased on main to resolve merge conflicts.

burmako (approved):

Issue #731, opened 2022-12-09
Revisit the spec for SelectAndScatter


sdasgup3:
### Request description

**Problem** what happens if the entire window of the `reduce_window` phase of `select_and_scatter` consists of padding values?

Specifically, per the [spec](https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md#stablehloselect_and_scatter)
what is going to the behavior of `selected_values = reduce_window_without_init(...)` when the data consists of only padding values. 

Ideally, The expected outcome sould be `selected_values = []`. But `reduce_window_without_init` does not convey that clearly.

A suggestion:

As per the hlo  implementation of SelectAndScatter,
[this](https://github.com/tensorflow/tensorflow/blob/2a91b8212a4d00f00bcbd7feb83c87687b1f8286/tensorflow/compiler/xla/hlo/evaluator/hlo_evaluator_typed_visitor.h#L1774) is the callback function which is used to go through every valid operand index and book-keep the most updated selection based on the operand indices seen so far.

As expected, in case where there is no valid index, the above callback is [never applied](https://github.com/tensorflow/tensorflow/blob/2a91b8212a4d00f00bcbd7feb83c87687b1f8286/tensorflow/compiler/xla/hlo/evaluator/hlo_evaluator_typed_visitor.h#L2405), leaving the result of selection  [empty](https://github.com/tensorflow/tensorflow/blob/2a91b8212a4d00f00bcbd7feb83c87687b1f8286/tensorflow/compiler/xla/hlo/evaluator/hlo_evaluator_typed_visitor.h#L1772).

One of the challenges that I found is: for the case of [selectAndScatter](https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md#stablehloselect_and_scatter) we define `reduce_window_without_init`  saying that it is based on `reduce_window`, but the reduce_window's spec [relies very much on "init_values"](https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md#stablehloreduce_window)  and thereby,  it might be hard to imagine how the *_withpout_init* version would look like.

How about we define `selectAndScatter` using `reduce_without_init`: Including just the valid indices in a window and applying  `reduce_without_init` on that. The valid indices will be along [this](https://github.com/tensorflow/tensorflow/blob/2a91b8212a4d00f00bcbd7feb83c87687b1f8286/tensorflow/compiler/xla/hlo/evaluator/hlo_evaluator_typed_visitor.h#L2391) line.

PR #771, opened 2022-12-14
Spec the behavior of select in SelectAndScatterOp

sdasgup3:
fixes https://github.com/openxla/stablehlo/issues/731

As per the formulation,  there will not be a valid `index` in `selected_values`, in case the associated window on which the `select` reduction needs to be computed is all padding,

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -4106,11 +4106,11 @@ The following diagram shows how elements in `results[k]` are computed from
 
 ![](images/spec/reduce_window.svg)
 
-More formally, `results[:][result_index] = reduce(windows, init_values, axes(inputs[:]), body)` where:
+More formally, `results[:][result_index] = reduce(window, init_values, axes(inputs[:]), body)` where:
 
   * `padded_inputs = pad(inputs[:], init_values[:], padding[:, 0], padding[:, 1], base_dilations)`.
   * `window_start = result_index * window_strides`.
-  * `windows = slice(padded_inputs[:], window_start, window_start + window_dimensions, window_dilations)`.
+  * `window = slice(padded_inputs[:], window_start, window_start + window_dimensions, window_dilations)`.
 
 ### Inputs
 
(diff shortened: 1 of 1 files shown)


docs/spec.md
```diff
@@ -4106,11 +4106,11 @@ The following diagram shows how elements in `results[k]` are computed from
 
 ![](images/spec/reduce_window.svg)
 
-More formally, `results[:][result_index] = reduce(windows, init_values, axes(inputs[:]), body)` where:
+More formally, `results[:][result_index] = reduce(window, init_values, axes(inputs[:]), body)` where:
 
   * `padded_inputs = pad(inputs[:], init_values[:], padding[:, 0], padding[:, 1], base_dilations)`.
   * `window_start = result_index * window_strides`.
-  * `windows = slice(padded_inputs[:], window_start, window_start + window_dimensions, window_dilations)`.
+  * `window = slice(padded_inputs[:], window_start, window_start + window_dimensions, window_dilations)`.
```
burmako, line 4113: There can be multiple inputs, so I think it is reasonable to say that there are multiple windows?

docs/spec.md
```diff
@@ -4770,22 +4770,27 @@ The following diagram shows how elements in `result` are computed from
 
 More formally:
 
- * `selected_values = reduce_window_without_init(...)` with the following inputs:
-   * `inputs` $=$ [ `operand` ].
-   * `window_dimensions`, `window_strides`, and `padding` which are used as is.
-   * `base_dilations` $=$ `windows_dilations` $=$ `[1, ..., 1]`.
+ * `selected_values[index] = reduce_without_init(window, axes(operand), body)` where:
```
burmako, line 4773: I think the current problem is not caused by the formalism using `reduce_window` (which I think fits into this formalism nicely, allowing to avoid speccing how to compute windows). The problem is that `selected_values[index]` needs to be populated by something for each `index`, and it's unclear what that something should be if the entire window consists of padding values. This formulation doesn't address this problem either.

burmako (changes requested):

sdasgup3:
Let us close the PR for now.  I am planning to revisit the formulation of this with more thinking. 

From https://github.com/openxla/stablehlo/issues/731, 
```
Ideally, The expected outcome should be selected_values = []. But reduce_window_without_init does not convey that clearly.
```

I note here that not finalizing on how to concisely specify the details requested by #731 should not block the interpreter implementation.

Issue #730, opened 2022-12-09
Spec the behavior of window_reversal in ConvolutionOp


sdasgup3:
### Request description

As per https://github.com/openxla/stablehlo/pull/441#discussion_r1041744301, we decided to spec the `window_reversal` component of the op in stages.

PR #772, opened 2022-12-14
add spec for window_reversal

sdasgup3:
fixes https://github.com/openxla/stablehlo/issues/730

[Reference](https://github.com/tensorflow/tensorflow/blob/ffefd39ca3f3ebdab56883dd962c2ba7a2f80eeb/tensorflow/compiler/xla/hlo/evaluator/hlo_evaluator_typed_visitor.h#L1024)


docs/spec.md
```diff
@@ -1918,7 +1918,8 @@ If `feature_group_count = 1` and `batch_group_count = 1`, then for all
   * `padded_lhs = pad(lhs, 0, lhs_padding[:, 0], lhs_padding[:, 1], lhs_base_dilations)`.
   * `lhs_window_start = lhs_shape(0, output_spatial_index, 0) * lhs_window_strides`.
   * `lhs_window = slice(padded_lhs, lhs_window_start, lhs_window_start + lhs_window_dimensions, lhs_window_dilations)`.
-  * `dot_product = dot_general(lhs_window, rhs,
+  * `reversed_rhs = reverse(rhs, [kernel_spatial_dimensions[dim] for dim in [0, size(window_reversal) and window_reversal[dim] = true])`.
```
burmako, line 1921: Given that the corresponding input is called `window_reversal`, let's specifically reverse the window? I don't think it matters whether we reverse lhs or rhs, so the proposed formalism is mathematically correct too, but I think it'll read better if it'll be `reversed_lhs_window` and `rhs` rather than the other way around.

burmako (changes requested):

burmako (approved):

PR #776, opened 2022-12-15
Use empty() method to check for emptiness

ghpvnist:


sdasgup3 (approved):

burmako:
Since this is "Migrate from MHLO", what is the corresponding MLIR-HLO commit?

Issue #778, opened 2022-12-16
Disable shape functions for ConvolutionOp and DotOp


zhouxin913:
For many ops, not all components of return type can be inferred, for example,
- The return element type of DotOp / DotGeneralOp
- The return shape of ReshapeOp / RngBitGeneratorOp 

We could add the type inference in ODS, with a less restricted isCompatibleReturnTypes() which only checks the compatibility of some components while ignoring others. However, this will crash the auto generated Python API and JAX. So that we should only implement verifiers for this ops and call the shape function if needed instead.

burmako:
"However, this will crash the auto generated Python API and JAX". To elaborate on this: these APIs will not crash, but they will be unusable from JAX because they produce wrong results (since e.g. shape functions for ConvolutionOp, DotOp and DotGeneralOp don't take preferred element type into account).

ghpvnist:
With migrating MHLO changes to StableHLO and some StableHLO changes migrating to HLO, there is a conflict with a negative test for DotOp's precision_config verification. #780 removes this test, which may be added later once this lands.

PR #441, opened 2022-11-04
Add spec for ConvolutionOp

sdasgup3:
fixes #359 

The PR addresses the followings:
1. Spec of ConvolutionOp
2. Clarify the semantics of `precision_config` : The precision_config parameter is a array of enums without any constraint on its size. Need to resolve this.  
    - update: Added constraints on the parameter.  With that the verifier is in sync with this spec. Also added https://github.com/openxla/stablehlo/issues/445 for further exploration.
4. Fix https://github.com/openxla/stablehlo/pull/360#issuecomment-1282808168
5. Avoid disabling clang formatting in StablehloOps.cpp.
6. Address https://github.com/openxla/stablehlo/issues/399

Only missing peice:

The constraint between output feature size and input batch size. Working on getting a better understanding on this: Done

Type inference should be "revisit" as well because of #600.

PR #780, opened 2022-12-16
Remove DotOp verifier

### LINKED ISSUE #682: Consider remove DotOp verifier
### Request description

This is redundant with the automatic verification from `InferTypeOpInterface` and also doesn't allow a more dynamic output type than the input types, unlike all other mhlo ops.

https://github.com/tensorflow/mlir-hlo/commit/e2506b2bd10911f834db447c636be5963d1af107

ghpvnist:
Removes negative DotOp test since the verifier is removed.
closes #682

### DIFF
--- a/stablehlo/tests/ops_stablehlo.mlir
+++ b/stablehlo/tests/ops_stablehlo.mlir
@@ -1513,7 +1513,7 @@ func.func @dot_more_dynamic_output_type(%arg0: tensor<3xf32>, %arg1: tensor<?x3x
 // -----
 
 func.func @dot_illegal_result_type(%arg0: tensor<?x3xf32>, %arg1: tensor<3xf32>) -> tensor<3x?xf32> {
-// expected-error@+1 {{'mhlo.dot' op inferred type(s) 'tensor<?xf32>' are incompatible with return type(s) of operation 'tensor<3x?xf32>'}}
+// expected-error@+1 {{'stablehlo.dot' op inferred type(s) 'tensor<?xf32>' are incompatible with return type(s) of operation 'tensor<3x?xf32>'}}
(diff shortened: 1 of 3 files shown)


ghpvnist:
Depending on the integrate/PR merge timeline, this change may not need to be propagated to MHLO.

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -1505,16 +1497,15 @@ func.func @dot_bad_precision_config(%arg0: tensor<2x2xi32>, %arg1: tensor<2x2xi3
 
 // -----
 
-func.func @dot_illegal_input_type(%arg0: tensor<3xf32>, %arg1: tensor<?x3xf32>) -> tensor<?xf32> {
-  // expected-error@+1 {{Unexpected result type: has 'tensor<?xf32>' but inferred 'tensor<3xf32>' from operands 'tensor<3xf32>' and 'tensor<?x3xf32>}}
+func.func @dot_more_dynamic_output_type(%arg0: tensor<3xf32>, %arg1: tensor<?x3xf32>) -> tensor<?xf32> {
```
sdasgup3, line 1500: It seems that you keep the tests as they are there in https://github.com/tensorflow/tensorflow/blob/82c38106d03c15efe3ee53401e3527348a62a8cb/tensorflow/compiler/xla/mlir_hlo/tests/Dialect/mhlo/ops.mlir#L1099.

No action.

sdasgup3 (approved):

stablehlo/tests/ops_stablehlo.mlir
ghpvnist: The above comment was for removing the test related to DotOp precision_config since the recent change in StableHLO #441, once migrated, will fail in MHLO since DotOp has already been removed there and there are currently discrepancies between the two, which #778 will aim to fix.

Issue #790, opened 2022-12-20
Verify dimension in GetDimensionSizeOp


smit-hinsu:
### Request description

get_dimension_size op doesn't verify dimension attribute and the corresponding dimension size it is within supported limit.

PR #791, opened 2022-12-20
Add spec for GetDimensionSizeOp

smit-hinsu:
Verification is missing two constraints. Filed https://github.com/openxla/stablehlo/issues/790 for that.

RFC discussing this op: https://github.com/openxla/stablehlo/pull/194

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -2833,7 +2833,7 @@ behavior is undefined. More formally, for all `id < jd` from `indices(result)`,
 
 ### Semantics
 
-Returns the size of the given `dimension` of the `operand`.
+Produces the size of the given `dimension` of the `operand`.
 
 ### Inputs
 
@@ -2850,8 +2850,7 @@ Returns the size of the given `dimension` of the `operand`.
 
 ### Constraints
 
-  * (C1) 0 $\le$ `dimension` $\lt$ rank of `operand`. [todo](https://github.com/openxla/stablehlo/issues/790)
-  * (C2) dim(operand, dimension) $\le$ std::numeric_limits<int32_t>::max(). [todo](https://github.com/openxla/stablehlo/issues/790)
+  * (C1) 0 $\le$ `dimension` $\lt$ `rank(operand)`. [todo](https://github.com/openxla/stablehlo/issues/790)
 
 ### Examples
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -2837,10 +2837,10 @@ Returns the size of the given `dimension` of the `operand`.
 
 ### Inputs
 
-| Name      | Type                         |
-|-----------|------------------------------|
-| `operand` | tensor of any supported type |
-| `index`   | constant of type `si64`      |
+| Name          | Type                         |
+|---------------|------------------------------|
+| `operand`     | tensor of any supported type |
+| `dimension`   | constant of type `si64`      |
 
 ### Outputs
 
@@ -2852,7 +2852,6 @@ Returns the size of the given `dimension` of the `operand`.
 
   * (C1) 0 $\le$ `dimension` $\lt$ rank of `operand`. [todo](https://github.com/openxla/stablehlo/issues/790)
   * (C2) dim(operand, dimension) $\le$ std::numeric_limits<int32_t>::max(). [todo](https://github.com/openxla/stablehlo/issues/790)
-  * (C3) `result` is of type `tensor<i32>`
 
 ### Examples
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -413,6 +413,7 @@ syntax.
    * [fft](#stablehlofft)
    * [floor](#stablehlofloor)
    * [gather](#stablehlogather)
+   * [get_dimension_size](#stablehloget_dimension_size)
    * [get_tuple_element](#stablehloget_tuple_element)
    * [if](#stablehloif)
    * [imag](#stablehloimag)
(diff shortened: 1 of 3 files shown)


docs/spec.md
```diff
@@ -2828,6 +2829,43 @@ behavior is undefined. More formally, for all `id < jd` from `indices(result)`,
...
+### Semantics
+
+Returns the size of the given `dimension` of the `operand`.
+
+### Inputs
+
+| Name      | Type                         |
+|-----------|------------------------------|
+| `operand` | tensor of any supported type |
+| `index`   | constant of type `si64`      |
```
sdasgup3, line 2843: `dimension`

docs/spec.md
```diff
@@ -2828,6 +2829,43 @@ behavior is undefined. More formally, for all `id < jd` from `indices(result)`,
...
+
+| Name     | Type                                |
+|----------|-------------------------------------|
+| `result` | 0-dimensional tensor of type `si32` |
+
+### Constraints
+
+  * (C1) 0 $\le$ `dimension` $\lt$ rank of `operand`. [todo](https://github.com/openxla/stablehlo/issues/790)
+  * (C2) dim(operand, dimension) $\le$ std::numeric_limits<int32_t>::max(). [todo](https://github.com/openxla/stablehlo/issues/790)
+  * (C3) `result` is of type `tensor<i32>`
```
sdasgup3, line 2855: seems redundant as they are already mentioned in Outputs table

docs/spec.md
```diff
@@ -2828,6 +2829,43 @@ behavior is undefined. More formally, for all `id < jd` from `indices(result)`,
...
+### Outputs
+
+| Name     | Type                                |
+|----------|-------------------------------------|
+| `result` | 0-dimensional tensor of type `si32` |
+
+### Constraints
+
+  * (C1) 0 $\le$ `dimension` $\lt$ rank of `operand`. [todo](https://github.com/openxla/stablehlo/issues/790)
+  * (C2) dim(operand, dimension) $\le$ std::numeric_limits<int32_t>::max(). [todo](https://github.com/openxla/stablehlo/issues/790)
```
sdasgup3, line 2854: Do we know what is  special about the dimension size of this particular `dimension` axis that https://github.com/tensorflow/tensorflow/blob/963d7a175b0c03b0d7b03286c0d9eb7098c29e5e/tensorflow/compiler/xla/service/shape_inference.cc#L2486 has a dedicated check only for it but not for other axes.

docs/spec.md
smit-hinsu: Fixed

docs/spec.md
smit-hinsu: Removed

docs/spec.md
smit-hinsu: That is because the result type of get_dimension_size and operand type of set_dimension_size is i32.

sdasgup3: makes sense. Thanks!

docs/status.md
```diff
@@ -92,7 +92,7 @@ one of the following tracking labels.
 | fft                      | yes           | revisit      | yes            | yes             | no          |
 | floor                    | yes           | yes          | yes            | yes             | yes         |
 | gather                   | yes           | yes          | yes            | no              | no          |
-| get_dimension_size       | no            | yes*         | yes*           | yes             | no          |
```
sdasgup3, line 95: Shouldn't the type inference be yes

https://github.com/openxla/stablehlo/blob/93e7cf81bf8258021ef5d96c31682ab893cb82da/stablehlo/dialect/StablehloOps.cpp#L1267

sdasgup3 (approved):
LGTM except for one last comment.

docs/status.md
smit-hinsu: updated.

docs/spec.md
```diff
@@ -2828,6 +2829,42 @@ behavior is undefined. More formally, for all `id < jd` from `indices(result)`,
 
 [Back to Ops](#index-of-ops)
 
+## stablehlo.get_dimension_size
+
+### Semantics
+
+Returns the size of the given `dimension` of the `operand`.
```
burmako, line 2836: In the rest of the spec, we're typically saying "produces" to avoid the ambiguity with returning from a function.

docs/spec.md
```diff
@@ -2828,6 +2829,42 @@ behavior is undefined. More formally, for all `id < jd` from `indices(result)`,
...
+
+### Outputs
+
+| Name     | Type                                |
+|----------|-------------------------------------|
+| `result` | 0-dimensional tensor of type `si32` |
+
+### Constraints
+
+  * (C1) 0 $\le$ `dimension` $\lt$ rank of `operand`. [todo](https://github.com/openxla/stablehlo/issues/790)
```
burmako, line 2853: \`rank(operand)\`, similarly to other places in the spec.

docs/spec.md
```diff
@@ -2828,6 +2829,42 @@ behavior is undefined. More formally, for all `id < jd` from `indices(result)`,
...
+  * (C2) dim(operand, dimension) $\le$ std::numeric_limits<int32_t>::max(). [todo](https://github.com/openxla/stablehlo/issues/790)
```
burmako, line 2854: Let's drop C2. The "Constraints" section conventionally contains only the stuff that we can check statically (i.e. stuff that we can put into a verifier or into a shape function in the StableHLO dialect).

burmako (changes requested):

docs/spec.md
smit-hinsu: updated.

docs/spec.md
smit-hinsu: Done. Sent https://github.com/openxla/stablehlo/pull/815 for a few existing places.

docs/spec.md
smit-hinsu: Done.

burmako (approved):

Issue #792, opened 2022-12-20
Nonconforming syntax in multiple markdown files


scottamain:
### What happened?

Although some MD parsers are pretty forgiving (including GitHub's), the current markdown files in stablehlo include a variety of syntax errors that do not conform to the basic [MD syntax rules](https://daringfireball.net/projects/markdown/syntax). Examples include improper indentation for lists, unescaped asterisks signs (meant for bold or italic text), and missing blank lines between headings and paragraphs.

We should follow traditional MD syntax rules to ensure well-structured files that are easy to read, use an agreed-upon structure that can be tested, and that are compatible with more than one MD parser.

The simplest solution is to [markdownlint](https://github.com/markdownlint/markdownlint) and fix various syntax and style issues. 

We can submit fixes for individual files, and make exceptions as appropriate because this tool allows you to disable certain rules in the linter. Then, once the files are cleaned up, we should enable a GitHub Action to run markdownlint on all new pull requests with .md files so we can avoid re-introduction of syntax errors.

burmako:
@scottamain Thank you for looking into this! This will be a nice improvement to the state of the art. We had prior discussions about Markdown linting but haven't had the time to follow up yet.

burmako:
Reopening the ticket because #793 said that it only partially fixes it.

GleasonK:
Not sure if this lint has been run on the `rfcs/` directory as well? I didn't see the Compatibility RFC in #793, and I'm doubtful that I used perfect MD syntax all the way through.

Here are [all md files in the repo](https://github.com/openxla/stablehlo/search?q=filename%3A.md).

scottamain:
Hehe, yeah @GleasonK I didn't do the rfc files in the first CL just to keep my first PR simple. I'm sure you saw I've now cleaned those ones as well.

I think there's just two items remaining before closing this issue (in order):
1. Fix all lint errors in spec.md. I'm happy to do this but I don't want to cause confusing merge conflicts for others because this appears to be a fairly active document. @burmako Should we wait a little longer for this?
2. Enable the GitHub Action to run markdownlint on all .md file PRs.

GleasonK:
> Should we wait a little longer for this?

Spoke with Eugene, totally fine to cause merge conflicts. There is only one open PR touching spec now, which makes this a good time to make that change. That plan sounds good to me.

Got the script running locally, looks like its only spec:

```
[...]

PR #795, opened 2022-12-20
More MD syntax fixes, mostly line length

scottamain:
More fixes for #792 with expanded scope to MD files not under /docs. (Still not touching the spec yet.)

This includes markdownlint fixes that could not be fixed by `markdownlint-cli`, such as line length (80 char), consistent list item indications, proper heading levels, removal of bash "$" character when not necessary ([rationale](https://cirosantilli.com/markdown-style-guide/#dollar-signs-in-shell-code); if the snippet includes multiple commands, then add a line break).

### DIFF
--- a/docs/status.md
+++ b/docs/status.md
@@ -1,4 +1,4 @@
-# StableHLO progress
+# StableHLO status
 
 When bootstrapping StableHLO from MHLO, we have inherited MHLO's implementation
 of many things, including prettyprinting, verification and shape inference.
--- a/docs/status.md
+++ b/docs/status.md
@@ -1,4 +1,4 @@
-## About
+# StableHLO progress
 
 When bootstrapping StableHLO from MHLO, we have inherited MHLO's implementation
 of many things, including prettyprinting, verification and shape inference.
@@ -10,7 +10,7 @@ This live document is for the developers and the users to track the progress on
 various aspects of the opset - specification, verification, type inference,
 pretty printing, interpreter, etc.
 
-### How to use it
+## How to use it
 
 The progress of a StableHLO op, as mentioned in the corresponding row, on a
 particular aspect, as mentioned in the corresponding column, is tracked using
--- a/build_tools/README.md
+++ b/build_tools/README.md
@@ -3,30 +3,38 @@
 To build the code in this repository, you need a clone of the LLVM/MLIR
 git repository:
 
-`$ git clone https://github.com/llvm/llvm-project.git`
+```bash
+git clone https://github.com/llvm/llvm-project.git
+````
 
 You need to make sure you have the right commit checked out in
 the LLVM repository (you need to do this every time you pull from this repo):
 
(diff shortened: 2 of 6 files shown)


docs/status.md
```diff
@@ -1,4 +1,4 @@
-## About
+# StableHLO progress
```
burmako, line 1: How about "StableHLO Status"?

burmako (changes requested):

docs/status.md
GleasonK: Was this made by markdownlint? Or was this made separately because it is more readable?

GleasonK:
This is great - thanks so much for looking into this! 

Excited to have this CI in place. Just a note for if/when you get around to CI, we follow a principle of "bash script as much as possible in GH actions". This was a practice I took from IREE, which allows us to run the CI commands locally as well.

For example: Have a script `./build_tools/github_actions/lint_markdown.sh` which uses `markdownlint` to validate, and have the GH action with 2 steps: 1) install markdownlint on runner. 2) call `lint_markdown.sh`

docs/status.md
scottamain: SGTM, done. I made this change because "About" was an H2 heading, so this document didn't actually have a proper title (markdownlint requires each file have an H1 heading). I chose "progress" based on the sentence in here that says, "The progress of a StableHLO op... is tracked using one of the following labels," and also because "Status" is already used in an H2 heading, but I'm fine with either.

scottamain:
Hey @GleasonK, so does that mean you frown upon using any pre-defined GitHub Actions? Here's my working version with a couple `uses` actions:
https://github.com/scottamain/stablehlo/blob/markdownlint/.github/workflows/lint_docs.yml
I'm happy to abide by whatever style you prefer, so I wonder where you draw the line between what sort of stuff can be done with a pre-defined action (with `uses`) or in our own shell script (is `actions/checkout` okay?).

GleasonK:
I'd say it depends on how easy it is to fix locally - what does a GH actions failure look like and what do we expect a user to do after encountering a failure? If there is an automated / minimalist way to go from non-conformant MD to conformant MD, then should be ok.

Currently we run clang-format and trailing whitespace lint checks.

```
$ ./build_tools/github_actions/lint_clang_format.sh     # Check for failures
$ ./build_tools/github_actions/lint_clang_format.sh -f  # Autofix any failures

$ ./build_tools/github_actions/lint_whitespace_checks.sh     # Check for failures
$ ./build_tools/github_actions/lint_whitespace_checks.sh -f  # Autofix any failures
```

burmako (approved):

scottamain:
Thanks @GleasonK. I definitely appreciate how using a bash script allows users to run the same process locally before sending a PR so you can easily preview the lint results. Likewise, it gives me confidence to merge the script early because we can all experiment with it locally instead of relying on the Action events.

PR #797, opened 2022-12-20
Separate VhloTypes into separate tablegen file

GleasonK:
This is similar to how MHLO manages their custom types in [hlo_ops_typedefs.td](https://github.com/tensorflow/mlir-hlo/blob/master/mhlo/IR/hlo_ops_typedefs.td). 

These typedefs will also grow in upcoming submissions as other types are forked.

This allows us to use `VhloBase.td` as a tablegen target in a bazel build since it does not have complicated dependencies (adding typedefs makes a new dependency to VHLO_Dialect from VhloOps.td).

### DIFF
--- a/stablehlo/dialect/VhloBase.td
+++ b/stablehlo/dialect/VhloBase.td
@@ -17,70 +17,7 @@ limitations under the License.
 #ifndef STABLEHLO_DIALECT_VHLO_BASE
 #define STABLEHLO_DIALECT_VHLO_BASE
 
-include "mlir/IR/AttrTypeBase.td"
 include "mlir/IR/OpBase.td"
-
-//===----------------------------------------------------------------------===//
-// VHLO Versioning Interfaces
-//===----------------------------------------------------------------------===//
-
-def VersionedOpInterface : OpInterface<"VersionedOpInterface"> {
-  let methods = [
-    InterfaceMethod<
-      "Returns the minimum version of the VHLO dialect an op is supported in.",
-      "mlir::vhlo::Version", "getMinVersion">,
-    InterfaceMethod<
-      "Returns the maximum version (inclusive) of the VHLO dialect an op is supported in.",
-      "mlir::vhlo::Version", "getMaxVersion">,
-  ];
-}
-
-def VHLO_VersionedAttrInterface : AttrInterface<"VersionedAttrInterface"> {
-  let cppNamespace = "::mlir::vhlo";
-  let methods = [
-    InterfaceMethod<
-      "Returns the minimum version of the VHLO dialect an attribute is supported in.",
-      "mlir::vhlo::Version", "getMinVersion">,
-    InterfaceMethod<
-      "Returns the maximum version (inclusive) of the VHLO dialect an attribute is supported in.",
-      "mlir::vhlo::Version", "getMaxVersion">,
-  ];
-}
-
-def VHLO_VersionedTypeInterface : TypeInterface<"VersionedTypeInterface"> {
-  let cppNamespace = "::mlir::vhlo";
-  let methods = [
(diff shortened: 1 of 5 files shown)


stablehlo/dialect/VhloBase.td
```diff
@@ -59,43 +59,6 @@ def VHLO_VersionedTypeInterface : TypeInterface<"VersionedTypeInterface"> {
   ];
 }
```
burmako, line 60: Should `VHLO_VersionedTypeInterface` move to `VhloTypes.td` (and the same thing about other `VHLO_VersionedFooInterface` and the corresponding `VhloFoos.td` files).

stablehlo/dialect/VhloBase.td
```diff
@@ -59,43 +59,6 @@ def VHLO_VersionedTypeInterface : TypeInterface<"VersionedTypeInterface"> {
...
-}
-
-//===----------------------------------------------------------------------===//
-// VHLO Type Definitions.
-//===----------------------------------------------------------------------===//
-
-// VHLO is intended to represent the layout only, as such uses AnyType everywhere.
-def VHLO_AnyType : AnyTypeOf<[AnyType]>;
-def VHLO_AnyAttr : AnyAttrOf<[AnyAttr]>;
-def VHLO_AnyRegion : Region<CPred<"true">, "any region">;
```
burmako, line 92: I would we imagine we want to keep these three defs in this file, given that they are not TypeDefs or related to TypeDefs?

burmako (changes requested):

stablehlo/dialect/VhloBase.td
burmako: This seems to work out well, except that we need to have `VhloEnums.td` include `VhloAttrs.td` (and not the other way around, which is how things currently are in the code.).

burmako: If we decide to go this route the `include "mlir/IR/AttrTypeBase.td"` can be moved from `VhloBase.td` to `VhloTypes.td`.

stablehlo/dialect/VhloBase.td
GleasonK: I'll leave them in base for now. I think it would be a good sanity check to define:

```
def VHLO_AnyType : AnyTypeOf<[VHLO_TokenV1Type, VHLO_WrappedType, VHLO_IntegerV1Type, ...]>;
def VHLO_AnyAttr : AnyTypeOf<[VHLO_ChannelHandle, ....]>;
```

That said, this adds overhead to the process of adding a type/attr and may be better as a one-time experiment instead of something we submit.

GleasonK: Moved back to VhloBase.td

stablehlo/dialect/VhloBase.td
GleasonK: > we need to have VhloEnums.td include VhloAttrs.td

Could you clarify this? Currently Attrs includes Enums, are you suggesting the proper way is to reverse that dependency?

> Should VHLO_VersionedTypeInterface move to VhloTypes.td

Not necessarily. We could organize it that way, but interfaces have no dependencies to (op|attr|type)-defs, so the current organization is "base declares all interfaces, and the more specific files implement the interfaces for definitions."

I'm impartial, let me know what you think.

> If we decide to go this route the include "mlir/IR/AttrTypeBase.td" can be moved from VhloBase.td to VhloTypes.td.

Done.

burmako: Discussed this offline and decided to move interfaces.

burmako (changes requested):

burmako (approved):

PR #814, opened 2022-12-21
Add shell script and GitHub Action to run markdownlint

scottamain:
As per #792, here's a script to run `markdownlint`. The bash script includes usage instructions. You just need to pass it a file, directory, or glob pattern for the MD files you want to lint. It uses a Docker image because I fear the npm package version dependency hell, but the docker image is versioned.

~~The GitHub Action is inert because it currently has no trigger event.~~ [Removed the workflow file.] I just want to get the script reviewed and merged for now, so we can all experiment with the lint results. At this point, I believe the only file that does not pass the lint test (using the provided lint config, `markdownlint.yaml`) is `docs/spec.md`.


.github/workflows/lint_markdown.yml
```diff
@@ -0,0 +1,19 @@
+# Copyright 2022 The StableHLO Authors.
+
+name: Markdown Lint
```
GleasonK, line 3: Can we remove this from the PR? We currently have GH set up to require that all checks run in order to merge, I'm worried this not running will prevent that.

GleasonK (changes requested):
The bash script looks good, config looks good. Let's remove the actions YML for now until we're able to enable the CI.

Agree that it would be good to submit before CI for local testing and design iteration. I'll start using it immediately and provide feedback / enhancements.

.github/workflows/lint_markdown.yml
scottamain: Sounds good. Done.

scottamain:
FYI, the reason for `reference-links-images: false` is this bug: https://github.com/DavidAnson/markdownlint/issues/686

It's not a critical rule we need, so I think it's okay to keep disabled unless there's a way we can re-write this spec.md text as a workaround (in the "Constraints" list for stablhlo.reduce).

GleasonK (approved):

PR #816, opened 2022-12-21
Add interpreter testing guidelines

sdasgup3:


bhack:
Probably you are not interested as it is a fronted only issue. But check if you could find something interesting in https://discuss.tensorflow.org/t/best-place-to-submit-rfc-for-tensorflow-core-developers/13564

docs/reference.md
```diff
@@ -131,3 +131,57 @@ dedicated test-suite, consisting of several tests exercising various runtime
...
+While implementing an op, if we need to introduce code in the corresponding
+`eval` function to handle types differently, then it is imperative to have tests
+to cover for all those types. But in cases where a set of types are handled
+uniformly, like for `add` op, all the supported integer types are handled alike
+using `llvm::APInt` APIs, then we can skip adding tests for all those types and,
+instead, add a single test with a type, with maximum bit-width (an arbitrary
+choice), from the set.
+
+For quantization ops, like `uniform_dequantize`, `uniform_quantize`, the
+decision is TBD.
```
burmako, line 150: We don't yet have these ops specced, so I don't think we need this paragraph.

docs/reference.md
```diff
@@ -131,3 +131,57 @@ dedicated test-suite, consisting of several tests exercising various runtime
...
+
+We can use the following thumb rule to decide it:
+
+While implementing an op, if we need to introduce code in the corresponding
+`eval` function to handle types differently, then it is imperative to have tests
+to cover for all those types. But in cases where a set of types are handled
+uniformly, like for `add` op, all the supported integer types are handled alike
+using `llvm::APInt` APIs, then we can skip adding tests for all those types and,
+instead, add a single test with a type, with maximum bit-width (an arbitrary
+choice), from the set.
```
burmako, line 147: I'm not sure I exactly follow how many tests we need for `add`? Just one? One per type category (i.e. one for ints, one for floats, etc)? By the way, I really like like the idea of not writing tests for llvm::APInt and llvm::APFloat - but let's document this in a bit more detail.

burmako, line 147: Also, what happens to existing tests, e.g. https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/interpret_add.mlir?

docs/reference.md
```diff
@@ -131,3 +131,57 @@ dedicated test-suite, consisting of several tests exercising various runtime
...
+The interpreter infrastructure is mostly straightforward and can be added it in
+our trust base. The only non-trivial part is how various types are packed into
+and unpacked from the underlying interpreter storage. As discussed in (G1), we
+will be testing only those types of an op which are handled differently. With
+that is possible that the packing/un-packing code, corresponding to different
+variants of integer/floating point types, might not get fully covered during
+testing.
+
+To ensure that we can chose an op, like `add`, which supports all the StableHLO
+element types and write exhaustive tests. Fortunately, we already have it.
```
burmako, line 163: How about instead we have a dedicated file for this, separate from `interpreter_add.mlir`? E.g. `interpreter_types.mlir` or something?

docs/reference.md
```diff
@@ -131,3 +131,57 @@ dedicated test-suite, consisting of several tests exercising various runtime
...
+
+**(G3) If the implementation of an op depends other ops, should be write tests for the later ?**
+
+No. For example, the implementation of `batch_norm_grad` can be based on
+`divide`, `subtract`, `multiply` and others, we should avoid testing the later
```
burmako, line 168: Typo: "latter".

docs/reference.md
```diff
@@ -131,3 +131,57 @@ dedicated test-suite, consisting of several tests exercising various runtime
 behaviors, for each StableHLO Op. The tests can be found
 [here](https://github.com/openxla/stablehlo/tree/main/stablehlo/tests/) (e.g.
 interpret\_\*.mlir).
+
+### Testing guidelines
+
+**(G1) Do we need to test for all the supported types for every op ?**
```
burmako, line 137: Typo: we don't need whitespace before punctuation. (Here and below).

docs/reference.md
```diff
@@ -131,3 +131,57 @@ dedicated test-suite, consisting of several tests exercising various runtime
...
+
+No. For example, the implementation of `batch_norm_grad` can be based on
+`divide`, `subtract`, `multiply` and others, we should avoid testing the later
+ops while testing the former.
+
+**(G4) Should we write tests to exercise the implementation-defined / undefined behaviors ?**
+
+We should not write tests which exercises the implementation defined or
+undefined behaviors of the op. Tests exercising implementation defined behaviors
+demonstrates a local behavior of the interpreter which cannot be generalized.
```
burmako, line 175: Consider replacing "cannot" with "should not".

burmako (changes requested):

hrbigelow:
> Probably you are not interested as it is a fronted only issue. But check if you could find something interesting in https://discuss.tensorflow.org/t/best-place-to-submit-rfc-for-tensorflow-core-developers/13564

If the proposal in the above thread is of interest for StableHLO, this is a good place to start:   [opschema intro](https://github.com/hrbigelow/opschema/blob/master/introduction.md)

docs/reference.md
sdasgup3: This is my plan:

As suggested in https://github.com/openxla/stablehlo/pull/816#discussion_r1056874546, we will have a separate file `interpret_types.mlir` for testing how types are handled. We can use any particular op to write those tests. Let's say `add` op.

For the existing  interpret_<op>.mlir files, which  violates the guidelines mentioned in this OR, we need to trim them accordingly. As as example, https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/interpret_add.mlir will have fewer tests. Same with other , and other similar files. 

I wish to include the new interpret_types.mlir plus  the trimmed versions of existing test files all in this PR only once we LGTM on the rules.

docs/reference.md
sdasgup3: Please see my reply above.

docs/reference.md
```diff
@@ -69,7 +69,7 @@ we encapsulate details about how different element types are handled in
 
 We can use the interpreter mechanism to fold operations with constant operand
 values. The following code snippet demonstrates an idea of the implementation
-for folding `stablehlo::AddOp` with floating-point typed operands:
+for folding `stablehlo::AddOp` with Floating-point typed operands:
```
burmako, line 72: Canonical spelling in the spec doesn't capitalize "Integer", "Floating-point" etc. unless they are at the beginning of a sentence. Let's follow the same convention here and below?

docs/reference.md
```diff
@@ -131,3 +131,66 @@ dedicated test-suite, consisting of several tests exercising various runtime
...
+**(G1) Do we need to test for all the supported types for every op?**
+
+We can use a combination of following rules to decide it:
+
+  1. While implementing an op, if there exists code in the corresponding `eval`
+     function to handle a particular type, then it is imperative to have test(s)
+     to cover for that type. As an example, for `add` op, there is exclusive
+     code to handle Integer, Boolean, Floating-point, and Complex types, and hence we
+     need one test for each type.
+  2. If a set of types are handled uniformly in the corresponding
```
burmako, line 146: Typo: "is handled".

docs/reference.md
```diff
@@ -131,3 +131,66 @@ dedicated test-suite, consisting of several tests exercising various runtime
...
+
+**(G5) While writing tests for Floating-point type, to what precision the results need to be specified in llvm lit checks?**
+
+The current lit-based interpreter testing fails if the result is computed with a
+different precision than what is mentioned in the lit CHECK directives. As a
+quick-fix, we allow the lit checks to measure the accuracy up to an arbitrary
+places after the decimal point. But the solution is far from ideal. We plan to
+resolve it using [ticket](https://github.com/openxla/stablehlo/issues/268).
+
+
```
burmako, line 196: Typo: extraneous newline.

docs/reference.md
```diff
@@ -131,3 +131,66 @@ dedicated test-suite, consisting of several tests exercising various runtime
...
+**(G3) If the implementation of an op depends other ops, should be write tests for the latter?**
+
+No. For example, the implementation of `batch_norm_grad` can be based on
+`divide`, `subtract`, `multiply` and others, we should avoid testing the latter
+ops while testing the former.
+
+**(G4) Should we write tests to exercise the implementation-defined / undefined behaviors?**
+
+We should not write tests which exercises the implementation defined or
```
burmako, line 182: Typo: "exercise". (As well as "demonstrate" two lines below and "do not" three lines below).

docs/reference.md
```diff
@@ -131,3 +131,66 @@ dedicated test-suite, consisting of several tests exercising various runtime
...
+
+**(G1) Do we need to test for all the supported types for every op?**
+
+We can use a combination of following rules to decide it:
+
+  1. While implementing an op, if there exists code in the corresponding `eval`
+     function to handle a particular type, then it is imperative to have test(s)
+     to cover for that type. As an example, for `add` op, there is exclusive
+     code to handle Integer, Boolean, Floating-point, and Complex types, and hence we
+     need one test for each type.
```
burmako, line 145: "for each category of types"?

docs/reference.md
burmako: Sounds like a plan. Instead of updating the tests right in this PR, consider opening a follow-up ticket for that, so that we can expedite merging this PR.

burmako (changes requested):

burmako:
@bhack @hrbigelow Thank you for sharing! We'll take a look after the winter holidays and follow up in early January.

docs/reference.md
sdasgup3: Here is the [ticket](https://github.com/openxla/stablehlo/issues/918).

sdasgup3:
> > Probably you are not interested as it is a fronted only issue. But check if you could find something interesting in https://discuss.tensorflow.org/t/best-place-to-submit-rfc-for-tensorflow-core-developers/13564
> 
> If the proposal in the above thread is of interest for StableHLO, this is a good place to start: [opschema intro](https://github.com/hrbigelow/opschema/blob/master/introduction.md)

Thanks for sharing this super interesting work! 
From my past experience in formally defining op semantics, I can certainly appreciate the benefit of frameworks (like opschema) which can provide the various utilities (including the ones mentioned in https://github.com/hrbigelow/opschema#introduction), almost for free, once we define the opset in a structured (or, for some fraemworks, even formal) setup.

We do have a future goal to auto generate spec to reduce the TCB and tools like the above will definitely help such exploration.

docs/reference.md
```diff
@@ -131,3 +131,64 @@ dedicated test-suite, consisting of several tests exercising various runtime
...
+     `llvm::APInt` APIs, and hence we can skip adding tests for each of those
+     variants, and instead, add a single representative test. To avoid ambiguity
+     in selecting the representative, we should use the following guidelines:
+
+       * If all the types, handled uniformaly, have the same primitive type
+         (i.e., if all are integer, or floating-point, or complex types), then
+         choose the one with maximum bit-width.
+       * If all the types, handled uniformaly, have a mix of primitive types,
+         then choose the one with the following primitive type, in decreasing
+         order of preference: integer, floating-point, boolean, complex.
```
subhankarshah, line 159: 1. What is the rationale for this order?

docs/reference.md
```diff
@@ -131,3 +131,64 @@ dedicated test-suite, consisting of several tests exercising various runtime
...
+
+**(G2) How about adding tests dedicated for testing the interpreter infrastructure?**
+
+The interpreter infrastructure is mostly straightforward and can be added it in
```
subhankarshah, line 163: added to

docs/reference.md
```diff
@@ -131,3 +131,64 @@ dedicated test-suite, consisting of several tests exercising various runtime
...
+our trust base. The only non-trivial part is how various types are packed into
+and unpacked from the underlying interpreter storage. As discussed in (G1), we
+will be testing only those types of an op which are handled differently. With
+that is possible that the packing/un-packing code, corresponding to different
```
subhankarshah, line 167: , it is possible

docs/reference.md
```diff
@@ -131,3 +131,64 @@ dedicated test-suite, consisting of several tests exercising various runtime
...
+variants of integer/floating-point types, might not get fully covered during
+testing.
+
+To ensure that we can chose an op, like `add`, which supports all the StableHLO
```
subhankarshah, line 171: can we add this in previous paragraph.

docs/reference.md
```diff
@@ -131,3 +131,64 @@ dedicated test-suite, consisting of several tests exercising various runtime
...
+element types and write exhaustive tests. Fortunately, we already have it.
+
+**(G3) If the implementation of an op depends other ops, should be write tests for the latter?**
+
+No. For example, the implementation of `batch_norm_grad` can be based on
+`divide`, `subtract`, `multiply` and others, we should avoid testing the latter
+ops while testing the former.
```
subhankarshah, line 178: This seems right, I have a question - We would still be able to interpret `batch_norm_grad` right? As in it would be lowered to other ops and those ops would be interpreted and give a result?

subhankarshah (changes requested):
Added a few comments.

docs/reference.md
sdasgup3: Thanks for bringing this up.

* If all the types, handled uniformaly, have a mix of primitive types,
         then choose..
An example of the ops which will fall into this bucket are ones in category "Data movement" like scatter/gather/sort etc.. 

One option I was considering  was to have the order same as https://github.com/openxla/stablehlo/blob/main/docs/spec.md#types.
which proposes to prefer boolean types over other types.
 
My current proposal is to order the types in a way such the the tests will add value to the understanding of the op.  

For example, writing tests, for say sort op, with boolean values vs integer values. IMO, the latter is better when it comes to understanding what the op is doing. This might be a trivial example, but the point is, in order to understand the semantics, often times we track the data movement to get an empirical understanding of the functioning of the op. For example, with gather/scatter, movements of 0/1 are much difficult to track that unique integer  values. 

Overall, the proposal is discouraging writing tests in boolean and complex over integers and floating-point values. 

The following (remaining) choices are  based on ease of representation. But I agree that it could be debatable.  
  - Between integer and floating-point, integer is preferred 
  - Between boolean and complex, boolean is preffered   
  
  Please let me know if this makes sense.

docs/reference.md
sdasgup3: You that is right! Given that the interpreter needs to be a readable implementation of the spec, it is expected that the implementation of a batch* op follow the from the evaluation of  the decomposed ops.

docs/reference.md
sdasgup3: sorry that I did not follow this. Do you want this to be added under G1?

subhankarshah: yes, immediately after `... covered during testing. In order to ensure this, we can choose an op like ...`
typo in spelling of choose

sdasgup3: done. Thanks!

hrbigelow:
> > > Probably you are not interested as it is a fronted only issue. But check if you could find something interesting in https://discuss.tensorflow.org/t/best-place-to-submit-rfc-for-tensorflow-core-developers/13564
> > 
> > 
> > If the proposal in the above thread is of interest for StableHLO, this is a good place to start: [opschema intro](https://github.com/hrbigelow/opschema/blob/master/introduction.md)
> 
> Thanks for sharing this super interesting work! From my past experience in formally defining op semantics, I do certainly appreciate the benefit of frameworks (like opschema) which can provide the various utilities (including the ones mentioned in https://github.com/hrbigelow/opschema#introduction), almost for free, once we define the opset in a structured (or, for some fraemworks, even formal) setup.
> 
> We do have a future goal to auto generate spec to reduce the TCB and tools like the above will definitely help such exploration.

Hi Sandeep,

Actually, opschema began as a previous project called [einsum-tuple](https://github.com/hrbigelow/einsum-tuple) which was a mini-language for fully defining such tensor operations in a quasi-einsum notation.  It is actually very similar to a lot of the informal pseudo-code showing index arithmetic in tensorflow docs, and I was inspired to try to design it after reading a lot of the MLIR docs and polyhedral analysis.

I thought it would be a good thing to have as a high-level 'spec' defining each op's computation.  After awhile I thought maybe it wouldn't be useful to end-users, only developers.  But, would be great to hear your feedback on it and any pointers to similar work you have done?

Some examples of the specs are here:  [gather](https://github.com/hrbigelow/einsum-tuple#gather), [meshgrid](https://github.com/hrbigelow/einsum-tuple#meshgrid), [space_to_depth](https://github.com/hrbigelow/einsum-tuple#space-to-depth).

Best,

Henry

docs/reference.md
subhankarshah: Thanks for the detailed answer, between integer and floating-point, if floating point results do not provide any useful information from a precision standpoint then integer saves us a lot of hassle and gets the point across in terms of explaining the behaviour or logic of the op.

docs/reference.md
```diff
@@ -131,3 +131,62 @@ dedicated test-suite, consisting of several tests exercising various runtime
 behaviors, for each StableHLO Op. The tests can be found
 [here](https://github.com/openxla/stablehlo/tree/main/stablehlo/tests/) (e.g.
 interpret\_\*.mlir).
+
+### Testing guidelines
+
+**(G1) Do we need to test for all the supported types for every op?**
+
+We can use a combination of following rules to decide it:
```
burmako, line 139: Indentation for this list doesn't seem to follow the recently approved guidelines. PTAL.

burmako (approved):

docs/reference.md
sdasgup3: Sure.

subhankarshah (approved):
LGTM

sdasgup3:
@hrbigelow 
Dear Henry
First of all sorry for the late response. Thanks for sharing your work. It's enlightening. 

<<  I thought it would be a good thing to have as a high-level 'spec' defining each op's computation. After awhile I thought maybe it wouldn't be useful to end-users, only developers. 

I agree with your point. The acceptance to end-users is also proportional to the "incentives"  provided by the high-level specification framework.  For example, users of the `validator` will be interested to write the operation in  `einsum-tuple`.

>> any pointers to similar work you have done?

Sorry! Not exactly related to the problem you people are solving, but with a similar theme (writing operational spec of a language  in a high-level specification framework (e.g. a proof-assistant)

ghpvnist (approved):

hrbigelow:
> @hrbigelow Dear Henry First of all sorry for the late response. Thanks for sharing your work. It's enlightening.
> 
> << I thought it would be a good thing to have as a high-level 'spec' defining each op's computation. After awhile I thought maybe it wouldn't be useful to end-users, only developers.
> 
> I agree with your point. The acceptance to end-users is also proportional to the "incentives" provided by the high-level specification framework. For example, users of the `validator` will be interested to write the operation in `einsum-tuple`.
> 
> > > any pointers to similar work you have done?
> 
> Sorry! Not exactly related to the problem you people are solving, but with a similar theme (writing operational spec of a language in a high-level specification framework (e.g. a proof-assistant)

Hi Sandeep,

No problem - really appreciate the response and thanks for your insights.  Yes, a validator was one of the things I had in mind.  Was a lot of fun to work on and educational for me.  I'll check out your operational spec work.  Cheers!

Issue #600, opened 2022-11-24
Consider adding preferred_element_type to DotGeneralOp/ConvolutionOp


burmako:
One of the requirements for DotGeneralOp/ConvolutionOp is to support different element types for operands and results. E.g. for `i8` inputs we might very well want to support `i32` outputs.

At the moment, this is implemented in the StableHLO dialect via load-bearing result types, which makes type inference for these ops infeasible. That's kind of fine, but an alternative design would involve specifying the element type of the result via an attribute, which doesn't look like a big imposition on producers and would unlock type inference.

PR #817, opened 2022-12-21
Disable shape functions for ConvolutionOp, DotOp and DotGeneralOp

burmako:
In their current formulation, these ops cannot have shape functions because their lhs/rhs element types can be different from their result element type.

Perhaps in the future we will add the `preferred_element_type` attribute to these ops (see #600), and then shape functions will be feasible to write again, but today we have to disable them.

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -2300,13 +2300,15 @@ LogicalResult verifyDotOp(Optional<Location> location, Value lhs, Value rhs,
   auto lhsType = lhs.getType().dyn_cast<RankedTensorType>();
   auto rhsType = rhs.getType().dyn_cast<RankedTensorType>();
   auto resultType = result.getType().dyn_cast<RankedTensorType>();
-  if (!lhsType || !rhsType || !resultType) return success();
+  if (!lhsType || !rhsType) return success();
 
   SmallVector<int64_t> inferredShape;
   if (failed(inferDotShape(lhsType, rhsType, inferredShape)))
-    return emitOptionalError(location, "failed to infer shape");
+    return emitOptionalError(location, "failed to infer shape for lhs ",
+                             lhsType, " and rhs ", rhsType);
 
-  if (failed(verifyCompatibleShape(inferredShape, resultType.getShape())))
+  if (resultType &&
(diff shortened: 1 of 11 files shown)


stablehlo/dialect/TypeInference.cpp
```diff
@@ -2069,6 +2172,280 @@ LogicalResult verifyCollectivePermuteOp(
...
+                          Optional<ArrayAttr> /*precisionConfig*/,
+                          Value result) {
+  auto lhsType = lhs.getType().dyn_cast<RankedTensorType>();
+  auto rhsType = rhs.getType().dyn_cast<RankedTensorType>();
+  auto resultType = result.getType().dyn_cast<RankedTensorType>();
+  if (!lhsType || !rhsType || !resultType) return success();
+
+  SmallVector<int64_t> inferredShape;
+  if (failed(inferDotShape(lhsType, rhsType, inferredShape)))
+    return emitOptionalError(location, "failed to infer shape");
```
zhouxin913, line 2307: Looks like there is no tests for this error? A more comprehensive error message would be helpful like: 
"failed to infer shape with lhsType of shape [...] and rhsType of shape [...]".

stablehlo/dialect/TypeInference.cpp
```diff
@@ -2069,6 +2172,280 @@ LogicalResult verifyCollectivePermuteOp(
...
+  return success();
+}
+
+LogicalResult verifyDotOp(Optional<Location> location, Value lhs, Value rhs,
+                          Optional<ArrayAttr> /*precisionConfig*/,
+                          Value result) {
+  auto lhsType = lhs.getType().dyn_cast<RankedTensorType>();
+  auto rhsType = rhs.getType().dyn_cast<RankedTensorType>();
+  auto resultType = result.getType().dyn_cast<RankedTensorType>();
+  if (!lhsType || !rhsType || !resultType) return success();
```
zhouxin913, line 2303: Your `verifyConvolutionop()` inspire me that this `if (!resultType) return success() ` should show up after all the verifications are done for the inputs/attrs, right? I mean to move it below the `inferDotShape()` part.

zhouxin913 (approved):
Approved with 2 small comments: if rushing time, please consider add TODO for it if it make sense. Thanks!

PR #819, opened 2022-12-21
Handle bounds among branches

zhouxin913:
This PR proposes an approach to infer the sizes and bounds for ops with branches.
Back to months ago, https://github.com/openxla/stablehlo/pull/229 tried to resolve this exactly same question in a restrict approach, inferring the most restrict shape from the result type among all branches: but that PR does not submit, leaving an open question of this topic. 

Inspired by how HLO handle this problem https://github.com/tensorflow/tensorflow/blob/v2.11.0/tensorflow/compiler/xla/service/shape_inference.cc#L2988
```
// For each subshape, If any of the branch is dynamic, we say result is
  // dynamic:
  //   true_branch  (s32[<=4])
  //   false_branch (s32[4])
  // Result is s32[<=4].
``` 
this PR proposes a most least specific inference approach with rules as below (assuming dim of lhs and dim of rhs are compatible): 
```
// Inference rules for conditional branches (lhs/rhs are commutative):
//       Dim of lhs     Dim of rhs      Infer
//  c0:  X              X               X
//  c1:  X              ?               ?
//  c2:  X              ?, B            ?, max(X, B)
//  c3:  ?              ?               ?
//  c4:  ?              ?, B            ?
//  c5:  ?, B           ?, C            ?, max(B, C)
```
Note that C2 is a superset of the XLA code above.

The reasoning behind:
1. For binary op `stablehlo.add`, the result type must match with both `lhs` and `rhs`, thus the inferred type is of cause generated from the most restricted way.
2. However, for ops with branches, it is unknown whether every branch will be actually executed in the runtime, thus inferring a restrict return type would raise false positive errors, even tho it may help reduce false negative errors. For example a `case` op with branch_0 return  `tensor<3x?x>` and branch_1 return `tensor<?x4x>`: inferring `tensor<3x4x>` is too strong if branch_0 is always executed in runtime with a valid output with shape `<3x100x>` and consumed by the following ops who don't care branch_1 at all and assume the input are always like `<3x?x>`.

Add reviewer: @smit-hinsu


stablehlo/dialect/Base.cpp
```diff
@@ -261,67 +261,96 @@ FailureOr<std::pair<int64_t, int64_t>> inferMergedDimAndBound(
...
-// TODO(zhouxin) Refactor to better handle errors and return single type
-LogicalResult inferMostSpecificType(
+// Inference rules for conditional branches (lhs/rhs are commutative):
+//       Dim of lhs     Dim of rhs      Infer
+//  c0:  X              X               X
+//  c1:  X              ?               ?
+//  c2:  X              ?, B(>=X)       ?, B(>=X)
+//  c3:  X              ?, B(<X)        ?, B(<X)
+//  c4:  ?              ?               ?
+//  c5:  ?              ?, B            ?, B
```
smit-hinsu, line 271: should C5 be,

? <=B to ?

If we end up taking the first branch, then the size could be larger.

zhouxin913: Agreed.

stablehlo/dialect/Base.cpp
```diff
@@ -261,67 +261,96 @@ FailureOr<std::pair<int64_t, int64_t>> inferMergedDimAndBound(
...
 }
 
-// TODO(zhouxin) Refactor to better handle errors and return single type
-LogicalResult inferMostSpecificType(
+// Inference rules for conditional branches (lhs/rhs are commutative):
+//       Dim of lhs     Dim of rhs      Infer
+//  c0:  X              X               X
+//  c1:  X              ?               ?
+//  c2:  X              ?, B(>=X)       ?, B(>=X)
+//  c3:  X              ?, B(<X)        ?, B(<X)
```
burmako, line 269: From what I can see, [the HLO code mentioned in PR's description](https://github.com/tensorflow/tensorflow/blob/d5b57ca93e506df258271ea00fc29cf98383a374/tensorflow/compiler/xla/service/shape_inference.cc#L2999) only allows `B = X`, but not any other values for B. What are the pros and cons of doing the same thing as HLO vs being more general as proposed in this pull request?

burmako (changes requested):

stablehlo/dialect/Base.cpp
zhouxin913: I give a detailed proposal in the PR description, which answer this question as well, PTAL.

smit-hinsu (commented):
Thanks!

stablehlo/dialect/Base.cpp
```diff
@@ -261,67 +267,105 @@ FailureOr<std::pair<int64_t, int64_t>> inferMergedDimAndBound(
   return std::make_pair(size, bound);
 }
 
-// TODO(zhouxin) Refactor to better handle errors and return single type
-LogicalResult inferMostSpecificType(
+// Inference rules for conditional branches (lhs/rhs are commutative):
+//       Dim of lhs     Dim of rhs      Infer
+//  c0:  X              X               X
+//  c1:  X              ?               ?
+//  c2:  X              ?, B(<X)        Will error out by compatible checks
```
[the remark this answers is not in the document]
zhouxin913: Makes sense.

stablehlo/dialect/Base.cpp
```diff
@@ -261,67 +267,105 @@ FailureOr<std::pair<int64_t, int64_t>> inferMergedDimAndBound(
...
+    if (isLeftStaticBound && isRightStaticBound)
+      inferredbound = std::max(leftBound, rightBound);
+  }
+  return std::make_pair(inferredSize, inferredbound);
+}
+
+FailureOr<Type> inferMostSpecificType(
     Optional<Location> location, TypeRange inputTypes,
-    SmallVectorImpl<Type>& inferredReturnTypes) {
+    std::function<FailureOr<std::pair<int64_t, int64_t>>(
```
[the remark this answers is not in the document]
zhouxin913: Resolved and refactored.

stablehlo/dialect/Base.cpp
```diff
@@ -261,67 +267,116 @@ FailureOr<std::pair<int64_t, int64_t>> inferMergedDimAndBound(
...
+
+  if (isLeftStaticDim || isRightStaticDim) {
+    if (isLeftStaticDim && isRightStaticDim) {
+      if (leftSize != rightSize)
+        return emitOptionalError(location, "Mismatched dimension sizes ",
+                                 leftSize, " and ", rightSize, " in dimension ",
+                                 dim);
+      inferredSize = leftSize;
+    } else if (isLeftStaticBound || isRightStaticBound) {
+      int64_t size = isLeftStaticDim ? leftSize : rightSize;
```
smit-hinsu, line 296: nit: we should encode the fact here that if isLeftStaticDim then isLeftStaticBound is false so we could have just one condition. Either with tuple or simple if/else.

isLeftStaticDim ? (leftSize, rightBound) : (rightSize, leftBound)

stablehlo/dialect/Base.h
```diff
@@ -73,16 +73,27 @@ std::pair<int64_t, int64_t> inferConcatenatedDimAndBound(int64_t leftSize,
...
+
 // Infer single most specific return type from inputTypes with support for
 // bounds. (Size, bound) of each dimension of the return type will be merged
-// from corresponding dimensions of every inputType by merging them.
-LogicalResult inferMostSpecificType(Optional<Location> location,
-                                    TypeRange inputTypes,
-                                    SmallVectorImpl<Type> &inferredReturnTypes);
+// from corresponding dimensions of every inputType by extracting the most
+// specific one. Return unranked tensor if all inputs are unranked.
+FailureOr<Type> inferMostSpecificType(Optional<Location> location,
```
smit-hinsu, line 95: we could return ShapedType from here to avoid cast at the callsite?

smit-hinsu (approved):
Thanks!

ghpvnist:
SliceOp would also benefit from introducing the `inferLeastSpecificDimAndBound` as this op also selects one of two tensors based on the `pred` input tensor.

stablehlo/dialect/Base.cpp
zhouxin913: That's neat! Changed to
inferredBound = isLeftStaticDim ? std::max(leftSize, rightBound) : std::max(rightSize, leftBound);

stablehlo/dialect/Base.h
zhouxin913: Done.

stablehlo/tests/infer_stablehlo.mlir
```diff
@@ -983,6 +981,64 @@ func.func @concat_bounds_unranked_c1(
...
+    %branch_2_operand : tensor<?xf32, #stablehlo.type_extensions<bounds = [3]>>) -> tensor<*xindex> {
+  %0 = "stablehlo.case"(%index) ({
+      "stablehlo.return"(%branch_0_operand) : (tensor<2xf32, #stablehlo.type_extensions<bounds = [?]>>) -> ()
+  }, {
+      "stablehlo.return"(%branch_0_operand) : (tensor<2xf32, #stablehlo.type_extensions<bounds = [?]>>) -> ()
+  }, {
+      "stablehlo.return"(%branch_2_operand) : (tensor<?xf32, #stablehlo.type_extensions<bounds = [3]>>) -> ()
+  }, {
+      "stablehlo.return"(%branch_2_operand) : (tensor<?xf32, #stablehlo.type_extensions<bounds = [3]>>) -> ()
+  }) : (tensor<i32>) -> tensor<*xf32>
```
burmako, line 1034: Let's reduce the number of branches to simplify the test? It doesn't seem like duplicating the return types helps test additional logic?

stablehlo/dialect/Base.h
```diff
@@ -237,9 +248,10 @@ class CompatibleOperandsAndResultType
           location,
           "Expected non-empty operands for [CompatibleOperandsAndResultType]");
 
-    if (failed(inferMostSpecificType(location, operands.getTypes(),
-                                     inferredReturnTypes)))
-      return failure();
+    auto inferredTypeOrErr =
+        inferMostSpecificType(location, operands.getTypes());
+    if (failed(inferredTypeOrErr)) return failure();
+    inferredReturnTypes.emplace_back((*inferredTypeOrErr).cast<ShapedType>());
```
burmako, line 254: I don't think the cast is needed here, because `*inferredTypeOrErr` is already `ShapedType`.

stablehlo/dialect/Base.h
```diff
@@ -73,16 +73,27 @@ std::pair<int64_t, int64_t> inferConcatenatedDimAndBound(int64_t leftSize,
...
 // Infer single most specific return type from inputTypes with support for
 // bounds. (Size, bound) of each dimension of the return type will be merged
-// from corresponding dimensions of every inputType by merging them.
-LogicalResult inferMostSpecificType(Optional<Location> location,
-                                    TypeRange inputTypes,
-                                    SmallVectorImpl<Type> &inferredReturnTypes);
+// from corresponding dimensions of every inputType by extracting the most
+// specific one. Return unranked tensor if all inputs are unranked.
+FailureOr<ShapedType> inferMostSpecificType(Optional<Location> location,
+                                            TypeRange inputTypes);
```
burmako, line 96: Any particular reason to switch from LogicalResult to FailureOr? This makes the code in this PR more verbose overall, and the upside is not entirely clear (at least, not in this PR).

stablehlo/dialect/Base.cpp
```diff
@@ -261,67 +267,114 @@ FailureOr<std::pair<int64_t, int64_t>> inferMergedDimAndBound(
...
+    inferredBound = std::max(leftBound, rightBound);
   }
+  return std::make_pair(inferredSize, inferredBound);
+}
 
+FailureOr<TensorType> inferTypeWithCustomFunc(
+    Optional<Location> location, SmallVector<RankedTensorType> rankedTypes,
+    std::function<FailureOr<std::pair<int64_t, int64_t>>(
+        Optional<Location>, int64_t, int64_t, int64_t, int64_t, int64_t)>
+        inferDimAndBoundFunc) {
```
burmako, line 309: How about simplifying this to `inferDimAndBound`?

stablehlo/dialect/Base.cpp
```diff
@@ -261,67 +267,114 @@ FailureOr<std::pair<int64_t, int64_t>> inferMergedDimAndBound(
...
+  for (unsigned i = 0; i < rankedTypes.size(); ++i) {
+    RankedTensorType rankedType = rankedTypes[i];
     ArrayRef<int64_t> bounds = encodingToBounds(rankedType.getEncoding());
     if (!bounds.empty()) anyInputHaveBounds = true;
 
+    if (i == 0) {
+      inferredSizes = to_vector(rankedType.getShape());
+      if (!bounds.empty()) inferredBounds = to_vector(bounds);
+      continue;
+    }
```
burmako, line 324: How about moving the `i == 0` part of the loop above where `inferredSizes` and `inferredBounds` are declared and then start the loop with `i = 1`? This way the variables will be consistently initialized where they are declared, and the loop won't have a special case.

stablehlo/dialect/Base.cpp
```diff
@@ -261,67 +267,114 @@ FailureOr<std::pair<int64_t, int64_t>> inferMergedDimAndBound(
...
+//  c5:  ?, B           ?, C            ?, max(B, C)
+FailureOr<std::pair<int64_t, int64_t>> inferLeastSpecificDimAndBound(
+    Optional<Location> location, int64_t dim, int64_t leftSize,
+    int64_t rightSize, int64_t leftBound, int64_t rightBound) {
+  bool isLeftStaticDim = !isDynamicDimSize(leftSize);
+  bool isRightStaticDim = !isDynamicDimSize(rightSize);
+  bool isLeftStaticBound = !isDynamicDimSize(leftBound);
+  bool isRightStaticBound = !isDynamicDimSize(rightBound);
+  int64_t inferredSize = ShapedType::kDynamic;
+  int64_t inferredBound = ShapedType::kDynamic;
```
burmako, line 286: These two variables are called `size` and `bound` above in `inferMostSpecificDimAndBound`. Let's consolidate the naming to be the same across the two functions. (I don't have a strong preference for what it should be).

burmako (changes requested):

stablehlo/dialect/Base.cpp
```diff
@@ -150,6 +150,12 @@ Type createRealType(TensorType type) {
   return hlo::getSameShapeTensorType(type, elementTy);
 }
 
+// TODO(zhouxin) move this section to TypeInference.h/cpp
```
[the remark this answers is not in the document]
zhouxin913: Sure, opened https://github.com/openxla/stablehlo/issues/955.

stablehlo/dialect/Base.h
zhouxin913: It was intended to be moved. Fixed.

stablehlo/tests/infer_stablehlo.mlir
zhouxin913: This in intended to test that CaseOp supports more branches than IfOp (the comment above the test says this as well. I agree to reduce this test from 4 branches to only 3 branches.

stablehlo/dialect/Base.h
zhouxin913: We discussed a while ago to refine this function to return Type instead of passing in `SmallVectorImpl<Type> &inferredReturnTypes`, because there will only 1 inferred type returned.
Besides, this util function should handle error inside it, thus `FailureOr<ShapedType>` is the final decision. Please let me know if you have better thoughts for this.

stablehlo/dialect/Base.cpp
zhouxin913: I prefer to simplify it to inferDimAndBoundFn, I have not find a code style for this in LLVM, but I think it provides a clear clue that this is a fn to be used, like https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Transforms/DialectConversion.h#L237

stablehlo/dialect/Base.cpp
zhouxin913: Makes sense! Updated.

stablehlo/dialect/Base.cpp
zhouxin913: I prefer `inferredSize ` than `size`, even tho it's longer, because `size` is used as a local var to hold some intermediate value. Updated the other functions.

stablehlo/dialect/Base.h
burmako: I don't have a strong opinion about this one. Let's proceed with what's implemented in the PR.

burmako (approved):

PR #820, opened 2022-12-21
Fix MD lint errors for spec.md

scottamain:
Fixes to the markdown include:
+ Use a consistent 2-space indentation for unordered lists.
+ Set line lengths to 80 characters, except for cases in which doing so would adversely affect readability for the source file such as for long formulas (in which case, the rule is temporarily disabled with the `markdownlint-disable` tag). (Unfortunately the line-length rule cannot be customized for code font—only for pre-formatted code blocks.)
+ Specify code language for all code blocks.

Modify the configuration such that:
+ Allow duplicate heading names as long as the headings are not siblings (separate sections can use the same sub-headings).
+ Allow images without alt-text, due to the complexity of images and relatively small number of readers (certainly something to revisit if a11y is an issue).
+ Prefer 2-space indentation for unordered lists (there is no matching config for ordered lists, so those should use no indentation for the top-level list items)


burmako:
Thank you for the continuous cleanup of markdown files! This is a big change. Assigned both Kevin and myself to review this. @GleasonK can you take a look at the changes to actions, and I'll take a look at the spec later today?

GleasonK:
Config changes look good. I'll take a look if there are any other  @burmako the yaml file in this CL is for config. Thoughts on moving this to the repo root and renaming to `.markdownlint.yml`? This seems more conventional for configs, also has some precedence in other repos:

- https://github.com/DavidAnson/markdownlint/blob/main/.markdownlint.json
- https://github.com/iree-org/iree/blob/main/.yamllint.yml

@scottamain do you have any thoughts on the rules we use? Does the default check enforce most of these? Any others worth considering in the future?
https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md

scottamain:
@GleasonK I'm indifferent on the config file location. At the root SGTM if that's what you like. Shall I move it there?

By default, markdownlint enforces all those rules. I included `default: true` at the top of the config just to be explicit, and then the config parameters that follow it are our overrides to the default behaviors (such as allow line-length to exceed 80 chars if in a table or code block). So I prefer to use all the default rules until we encounter a reason to do otherwise, which is how I arrived at this current configuration.

GleasonK:
Will leave final word on config location to Eugene's review. I'm supportive of root and prefix file with dot.

docs/spec.md
```diff
@@ -58,11 +58,11 @@ At the moment, StableHLO only supports dense tensors, so each tensor has
...
-  * `tensor<2x3xf32>` has 6 elements whose indices are
+* `tensor<2x3xf32>` has 6 elements whose indices are
     `{0, 0}`, `{0, 1}`, `{0, 2}`, `{1, 0}`, `{1, 1}` and `{1, 2}`.
-  * Tensors of rank zero, e.g `tensor<f32>`, have 1 element. Such tensors are
+* Tensors of rank zero, e.g `tensor<f32>`, have 1 element. Such tensors are
     allowed and are useful to model scalars.
-  * Tensors with dimensions of size zero, e.g. `tensor<2x0xf32>`, have
+* Tensors with dimensions of size zero, e.g. `tensor<2x0xf32>`, have
     0 elements. Such tensors are allowed and are useful in rare cases, e.g.
     to model empty slices.
```
burmako, line 67: Should indentation of lines 66 and 67 (as well as some lines above) be changed to align with the first line of the bullet? (Same for lines 605-619, 679-693, 736, 1032, 1115, 1203, 1234-1249, 1270, 1323, 1570-1583, 1611, 1677-1695, 2258-2260, 2270-2282, and more).

I understand that this is non-trivial work to align these indentations, so I can do it on my own too, just wondering about the canonical rules for indenting multiline bullets.

docs/spec.md
```diff
@@ -154,6 +154,7 @@ before their uses. Possible execution orders of the example program above are
 → `return`.
 
 More formally, a **StableHLO process** is a combination of:
+
 1) a StableHLO program, 2) operation statuses (not executed yet,
```
burmako, line 158: Actually this was not supposed to be a list, just a regular paragraph. How do we express that? I guess, "1\)"?

burmako (changes requested):

burmako:
"I'm supportive of root and prefix file with dot". Sounds good to me as well.

docs/spec.md
scottamain: Ah. Yes, full parentheses solves it. That is: "(1)"

docs/spec.md
scottamain: Dang, I didn't notice this. It was caused by the auto-fix for list items that were unnecessarily indented ([rule MD007](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md007---unordered-list-indentation)), but the autofix does not remove indentation from the wrapped lines. Perhaps I should revert that part of the change (fortunately it was a separate commit) and then we can manually respond to all the MD007 errors manually, so that we're sure to also un-indent the wrapped lines? The alternative is to actually *prefer* two-space indentation for all list items, which we can actually specify by setting [`start_indented: true`](https://github.com/DavidAnson/markdownlint/blob/main/schema/.markdownlint.yaml#L33-L40). Then MD007 shouldn't complain about the original formatting at all.

burmako: From your perspective, what do you prefer `start_indented: true` or `start_indented: false`? Our convention so far was the former, so we might as well keep it if you don't feel strongly about this / there are no significant downsides.

scottamain: Considering the convention has been to use two-space indents for lists, I think we should keep that just for simplicity because this is purely a style choice. So I've updated the lint config to prefer two-space indents (beware that this applies only to unordered/bulleted lists; there is no matching config for ordered lists), and I reverted the change that applied the indentation changes.

burmako (approved):

PR #824, opened 2022-12-23
Handle Dynamic Bounds in Shape Function of DynamicUpdateSliceOp

### LINKED ISSUE #799: Handle bounds in `DynamicUpdateSliceOp`shape function

sdasgup3:
The PR fixes #799 (please refer this for more details)

 Adding @smit-hinsu for review.

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -1724,6 +1724,7 @@ LogicalResult inferDynamicUpdateSliceOp(
   auto operandType = operand.getType().cast<ShapedType>();
   auto updateType = update.getType().cast<ShapedType>();
 
+  // (C3)
   if (updateType.hasRank() && operandType.hasRank() &&
       updateType.getRank() != operandType.getRank())
     return emitOptionalError(
@@ -1731,12 +1732,14 @@ LogicalResult inferDynamicUpdateSliceOp(
         "update rank does not match operand rank: ", updateType.getRank(),
         " vs ", operandType.getRank(), ".");
 
+  // (C4)
   if (operandType.hasRank() &&
       (int64_t)startIndices.size() != operandType.getRank())
     return emitOptionalError(
         location, "expects number of start_indices to match operand rank: ",
         startIndices.size(), " vs ", operandType.getRank(), ".");
 
+  // (C5)
   if (!startIndices.empty()) {
     auto firstIndexType = startIndices[0].getType().cast<ShapedType>();
     Type firstIndexElement = firstIndexType.getElementType();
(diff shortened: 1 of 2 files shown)


smit-hinsu (approved):
Thanks!

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1337,22 +1337,19 @@ LogicalResult inferDynamicUpdateSliceOp(
   auto operandType = operand.getType().cast<ShapedType>();
   auto updateType = update.getType().cast<ShapedType>();
 
-  // (C3)
```
ghpvnist, line 1340: Any reason for removing the constraints comment?

sdasgup3: Good catch!
I thought them to be stray comments. But realised that they are linked 1-1 to the spec constraints. Added them back.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1388,8 +1388,9 @@ LogicalResult inferDynamicUpdateSliceOp(
 
   // (C1)
   if (operandType.hasRank()) {
-    inferredReturnShapes.emplace_back(operandType.getShape(),
-                                      operandType.getElementType());
+    inferredReturnShapes.emplace_back(
+        operandType.getShape(), operandType.getElementType(),
+        operandType.cast<RankedTensorType>().getEncoding());
```
burmako, line 1393: What are our plans for updating the rest of the code? More specifically, what's going to happen to C6?

zhouxin913: I think this leads to a more general question: what do we do for an existing "X <= Y" check where X is a dim from one tensor, and Y is a dim from a different tensor, in the new bounds scenario? 
I would propose a util function below to replace:
```
assert(X <= Y);
```
with 
```
assert(LessEqualThanWithBounds(X, Y, A, B);
```
```
//  c4:  ?              any             true
//  c5:  any            ?               true
//  c0:  X              Y               assert X <= Y
//  c2:  X              ?, B            assert X <= B
//  c2:  ?, A           Y               true(?)
//  c6:  ?, B           ?, C            true(?)
LogicalResult LessEqualThanWithBounds(
    Optional<Location> location, int64_t dim, 
    int64_t leftSize, int64_t rightSize, 
    int64_t leftBound, int64_t rightBound) {
  // TODO
}
```

sdasgup3: Thanks for summarizing. 
Added a ticket  to track this.

burmako (approved):

zhouxin913 (approved):

PR #794, opened 2022-12-20
Add channel_id to AllToAllOp

burmako:
Rationale for approval is the same as in #272, #388, #403 and #673: this is a non-controversial backward-compatible change, and accepting it doesn't violate any of the existing commitments (it sticks to existing HLO semantics, and it is compatible with the extent of the current compatibility commitments).

MLIR-HLO commit: tensorflow/mlir-hlo@bd07cb9.

PR #826, opened 2022-12-24
Update the spec of AllToAllOp

burmako:
This PR follows up on #794 and updates the spec which I forgot to update originally.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -682,8 +682,12 @@ Within each process group in the StableHLO grid, splits the values of the
 between the processes, concatenates the scattered parts along `concat_dimension`
 and produces a `result` tensor.
 
-The operation splits the StableHLO grid into process groups using the
-`cross_replica(replica_groups)` strategy.
+The operation splits the StableHLO grid into `process_groups` as follows:
+
+  * `channel_id <= 0`,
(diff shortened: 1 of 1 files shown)


docs/spec.md
```diff
@@ -764,8 +764,12 @@ Within each process group in the StableHLO grid, splits the values of the
...
 and produces a `result` tensor.
 
-The operation splits the StableHLO grid into process groups using the
-`cross_replica(replica_groups)` strategy.
+The operation splits the StableHLO grid into `process_groups` as follows:
+
+  * `channel_id <= 0`,
+    `cross_replica(replica_groups)`.
+  * `channel_id > 0`,
+    `cross_partition(replica_groups)`.
```
sdasgup3, line 772: Shouldn't the constraint C6 also needs to change ?

(C6) `size(replica_groups)` depends on the process grouping strategy:
    * If `cross_replica`, `num_replicas`.
    * If `cross_partition`, `num_partition`.

sdasgup3 (commented):
LGTM with a single change suggestion.

docs/spec.md
burmako: Nice catch!

PR #840, opened 2022-12-27
First step towards turning spec.md into a programming language spec

burmako:
As the first step towards implementing #484, this PR introduces an EBNF grammar for the StableHLO language, modelling it after the generic assembly format of the StableHLO dialect.

In addition to creating a foundation for future work as described in
 #484, this PR immediately improves clarity of the Types and Constants
sections of the specification.

Next steps involve: 1) writing the Notation section along the lines of
 #289, 2) formalizing execution of the StableHLO language in the
Execution section.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -1,129 +1,32 @@
 # StableHLO Specification
 
-## Types
+StableHLO is an operation set for high-level operations (HLO) in machine
+learning (ML) models. StableHLO works as a portability layer between different
+ML frameworks and ML compilers: ML frameworks that produce StableHLO programs
+are compatible with ML compilers that consume StableHLO programs.
 
-Following are the supported element types in StableHLO:
+Our goal is to simplify and accelerate ML development by creating more
+interoperability between various ML frameworks (such as TensorFlow, JAX and
+PyTorch) and ML compilers (such as XLA and IREE). Towards that end, this
+document provides a specification for the StableHLO programming language.
 
-  * **Integer types**
-    * Signed integer with two’s complement representation. Referred to in the
-    document as `si<N>`, where the bit-width N ∊ {4, 8, 16, 32, 64}.
-    * Unsigned integer referred to in the document as `ui<N>`, where the
-    bit-width N ∊ {4, 8, 16, 32, 64}.
-  * **Boolean type** referred to in the document as `i1`. Exact
-  representation of boolean types (e.g. 1 byte per boolean vs 1 bit per boolean)
-  is implementation-defined.
-  * **Floating-point types**
-    * Single precision `f32`, double precision `f64` and half precision `f16`
-    floating-points complying with [IEEE 754-2019
-    format](https://ieeexplore.ieee.org/document/8766229).
-    * Bfloat16 `bf16` floating-point complying with [BFloat16 format](https://cloud.google.com/blog/products/ai-machine-learning/bfloat16-the-secret-to-high-performance-on-cloud-tpus).
-    Provides the same number of exponent bits as `f32`, so that it matches its
-    dynamic range, but with greatly reduced precision. This also ensures
-    identical behavior for underflows, overflows, and NaNs. However, `bf16`
-    handles denormals differently from `f32`: it flushes them to zero.
-    * FP8 `f8E4M3FN` and `f8E5M2` types corresponding to respectively the
-    `E4M3` and `E5M2` types from the whitepaper
-    [FP8 Formats for Deep Learning](https://arxiv.org/abs/2209.05433).
-  * **Complex types** represent a pair of floating-point types. Supported ones
-    are `complex<f32>` (represents a par of `f32`) and `complex<f64>`
-    (represents a pair of `f64`). Exact representation of complex types
-    (e.g. whether the real part or the imaginary part comes first in memory)
-    is implementation-defined.
-
-**Tensor types** are the cornerstone of the StableHLO type system. They model
-immutable n-dimensional arrays and are referred to in the document as
-`tensor<SxE>` where:
-
-  * **Shape** `S` represented as `(d0)x(d1)x...x(dR-1)` is a 1-dimensional array
-  of **dimension sizes** `di`, in the increasing order of the corresponding
-  **dimensions** (which are also called **axes**) 0, 1, ..., R-1.
-  The size `R` of this array is called **rank**. Dimension sizes have type
-  `si64` and are non-negative (dimension sizes equal to zero are allowed,
-  and their meaning is described below). Ranks equal to zero are also allowed,
-  and their meaning is also described below.
-  * **Element type** `E` is any one of the supported element types mentioned
-  above.
-
-For example, `tensor<2x3xf32>` is a tensor type with shape `2x3` and element
-type `f32`. It has two dimensions (or, in other words, two axes) whose sizes
-are 2 and 3. Its rank is 2.
-
-At the logical level, a `tensor<SxE>` maps a 1-dimensional array of **indices**
-`{i0, i1, ..., iR-1}` on **elements** of type `E`. If a tensor `t` maps an index
-`i` on an element `e`, we say that `t[i0, i1, ..., iR-1] = e`.
-
-Individual indices have type `si64` and are within the range `[0, di)` defined
-by the corresponding dimension. The size of the index array is equal to `R`.
-At the moment, StableHLO only supports dense tensors, so each tensor has
-`1*(d0)*(d1)*...*(dR-1)` elements whose indices are drawn from an
-**index space** which is a Cartesian product of its dimensions. For example:
-
-  * `tensor<2x3xf32>` has 6 elements whose indices are
-    `{0, 0}`, `{0, 1}`, `{0, 2}`, `{1, 0}`, `{1, 1}` and `{1, 2}`.
-  * Tensors of rank zero, e.g `tensor<f32>`, have 1 element. Such tensors are
-    allowed and are useful to model scalars.
-  * Tensors with dimensions of size zero, e.g. `tensor<2x0xf32>`, have
-    0 elements. Such tensors are allowed and are useful in rare cases, e.g.
-    to model empty slices.
-
-**Canonical representation** of a tensor is a 1-dimensional array of elements
-which correspond to indices ordered lexicographically. For example, for a
-`tensor<2x3xf32>` with the following mapping from indices to elements:
-`{0, 0} => 1`, `{0, 1} => 2`, `{0, 2} => 3`, `{1, 0} => 4`, `{1, 1} => 5`,
-`{1, 2} => 6` - the canonical representation would be: `[1, 2, 3, 4, 5, 6]`.
-
-Exact representation of tensors is implementation-defined. This specification
-does not define in which order tensor elements are laid out in memory (e.g.
-whether/when they follow the canonical order) and how individual tensor elements
-in a particular order are packed together into a tensor (e.g. how these elements
-are aligned, whether they are stored contiguously, etc).
-
-**Token type** Values of this type are used for imposing order on execution of
-side-effecting operations using data dependencies.
-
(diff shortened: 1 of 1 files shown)


sdasgup3 (approved):

docs/spec.md
```diff
@@ -140,240 +43,355 @@ stablehlo.func @main(
...
+### Functions
 
-### Sequential execution
+```ebnf
+Func        ::= 'stablehlo' '.' 'func' FuncId FuncInputs FuncOutputs '{' FuncBody '}'
+FuncInputs  ::= '(' [FuncInput {',' FuncInput}] `)`
+FuncInput   ::= '%' ValueId ':' ValueType
+FuncOutputs ::= ['->' FuncOutput, {',' FuncOutput}]
+FuncOutput  ::= ValueType
+FuncBody    ::= {Op}
```
GleasonK, line 54: A notation section would be great, saw in separate thread that it is underway which is great.

In the meantime, does `{ Thing }` mean zero-or-more, or one-or-more?

We allow empty programs `Program ::= Func*`, we do not allow empty function bodies `FuncBody ::= Op+`, or maybe `FuncBody ::= {Op} ReturnOp` may be more accurate if we want to keep the current notation.

_Unless_ we are capturing the language of unverified programs, then this is accurate.

docs/spec.md
```diff
@@ -140,240 +43,355 @@ stablehlo.func @main(
...
+### Identifiers
 
-More formally, a **StableHLO process** is a combination of:
-(1) a StableHLO program, (2) operation statuses (not executed yet,
-already executed), and (3) intermediate values that the process is working on.
-The process starts with input values to the `main` function, progresses through
-the graph of ops updating operation statuses and intermediate values and
-finishes with output values. Further formalization is TBD.
+```ebnf
+FuncId  ::= '@' letter {letter | digit}
```
GleasonK, line 65: This is not enforced currently. We currently support MLIR identifiers: https://mlir.llvm.org/docs/LangRef/#identifiers-and-keywords

```
  stablehlo.func @func.with.dot() {
    stablehlo.return
  }
```

docs/spec.md
GleasonK: Just seeing below that we have `digit {digit}` which makes me think it means zero-or-more. In that case I'd suggest the return op change

docs/spec.md
GleasonK: This section may be easier to write if we define an "identifier" production of our own

docs/spec.md
```diff
@@ -140,240 +43,355 @@ stablehlo.func @main(
...
+Type         ::= ValueType | NonValueType
+ValueType    ::= TensorType | TokenType | TupleType
+NonValueType ::= ElementType | FunctionType | StringType
+```
 
-The size of the grid is known statically for every program, and the position
-within the grid is known statically for every process. Each process has access
-to its position within the grid via the `replica_id` and `partition_id` ops.
+**StableHLO types** are categorized into **value types** (which are also called
+**first-class types**) which represent StableHLO values and **non-value types**
```
GleasonK, line 86: "StableHLO values"

Is the definition of a value: A type that can be an input or output of an Op ?

docs/spec.md
```diff
@@ -140,240 +43,355 @@ stablehlo.func @main(
...
+used to represent variadic inputs and outputs. In StableHLO, variadic inputs and
+outputs are supported natively, and the only use of tuples in StableHLO is to
+comprehensively represent HLO ABI where e.g. `T`, `tuple<T>` and
+`tuple<tuple<T>>` may be materially different depending on a particular
+implementation.
+
+```ebnf
+ElementType ::= BooleanType | IntegerType | FloatType | ComplexType
+BooleanType ::= 'i1'
+IntegerType ::= 'si4' | 'si8' | 'si16' | 'si32' | 'si64'
```
GleasonK, line 130: We support `4/8/16/32/64-bit signless integer or 4/8/16/32/64-bit unsigned integer`.

This should probably read `'i4' | 'i8' | 'i16' | 'i32' | 'i64'` as we error on `si`.

docs/spec.md
```diff
@@ -140,240 +43,355 @@ stablehlo.func @main(
...
 
-### Point-to-point communication
+**Element types** represent elements of tensor types. Unlike in many programming
+languages, these types are not first class in StableHLO. This means that
+StableHLO programs cannot directly represent values of these types (as a result,
+it is idiomatic to represent scalar values of type `T` with 0-dimensional tensor
+values of type `tensor<T>`).
+
+  * **Boolean type** represents boolean values `true` and `false`.
+  * **Integer types** can be either signed (`si`) or unsigned (`ui`) and have
```
GleasonK, line 143: Same `si` feedback

docs/spec.md
```diff
@@ -140,240 +43,355 @@ stablehlo.func @main(
...
+    `binary16` ("half precision"), `binary32` ("single precision") and
+    `binary64` ("double precision") formats described in
+    [the IEEE 754 standard](https://ieeexplore.ieee.org/document/8766229).
+  * **Complex types** represent complex values that have a **real part**
+    and an **imaginary part** of the same **element type**. Supported complex
+    types are `complex<f32>` (both parts are of type `f32`) and `complex<f64>`
+    (both parts are of type `f64`).
+
+```ebnf
+FunctionType ::= '(' [ValueType {',' ValueType}] ')' '->' '(' [ValueType {',' ValueType}] ')'
```
GleasonK, line 164: Could define `ValueTypes` alongside Tuple and reuse it here. Totally optional, if we want to avoid unnecessary productions this is fine.

docs/spec.md
```diff
@@ -140,240 +43,355 @@ stablehlo.func @main(
...
+```
 
-StableHLO processes can communicate with each other through
-**StableHLO channels**. A channel is represented by a positive id of type
-`si64`. Through various ops, it is possible to send values to channels and
-receive them from channels.
+**Function types** represent both named and anonymous functions. They have
+input types (the list of types on the left-hand side of `->`) and output types
+(the list of types on the right-hand side of `->`). In many programming
+languages, function types are first class, but not in StableHLO.
```
GleasonK, line 170: Via reflection / metaprogramming API's? This represents the functions signature IIUC

docs/spec.md
```diff
@@ -140,240 +43,355 @@ stablehlo.func @main(
...
 
-Further formalization, e.g. where these channel ids are coming from, how
-processes programs become aware of them and what kind of synchronization is
-introduced by them, is TBD.
+```ebnf
+StringType ::= 'string'
```
GleasonK, line 173: Confused by this. Do we have this type? Should this be `'"' { chars } '"'`

This notation makes me think that ` : string` can appear in a program, which I'm not sure it can?

```
  %0 = "stablehlo.unary_einsum"(%arg0) {
    einsum_config = "ab->a" : string
  } : (tensor<8x16xf32>) -> tensor<8xf32>


stablehlo-opt /tmp/t.mlir -mlir-print-op-generic
/tmp/t.mlir:3:30: error: expected non-function type
    einsum_config = "ab->a" : string
```

docs/spec.md
```diff
@@ -140,240 +43,355 @@ stablehlo.func @main(
...
+the name "anonymous"), 2) they don't declare output types (output types are
+inferred from the `return` op within the function).
+
+The syntax for input functions includes a currently unused part (see the
+`Unused` production above) which is there for compatibility with MLIR. In MLIR,
+there is a more general concept of "regions" which can have multiple "blocks"
+of ops connected together via jump ops. These blocks have ids which correspond
+to the `Unused` production, so that they can be distinguished from each other.
+StableHLO doesn't have jump ops, so the corresponding part of MLIR syntax is
+unused (but is still there).
```
GleasonK, line 235: Huh, block ID's aren't even required to be unique. Didn't know that. Section is accurate!

docs/spec.md
```diff
@@ -140,240 +43,355 @@ stablehlo.func @main(
...
+
+```ebnf
+OpInputAttr      ::= OpInputAttrName '=' OpInputAttrValue
+OpInputAttrName  ::= letter {letter | digit}
```
GleasonK, line 239: Similar to functions, these allow dots as well. However, we leave that area rather unspecced since there isn't a good use case for these. So I think leaving it out is fine.

docs/spec.md
GleasonK: Ok I think I figured this out after reading into the later sections, this is an isolated production that will never be produced. Might be worth adding a note that this is an implicit type that doesn't appear in the text of the program.

docs/spec.md
```diff
@@ -140,240 +43,355 @@ stablehlo.func @main(
...
+the `slice` op uses multiple attributes like `start_indices` and `limit_indices`
+to specify the bounds that are used to slice the input value.
 
-If the process group involves cross-partition communication, i.e. there are
-processes in the process group whose partition ids are different, then execution
-of the collective op needs a channel, and the collective op must provide a
-positive `channel_id` of type `si64`. Cross-replica communication doesn't need
-channels.
+```ebnf
+OpSignature ::= '(' [ValueType {',' ValueType}] ')' '->' '(' [ValueType {',' ValueType}] ')'
```
GleasonK, line 251: The following line  (a few lines below) violates this definition:

```
%result = "stablehlo.select_and_scatter"(%operand, %source, %init_value) ({
  ^bb0(%arg0: tensor<i32>, %arg1: tensor<i32>):
    %0 = "stablehlo.compare"(%arg0, %arg1) {
      comparison_direction = #stablehlo<comparison_direction GE>
    } : (tensor<i32>, tensor<i32>) -> tensor<i1>
    "stablehlo.return"(%0) : (tensor<i1>) -> ()
}
```

Specifically `(tensor<i32>, tensor<i32>) -> tensor<i1>` would need to be `(tensor<i32>, tensor<i32>) -> (tensor<i1>)` under the current grammar. Could either resolve by updating the production or the example.

docs/spec.md
```diff
@@ -140,240 +43,355 @@ stablehlo.func @main(
...
+**Boolean constants** represent boolean values `true` and `false`. Boolean
+constants have type `i1`.
+
+```ebnf
+IntegerConstant   ::= IntegerLiteral ':' IntegerType
+IntegerLiteral    ::= ['-' | '+'] DecimalDigits
+                    | ['-' | '+'] '0x' HexadecimalDigits
+DecimalDigits     ::= decimalDigit {decimalDigit}
+HexadecimalDigits ::= hexadecimalDigit {hexadecimalDigit}
+decimalDigit      ::= '0' | ... | '9'
```
GleasonK, line 316: What is the lowercase vs capital letters scheme used here? If it is that lowercase means it only produces terminals, should it be used for `booleanLiteral` above as well?

docs/spec.md
```diff
@@ -140,240 +43,355 @@ stablehlo.func @main(
...
-  * **Tensor constants** use NumPy notation. For example,
-  `[[1, 2, 3], [4, 5, 6]]` is a constant of type `tensor<2x3xf32>` with the
-  following mapping from indices to elements: `{0, 0} => 1`, `{0, 1} => 2`,
-  `{0, 2} => 3`, `{1, 0} => 4`, `{1, 1} => 5`, `{1, 2} => 6`.
-  * **String constants** String constants are represented as a sequence of
-  bytes enclosed in double quotation mark symbols, e.g. "foo123?" (in ASCII
-  encoding) or "\18\A3" (in hex encoding).
+**String literals** consist of bytes specified using ASCII characters and
+escape sequences. They are encoding-agnostic, so the interpretation of these
+bytes is implementation-defined. String literals have type `string`.
```
GleasonK, line 394: "String literals have the implicit type `string`"

GleasonK (commented):
This is great! Left a bunch of nitpicks.

soraros:
I think the definition for inline comment (`// ...`) is also missing from the grammar?

burmako:
Thank you everyone for your interest and feedback! I'll reply in early January after the winter holidays.

PR #841, opened 2022-12-27
Lint cleanup for list indentation

scottamain:
Clarify the style/lint rules for lists and apply new no-indentation for list items (except nested lists).


docs/bytecode.md
```diff
@@ -31,48 +31,48 @@ in the following code comments:
...
+  - `HLO_UInt`
 
 **Special Cases:**
 
-- `StableHLO_ConvolutionAttributes`
-  - Despite its name,  is not an attribute and is not encoded.
+  - `StableHLO_ConvolutionAttributes`
+    - Despite its name,  is not an attribute and is not encoded.
     Rather, it is a dag which gets expanded into several attributes
     which are all encoded separately.
```
burmako, line 73: If I recall correctly, indenting multiline bullets is not something that is currently done automatically by the tool, right? How should we approach this - can this only be fixed manually (doesn't have to be done in this PR, just trying to understand this better)? Also, is this something that the linter will flag in the future?

burmako (approved):

docs/bytecode.md
scottamain: Yeah. Well, I didn't realize that this indentation style was inconsistent across docs until now. It appears that spec.md was the only file that used 2-space indentation for all lists. The rest of these had none and now changing them to two creates this misalignment. 

So, there's some manual effort required for this no matter what (to make multiline list items left aligned). As such, the decision about whether to use `start_indented: true` should be either based on reducing the total amount of manual effort required (which is probably to use it---keep it `true`---because spec.md is a much larger document than the rest of these combined) or based on logic (which I believe leads to not using it because the extra indentation serves no real purpose and is actually harmful because it subtracts 2 characters from the available line limit, and arguably slows the writing process).

That is, I think using `start_indented: false` (no indentation for the list) is better if we were starting from scratch because it saves space and the only reason to keep it set `true` is because spec.md already uses 2-space indents.

You and your team are ultimately the authors here, so I think the decision is yours. Although the evidence suggests there's no strong preference to indent because not all documents do it, I'll honor your wishes if you want to do it that way for all these pages as well.

scottamain: This actually isn't that much work to align all these, either way, and I'll do it. So just confirm for me which style you prefer.

.markdownlint.yaml
```diff
@@ -13,11 +13,32 @@
 
 # markdownlint-cli configuration file.
 
-# Enable all rules by default
+# Enable all markdownlint rules by default
 default: true
 
-# Prefer indented lists for legacy compatibility
+# List formatting:
```
ghpvnist, line 19: Nit: blank space here.

scottamain: Well, this isn't actually markdown. :)

docs/spec.md
```diff
@@ -139,26 +139,26 @@ StableHLO programs cannot directly represent values of these types (as a result,
...
+* **Boolean type** represents boolean values `true` and `false`.
+* **Integer types** can be either signed (`si`) or unsigned (`ui`) and have
+  one of the supported bit widths (`4`, `8`, `16`, `32` or `64`).
+  Signed `siN` types represent integer values from `-2^(N-1)` to `2^(N-1)-1`
+  inclusive, and unsigned `uiN` types represent integer values from `0` to
+  `2^N-1` inclusive.
+* **Floating-point types** can be one of the following:
+  * `f8E4M3FN` and `f8E5M2` types corresponding to respectively the
+  `E4M3` and `E5M2` encodings of the FP8 format described in
+  [FP8 Formats for Deep Learning](https://arxiv.org/abs/2209.05433).
```
burmako, line 151: Re: "In multiline list items, subsequent lines are indented by 2 spaces". In some parts of our Markdown files, e.g. this one, this convention doesn't hold. Do we want to update these files in this PR, or that will be done separately?

.markdownlint.yaml
```diff
@@ -12,13 +12,35 @@
...
+# * This is a list item that has multiple
+#   lines and each line aligns with the text from the first line.
+#   * This is a nested list, also aligned with the first line.
+#
+# For ordered lists, that means three spaces for wrapped lines:
+#
+# 1. This is an ordered list item.
+#    1. The nested list aligns with the first line.
 ul-indent:
-  start_indented: true
```
burmako, line 21: Perhaps let's explicitly call out `start_indented: false`?

burmako (approved):
Scott, thank you for pushing this to a resolution! I appreciate that we now have clear conventions for Markdown.

docs/spec.md
scottamain: Ah, that one was a missed fix. I've now indented these lines. Unfortunately, markdownlint does not check for alignment like this, so we just have to make a best-effort. There is also an exception for lists of numbered constraints: there's an unspoken convention to align with the text following the C number. For example:

```
* (C7) $0 \le$ `replica_groups`[i] $\lt$ size(`replica_groups`) $\forall i$
       in `indices(replica_groups)`.
```

I think this is well and good to aid in readability. I think the general idea is to indent *at least* 2 characters, but I worry that explaining this exception would require far too much text. So let's just be reasonable with adherence to the styles that aren't lint checked. :)

.markdownlint.yaml
scottamain: I'd prefer not. There are a lot of default properties like this ([see them all here](https://github.com/DavidAnson/markdownlint/blob/main/schema/.markdownlint.yaml), so I think it's best to just keep our config limited to the overrides.

scottamain: Oops. I'm a hypocrite. `indent: 2` is also the default, but if we don't have that rule declared here then we have no context for our style guide for list items. So I think we should keep that one here until we have another document where we describe all the Markdown style guidelines such as these list styles.

PR #843, opened 2022-12-27
Remove "stablehlo" from "#stablehloadd" links

burmako:
The "stablehlo" part of spec sections and spec permalinks is redundant.
It's a StableHLO spec, so it is reasonable to assume that it specs
StableHLO ops.


sdasgup3 (approved):

burmako:
Force merging because changes in StablehloOps.td are documentation only.

PR #849, opened 2022-12-28
Fork Attributes and Types for VHLO

### LINKED ISSUE #674: Add forked types and attributes for VHLO

GleasonK:
- Change `CustomCallApiVersion` to an enum instead of an integer value.
  + To avoid transforming _all_ integers, the transformation from StableHLO --> VHLO special cases this conversion
- Attributes
  + Forked Attributes: `IntegerAttr, StringAttr, UnitAttr, ArrayAttr`, `DenseIntOrFPElementsV1Attr, FlatSymbolRefV1Attr, FloatV1Attr`
- Types
  + Forked Types: `RankedTensorType, UnrankedTensorType, TupleType, WitnessType`, `BFloat16V1Type, Float16V1Type, Float32V1Type, Float64V1Type, IndexV1Type, ComplexV1Type, IntegerV1Type, UniformQuantizedV1Type`
- Bytecode implementations forked from [BuiltinDialectBytecode.cpp](https://github.com/llvm/llvm-project/blob/c48e0cf03a50bb8a2043ac4bb5e9a83ff135247a/mlir/lib/IR/BuiltinDialectBytecode.cpp)

Need a closer review on FloatAttr since it deals with numerics, I have a FIXME comment next to that line still. I may revert from forking that value so that the implementation doesn't have to deal with numeric value conversions.

Closes #674.


stablehlo/dialect/VhloAttrs.td
```diff
@@ -56,6 +56,34 @@ class VHLO_AttrDef<string name,
   }];
 }
 
+//===----------------------------------------------------------------------===//
+// Wrapped Attributes
+//===----------------------------------------------------------------------===//
+
+def VHLO_AttrWrap : VHLO_AttrDef<"Wrapped"> {
```
burmako, line 63: I probably won't have time to review this in detail today, but I wanted to ask a high-level question that we could already iterate on. Given that both the attributes and the types are currently split between forking and not forking, how about we consistently fork everything? How much additional work would that be?

burmako (changes requested):

stablehlo/dialect/VhloAttrs.td
GleasonK: Not too much. I agree this will be helpful. We'll have to be a bit careful on forking `FloatAttr` and `IntegerAttr` to make sure we maintain proper textual parsing/printing of numerics, but otherwise I think it's fairly simple. I'll get to that addition.

GleasonK: Forked all attributes. Will do the same for types soon. I still wrapped integer, but with an integer type that validates the underlying value and will likely do the same for float. It pushes a bit of code into bytecode, but I think it works out better for printing/parsing/memory storage of values.

stablehlo/dialect/VhloBytecode.cpp
```diff
@@ -696,6 +927,185 @@ void VhloBytecodeInterface::write(TypeExtensionsAttr attr,
   writer.writeSignedVarInts(attr.getBounds());
 }
 
+//===----------------------------------------------------------------------===//
+// Forked Attributes
+//===----------------------------------------------------------------------===//
+
+namespace {
+template <typename AttrOrType>
+bool assertFromVhlo(AttrOrType val) {
```
burmako, line 936: I wonder if we can check this via verifiers instead - both for attributes and types. This would allow for much better error messages.

stablehlo/dialect/VhloTypes.td
```diff
@@ -60,9 +63,92 @@ class VHLO_TypeDef<string name,
...
+
+def VHLO_Double : APFloatParameter<""> {
+  // Used by UniformQuantizedV1 to have APFloat with custom parser
+  let parser = [{
+    [&]() -> FailureOr<llvm::APFloat> {
+      double value;
+      if (failed($_parser.parseFloat(value))) {
+        return failure();
+      }
+      return APFloat(value);
```
burmako, line 129: Why do we need `APFloat` here? The upstream type uses `double`.

stablehlo/transforms/TypeConversion.h
```diff
@@ -28,99 +33,189 @@ namespace vhlo {
...
+                                     convertedElementType, convertedEncoding);
+    });
+    addConversion([&](TupleType type) -> Type {
+      SmallVector<Type> convertedTypes;
+      if (failed(convertTypes(type.getTypes(), convertedTypes))) return {};
+      return vhlo::TupleV1Type::get(type.getContext(), convertedTypes);
+    });
+    addConversion([&](quant::UniformQuantizedType type) -> Type {
+      Type storage = convertType(type.getStorageType());
+      Type expressed = convertType(type.getExpressedType());
```
burmako, line 94: `convertedStorageType` and `convertedExpressedType`? (Same below).

stablehlo/transforms/TypeConversion.h
```diff
@@ -28,99 +33,189 @@ namespace vhlo {
...
+      if (!storage || !expressed) return {};
+      return vhlo::UniformQuantizedV1Type::get(
+          type.getContext(), type.getFlags(), storage, expressed,
+          APFloat(type.getScale()), type.getZeroPoint(),
+          type.getStorageTypeMin(), type.getStorageTypeMax());
+    });
+    addConversion([&](UnrankedTensorType type) -> Type {
+      auto convertedElementType = convertType(type.getElementType());
+      if (!convertedElementType) return {};  // unsupported element type
```
burmako, line 103: I don't think this comment buys us much.

stablehlo/transforms/TypeConversion.h
```diff
@@ -28,99 +33,189 @@ namespace vhlo {
...
+    });
+    addConversion([&](shape::WitnessType type) -> Type {
+      return vhlo::WitnessV1Type::get(type.getContext());
+    });
   }
 
-  bool isSourceDialect(Dialect& dialect) final {
-    return dialect.getNamespace() ==
-           stablehlo::StablehloDialect::getDialectNamespace();
+  bool isTargetDialect(Dialect& dialect) {
```
burmako, line 111: This function seems to only be used once. Inline?

stablehlo/dialect/VhloBytecode.cpp
```diff
@@ -696,6 +927,185 @@ void VhloBytecodeInterface::write(TypeExtensionsAttr attr,
...
+  if (val.getDialect().getNamespace() != "vhlo") {
```
burmako, line 937: `vhlo::VhloDialect::getDialectNamespace()` instead of `"vhlo"`? (Here and in a few other places).

stablehlo/transforms/VhloLegalizeToStablehlo.cpp
```diff
@@ -100,25 +102,56 @@ Attribute convertAttrToStablehlo(Attribute vhloAttr) {
...
-    for (auto vhloAttr : vhloAttrs) {
-      auto stablehloAttr = convertAttrToStablehlo(vhloAttr);
+    for (auto vhloAttr : vhloAttrs.getValue()) {
+      auto stablehloAttr = convertAttrToStablehlo(vhloAttr, typeConverter);
       if (!stablehloAttr) return {};
       stablehloAttrs.push_back(stablehloAttr);
     }
     return ArrayAttr::get(vhloAttrs.getContext(), stablehloAttrs);
   }
+  if (auto denseElements =
```
burmako, line 116: The code above would use `attr` to name these auto variables.

stablehlo/transforms/VhloLegalizeToStablehlo.cpp
```diff
@@ -100,25 +102,56 @@ Attribute convertAttrToStablehlo(Attribute vhloAttr) {
...
+          vhloAttr.dyn_cast<vhlo::DenseIntOrFPElementsV1Attr>()) {
+    auto builtinType = typeConverter->convertType(denseElements.getType());
+    if (!builtinType) return {};
+    return DenseIntOrFPElementsAttr::getFromRawBuffer(
+        builtinType, denseElements.getRawData());
+  }
+  if (auto flatSymAttr = vhloAttr.dyn_cast<vhlo::FlatSymbolRefV1Attr>()) {
+    auto rootRef =
```
burmako, line 124: `builtinRootRef`? (Same for the temporary variables in the code below).

stablehlo/transforms/VhloLegalizeToStablehlo.cpp
```diff
@@ -100,25 +102,56 @@ Attribute convertAttrToStablehlo(Attribute vhloAttr) {
...
+  // All VHLO Attributes must be converted by now.
+  if (vhloAttr.getDialect().getNamespace() ==
+      vhlo::VhloDialect::getDialectNamespace()) {
+    // All VHLO attributes must have counterparts in StableHLO.
+    return {};
+  }
+
+  // This should be unreachable unless program is a mix of VHLO and other
+  // due to user edits to textual assembly format.
   return vhloAttr;
```
burmako, line 155: `return {}` then?

stablehlo/transforms/VhloToVersion.cpp
```diff
@@ -78,59 +87,90 @@ FailureOr<Version> validateTargetVersion(llvm::StringRef versionRef,
...
+        succeeded(isLegalType(quant.getStorageType(), targetVersion)) &&
+        succeeded(isLegalType(quant.getExpressedType(), targetVersion)));
+  }
+  if (auto unranked = type.dyn_cast<UnrankedTensorV1Type>()) {
+    return isLegalType(unranked.getElementType(), targetVersion);
+  }
+  if (auto tuple = type.dyn_cast<TupleV1Type>()) {
+    return success(llvm::all_of(tuple.getTypes(), [&](Type ele) {
+      return succeeded(isLegalType(ele, targetVersion));
+    }));
```
burmako, line 157: If this is supposed to be ordered alphabetically, `TupleV1Type` is out of place.

stablehlo/transforms/VhloToVersion.cpp
```diff
@@ -78,59 +87,90 @@ FailureOr<Version> validateTargetVersion(llvm::StringRef versionRef,
...
   // Validate op
   auto opInterface = dyn_cast<VersionedOpInterface>(op);
   if (!opInterface) return false;
   if (!isLegalVersion(opInterface, targetVersion)) return false;
+  LLVM_DEBUG(llvm::dbgs() << "Legal version for target. " << op << '\n');
 
   // Validate attributes
-  auto isLegalAttrFn = [&](NamedAttribute attr) {
-    return succeeded(isLegalAttribute(attr, targetVersion));
+  auto isLegalAttrFn = [&](NamedAttribute const& attr) {
```
burmako, line 172: `const NamedAttribute&`?

burmako (changes requested):

stablehlo/dialect/VhloAttrs.td
```diff
@@ -203,4 +166,55 @@ def VHLO_ConvolutionAttributes {
   );
 }
 
+//===----------------------------------------------------------------------===//
+// Forked Attributes
+//===----------------------------------------------------------------------===//
+
+def VHLO_ArrayAttr : VHLO_AttrDef<"ArrayV1"> {
```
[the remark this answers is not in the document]
GleasonK: This is more important for types/attrs than ops because it gets very confusing what `ArrayAttr` is referring to in a namespace, since we nest namespaces `namespace mlir { namespace vhlo { ... ArrayAttr::get(...)  }}` in the same namespace. If we include the file with `vhlo::ArrayAttr` this is resolved to VHLO, otherwise it gets resolved to `mlir::ArrayAttr`. Adding a V1 to the type name prevents any ambiguity. This impacts ops less since we don't nest namespace VHLO/StableHLO anywhere.

I'll update all names to be `VHLO_ArrayAttrV1`, but will consider suffixing all ops in a later submission.

stablehlo/dialect/VhloAttrs.td
```diff
@@ -203,4 +166,55 @@ def VHLO_ConvolutionAttributes {
...
+  let mnemonic = "array";
+  let parameters = (ins ArrayRefParameter<"mlir::Attribute">:$value);
```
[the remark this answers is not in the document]
GleasonK: Tried to avoid wrapping as much as possible. Rationale, in Bytecode we call into `writer.writeType`/`writer.writeAttriute` to write any nested types (nested in the sense that IntegerAttr holds an IntegerType, or TensorType has an Attribute encoding). If we wrap, we cannot use those methods and have non-standard bytecode which may cause issues. The `writeType` method logs some information, then calls into the dialect interface to write the type. It is also super useful to rely on the nested type pretty printers for numerics.

The only time I wrap currently (I believe) is IntegerType/IntegerAttr. I have heavily considered wrapping FloatAttr as well. The less we deal with numerics in VHLO the more concise it can be. You'll notice that VhloBytecode has to do special handling for IntegerAttr type as a result of the wrapping, but I think it is worth it in those cases.

stablehlo/dialect/VhloBytecode.cpp
```diff
@@ -163,6 +172,52 @@ enum AttributeCode {
   ///     operandTupleIndices: svarint[]
   ///   }
   kOutputOperandAlias = 14,
+
+  ///   CustomCallApiVersionAttr
+  ///     value: varint (encoded enum)
+  ///   }
+  kCustomCallApiVersionAttr = 15,
+
+  // Gap in numbers to separate StableHLO types from forked types //
```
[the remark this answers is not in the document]
GleasonK: We don't. Was just in case we wanted a few spaces to keep StableHLO attrs next to eachother so StablehloBytecode and VhloBytecode numbers could resemble eachother for a short while.

Will remove the gap.

stablehlo/dialect/VhloBytecode.cpp
```diff
@@ -696,6 +927,185 @@ void VhloBytecodeInterface::write(TypeExtensionsAttr attr,
...
+//===----------------------------------------------------------------------===//
+// ArrayV1Attr
+
+ArrayV1Attr VhloBytecodeInterface::readArrayAttr(
+    DialectBytecodeReader &reader) const {
+  LOG_READ_CALL;
+  SmallVector<Attribute> elements;
+  if (failed(reader.readAttributes(elements))) return ArrayV1Attr();
+
+  llvm::all_of(elements, assertFromVhlo<Attribute>);
```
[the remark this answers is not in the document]
GleasonK: It would be a `report_fatal_error` regardless of `NDEBUG` (assert is just a function name). Writing wasn't made for error reporting, much like asm printing isn't made for error reporting. Parsing/reading are.

I can flip the reader to generate errors on reading, but I still think the fatal error would be needed for printing.

stablehlo/dialect/VhloBytecode.cpp
```diff
@@ -696,6 +927,185 @@ void VhloBytecodeInterface::write(TypeExtensionsAttr attr,
...
+void VhloBytecodeInterface::write(IntegerV1Attr attr,
+                                  DialectBytecodeWriter &writer) const {
+  writer.writeVarInt(vhlo_encoding::kIntegerAttr);
+  LLVM_DEBUG(llvm::dbgs() << "IntegerAttr: " << attr << '\n');
+  auto intAttr = attr.getValue();
+  if (intAttr.getType().isa<IndexType>()) {
+    writer.writeType(IndexV1Type::get(getContext()));
+  } else {
+    assert(intAttr.getType().isa<IntegerType>());
+    // Wrap integer type in vhlo::IntegerV1Type
```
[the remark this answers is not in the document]
GleasonK: Updated the comment. I think it is important because this is a non-standard piece, since this is one of the only times we wrap builtin types/attrs:

```
    // Note, since IntegerV1Type wraps builtin IntegerAttr, we need to explicitly
    // convert to IntegerV1Type here to avoid calling into builtin dialect bytecode
    // writer.
```

stablehlo/dialect/VhloBytecode.cpp
```diff
@@ -723,6 +1160,35 @@ LogicalResult VhloBytecodeInterface::writeType(
...
         return success();
       })
+      .Case<ComplexV1Type, IntegerV1Type, RankedTensorV1Type, TupleV1Type,
+            UnrankedTensorV1Type, UniformQuantizedV1Type>([&](auto type) {
+        LOG_WRITE_CALL;
+        return write(type, writer), success();
+      })
+      .Case([&](BFloat16V1Type) {
+        LOG_WRITE_CALL;
+        return writer.writeVarInt(vhlo_encoding::kBFloat16Type), success();
```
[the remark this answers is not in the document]
GleasonK: Did this for consistency with the types being forked:
https://github.com/llvm/llvm-project/blob/main/mlir/lib/IR/BuiltinDialectBytecode.cpp#L950

stablehlo/dialect/VhloOps.cpp
```diff
@@ -56,6 +187,8 @@ VhloDialect::VhloDialect(MLIRContext* context)
 #define GET_ATTRDEF_LIST
 #include "stablehlo/dialect/VhloAttrs.cpp.inc"
       >();
+  context->loadDialect<shape::ShapeDialect>();
+  context->loadDialect<quant::QuantizationDialect>();
```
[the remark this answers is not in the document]
GleasonK: Ah, is that how we do it for StableHLO? How do we write tests for these types? These are needed for construction of the type, AKA if type conversion is ever called and these are not loaded we fail

stablehlo/dialect/VhloOps.cpp
```diff
@@ -16,11 +16,142 @@ limitations under the License.
...
+#include "mlir/IR/BuiltinAttributes.h"
+#include "mlir/IR/BuiltinTypes.h"
 #include "mlir/IR/TypeUtilities.h"
 #include "mlir/Support/LogicalResult.h"
 #include "stablehlo/dialect/AssemblyFormat.h"
 #include "stablehlo/dialect/VhloBytecode.h"
+#include "stablehlo/transforms/TypeConversion.h"
+
+namespace mlir {
+namespace vhlo {
```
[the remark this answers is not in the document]
GleasonK: See rationale for forking over wrapping in previous comment

stablehlo/tests/legalize_stablehlo_to_vhlo.mlir
```diff
@@ -382,11 +382,11 @@ func.func @op_all_reduce(%arg0: tensor<f32>) -> tensor<f32> {
 func.func @op_all_to_all(%arg0: tensor<4x16xf32>) -> tensor<16x4xf32> {
   //               CHECK: "vhlo.all_to_all_v2"(%arg0) {
   //          CHECK-SAME:   channel_handle = #vhlo.channel_handle<handle = 1, type = 0>
-  //          CHECK-SAME:   concat_dimension = 0 : i64,
-  // CHECK-SAME{LITERAL}:   replica_groups = dense<[[0, 1, 2, 3]]> : tensor<1x4xi64>,
-  //          CHECK-SAME:   split_count = 4 : i64,
-  //          CHECK-SAME:   split_dimension = 1 : i64
-  //          CHECK-SAME: } : (tensor<4x16xf32>) -> tensor<16x4xf32>
+  //          CHECK-SAME:   concat_dimension = #vhlo.integer<0 : i64>,
+  // CHECK-SAME{LITERAL}:   replica_groups = #vhlo.dense<dense<[[0, 1, 2, 3]]> : tensor<1x4xi64>>,
```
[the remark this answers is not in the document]
GleasonK: To avoid having to copy the entire implementation for dense printing/parsing, I create a temporary dense array and print/parse:

```
void DenseIntOrFPElementsV1Attr::print(mlir::AsmPrinter& p) const {
  VhloToStablehloTypeConverter conv;
  p << '<'
    << DenseIntOrFPElementsAttr::getFromRawBuffer(conv.convertType(getType()),
                                                  getRawData())
    << '>';
}

Attribute DenseIntOrFPElementsV1Attr::parse(AsmParser& parser, mlir::Type) {
  StablehloToVhloTypeConverter conv;
  DenseIntOrFPElementsAttr attr;
  if (failed(parser.parseLess()) || failed(parser.parseAttribute(attr)) ||
      failed(parser.parseGreater())) {
    return DenseIntOrFPElementsV1Attr();
  }
  return DenseIntOrFPElementsV1Attr::get(
      parser.getContext(), conv.convertType(attr.getType()), attr.getRawData());
}
```

As for "why not wrap", as described above, this would push conversions into the VhloBytecodeInterface code if we want to use the standard `writeType/Attribute` functions.

stablehlo/tests/vhlo_to_version_attrtype_invalid.mlir
```diff
@@ -0,0 +1,106 @@
...
+// -----
+
+// This simulates version validation if a new string attr is introduced.
+func.func @vhlo_illegal_attr_symbolref(%arg0: !vhlo.tensor<!vhlo.f32>) -> !vhlo.tensor<!vhlo.f32> {
+  // expected-error @+1 {{failed to legalize operation 'vhlo.custom_call_v2' that was explicitly marked illegal}}
+  %0 = "vhlo.custom_call_v2"(%arg0) {
+    api_version = #vhlo<api_version API_VERSION_ORIGINAL>,
+    backend_config = #vhlo.string<"">,
+    call_target_name = #vhlo.string<"foo">,
+    called_computations = #vhlo.array<[#vhlo.sym<dense<1> : tensor<i16>>]>
```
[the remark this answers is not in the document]
GleasonK: Hmm. I'm realizing that these attributes _are_ optional in VHLO, but I'm not sure they should be. 

```
// StablehloOps.td > CustomCallOp
let arguments = (ins
    Variadic<HLO_TensorOrTokenOrTuple>:$inputs,
    StrAttr:$call_target_name,
    DefaultValuedOptionalAttr<BoolAttr, "false">:$has_side_effect,
    DefaultValuedStrAttr<StrAttr, "">:$backend_config,
    DefaultValuedOptionalAttr<
        StableHLO_CustomCallApiVersionAttr,
        "::mlir::stablehlo::CustomCallApiVersion::API_VERSION_ORIGINAL">:
        $api_version,
    DefaultValuedOptionalAttr<StableHLO_FlatSymbolRefArrayAttr, "{}">:$called_computations,
    OptionalAttr<StableHLO_ArrayOfLayoutAttr>:$operand_layouts,
    OptionalAttr<StableHLO_ArrayOfLayoutAttr>:$result_layouts,
    DefaultValuedOptionalAttr<
        TypedArrayAttrBase<
            StableHLO_OutputOperandAlias,
            "Aliasing attribute for outputs and operands of CustomCall">,
        "{}">:$output_operand_aliases
  );
```

We decided that `DefaultValuedAttr --> VHLO_AnyAttr`, `OptionalAttr --> OptionalAttr<VHLO_AnyAttr>`, I'm not sure where we landed on `DefaultValuedOptionalAttr`. Still not a big fan of this descriptor, it holds implicit information in the op. If we keep VHLO as-is and the default value changes for a `DefaultValuedOptionalAttr`, this breaks compatibility. The same is not true for `DefaultValuedAttrs`

stablehlo/tests/legalize_stablehlo_to_vhlo.mlir
```diff
@@ -399,24 +399,24 @@ func.func @op_all_to_all(%arg0: tensor<4x16xf32>) -> tensor<16x4xf32> {
...
 }
 // CHECK-LABEL: "op_atan2"
 
 func.func @op_batch_norm_grad(%arg0: tensor<16x16x16x16xf32>, %arg1: tensor<16xf32>, %arg2: tensor<16xf32>, %arg3: tensor<16xf32>, %arg4: tensor<16x16x16x16xf32>) -> (tensor<16x16x16x16xf32>, tensor<16xf32>, tensor<16xf32>) {
   //      CHECK: "vhlo.batch_norm_grad"(%arg0, %arg1, %arg2, %arg3, %arg4) {
-  // CHECK-SAME:   epsilon = 1.000000e-03 : f32,
-  // CHECK-SAME:   feature_index = 0 : i64
-  // CHECK-SAME: } : (tensor<16x16x16x16xf32>, tensor<16xf32>, tensor<16xf32>, tensor<16xf32>, tensor<16x16x16x16xf32>) -> (tensor<16x16x16x16xf32>, tensor<16xf32>, tensor<16xf32>)
+  // CHECK-SAME:   epsilon = #vhlo.float<1.000000e-03 : !vhlo.f32>,
+  // CHECK-SAME:   feature_index = #vhlo.integer<0 : i64>
```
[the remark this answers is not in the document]
GleasonK: Float is forked, integer is wrapped. I'm thinking I'll wrap float as well, because I'm not sure I'm doing the float conversions correctly, and wrapping would prevent us from needing to worry about that. Would just take a bit of extra logic in the bytecode writer.

stablehlo/tests/legalize_stablehlo_to_vhlo.mlir
```diff
@@ -382,11 +382,11 @@ func.func @op_all_reduce(%arg0: tensor<f32>) -> tensor<f32> {
 func.func @op_all_to_all(%arg0: tensor<4x16xf32>) -> tensor<16x4xf32> {
   //               CHECK: "vhlo.all_to_all_v2"(%arg0) {
   //          CHECK-SAME:   channel_handle = #vhlo.channel_handle<handle = 1, type = 0>
-  //          CHECK-SAME:   concat_dimension = 0 : i64,
-  // CHECK-SAME{LITERAL}:   replica_groups = dense<[[0, 1, 2, 3]]> : tensor<1x4xi64>,
-  //          CHECK-SAME:   split_count = 4 : i64,
```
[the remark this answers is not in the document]
GleasonK: typo, added comma back

stablehlo/dialect/VhloOps.cpp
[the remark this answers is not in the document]
GleasonK: I had to flip the `preloadDialects` flag in the stablehlo opt tool to get test cases to work without these explicit loads:

```
  return failed(mlir::MlirOptMain(argc, argv, "StableHLO optimizer driver\n",
                                  registry, /*preloadDialectsInContext=*/true));
```

stablehlo/dialect/VhloAttrs.td
```diff
@@ -203,4 +166,55 @@ def VHLO_ConvolutionAttributes {
...
+  let assemblyFormat = "`<` custom<AttributeArray>($value) `>`";
+}
+
+def VHLO_DenseIntOrFPElements : VHLO_AttrDef<"DenseIntOrFPElementsV1"> {
+  let mnemonic = "dense";
+  let parameters = (ins "::mlir::Type":$type, "ArrayRef<char>":$raw_data);
```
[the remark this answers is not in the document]
GleasonK: Changed to:

```
def VHLO_DenseIntOrFPElementsData : TypeParameter<"::llvm::ArrayRef<char>", "Array of int"> {
  // Custom allocator to copy dense elements data into MLIR Context
  let allocator = "$_dst = $_allocator.copyInto($_self);";
}
def VHLO_DenseIntOrFPElementsAttrV1 : VHLO_AttrDef<"DenseIntOrFPElementsV1"> {
  let mnemonic = "dense";
  let parameters = (ins "::mlir::Type":$type, VHLO_DenseIntOrFPElementsData:$raw_data);
  let hasCustomAssemblyFormat = 1;
}
```

Curious what impact this will have on performance. Might not be too important since this is a serialization dialect.

stablehlo/dialect/VhloBytecode.cpp
GleasonK: I like this idea, will try it out.

stablehlo/dialect/VhloTypes.td
GleasonK: The resulting tablegen files aren't happy with "double", I couldn't find any TD files for attrs/types that explicitly use the double type either, end up with a build warning on hashing, happy to discuss offline we want to investigate further:

```
/llvm-project/llvm/include/llvm/ADT/Hashing.h:382:10: error: no matching function for call to 'hash_value'
  return hash_value(value);
         ^~~~~~~~~~
/llvm-project/llvm/include/llvm/ADT/Hashing.h:563:63: note: in instantiation of function template specialization 'llvm::hashing::detail::get_hashable_data<double>' requested here
    buffer_ptr = combine_data(length, buffer_ptr, buffer_end, get_hashable_data(arg));
                                                              ^
/llvm-project/llvm/include/llvm/ADT/Hashing.h:566:12: note: in instantiation of function template specialization 'llvm::hashing::detail::hash_combine_recursive_helper::combine<double, long, long, long>' requested here
    return combine(length, buffer_ptr, buffer_end, args...);
           ^
```

GleasonK: Looks like UniformQuantizedType gets around this by specifying its own hash function and casting away from double:
```
    unsigned getHashValue() const {
      int64_t scaleBits = llvm::bit_cast<int64_t>(scale);
      return llvm::hash_combine(flags, storageType, expressedType, scaleBits,
                                zeroPoint, storageTypeMin, storageTypeMax);
    }
```

stablehlo/dialect/VhloBytecode.cpp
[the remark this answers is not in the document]
GleasonK: Change of plan - this is all going to move into VHLO op verifiers so we don't need assertions. I'm going to leave this ticket open until I implement that as a reminder to clean up this function.

stablehlo/dialect/VhloTypes.td
```diff
@@ -60,9 +63,92 @@ class VHLO_TypeDef<string name,
...
+      llvm::SmallVector<unsigned> validWidths{4, 8, 16, 32, 64};
+      bool isPred = (value.getWidth() == 1 && value.isSignless());
+      if (!isPred && !llvm::is_contained(validWidths, value.getWidth())) {
+        // return emitError() << "invalid integer width " << value;
+      }
+      //if (!value.isSignless() && !value.isUnsigned()) {
+      //  return emitError() << "invalid integer signedness " << value;
+      //}
+      return success();
+    }
```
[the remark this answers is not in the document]
GleasonK: The alternative is to make separate types for `!vhlo.i32`, `!vhlo.ui32`, ... etc (11 type defs, 5signed, 5 unsigned, 1 pred) and move this logic into the type converter.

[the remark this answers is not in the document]
GleasonK: Uncommented this verifier, this was commented out from early dev, should be re-enabled.

stablehlo/dialect/VhloBytecode.cpp
GleasonK: Still working on this, but I don't think it blocks re-review of the other bits. I'm thinking of how to make this elegant.

stablehlo/dialect/VhloAttrs.td
```diff
@@ -204,4 +168,60 @@ def VHLO_ConvolutionAttributes {
...
+//===----------------------------------------------------------------------===//
+
+def VHLO_ArrayAttrV1 : VHLO_AttrDef<"ArrayV1"> {
+  let mnemonic = "array";
+  let parameters = (ins ArrayRefParameter<"mlir::Attribute">:$value);
+  let assemblyFormat = "`<` custom<AttributeArray>($value) `>`";
+}
+
+
+def VHLO_DenseIntOrFPElementsData : TypeParameter<"::llvm::ArrayRef<char>", "Array of int"> {
```
burmako, line 182: `AttrParameter` perhaps?

stablehlo/dialect/VhloBytecode.cpp
```diff
@@ -663,39 +889,218 @@ void VhloBytecodeInterface::write(ScatterDimensionNumbersAttr attr,
...
   writer.writeSignedVarInts(attr.getBounds());
 }
 
+//===----------------------------------------------------------------------===//
+// Forked Attributes
+//===----------------------------------------------------------------------===//
+
+namespace {
+template <typename AttrOrType>
+void assertFromVhlo(AttrOrType val) {
```
burmako, line 931: Since we aren't asserting in this function anymore, let's rename it to something else to avoid giving a wrong impression. E.g. `verifyFromVhlo` or `checkFromVhlo`.

stablehlo/dialect/VhloBytecode.cpp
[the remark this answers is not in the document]
burmako: "I'm going to leave this ticket open". Do you mean leaving this conversation unresolved? Or you're planning to open a ticket for this?

stablehlo/transforms/TypeConversion.h
burmako: I don't think this has been addressed.

stablehlo/dialect/VhloOps.cpp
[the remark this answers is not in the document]
burmako: Should we perhaps add these two dialects (as well as `tensor::TensorDialect` which is loaded in `StablehloDialect`) to registerAllDialects instead? Will we need `preloadDialectsInContext` then?

stablehlo/tests/vhlo_to_version_attrtype_invalid.mlir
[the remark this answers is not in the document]
burmako: How about 2, and then if the default value changes then we bump the version? (From what I understand, this won't need any changes to already existing code, just a mental note to make for the future?).

stablehlo/dialect/VhloTypes.td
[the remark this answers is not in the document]
burmako: Having 11 separate types sounds the best, given that their definitions will be oneliners.

stablehlo/dialect/VhloTypes.td
burmako: Would be nice to get rid of `VHLO_DoubleV1` if we can pull this off.

stablehlo/dialect/VhloTypes.td
```diff
@@ -61,9 +64,92 @@ class VHLO_TypeDef<string name,
...
+  }];
+}
+
+def VHLO_RankedTensorV1 : VHLO_TypeDef<"RankedTensorV1", "tensor"> {
+  let parameters = (ins
+    VHLO_Dims:$shape,
+    "::mlir::Type":$elementType,
+    "::mlir::Attribute":$encoding
+  );
+  let assemblyFormat = "`<` custom<TensorShape>($shape) `` $elementType `` custom<Encoding>($encoding) `>`";
```
burmako, line 114: Perhaps just `Shape` (instead of `TensorShape`), given that the field is called `shape`?

burmako (changes requested):

stablehlo/dialect/VhloBytecode.cpp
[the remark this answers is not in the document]
GleasonK: *Leave this conversation unresolved until it is fixed.

Will likely be back to addressing feedback on this tomorrow.

stablehlo/dialect/VhloTypes.td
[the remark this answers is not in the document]
GleasonK: Sounds good. That is 11 types, 11 type conversions, 11 bytecode constants. But this is still probably a more sound design for serialization.

stablehlo/dialect/VhloOps.cpp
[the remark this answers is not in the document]
GleasonK: Figured it out - These dialects need to be declared as dependent in passes that may construct objects from the dialect, namely `VHLO -> StableHLO`. Declared as dependent dialect and all working out now. Probably should include these along with `TensorDialect` in the `registerAllDialects` though? Now sure what happens if a pass depends on a dialect that isn't registered

stablehlo/transforms/TypeConversion.h
GleasonK: Fixed it in the StableHLO->VHLO conversion, missed it in the VHLO->StableHLO. Fixed in both now.

stablehlo/dialect/VhloBytecode.cpp
GleasonK: Changed to `verifyFromVhlo`. This will still be needed for bytecode read, a verifier on the VHLO ops should cover the VHLO write pieces.

stablehlo/dialect/VhloTypes.td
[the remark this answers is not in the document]
GleasonK: Done, the current setup has type conversion in the bytecode reader since we wrap IntegerAttr, but it should never fail. Curious on your thoughts

stablehlo/dialect/VhloTypes.td
GleasonK: Chose the name from spec.md ebnf:

```
TensorType    ::= 'tensor' '<' TensorShape ElementType '>'
TensorShape   ::= {DimensionSize 'x'}
DimensionSize ::= digit {digit}
```

stablehlo/dialect/VhloTypes.td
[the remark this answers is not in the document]
GleasonK: Refactored again. Decided against wrapping any upstream types. This removes the need for type conversions during bytecode reading/writing. Marking this as done.

stablehlo/tests/vhlo_to_version_attrtype_invalid.mlir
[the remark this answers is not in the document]
GleasonK: Sounds good, it may be a bit tricky, but can be done when needed. I don't think any of these default values are likely to change any time soon.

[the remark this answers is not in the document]
GleasonK: Modified test case to only have the following attrs:

```
    backend_config = #vhlo.string<"">,
    call_target_name = #vhlo.string<"foo">,
    called_computations = #vhlo.array<[#vhlo.sym<dense<1> : tensor<i16>>]>
```

These are default, not optional, in StableHLO and therefore required in VHLO.

stablehlo/dialect/VhloTypes.td
GleasonK: Do you have numeric concerns? I'm not entirely sure of the differences in APFloat-of-double vs double.

This type is also used by FloatAttr so that we can get printing and parsing without needing to implement a custom printer/parser in `VhloOps.cpp` for APFloat types.

stablehlo/dialect/VhloBytecode.cpp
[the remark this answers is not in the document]
GleasonK: Added verifiers to all VHLO attrs/types that hold nested attrs/types

stablehlo/dialect/VhloTypes.td
burmako: I think we should use APFloat because it also carries float semantics - something that would be good to preserve. Also, I've just noticed that upstream FloatAttr is using APFloat: https://github.com/llvm/llvm-project/blob/62bd944e42472b7aa01fe6f662d848d76a96b8f8/mlir/include/mlir/IR/BuiltinAttributes.td#L617-L618.

stablehlo/dialect/VhloTypes.td
GleasonK: Changed to `printShape`

stablehlo/dialect/VhloAttrs.td
```diff
@@ -204,4 +168,99 @@ def VHLO_ConvolutionAttributes {
...
+}
+
+def VHLO_FloatAttrV1 : VHLO_AttrDef<"FloatV1"> {
+  let mnemonic = "float";
+  let parameters = (ins "mlir::Type":$type, VHLO_DoubleV1:$value);
+  let genVerifyDecl = 1;
+  let extraClassDefinition = [{
+    LogicalResult FloatV1Attr::verify(
+        llvm::function_ref<mlir::InFlightDiagnostic ()> errFn, mlir::Type type, APFloat value) {
+      if (!isFromVhlo(type)) return errFn() << "expected VHLO float type";
```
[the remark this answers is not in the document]
GleasonK: Sounds good. Another issue I'm having is that these `errFn` produce multiple line error messages, and I'm unable to get `-verify-diagnostics` to be happy with multi-line messages. These all work, but verifiers are untested currently. I have a separate file with test cases.

stablehlo/dialect/VhloTypes.td
GleasonK: This currently uses APFloat, I renamed the type to VHLO_APFloatV1 accordingly.

stablehlo/dialect/VhloAttrs.td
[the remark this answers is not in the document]
GleasonK: Updated the error message as suggested.

Per my follow-up question, these test points were deleted in this PR but added back as legalization tests (not verifier tests) in https://github.com/openxla/stablehlo/pull/925/files#diff-a0dbcff346cd4a84a6267d153cdf699aa68d7672270b80ec0a7aadb21debe099

This was done because that PR adds FP8, so we are able to test that `0.3.0` fails on downgrade when that type is used.

GleasonK:
There was a cyclic dependency between type conversion and VhloOps, since VHLO pretty printing relies on conversion. We could fork all upstream printers, but that is excessive and potentially error-prone, since the numeric printing and the ElementsAttr printing are fairly intricate.

This issue cannot be avoided by wrapping types either, it would just push the dependency into validation instead of printing, we would need to convert wrapped type to VHLO type to confirm proper version - i.e. if we just wrapped `FloatAttr<1 : f8E5M2>`, we need to convert the f8 during validation to ensure that the type is valid in the target version.

burmako (changes requested):
Thank you for patiently working out the VHLO architecture! This looks like a very hairy problem - both the design and the implementation.

Even though it's a lot of code, I'm quite pleased with how things are going. Our high-level design is fairly approachable - we've got a dedicated dialect, three crisp passes that operate on it and the rest is implementation details. E.g. if needed, each pass can be rewritten from scratch with little effect on the rest of the system.

When I think about the VHLO dialect aka our serialization schema, I am also happy with how things are turning out - there are some rough edges around ops and attrs, but there's a clear evolution path towards addressing the irregularities, captured in the tickets in our issue tracker.

BUILD.bazel
```diff
@@ -360,6 +360,7 @@ cc_library(
     hdrs = [
         "stablehlo/dialect/VhloBytecode.h",
         "stablehlo/dialect/VhloOps.h",
+        "stablehlo/transforms/VhloBuiltinTypeConversion.h",
```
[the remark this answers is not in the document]
GleasonK: This is the cyclic dependency I struggled to resolve. Conversions requires VHLO ops and if we want to use the pretty printers, then ops requires conversions.

stablehlo/transforms/VhloBuiltinTypeConversion.h
```diff
@@ -0,0 +1,231 @@
+/* Copyright 2023 The StableHLO Authors.
```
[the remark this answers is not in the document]
GleasonK: Two functions and a virtual method `convertEncoding`. Modified and moved impls to cpp file

stablehlo/transforms/TypeConversion.h
```diff
@@ -20,107 +20,89 @@ limitations under the License.
 #include "mlir/Transforms/DialectConversion.h"
```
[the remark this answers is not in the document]
GleasonK: Split type converters. There is still a `TypeConversion.h` for the register function type conversions, but that will disappear with the `vhlo::FuncOp` PR.

BUILD.bazel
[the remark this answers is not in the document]
GleasonK: Refactored into separate Types target. Since we only need the converter for attribute definition, we are able to make this separation.

stablehlo/dialect/VhloTypes.h
```diff
@@ -0,0 +1,67 @@
...
+#define STABLEHLO_TRANSFORMS_VHLOTYPES_H
+
+#include "mlir/Transforms/DialectConversion.h"
+#include "stablehlo/dialect/Version.h"
+#include "stablehlo/dialect/VhloDialect.h"
+
+namespace mlir {
+namespace vhlo {
+
+class VersionedTypeConverterBase : public TypeConverter {
```
burmako: Wdyt about renaming VersionedTypeConverterBase to VhloTypeConverterBase? The original name is probably a remnant from the times when the dialect was called VersionedHLO?

stablehlo/dialect/VhloTypes.h
```diff
@@ -0,0 +1,67 @@
...
+
+// This class is used to manage conversions between VHLO and Builtin
+// dialects.
+class VhloTypeConverter : public VersionedTypeConverterBase {
+ public:
+  VhloTypeConverter() : VersionedTypeConverterBase() {}
+
+  virtual ~VhloTypeConverter() = default;
+
+  virtual Attribute convertEncoding(Attribute attr) override = 0;
```
burmako, line 43: Do we need to declare this method here given that it's inherited from the base class?

stablehlo/dialect/VhloDialect.h
```diff
@@ -0,0 +1,54 @@
...
+limitations under the License.
+==============================================================================*/
+
+#ifndef STABLEHLO_DIALECT_VHLO_DIALECT_H
+#define STABLEHLO_DIALECT_VHLO_DIALECT_H
+
+#include "llvm/ADT/StringRef.h"
+#include "mlir/IR/Attributes.h"
+#include "mlir/IR/Dialect.h"
+#include "mlir/IR/DialectImplementation.h"
```
burmako, line 22: Looks like we'll also want `#include "stablehlo/dialect/Version.h"` here? And maybe `"mlir/IR/Type.h"` as well?

stablehlo/dialect/VhloOps.h
```diff
@@ -18,53 +18,13 @@ limitations under the License.
...
 #include "mlir/IR/Attributes.h"
-#include "mlir/IR/Builders.h"
-#include "mlir/IR/BuiltinAttributes.h"
-#include "mlir/IR/BuiltinTypes.h"
 #include "mlir/IR/Dialect.h"
 #include "mlir/IR/DialectImplementation.h"
-#include "mlir/IR/Location.h"
 #include "mlir/IR/MLIRContext.h"
-#include "mlir/IR/OpDefinition.h"
 #include "stablehlo/dialect/Version.h"
```
burmako, line 25: We can probably clean up some or maybe even all of these includes from this file, given that they all seem to come from included files anyway?

stablehlo/dialect/VhloTypes.cpp
```diff
@@ -0,0 +1,188 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#include "stablehlo/dialect/VhloTypes.h"
+
+#include "llvm/Support/Debug.h"
```
burmako, line 18: Is this used anywhere in this file?

stablehlo/dialect/CMakeLists.txt
```diff
@@ -174,19 +174,37 @@ mlir_tablegen(VhloTypeDefs.cpp.inc -gen-typedef-defs --typedefs-dialect=vhlo)
...
+  LINK_LIBS PUBLIC
+  VhloVersion
+  MLIRIR
+  MLIRSupport
+  MLIRTransformUtils
+)
+
 add_mlir_dialect_library(VhloOps
   PARTIAL_SOURCES_INTENDED
+  Version.cpp
```
burmako, line 199: Is it intended that Version.cpp is included in two different targets?

stablehlo/dialect/CMakeLists.txt
```diff
@@ -174,19 +174,37 @@ mlir_tablegen(VhloTypeDefs.cpp.inc -gen-typedef-defs --typedefs-dialect=vhlo)
...
   DEPENDS
   VhloOpsIncGen
 
   LINK_LIBS PUBLIC
   StablehloAssemblyFormat
-  StablehloBase
-  MLIRIR
   MLIRSupport
-  MLIRQuantDialect
+  MLIRIR
```
burmako, line 209: Non-trivial reshufflings of our CMake build always make me nervous given that we experienced downstream breakages in the past. In the next integrate, let's take extra care to make sure that we don't break downstream projects. Nothing actionable for this PR though.

stablehlo/dialect/VhloOps.cpp
```diff
@@ -16,12 +16,140 @@ limitations under the License.
...
+      parser.getContext(), convertTypeToVhloForParse(attr.getType()),
+      attr.getRawData());
+}
+
+}  // namespace vhlo
+}  // namespace mlir
+
 // Include order matters
 #include "stablehlo/dialect/VhloTypeInterfaces.cpp.inc"
 #define GET_TYPEDEF_CLASSES
```
burmako, line 155: Should we move these into VhloTypes.cpp as well?

stablehlo/dialect/CMakeLists.txt
```diff
@@ -174,19 +174,37 @@ mlir_tablegen(VhloTypeDefs.cpp.inc -gen-typedef-defs --typedefs-dialect=vhlo)
 add_public_tablegen_target(VhloOpsIncGen)
 add_dependencies(mlir-headers VhloOpsIncGen)
 
+
+add_mlir_dialect_library(VhloVersion
+  PARTIAL_SOURCES_INTENDED
+  Version.cpp
```
burmako, line 180: Seems to be using `MLIRIR` and `MLIRSupport`?

stablehlo/dialect/CMakeLists.txt
```diff
@@ -174,19 +174,37 @@ mlir_tablegen(VhloTypeDefs.cpp.inc -gen-typedef-defs --typedefs-dialect=vhlo)
...
+  PARTIAL_SOURCES_INTENDED
+  VhloTypes.cpp
+
+  DEPENDS
+  VhloOpsIncGen
+
+  LINK_LIBS PUBLIC
+  VhloVersion
+  MLIRIR
+  MLIRSupport
```
burmako, line 193: I don't think `MLIRSupport` is used but `MLIRQuantDialect` is.

stablehlo/dialect/CMakeLists.txt
burmako, line 209: Also LINK_LIBS is no longer ordered alphabetically it would seem.

burmako, line 209: Let's also add `MLIRShapeDialect`?

BUILD.bazel
```diff
@@ -351,6 +351,39 @@ gentbl_cc_library(
     ],
 )
 
+cc_library(
```
burmako, line 354: How about ordering the new targets alphabetically?

stablehlo/dialect/VhloDialect.h
```diff
@@ -0,0 +1,54 @@
...
+
+namespace mlir {
+namespace vhlo {
+
+class VhloDialect : public Dialect {
```
burmako, line 27: Ah now I see what you meant when you asked offline what I think about creating VhloDialect.h.

What seems unfortunate about the proposed arrangement is that we don't have VhloDialect.cpp (which we cannot have because we'll be back to circular dependencies), so depending on the `vhlo_dialect` target alone is insufficient for the program to link correctly. Let's try to avoid that.

It looks like the only dependency from `VhloTypes.h` to `VhloDialect.h` is `VhloDialect::getCurrentVersion()`. Given that, how do you feel about moving `getCurrentVersion` (and `getMinimumVersion`) into `VhloVersion.h` instead of creating this file and this target?

This is not perfect from the purity perspective (I remember we discussed this some time ago), but it seems preferable to having the `vhlo_dialect` target as it stands right now.

burmako (changes requested):

stablehlo/dialect/VhloDialect.h
GleasonK: file deleted

stablehlo/dialect/VhloOps.cpp
GleasonK: Let's think about that later. I added a note to the Vhlo sync to spec issue. This cpp is required by things later in this file.

stablehlo/dialect/CMakeLists.txt
GleasonK: Shape dialect as well for witness

stablehlo/dialect/VhloOps.h
GleasonK: Removed several, with Dialect back in this file I think the list is good. Don't want to rely too much on headers including other headers for us

stablehlo/dialect/VhloTypes.h
```diff
@@ -0,0 +1,62 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#ifndef STABLEHLO_TRANSFORMS_VHLOTYPES_H
+#define STABLEHLO_TRANSFORMS_VHLOTYPES_H
```
burmako, line 17: `STABLEHLO_DIALECT_...`.

stablehlo/dialect/VhloOps.cpp
burmako: I think this is important because keeping these `.cpp.inc` files here will make the `vhlo_types` target non self-contained. To address the dependency from this file to these definitions, can we include `VhloTypes.h` from here?

burmako (approved):

stablehlo/dialect/VhloOps.cpp
burmako: We agreed to fix this early next week.

PR #853, opened 2022-12-29
Move inferFftOp and verifyRngOp with Enums

zhouxin913:
Note in `inferFftOp()` this old message
```
return emitOptionalError(
    location, stringifyFftType(fftType),
    " takes a complex tensor as input, but is given ", operandType, ".");
}
```
is replaced by
```
return emitOptionalError(
    location, "FFT/IFFT/IRFFT",
    " takes a complex tensor as input, but is given ", operandType, ".");
)
```
To avoid passing the arg FFType `fftType` to `inferFftOp()`. No tests changes for this because the test message happens to be "IRFFT takes a complex tensor as input".

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -1788,10 +1788,10 @@ LogicalResult inferDynamicUpdateSliceOp(
 // P2. Element types agree with fft_type
 // P3. Operand shape dimensions agree with fft_length for the given fft_type
 LogicalResult inferFftOp(
-    Optional<Location> location, Value operand, bool isRfft, bool isIrfft,
-    DenseIntElementsAttr fftLength,
+    Optional<Location> location, Value operand, bool isFftTypeRfft,
+    bool isFftTypeIrfft, DenseIntElementsAttr fftLength,
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
-  auto fftLengthes = fftLength.getValues<int64_t>();
+  auto fftLengthValues = fftLength.getValues<int64_t>();
   int64_t fftRank = fftLength.size();
 
   // P1.
@@ -1807,7 +1807,7 @@ LogicalResult inferFftOp(
   auto operandType = operand.getType().cast<TensorType>();
   Type operandElementType = operandType.getElementType();
   // Check the input element type and infer return element type
-  if (isRfft) {
+  if (isFftTypeRfft) {
     if (!operandElementType.isF32() && !operandElementType.isF64()) {
       return emitOptionalError(
           location, "RFFT requires f32 or f64 input type, but is given ",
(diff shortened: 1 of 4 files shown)


stablehlo/dialect/TypeInference.h
```diff
@@ -129,6 +129,11 @@ LogicalResult inferDynamicUpdateSliceOp(
     ValueRange startIndices,
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes);
 
+LogicalResult inferFftOp(
+    Optional<Location> location, Value operand, bool isRfft, bool isIrfft,
```
burmako, line 133: How about `isFftTypeRfft` and `isFftTypeIrfft`? That would be consistent with `isTransposeAInvalid` and `isRngDistributionUniform` in that it would also call out the name of the attribute, not just its value.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1783,6 +1783,105 @@ LogicalResult inferDynamicUpdateSliceOp(
...
+    if (!operandElementType.isF32() && !operandElementType.isF64()) {
+      return emitOptionalError(
+          location, "RFFT requires f32 or f64 input type, but is given ",
+          operandElementType, ".");
+    }
+  } else {
+    if (!operandElementType.isa<ComplexType>())
+      return emitOptionalError(
+          location, "FFT/IFFT/IRFFT",
+          " takes a complex tensor as input, but is given ", operandType, ".");
```
burmako, line 1820: Typo: "take".

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1783,6 +1783,105 @@ LogicalResult inferDynamicUpdateSliceOp(
...
+  if (isRfft) {
+    if (!operandElementType.isF32() && !operandElementType.isF64()) {
+      return emitOptionalError(
+          location, "RFFT requires f32 or f64 input type, but is given ",
+          operandElementType, ".");
+    }
+  } else {
+    if (!operandElementType.isa<ComplexType>())
+      return emitOptionalError(
+          location, "FFT/IFFT/IRFFT",
```
burmako, line 1819: It's neat that technically the test doesn't need updating, but let's still update it to demonstrate the updated error message.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1783,6 +1783,105 @@ LogicalResult inferDynamicUpdateSliceOp(
...
 
+// We intend to verify the following properties
+// P1. 1 <= rank <= 3
+// P2. Element types agree with fft_type
+// P3. Operand shape dimensions agree with fft_length for the given fft_type
+LogicalResult inferFftOp(
+    Optional<Location> location, Value operand, bool isRfft, bool isIrfft,
+    DenseIntElementsAttr fftLength,
+    SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
+  auto fftLengthes = fftLength.getValues<int64_t>();
```
burmako, line 1794: How about `fftLengthValues`? This is more verbose but avoids confusion between two very similarly named variables.

burmako (changes requested):

stablehlo/dialect/TypeInference.cpp
zhouxin913: Agreed.

stablehlo/dialect/TypeInference.cpp
zhouxin913: Makes sense, updated.

stablehlo/dialect/TypeInference.h
zhouxin913: Yeah I hesitated on this but it seems better to be aligned. Updated.

burmako (approved):

Issue #846, opened 2022-12-28
Review the spec for RemainderOp


burmako:
The spec currently says "the remainder is calculated as `lhs - d * rhs`, where `d = stablehlo.divide`", and I'm not sure what this means exactly. Perhaps this was a typo that went unnoticed.

PR #877, opened 2023-01-03
Added missing detail for remainder op

sdasgup3:
For more details please refer to the ticket below.

fixes https://github.com/openxla/stablehlo/issues/846

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -4013,10 +4013,10 @@ produces a `result` tensor.
 
 More formally, the sign of the result is taken from the dividend, and the
 absolute value of the result is always less than the divisor's absolute value.
-The remainder is calculated as `lhs - d * rhs`, where `d = stablehlo.divide`.
-For floating-point element types, this is in contrast with the `remainder`
-operation from IEEE-754 specification where `d` is an integral value nearest to
-the exact value of `lhs/rhs` with ties to even.
+The remainder is calculated as `lhs - d * rhs`, where `d = stablehlo.divide(lhs,
+rhs)`. For floating-point element types, this is in contrast with the
+`remainder` operation from IEEE-754 specification where `d` is an integral value
+nearest to the exact value of `lhs/rhs` with ties to even.
 
 ### Inputs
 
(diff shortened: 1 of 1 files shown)


docs/spec.md
```diff
@@ -4008,10 +4008,10 @@ produces a `result` tensor.
 
 More formally, the sign of the result is taken from the dividend, and the
 absolute value of the result is always less than the divisor's absolute value.
-The remainder is calculated as `lhs - d * rhs`, where `d = stablehlo.divide`.
-For floating-point element types, this is in contrast with the `remainder`
-operation from IEEE-754 specification where `d` is an integral value nearest to
-the exact value of `lhs/rhs` with ties to even.
+The remainder is calculated as `lhs - d * rhs`, where `d = stablehlo.divide(lhs,
```
burmako, line 4011: Do we also need to apply some kind of rounding to `d * rhs`? Otherwise for floating-point values, `lhs - (lhs / rhs) * rhs` will mostly return values that are very close to zero?

burmako (changes requested):

docs/spec.md
sdasgup3: Realized that I completely lost the "with its fractional part truncated part". While I am looking at the old reviews
https://github.com/openxla/stablehlo/pull/106/commits/81ae9d8e0f2222c5e4f707979de23da09aacfec7#r967534834 I can see that it was very much there.

burmako: Is it possible to tie "fractional part truncated" to some IEEE-754 operation? Sorry for the late response!

burmako (changes requested):

docs/spec.md
sdasgup3: No problem!

PTAL of the option I already incorporated in the spec. I liked it better than the following alternative because it is more concise
 
```
...
* For floats: `roundToIntegralTowardZero(lhs/rhs)` from IEEE-754 specification, where `lhs/rhs` is the exact  result of `divison`.
...
```

sdasgup3: Also, IMO we should get rid of the complex type for this op (via a separate issue). Its evaluation not supported in XLA and neither it is produced by JAX.

burmako (approved):

docs/spec.md
burmako: Thank you for the update! LGTM.

Let's also open a ticket for complex remainder.

PR #881, opened 2023-01-05
Fix typos in infer_stablehlo.mlir tests

ghpvnist:
I found a couple of inconsistencies in the tests in regards to what value to return (and lack thereof), so this PR fixes it.


zhouxin913 (approved):
Thank you for making the tests a better world!

PR #883, opened 2023-01-05
Reinstate util functions in TypeInference.h

zhouxin913:
The declaration of these util functions are removed in previous PRs because they are used in `TypeInference.cpp` only. But it turns out that they are used by MHLO as well: so need reinstate them.


burmako (approved):
Can you update the description of the PR to explain the rationale? Also I think the PR title could use an update as well along the lines of "Reinstate helper functions in TypeInference.h".

PR #885, opened 2023-01-05
Handle bounds in the SortOp shape function

### LINKED ISSUE #321: Add verification for SortOp
### Request description

Add verification for following cases of SortOp -
  * Variadic tensor `inputs` should have at least 1 tensor

### LINKED ISSUE #803: Handle bounds in `SortOp`shape function

subhankarshah:
Closes #803

Infer the most specific type from all input tensors and use the most specific type for every result tensor with the element type of the corresponding input tensor.

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -2393,7 +2393,7 @@ LogicalResult inferSliceOp(Optional<Location> location, Value operand,
 LogicalResult inferSortOp(
     Optional<Location> location, ValueRange inputs,
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
-  // Infermost specific type from all operands
+  // Infer most specific type from all operands
   SmallVector<Type> mostSpecificType;
   if (failed(hlo::inferMostSpecificType(location, inputs.getTypes(),
                                         mostSpecificType)))
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -2391,10 +2391,24 @@ LogicalResult inferSliceOp(Optional<Location> location, Value operand,
 }
 
 LogicalResult inferSortOp(
-    Optional<Location>, ValueRange inputs,
+    Optional<Location> location, ValueRange inputs,
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
-  for (auto resultType : inputs.getTypes())
-    inferredReturnShapes.emplace_back(resultType.cast<ShapedType>());
+  // Infermost specific type from all operands
+  SmallVector<Type> mostSpecificType;
+  if (failed(hlo::inferMostSpecificType(location, inputs.getTypes(),
+                                        mostSpecificType)))
+    return failure();
+
+  auto rankedOutput = mostSpecificType.front().dyn_cast<RankedTensorType>();
+  for (auto inputType : inputs.getTypes()) {
+    auto inputShapedType = inputType.cast<ShapedType>();
+    if (rankedOutput)
+      inferredReturnShapes.emplace_back(rankedOutput.getShape(),
+                                        inputShapedType.getElementType(),
+                                        rankedOutput.getEncoding());
+    else
+      inferredReturnShapes.emplace_back(inputShapedType);
+  }
   return success();
 }
 
--- a/stablehlo/tests/infer_stablehlo.mlir
+++ b/stablehlo/tests/infer_stablehlo.mlir
@@ -1228,8 +1228,7 @@ func.func @sort_unknown_rank(%input0: tensor<*xf32>, %input1: tensor<*xi32>) {
 // -----
 
 // CHECK-LABEL: @sort_bounds_and_unknown_rank
-func.func @sort_bounds_and_unknown_rank(%input0: tensor<*xf32>, %input1: tensor<5x?x?xi32, #stablehlo.type_extensions<bounds = [?, 7, 6]>>
-) {
+func.func @sort_bounds_and_unknown_rank(%input0: tensor<*xf32>, %input1: tensor<5x?x?xi32, #stablehlo.type_extensions<bounds = [?, 7, 6]>>) {
   %0, %1 = "stablehlo.sort"(%input0, %input1) ({
   ^bb0(%arg0: tensor<f32>, %arg1: tensor<f32>, %arg2: tensor<i32>, %arg3: tensor<i32>):
     %pred = "stablehlo.compare"(%arg0, %arg1) {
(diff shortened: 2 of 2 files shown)


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2171,12 +2171,12 @@ void SortOp::build(OpBuilder& builder, OperationState& state,
   state.addRegion();
 }
 
-LogicalResult SortOp::inferReturnTypeComponents(
-    MLIRContext*, Optional<Location> location, ValueShapeRange operands,
+LogicalResult SortOp::inferReturnTypes(
```
burmako, line 2174: What is the reason for refactoring to `inferReturnTypes`? I think that `inferReturnTypeComponents` recently added support for encodings as well.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -2390,11 +2390,26 @@ LogicalResult inferSliceOp(Optional<Location> location, Value operand,
...
-    Optional<Location>, ValueRange inputs,
-    SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
-  for (auto resultType : inputs.getTypes())
-    inferredReturnShapes.emplace_back(resultType.cast<ShapedType>());
+LogicalResult inferSortOp(Optional<Location> location, ValueRange inputs,
+                          SmallVectorImpl<Type>& inferredReturnTypes) {
+  // Infermost specific type from all operands
+  SmallVector<Type> mostSpecificType;
+  if (failed(hlo::inferMostSpecificType(location, inputs.getTypes(),
+                                        mostSpecificType))) {
```
burmako, line 2398: Sounds like we would also want an overload for `inferMostSpecificType` which takes `Type&`.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -2390,11 +2390,26 @@ LogicalResult inferSliceOp(Optional<Location> location, Value operand,
...
+  if (!rankedOutput) {
+    for (auto input : inputs.getTypes())
+      inferredReturnTypes.emplace_back(input);
+    return success();
+  }
+
+  for (auto input : inputs.getTypes())
+    inferredReturnTypes.emplace_back(RankedTensorType::get(
+        rankedOutput.getShape(), input.dyn_cast<TensorType>().getElementType(),
+        rankedOutput.getEncoding()));
```
burmako, line 2412: This is a non-trivial change, and we'll need to document why we're making it. PR description might be a good place for that. On an unrelated note, the description currently links to a wrong ticket.

burmako (changes requested):
Added a few high-level comments, didn't review the code in details yet - will leave that to the first pass by Xin.

stablehlo/dialect/StablehloOps.cpp
zhouxin913: +1. The "recently added support" Eugene mentioned is here: https://github.com/llvm/llvm-project/commit/b2505ca2ece3
In short, in old days the PRs like this have to change to use `inferReturnTypes` but today it is not necessary any more

zhouxin913 (changes requested):

subhankarshah:
@burmako As to why we are inferring the most specific type, I am not sure if it is an overkill or we can go with the most specific type, I have added tests for the same, let me know if we should it simple.

zhouxin913 (commented):
There are 2 ways of addressing this multiple-inputs op, SortOp:
1. Simply copying over tensors with bounds from inputs to inferred results
2. As all the inputs care compatible, we can also have (like what current commit does) inferredType[i] = inferMostSpecificType(all_inputs) for each tensor in result.

@smit-hinsu Are you expecting (2)? If so, we will go with it. If not, we will have a decision on our own on situations like this.

smit-hinsu:
We don't need to go with 2 and can just copy over the input shapes to begin with. That means just adding a test in this PR.

IIRC, we did 1 for the concat op and 2 for the binary ops. Note that this choice is not limited just to bounded shapes and applies to unbounded shapes as well. Like, with inputs tensor<2x?xf32> and tensor<?x3xf32>, we can say infer the result as tensor<2x3xf32>.

zhouxin913:
> We don't need to go with 2 and can just copy over the input shapes to begin with. That means just adding a test in this PR.
> 
> IIRC, we did 1 for the concat op and 2 for the binary ops. Note that this choice is not limited just to bounded shapes and applies to unbounded shapes as well. Like, with inputs tensor<2x?xf32> and tensor<?x3xf32>, we infer the result as tensor<2x3xf32>.

Thanks @smit-hinsu for confirmation. We did go with 2 for Concat as well: it's a 
 multiple-inputs op and all the dims are inferred in a hard way https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/TypeInference.cpp#L1565

Could you elaborate more about the reason behind your preference? Do you mean that for all those op with mulitple-input-multiple-output, you always prefer the simple way 1? 
Otherwise, if it's case by case, then we will need agree on collaboration pattern: for example, discuss with you with a draft PR with description, and then the author actually dives into implementation. What do you think of this?

smit-hinsu:
yes, you are right! Seems like we went with 2 for Concat in the final version.

I don't have any preferences between these two approaches. Both are correct and enough for our immediate use-cases given XLA only has static shapes. The second approach provides a stricter result shape. We could choose to go with the first approach if that helps implementation and punt on going to second option for second iteration or doing it later on. We could also choose to go with the first approach for all ops for consistency. 

I will leave this to you to decide based on how much effort each options require and your other priorities.

zhouxin913 (approved):
Even through we can go with the simple way without specific inference, the current implementation is still very valuable and worth checkin. And no future issue needs to be created.

zhouxin913 (commented):
LGTM

stablehlo/dialect/TypeInference.cpp
```diff
@@ -2391,10 +2391,24 @@ LogicalResult inferSliceOp(Optional<Location> location, Value operand,
...
+  SmallVector<Type> mostSpecificType;
+  if (failed(hlo::inferMostSpecificType(location, inputs.getTypes(),
+                                        mostSpecificType)))
+    return failure();
+
+  auto rankedOutput = mostSpecificType.front().dyn_cast<RankedTensorType>();
+  for (auto inputType : inputs.getTypes()) {
+    auto inputShapedType = inputType.cast<ShapedType>();
+    if (rankedOutput)
+      inferredReturnShapes.emplace_back(rankedOutput.getShape(),
```
smit-hinsu, line 2406: add a comment mentioning that we preserve the original type and only use shape/bounds from the inferred result.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -2391,10 +2391,24 @@ LogicalResult inferSliceOp(Optional<Location> location, Value operand,
 }
 
 LogicalResult inferSortOp(
-    Optional<Location>, ValueRange inputs,
+    Optional<Location> location, ValueRange inputs,
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
-  for (auto resultType : inputs.getTypes())
-    inferredReturnShapes.emplace_back(resultType.cast<ShapedType>());
```
burmako, line 2397: Given the offline discussion, let's update the PR to just emplace_back the components of `resultType`. Unfortunately `emplace_back(resultType)` won't work because it doesn't account for encoding, so we'll need to explicitly call out the three components.

burmako, line 2397: Can you also update the PR description to provide the updated rationale?

burmako (changes requested):

PR #886, opened 2023-01-05
Integrate small changes of gather and dynamicslice

zhouxin913:
There 2 small fixes aiming to fill in the gap against MHLO in `TypeInference.cpp`.
The diverge of the dynamic issue in GatherOp come from https://github.com/tensorflow/mlir-hlo/commit/77f83757df60fccfe1c70fb847040dc9dd618350

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -65,6 +65,8 @@ namespace hlo {
 // Utils for shape functions.
 //===----------------------------------------------------------------------===//
 
+static constexpr int64_t kDynamicSizePrintValue = -1;
+
 // Checks if the vector `nums` has duplicates.
 const auto hasDuplicates = [](const ArrayRef<int64_t> nums) {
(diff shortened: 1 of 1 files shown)


stablehlo/dialect/TypeInference.cpp
```diff
@@ -65,6 +65,8 @@ namespace hlo {
 // Utils for shape functions.
 //===----------------------------------------------------------------------===//
 
+static constexpr int64_t kDynamicSizePrintValue = -1;
```
burmako, line 68: In the PR description, can you explain the rationale behind these changes? Also, why does it matter to type inference how dynamic sizes are printed? Aren't they printed as question marks anyway?

burmako (changes requested):

burmako:
We've decided that this shouldn't be part of TypeInference.cpp, so we'll fix this on the MHLO side by removing the logic in the TF/XLA bridge that relies on this behavior.

PR #887, opened 2023-01-06
Handle bounds in the CholeskyOp shape function

### LINKED ISSUE #804: Handle bounds in `CholeskyOp`shape function

ghpvnist:
Cholesky op's [constraint (C3)](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#cholesky) requires the last two dimensions to be the same. The bound rules are as follows (3 being an arbitrary static shape):
```
dynamic type inference rules for the last two dims of A (case0 is sanity
check for dynamic batch dims):
       dim R-2 | dim R-1 | inferred R-2 | inferred R-1
case0: 3       | 3       | 3            | 3
case1: ?       | 3       | 3            | 3
case2: ?       | ?       | ?            | ?

dynamic bound infererence rules for the last two dims of A (case0 is
sanity check for dynamic batch dims):
       dim R-2 | dim R-1 | inferred R-2 | inferred R-1
case0: 3, ?    | 3, ?    | 3, ?         | 3, ?
case1: ?, ?    | 3, ?    | 3, ?         | 3, ?
case2: ?, A<3  | 3, ?    | error        | error
case3: ?, A>=3 | 3, ?    | 3, ?         | 3, ?
case4: ?, ?    | ?, ?    | ?, ?         | ?, ?
case5: ?, A    | ?, ?    | ?, A         | ?, A
case6: ?, A    | ?, B    | ?, min(A,B)  | ?, min(A,B)
```

The rules proposed above are reviewed and agreed upon, but there will be no immediate implementation in the current PR to follow recently updated guidelines.

closes #804

### DIFF
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -547,12 +547,12 @@ void CustomCallOp::getEffects(
 // CholeskyOp
 //===----------------------------------------------------------------------===//
 
-LogicalResult CholeskyOp::inferReturnTypes(
-    MLIRContext*, Optional<Location> location, ValueRange operands,
+LogicalResult CholeskyOp::inferReturnTypeComponents(
+    MLIRContext*, Optional<Location> location, ValueShapeRange operands,
     DictionaryAttr attributes, RegionRange regions,
-    SmallVectorImpl<Type>& inferredReturnTypes) {
+    SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
   CholeskyOp::Adaptor adaptor(operands, attributes, regions);
-  return hlo::inferCholeskyOp(location, adaptor.getA(), inferredReturnTypes);
+  return hlo::inferCholeskyOp(location, adaptor.getA(), inferredReturnShapes);
 }
 
 //===----------------------------------------------------------------------===//
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -547,12 +547,12 @@ void CustomCallOp::getEffects(
 // CholeskyOp
 //===----------------------------------------------------------------------===//
 
-LogicalResult CholeskyOp::inferReturnTypeComponents(
-    MLIRContext*, Optional<Location> location, ValueShapeRange operands,
+LogicalResult CholeskyOp::inferReturnTypes(
+    MLIRContext*, Optional<Location> location, ValueRange operands,
     DictionaryAttr attributes, RegionRange regions,
-    SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
+    SmallVectorImpl<Type>& inferredReturnTypes) {
   CholeskyOp::Adaptor adaptor(operands, attributes, regions);
-  return hlo::inferCholeskyOp(location, adaptor.getA(), inferredReturnShapes);
+  return hlo::inferCholeskyOp(location, adaptor.getA(), inferredReturnTypes);
 }
 
 //===----------------------------------------------------------------------===//
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -1454,9 +1454,10 @@ LogicalResult inferCholeskyOp(
   int64_t penultimateDim = aShape[aShape.size() - 2];
   auto resultShape = aRankedType.getShape().vec();
   int64_t staticSize = isStaticDimSize(lastDim) ? lastDim : penultimateDim;
+  int64_t rank = aRankedType.getRank();
   if (isStaticDimSize(lastDim) || isStaticDimSize(penultimateDim)) {
-    resultShape[resultShape.size() - 1] = staticSize;
-    resultShape[resultShape.size() - 2] = staticSize;
+    resultShape[rank - 1] = staticSize;
+    resultShape[rank - 2] = staticSize;
   }
   auto resultBounds = encodingToBounds(aRankedType.getEncoding()).vec();
   if (resultBounds.empty()) {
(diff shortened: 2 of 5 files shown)


ghpvnist:
Cholesky op's [constraint (C3)](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#cholesky) requires the last two dimensions to be the same. The bound rules are as follows:
```
       dim R-2 | dim R-1 | inferred R-2 | inferred R-1
case0: 3       | 3       | 3            | 3
case1: ?       | 3       | 3            | 3
case2: ?, A>=3 | 3       | 3            | 3
case3: 3       | ?       | 3            | 3
case4: 3       | ?, A>=3 | 3            | 3
case5: ?       | ?       | ?            | ?
case6: ?, A    | ?       | ?, A         | ?, A
case7: ?       | ?, A    | ?, A         | ?, A
case8: ?, A    | ?, B    | ?, min(A,B)  | ?, min(A,B)
```

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -547,12 +547,12 @@ void CustomCallOp::getEffects(
 // CholeskyOp
 //===----------------------------------------------------------------------===//
 
-LogicalResult CholeskyOp::inferReturnTypeComponents(
-    MLIRContext*, Optional<Location> location, ValueShapeRange operands,
+LogicalResult CholeskyOp::inferReturnTypes(
```
burmako, line 550: Same question as in #885. What is the reason for refactoring to `inferReturnTypes`? I think that `inferReturnTypeComponents` recently added support for encodings as well.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1484,8 +1482,57 @@ LogicalResult inferCholeskyOp(
...
+                            ? resultBounds[resultBounds.size() - 1]
+                            : resultBounds[resultBounds.size() - 2];
+  if (isLastDimStaticBound && isPenultimateDimStaticBound)
+    staticBound = std::min(resultBounds[resultBounds.size() - 1],
+                           resultBounds[resultBounds.size() - 2]);
+  resultBounds[resultBounds.size() - 1] = staticBound;
+  resultBounds[resultBounds.size() - 2] = staticBound;
+  inferredReturnTypes.push_back(RankedTensorType::get(
+      aRankedType.getShape(), aRankedType.getElementType(),
+      boundsToEncoding(aRankedType.getEncoding(), resultBounds)));
```
burmako, line 1535: This is a lot of code - can we make this simpler? E.g. can we use `inferMostSpecificSomething` here instead?

burmako, line 1535: Also, we'll want to document the rationale behind this change which is currently not obvious to me. E.g. what is the driving use case? At the moment, the PR description just links to the ticket which doesn't have a description provided.

burmako (changes requested):
Similarly to #885, I've added a few high-level comments, but didn't review the code in details yet - will leave that to the first pass by Xin.

stablehlo/dialect/StablehloOps.cpp
ghpvnist: Yes, you're correct. I rewrote this with the assumption that the support wasn't added yet. I've reverted the changes.

stablehlo/dialect/TypeInference.cpp
ghpvnist: Can you clarify what change you are referring to? If you mean the rationale for changing the interface, I've reverted back to using InferTensorType as mentioned in the other comment.

ghpvnist:
I've listed the infer rules for this op, but all of them are taken care of by `inferMostSpecificTypeComponents` function. Are these tests necessary?
CC: @smit-hinsu

stablehlo/dialect/TypeInference.cpp
ghpvnist: Ah okay. I realized that all this code is not needed, and is deleted in the recent commit as they are all related to type inference with dynamism not bounded dynamism. However for the last case8 which does have bounded dynamism, it is also conveniently handled by inferMostSpecificTypeComponents, so this PR ends up being a one line change.

ghpvnist: We can't :( `inferMostSpecificXXX` functions are inferring using variadic types and using information from multiple types to find a single most-specific type, but since this op is dealing with tightening a single type, it doesn't fit our use case. I've actually made some significant changes since yesterday so it requires rereview. This op was confusing at first, but after figuring out that some rules are not bound related, things made much more sense.

ghpvnist: We cannot use any previous utility functions as Cholesky op deals with inference working with its dimensions not between types of multiple operands.

> At the moment, the PR description just links to the ticket which doesn't have a description provided.

I see that in a recent PR #799 that the discussions are happening in the issue not the PR. Since I've started some discussions already in the PR and not the issue, should I copy over and continue the discussion in the ticket? I'll provide early review in the issue itself for future shape functions.

zhouxin913:
> Cholesky op's [constraint (C3)](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#cholesky) requires the last two dimensions to be the same. The bound rules are as follows (3 being an arbitrary static shape):
> 
> ```
>        dim R-2 | dim R-1 | inferred R-2 | inferred R-1
> case0: 3       | 3       | 3            | 3
> case1: ?       | 3       | 3            | 3
> case2: ?, A>=3 | 3       | 3            | 3
> case3: 3       | ?       | 3            | 3
> case4: 3       | ?, A>=3 | 3            | 3
> case5: ?       | ?       | ?            | ?
> case6: ?, A    | ?       | ?, A         | ?, A
> case7: ?       | ?, A    | ?, A         | ?, A
> case8: ?, A    | ?, B    | ?, min(A,B)  | ?, min(A,B)
> ```
> 
> * Update: ignore the above table I realized some of these rules are not bound related and are instead specific to dynamic type inference. Since I've already implemented both, I'll include the rules I used for it in addition to the bound rules, which are updated for clarity, removing commutative rules, and reordering.
> 
> ```
> dynamic type inference rules for the last two dims of A (case0 is sanity
> check for dynamic batch dims):
>        dim R-2 | dim R-1 | inferred R-2 | inferred R-1
> case0: 3       | 3       | 3            | 3
> case1: ?       | 3       | 3            | 3
> case2: ?       | ?       | ?            | ?
> 
> dynamic bound infererence rules for the last two dims of A (case0 is
> sanity check for dynamic batch dims):
>        dim R-2 | dim R-1 | inferred R-2 | inferred R-1
> case0: 3, ?    | 3, ?    | 3, ?         | 3, ?
> case1: ?, ?    | 3, ?    | 3, ?         | 3, ?
> case2: ?, A<3  | 3, ?    | error        | error
> case3: ?, A>=3 | 3, ?    | 3, ?         | 3, ?
> case4: ?, ?    | ?, ?    | ?, ?         | ?, ?
> case5: ?, A    | ?, ?    | ?, A         | ?, A
> case6: ?, A    | ?, B    | ?, min(A,B)  | ?, min(A,B)
> ```

Hi Gunhyun, could you move this long comment and fill it in the PR description with the updated version ? So that other reviewer will review it directly. Thanks!

stablehlo/dialect/TypeInference.cpp
burmako: No strong preference on my side about the location of these discussions. If it's in the issue, that's fine. If it's in a PR, that's fine too. If it's split, that's great as well. As long as we make a note somewhere about the motivation behind the tricky changes :)

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1443,8 +1443,83 @@ LogicalResult inferCholeskyOp(
     return emitOptionalError(
         location, "minor dimensions of 'a' must have equal size, got shape ",
         aShape, ".");
-  inferredReturnShapes.emplace_back(aRankedType.getShape(),
-                                    aRankedType.getElementType());
```
burmako, line 1447: Similarly to #885, I'd like to propose to update the PR to just `emplace_back` the components of `aRankedType`. Unfortunately `emplace_back(aRankedType)` won't work because it doesn't account for encoding, so we'll need to explicitly call out the three components.

burmako, line 1447: Thank you for your work going into detail of all the possible cases for inferring the most specific dims for the result type! This was some meticulous work that was important in influencing the overall guidelines that we've worked out for bounded dynamism in shape functions.

burmako (changes requested):

burmako (approved):

stablehlo/tests/infer_stablehlo.mlir
```diff
@@ -1037,6 +1036,14 @@ func.func @pad(%arg0: tensor<?x48x48x32xf32>) -> tensor<4xindex> {
 
 // -----
 
+// CHECK-LABEL: func @cholesky_bounds
+func.func @cholesky_bounds(%input: tensor<2x?x?xf32, #stablehlo.type_extensions<bounds = [?, 5, ?]>>) -> tensor<2x?x?xindex> {
+  %0 = "stablehlo.cholesky"(%input) { lower = true } : (tensor<2x?x?xf32, #stablehlo.type_extensions<bounds = [?, 5, ?]>>) -> tensor<2x?x?xf32>
```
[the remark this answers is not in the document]
zhouxin913: Both are fine. I learned this (tensor<*xf32>) in from https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/infer_stablehlo.mlir#L664
Personally I prefer this way because 1) it's shorter, 2) we return tensor<*xf32> from the op we are testing, and then in `hlo_test_infer.return_types` it's more clear that the test expect a very concrete type in "type0 = tensor<some_concrete_shape_with_bound>.

zhouxin913 (approved):

stablehlo/tests/infer_stablehlo.mlir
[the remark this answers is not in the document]
ghpvnist: I also noticed that some tests omit `%1 = "hlo_test_infer.return_types"(%0) {...} : (tensor<*xf32>) -> tensor<*xindex>` boilerplate in the CHECK directive, which is also a good idea.

Issue #622, opened 2022-11-25
Enable specializing dynamically-shaped programs to static shapes


burmako:
With StableHLO, it should be possible to take dynamically-shaped programs like this one:

```mlir
func.func @main(
    %arg0: tensor<i32>, 
    %arg1: tensor<?x4xf32>, 
    %arg2: tensor<2x?x4xf32>) -> (tensor<2x?x4xf32>, tensor<2x?x4xf32>) {
  %0 = stablehlo.constant dense<2> : tensor<1xi32>
  %1 = stablehlo.reshape %arg0 : (tensor<i32>) -> tensor<1xi32>
  %2 = stablehlo.constant dense<4> : tensor<1xi32>
  %3 = "stablehlo.concatenate"(%0, %1, %2) {dimension = 0 : i64} : (tensor<1xi32>, tensor<1xi32>, tensor<1xi32>) -> tensor<3xi32>
  %4 = "stablehlo.dynamic_broadcast_in_dim"(%arg1, %3) {broadcast_dimensions = dense<[1, 2]> : tensor<2xi64>} : (tensor<?x4xf32>, tensor<3xi32>) -> tensor<2x?x4xf32>
  %5 = stablehlo.add %4, %arg2 : (tensor<2x?x4xf32>, tensor<2x?x4xf32>) -> tensor<2x?x4xf32>
  func.return %4, %6 : tensor<2x?x4xf32>, tensor<2x?x4xf32>
}
```

And then given sufficient static information (e.g. the fact that `%arg0` is equal to the dynamic dimension sizes of `%arg1` and `%arg2`, and the fact that `%arg0` is equal to `3`), specialize these programs to static shapes:

```mlir
func.func @main(
    %arg1: tensor<3x4xf32>, 
    %arg2: tensor<2x3x4xf32>) -> (tensor<2x3x4xf32>, tensor<2x3x4xf32>) {
  %4 = "stablehlo.broadcast_in_dim"(%arg1) {broadcast_dimensions = dense<[1, 2]> : tensor<2xi64>} : (tensor<3x4xf32>) -> tensor<2x3x4xf32>
  %5 = stablehlo.add %4, %arg2 : (tensor<2x3x4xf32>, tensor<2x3x4xf32>) -> tensor<2x3x4xf32>
  func.return %4, %6 : tensor<2x3x4xf32>, tensor<2x3x4xf32>
}
```

[...]

Issue #851, opened 2022-12-28
Design a mechanism to customize shape inference for CustomCallOp


burmako:
This will be needed to some real-world uses of specializing dynamically-shaped programs to static shapes (see #622 for details).

burmako:
One way of going about this would be to add a shape argument to `custom_call`, similarly to how e.g. `dynamic_iota` and `dynamic_reshape` have shape arguments. There are compatibility implications but that's why we have our compatibility protocols. Overall, I like this idea.

burmako:
This was prototyped in #1169. If practical evaluation works out, I'll write a StableHLO RFC to turn the introduced experimental functionality into a bona fide feature of CustomCallOp.

burmako:
The prototype has been working fairly well within jax2tf. I'm planning to prototype some more functionality requested by JAX folks (#1258, #1344, #1366, #1396), and then wrap this up in an RFC, based on our experimental evaluation.

PR #891, opened 2023-01-06
Consistently use HLO_StaticShapeTensor

burmako:
Some ops in the StableHLO dialect have "load-bearing" result shapes, meaning that these shapes cannot be inferred from op's operands, attributes and regions.

For these ops, the current design philosophy involves one of the two approaches:
  1) The shape is specified statically in the result type, e.g. see
     ReshapeOp.
  2) The shape is passed dynamically as an operand (or can be computed
     indirectly from shape-related operands), e.g. see DynamicReshapeOp.

In the future (#8), we want to have one unified approach in these situations, but we're not there yet, and this PR is about an incremental improvement to the state of the art.

One important piece of the design philosophy for 1) is that these ops need to be restricted to produce HLO_StaticShapeTensor, otherwise it is unclear what their semantics is.

  %1 = "stablehlo.reshape"(%0) : (tensor<2x4xf32>) -> tensor<?x?xf32>

For example, the code snippet above doesn't make sense because we need an actual shape to execute the op - either passed at compile time or passed at runtime - and neither is happening here.

When working on #622, I discovered that there are some ops which fall into 1) but don't have their result types restricted to be static: InfeedOp, IotaOp, RecvOp and RngBitGeneratorOp.

This means that StableHLO producers can currently produce these ops with dynamic result types, which would be unsound. This PR fixes this issue by changing the result types to HLO_StaticShapeTensor or equivalent.

### DIFF
--- a/stablehlo/dialect/Base.td
+++ b/stablehlo/dialect/Base.td
@@ -135,11 +135,6 @@ def HLO_DimensionValue : AnyTypeOf<[Index, HLO_Int]>;
 // Dynamic representation of a shape vector as a tensor.
 def HLO_DimensionTensor : 1DTensorOf<[HLO_DimensionValue]>;
 
-// In general, static shaped tensor constraints should be avoided unless
-// it is for a legacy op which is only correct with static shapes.
-def HLO_StaticShapeTensor : StaticShapeTensorOf<[
-      HLO_Float, HLO_Pred, HLO_Int, HLO_Complex, HLO_QuantizedInt]>;
-
 //===----------------------------------------------------------------------===//
 // HLO combined type definitions.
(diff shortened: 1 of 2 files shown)


sdasgup3 (commented):
Is there a reason we left

custom_call and 
broadcast_in_dim?

In both the case we cannot infer the shape.

burmako:
Good questions! `broadcast_in_dim` already has `let results = (outs HLO_StaticShapeTensor)`, so it's already handled. `custom_call` will be handled in #851.

sdasgup3 (approved):
Thanks!
I was looking at `BroadcastOp` by mistake (instead of BroadcastInDim).

LGTM.

PR #894, opened 2023-01-06
Make CHLO ops inlinable

burmako:
When working on #622, I discovered that functions with CHLO ops aren't getting inlined, which is a miss because there's no reason for these ops to not be inlinable. For comparison, StableHLO and MHLO both implement this interface.

This PR fixes that and adds tests to make sure that things stay this way in the future as well.

### DIFF
--- a/stablehlo/dialect/ChloOps.cpp
+++ b/stablehlo/dialect/ChloOps.cpp
@@ -529,8 +529,7 @@ struct ChloDialectInlinerInterface : public DialectInlinerInterface {
                        BlockAndValueMapping& valueMapping) const final {
     return true;
   }
-  // Operations in StableHLO dialect are always legal to inline since they are
-  // pure.
+  // Operations in CHLO dialect are always legal to inline since they are pure.
   bool isLegalToInline(Operation*, Region*, bool,
                        BlockAndValueMapping&) const final {
     return true;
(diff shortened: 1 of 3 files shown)


stablehlo/dialect/ChloOps.cpp
```diff
@@ -513,6 +514,29 @@ LogicalResult ConstantOp::inferReturnTypes(
...
+    return true;
+  }
+  // We don't have any special restrictions on what can be inlined into
+  // destination regions (e.g. while/conditional bodies). Always allow it.
+  bool isLegalToInline(Region* dest, Region* src, bool wouldBeCloned,
+                       BlockAndValueMapping& valueMapping) const final {
+    return true;
+  }
+  // Operations in CHLO dialect are always legal to inline since they are pure.
+  bool isLegalToInline(Operation*, Region*, bool,
```
GleasonK, line 533: Not sure if clang-tidy complains about unused variables in inlined class method definitions?

GleasonK (approved):

stablehlo/dialect/ChloOps.cpp
burmako: Hmm this is what we're doing in StableHLO right now (I copypasted the implementation from there). If this becomes a problem during integrate, I'll get back and fix it here.

PR #895, opened 2023-01-06
Rename .cpp files for legalize passes to match pass names

burmako:


burmako:
@GleasonK Sounds like there might be a bug in the lint action?

GleasonK (approved):
LGTM. I've recently been thinking about just renaming these to `-vhlo-to-stablehlo`, etc. Then we'd have:

```
-stablehlo-to-vhlo
-vhlo-to-stablehlo
-vhlo-to-version
```

This seems consistent, but I still think that the current naming convention is far more consistent with MHLO. We have many `--A-legalize-to-B` passes in mlir-hlo.

GleasonK:
Rebase on main should fix lint failure. Had a similar failure in #899 which removed `TypeConversion.cpp` which now passes after rebase on the latest lint fix.

burmako:
Thank you for the review! I'll just force-merge to avoid rerunning build and test which will take a long time given that we've recently done an LLVM commit bump.

PR #898, opened 2023-01-06
Create inferFooOp functions for some infeasible ops

burmako:
In this PR, we introduce inferFooOp functions for ops where result's element type is load-bearing: convert, convolution, dot, dot_general, uniform_quantize.

These functions return SmallVectorImpl<ShapedTypeComponents>& and only fill in the shape part. Even though these functions aren't be usable from the existing type inference framework (missing element types will mean that type inference will fail), they will allow sharing shape inference logic with #622.

bitcast_convert is an odd one. It does belong in this category, but its shape curiously depends on the relationship between the operand type and the result type, so implementing inferBitcastConvertOp is not possible, even if it returns ShapedTypeComponents.

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -897,35 +897,6 @@ LogicalResult verifyConvolutionAttributes(
   return success();
 }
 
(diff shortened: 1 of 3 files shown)


zhouxin913 (approved):
Thanks for the change, especially optimize the old "failed to infer shape" message.

PR #899, opened 2023-01-06
Fork FuncOp and CallOp to VHLO

### LINKED ISSUE #850: Implement compatibility for Module/Func/Call/Return ops
After these ops are added to the StableHLO opset in #425, we need to implement compatibility guarantees for them - either for their `stablehlo.foo` versions if we decide to fork them or for their `vhlo.foo` versions if not.

GleasonK:
Leaving out `ModuleOp` for now. It is awkward to convert ModuleOp to something else, also would also require adding `--no-implicit-module` to all calls which is less ergonomic and easy to miss. May decide to avoid forking `ModuleOp`, and instead just enforce that the op has valid types/attrs. If that op disappears we can get clever in the BytecodeReader. We can discuss this further next week.

This PR (built on top of ):
- Fork Ops: `FuncOp, CallOp`
- Fork Types: `FunctionType`
- Fork Attributes: `DictionaryAttr, TypeAttr` (`TypeAttr` used for `FunctionType` attribute)
- Add conversions: 
  + `func.func` <--> `vhlo.func`
  + `func.call` <--> `vhlo.call`
  + `func.return` --> `vhlo.return`
  + `vhlo.return` --> `func.return` if parens is `vhlo.func`, else `stablehlo.return`
- Added pretty print for func ops, since these are so common and generic format of functions is difficult to write tests for.

Closes #850.


stablehlo/dialect/VhloAttrs.td
```diff
@@ -206,6 +212,19 @@ def VHLO_StringAttr : VHLO_AttrDef<"StringV1"> {
...
+  let parameters = (ins "::mlir::Type":$value);
+  let assemblyFormat = "`<` $value `>`";
+}
+/*
+def VHLO_DictionaryAttr : VHLO_AttrDef<"DictionaryV1"> {
+  let mnemonic = "dict";
+  let parameters = (ins Variadic<VHLO_AnyType>:$entries); // NamedAttributes
+  let assemblyFormat = "`<` $entries `>`";
+}
+*/
```
burmako, line 226: Let's delete commented out code?

stablehlo/dialect/VhloBytecode.cpp
```diff
@@ -192,32 +193,43 @@ enum AttributeCode {
   ///   }
   kDenseIntOrFPElementsAttr = 21,
 
+  ///   ArrayAttr {
+  ///   DictionaryAttr {
+  ///     attrs: <StringAttr, Attribute>[]
+  ///   }
+  kDictionaryAttr = 22,
```
burmako, line 200: Reusing this constant in this specific case LGTM since we don't provide stability guarantees for VHLO bytecode yet. Same below.

stablehlo/dialect/VhloOps.cpp
```diff
@@ -68,10 +77,36 @@ ParseResult parseTensorShape(AsmParser& parser,
   return success();
 }
 
+// Print types in parns: (!vhlo.type, !vhlo.type)
```
burmako, line 80: "parentheses"? I don't think this is a particularly conventional abbreviation.

stablehlo/tests/vhlo_to_version_attrtype_invalid.mlir
```diff
@@ -10,97 +10,116 @@
 
 
 // This simulates version validation if a new numeric type is introduced.
```
burmako, line 12: We probably also want to have negative tests for invalid function types?

stablehlo/transforms/CMakeLists.txt
```diff
@@ -16,18 +16,6 @@ set(LLVM_TARGET_DEFINITIONS Passes.td)
 mlir_tablegen(Passes.h.inc -gen-pass-decls)
 add_public_tablegen_target(PassesIncGen)
 
-add_mlir_dialect_library(StablehloTypeConversion
-  PARTIAL_SOURCES_INTENDED
-  TypeConversion.cpp
```
burmako, line 21: How about we move the definitions of these functions from the .h file to the .cpp file? The header is already 200+ lines of code.

stablehlo/transforms/VhloLegalizeToStablehlo.cpp
```diff
@@ -120,6 +123,16 @@ Attribute convertAttrToStablehlo(Attribute vhloAttr,
     return DenseIntOrFPElementsAttr::getFromRawBuffer(
         builtinType, denseElements.getRawData());
   }
+  if (auto dictAttr = vhloAttr.dyn_cast<vhlo::DictionaryV1Attr>()) {
+    SmallVector<NamedAttribute> vhloAttrs;
+    for (auto namedAttr : dictAttr.getValue()) {
+      auto vhloName = convertAttrToStablehlo(namedAttr.first, typeConverter);
```
burmako, line 129: `builtinName`, etc?

stablehlo/transforms/VhloLegalizeToStablehlo.cpp
```diff
@@ -247,7 +274,8 @@ void populateVhloToStablehloPatterns(RewritePatternSet* patterns,
   populateVhloToStablehloPatterns<
 #define GET_OP_LIST
 #include "stablehlo/dialect/StablehloOps.cpp.inc"
-      >(patterns, converter, context);
+      , func::CallOp, func::FuncOp>(patterns, converter, context);
```
burmako, line 277: Why do we need to handle `func::CallOp` and `func::FuncOp` here at all?

stablehlo/transforms/VhloToVersion.cpp
```diff
@@ -226,12 +238,12 @@ struct VhloToVersionPass : public VhloToVersionPassBase<VhloToVersionPass> {
         [&targetVersion](Operation* op) {
           return isLegalOperation(op, targetVersion);
         });
+    target.addIllegalDialect<stablehlo::StablehloDialect, func::FuncDialect>();
```
burmako, line 241: Why are these dialects called out here?

burmako (changes requested):

stablehlo/transforms/VhloLegalizeToStablehlo.cpp
GleasonK: Since we create patterns based on Mapping results, these two create the following patterns:

```
vhlo.func --> func.func
vhlo.call --> func.call
```

When these ops are added to StableHLO (and therefore included by `GET_OP_LIST`) we can remove these.

stablehlo/dialect/VhloBytecode.cpp
GleasonK: That's the thought. After we roll out our first version, these numbers will not be allowed to change.

stablehlo/dialect/VhloOps.cpp
GleasonK: meant parens. Changed to "parentheses"

stablehlo/transforms/VhloToVersion.cpp
GleasonK: Caught some issues with tests using this one. For the `vhlo-to-version` pass, everything should be VHLO, except builtin.module.

stablehlo/transforms/StablehloLegalizeToVhlo.cpp
```diff
@@ -175,6 +175,16 @@ Attribute convertAttrToVhlo(Attribute stablehloAttr,
     return vhlo::DenseIntOrFPElementsV1Attr::get(attr.getContext(), vhloType,
                                                  attr.getRawData());
   }
+  if (auto dictAttr = stablehloAttr.dyn_cast<DictionaryAttr>()) {
```
GleasonK, line 178: Change to `attr =`

stablehlo/transforms/VhloLegalizeToStablehlo.cpp
```diff
@@ -143,6 +145,18 @@ Attribute convertAttrToStablehlo(Attribute vhloAttr,
     return DenseIntOrFPElementsAttr::getFromRawBuffer(builtinType,
                                                       attr.getRawData());
   }
+  if (auto dictAttr = vhloAttr.dyn_cast<vhlo::DictionaryV1Attr>()) {
```
GleasonK, line 148: Change to `attr =`

GleasonK:
Rebased on main, makes the review much easier.

stablehlo/tests/vhlo_to_version_attrtype_invalid.mlir
GleasonK: Good catch, turns out we don't validate function types. Fixed and tested

stablehlo/dialect/VhloBytecode.cpp
```diff
@@ -191,31 +192,42 @@ enum AttributeCode {
   ///   }
   kDenseIntOrFPElementsAttr = 17,
 
+  ///   ArrayAttr {
+  ///   DictionaryAttr {
+  ///     attrs: <StringAttr, Attribute>[]
```
burmako, line 197: `attrs: <Attribute, Attribute>[]` based on how this is defined in VhloAttrs.td?

stablehlo/dialect/VhloAttrs.td
```diff
@@ -252,6 +269,20 @@ def VHLO_StringAttrV1 : VHLO_AttrDef<"StringV1"> {
...
 }
 
+def VHLO_TypeAttrV1 : VHLO_AttrDef<"TypeV1"> {
+  let mnemonic = "type";
+  let parameters = (ins "::mlir::Type":$value);
+  let genVerifyDecl = 1;
+  let extraClassDefinition = [{
+    LogicalResult TypeV1Attr::verify(
+        llvm::function_ref<mlir::InFlightDiagnostic ()> errFn, mlir::Type value) {
+      if (!isFromVhlo(value)) return errFn() << "expected VHLO types";
```
burmako, line 279: Typo: "type".

stablehlo/dialect/VhloOps.cpp
```diff
@@ -102,13 +107,84 @@ Attribute IntegerV1Attr::parse(AsmParser& parser, mlir::Type) {
                             attr.getValue());
 }
 
+static void printDictionary(AsmPrinter& os,
```
burmako, line 110: `printAttributeDictionary` / `parseAttributeDictionary` based on `printAttributeArray` / `parseAttributeArray`?

stablehlo/dialect/VhloOps.cpp
```diff
@@ -102,13 +107,84 @@ Attribute IntegerV1Attr::parse(AsmParser& parser, mlir::Type) {
...
+  SmallVector<Type> inputTypes;
+  SmallVector<Type> resultTypes;
+  if (failed(parser.parseSymbolName(strName)) ||
+      failed(
+          parser.parseArgumentList(args, AsmParser::Delimiter::Paren, true)) ||
+      failed(parser.parseArrowTypeList(resultTypes)) ||
+      failed(parser.parseRegion(region, args))) {
+    return failure();
+  }
+  name = vhlo::StringV1Attr::get(parser.getContext(), strName.getValue());
```
burmako, line 169: Do we need `vhlo::` here?

stablehlo/dialect/VhloOps.cpp
```diff
@@ -102,13 +107,84 @@ Attribute IntegerV1Attr::parse(AsmParser& parser, mlir::Type) {
...
+}
+
 void DenseIntOrFPElementsV1Attr::print(mlir::AsmPrinter& p) const {
   p << '<'
     << DenseIntOrFPElementsAttr::getFromRawBuffer(
            convertTypeToBuiltinForPrint(getType()), getRawData())
     << '>';
 }
 
+// Parse dense elements using DenseIntOfFPElementsAttr printing.
```
burmako, line 187: Typo: "...OrFP".

stablehlo/dialect/VhloOps.td
```diff
@@ -956,4 +956,22 @@ def VHLO_CstrReshapableOpV1 : VHLO_Op<"cstr_reshapable"> {
...
+  let regions = (region VHLO_AnyRegion:$body);
+  let assemblyFormat = "custom<FunctionBody>($sym_name, $body, $function_type) attr-dict";
+}
+
+def VHLO_CallOpV1 : VHLO_Op<"call"> {
+  let arguments = (ins VHLO_AnyAttr:$callee, Variadic<VHLO_AnyType>:$operands);
+  let results = (outs Variadic<VHLO_AnyType>);
+  let assemblyFormat = [{
+    $callee `(` $operands `)` attr-dict `:` functional-type($operands, results)
+  }];
```
burmako, line 974: Sounds like we could drop the assembly format for `vhlo.call` for now? Agreed that having a `vhlo.func` assembly format is important.

stablehlo/dialect/VhloTypes.td
```diff
@@ -90,7 +90,25 @@ def VHLO_Float64V1 : VHLO_TypeDef<"Float64V1","f64">;
 
 def VHLO_Float8E4M3FNV1 : VHLO_TypeDef<"Float8E4M3FNV1", "f8E4M3FN", "0.4.0">;
 
-def VHLO_Float8E5M2 : VHLO_TypeDef<"Float8E5M2V1", "f8E5M2", "0.4.0">;
+def VHLO_Float8E5M2V1 : VHLO_TypeDef<"Float8E5M2V1", "f8E5M2", "0.4.0">;
+
+def VHLO_FunctionTypeV1 : VHLO_TypeDef<"FunctionV1", "func"> {
```
burmako, line 95: Other types in this file are called `VHLO_SomethingV1` rather than `VHLO_SomethingTypeV1`. Let's update this name accordingly?

stablehlo/transforms/Passes.td
```diff
@@ -36,7 +36,8 @@ def StablehloRefineShapesPass : Pass<"stablehlo-refine-shapes", "ModuleOp"> {
 
 def VhloLegalizeToStablehloPass : Pass<"vhlo-legalize-to-stablehlo", "ModuleOp"> {
   let summary = "Legalize VHLO to StableHLO.";
-  let dependentDialects = ["mlir::stablehlo::StablehloDialect", "mlir::shape::ShapeDialect", "mlir::quant::QuantizationDialect"];
+  let dependentDialects = ["mlir::func::FuncDialect", "mlir::stablehlo::StablehloDialect",
+                           "mlir::shape::ShapeDialect", "mlir::quant::QuantizationDialect"];
```
burmako, line 40: By the way, this reminds me that we [added a bunch of dialects](https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/Register.cpp) to `registerAllDialects` and then discovered that `dependentDialects` achieve the same effect. Can we revert `registerAllDialects` given that?

stablehlo/transforms/VhloLegalizeToStablehlo.cpp
```diff
@@ -143,6 +145,18 @@ Attribute convertAttrToStablehlo(Attribute vhloAttr,
     return DenseIntOrFPElementsAttr::getFromRawBuffer(builtinType,
                                                       attr.getRawData());
   }
+  if (auto attr = vhloAttr.dyn_cast<vhlo::DictionaryV1Attr>()) {
+    SmallVector<NamedAttribute> vhloAttrs;
+    for (auto namedAttr : attr.getValue()) {
+      auto builtinName = convertAttrToStablehlo(namedAttr.first, typeConverter);
```
burmako, line 151: `.dyn_cast_or_null<StringAttr>()`. This way, you won't need isa and cast below.

burmako (changes requested):
Left some relatively minor comments, otherwise all looks good. Thank you for the thorough work!

stablehlo/dialect/VhloOps.cpp
GleasonK: removed all `vhlo::` from this file

stablehlo/transforms/Passes.td
GleasonK: Reverted to just `registry.insert<mlir::sparse_tensor::SparseTensorDialect>();`

burmako (approved):

PR #902, opened 2023-01-07
Support bounds in hlo_test_infer.get_return_type_components rewrite pattern.

subhankarshah:


subhankarshah:
@smit-hinsu PTAL.

smit-hinsu:
It wasn't obvious to me why we are making this change and took some time to understand. Good job fixing the gap that I had missed!

Please add some description:
See https://google.github.io/eng-practices/review/developer/cl-descriptions.html#:~:text=A%20CL%20description%20is%20a,CL%20based%20on%20its%20description.

get_return_types works for all the ops regardless if they have inferReturnTypeComponents or inferReturnTypes so should we just always use get_return_types and simply remove InferReturnTypeComponentsPattern for consistency? We could update the output format of get_return_types as required.

burmako:
"so should we just always use get_return_types and simply remove InferReturnTypeComponentsPattern for consistency?". Nice idea! I think we should 👍

zhouxin913:
Thanks Subhankar for this careful fix! I would agree with Smit and Eugene that it is better to keep `inferReturnTypes` only, because for now I'm now aware of any strong reason to keep them both.
We can either update this PR or create a new PR, trying to remove `InferReturnTypeComponentsPattern`. But it is of a little  lower priority than other bounded dynamism issues I believe.

zhouxin913:
Unassign myself: as the reviewers provide another solution to optimize and we do not intent to submit in current way (Please correct me if you have other thoughts)

zhouxin913 (approved):
We agree on submit this code to fill in the gap, and open a ticket for next steps.

PR #908, opened 2023-01-10
Handle bounds of Gather op

sdasgup3:
The dimension sizes of result, corresponding to offset dimensions, depend on attributes (like `collapsed_slice_dims` and `slice_sizes`) and hence are always static. 

Whereas, the dimension sizes of result, corresponding to batch dimensions, depends on input `start_indices` and could be dynamic. The corresponding bounds, in that case,  are propagated from the `start_indices`.

Side notes:
I deliberately avoid making any changes in `inferGatherShape` which is currently shared by `GatherOp::inferReturnTypeComponents` and `GatherOp::reifyReturnTypeShapes`. The reason being I do not `reifyReturnTypeShapes` needs to handle the bounds. 

Because of the decision, we have  some duplicate code in line 1237 (duplicate of https://github.com/openxla/stablehlo/blob/afcbb953c12a2a3a8564cab73278f4b47021ec01/stablehlo/dialect/TypeInference.cpp#L1149).

PTAL @smit-hinsu

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -1298,17 +1298,15 @@ static LogicalResult inferGatherReturnTypeComponents(
       startIndices.getType().cast<RankedTensorType>().getEncoding());
   SmallVector<int64_t> inferredBounds(resultRank, ShapedType::kDynamic);
   if (!startIndicesBounds.empty()) {
-    SmallVector<int64_t> batchDims;
-    for (int dim = 0; dim < resultRank; ++dim)
-      if (!llvm::is_contained(offsetDims, dim)) batchDims.push_back(dim);
+    llvm::SmallDenseSet<int64_t> offsetDimSet(offsetDims.begin(),
+                                              offsetDims.end());
+    int64_t index = 0;
+    for (int dim = 0; dim < resultRank; ++dim) {
+      if (offsetDimSet.count(dim)) continue;
 
-    for (int i = 0; i < resultRank; ++i) {
-      auto* batchDimsIt = std::find(batchDims.begin(), batchDims.end(), i);
-      if (batchDimsIt == batchDims.end()) continue;
-
-      auto index = std::distance(batchDims.begin(), batchDimsIt);
-      if (index >= indexVectorDim) ++index;
-      inferredBounds[i] = startIndicesBounds[index];
+      if (index == indexVectorDim) ++index;
+      inferredBounds[dim] = startIndicesBounds[index];
+      index++;
     }
   }
 
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -1295,28 +1295,25 @@ static LogicalResult inferGatherReturnTypeComponents(
   // batch dimensions, depends on input `start_indices` and could be dynamic.
   // The corresponding bounds, in that case,  are propagated from the
   // `start_indices`.
-  ArrayRef<int64_t> startIndicesBounds = encodingToBounds(
-      startIndices.getType().cast<RankedTensorType>().getEncoding());
+  Attribute encoding =
+      startIndices.getType().cast<RankedTensorType>().getEncoding();
+  ArrayRef<int64_t> startIndicesBounds = encodingToBounds(encoding);
   SmallVector<int64_t> inferredBounds(resultRank, ShapedType::kDynamic);
   if (!startIndicesBounds.empty()) {
     llvm::BitVector isOffsetDim(resultRank);
     for (auto offsetDim : offsetDims) isOffsetDim.set(offsetDim);
 
-    int64_t index = 0;
-    for (int dim = 0; dim < resultRank; ++dim) {
-      if (isOffsetDim.test(dim)) continue;
+    int64_t startIndicesDim = 0;
+    for (int resultDim = 0; resultDim < resultRank; ++resultDim) {
+      if (isOffsetDim.test(resultDim)) continue;
 
-      if (index == indexVectorDim) ++index;
-      inferredBounds[dim] = startIndicesBounds[index];
-      index++;
+      if (startIndicesDim == indexVectorDim) ++startIndicesDim;
+      inferredBounds[resultDim] = startIndicesBounds[startIndicesDim++];
     }
   }
 
-  inferredReturnShapes.emplace_back(
-      shape, elementType,
-      boundsToEncoding(
-          startIndices.getType().cast<RankedTensorType>().getEncoding(),
-          inferredBounds));
+  inferredReturnShapes.emplace_back(shape, elementType,
+                                    boundsToEncoding(encoding, inferredBounds));
   return success();
 }
 
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -1252,13 +1252,13 @@ void reifyGatherDimSizes(int64_t resultRank,
 //  P1. Verify 0 <= offset_dims[i] < output_shape_rank, for every i.
 //      (output_shape_rank = size(offset_dims) + rank(start_indices) -1)
 static LogicalResult inferGatherReturnTypeComponents(
-    std::optional<Location> location, ShapeAdaptor operandShape,
-    ShapeAdaptor startIndicesShape,
+    Optional<Location> location, ShapeAdaptor operandShape, Value startIndices,
     llvm::function_ref<int64_t(int64_t)> getSliceDim,
     ArrayRef<int64_t> offsetDims, ArrayRef<int64_t> collapsedSliceDims,
     ArrayRef<int64_t> startIndexMap, int64_t indexVectorDim,
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
   Type elementType = operandShape.getElementType();
+  ShapeAdaptor startIndicesShape(startIndices.getType());
 
   // We need this to determine the result rank. We could still place bounds on
   // the result rank if that was something ShapedTypeComponents could express.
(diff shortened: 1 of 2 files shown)


smit-hinsu (commented):
Thanks!

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1223,7 +1223,35 @@ static LogicalResult inferGatherReturnTypeComponents(
                             offsetDims, collapsedSliceDims, startIndexMap,
                             indexVectorDim, shape);
 
-  inferredReturnShapes.emplace_back(shape, elementType);
+  // The dimension sizes of result, corresponding to offset dimensions, depend
```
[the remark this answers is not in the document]
sdasgup3: Hi Smit,

That would be nice and I considered the same.  PTAL at the PR description where I noted my findings. Let me know if I am missing anything.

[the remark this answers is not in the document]
smit-hinsu: ah, I see. sorry, didn't read the description.

I personally would avoid the duplication as this shouldn't negatively affect the other user in practice and will be easier to follow.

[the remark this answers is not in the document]
zhouxin913: How about we apply and patch this PR in MHLO, if everything is OK then we can go on with shared code among `reifyReturnTypeShapes` with confidence?

zhouxin913 (changes requested):
The bounds logic looks good to me. Leave a comment to support deduplicate the code.

stablehlo/dialect/TypeInference.cpp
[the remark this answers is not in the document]
sdasgup3: That sounds good. 

@smit-hinsu @zhouxin913 
Sharing the code might still have problems.

The main reason I did not share [the](https://github.com/openxla/stablehlo/blob/afcbb953c12a2a3a8564cab73278f4b47021ec01/stablehlo/dialect/TypeInference.cpp#L1119) between  `GatherOp::inferReturnTypeComponents` and `GatherOp::reifyReturnTypeShapes` is because I do not think bounds are currently handled in reify functions. For example, which `op` to create to reify bounds from input arguments.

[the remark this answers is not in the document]
smit-hinsu: We don't need to reify bounds in reifyReturnTypeShapes.

Can't we just use inferGatherShape to infer shape and bounds. Use both in inferReturnTypeComponents and drop it within reifyGatherDimSizes. (I don't have the full context around reify.)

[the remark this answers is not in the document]
sdasgup3: Hi Smit,

Here is what I found to be done if we want to share  `inferGatherShape` among  ` reifyGatherDimSizes` and `inferReturnTypeComponents`

Both ` reifyGatherDimSizes` and `inferReturnTypeComponents` need to provide callbacks (similar to `getStartIndicesDim` and `getSliceDim`) which, for `inferReturnTypeComponents`, will help compute the concrete bounds and, for `reifyGatherDimSizes`, help reify the bounds.

The callbacks for computing the concrete bounds are straightforward. Something like
```
  ArrayRef<int64_t> startIndicesBounds = encodingToBounds(
      startIndices.getType().cast<RankedTensorType>().getEncoding());
  
auto getOffsetIndicesBound = [&](int64_t index) {
    return ShapedType::kDynamic;
  };

  auto  getBatchIndicesBound = [&](int64_t index) {
    if(startIndicesBounds.empty())
      return ShapedType::kDynamic;
    return startIndicesBounds[index];
  };

```

For for reifying the bounds, I need to pass some dummy callbacks which will end up creating some dead `mlir::Value` typed objects, probably of type `ConstantIndexOp`, using builder API.

As we are not sure about the full context around reify, my proposal is to keep them separate for now. With more context (which IMO will arise once we have a need to support bounds in reify methods) we can make an  effort to clean this tech debt.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1223,7 +1223,35 @@ static LogicalResult inferGatherReturnTypeComponents(
...
+  // always static. Whereas, the dimension sizes of result, corresponding to
+  // batch dimensions, depends on input `start_indices` and could be dynamic.
+  // The corresponding bounds, in that case,  are propagated from the
+  // `start_indices`.
+  ArrayRef<int64_t> startIndicesBounds = encodingToBounds(
+      startIndices.getType().cast<RankedTensorType>().getEncoding());
+  SmallVector<int64_t> inferredBounds(resultRank, ShapedType::kDynamic);
+  if (!startIndicesBounds.empty()) {
+    SmallVector<int64_t> batchDims;
+    for (int dim = 0; dim < resultRank; ++dim)
```
smit-hinsu, line 1237: can we do something like the following? It can avoid find, distance and is_contained.

```
int64_t index = 0;
// make a bit map of isOffsetDim
for (int dim = 0; dim < resultRank; ++dim) {
  if (isOffsetDim(dim)) continue;
  if (index == indexVectorDim) ++index;
  inferredBounds[dim] = startIndicesBounds[index];
  ++index;
}
```

stablehlo/dialect/TypeInference.cpp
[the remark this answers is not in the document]
smit-hinsu: That sounds completely reasonable. Thanks for digging in!

The code duplication is not large anyway.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1223,7 +1223,35 @@ static LogicalResult inferGatherReturnTypeComponents(
...
+      if (!llvm::is_contained(offsetDims, dim)) batchDims.push_back(dim);
+
+    for (int i = 0; i < resultRank; ++i) {
+      auto* batchDimsIt = std::find(batchDims.begin(), batchDims.end(), i);
+      if (batchDimsIt == batchDims.end()) continue;
```
smit-hinsu, line 1242: for non batch dims, there is this handling in XLA.
https://github.com/tensorflow/tensorflow/blob/85d9f5aef28d6a0aaea6aff0b261e11254f9320c/tensorflow/compiler/xla/service/shape_inference.cc#L3572

Current StableHLO spec with static doesn't allow slice sizes to be larger than the input but we need to figure out semantics in dynamism RFC. Either follow the XLA semantics (as per the implementation) or disallow that. Check with Eugene on what to do here until then.

smit-hinsu (commented):
Thanks!

stablehlo/dialect/TypeInference.cpp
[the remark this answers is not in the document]
sdasgup3: Thanks!

stablehlo/tests/infer_stablehlo.mlir
```diff
@@ -154,6 +154,27 @@ func.func @gather(%operand : tensor<2x4x9xi32>, %start_indices : tensor<1x5x2xi3
 
 // -----
 
+// CHECK-LABEL: @gather_bounds
+func.func @gather_bounds(%operand : tensor<2x4x9xi32>, %start_indices : tensor<?x?x?xi32, #stablehlo.type_extensions<bounds = [16, 32, 64]>>) -> tensor<*xindex> {
```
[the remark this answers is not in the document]
sdasgup3: done.

stablehlo/dialect/TypeInference.cpp
sdasgup3: Thanks!
@smit-hinsu  PTAL.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1277,7 +1277,33 @@ static LogicalResult inferGatherReturnTypeComponents(
...
+  // on attributes (like `collapsed_slice_dims` and `slice_sizes`) and hence are
+  // always static. Whereas, the dimension sizes of result, corresponding to
+  // batch dimensions, depends on input `start_indices` and could be dynamic.
+  // The corresponding bounds, in that case,  are propagated from the
+  // `start_indices`.
+  ArrayRef<int64_t> startIndicesBounds = encodingToBounds(
+      startIndices.getType().cast<RankedTensorType>().getEncoding());
+  SmallVector<int64_t> inferredBounds(resultRank, ShapedType::kDynamic);
+  if (!startIndicesBounds.empty()) {
+    llvm::SmallDenseSet<int64_t> offsetDimSet(offsetDims.begin(),
```
smit-hinsu, line 1290: llvm::bitVector could come handy here.

smit-hinsu (approved):
Thanks!

stablehlo/dialect/TypeInference.cpp
sdasgup3: That is a good point. Thanks for pointing! 
So, if the size sizes is larger than input, then then the bounds of input should be propagated to result's bounds.   

cc @burmako Can you please help here?

stablehlo/dialect/TypeInference.cpp
sdasgup3: Sure. done.

smit-hinsu (approved):

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1277,7 +1278,34 @@ static LogicalResult inferGatherReturnTypeComponents(
...
 
-  inferredReturnShapes.emplace_back(shape, elementType);
+  // The dimension sizes of result, corresponding to offset dimensions, depend
+  // on attributes (like `collapsed_slice_dims` and `slice_sizes`) and hence are
+  // always static. Whereas, the dimension sizes of result, corresponding to
+  // batch dimensions, depends on input `start_indices` and could be dynamic.
+  // The corresponding bounds, in that case,  are propagated from the
+  // `start_indices`.
+  ArrayRef<int64_t> startIndicesBounds = encodingToBounds(
+      startIndices.getType().cast<RankedTensorType>().getEncoding());
```
burmako, line 1288: Would be good to extract `startIndices.getType().cast<RankedTensorType>().getEncoding()` into a temporary variable like `startIndicesEncoding` since it's a pretty sizable expression and is used more than once.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1277,7 +1278,34 @@ static LogicalResult inferGatherReturnTypeComponents(
...
+  SmallVector<int64_t> inferredBounds(resultRank, ShapedType::kDynamic);
+  if (!startIndicesBounds.empty()) {
+    llvm::BitVector isOffsetDim(resultRank);
+    for (auto offsetDim : offsetDims) isOffsetDim.set(offsetDim);
+
+    int64_t index = 0;
+    for (int dim = 0; dim < resultRank; ++dim) {
```
burmako, line 1295: Let's be a bit more explicit and rename `index` to `startIndicesDim` and `dim` to `resultDim`?

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1277,7 +1278,34 @@ static LogicalResult inferGatherReturnTypeComponents(
...
+      if (isOffsetDim.test(dim)) continue;
+
+      if (index == indexVectorDim) ++index;
+      inferredBounds[dim] = startIndicesBounds[index];
+      index++;
```
burmako, line 1300: Merge `index++` into the previous line?

stablehlo/dialect/TypeInference.cpp
burmako: That's a nice subtlety - thank you for calling this out! Let's keep the current implementation (i.e. what's in the PR) for now, and I'll make a note for the dynamism RFC.

burmako (approved):

zhouxin913 (approved):

burmako (approved):

smit-hinsu (approved):

PR #909, opened 2023-01-10
Handle bounds in the WhileOp shape function

zhouxin913:
WhileOp has trivial shape function which use operands types as result type. So nothing changes in this PR except providing a unit test cover that the result types are same as operand types with bounds. 

@smit-hinsu for review.

### DIFF
--- a/stablehlo/tests/infer_stablehlo.mlir
+++ b/stablehlo/tests/infer_stablehlo.mlir
@@ -992,6 +992,36 @@ func.func @concat_bounds_unranked_c1(
 
 // -----
 
+// CHECK-LABEL: while_bounds
+func.func @while_bounds(
+  %arg00: tensor<3xf32>,
+  %arg11: tensor<1xi32>,
+  %arg22: tensor<2x?xi32, #stablehlo.type_extensions<bounds = [?, 4]>>,
+  %arg33: tensor<1xf32>) -> tensor<*xindex> {
+  %1:4 = "stablehlo.while"(%arg11, %arg22, %arg33, %arg00) ({
+  ^bb0(%arg1: tensor<1xi32>, %arg2: tensor<2x?xi32, #stablehlo.type_extensions<bounds = [?, 4]>>, %arg3: tensor<1xf32>, %arg4: tensor<3xf32>):
+    %2 = arith.constant dense<0> : tensor<i32>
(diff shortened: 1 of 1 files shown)


stablehlo/tests/infer_stablehlo.mlir
```diff
@@ -992,6 +992,36 @@ func.func @concat_bounds_unranked_c1(
...
+      (tensor<2x?xi32, #stablehlo.type_extensions<bounds = [?, 4]>>) -> tensor<1x1xi32>
+    %4 = "stablehlo.compare"(%3, %3) {comparison_direction = #stablehlo<comparison_direction LT>} : (tensor<1x1xi32>, tensor<1x1xi32>) -> tensor<1x1xi1>
+    %5 = "stablehlo.reshape"(%4) : (tensor<1x1xi1>) -> tensor<i1>
+    "stablehlo.return"(%5) : (tensor<i1>) -> ()
+  },  {
+  ^bb0(%arg1: tensor<1xi32>, %arg2: tensor<2x?xi32, #stablehlo.type_extensions<bounds = [?, 4]>>, %arg3: tensor<1xf32>, %arg4: tensor<3xf32>):
+    %3 = "stablehlo.broadcast_in_dim"(%arg3) {broadcast_dimensions = dense<0> : tensor<1xi64>} : (tensor<1xf32>) -> tensor<3xf32>
+    %4 = stablehlo.add %3, %arg4 : tensor<3xf32>
+    "stablehlo.return"(%arg1, %arg2, %arg3, %4) : (tensor<1xi32>, tensor<2x?xi32, #stablehlo.type_extensions<bounds = [?, 4]>>, tensor<1xf32>, tensor<3xf32>) -> ()
+  }) : (tensor<1xi32>, tensor<2x?xi32, #stablehlo.type_extensions<bounds = [?, 4]>>, tensor<1xf32>, tensor<3xf32>) -> (tensor<*xi32>, tensor<*xi32>, tensor<*xf32>, tensor<*xf32>)
```
burmako, line 1014: How about simplifying the while loop as much as possible? E.g. the cond could be replaced with `return true`, and the body could be replaced with `return arguments`. Similarly, do we need all the 4 operands for this while op? types0, types2, types3 test exactly the same thing as far as this test is concerned.

burmako (changes requested):

stablehlo/tests/infer_stablehlo.mlir
zhouxin913: Good suggestion! Fixed.
I keep 2 args: 1 with bounds and 1 not, for testing that bounds and non-bounds can exist together in this multiple case.

burmako (approved):

Issue #544, opened 2022-11-14
Change embedded CI to use strict warnings


GleasonK:
### Request description

Build LLVM CI with warnings as errors, all warnings, and pedantic warnings enabled.

GleasonK:
We need a way to build _only_ StableHLO on embedded LLVM builds using the LLVM strict build flags. We cannot call back into `HandleLLVMOptions.cmake` to set these flags for us.

1. We could hardcode `set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}  -Werror -Wall -Wno-long-long -Wc++98-compat-extra-semi -pedantic")` but that could go out of sync if upstream changes. This may not be a big deal since we will build the non-embedded build using the proper flags.
2. We could build all of LLVM with `LLVM_ENABLE_WARNINGS, LLVM_ENABLE_WERROR, LLVM_ENABLE_PEDANTIC` set to `On`, but this would lead to build failures on embedded builds when upstream has build warnings.

I wonder if there's any way to build embedded with LLVM defaults (Werror off), and notify, instead of faili, when there are build warnings for the embedded build.

burmako:
"We could hardcode set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror -Wall -Wno-long-long -Wc++98-compat-extra-semi -pedantic") but that could go out of sync if upstream changes". This sounds like the best option, given the tradeoffs.

PR #915, opened 2023-01-11
Replace makeArrayRef with ArrayRef in Python bindings

burmako:
The recent integrate PR (#912) missed fixing a few deprecated makeArrayRef calls because our build infrastructure didn't flag them (see #544 for the ticket that talks about the details).

This PR handles the missed calls.


GleasonK:
I'm confused why CI didn't catch this. We still run non-embedded build in strict mode. My local build fails as well, which is using the same build script as CI.

GleasonK (approved):

GleasonK:
Ah this is just in a file that builds embedded, that makes more sense.

PR #900, opened 2023-01-06
Infer bounds for BatchNorm* Ops.

subhankarshah:
closes #809, closes #810, closes #811, closes #869

Unify verification and inference for BatchNorm* Ops

  * For all `batch_norm_*` ops infer all shapes and bounds from `operand` only.
  * Verify `operand` feature dimension compatibility with `dim[0]` of `scale`.
  * Remove `AllShapesMatch` and convert `AllTypesMatch` to `AllElementTypesMatch`.
  * Check for shape compatibility instead of equality.

PR #917, opened 2023-01-11
Handle bounds in the SelectAndScatterOp shape function

### LINKED ISSUE #754: Handle bounds in the SelectAndScatterOp shape function

ghpvnist:
SelectAndScatterOp's bound inference is very simple. There are two inputs: `operand` and `source`. The `source` contains the indices to the `result` tensor to place the reduced values from `operand`, so it does not affect the bound of the `result` tensor. As for `operand`, the type is equal to the `result` tensor. There is only a movement of values happening in this op (after reduction is applied, but this also does not affect the bounds), so we can propagate the bounds with no further inference done.

closes #754

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -2546,15 +2546,7 @@ LogicalResult inferSelectOp(
 
 LogicalResult inferSelectAndScatterOp(
     Value operand, SmallVectorImpl<Type>& inferredReturnTypes) {
-  auto operandRankedType = operand.getType().dyn_cast<RankedTensorType>();
-  if (!operandRankedType) {
-    inferredReturnTypes.push_back(operand.getType());
-  } else {
-    auto bounds = encodingToBounds(operandRankedType.getEncoding()).vec();
-    inferredReturnTypes.push_back(RankedTensorType::get(
-        operandRankedType.getShape(), operandRankedType.getElementType(),
-        boundsToEncoding(operandRankedType.getEncoding(), bounds)));
-  }
+  inferredReturnTypes.push_back(operand.getType());
   return success();
 }
 
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -2546,7 +2546,15 @@ LogicalResult inferSelectOp(
 
 LogicalResult inferSelectAndScatterOp(
     Value operand, SmallVectorImpl<Type>& inferredReturnTypes) {
-  inferredReturnTypes.push_back(operand.getType());
+  auto operandRankedType = operand.getType().dyn_cast<RankedTensorType>();
+  if (!operandRankedType) {
+    inferredReturnTypes.push_back(operand.getType());
+  } else {
+    auto bounds = encodingToBounds(operandRankedType.getEncoding()).vec();
+    inferredReturnTypes.push_back(RankedTensorType::get(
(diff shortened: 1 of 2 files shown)


stablehlo/dialect/TypeInference.cpp
```diff
@@ -2544,7 +2544,15 @@ LogicalResult inferSelectOp(
 
 LogicalResult inferSelectAndScatterOp(
     Value operand, SmallVectorImpl<Type>& inferredReturnTypes) {
-  inferredReturnTypes.push_back(operand.getType());
+  auto operandRankedType = operand.getType().dyn_cast<RankedTensorType>();
+  if (!operandRankedType) {
```
zhouxin913, line 2548: nit: could you reverse the 2 branches of if-else, so that "If(!xxx)" can be avoid.

zhouxin913 (commented):
LGTM.

add @smit-hinsu and @burmako for review

stablehlo/dialect/TypeInference.cpp
```diff
@@ -2544,7 +2544,15 @@ LogicalResult inferSelectOp(
...
+    inferredReturnTypes.push_back(operand.getType());
+  } else {
+    auto bounds = encodingToBounds(operandRankedType.getEncoding()).vec();
+    inferredReturnTypes.push_back(RankedTensorType::get(
+        operandRankedType.getShape(), operandRankedType.getElementType(),
+        boundsToEncoding(operandRankedType.getEncoding(), bounds)));
+  }
```
burmako, line 2555: Why do we need to change the original code at all? The `encodingToBounds` followed by `boundsToEncoding` seems to be a no-op?

burmako (changes requested):

stablehlo/dialect/TypeInference.cpp
ghpvnist: Good point, the bound information is carried by the type, so this separation of logic is not needed.

burmako (approved):

smit-hinsu (approved):
Thanks!

Unless we don't already have a ticket, let's file one to add AllTypesMatch/AllShapesMatch traits and then, we might be able to even avoid shape functions using some MLIR magic. Technically, shapes don't have to match and just need to be compatible but we require equal shapes for binary ops so should apply here as well.

ghpvnist:
After further look, adding AllTypesMatch/AllShapesMatch traits do not work when we need compatible types, and #900 is migrating away from them.

burmako:
"Technically, shapes don't have to match and just need to be compatible but we require equal shapes for binary ops so should apply here as well". Overall, the trend in the last 6 months was to require compatible shapes wherever possible to better support dynamic programs. For example, binary ops say `HLO_CompatibleOperandsAndResultType` instead of `SameOperandsAndResultType`, and we've made similar relaxations to other ops too. Here's an example: https://github.com/openxla/stablehlo/blob/067750f1bfaf3361611571fb2f545f5db0351d2e/stablehlo/dialect/StablehloOps.td#L675-L676.

smit-hinsu:
Ah, right. This makes sense. 

I was looking at StableHLO_BinaryElementwiseOp and got confused. That is still adding the SameOperandsAndResultShape trait. This trait is named same but actually checks for compatibility. I looked at the 15 ops deriving from this and only MHLO_ComplexOp needs that. Even for that we should probably defined and use  HLO_CompatibleOperandsAndResultShape trait. Similarly, for unary and other ops using SameOperandsAndResultShape trait but that is more involved as the element types differ.

Issue #481, opened 2022-11-08
Decide on signedness of various attributes


burmako:
MHLO/StableHLO dialects frequently use `I64Attr` and `I64ElementsAttr` which are signless. HLO uses signed integers. In the spec, we're using signed integers. Let's review and make a consistent decision once and for all.

ghpvnist:
To give an example, PR for ReducePrecisionOp #482 uses `I32Attr` for `exponent_bits` and `mantissa_bits`. The ODS enforces that these are `uint32`, but the spec currently says "constant of type `si32`".

Related to this, there may be inconsistencies between the use of "constant of type x" and "1-dimensional tensor constant of type x" for these attributes which may need to be sorted out.

burmako:
I don't think that the ODS enforces anything. Since `I32Attr` means "signless", it just bitcasts the value to `uint32_t`. For example, this passes the verifier at the moment:

```mlir
func.func @main(%arg: tensor<f32>) -> tensor<f32> {
  %0 = "stablehlo.reduce_precision"(%arg) {
    exponent_bits = -5 : i32,
    mantissa_bits = 2 : i32
  } : (tensor<f32>) -> tensor<f32>
  func.return %0 : tensor<f32>
}
```

GleasonK:
Signed integers in MLIR are very unused, I believe because they were added much later than signless integers. As such I don't think its a good idea to diverge from the MLIR standard way of representing signed values using the signless int type. Doing so would complicate all conversions to/from StableHLO, and make dialect interop much more difficult.

Closing this issue, if a reason to start considering this distinction comes up, or if upstream / MLIR community starts to use the type more, we can reconsider.

PR #919, opened 2023-01-12
Verify exponent and mantissa in ReducePrecisionOp

### LINKED ISSUE #488: Verify exponent and mantissa in ReducePrecisionOp
### Request description

https://github.com/openxla/stablehlo/blob/2988c0f5ed660d5007b8a780525bbda94c6ffb18/stablehlo/dialect/StablehloOps.cpp#L2820 should be
```
if ((int32_t)getExponentBits() < 1) {
```

and a check for mantissa bits need to be added.
```
if ((int32_t)getMantissaBits() < 0) {
```

ghpvnist:
closes #488

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -3434,9 +3434,12 @@ LogicalResult verifyReduceOp(Optional<Location> location, ValueRange inputs,
 // We intend to verify the following properties
 //  P2. exponent_bits >= 1
 LogicalResult verifyReducePrecisionOp(Optional<Location> location,
-                                      int32_t exponentBits) {
-  if (exponentBits < 1)
+                                      uint32_t exponentBits,
+                                      uint32_t mantissaBits) {
+  if ((int32_t)exponentBits < 1)
     return emitOptionalError(location, "exponent_bits must be at least 1.");
+  if ((int32_t)mantissaBits < 0)
+    return emitOptionalError(location, "mantissa_bits must be at least 0.");
   return success();
 }
 
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -3428,18 +3428,14 @@ LogicalResult verifyReduceOp(Optional<Location> location, ValueRange inputs,
   return success();
 }
 
-// The following property is already enforced by the ODS:
-//  P0. operand element type is float
-// We intend to verify the following properties (since I32Attr is currently
-// represented by an unsigned 32-bit integer type):
-//  P1. exponent_bits >= 1
-//  P2. mantissa_bits >= 0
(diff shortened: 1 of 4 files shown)


stablehlo/dialect/TypeInference.cpp
```diff
@@ -3434,9 +3434,12 @@ LogicalResult verifyReduceOp(Optional<Location> location, ValueRange inputs,
 // We intend to verify the following properties
 //  P2. exponent_bits >= 1
```
sdasgup3, line 3435: update the comment.

sdasgup3 (approved):
Thanks!

LGTM with a minor comment.

stablehlo/dialect/TypeInference.h
```diff
@@ -412,7 +412,8 @@ LogicalResult verifyReduceOp(Optional<Location> location, ValueRange inputs,
                              DenseIntElementsAttr dimensions, Region& body);
 
 LogicalResult verifyReducePrecisionOp(Optional<Location> location,
-                                      int32_t exponentBits);
+                                      uint32_t exponentBits,
+                                      uint32_t mantissaBits);
```
burmako, line 416: At the moment, it is conventional to use signed integers for metadata like this (e.g. TypeInference.h doesn't feature any uint types). #481 is still pending, so this is not a final decision yet, but I think it would be good to be consistent with the rest of the file (especially since the implementation immediately casts these parameters to int32_t).

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3430,13 +3430,17 @@ LogicalResult verifyReduceOp(Optional<Location> location, ValueRange inputs,
 
 // The following property is already enforced by the ODS:
 //  P0. operand element type is float
-//  P1. mantissa_bits >= 0
-// We intend to verify the following properties
-//  P2. exponent_bits >= 1
+// We intend to verify the following properties (since I32Attr is currently
+// represented by an unsigned 32-bit integer type):
+//  P1. exponent_bits >= 1
+//  P2. mantissa_bits >= 0
```
burmako, line 3436: What do you folks think about dropping these comments and referring to individual constraints from the spec similarly to #686?

burmako (changes requested):

stablehlo/dialect/TypeInference.cpp
ghpvnist: I agree, we should refer to the spec constraints.

stablehlo/dialect/TypeInference.h
ghpvnist: IIRC a while back, having a -1 as the exponentBits passed the test (somehow interpreting this as the max unsigned value), which prompted to filing this bug. It looks that problem doesn't exist anymore. I'll make the changes in a bit.

burmako (approved):

PR #920, opened 2023-01-12
Handle bounds for ScatterOp shape function

### LINKED ISSUE #749: Handle bounds in the ScatterOp shape function

sdasgup3:
fixes #749 

For the ScatterOp, where the 
(C16) inputs[k] and result[k] have the same type for any k  [0, N)

and this is how the result type is inferred. 
(https://github.com/openxla/stablehlo/blob/81d520685c5cfe2620f995dbb09ac4399d000c9a/stablehlo/dialect/TypeInference.cpp#L2513).
 
IMO, we do not need to do anything special to handle the bounds as the bounds should get copied as part of copying the types. I wrote a test to check that input bounds are indeed propagated. Also I note that I do not have to use `inferReturnTypeComponents`  at all, for this case.

### DIFF
--- a/stablehlo/tests/infer_stablehlo.mlir
+++ b/stablehlo/tests/infer_stablehlo.mlir
@@ -472,7 +472,7 @@ func.func @scatter(%input_tensor: tensor<200x100x300xf32>,
 // -----
 
 // CHECK-LABEL: func @scatter_bounds
-func.func @scatter_bounds(%input_tensor: tensor<200x?x?xf32, #stablehlo.type_extensions<bounds = [?, 101, 301]>>,
+func.func @scatter_bounds(%input_tensor: tensor<200x?x?xf32, #stablehlo.type_extensions<bounds = [?, ?, 301]>>,
     %scatter_indices: tensor<10x2xi32>, %updates: tensor<10x300xf32>) ->
       tensor<*xindex> {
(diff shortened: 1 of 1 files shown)


stablehlo/tests/infer_stablehlo.mlir
```diff
@@ -471,6 +471,33 @@ func.func @scatter(%input_tensor: tensor<200x100x300xf32>,
 
 // -----
 
+// CHECK-LABEL: func @scatter_bounds
+func.func @scatter_bounds(%input_tensor: tensor<200x?x?xf32, #stablehlo.type_extensions<bounds = [?, 101, 301]>>,
```
zhouxin913, line 475: nit: the last 2 dims are testing the same thing, how about we change to:
tensor<200x?x?xf32, #stablehlo.type_extensions<bounds = [?, ?, 301]>>

zhouxin913 (approved):
Approve with a small nit suggestion.

zhouxin913:
@smit-hinsu I believe this op is trivial too. Just add you here for visibility.

stablehlo/tests/infer_stablehlo.mlir
sdasgup3: Good point!

PR #927, opened 2023-01-12
Verify dimension in GetDimensionSizeOp

### LINKED ISSUE #790: Verify dimension in GetDimensionSizeOp

ghpvnist:
Of the two constraints mentioned in #790:
* (C1) is already checked by [verifyDimAttr](https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/StablehloOps.cpp#L117-L120), but have added a negative test
* (C2) is not added per discussion [here](https://github.com/openxla/stablehlo/pull/791#discussion_r1053899143)

closes #790

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -2716,7 +2716,7 @@ Produces the size of the given `dimension` of the `operand`.
 
 ### Constraints
 
-  * (C1) 0 $\le$ `dimension` $\lt$ `rank(operand)`.
+* (C1) 0 $\le$ `dimension` $\lt$ `rank(operand)`.
 
 ### Examples
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -2716,8 +2716,7 @@ Produces the size of the given `dimension` of the `operand`.
 
 ### Constraints
 
-* (C1) 0 $\le$ `dimension` $\lt$ `rank(operand)`.
-  [todo](https://github.com/openxla/stablehlo/issues/790)
+  * (C1) 0 $\le$ `dimension` $\lt$ `rank(operand)`.
 
 ### Examples
 
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -105,17 +105,20 @@ void createArgs(ArrayRef<OpAsmParser::UnresolvedOperand> operands,
 template <typename OpT>
 static LogicalResult verifyDimAttr(OpT op) {
   int64_t rank = -1;
-  if (auto ty =
-          op.getOperand().getType().template dyn_cast<RankedTensorType>()) {
-    rank = ty.getRank();
+  auto operandTy =
+      op.getOperand().getType().template dyn_cast<RankedTensorType>();
+  if (operandTy) {
+    rank = operandTy.getRank();
   } else if (auto ty = op.getType().template dyn_cast<RankedTensorType>()) {
     rank = ty.getRank();
-  } else {
-    return success();
   }
-
   int64_t dim = op.getDimension();
-  if (dim < 0 || dim >= rank)
+  if (dim < 0) {
+    return op.emitOpError()
+           << "requires non-negative dimension attribute; found (" << dim
+           << ")";
+  }
+  if (operandTy && dim >= rank)
     return op.emitOpError() << "requires dimension attribute in range [0, "
                             << rank << "); found (" << dim << ")";
   return success();
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -100,27 +100,13 @@ void createArgs(ArrayRef<OpAsmParser::UnresolvedOperand> operands,
 // Utilities for the canonicalize patterns
 //===----------------------------------------------------------------------===//
 
-// Verifies that dimension attribute for the op correctly indexes in operand or
-// result shape.
-template <typename OpT>
-static LogicalResult verifyDimAttr(OpT op) {
-  int64_t rank = -1;
-  auto operandTy =
-      op.getOperand().getType().template dyn_cast<RankedTensorType>();
-  if (operandTy) {
-    rank = operandTy.getRank();
-  } else if (auto ty = op.getType().template dyn_cast<RankedTensorType>()) {
-    rank = ty.getRank();
-  }
-  int64_t dim = op.getDimension();
-  if (dim < 0) {
-    return op.emitOpError()
-           << "requires non-negative dimension attribute; found (" << dim
-           << ")";
-  }
-  if (operandTy && dim >= rank)
-    return op.emitOpError() << "requires dimension attribute in range [0, "
-                            << rank << "); found (" << dim << ")";
+LogicalResult verifyDimInBounds(Location loc, ShapedType type, int64_t dim) {
+  if (dim < 0)
+    return emitOptionalError(
+        loc, "requires non-negative dimension attribute; found (", dim, ")");
+  if (type.hasRank() && dim >= type.getRank())
+    return emitOptionalError(loc, "requires dimension attribute in range [0, ",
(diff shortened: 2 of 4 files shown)


docs/spec.md
```diff
@@ -2716,8 +2716,7 @@ Produces the size of the given `dimension` of the `operand`.
 
 ### Constraints
 
-* (C1) 0 $\le$ `dimension` $\lt$ `rank(operand)`.
-  [todo](https://github.com/openxla/stablehlo/issues/790)
+  * (C1) 0 $\le$ `dimension` $\lt$ `rank(operand)`.
```
burmako, line 2719: Per https://github.com/openxla/stablehlo/blob/main/.markdownlint.yaml#L21, this shouldn't be indented.

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -4177,6 +4177,14 @@ func.func @get_dimension_size(%I: tensor<1x128x512xf32>) -> tensor<i32> {
 
 // -----
 
+func.func @get_dimension_size_invalid_dimension(%I: tensor<1x128x512xf32>) -> tensor<i32> {
+  // expected-error@+1 {{requires dimension attribute in range [0, 3); found (-1)}}
+  %size = "stablehlo.get_dimension_size"(%I) {dimension = -1 : i64} : (tensor<1x128x512xf32>) -> tensor<i32>
```
burmako, line 4182: Looks like `verifyDimAttr` accepts negative dimensions if the input is unranked. Let's fix that?

burmako (changes requested):

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -105,17 +105,20 @@ void createArgs(ArrayRef<OpAsmParser::UnresolvedOperand> operands,
...
   }
-
   int64_t dim = op.getDimension();
-  if (dim < 0 || dim >= rank)
+  if (dim < 0) {
+    return op.emitOpError()
+           << "requires non-negative dimension attribute; found (" << dim
+           << ")";
+  }
+  if (operandTy && dim >= rank)
```
burmako, line 121: From what I can see, this function is also used from SetDimensionSizeOp::verify, in which case we also want to somehow account for resultTy here.

You know, given that the contract of `verifyDimAttr` is not straightforward, I propose that we inline it into callers to make things easier to understand. To deduplicate, we could introduce the following function, with a straightforward contract, and use it from both places:

```C++
LogicalResult verifyDimInBounds(ShapedType type, int64_t dim) {
  // Check that dim >= 0
  // if hasRank(), check that dim < getRank()
}
```

burmako (changes requested):

stablehlo/dialect/StablehloOps.cpp
ghpvnist: Can you clarify? SetDimensionSizeOp::verify() takes care of the resultTy separately already. I think implementing something like this still has "duplication" whereas imo keeping the existing `verifyDimAttr` already factors the code well.
```
LogicalResult verifyDimInBounds(ShapedType type, int64_t dim, int64_t rank,
     mlir::InFlightDiagnostic emitOpError) {
  if (dim < 0) {
    return emitOpError
        << "requires non-negative dimension attribute; found (" << dim
        << ")";
  }
  if (type && dim >= rank)
    return emitOpError << "requires dimension attribute in range [0, "
                       << rank << "); found (" << dim << ")";
  return success();
}

...
LogicalResult GetDimensionSizeOp::verify() {
  // refactored code below
  int64_t rank = -1;
  auto operandTy = getOperand().getType().template dyn_cast<RankedTensorType>();
  if (operandTy) {
    rank = operandTy.getRank();
  } else if (auto ty = getType().template dyn_cast<RankedTensorType>()) {
    rank = ty.getRank();
  }
  int64_t dim = getDimension();
  return verifyDimInBounds(operandTy, dim, rank, emitOpError());
}

...
LogicalResult SetDimensionSizeOp::verify() {
  // already handles result type here
  if (auto size = this->getSize().getType().dyn_cast<RankedTensorType>()) {
    if (size.getRank() != 0)
      return emitOpError() << "size operand should be of rank-0";
  }
  // refactored code below
  int64_t rank = -1;
  auto operandTy = getOperand().getType().template dyn_cast<RankedTensorType>();
  if (operandTy) {
    rank = operandTy.getRank();
  } else if (auto ty = getType().template dyn_cast<RankedTensorType>()) {
    rank = ty.getRank();
  }
  int64_t dim = getDimension();
  return verifyDimInBounds(operandTy, dim, rank, emitOpError());
}
```

burmako: What I think can be improved about `verifyDimAttr` is how it tries to do multiple things at once - check operand type, check result type, and also mix that together instead of checking these types separately. As a result, when reading the code of this function, I am wondering which ops the function applies to, which of these ops have an operand type that needs to be checked, which of these ops have a result type that needs to be checked, and how they interact with each other. Lots of things to think about!

Here's the simplified code that I had in mind:
  1) I think that `verifyDimInBounds` (or something along these lines) is a more descriptive name that `verifyDimAttr`, so it looks like an improvement.
  2) Note that we don't need to verify the result type at all, because it's already checked through type inference. This leaves just the operand type. 
  3) Now, what do we do with the operand type? We could have a template parameter in `verifyDimInBounds`, pass `T op` into the function, and then do `op.getOperand().getType()` within the function aiming to reduce duplication.  That would establish an implicit contract for this function that the op that goes in must have `getOperand()` which I don't think is ideal from the readability point of view. Given that the amount of duplication that this saves is minor, I think it's better to have a more explicit contract.
  4) Same rationale for passing `dim` explicitly rather than doing `op.getDimension()` inside the function.

```C++
LogicalResult verifyDimInBounds(Location loc, ShapedType type, int64_t dim) {
  if (dim < 0) {
    return emitOptionalError(loc, 
        "requires non-negative dimension attribute; found (", dim, ")");
  }
  if (type.hasRank() && dim >= type.getRank())
    return emitOptionalError(loc, 
        "requires dimension attribute in range [0, ", rank, "); found (", 
        dim, ")");
  return success();
}

...
LogicalResult GetDimensionSizeOp::verify() {
  auto operandTy = getOperand().getType().cast<ShapedType>();
  int64_t dim = getDimension();
  return verifyDimInBounds(op.getLoc(), operandTy, dim);
}

...
LogicalResult SetDimensionSizeOp::verify() {
  if (auto size = this->getSize().getType().dyn_cast<RankedTensorType>()) {
    if (size.getRank() != 0)
      return emitOpError() << "size operand should be of rank-0";
  }
  auto operandTy = getOperand().getType().cast<ShapedType>();
  int64_t dim = getDimension();
  return verifyDimInBounds(op.getLoc(), operandTy, dim);
}
```

burmako (changes requested):

stablehlo/dialect/StablehloOps.cpp
ghpvnist: Thanks for your input! Some comments:

2. Right, we do not need verification for result type since the return type is always a scalar hence type inference is guaranteed to work as long as verify is successful (type inference runs after verify afaik).

One thing to note, `SetDimensionSizeOp` does not have a spec yet and marked `yes*` for verification and type inference. We'll need to revisit it, but in hindsight, it is not much different from GetDimensionSizeOp.

burmako: "One thing to note, SetDimensionSizeOp does not have a spec yet and marked yes* for verification and type inference". Good catch! That's intentional though - set_dimension_size has pretty tricky semantics which is tied to dynamism, so it won't be easy (or possible) to spec it without dynamism support in the spec. In comparison, get_dimension_size works for static shapes too.

burmako (approved):

PR #930, opened 2023-01-13
Handle bounds in `TriangularSolveOp` shape function

### LINKED ISSUE #807: Handle bounds in `TriangularSolveOp`shape function

ghpvnist:
The shape and bounds of batching dimensions is inferred using inferMergedDimAndBound.

If non-batching dimensions of both A and B are dynamic with bound, then the following inference rule applies. The shape is propagated the same way bounds are propagated. Arbitrary bounds are provided for A, B, and X to visually aid in how the they are moved:
```
A * X = B, R=rank

                | A   | X   | B   | X[R-2], X[R-1]
---------------------------------------------------------------
left_side=false | 3x3 | 3x5 | 3x5 | min(A[R-1], A[R-2]), B[R-1]

X * A = B, R=rank

                | X   | A   | B   | X[R-2], X[R-1]
---------------------------------------------------------------
left_side=true  | 2x3 | 3x3 | 2x3 | B[R-2], min(A[R-1], A[R-2])
```

closes #807

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -2818,13 +2818,12 @@ LogicalResult inferTransposeOp(std::optional<Location> loc, Value operand,
 
 LogicalResult inferTriangularSolveOp(
     std::optional<Location> location, Value a, Value b, bool leftSide,
-    bool isTransposeAInvalid, bool isTransposeOrAdjoint,
+    bool isTransposeAInvalid,
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
   // ODS enforces that a and b are of same element type: float or complex.
   auto elementType = a.getType().cast<ShapedType>().getElementType();
   auto aType = a.getType().dyn_cast<RankedTensorType>();
-  auto bType = b.getType().dyn_cast<RankedTensorType>();
-  if (!aType || !bType) {
+  if (!aType) {
     inferredReturnShapes.emplace_back(elementType);
     return success();
   }
@@ -2839,6 +2838,12 @@ LogicalResult inferTriangularSolveOp(
     return emitOptionalError(location,
                              "two minor dimensions of operand 'a' must ",
                              "be compatible, but got ", aType);
+
+  auto bType = b.getType().dyn_cast<RankedTensorType>();
+  if (!bType) {
+    inferredReturnShapes.emplace_back(elementType);
+    return success();
+  }
 
   auto bRank = bType.getRank();
   if (aRank != bRank)
@@ -2869,34 +2874,13 @@ LogicalResult inferTriangularSolveOp(
     return success();
   }
   auto resultShape = bType.getShape().vec();
-  auto resultBounds = encodingToBounds(aType.getEncoding()).vec();
+  auto resultBounds = encodingToBounds(bType.getEncoding()).vec();
   for (int64_t i = 0; i < aRank - 2; i++) {
     auto inferredDimAndBoundOrErr = inferMergedDimAndBound(
         location, i, aBatchDims[i], bBatchDims[i], aBounds[i], bBounds[i]);
     if (failed(inferredDimAndBoundOrErr)) return failure();
     resultShape[i] = (*inferredDimAndBoundOrErr).first;
     resultBounds[i] = (*inferredDimAndBoundOrErr).second;
-  }
-  if (leftSide) {
-    resultShape[bRank - 2] = bType.getShape()[bRank - 2];
-    resultBounds[bRank - 2] = bBounds[bRank - 2];
-    if (isTransposeOrAdjoint) {
-      resultShape[bRank - 1] = aType.getShape()[aRank - 1];
-      resultBounds[bRank - 1] = aBounds[aRank - 1];
-    } else {
-      resultShape[bRank - 1] = aType.getShape()[aRank - 2];
-      resultBounds[bRank - 1] = aBounds[aRank - 2];
-    }
-  } else {
-    resultShape[bRank - 1] = bType.getShape()[bRank - 1];
-    resultBounds[bRank - 1] = bBounds[bRank - 1];
-    if (isTransposeOrAdjoint) {
-      resultShape[bRank - 2] = aType.getShape()[aRank - 2];
-      resultBounds[bRank - 2] = aBounds[aRank - 2];
-    } else {
-      resultShape[bRank - 2] = aType.getShape()[aRank - 1];
-      resultBounds[bRank - 2] = aBounds[aRank - 1];
-    }
   }
   inferredReturnShapes.emplace_back(
       resultShape, bType.getElementType(),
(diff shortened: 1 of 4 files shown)


stablehlo/dialect/TypeInference.h
```diff
@@ -313,7 +313,7 @@ LogicalResult inferTransposeOp(Optional<Location> loc, Value operand,
 
 LogicalResult inferTriangularSolveOp(
     Optional<Location> location, Value a, Value b, bool leftSide,
-    bool isTransposeAInvalid,
+    bool isTransposeAInvalid, bool isTransposeOrAdjoint,
```
burmako, line 316: Let's split this into two flags `bool isTransposeATranspose` and `bool isTransposeAAdjoint`, to be consistent with `inferFftOp`.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -2708,12 +2708,13 @@ LogicalResult inferTransposeOp(Optional<Location> loc, Value operand,
...
     Optional<Location> location, Value a, Value b, bool leftSide,
-    bool isTransposeAInvalid,
+    bool isTransposeAInvalid, bool isTransposeOrAdjoint,
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
   // ODS enforces that a and b are of same element type: float or complex.
   auto elementType = a.getType().cast<ShapedType>().getElementType();
   auto aType = a.getType().dyn_cast<RankedTensorType>();
-  if (!aType) {
+  auto bType = b.getType().dyn_cast<RankedTensorType>();
+  if (!aType || !bType) {
```
burmako, line 2717: I don't think we should move the `bType` check here because then we'll regress verification for the case when `aType` is ranked but `bType` is unranked.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -2757,7 +2752,45 @@ LogicalResult inferTriangularSolveOp(
...
+    auto inferredDimAndBoundOrErr = inferMergedDimAndBound(
+        location, i, aBatchDims[i], bBatchDims[i], aBounds[i], bBounds[i]);
+    if (failed(inferredDimAndBoundOrErr)) return failure();
+    resultShape[i] = (*inferredDimAndBoundOrErr).first;
+    resultBounds[i] = (*inferredDimAndBoundOrErr).second;
+  }
+  if (leftSide) {
+    resultShape[bRank - 2] = bType.getShape()[bRank - 2];
+    resultBounds[bRank - 2] = bBounds[bRank - 2];
+    if (isTransposeOrAdjoint) {
```
burmako, line 2773: Why does it matter whether `isTransposeOrAdjoint` is true or false? This doesn't come up in [the spec](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#triangular_solve) or in [shape_inference.cc](https://github.com/tensorflow/tensorflow/blob/6a7da3957b5c2c35ad5b7b92e364068f9814235a/tensorflow/compiler/xla/service/shape_inference.cc#L2006).

stablehlo/dialect/TypeInference.cpp
```diff
@@ -2757,7 +2752,45 @@ LogicalResult inferTriangularSolveOp(
     return emitOptionalError(
         location, "Invalid transpose option value for triangular solve");
 
-  inferredReturnShapes.emplace_back(bType.cast<ShapedType>());
```
burmako, line 2760: Given that [the spec](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#triangular_solve) says that "(C4) `b` and `result` have the same type", would it be okay to just inherit bounds from `bType`? That would considerably simplify the resulting code.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -2757,7 +2752,45 @@ LogicalResult inferTriangularSolveOp(
...
+    resultShape[bRank - 1] = bType.getShape()[bRank - 1];
+    resultBounds[bRank - 1] = bBounds[bRank - 1];
+    if (isTransposeOrAdjoint) {
+      resultShape[bRank - 2] = aType.getShape()[aRank - 2];
+      resultBounds[bRank - 2] = aBounds[aRank - 2];
+    } else {
+      resultShape[bRank - 2] = aType.getShape()[aRank - 1];
+      resultBounds[bRank - 2] = aBounds[aRank - 1];
+    }
+  }
```
burmako, line 2790: Also, there's a considerable degree of duplication between the branches that looks like it can be relatively easily avoided. If we decide to keep this code (rather than a simpler "inherit bounds from `bType`" strategy), let's discuss how to improve this.

burmako (changes requested):

stablehlo/dialect/TypeInference.cpp
ghpvnist: It doesn't show up in shape_inference.cc and the spec since this change is specific to bounds. TriangularSolveOp (C3) requires the last two dimensions to be square, so it does not need any shape inference. However, TRANSPOSE or ADJOINT both require a transpose of the last two dimensions, and if the bounds differ, they need to be swapped. There are two options, keep the current PR's implementation or find the min bound and set both dimensions to it. I went with this approach since we have not decided on inferring the tightest bound, but ideally, we should just be setting the same minimum bounds.

stablehlo/dialect/TypeInference.cpp
ghpvnist: With further thinking, I agree that we could just infer the bound from `bType`.

stablehlo/dialect/TypeInference.cpp
ghpvnist: Let's inherit bounds from bType and leave more specific bound inference for future work.

stablehlo/dialect/TypeInference.h
ghpvnist: We won't need these flags in the latest commit

stablehlo/dialect/TypeInference.cpp
```diff
@@ -2789,7 +2789,24 @@ LogicalResult inferTriangularSolveOp(
     return emitOptionalError(
         location, "Invalid transpose option value for triangular solve");
 
-  inferredReturnShapes.emplace_back(bType.cast<ShapedType>());
+  auto aBounds = encodingToBounds(aType.getEncoding()).vec();
+  auto bBounds = encodingToBounds(bType.getEncoding()).vec();
```
burmako, line 2793: I don't think we need 2x `.vec()` here.

stablehlo/dialect/TypeInference.cpp
burmako: The code below doesn't seem to be doing that though. Did you mean to keep the `inferMergedDimAndBound` part? If yes, is it intended for it to not apply based on a/b's bounds even though it affects a/b's shapes as well?

burmako (changes requested):

stablehlo/dialect/TypeInference.cpp
ghpvnist: Thanks, I meant to remove it as well, fixed in the recent commit.

burmako (approved):

PR #933, opened 2023-01-13
Add new intro and build steps for readme

scottamain:
More to come...


README.md
```diff
@@ -1,28 +1,95 @@
...
+and compilers, even as StableHLO continues to evolve.
+
+This repository includes the [StableHLO
+specification](https://github.com/openxla/stablehlo/blob/main/docs/spec.md)
+along with implementations in C++ and Python, which you can use to build
+StableHLO programs.
+
+## Build steps
+
+  Here's how to build the StableHLO repo:
```
burmako, line 26: Do we need the indentation here?

README.md
```diff
@@ -1,28 +1,95 @@
...
+
+1. First clone this repo and the LLVM/MLIR git repository:
```
burmako, line 28: Let's simplify to "the LLVM repository", given that MLIR is part of LLVM?

README.md
```diff
@@ -1,28 +1,95 @@
...
+   ```
+
+2. Make sure you check out the correct commit in the LLVM repository:
+
+   ```sh
+   (cd llvm-project && git fetch && git checkout $(cat ../build_tools/llvm_version.txt))
+   ```
+
+   You need to do this every time you run git pull in the stablehlo repo
+   (in case `llvm_version.txt` changes).
```
burmako, line 45: Since we're here, how about we simplify this a bit to say "You need to do this every time `llvm_version.txt` changes"? This will avoid the need to talk about git pull. Also, llvm_version.txt can change due to other reasons.

README.md
```diff
@@ -1,28 +1,95 @@
...
+
+3. Configure and build LLVM/MLIR:
```
burmako, line 47: How about just "MLIR", given that the script is called "build_mlir.sh"?

README.md
```diff
@@ -1,28 +1,95 @@
...
+   ```sh
+   ninja check-stablehlo
+   ```
+
+   You should see results like this:
+
+   ```txt
+   Testing Time: 5.99s
+     Passed: 47
+   ```
```
burmako, line 80: Given that we now have build instructions in the readme, let's delete `build_tools/README.md` altogether. The non-Python part is covered by this file, and the Python part is an internal implementation detail and is captured in `build_tools/github_actions/ci_build_stablehlo_python_api.sh`.

burmako (changes requested):

README.md
```diff
@@ -1,28 +1,95 @@
...
+
+   ```sh
+   mkdir build && cd build
+
+   cmake .. -GNinja \
+     -DLLVM_ENABLE_LLD=ON \
+     -DCMAKE_BUILD_TYPE=Release \
+     -DLLVM_ENABLE_ASSERTIONS=On \
+     -DMLIR_DIR=${PWD}/../llvm-build/lib/cmake/mlir
+   ```
```
burmako, line 67: Can you also document your recent experience with `lld` - the problem that you ran into, how you fixed it, and how the initial application of the fix didn't work?

burmako (changes requested):

README.md
scottamain: Of course not. :) Sad that markdownlint doesn't have a rule for this (should be either 0 indent for a paragraph or 4 space indent for code block).

README.md
scottamain: Thanks

README.md
scottamain: SGTM

README.md
scottamain: SGTM

README.md
scottamain: Oops, meant to do that, thanks.

README.md
scottamain: Thank you, I meant to do that! I think it's easiest if users just be sure they have it installed before running any builds. So I've added the install command as the first step.

README.md
```diff
@@ -1,28 +1,103 @@
 # StableHLO
 
-StableHLO is an operation set that expresses ML computations. It has been
-originally bootstrapped from
-[the MHLO dialect](https://github.com/tensorflow/mlir-hlo#meta-hlo-dialect-mhlo)
-and enhances it with additional functionality, including serialization and
-versioning.
+StableHLO is an operation set for high-level operations (HLO) in machine
```
bhack, line 3: We don't mention MLIR but then we require MLIR in the build steps

README.md
```diff
@@ -1,28 +1,103 @@
...
+StableHLO is based on the MHLO dialect and enhances it with additional
+functionality, including serialization and versioning. This provides
+[backward and forward
+compatibility](https://github.com/openxla/stablehlo/blob/main/rfcs/20220912-compatibility.md)
+guarantees for StableHLO programs and ensures compatibility between frameworks
+and compilers, even as StableHLO continues to evolve.
+
+This repository includes the [StableHLO
+specification](https://github.com/openxla/stablehlo/blob/main/docs/spec.md)
+along with implementations in C++ and Python, which you can use to build
```
bhack, line 21: It is not clear what the input is, using the implementation, to produce a StableHLO programs.

README.md
scottamain: Okay, I added a mention for MLIR in the third paragraph.

README.md
scottamain: I'm not sure I understand. Are you suggesting that we explain here that the ML frameworks are the source for the ML programs and that's where these implementations are used? I think that would be overly specific and, IIRC, @burmako prefers that we not be prescriptive about who/what is actually building the StableHLO programs. But I might have misunderstood that. 

I revised this sentence slightly for clarity, but I don't think it addresses your concern. As is, this is intentionally ambiguous about who defines the actual ML program: This is just an API to build ML programs/graphs, so it can be used to write programs by hand or be implemented in another program (an ML framework) to generate the StableHLO output using some other input. I expect engineers will understand that, but if there's another way you think we can make that more clear, I'm happy to revise this.

bhack: Ps.
As we don't care to detail if are the Frameworks bridges developers to orchestrate these API or a developer directly to express his own ML program to obtain a StableHLO ML program output I think my previous example is general enough.

scottamain: I think I understand. I've now added that this StableHLO program can be fed to XLA/IREE, but still without specifying how the StableHLO is literally produced. Sound good?

README.md
```diff
@@ -1,6 +1,6 @@
 # StableHLO
 
-StableHLO is an operation set for high-level operations (HLO) in machine
+StableHLO is an operator set for high-level operations (HLO) in machine
```
burmako, line 3: The usual way of referring to StableHLO is "operation set". Both in [the spec](https://github.com/openxla/stablehlo/blob/main/docs/spec.md) and in [the MLIR language reference](https://mlir.llvm.org/docs/LangRef/#operations) that we're drawing inspiration from, ops are called operations, not operators. Let's use the more conventional terminology here as well?

burmako (changes requested):

README.md
scottamain: Oops. "Operator" ended up in our draft for the landing page and I thought that was from you so I copied it here. Reverted.

README.md
```diff
@@ -9,17 +9,17 @@ Our goal is to simplify and accelerate ML development by creating more
 interoperability between various ML frameworks (such as TensorFlow, JAX and
 PyTorch) and ML compilers (such as XLA and IREE).
 
-StableHLO is based on the MHLO dialect and enhances it with additional
-functionality, including serialization and versioning. This provides
-[backward and forward
+StableHLO is built using [MLIR](https://mlir.llvm.org/) and is based on the MHLO
+dialect but enhances it with additional functionality, including serialization
+and versioning. This provides [backward and forward
```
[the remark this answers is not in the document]
scottamain: SGTM

README.md
```diff
@@ -79,6 +84,9 @@ StableHLO programs.
      Passed: 47
    ```
 
+   This runs all the optimizer and interpreter tests in `stablehlo/tests/`.
```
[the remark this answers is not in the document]
scottamain: Done

README.md
```diff
@@ -9,17 +9,17 @@ Our goal is to simplify and accelerate ML development by creating more
...
+and versioning. This provides [backward and forward
 compatibility](https://github.com/openxla/stablehlo/blob/main/rfcs/20220912-compatibility.md)
 guarantees for StableHLO programs and ensures compatibility between frameworks
 and compilers, even as StableHLO continues to evolve.
 
 This repository includes the [StableHLO
 specification](https://github.com/openxla/stablehlo/blob/main/docs/spec.md)
-along with implementations in C++ and Python, which you can use to build
-StableHLO programs.
+along with op implementations in C++ and Python, which you can use to build ML
```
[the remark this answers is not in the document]
scottamain: Done, thanks

burmako:
Already getting mileage out of Markdown Lint! Nice :)

burmako (approved):

PR #935, opened 2023-01-16
Fix Reduce op bounds inference for scalar results

smit-hinsu:
encoding attribute is not required if the result is a scalar but the current code will end up incorrectly setting it to the input encoding attribute.

For the added test, the inferred result type was `tensor<f32, #stablehlo.type_extensions<bounds = [3]>>` but scalar tensors shouldn't have a bounds attributes. This was happening because boundsToEncoding returns the original encoding if the provided bounds are empty. boundsToEncoding doesn't differentiate between bounds being empty for scalars or bounds being empty for unbounded tensors.

I also checked that the other inference functions are not having a similar issue. (May have missed recently submitted shape functions.)

Reduce op inference also doesn't deal with sparse encoding attribute so we don't have to worry about that here.

cc @zhouxin913

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -452,7 +452,10 @@ LogicalResult verifyReduceOpInputsAndInferShape(
         }
       }
     }
-    if (!inputBounds.empty()) {
+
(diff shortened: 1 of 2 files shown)


smit-hinsu:
Gentle ping!

burmako (approved):

PR #940, opened 2023-01-18
Mention that mardkownlint does not check ol indentation

scottamain:


sdasgup3 (approved):
Thanks for accommodating this information.

PR #941, opened 2023-01-18
Add github action to run markdownlint on pull requests

scottamain:
This will resolve issue #792


GleasonK (changes requested):
Two notes:

1. Could we add this to our existing `lint.yml`? Would just need to past in there and remove the `paths-ignore` line I believe. Otherwise implementation of the workflow is elegant, let me know if there is a reason to keep this a separate workflow.

2. No need to take action on this, but if the `./**/*.md` directory traversal turns out to be costly, we could likely do something similar to: [lint_clang_format.sh#L38](https://github.com/openxla/stablehlo/blob/main/build_tools/github_actions/lint_clang_format.sh#L38) to gather file diff. That said, this repo is still small, and whitespace checks is already doing a full traversal with `find`.

scottamain:
Thanks Kevin!

1) It did cross my mind to add this to lint.yml but I figured it was more appropriate to run this task only when the PR includes MD files, and my understanding is that you can specify the "on" triggers on a per-file basis. (In an abundance of caution about variation in dependencies, I thought it would be best that this script not run in the same virtual environment as the other lint jobs. So then this could be a separate job in the same workflow, but then it seems like wasteful compute cost to spin up a new virtual environment for PRs that don't even include MD files.) Let me know if you think otherwise.

2) It didn't occur to me that this globbing would be a latency bottleneck but I haven't noticed any consequences that would warrant investigation. If you have suspicions about it, then it might be worth a look as a followup to this PR, but so far I think it performs at a reasonable speed.

GleasonK (approved):
This LGTM. We may do some reorganizing of lint actions in the near future. I think the "paths-ignore" and "paths" that I pointed out cause some merge issues (for PR's that don't run every check, Eugene has to override the merge).

That work is on my schedule though and should not block this submit in any way.

PR #942, opened 2023-01-19
Verify that start_indices have the same type in DynamicSliceOp

### LINKED ISSUE #557: Verify that start_indices have the same type in DynamicSliceOp
~Current verification does not check that `start_indices` is in bounds and whether `start_indices` combined with `slice_sizes` is also in range. See constraints in #556.~

~As commented below, the element type of individual `start_indices` should be `i64`.~

`start_indices` should all have the same type since during discussion below, it's decided that they should be integer type. Verification is set to `revisit` in #556.

ghpvnist:
`inferDynamicUpdateSliceOp` also has the same verification logic, so I've moved the duplicate logic to `variadicTensorsHaveSameElType` function.
closes #557

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -87,6 +87,24 @@ bool tensorsHaveSameElType(Type type1, Type type2, bool ignoreFpPrecision) {
   return tensorTy1.getElementType() == tensorTy2.getElementType();
 }
 
+LogicalResult variadicTensorsHaveSameElType(ValueRange tensors,
+                                            std::optional<Location> location) {
+  if (!tensors.empty()) {
+    auto tensorTy1 = tensors[0].getType().cast<ShapedType>();
+    Type tensorEl1 = tensorTy1.getElementType();
+    for (auto otherTensor : llvm::drop_begin(tensors, 1)) {
+      auto tensorTy2 = otherTensor.getType().cast<ShapedType>();
+      Type tensorEl2 = tensorTy2.getElementType();
+      if (tensorEl1 != tensorEl2)
+        return emitOptionalError(
+            location,
+            "start indices must have same element type (encountered mismatch: ",
+            tensorEl1, " vs ", tensorEl2, ")");
+    }
+  }
+  return success();
+}
+
 // Return true if type1 and type2 are shape-compatible and have same element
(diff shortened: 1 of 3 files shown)


ghpvnist:
The function `tensorsHaveSameElType` right above `variadicTensorsHaveSameElType` is very similar in terms of functionality. We can consider to merge these functions together which requires some thinking of how to handle variadic tensors vs fixed-length inputs to the unified function.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -87,6 +87,24 @@ bool tensorsHaveSameElType(Type type1, Type type2, bool ignoreFpPrecision) {
   return tensorTy1.getElementType() == tensorTy2.getElementType();
 }
 
+LogicalResult variadicTensorsHaveSameElType(ValueRange tensors,
+                                            std::optional<Location> location) {
```
burmako, line 91: Agreed that we need to deduplicate `tensorsHaveSameElType` and `variadicTensorsHaveSameElType`. My recommendation would be along these lines:

```C++
bool tensorsHaveSameElType(Type type1, Type type2, bool ignoreFpPrecision = true) { 
  return tensorsHaveSameElType({type1, type2}, ignoreFpPrecision);
}

bool tensorsHaveSameElType(TypeRange types, bool ignoreFpPrecision = true) { 
  // Don't emit emitOptionalError here, do that at callsites.
  // We won't be able to provide a super precise error message (that includes
  // what types mismatch), and that's fine.
  ... 
}
```

burmako (changes requested):

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1972,6 +1981,14 @@ LogicalResult inferDynamicSliceOp(
         location, "has mismatched number of start indices (", numStartIndices,
         ") and the rank of operand (", operandType.getRank(), ")");
 
+  // (C3)
+  SmallVector<Type> startIndicesTypes;
+  for (auto index : startIndices) startIndicesTypes.push_back(index.getType());
+  if (!tensorsHaveSameElType(startIndicesTypes))
```
burmako, line 1987: Can we simplify this to `startIndices.getTypes()`? (Same below).

burmako (changes requested):

burmako (approved):

PR #949, opened 2023-01-20
Fix tblgen rules in CMake after the VHLO reorganization

d0k:


atondwal:
I'll test this against the next llvm integrate and then merge

burmako:
Force merging after both Anish and I tested this PR locally.

atondwal:
Thanks for the fix!

Issue #611, opened 2022-11-25
Standardize summaries


burmako:
We've got `let description` in the ODS under control thanks to [the checklist](https://github.com/openxla/stablehlo/blob/main/docs/spec_checklist.md), but `let summary` is still inconsistent across ops.

sdasgup3:
For any operation definition 
`def_StableHLO_XYZOp `, let us use the summary as 
```
let summary = "XYZ operator";
```

Note: 
1. Consistent across ops definitions
2. As the description is already load bearing, we can avoid re-inventing summary text for each op.  
3. Most of the ops definitions are already following the norm.
4. Some ops like rng deviates from the norm. 
```
let summary = "RNG operator with uniform or normal distribution.".
```

In such cases, we can choose to follow (1) OR make exceptions if the summary provides information which is (A) correct, and (B) is not repeated verbatim in description including the link to spec.

I vote for (1) as that way there less maintenance: One less thing to worry if op semantics changes.

burmako:
LGTM, except that "operator" isn't currently a thing - neither in [the spec](https://github.com/openxla/stablehlo/blob/main/docs/spec.md) nor in [MLIR Language Reference](https://mlir.llvm.org/docs/LangRef/) which we use for inspiration for many things. How about "XYZ operation"?

sdasgup3:
makes sense.
LGTM

burmako:
Assigned this ticket to myself by mistake. It will be fixed anyway over time via https://github.com/openxla/stablehlo/blob/main/docs/reference_checklist.md.

PR #123, opened 2022-09-15
Reorder interpret_add.mlir tests and fix typos in spec_draft.md

ghpvnist:

PR #951, opened 2023-01-20
Interpreter implementation and review guidelines

sdasgup3:
Provides a doc to be followed while introducing the interpreter implementation of an op. 

Added @GleasonK.
Can you please take a look the the item related to running the `ccov`.

### DIFF
--- a/docs/interpreter_checklist.md
+++ b/docs/interpreter_checklist.md
@@ -7,27 +7,26 @@ those fronts alongside the interpreter implementation.
 
 1. While implementing the interpreter for an op, double check with related
    implementations (e.g.
-   <!-- markdownlint-disable line-length -->
    [hlo_interpreter](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/hlo/evaluator)
-   <!-- markdownlint-enable line-length -->
    and
-   <!-- markdownlint-disable line-length -->
    [mlir_interpreter](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/compiler/xla/mlir_hlo/tools/mlir_interpreter)).
-   <!-- markdownlint-enable line-length -->
    This action item will allow mutual improvements.
+1. Make sure that the ODS follows standard [summary](https://github.com/openxla/stablehlo/issues/611).
 1. Replace the comments, related to input/output constraints, in
    the implementations, specifically for verifiers and type inference, with
-   "constraint-labels" (e.g. `Cn` in
-   <!-- markdownlint-disable line-length -->
-   [here](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#constraints-2) )
-   <!-- markdownlint-enable line-length -->
-   used in the
-   <!-- markdownlint-disable line-length -->
+   some unique identifier containing "constraint-labels", e.g. `Cn` in
+   [here](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#constraints-2)
+   , used in the
    [specification](https://github.com/openxla/stablehlo/blob/main/docs/spec.md).
-   <!-- markdownlint-enable line-length -->
+   The unique identifier should have the format `<op_name>_Cn`.
    This is to explicitly align all the implementations with the specification:
    the single source of truth, and help to track gaps in those implementations
    back to the spec.
+
+   **Why Unique**: The fact that the spec will be augmented with Dynamism,
+   Quantization can affect the suffix number in the constraint label or might
+   change it to something else. A unique identifier will enable faster updates
+   to comments we are planning to add here.
    1. It is OK to have a comment with multiple constraint-labels or to have
       multiple comments with the same constraint-label. It all depends on how
       the constraint-label(s) are coded in the implementations.
@@ -37,19 +36,32 @@ those fronts alongside the interpreter implementation.
 1. Make sure that the there are sufficient tests to cover for the
    implementations (e.g. verifiers, type inference, interpreter) of the
    specification. Specifically,
-   1. Write tests for an op interpreter following the
-      <!-- markdownlint-disable line-length -->
-      [testing guidelines](https://github.com/openxla/stablehlo/blob/main/docs/reference.md#testing-guidelines).
-      <!-- markdownlint-enable line-length -->
-   1. Double check that there is at least one test for each constraint in
-      verifier and type inference methods.
+   1. Regarding interpreter tests
+      1. For interpreter: Write tests following the
+         [testing guidelines](https://github.com/openxla/stablehlo/blob/main/docs/reference.md#testing-guidelines).
+      1. Add a link of the test file in the spec (e.g.
+         [More Examples](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#add)).
+   1. For verifier and type inference: Affecting [this file](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/ops_stablehlo.mlir).
+      1. Make sure that there is at least one test for each constraint in
+         verifier and type inference methods. These tests will mostly be
+         negative tests exercising the fact that the constraints are not met.
+         Make sure to add at least one positive tests in case there are none.
+         Note that it is out of scope to revisit the correctness of type
+         inference as in
+         [file](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/infer_stablehlo.mlir).
+      1. Make sure that all the tests related to the op under test, are placed
+         together.
+      1. Make sure that all the tests related to the op under test, are
+         prepended with `CHECK-LABEL` lit macro.
+      1. Chose the function name of the tests using the format
+         `op_name_Cn1_Cn2_...` for a function testing constraints `Cn1`, `Cn2`
+         for op `op_name`. In cases when the proposed format does not apply,
+         keep the existing name.
 1. Locally run the
-   <!-- markdownlint-disable line-length -->
    [ccov](https://github.com/openxla/stablehlo/blob/main/build_tools/github_actions/ci_build_stablehlo_code_coverage.sh)
-   <!-- markdownlint-enable line-length -->
    tool to make sure that the followings yield >= 90% coverage. If not, then add
    tests to achieve the goal.
    1. Code and tests introduced for implementing the interpreter.
    1. The existing code and tests for verifier and type inference.
-1. Cross validate the test results of the interpreter implementation against
-   hardware. TBD
+1. Once all the items are finsihed, update the interpreter column in status.md
+   to `yes`.
--- a/docs/interpreter_checklist.md
+++ b/docs/interpreter_checklist.md
@@ -25,7 +25,6 @@ After implementing the interpreter:
    and [StablehloOps.cpp](https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/StablehloOps.cpp):
     1. Delete comments that say things like "Verify the following properties:
        ...".
-
     1. Add comments that reference constraint labels from the spec in the format
        `<op_name>_Cn`, e.g. `SliceOp_C1`, to identify which parts of verifiers
        and shape functions correspond to which constraints in the specification.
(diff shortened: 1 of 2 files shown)


docs/interpreter_checklist.md
```diff
@@ -0,0 +1,55 @@
...
+      [testing guidelines](https://github.com/openxla/stablehlo/blob/main/docs/reference.md#testing-guidelines).
+      <!-- markdownlint-enable line-length -->
+   1. Double check that there is at least one test for each constraint in
+      verifier and type inference methods.
+1. Locally run the
+   <!-- markdownlint-disable line-length -->
+   [ccov](https://github.com/openxla/stablehlo/blob/main/build_tools/github_actions/ci_build_stablehlo_code_coverage.sh)
+   <!-- markdownlint-enable line-length -->
+   tool to make sure that the followings yield a 100% coverage. If not, then add
+   tests to achieve the goal.
```
GleasonK, line 51: There are times where 100% isn't feasible, i.e. lines with asserts that should never be hit, switch statements with default values that should never be hit. I'm not sure how much this is the case with interpreter, but it could be worth adding the clause _"... 100% coverage, except for lines with unreachable assertions."_

For reference our code coverage goals for the StableHLO repo were 90%. I'd say 95% is likely feasible as well, anything higher will likely have edge cases.

sdasgup3: That is a very good catch! 
Thanks for correction.

sdasgup3: settle on `>= 90%`

docs/interpreter_checklist.md
```diff
@@ -0,0 +1,55 @@
...
+   back to the spec.
+   1. It is OK to have a comment with multiple constraint-labels or to have
+      multiple comments with the same constraint-label. It all depends on how
+      the constraint-label(s) are coded in the implementations.
+   1. In case there is a mismatch between the constraints in the
+      implementation VS those in the specification, make sure there is an
+      open issue reflecting that discrepancy.
+1. Make sure that the there are sufficient tests to cover for the
+   implementations (e.g. verifiers, type inference, interpreter) of the
+   specification. Specifically,
```
zhouxin913, line 39: Are you suggesting that for positive cases we add a  interpreter test, and for negative cases (all kinds of failures) we add/check tests in verifiers/type_inferences test files?

sdasgup3: Thanks for the input!
I revised the text with additional information. PTAL.

GleasonK (approved):
LGTM, reviewed the coverage portion.

docs/interpreter_checklist.md
```diff
@@ -0,0 +1,60 @@
...
+      <!-- markdownlint-disable line-length -->
+      [testing guidelines](https://github.com/openxla/stablehlo/blob/main/docs/reference.md#testing-guidelines).
+      <!-- markdownlint-enable line-length -->
+   1. For verifier and type inference: Make sure that there is at least one test
+      for each constraint in verifier and type inference methods. These tests
+      will mostly be negative tests exercising the fact that the constraints are
+      not met. Make sure to add at least one positive tests in case there are
+      none. Note that the action item here is not related to tests for checking
+      if the types are correctly inferred or not, which we believe is already
+      done with sufficient thoroughness.
```
subhankarshah, line 50: * Group tests by ops alphabetically and before every op add a comment section with the ops name?
  * Within a single op, order the tests in some manner like positive test followed by negative tests?
  * Maybe write the constraints `Cn` in comments in tests for testing constraints?

This will make the tests also organised, making sure all the cases are extensibly and exhaustively covered and tie it to the spec.

subhankarshah: @sdasgup3

subhankarshah (changes requested):
Added a minor comment.

docs/interpreter_checklist.md
subhankarshah: #123

sdasgup3: Nice suggestion!
I agree that there is a lot opportunity to to sanitize https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/ops_stablehlo.mlir and I have opened a [ticket](https://github.com/openxla/stablehlo/issues/993) enumerating some of my thoughts on (1) and related items. 

The action items, related to testing verifier and type inferencers, are a few auxiliary things that we want to accomplish at a per op level.  Given that the scope of (1) is beyond "per op" level, we can address this using the above ticket. For other I have updated the guidelines. PTAL

docs/interpreter_checklist.md
```diff
@@ -0,0 +1,63 @@
...
+   1. It is OK to have a comment with multiple constraint-labels or to have
+      multiple comments with the same constraint-label. It all depends on how
+      the constraint-label(s) are coded in the implementations.
+   1. In case there is a mismatch between the constraints in the
+      implementation VS those in the specification, make sure there is an
+      open issue reflecting that discrepancy.
+1. Make sure that the there are sufficient tests to cover for the
+   implementations (e.g. verifiers, type inference, interpreter) of the
+   specification. Specifically,
+   1. For interpreter: Write tests following the
```
ghpvnist, line 38: Can you also add a section to:
1. Make sure the spec only has 1 example
2. The spec example is the same as what is in the ODS (it may not be the same if the spec has > 1 examples)
3. Add a link for "More Examples" in the spec to the interpreter tests
4. Update the interpreter column in status.md to `yes`
5. Default the spec test to be the first test in the `interpreter_foo.mlir`
6. Make sure to use the actual name of the inputs/outputs instead of defaulting to SSA values (e.g. %0, %1, etc.)

ghpvnist (changes requested):

docs/interpreter_checklist.md
ghpvnist: We could also add another action item to standardize summary in the ODS to address #611

sdasgup3: Thanks!
I have included all the comments except for
1 and 2: IIRC as part of spec work we already did that 
5. The tests in the spec are not written following the testing guidelines. I am afraid that they can deviate. Plus, IMO with "More Examples" we are expected to add new information.  

Please let me know your thoughts

ghpvnist: Actually, the spec work items were introduced after we started the spec work, so we still may have (I haven't gone through all the ops) ops with more than one example (which is the case in https://github.com/openxla/stablehlo/pull/996/files). By removing the example which is also in ODS, they also have to be changed.

Yes, I also realized this, but I also think it is good to include it to:
1. Make sure the spec example test we wrote is working (aka no typos in the spec example)
2. Since the spec example already covers most, if not all, the interpreter nuances (for example adjusted start index for DynamicUpdateSliceOp), creating another similar test and excluding the spec example in the test is a bit odd imo. And I agree, we should still be adding new tests.

The decision is still up for discussion though. Lmk your thoughts.

docs/interpreter_checklist.md
```diff
@@ -0,0 +1,67 @@
...
+related to verifier and type inference, with the idea of making progress in
+those fronts alongside the interpreter implementation.
+
+1. While implementing the interpreter for an op, double check with related
+   implementations (e.g.
+   [hlo_interpreter](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/hlo/evaluator)
+   and
+   [mlir_interpreter](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/compiler/xla/mlir_hlo/tools/mlir_interpreter)).
+   This action item will allow mutual improvements.
+1. Make sure that the ODS follows standard [summary](https://github.com/openxla/stablehlo/issues/611).
```
burmako, line 14: What is the standard summary? Let's write this down in the ticket.

docs/interpreter_checklist.md
```diff
@@ -0,0 +1,67 @@
...
+         Make sure to add at least one positive tests in case there are none.
+         Note that it is out of scope to revisit the correctness of type
+         inference as in
+         [file](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/infer_stablehlo.mlir).
+      1. Make sure that all the tests related to the op under test, are placed
+         together.
+      1. Make sure that all the tests related to the op under test, are
+         prepended with `CHECK-LABEL` lit macro.
+      1. Chose the function name of the tests using the format
+         `op_name_Cn1_Cn2_...` for a function testing constraints `Cn1`, `Cn2`
```
burmako, line 57: op_mnemonic? Same below.

docs/interpreter_checklist.md
```diff
@@ -0,0 +1,67 @@
...
+         for op `op_name`. In cases when the proposed format does not apply,
+         keep the existing name.
+1. Locally run the
+   [ccov](https://github.com/openxla/stablehlo/blob/main/build_tools/github_actions/ci_build_stablehlo_code_coverage.sh)
+   tool to make sure that the followings yield >= 90% coverage. If not, then add
+   tests to achieve the goal.
+   1. Code and tests introduced for implementing the interpreter.
+   1. The existing code and tests for verifier and type inference.
+1. Once all the items are finsihed, update the interpreter column in status.md
+   to `yes`.
```
burmako, line 67: This is a good checklist, but I think it can be sharpened up to make instructions more straightforward, similarly to the spec checklist. Here's how I would've written it:

```
While implementing the interpreter:

1. Consult [hlo_evaluator](link) and [mlir_interpreter](link) to identify tricky
   implementation details and potential functionality gaps.

2. File tickets for the corresponding software components if you find any bugs
   or missing functionality.

After implementing the interpreter:

1. In [StablehloOps.td](link):
  * Make sure that the `summary` in op's ODS follows the standard format.

2. In [TypeInference.cpp](link):
  * Delete comments that say things like "Verify the following properties: ...".
  * Add comments that reference constraint labels from the spec in the format
    `<op_name>_Cn`, e.g. `SliceOp_C1` to identify which parts of verifiers and 
    shape functions correspond to which constraints in the specification.
      * It is OK to have a comment with multiple constraint labels or to have 
        multiple comments with the same constraint label. It all depends on how 
        the constraints are implemented.
      * In case there is a mismatch between the constraints in the 
        implementation VS those in the specification, make sure there is an open 
        issue reflecting that discrepancy.

3. In [interpreter tests](link to the directory):
  * Add file called `interpret_<op_mnemonic>.mlir`.
  * Write tests following the [testing guidelines](link).

4. In [ops_stablehlo.mlir](link):
  * Make sure that there is at least one test for each constraint...
  * Make sure that all the tests related to the op under test...
  * Make sure that all the tests related to the op under test...
  * Chose the function name of the tests using the format...
  * Keep adding tests until the [ccov](link) shows >= 90% coverage for the op.

5. In [spec.md](link):
  * Add a link to `interpret_<op_mnemonic>.mlir` to the "Examples" section.

6. In [status.md](link):
  * Update the "Interpreter" column to `yes`.
============

File: docs/interpreter_checklist.md [hunk 8]
```diff
@@ -0,0 +1,67 @@
+# StableHLO Interpreter Checklist
+
+In this document, we summarize the guidelines while implementing and reviewing
+interpreter for an op. We have intentionally tied a few auxiliary action items,
+related to verifier and type inference, with the idea of making progress in
+those fronts alongside the interpreter implementation.
```
burmako, line 6: What about ops that have already been implemented?

burmako (changes requested):

docs/interpreter_checklist.md
ghpvnist: 4. Should we need to add test for constraints covered by the ODS? What do we do with tests covered in Types section (e.g. 1-dimensional tensor... and the test "x should be rank 1...") Should we add them (if we don't have it and not covered by ODS)?

docs/interpreter_checklist.md
```diff
@@ -0,0 +1,67 @@
...
+         inference as in
+         [file](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/infer_stablehlo.mlir).
+      1. Make sure that all the tests related to the op under test, are placed
+         together.
+      1. Make sure that all the tests related to the op under test, are
+         prepended with `CHECK-LABEL` lit macro.
+      1. Chose the function name of the tests using the format
+         `op_name_Cn1_Cn2_...` for a function testing constraints `Cn1`, `Cn2`
+         for op `op_name`. In cases when the proposed format does not apply,
+         keep the existing name.
```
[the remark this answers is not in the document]
sdasgup3: Added.

docs/interpreter_checklist.md
sdasgup3: All the ops which has the interpreter will be assigned (to me) and checked for compliance with this checklist.

docs/interpreter_checklist.md
sdasgup3: That is right.  we should add tests only if if we don't have it and not covered by ODS. IMO, It does not worth testing semantics of various mlir traits.

docs/interpreter_checklist.md
sdasgup3: updated the ticket.

docs/interpreter_checklist.md
sdasgup3: Here is the ticket for the assignment of those ops
 https://github.com/openxla/stablehlo/issues/1002

docs/interpreter_checklist.md
burmako: Agreed.

burmako (approved):

zhouxin913 (approved):

subhankarshah (approved):
LGTM

ghpvnist (approved):

PR #953, opened 2023-01-20
Make build instructions repeatable

atondwal:
use `mkdir -p`, so that it doesn't fail if the dir already exists


burmako (commented):
LGTM

Issue #583, opened 2022-11-23
Remove `COMPARISON_TYPE_NOTYPE` enum from CompareOp


burmako:
Similarly to #534, due to similar reasons - this value is used as a sentinel in the HLO protobuf schema where it's not possible to specify that it's mandatory to provide a value for the attribute. This is not a thing in StableHLO, so this value is not needed in StableHLO.

PR #958, opened 2023-01-23
Verify compare_type for compare_op.

### LINKED ISSUE #571: Verify comparison_direction and compare_type for CompareOp
Disallow illegal `compare_type` for all element types.

subhankarshah:
closes #571

### DIFF
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -2305,6 +2305,32 @@ LogicalResult CompareOp::inferReturnTypeComponents(
     ValueShapeRange operands, DictionaryAttr attributes, RegionRange regions,
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
   CompareOp::Adaptor adaptor(operands, attributes, regions);
+  if (adaptor.getCompareType().has_value() &&
+      adaptor.getCompareType().value() != ComparisonType::NOTYPE) {
+    auto elementType =
+        adaptor.getLhs().getType().cast<ShapedType>().getElementType();
+    ComparisonType compareType = adaptor.getCompareType().value();
+    if (elementType.isa<FloatType>() && compareType != ComparisonType::FLOAT &&
+        compareType != ComparisonType::TOTALORDER)
+      return emitOptionalError(
+          location, "compareType must be FLOAT or TOTALORDER, but got ",
+          stringifyEnum(compareType));
+    else if (elementType.isa<ComplexType>() &&
+             compareType != ComparisonType::FLOAT)
+      return emitOptionalError(location, "compareType must be FLOAT, but got ",
+                               stringifyEnum(compareType));
+    else if (elementType.isa<IntegerType>() &&
+             elementType.isUnsignedInteger() &&
+             compareType != ComparisonType::UNSIGNED)
+      return emitOptionalError(location,
+                               "compareType must be UNSIGNED, but got ",
+                               stringifyEnum(compareType));
+    else if (elementType.isa<IntegerType>() &&
+             elementType.isSignlessInteger() &&
+             compareType != ComparisonType::SIGNED)
+      return emitOptionalError(location, "compareType must be SIGNED, but got ",
+                               stringifyEnum(compareType));
+  }
   return hlo::inferCompareOp(context, location, adaptor.getLhs(),
                              inferredReturnShapes);
 }
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -2305,34 +2305,14 @@ LogicalResult CompareOp::inferReturnTypeComponents(
     ValueShapeRange operands, DictionaryAttr attributes, RegionRange regions,
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
   CompareOp::Adaptor adaptor(operands, attributes, regions);
-  if (adaptor.getCompareType().has_value() &&
-      adaptor.getCompareType().value() != ComparisonType::NOTYPE) {
-    auto elementType =
-        adaptor.getLhs().getType().cast<ShapedType>().getElementType();
-    ComparisonType compareType = adaptor.getCompareType().value();
-    if (elementType.isa<FloatType>() && compareType != ComparisonType::FLOAT &&
-        compareType != ComparisonType::TOTALORDER)
-      return emitOptionalError(
-          location, "compareType must be FLOAT or TOTALORDER, but got ",
-          stringifyEnum(compareType));
-    else if (elementType.isa<ComplexType>() &&
-             compareType != ComparisonType::FLOAT)
-      return emitOptionalError(location, "compareType must be FLOAT, but got ",
-                               stringifyEnum(compareType));
-    else if (elementType.isa<IntegerType>() &&
-             elementType.isUnsignedInteger() &&
-             compareType != ComparisonType::UNSIGNED)
-      return emitOptionalError(location,
-                               "compareType must be UNSIGNED, but got ",
-                               stringifyEnum(compareType));
-    else if (elementType.isa<IntegerType>() &&
-             elementType.isSignlessInteger() &&
-             compareType != ComparisonType::SIGNED)
-      return emitOptionalError(location, "compareType must be SIGNED, but got ",
-                               stringifyEnum(compareType));
-  }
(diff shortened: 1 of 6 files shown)


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2305,6 +2305,32 @@ LogicalResult CompareOp::inferReturnTypeComponents(
...
+      return emitOptionalError(location,
+                               "compareType must be UNSIGNED, but got ",
+                               stringifyEnum(compareType));
+    else if (elementType.isa<IntegerType>() &&
+             elementType.isSignlessInteger() &&
+             compareType != ComparisonType::SIGNED)
+      return emitOptionalError(location, "compareType must be SIGNED, but got ",
+                               stringifyEnum(compareType));
+  }
   return hlo::inferCompareOp(context, location, adaptor.getLhs(),
```
subhankarshah, line 2334: @ghpvnist As of now I did not add the flags isFloat, isComplex, isUnsigned, isSigned because they depend on the values taken by the ComparisonType enum and we would have to makes changes if the enum were to change (which is unlikely though) but it does not seem clean to have parameters tied to values of an enum like that.

subhankarshah: Other option is sending a lowercase string, which I personally find cleaner IMO.

ghpvnist: Since we are using boolean flags, it shouldn't matter since they are taken care of in StablehloOps.cpp. The enum values shouldn't change, and if so this part of code should be fixed regardless of the approach. PS. I think the code is more readable refactored this way, at the cost of less precise error messages (but the error should still be obvious since we are talking about an enum value):
```C++
// StablehloOps.cpp
// ...
return hlo::inferCompareOp(context, location,
                      compareType.has_value(),
                      compareType == ComparisonType::NOTYPE,
                      compareType == ComparisonType::FLOAT,
                      compareType == ComparisonType::TOTALORDER,
                      compareType == ComparisonType::UNSIGNED,
                      compareType == ComparisonType::SIGNED,
                      adaptor.getLhs());

// TypeInference.cpp
LogicalResult inferCompareOp(
    MLIRContext* context, std::optional<Location>, bool compareTypeHasValue,
    bool isNoType, bool isFloat, bool isTotalOrder, bool isUnsigned, bool isSigned,
    Value lhs, SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
  // ...
  if (compareTypeHasValue && !isNoType) {
    auto elementType = lhs.getType().cast<ShapedType>().getElementType();
    if (elementType.isa<FloatType>() && !isFloat && !isTotalOrder)
                               "compareType must be FLOAT or TOTALORDER");
    else if (elementType.isa<ComplexType>() && !isFloat)
      return emitOptionalError(location, "compareType must be FLOAT");
    else if (elementType.isa<IntegerType>() &&
             elementType.isUnsignedInteger() && !isUnsigned)
      return emitOptionalError(location, "compareType must be UNSIGNED");
    else if (elementType.isa<IntegerType>() &&
             elementType.isSignlessInteger() && !isSigned)
      return emitOptionalError(location, "compareType must be SIGNED");
  }
  // ...
  ```

Comment (subhankarshah) [reply stablehlo/dialect/StablehloOps.cpp:2334]: Made the change but keeping this open for now.

**Review by ghpvnist — APPROVED** [review]

File: stablehlo/dialect/TypeInference.h [hunk 2]
```diff
@@ -147,7 +147,9 @@ LogicalResult inferClampOp(
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes);
 
 LogicalResult inferCompareOp(
-    MLIRContext* context, std::optional<Location>, Value lhs,
+    MLIRContext* context, std::optional<Location>, bool compareTypeHasValue,
+    bool isNoType, bool isFloat, bool isTotalOrder, bool isUnsigned,
```
burmako, line 151: Other similar flags in this file are called `isFieldNameFieldValue`, e.g. `isCompareTypeNoType` etc. Let's rename accordingly?

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -1193,6 +1193,17 @@ func.func @comp_eq(%arg0: tensor<3xi32>, %arg1: tensor<3xi32>) -> tensor<3xi1> {
...
 // -----
 
+func.func @comp_bad_type(%arg0: tensor<3xi32>, %arg1: tensor<3xi32>) -> tensor<3xi1> {
+  // expected-error@+1 {{compareType must be SIGNED}}
+  %0 = "stablehlo.compare"(%arg0, %arg1) {
+    comparison_direction = #stablehlo<comparison_direction EQ>,
+    compare_type = #stablehlo<comparison_type TOTALORDER>
+  } : (tensor<3xi32>, tensor<3xi32>) -> tensor<3xi1>
+  func.return %0 : tensor<3xi1>
+}
```
burmako, line 1203: Let's have negative tests for all error messages produced by inferCompareOp.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1547,8 +1547,22 @@ LogicalResult inferClampOp(
 }
 
 LogicalResult inferCompareOp(
-    MLIRContext* context, std::optional<Location>, Value lhs,
+    MLIRContext* context, std::optional<Location> location,
+    bool compareTypeHasValue, bool isNoType, bool isFloat, bool isTotalOrder,
+    bool isUnsigned, bool isSigned, Value lhs,
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
+  if (compareTypeHasValue && !isNoType) {
```
burmako, line 1554: I don't think we should allow `ComparisonType::NOTYPE` (see #583 for rationale). How about we unconditionally emit an error when encountering this value of `compare_type`?

burmako (changes requested):

stablehlo/dialect/TypeInference.cpp
subhankarshah: There might be usages of NOTYPE to denote default which is same as leaving it empty also I had to remove a test from legalize-stablehlo-to-vhlo as stablehlo no longer allows NOTYPE. Why emit an error when we can remove it completely?

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1547,8 +1547,27 @@ LogicalResult inferClampOp(
 }
 
 LogicalResult inferCompareOp(
-    MLIRContext* context, std::optional<Location>, Value lhs,
+    MLIRContext* context, std::optional<Location> location,
+    bool compareTypeHasValue, bool isCompareTypeNoType, bool isCompareTypeFloat,
+    bool isCompareTypeTotalOrder, bool isCompareTypeUnsigned,
+    bool isCompareTypeSigned, Value lhs,
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
+  // Verify CompareType.
```
burmako, line 1555: Let's drop this comment because it is fairly evident from the nearby code.

stablehlo/dialect/TypeInference.cpp
burmako: Removing completely sounds good to me as well. Let's loop in @GleasonK to provide guidance on how to do this properly with respect to VHLO (and e.g. whether we need to call this out in VhloDialect.td).

burmako (approved):

stablehlo/dialect/TypeInference.cpp
subhankarshah: I discussed with @GleasonK  and we decided to remove it from StableHLO, keep it in VHLO and update VHLO to StableHLO conversion to skip NOTYPE. While removing it, I was unable to resolve the build errors resulting from these usages (I tried removing these and updating the builder, could not get it right)

https://github.com/openxla/stablehlo/blob/17af8e5c82d98cb22f62768a26ae1072eec81826/stablehlo/dialect/StablehloOps.td#L1572
https://github.com/openxla/stablehlo/blob/17af8e5c82d98cb22f62768a26ae1072eec81826/stablehlo/dialect/StablehloOps.cpp#L3433

Opening another issue to deprecate NOTYPE and we can merge this as as if that is all right @burmako .

burmako: Following up separately SGTM. There's already a ticket about that too: #583.

burmako:
Reverted in #1071.

Issue #955, opened 2023-01-22
Move bounds related util functions out of base.cpp


zhouxin913:

zhouxin913:
Will start work on this once https://github.com/openxla/stablehlo/pull/961 is submitted.

zhouxin913:
Add a test for https://github.com/openxla/stablehlo/pull/1046 as well.

zhouxin913:
The ideal plan is to move all bounds related methods from Base.h/cpp to TypeInference.h/cpp including `verifyBounds()` until the end of file. However there lead to a dependency problem for `CompatibleOperandsAndResultType` https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/Base.h#L221:

`CompatibleOperandsAndResultType` is a trait used by ~20 ops, which contains the method `inferReturnTypes()` defined for these ops. And `inferReturnTypes()` depend on `inferMostSpecificType()` which is planed to moved to `TypeInference.h/cpp`. This break the dependency direction (`TypeInference.h/cpp` depend on `Base.h/cpp` instead of the other way around). 

Note that `Base.h/cpp` contains no only several basic StableHLO types, but also several traits. A approach would be split these traits away to a separate `trait.td/h/cpp` and then make the dependency from:
`TypeInference.h/cpp` -> `Base.td/h/cpp`
to 
`Trait.td/h/cpp` (contains `CompatibleOperandsAndResultType`)  -> `TypeInference.h/cpp` -> `Base.td/h/cpp` 

Welcome to provide more advise on this. @burmako

zhouxin913:
There is not resolution yet. Close for now, and reopen when needed.

PR #961, opened 2023-01-25
Handle bounds in the Select shape function

### LINKED ISSUE #750: Handle bounds in the Select shape function

ghpvnist:
For all dimensions, it infers the most specific of the `onTrue` and `onFalse` bounds, following the same inference rule in `inferMostSpecificTypeComponents`, so there is no change required to the inference rule except that the code for `inferMostSpecificTypeComponents` is updated to also include bounds information.
closes #750

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -2671,9 +2671,24 @@ LogicalResult inferSelectOp(
 
   // The output shape should be derived from the most specific parts of the
   // `onTrue` and `onFalse` (see documentation for details).
-  SmallVector<Type> inferredReturnTypes;
-  return inferMostSpecificTypeComponents(location, {trueType, falseType},
-                                         inferredReturnShapes);
+  auto onTrueRankedTy = onTrue.getType().dyn_cast<RankedTensorType>();
+  auto onFalseRankedTy = onFalse.getType().dyn_cast<RankedTensorType>();
+  if (!onTrueRankedTy || !onFalseRankedTy ||
+      (encodingToBounds(onTrueRankedTy.getEncoding()).empty() &&
+       encodingToBounds(onFalseRankedTy.getEncoding()).empty()))
+    return inferMostSpecificTypeComponents(location, {trueType, falseType},
+                                           inferredReturnShapes);
+
+  // The output shape and bounds are derived using the least specifc rules for
+  // `onTrue` and `onFalse` (see documentation for details).
+  auto inferredTypeOrErr =
+      inferLeastSpecificType(location, {onTrueRankedTy, onFalseRankedTy});
+  if (failed(inferredTypeOrErr)) return failure();
+  auto resultType = (*inferredTypeOrErr).cast<RankedTensorType>();
+  inferredReturnShapes.emplace_back(resultType.getShape(),
+                                    resultType.getElementType(),
+                                    resultType.getEncoding());
+  return success();
 }
 
 LogicalResult inferSelectAndScatterOp(
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -2670,21 +2670,10 @@ LogicalResult inferSelectOp(
                                "requires the same shape for all operands");
 
   // The output shape should be derived from the most specific parts of the
-  // `onTrue` and `onFalse` when the bounds are missing (see documantation for
-  // `inferMostSpecificTypeComponents` or `inferMostSpecificType` for details).
-  auto onTrueRankedTy = onTrue.getType().dyn_cast<RankedTensorType>();
-  auto onFalseRankedTy = onFalse.getType().dyn_cast<RankedTensorType>();
-  if (!onTrueRankedTy || !onFalseRankedTy)
-    return inferMostSpecificTypeComponents(location, {trueType, falseType},
-                                           inferredReturnShapes);
-  auto inferredTypeOrErr =
-      inferMostSpecificType(location, {onTrueRankedTy, onFalseRankedTy});
-  if (failed(inferredTypeOrErr)) return failure();
-  auto resultType = (*inferredTypeOrErr).cast<RankedTensorType>();
-  inferredReturnShapes.emplace_back(resultType.getShape(),
-                                    resultType.getElementType(),
-                                    resultType.getEncoding());
-  return success();
+  // `onTrue` and `onFalse` (see documentation for details).
+  SmallVector<Type> inferredReturnTypes;
+  return inferMostSpecificTypeComponents(location, {trueType, falseType},
+                                         inferredReturnShapes);
 }
 
 LogicalResult inferSelectAndScatterOp(
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -2671,20 +2671,14 @@ LogicalResult inferSelectOp(
 
   // The output shape should be derived from the most specific parts of the
   // `onTrue` and `onFalse` when the bounds are missing (see documantation for
-  // `inferMostSpecificTypeComponents` for details).
+  // `inferMostSpecificTypeComponents` or `inferMostSpecificType` for details).
   auto onTrueRankedTy = onTrue.getType().dyn_cast<RankedTensorType>();
   auto onFalseRankedTy = onFalse.getType().dyn_cast<RankedTensorType>();
-  if (!onTrueRankedTy || !onFalseRankedTy ||
-      (encodingToBounds(onTrueRankedTy.getEncoding()).empty() &&
-       encodingToBounds(onFalseRankedTy.getEncoding()).empty()))
+  if (!onTrueRankedTy || !onFalseRankedTy)
     return inferMostSpecificTypeComponents(location, {trueType, falseType},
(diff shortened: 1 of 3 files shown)


stablehlo/dialect/TypeInference.cpp
```diff
@@ -2649,9 +2649,24 @@ LogicalResult inferSelectOp(
 
   // The output shape should be derived from the most specific parts of the
```
sdasgup3, line 2650: We can update the comment here with something like 
`The output shape ..... when the bounds are missing`
(same below)

Also `see documentation for details`:  I think I am missing which documentation? same below

ghpvnist: Sure. Re: documentation, I clarified to refer to `inferXXX`'s documentation.

zhouxin913 (approved):
Approved. 

Other reviewers, this PR may seems verbose in some place: that's because the current infra is not clean enough: 
1. Lack of a `inferLeastSpecificTypeComponents()`
2. Inconsistant on the return type of methods: `FailureOr<ShapedType>` VS `LogicalResult`

I will make this right in next PR with issue https://github.com/openxla/stablehlo/issues/955.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -2670,10 +2670,27 @@ LogicalResult inferSelectOp(
...
-                                         inferredReturnShapes);
+  // `onTrue` and `onFalse` when the bounds are missing (see documantation for
+  // `inferMostSpecificTypeComponents` for details).
+  auto onTrueRankedTy = onTrue.getType().dyn_cast<RankedTensorType>();
+  auto onFalseRankedTy = onFalse.getType().dyn_cast<RankedTensorType>();
+  if (!onTrueRankedTy || !onFalseRankedTy ||
+      (encodingToBounds(onTrueRankedTy.getEncoding()).empty() &&
+       encodingToBounds(onFalseRankedTy.getEncoding()).empty()))
+    return inferMostSpecificTypeComponents(location, {trueType, falseType},
+                                           inferredReturnShapes);
```
burmako, line 2681: Hmm I'm not sure I follow. If we want to infer the least specific type, why do we need this code block that infers the most specific type under some circumstances?

burmako (changes requested):

stablehlo/dialect/TypeInference.cpp
ghpvnist: For this part of the check, since previous code was inferring the most specific shape before the bounds were introduced, I followed suit to infer the most specific shape when there are no static bounds. I intended to infer the least specific type only when the bounds are present. We end up having most specific shape w/o bounds and least specific shape/bounds if bounds are present. Does this pose issues down the line?

burmako: "Does this pose issues down the line?". Since we don't have a formal calculus for type inference yet (theoretical part), and since type inference with dynamic shapes is much less used than type inference with static shapes (practical part), it's hard to answer this definitively.

Based on my experience with working on programming languages and type systems, I would find this behavior quite surprising. Intuitively, we'd want to treat unbounded and bounded dynamism as similarly as possible, and this code would contradict this intuition. Also, when bounds are present, the behavior seems to flip for all dims (not just for bounded dims, but for unbounded dims), which introduces an additional element of surprise.

Overall, I would recommend that we stick to unsurprising solutions / solutions that align with other programming languages, unless there's a very good reason to deviate.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -2670,10 +2670,27 @@ LogicalResult inferSelectOp(
...
+
+  // The output shape and bounds are derived using the least specifc rules for
+  // `onTrue` and `onFalse` when the bounds are present (see documentation for
+  // `inferLeastSpecificType` for details).
+  auto inferredTypeOrErr =
+      inferLeastSpecificType(location, {onTrueRankedTy, onFalseRankedTy});
```
burmako, line 2687: Also, why are we using inferLeastSpecificType for SelectOp?

Perhaps, that's because CaseOp/IfOp use inferLeastSpecificType, and both perform operations conditionally? If that's what you had in mind, then I think that SelectOp is materially different.

For CaseOp/IfOp, execution can take one of the several available paths, so it makes sense to use inferLeastSpecificType, because we don't know in advance which path will be taken. Also, in many other programming languages, conditional expressions use least upper bounds for type inference. That checks out.

For SelectOp, execution can take only one available path. Exactly the same as other elementwise ops, the op requires `on_true` and `on_false` to have the same shape at runtime because it needs to zip them together. Given that, I don't think there's a case here to deviate from inferMostSpecificType that's used in other elementwise ops.

burmako (changes requested):

stablehlo/dialect/TypeInference.cpp
zhouxin913: Hmm. This claim changed my mind (take back my previous approval, sorry for this confusion).
`inferLeastSpecificType` is not proper for `select`, For example: 
If lhs is `tensor<4xi32>` and rhs is `tensor<?xi32>`, then the result have to be `tensor<4xi32>` instead of any other form, so as rhs and `pred`:  If any dim is larger or smaller than 4, then the semantic meaning of this op does not make sense anymore.
If both lhs and rhs are `tensor<?xi32>` but have difference bounds, then the result must be align with the operand with smaller the bound.
So we need stick with `inferMostSpecificType` to infer shape (not element type) from all `pred` (if rank(pred) 
 !=0), `lhs`, `rhs`.

zhouxin913 (changes requested):
See comment https://github.com/openxla/stablehlo/pull/961#discussion_r1090033280

stablehlo/dialect/TypeInference.cpp
ghpvnist: Makes sense, thanks for the input.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -2670,10 +2670,21 @@ LogicalResult inferSelectOp(
...
+    return inferMostSpecificTypeComponents(location, {trueType, falseType},
+                                           inferredReturnShapes);
+  auto inferredTypeOrErr =
+      inferMostSpecificType(location, {onTrueRankedTy, onFalseRankedTy});
+  if (failed(inferredTypeOrErr)) return failure();
+  auto resultType = (*inferredTypeOrErr).cast<RankedTensorType>();
+  inferredReturnShapes.emplace_back(resultType.getShape(),
+                                    resultType.getElementType(),
+                                    resultType.getEncoding());
+  return success();
```
zhouxin913, line 2687: Thanks for the update. How about we simplify all the new code to:
```
auto inferredTypeOrErr =
      inferMostSpecificType(location, {trueType, falseType});
  if (failed(inferredTypeOrErr)) return failure();
  auto resultType = (*inferredTypeOrErr).cast<RankedTensorType>();
  inferredReturnShapes.emplace_back(resultType.getShape(),
                                    resultType.getElementType(),
                                    resultType.getEncoding());
  return success();
```
So that we don't need to call twice, one with `inferMostSpecificTypeComponents` and another one with `inferMostSpecificType()`. Because it can handle unranked cases already.

zhouxin913 (changes requested):

zhouxin913 (approved):

stablehlo/dialect/TypeInference.cpp
burmako: Why do we need to make any changes to the code at all? I think that `inferMostSpecificTypeComponents` [already supports bounds](https://github.com/openxla/stablehlo/blob/b8f97cf15495617e8fbaea98b2f3977df66ecfa8/stablehlo/dialect/Base.cpp#L303-L310)?

burmako (changes requested):

stablehlo/dialect/TypeInference.cpp
ghpvnist: I also remember for a few ops we switched from using `inferReturnTypeComponents` to `inferredReturnTypes` but switched back because `inferReturnTypeComponents` now added support for encodings.

`inferSelectOp` passes `inferredReturnShapes` as a parameter (other ops use `inferReturnTypes` which uses `inferredReturnTypes` but this op passes `inferReturnTypeComponents`), so the encoding has to be explicitly passed, but `inferMostSpecificTypeComponents` only passes the shape.
https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/Base.cpp#L363-L370

burmako: Let's fix the linked code in Base.cpp to also pass the encoding to ShapedTypeComponents, then we'll be able to avoid code changes in TypeInference.cpp? I.e. replace `inferredReturnShapes.emplace_back(*inferredTypeOrErr);` with something like:
```C++
  inferredReturnShapes.emplace_back(resultType.getShape(),
                                    resultType.getElementType(),
                                    resultType.getEncoding());
```

burmako (changes requested):

stablehlo/dialect/TypeInference.cpp
ghpvnist: SGTM :)

burmako (approved):

PR #996, opened 2023-01-26
Add interpreter for SliceOp

### LINKED ISSUE #990: Add interpreter for slice

ghpvnist:

```
// Ix constraints come from the Inputs table
(I1) operand tensor
(I2) start_indices 1-dimensional tensor constant of type si64
(I3) limit_indices 1-dimensional tensor constant of type si64
(I4) strides 1-dimensional tensor constant of type si64
(C1) operand and result have the same element type.
(C2) size(start_indices) = size(limit_indices) = size(strides) = rank(operand).
(C3) 0 <= start_indices[d] <= limit_indices[d] <= dim(operand, d) for all dimension d.
(C4) 0 < strides[d] for all dimension d.
(C5) dim(result, d) = ... for all dimension d in operand.
```

```
I2: a) start_indices is not a tensor. (Covered by ODS).
    b) rank(start_indices) != 1.
    c) element_type(start_indices) != si64. (Covered by ODS).
I3: a) limit_indices is not a tensor. (Covered by ODS).
    b) rank(limit_indices) != 1. (Covered by ODS).
    c) element_type(limit_indices) != si64. (Covered by ODS).
I4: a) strides is not a tensor. (Covered by ODS).
    b) rank(strides) != 1. (Covered by ODS).
    c) element_type(strides) != si64. (Covered by ODS).
C1: element_type(operand) != element_type(result). (Covered by ODS).
C2: a) size(start_indices) != rank(operand).
    b) size(limit_indices) != rank(operand). (Covered by ODS).
    c) size(strides) != rank(operand). (Covered by ODS).
C3: a) start_indices < 0.
    b) start_indices > limit_indices.
    c) limit_indices > shape(operand).
C4: a) strides <= 0.
C5: no negative test needed since it's just inferring the shape
```

```
I2b: rank(start_indices) != 1.
C2a: size(start_indices) != rank(operand).
C3a: start_indices < 0.
C3b: start_indices > limit_indices.
C3c: limit_indices > shape(operand).
C4a: strides <= 0.
```
closes #990

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -4955,18 +4955,6 @@ dimension `d` in `operand`.
 #### Examples
 
 ```mlir
-// 1-dimensional slice
-
-// %operand: [0, 1, 2, 3, 4]
-%result = "stablehlo.slice"(%operand) {
-  start_indices = dense<2> : tensor<1xi64>,
-  limit_indices = dense<4> : tensor<1xi64>,
-  strides = dense<1> : tensor<1xi64>
-} : (tensor<5xi64>) -> tensor<2xi64>
-// %result: [2, 3]
-
-// 2-dimensional slice
-
 // %operand: [
 //            [0, 0, 0, 0],
 //            [0, 0, 1, 1],
@@ -4982,6 +4970,8 @@ dimension `d` in `operand`.
 //            [1, 1]
 //           ]
 ```
+
+&nbsp;[More Examples](../stablehlo/tests/interpret_slice.mlir)
 
 ### sort
 
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -1584,8 +1584,8 @@ def StableHLO_CompareOp: StableHLO_Op<"compare", [Pure, SameOperandsElementType,
 
 def StableHLO_SliceOp: StableHLO_Op<
       "slice",
-      [Pure, SameOperandsAndResultElementType,
-       AllTypesMatch<["start_indices", "limit_indices", "strides"]>,
+      [Pure, SameOperandsAndResultElementType /*slice_c1*/,
+       AllTypesMatch<["start_indices", "limit_indices", "strides"]> /*slice_c2*/,
        DeclareOpInterfaceMethods<InferTypeOpInterface>]> {
   let summary = "Slice operation";
   let description = [{
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -1587,7 +1587,7 @@ def StableHLO_SliceOp: StableHLO_Op<
       [Pure, SameOperandsAndResultElementType,
        AllTypesMatch<["start_indices", "limit_indices", "strides"]>,
        DeclareOpInterfaceMethods<InferTypeOpInterface>]> {
-  let summary = "Slice operator";
+  let summary = "Slice operation";
   let description = [{
     Extracts a slice from the `operand` using statically-computed starting
     indices and produces a `result` tensor.
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -1598,10 +1598,10 @@ def StableHLO_SliceOp: StableHLO_Op<
     Example:
     ```mlir
     %result = "stablehlo.slice" (%operand) {
-      start_indices = dense<2> : tensor<1xi64>,
-      limit_indices = dense<4> : tensor<1xi64>,
-      strides = dense<1> : tensor<1xi64>
-    } : (tensor<5xi64>) -> tensor<2xi64>
+      start_indices = dense<[1, 2]> : tensor<2xi64>,
+      limit_indices = dense<[3, 4]> : tensor<2xi64>,
+      strides = dense<1> : tensor<2xi64>
+    } : (tensor<3x4xi64>) -> tensor<2x2xi64>
     ```
   }];
 
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -2689,12 +2689,12 @@ LogicalResult inferSendOp(Dialect* dialect, std::optional<Location> location,
   return success();
 }
 
-LogicalResult inferSliceOp(std::optional<Location> location, Value operand,
+LogicalResult inferSliceOp(std::optional<Location> location, Type operandType,
                            DenseIntElementsAttr startIndices,
                            DenseIntElementsAttr limitIndices,
                            DenseIntElementsAttr strides,
                            SmallVectorImpl<Type>& inferredReturnTypes) {
-  RankedTensorType rankedTy = operandType.dyn_cast<RankedTensorType>();
+  auto rankedTy = operandType.dyn_cast<RankedTensorType>();
   if (!rankedTy) {
     // The operand type is unranked, so the best we can infer for the result
     // type is an unranked tensor with the same element type as the operand
@@ -2703,7 +2703,7 @@ LogicalResult inferSliceOp(std::optional<Location> location, Value operand,
     return success();
   }
 
-  // slice_c2
+  // slice_i2
   ShapedType attrTy = startIndices.getType();
   if (attrTy.getRank() != 1)
     return emitOptionalError(location, "start_indices has rank ",
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -2689,15 +2689,7 @@ LogicalResult inferSendOp(Dialect* dialect, std::optional<Location> location,
   return success();
 }
 
-// The following properties are already enforced by the ODS:
-//  type(start_indices) == type(limit_indices) == type(strides).
-// Verify the following properties:
-//  P1. Verify rank(start_indices) == 1.
-//  P2. Verify size(start_indices) == rank(operand).
-//  P3~5. Verify 0 <= start_indices[i] <= limit_indices[i] <= shape(operand)[i].
-//  P6. Verify stride[i] > 0.
-// Note: for P4, use the bound size than dim size for bounded dynamism case.
-LogicalResult inferSliceOp(std::optional<Location> location, Type operandType,
+LogicalResult inferSliceOp(std::optional<Location> location, Value operand,
                            DenseIntElementsAttr startIndices,
                            DenseIntElementsAttr limitIndices,
                            DenseIntElementsAttr strides,
@@ -2711,23 +2703,19 @@ LogicalResult inferSliceOp(std::optional<Location> location, Type operandType,
     return success();
   }
 
+  // SliceOp_C2
   ShapedType attrTy = startIndices.getType();
-  // P1.
-  // Note: ODS has type(start_indices) == type(limit_indices) == type(strides)
-  // So this implies rank(limit_indices) == rank(strides) == 1 also.
-  if (attrTy.getRank() != 1) {
+  if (attrTy.getRank() != 1)
     return emitOptionalError(location, "start_indices has rank ",
                              attrTy.getRank(), " instead of required rank 1");
-  }
 
-  // P2.
+  // SliceOp_C2
   int64_t rank = rankedTy.getRank();
-  if (attrTy.getNumElements() != rank) {
+  if (attrTy.getNumElements() != rank)
     return emitOptionalError(
         location, "the number of elements in start_indices (",
         attrTy.getNumElements(), ") does not match the rank of the operand (",
         rank, ")");
-  }
 
   SmallVector<int64_t, 4> start(startIndices.getValues<int64_t>());
   SmallVector<int64_t, 4> limit(limitIndices.getValues<int64_t>());
@@ -2738,12 +2726,12 @@ LogicalResult inferSliceOp(std::optional<Location> location, Type operandType,
   SmallVector<int64_t> resultBounds(inputBounds.size(), ShapedType::kDynamic);
 
   for (int64_t i = 0, e = rank; i != e; i++) {
-    // P3.
+    // SliceOp_C3
     if (start[i] < 0)
       return emitOptionalError(location, "negative start index ", start[i],
                                " in dimension ", i);
 
-    // P4.
+    // SliceOp_C3
     bool isStaticDim = !isDynamicDimSize(rankedTy.getDimSize(i));
     bool isStaticBound =
         !inputBounds.empty() && !isDynamicDimSize(inputBounds[i]);
@@ -2757,20 +2745,22 @@ LogicalResult inferSliceOp(std::optional<Location> location, Type operandType,
                                  operandSizeOrBound, " in dimension ", i);
     }
 
-    // P5.
+    // SliceOp_C3
     if (start[i] > limit[i])
       return emitOptionalError(location, "start index ", start[i],
                                " is larger than limit index ", limit[i],
                                " in dimension ", i);
-    // P6.
+    // SliceOp_C4
     if (strideVals[i] <= 0)
       return emitOptionalError(location, "stride must be positive but got ",
                                strideVals[i], " in dimension ", i);
 
+    // SliceOp_C5
     shape[i] = static_cast<int64_t>(
         llvm::divideCeil(limit[i] - start[i], strideVals[i]));
   }
 
+  // SliceOp_C1
   inferredReturnTypes.push_back(RankedTensorType::get(
       shape, rankedTy.getElementType(),
       boundsToEncoding(rankedTy.getEncoding(), resultBounds)));
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -2731,7 +2731,6 @@ LogicalResult inferSliceOp(std::optional<Location> location, Value operand,
       return emitOptionalError(location, "negative start index ", start[i],
                                " in dimension ", i);
 
-    // SliceOp_C3
     bool isStaticDim = !isDynamicDimSize(rankedTy.getDimSize(i));
     bool isStaticBound =
         !inputBounds.empty() && !isDynamicDimSize(inputBounds[i]);
@@ -2739,6 +2738,7 @@ LogicalResult inferSliceOp(std::optional<Location> location, Value operand,
       int64_t operandSizeOrBound =
           isStaticDim ? rankedTy.getDimSize(i) : inputBounds[i];
       StringRef sizeOrBound = isStaticDim ? "size" : "bound";
+      // SliceOp_C3
       if (limit[i] > operandSizeOrBound)
         return emitOptionalError(location, "limit index ", limit[i],
                                  " is larger than dimension ", sizeOrBound, " ",
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -2703,13 +2703,13 @@ LogicalResult inferSliceOp(std::optional<Location> location, Value operand,
     return success();
   }
 
-  // SliceOp_C2
+  // slice_c2
   ShapedType attrTy = startIndices.getType();
   if (attrTy.getRank() != 1)
     return emitOptionalError(location, "start_indices has rank ",
                              attrTy.getRank(), " instead of required rank 1");
 
-  // SliceOp_C2
+  // slice_c2
   int64_t rank = rankedTy.getRank();
   if (attrTy.getNumElements() != rank)
     return emitOptionalError(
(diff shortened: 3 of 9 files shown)


docs/spec.md
```diff
@@ -4955,18 +4955,6 @@ dimension `d` in `operand`.
 #### Examples
 
 ```mlir
-// 1-dimensional slice
```
subhankarshah, line 4958: maybe you wanted to copy and not cut?

subhankarshah (approved):
LGTM

stablehlo/tests/interpret_slice.mlir
```diff
@@ -0,0 +1,51 @@
...
+    start_indices = dense<[1, 2]> : tensor<2xi64>,
+    limit_indices = dense<[3, 4]> : tensor<2xi64>,
+    strides = dense<1> : tensor<2xi64>
+  } : (tensor<3x4xi64>) -> tensor<2x2xi64>
+  func.return %result : tensor<2x2xi64>
+  // CHECK-NEXT: tensor<2x2xi64>
+  // CHECK-NEXT: 1 : i64
+  // CHECK-NEXT: 1 : i64
+  // CHECK-NEXT: 1 : i64
+  // CHECK-NEXT: 1 : i64
```
subhankarshah, line 16: Would it be better to have the results in the following form `[[1, 1], [1, 1]] : tensor<2x2xi64>`
It would require only one CHECK-NEXT line and make it easier to write and understand the results.
For tests with more dimensions it is harder to understand what the result is, we can still have a separate line for type.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -2736,20 +2724,22 @@ LogicalResult inferSliceOp(std::optional<Location> location, Value operand,
                                  operandSizeOrBound, " in dimension ", i);
```
sdasgup3, line 2724: missing // SliceOp_C3
`limit_indices[d] <= dim(operand, d) for all dimension d.`

stablehlo/tests/interpret_slice.mlir
```diff
@@ -0,0 +1,51 @@
...
+    limit_indices = dense<[3, 6]> : tensor<2xi64>,
+    strides = dense<[2, 3]> : tensor<2xi64>
+  } : (tensor<3x6xi64>) -> tensor<2x2xi64>
+  func.return %result : tensor<2x2xi64>
+  // CHECK-NEXT: tensor<2x2xi64>
+  // CHECK-NEXT: 1 : i64
+  // CHECK-NEXT: 1 : i64
+  // CHECK-NEXT: 1 : i64
+  // CHECK-NEXT: 1 : i64
+}
```
sdasgup3, line 51: Can we have a test with start index   == limit index for a dimension. The idea is that the result should be a tensor with shape having 0 in one of the dimensions and the content of the tensor is empty.

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -2195,22 +2195,18 @@ func.func @slice(%arg0: tensor<3x4xi32>) -> tensor<1x2xi32> {
...
-// CHECK-LABEL: func @slice
-func.func @slice_dynamic_dim(%arg0: tensor<3x?xi32>) -> tensor<1x?xi32> {
-  %0 = "stablehlo.slice"(%arg0) {
-    start_indices = dense<[1, 1]> : tensor<2xi64>,
-    limit_indices = dense<[2, 2]> : tensor<2xi64>,
-    strides = dense<[1, 1]> : tensor<2xi64>
-  } : (tensor<3x?xi32>) -> tensor<1x?xi32>
-  func.return %0 : tensor<1x?xi32>
+// CHECK-LABEL: func @slice_unranked
+func.func @slice_unranked(%arg0: tensor<*xi32>) -> tensor<*xi32> {
```
sdasgup3, line 2199: `slice_unranked, slice_dynamic_dim and slice_indices_mismatch` does not follow alphabetical ordering (4(v) of the checklist).

stablehlo/reference/Ops.cpp
```diff
@@ -186,6 +192,22 @@ Tensor eval(SineOp op, const Tensor &operand) {
...
+  for (auto resItr = result.index_begin(); resItr != result.index_end();
+       ++resItr) {
+    SmallVector<int64_t> opIdx;
+    for (auto i = 0; i < operand.getType().getRank(); ++i)
+      opIdx.push_back(startIndices[i] + (*resItr)[i] * strides[i]);
+    result.set(*resItr, operand.get(opIdx));
+  }
+  return result;
+}
+
```
sdasgup3, line 210: Refactored based on
1. Remove `getSExtValues` function. 
2. unused `limitIndices`
3. `resItr` --> `resultItr`, `i` -> `dim`

```
Tensor eval(SliceOp op, const Tensor &operand) {
  Tensor result(op.getType());
  auto startIndices = op.getStartIndices().getValues<int64_t>();
  auto strides = op.getStrides().getValues<int64_t>();

       ++resultIt) {
    SmallVector<int64_t> operandIdx;
    for (auto dim = 0; dim < operand.getType().getRank(); ++dim)
      operandIdx.push_back(startIndices[dim] + (*resultIt)[dim] * strides[dim]);
    result.set(*resultIt, operand.get(operandIdx));
  }
  return result;
}
```

stablehlo/reference/Ops.cpp
```diff
@@ -185,6 +185,22 @@ Tensor evalSineOp(const Tensor &operand, Type resultType) {
...
+                   DenseIntElementsAttr stridesAttr, Type resultType) {
+  Tensor result(resultType);
+  auto startIndices = startIndicesAttr.getValues<int64_t>();
+  auto strides = stridesAttr.getValues<int64_t>();
+
+  for (auto resultIt = result.index_begin(); resultIt != result.index_end();
+       ++resultIt) {
+    SmallVector<int64_t> opIdx;
+    for (auto dim = 0; dim < operand.getType().getRank(); ++dim)
+      opIdx.push_back(startIndices[dim] + (*resultIt)[dim] * strides[dim]);
```
sdasgup3, line 198: nit:
`operandIdx` or `operandIndex` for clarity.

sdasgup3 (approved):
LGTM except for a small suggestion.

subhankarshah (approved):
LGTM

ghpvnist:
We are currently using a mix of custom class (`Tensor`) and MLIR classes (`DenseIntElementsAttr` and `Type`, but `Type` is debatable). What do you all think about peeling the wrapper classes in the caller (`Interpreter.cpp`) and use custom or LLVM classes in `Ops.cpp`? This will simplify more involved ops in the future by avoiding the need to create MLIR wrappers for calling ops with attributes by passing an ArrayRef of values instead. @burmako @sdasgup3

stablehlo/dialect/StablehloOps.td
```diff
@@ -1587,7 +1587,7 @@ def StableHLO_SliceOp: StableHLO_Op<
       [Pure, SameOperandsAndResultElementType,
        AllTypesMatch<["start_indices", "limit_indices", "strides"]>,
```
burmako, line 1588: Looks like we should add a `SliceOp_C2` comment here as well?

stablehlo/tests/interpret_slice.mlir
```diff
@@ -0,0 +1,71 @@
+// RUN: stablehlo-interpreter --interpret -split-input-file %s | FileCheck %s
```
burmako, line 1: Same question as in https://github.com/openxla/stablehlo/pull/1006/files#r1090063327. How do we decide how many of these tests are needed? E.g. should we only have `slice_op_2d` since it kind of tests everything that other tests include? 

And furthermore, how do we write positive tests in general? Also paging @sdasgup3 to discuss this.

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -2250,55 +2250,15 @@ func.func @slice(%arg0: tensor<3x4xi32>) -> tensor<1x2xi32> {
...
-func.func @slice_indices_mismatch(%arg0: tensor<3x4xi32>) -> tensor<1x4xi32> {
-  // expected-error@+1 {{failed to verify that all of {start_indices, limit_indices, strides} have same type}}
-  %0 = "stablehlo.slice"(%arg0) {start_indices = dense<[1, 2]> : tensor<2xi64>, limit_indices = dense<[2, 4, 1]> : tensor<3xi64>, strides = dense<[1, 2]> : tensor<2xi64>} : (tensor<3x4xi32>) -> tensor<1x4xi32>
-  func.return %0 : tensor<1x4xi32>
-}
-
-// -----
-
-func.func @slice_operand_result_mismatch(%arg0: tensor<3x4xi32>) -> tensor<1x4xf32> {
+func.func @slice_C1(%arg0: tensor<3x4xi32>) -> tensor<1x4xf32> {
```
burmako, line 2253: Same question as in another PR - let's say `@slice_c1` here?

burmako, line 2253: Another policy question - are we targeting complete coverage of constraints in these tests? If yes, then we'll want to have more tests e.g. in C2, C3. If no, how do we track that work? cc @sdasgup3

burmako (changes requested):

sdasgup3:
> Using `ArrayRef<T>` instead of various `ElementsAttr` attributes totally SGTM. I think should be doing that in TypeInference.h as well.

Agreed. Currently there are only two places where `eval` of an op takes `ElementsAttr` or its relatives.

For `TransposeOp`m  `DenseIntElementsAttr` is definitely an overkill.
```
Tensor evalTransposeOp(Type resultType, const Tensor &operand,
                       DenseIntElementsAttr permutation) {
```

But for `ConstantOp`

```
Tensor evalConstantOp(ElementsAttr value) {
  return makeTensor(value.cast<DenseElementsAttr>());
}
```

We can instead use two arguments of types `mlir::Type` and `ArrayRef<T>` .

burmako:
Yeah good point about ElementsAttr in ConstantOp. I think we should keep that for now, similarly to Type, for similar reasons - I believe it'll be non-trivial to reimplement on our own, and the benefit of that is not significant enough to prioritize right now, compared with other tasks.

stablehlo/tests/ops_stablehlo.mlir
sdasgup3: >> are we targeting complete coverage of constraints in these tests
Yes, that is right except for the checks enforced by the ODS using Type/Attr constraints and traits.
The idea was to test the code we introduced in the cpp functions and skip testing the mlir specific code implementing the ODS constraints.

On a side note: For this op, the constraint C2 is split between ODS and cpp files. We have a test,  labeled `slice_C2`, for only the part tested in cpp file.

sdasgup3: As @zhouxin913 pointed out to me that
https://github.com/openxla/stablehlo/blob/main/docs/type_inference.md#p4-establish-testing-guidelines  seems contradictory to what I am proposing here. I will get back to that asap.

stablehlo/tests/interpret_slice.mlir
ghpvnist: Good question. Should we include one comprehensive test that covers all the nuances or test the behaviors separately. E.g. DynamicUpdateSlice, testing that the op with the simplest formulation works, testing clamping of start indices, etc. or include them bunch in a single test.

stablehlo/tests/ops_stablehlo.mlir
burmako: "are we targeting complete coverage of constraints in these tests? Yes, that is right except for the checks enforced by the ODS using Type/Attr constraints and traits". Okay, then I have the following feedback. (And I'll also include my thought process around this).

```
// Ix constraints come from the Inputs table
(I1) operand tensor
(I2) start_indices 1-dimensional tensor constant of type si64
(I3) limit_indices 1-dimensional tensor constant of type si64
(I4) strides 1-dimensional tensor constant of type si64
(C1) operand and result have the same element type.
(C2) size(start_indices) = size(limit_indices) = size(strides) = rank(operand).
(C3) 0 <= start_indices[d] <= limit_indices[d] <= dim(operand, d) for all dimension d.
(C4) 0 < strides[d] for all dimension d.
(C5) dim(result, d) = ... for all dimension d in operand.
```

From what I can see, these constraints will be comprehensively covered by the following tests:

```
I2: a) start_indices is not a tensor. (Covered by ODS).
    b) rank(start_indices) != 1.
    c) element_type(start_indices) != si64. (Covered by ODS).
I3: a) limit_indices is not a tensor. (Covered by ODS).
    b) rank(limit_indices) != 1. (Covered by ODS).
    c) element_type(limit_indices) != si64. (Covered by ODS).
I4: a) strides is not a tensor. (Covered by ODS).
    b) rank(strides) != 1. (Covered by ODS).
    c) element_type(strides) != si64. (Covered by ODS).
C1: element_type(operand) != element_type(result). (Covered by ODS).
C2: a) size(start_indices) != rank(operand).
    b) size(limit_indices) != rank(operand). (Covered by ODS).
    c) size(strides) != rank(operand). (Covered by ODS).
C3: a) start_indices < 0.
    b) start_indices > limit_indices.
    c) limit_indices > shape(operand).
C4: a) strides <= 0.
C5: a) limit_indices[d]-start_indices[d] is a multiple of strides[d].
    b) is not a multiple of strides[d].
```

```
I2b: rank(start_indices) != 1.
C2a: size(start_indices) != rank(operand).
C3a: start_indices < 0.
C3b: start_indices > limit_indices.
C3c: limit_indices > shape(operand).
C4a: strides <= 0.
C5a: limit_indices[d]-start_indices[d] is a multiple of strides[d].
C5b: is not a multiple of strides[d].
```

Comparing these test cases with the actual test cases results in the following feedback on my part:
  1) The test for @slice_c1 doesn't seem to be needed (covered by ODS)
  2) The second test for @slice_c3 ("negative start index -1 in dimension 1") is redundant (covered by the first test).
  3) The test for @slice_indices_mismatch doesn't seem to be needed (covered by ODS).
  4) The test for C5b is missing.
  5) Let's not worry about dynamic shapes for now, since they aren't part of the spec at the moment.

burmako: "Let's not worry about dynamic shapes for now". What I mean by this is "let's keep whatever tests for dynamic shapes that we have but not worry about comprehensively covering dynamism support for specific ops".

ghpvnist: Thanks for the detailed feedback! The brevity of breaking down constraints definitely helps see which negative tests are needed. It seems we agree on leaving constraints covered by the ODS and instead have a comment next to it. Also it's nice to point out that we should also cover requirements listed in the Inputs section.

stablehlo/tests/interpret_slice.mlir
sdasgup3: >> how many test
I would say that it desirable to have minimal tests which meets our set coverage criteria. For the 

>> how to decide   
The idea behind the testing is to make sure that the spec is comprehensively covered. That coverage must also get transcended to the interpreter implementation as it is supposed to be readable implementation of the spec. having said that the number we should aim for close to minimum while meeting our coverage criteria.

For a like SliceOp,  we could have a single test covering all the cases. This will not be the case if for an op if its implementation handles dimensions differently: That needs separate test to cover each code path. 
Taking assistance from the coverage tools might help us when to stop. 

 
I propose to write test with "best effort" covering  multiple behaviors. By doing that I believe most of the simpler ops will end up having one test. From collective "best efforts" from reviewer and author, I think we should be able to get a near minimal set

stablehlo/tests/ops_stablehlo.mlir
burmako: "Also it's nice to point out that we should also cover requirements listed in the Inputs section". If we're in alignment on the proposed strategy, let's also update the checklist. Also cc @sdasgup3.

stablehlo/tests/interpret_slice.mlir
burmako: LGTM. Let's update the checklist accordingly?

burmako: Given the discussion above, do we still want to have the `slice_op_empty_slice` test?

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -2250,55 +2250,7 @@ func.func @slice(%arg0: tensor<3x4xi32>) -> tensor<1x2xi32> {
...
-func.func @slice_operand_result_mismatch(%arg0: tensor<3x4xi32>) -> tensor<1x4xf32> {
-  // expected-error@+1 {{requires the same element type for all operands and results}}
-  %0 = "stablehlo.slice"(%arg0) {start_indices = dense<[1, 0]> : tensor<2xi64>, limit_indices = dense<[2, 4]> : tensor<2xi64>, strides = dense<[1, 2]> : tensor<2xi64>} : (tensor<3x4xi32>) -> tensor<1x4xf32>
-  func.return %0 : tensor<1x4xf32>
-}
-
-// -----
-
-func.func @slice_indices_not_rank_1(%arg0: tensor<3x4xi32>) -> tensor<1x2xi32> {
+func.func @slice_c2(%arg0: tensor<3x4xi32>) -> tensor<1x2xi32> {
```
burmako, line 2253: Strictly speaking, this is not `@slice_c2` but `@slice_i2` (following the proposed convention from the nearby comment).

stablehlo/tests/ops_stablehlo.mlir
burmako: Agreed about C5b. All other tests check out as well, except that the first slice_c2 test is actually a slice_i2 test (noted in a separate comment thread).

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -2250,55 +2250,7 @@ func.func @slice(%arg0: tensor<3x4xi32>) -> tensor<1x2xi32> {
```
burmako, line 2250: On a related note, how about also having the author of a PR to come up with an explicitly written down testing strategy (in a PR description) similarly to what I wrote in https://github.com/openxla/stablehlo/pull/996#discussion_r1093933011? This way, everyone including future code archaeologists will be able to verify whether tests were skipped on purpose or as oversight.

burmako (changes requested):

stablehlo/tests/ops_stablehlo.mlir
ghpvnist: I think it's definitely a good idea that having an explicitly written testing strategy will help see the overall constraints needing tests and also for future viewers which I have updated the PR description as a starter, but I don't think this will scale very well with more complex ops. For example, the PR description is already quite lengthy with 2 inputs and 5 constraints. Taking the largest op as an example, ConvolutionOp has 19 inputs and 28 constraints. This would amount to an enormous burden for authors of such ops, especially when there are parts to the constraints. Wdyt?

stablehlo/tests/interpret_slice.mlir
ghpvnist: Agreed. Will remove the test.

stablehlo/tests/ops_stablehlo.mlir
burmako: "This would amount to an enormous burden for authors of such ops, especially when there are parts to the constraints. Wdyt?". I agree that this is a whole lot of work, but otherwise how do we know that we have comprehensively covered the op with tests? I would recommend that we do this, so that we can resolve this question once and for all.

StableHLO is developed as a portability layer between different ML frameworks and ML compilers, and I think that for that goal there is tremendous value in achieving comprehensive coverage. For the complicated ops, this will cost us hours per op (both in author and reviewer time), but I wouldn't be surprised if this saves the community dozens or even hundreds of hours overall.

stablehlo/reference/Ops.h
```diff
@@ -38,6 +38,9 @@ Tensor evalNotOp(const Tensor &operand, Type resultType);
 Tensor evalOrOp(const Tensor &lhs, const Tensor &rhs, Type resultType);
 Tensor evalReshapeOp(const Tensor &operand, Type resultType);
 Tensor evalSineOp(const Tensor &operand, Type resultType);
+Tensor evalSliceOp(const Tensor &operand,
+                   const SmallVector<int64_t> startIndices,
```
burmako, line 42: Apologies - I'm just noticing this. This should be `ArrayRef<int64_t>` instead of `SmallVector<int64_t>`.

burmako (approved):
One minor comment, and we're good to go.

stablehlo/reference/Ops.h
ghpvnist: Thanks!

stablehlo/tests/ops_stablehlo.mlir
ghpvnist: Agreed. This will also save the reviewers time verifying that all constraints are addressed, and future referees of the PR will understand why the tests were written/omitted.

PR #998, opened 2023-01-26
Add interpreter for DynamicUpdateSliceOp

### LINKED ISSUE #974: Add interpreter for dynamic_update_slice

ghpvnist:
Unlike the spec, the interpreter uses its own implementation of clamp to calculate `adjustedStartIndices` since the start indices are given as variadic tensors, and formulating as one tensor to use the ClampOp is more loc.

```
(I1) operand tensor.
(I2) update tensor.
(I3) start_indices variadic number of 0-dimensional tensors of integer type.
(C1) `operand` and `result` have the same type.
(C2) element_type(`update`) $=$ element_type(`operand`).
(C3) rank(`update`) $=$ rank(`operand`).
(C4) size(`start_indices`) $=$ rank(`operand`).
(C5) All `start_indices` have the same type.
(C6) dim(`update`, `k`) $\in$ [0, dim(`operand`, `k`)] for all `k` $\in$ [0, rank(`operand`)).
```

```
I2: a) update is not a tensor. (Covered by ODS).
I3: a) start_indices does not have 0-dimensional tensors. (Covered by ODS).
    b) start_indices does not have integer type. (Covered by ODS).
C1: a) element_type(operand) != element_type(result). (Covered by ODS).
    b) shape(operand) != shape(result).
C2: a) element_type(update) != element_type(operand). (Covered by ODS).
C3: a) rank(update) != rank(operand).
C4: a) size(start_indices) != rank(operand).
C5: a) all start_indices does not have the same type.
C6: a) dim(update, k) < 0 for any k.
    b) dim(update, k) > dim(operand, k) for any k.
```

```
C1b: no negative test needed since result shape is inferred from operand shape
C3a: rank(update) != rank(operand).
C4a: size(start_indices) != rank(operand).
C5a: all start_indices does not have the same type.
C6a: dim(update, k) < 0 for any k.
C6b: dim(update, k) > dim(operand, k) for any k. 
```

closes #974

### DIFF
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -1666,8 +1666,8 @@ def StableHLO_DynamicUpdateSliceOp: StableHLO_Op<"dynamic_update_slice",
     ```
   }];
   let arguments = (ins
-    HLO_Tensor:$operand,
-    HLO_Tensor:$update,
+    HLO_Tensor:$operand /*dynamic_update_slice_i1*/,
+    HLO_Tensor:$update /*dynamic_update_slice_i2*/,
     Variadic<HLO_ScalarIntTensor>:$start_indices /*dynamic_update_slice_i3*/
   );
   let results = (outs HLO_Tensor:$result);
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -1648,7 +1648,7 @@ def StableHLO_DynamicSliceOp: StableHLO_Op<"dynamic_slice",
 }
 
 def StableHLO_DynamicUpdateSliceOp: StableHLO_Op<"dynamic_update_slice",
-      [Pure, AllElementTypesMatch<["operand", "update", "result"]>,
+      [Pure, AllElementTypesMatch<["operand", "update", "result"]> /*dynamic_update_slice_c1, dynamic_update_slice_c2*/,
        InferTensorType]> {
   let summary = "DynamicUpdateSlice operation";
   let description = [{
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -1650,7 +1650,7 @@ def StableHLO_DynamicSliceOp: StableHLO_Op<"dynamic_slice",
 def StableHLO_DynamicUpdateSliceOp: StableHLO_Op<"dynamic_update_slice",
       [Pure, AllElementTypesMatch<["operand", "update", "result"]>,
        InferTensorType]> {
-  let summary = "Dynamic Update Slice operator";
+  let summary = "DynamicUpdateSlice operation";
   let description = [{
     Produces a `result` tensor which is equal to the `operand` tensor except
     that the slice starting at `start_indices` is updated with the values in
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -1668,7 +1668,7 @@ def StableHLO_DynamicUpdateSliceOp: StableHLO_Op<"dynamic_update_slice",
   let arguments = (ins
     HLO_Tensor:$operand,
     HLO_Tensor:$update,
-    Variadic<HLO_ScalarIntTensor>:$start_indices
+    Variadic<HLO_ScalarIntTensor>:$start_indices /*dynamic_update_slice_i3*/
   );
   let results = (outs HLO_Tensor:$result);
 
--- a/stablehlo/reference/Ops.cpp
+++ b/stablehlo/reference/Ops.cpp
@@ -102,14 +102,13 @@ Tensor evalDynamicUpdateSliceOp(const Tensor &operand, const Tensor &update,
     adjustedStartIndices.push_back(std::min(
         std::max(startIndices[i].get({}).getIntegerValue().getSExtValue(), 0l),
         operandShape[i] - updateShape[i]));
-  for (auto resItr = result.index_begin(), updItr = update.index_begin();
-       resItr != result.index_end(); ++resItr) {
-    if (updItr != update.index_end() &&
-        llvm::equal(*resItr, addIndices(*updItr, adjustedStartIndices)))
-      result.set(*resItr, update.get(*updItr++));
-    else
-      result.set(*resItr, operand.get(*resItr));
-  }
+  for (auto resultIt = result.index_begin(); resultIt != result.index_end();
+       ++resultIt)
+    result.set(*resultIt, operand.get(*resultIt));
+  for (auto updateIt = update.index_begin(); updateIt != update.index_end();
+       ++updateIt)
+    result.set(addIndices(*updateIt, adjustedStartIndices),
+               update.get(*updateIt));
   return result;
 }
 
--- a/stablehlo/reference/Ops.cpp
+++ b/stablehlo/reference/Ops.cpp
@@ -92,7 +92,7 @@ Tensor evalCosineOp(const Tensor &operand, Type resultType) {
 }
 
 Tensor evalDynamicUpdateSliceOp(const Tensor &operand, const Tensor &update,
-                                const ArrayRef<Tensor> &startIndices,
+                                ArrayRef<Tensor> startIndices,
                                 Type resultType) {
   Tensor result(resultType);
   auto operandShape = operand.getType().getShape();
--- a/stablehlo/reference/Ops.cpp
+++ b/stablehlo/reference/Ops.cpp
@@ -17,6 +17,7 @@ limitations under the License.
 
 #include "llvm/ADT/APFloat.h"
 #include "llvm/ADT/APInt.h"
+#include "llvm/ADT/STLExtras.h"
 #include "llvm/Support/Errc.h"
 #include "llvm/Support/Error.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
@@ -24,6 +25,7 @@ limitations under the License.
 #include "mlir/Support/DebugStringHelper.h"
 #include "stablehlo/reference/Element.h"
 #include "stablehlo/reference/Errors.h"
+#include "stablehlo/reference/Tensor.h"
 #include "stablehlo/reference/Types.h"
 
 namespace mlir {
@@ -37,6 +39,13 @@ SmallVector<int64_t> permute(ArrayRef<int64_t> array, ArrayRef<int64_t> perm) {
   SmallVector<int64_t> result(array.size());
   for (size_t i = 0; i < array.size(); i++) result[i] = array[perm[i]];
   return result;
+}
+
+SmallVector<int64_t> addIndices(ArrayRef<int64_t> lhs, ArrayRef<int64_t> rhs) {
+  SmallVector<int64_t> combined;
+  for (auto [lhsIdx, rhsIdx] : llvm::zip(lhs, rhs))
+    combined.push_back(lhsIdx + rhsIdx);
+  return combined;
 }
 
 }  // namespace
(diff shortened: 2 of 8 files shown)


stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -2427,38 +2419,46 @@ func.func @dynamic_update_slice_invalid_start(%input: tensor<3x4xi64>, %update:
...
-func.func @dynamic_update_slice_invalid_update_size(%input: tensor<3x4xi64>, %update: tensor<1x5xi64>, %start1: tensor<i64>, %start2: tensor<i64>) -> tensor<3x4xi64> {
+func.func @dynamic_update_slice_C6(%input: tensor<3x4xi64>, %update: tensor<1x5xi64>, %start1: tensor<i64>, %start2: tensor<i64>) -> tensor<3x4xi64> {
   // expected-error@+1 {{expects size at dimension 1 of update to be in range [0, 4]. Got: 5.}}
   %0 = "stablehlo.dynamic_update_slice"(%input, %update, %start1, %start2) : (tensor<3x4xi64>, tensor<1x5xi64>, tensor<i64>, tensor<i64>) -> tensor<3x4xi64>
   func.return %0 : tensor<3x4xi64>
 }
 
 // -----
 
+// CHECK-LABEL: @dynamic_update_slice_dynamic_dim
```
subhankarshah, line 2454: can we have the positive test first and then the negative tests.

stablehlo/reference/Interpreter.cpp
```diff
@@ -58,6 +58,18 @@ llvm::Expected<SmallVector<Tensor>> eval(func::FuncOp func,
       if (it != stackFrame.end()) return it->second;
       llvm::report_fatal_error("Expected a terminator when evaluating func");
     };
+    auto fetchVariadic = [&](OperandRange range) -> ArrayRef<Tensor> {
+      SmallVector<Tensor> arr;
+      for (auto value : range) {
+        auto it = stackFrame.find(value);
+        if (it != stackFrame.end()) {
```
subhankarshah, line 65: can we remove braces from both if else statements

subhankarshah: maybe move the statements inside the block next to if else if there is enough space?

subhankarshah (changes requested):
Added a few comments

stablehlo/tests/ops_stablehlo.mlir
ghpvnist: Updated the test to be alphabetical per [testing guidelines](https://github.com/openxla/stablehlo/blob/main/docs/reference.md#testing-guidelines).

stablehlo/reference/Interpreter.cpp
ghpvnist: Updated the changes for your first comment. The second comment will not be allowed by the formatter we are using.

subhankarshah (approved):
LGTM

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -2459,85 +2459,93 @@ func.func @dynamic_slice_slice_size_too_large(%arg0: tensor<3x4xi32>, %arg1: ten
...
-func.func @dynamic_update_slice_dynamic_sizes(%input: tensor<?x4xi64>, %update: tensor<1x?xi64>, %start1: tensor<i64>, %start2: tensor<i64>) -> tensor<?x4xi64> {
-  %0 = "stablehlo.dynamic_update_slice"(%input, %update, %start1, %start2) : (tensor<?x4xi64>, tensor<1x?xi64>, tensor<i64>, tensor<i64>) -> tensor<?x4xi64>
+func.func @dynamic_update_slice_dynamic_sizes(%operand: tensor<?x4xi64>, %update: tensor<1x?xi64>, %start1: tensor<i64>, %start2: tensor<i64>) -> tensor<?x4xi64> {
+  %0 = "stablehlo.dynamic_update_slice"(%operand, %update, %start1, %start2) : (tensor<?x4xi64>, tensor<1x?xi64>, tensor<i64>, tensor<i64>) -> tensor<?x4xi64>
   func.return %0 : tensor<?x4xi64>
 }
 
 // -----
 
+func.func @dynamic_update_slice_invalid_start(%operand: tensor<3x4xi64>, %update: tensor<1x2xi64>, %start: tensor<2xi64>) -> tensor<3x4xi64> {
```
sdasgup3, line 2541: remove ODS tests?

stablehlo/reference/Interpreter.cpp
```diff
@@ -58,6 +58,16 @@ llvm::Expected<SmallVector<Tensor>> eval(func::FuncOp func,
       if (it != stackFrame.end()) return it->second;
       llvm::report_fatal_error("Expected a terminator when evaluating func");
     };
+    auto fetchVariadic = [&](OperandRange range,
```
sdasgup3, line 61: `fetchVariadicOperand` and let the function return an `ArrayRef<Tensor>`.

stablehlo/reference/Interpreter.cpp
```diff
@@ -58,6 +58,17 @@ llvm::Expected<SmallVector<Tensor>> eval(func::FuncOp func,
       if (it != stackFrame.end()) return it->second;
       llvm::report_fatal_error("Expected a terminator when evaluating func");
     };
+    auto fetchVariadic = [&](OperandRange range) -> ArrayRef<Tensor> {
```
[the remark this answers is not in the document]
ghpvnist: I think the code will look ugly dealing with std::variants, wrapping and unwrapping ArrayRef for ops with non-variadic operands. Do you have a particular solution in mind?

stablehlo/dialect/StablehloOps.td
```diff
@@ -1648,9 +1648,9 @@ def StableHLO_DynamicSliceOp: StableHLO_Op<"dynamic_slice",
...
 
 def StableHLO_DynamicUpdateSliceOp: StableHLO_Op<"dynamic_update_slice",
-      [Pure, AllElementTypesMatch<["operand", "update", "result"]>,
+      [Pure, AllElementTypesMatch<["operand", "update", "result"]> /*dynamic_update_slice_c1, dynamic_update_slice_c2*/,
        InferTensorType]> {
-  let summary = "Dynamic Update Slice operator";
+  let summary = "DynamicUpdateSlice operation";
   let description = [{
     Produces a `result` tensor which is equal to the `operand` tensor except
     that the slice starting at `start_indices` is updated with the values in
```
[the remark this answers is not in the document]
ghpvnist: Let's have it written for consistency.

stablehlo/reference/Interpreter.cpp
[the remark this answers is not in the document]
sdasgup3: Makes sense. Can we resue fetchOperand as follows:
```
 auto fetchVariadicOperands = [&](OperandRange values) {
    return llvm::to_vector(llvm::map_range(
        values, [&](Value value) { return fetchOperand(value); }));
  };
```

sdasgup3 (approved):
Approved with a suggestion.

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -2459,80 +2459,80 @@ func.func @dynamic_slice_slice_size_too_large(%arg0: tensor<3x4xi32>, %arg1: ten
...
-  func.return %0 : tensor<3x4xi64>
-}
-
-// -----
-
-func.func @dynamic_update_slice_invalid_start(%input: tensor<3x4xi64>, %update: tensor<1x2xi64>, %start: tensor<2xi64>) -> tensor<3x4xi64> {
-  // expected-error@+1 {{operand #2 must be 0D tensor of 4/8/16/32/64-bit signless integer or 4/8/16/32/64-bit unsigned integer values, but got 'tensor<2xi64>'}}
-  %0 = "stablehlo.dynamic_update_slice"(%input, %update, %start) : (tensor<3x4xi64>, tensor<1x2xi64>, tensor<2xi64>) -> tensor<3x4xi64>
-  func.return %0 : tensor<3x4xi64>
+func.func @dynamic_update_slice_c1(%operand: tensor<3x4xi64>, %update: tensor<1x4xi64>, %start1: tensor<i64>, %start2: tensor<i64>) -> tensor<3x5xi64> {
```
sdasgup3, line 2469: Please make sure we have the check-label present consistently
```
Make sure that all the tests related to the op under test, are prepended with CHECK-LABEL lit macro.
```

ghpvnist: Negative tests cannot have CHECK-LABELs as verifications fail before the output is created, resulting in a blank test.

stablehlo/reference/Ops.cpp
```diff
@@ -38,6 +38,13 @@ SmallVector<int64_t> permute(ArrayRef<int64_t> array, ArrayRef<int64_t> perm) {
   return result;
 }
 
+SmallVector<int64_t> addIndices(ArrayRef<int64_t> lhs, ArrayRef<int64_t> rhs) {
+  SmallVector<int64_t> combined;
+  for (auto [lhsIdx, rhsIdx] : llvm::zip(lhs, rhs))
+    combined.push_back(lhsIdx + rhsIdx);
+  return combined;
+}
```
burmako, line 46: Given that we've started introducing these helper functions, I think we should create `class Index` and supersede `ArrayRef<int64_t>` for these things in the very near future. Doesn't have to block this PR, but can we open a ticket so that we don't forget? See related discussion in https://github.com/openxla/stablehlo/pull/1005#discussion_r1103409763.

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -2376,80 +2376,80 @@ func.func @dynamic_slice_slice_size_too_large(%arg0: tensor<3x4xi32>, %arg1: ten
 // -----
 
 // CHECK-LABEL: @dynamic_update_slice
-func.func @dynamic_update_slice(%input: tensor<3x4xi64>, %update: tensor<1x4xi64>, %start1: tensor<i64>, %start2: tensor<i64>) -> tensor<3x4xi64> {
-  %0 = "stablehlo.dynamic_update_slice"(%input, %update, %start1, %start2) : (tensor<3x4xi64>, tensor<1x4xi64>, tensor<i64>, tensor<i64>) -> tensor<3x4xi64>
+func.func @dynamic_update_slice(%operand: tensor<3x4xi64>, %update: tensor<1x4xi64>, %start1: tensor<i64>, %start2: tensor<i64>) -> tensor<3x4xi64> {
```
burmako, line 2379: `start_indices0`, etc? (Here and below).

stablehlo/reference/Ops.cpp
```diff
@@ -72,6 +79,28 @@ Tensor evalCosineOp(const Tensor &operand, Type resultType) {
...
+        operandShape[i] - updateShape[i]));
+  for (auto resItr = result.index_begin(), updItr = update.index_begin();
+       resItr != result.index_end(); ++resItr) {
+    if (updItr != update.index_end() &&
+        llvm::equal(*resItr, addIndices(*updItr, adjustedStartIndices)))
+      result.set(*resItr, update.get(*updItr++));
+    else
+      result.set(*resItr, operand.get(*resItr));
+  }
+  return result;
```
burmako, line 101: Feedback about names: `resultIt` and `updateIt`.

stablehlo/reference/Ops.cpp
```diff
@@ -72,6 +79,28 @@ Tensor evalCosineOp(const Tensor &operand, Type resultType) {
...
+  SmallVector<int64_t> adjustedStartIndices;
+  for (size_t i = 0; i < startIndices.size(); ++i)
+    adjustedStartIndices.push_back(std::min(
+        std::max(startIndices[i].get({}).getIntegerValue().getSExtValue(), 0l),
+        operandShape[i] - updateShape[i]));
+  for (auto resItr = result.index_begin(), updItr = update.index_begin();
+       resItr != result.index_end(); ++resItr) {
+    if (updItr != update.index_end() &&
+        llvm::equal(*resItr, addIndices(*updItr, adjustedStartIndices)))
+      result.set(*resItr, update.get(*updItr++));
```
burmako, line 97: The part where we coiterate `result` and `update` made me a bit nervous, as in I could not immediately see why this code works. Or, more specifically, how do we know that each value of `updateIt` will be matched with the corresponding `resultIt`, given that we increment `updateIt` and `resultIt` separately? 

I mean, I got it after staring at the code for a bit and then writing down some notes, but it wasn't immediately obvious.

Given that we're optimizing for readability, how about we do exactly the same thing as the implementation of PadOp does? First, fill the result with operand elements. And then, have a separate loop that iterates update's index space?

burmako (changes requested):

burmako (approved):

stablehlo/reference/Ops.cpp
burmako: Opened #1163.

Issue #233, opened 2022-10-01
Further discussion on lifetime and mutability of tensors


sdasgup3:
### Request description

Follow up the discussion from [here](https://github.com/openxla/stablehlo/pull/117#issuecomment-1262869759) 

As mentioned in the [spec](https://github.com/openxla/stablehlo/blob/main/docs/spec_draft.md#types), StableHLO `Tensor` objects are immutable. However, the way we are modeling it in the interpreter not in sync with that. 

Any one of the followings might break the immutability proposition  
1. We allow `Tensor` objects to be updated using `Tensor::set` api. 
2. `Tensor` has a shared and reference counted underlying storage called `Buffer` and they are allowed to be copied. Together with (1) we can modify the a `Tensor` object `x`  updating all the `Tensor` objects sharing the `Buffer` of x.

This issue is about discussing the possible solutions.

PS: I will update this my proposal.

sdasgup3:
Here is my take on this:

At the moment, a `Tensor` is created using 
```
 (1)  explicit Tensor(ShapedType type);
  (2) explicit Tensor(DenseElementsAttr attr);
```

and modified using 

```
(3) void set(ArrayRef<int64_t> index, const Element &element);
```

(1) helps us define an default initialized tensor whose elements can be modified later. This is mostly used in implementing element-wise ops, via `eval` functions, like [this](https://github.com/openxla/stablehlo/blob/75d7b654e51729f1e733af6d7fa593514f14db6e/stablehlo/reference/Ops.cpp#L35) in an intuitive way.

An observation: Tensor are created and updated only in the eval function of various ops. Beyond those functions,  they are not (and should not be) modified.  In other words, `eval` functions  are the only source of generating `Tensor` objects. How about making `eval` the only source of generating **immutable** `Tensor` objects. 

Based on the above observation, the behavior, w.r.t a `Tensor`'s mutability, that we might look for is: A `Tensor` object to be mutable until some point in time after which it is frozen i.e.  not mutable anymore.
 
We might need the following to achieve that:
1. Make the constructors (1) and (2) available only to `eval`: One option is to make (1) and (2) private and let `eval`s the friend of `Tensor`.
[...]

PR #1005, opened 2023-01-27
Add interpreter for PadOp

### LINKED ISSUE #980: Add interpreter for pad

ghpvnist:

```
(I1) operand tensor
(I2) padding_value 0-dimensional tensor
(I3) edge_padding_low 1-dimensional tensor constant of type si64
(I4) edge_padding_high 1-dimensional tensor constant of type si64
(I5) interior_padding 1-dimensional tensor constant of type si64
(C1) `operand`, `padding_value`, `result` have the same element type.
(C2) `edge_padding_low`, `edge_padding_high`, `interior_padding` have the
size equal to `operand`'s rank.
(C3) 0 $\le$ `interior_padding[i]` for all `i` values in `interior_padding`.
(C4) 0 $\le$ `dim(result, i)` for all `i`th dimension of `operand`, where
`dim(result, i) = di + max(di - 1, 0) * interior_padding[i] + edge_padding_low[i] + edge_padding_high[i]`
and `di = dim(operand, i)`.
```

```
I2: a) padding_value is not a 0-dimensional tensor.
I3: a) edge_padding_low is not a 1-dimensional tensor.
    b) element_type(edge_padding_low) != si64. (Covered by ODS).
I4: a) edge_padding_high is not a 1-dimensional tensor.
    b) element_type(edge_padding_high) != si64. (Covered by ODS).
I5: a) interior_padding is not a 1-dimensional tensor.
    b) element_type(interior_padding) != si64. (Covered by ODS).
C1: a) element_type(operand) != element_type(padding_value) != element_type(result). (Covered by ODS).
C2: a) size(edge_padding_low) != rank(operand).
    b) size(edge_padding_high) != rank(operand).
    c) size(interior_padding) != rank(operand).
C3: a) interior_padding < 0.
C4: no negative test added since it's just inferring the shape.
```

```
I2a: padding_value is not a 0-dimensional tensor.
I3a: edge_padding_low is not a 1-dimensional tensor.
I4a: edge_padding_high is not a 1-dimensional tensor.
I5a: interior_padding is not a 1-dimensional tensor.
C2a: size(edge_padding_low) != rank(operand).
C2b: size(edge_padding_high) != rank(operand).
C2c: size(interior_padding) != rank(operand).
C3a: interior_padding < 0.
```

Note that for `I3a`, `I4a`, and `I5a`, they are partially covered by the `AllTypesMatch` trait, so only one test (`I2a`) is needed to test all three. Similarly with `C2a`, `C2b`, and `C2c` (`C2a` test is added to cover all three).

closes #980

### DIFF
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -2670,7 +2670,6 @@ def StableHLO_ReverseOp: StableHLO_Op<"reverse",
 
 def StableHLO_PadOp: StableHLO_ShapedInterfaceOp<"pad",
       [Pure, SameOperandsAndResultElementType /*pad_c1*/,
-      AllTypesMatch<["edge_padding_low", "edge_padding_high", "interior_padding"]> /*pad_c2_i3_i4_i5*/,
       DeclareOpInterfaceMethods<InferTypeOpInterface>]> {
   let summary = "Pad operation";
   let description = [{
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -2670,6 +2670,8 @@ def StableHLO_ReverseOp: StableHLO_Op<"reverse",
 
 def StableHLO_PadOp: StableHLO_ShapedInterfaceOp<"pad",
       [Pure, SameOperandsAndResultElementType /*pad_c1*/,
+      AllTypesMatch<["edge_padding_low", "edge_padding_high",
+      "interior_padding"]> /*pad_c2, pad_i3, pad_i4, pad_i5*/,
       DeclareOpInterfaceMethods<InferTypeOpInterface>]> {
   let summary = "Pad operation";
   let description = [{
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -2671,7 +2671,7 @@ def StableHLO_ReverseOp: StableHLO_Op<"reverse",
 def StableHLO_PadOp: StableHLO_ShapedInterfaceOp<"pad",
       [Pure, SameOperandsAndResultElementType,
       DeclareOpInterfaceMethods<InferTypeOpInterface>]> {
-  let summary = "Pad operator";
+  let summary = "Pad operation";
   let description = [{
     Expands `operand` by padding around the tensor as well as between the
     elements of the tensor with the given `padding_value`.
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -2669,7 +2669,8 @@ def StableHLO_ReverseOp: StableHLO_Op<"reverse",
 }
 
 def StableHLO_PadOp: StableHLO_ShapedInterfaceOp<"pad",
-      [Pure, SameOperandsAndResultElementType,
+      [Pure, SameOperandsAndResultElementType /*pad_c1*/,
+      AllTypesMatch<["edge_padding_low", "edge_padding_high", "interior_padding"]> /*pad_c2_i3_i4_i5*/,
       DeclareOpInterfaceMethods<InferTypeOpInterface>]> {
   let summary = "Pad operation";
   let description = [{
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -2686,11 +2686,11 @@ def StableHLO_PadOp: StableHLO_ShapedInterfaceOp<"pad",
     ```
   }];
   let arguments = (ins
-    HLO_Tensor:$operand,
-    HLO_Tensor:$padding_value,
-    I64ElementsAttr:$edge_padding_low,
-    I64ElementsAttr:$edge_padding_high,
-    I64ElementsAttr:$interior_padding
+    HLO_Tensor:$operand /*pad_i1*/,
+    HLO_Tensor:$padding_value /*pad_i2*/,
+    I64ElementsAttr:$edge_padding_low /*pad_i3*/,
+    I64ElementsAttr:$edge_padding_high /*pad_i4*/,
+    I64ElementsAttr:$interior_padding /*pad_i5*/
   );
 
   let results = (outs HLO_Tensor);
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -2435,23 +2435,17 @@ LogicalResult inferPadOp(std::optional<Location> location, Value operand,
                              "tensor, is rank ",
                              padType.getRank());
 
-  // PadOp_C2
+  // pad_i3
+  if (edgePaddingLow.getType().getRank() != 1)
+    return emitOptionalError(location, "start_indices has rank ",
+                             edgePaddingLow.getType().getRank(),
+                             " instead of required rank 1");
+
+  // pad_c2
   int64_t rank = inputType.getRank();
   if (edgePaddingLow.getType().getNumElements() != rank)
     return emitOptionalError(location, "edge_padding_low length (",
                              edgePaddingLow.getType().getNumElements(),
-                             ") must match operand rank (", rank, ")");
-
-  // PadOp_C2
-  if (edgePaddingHigh.getType().getNumElements() != rank)
-    return emitOptionalError(location, "edge_padding_high length (",
-                             edgePaddingHigh.getType().getNumElements(),
-                             ") must match operand rank (", rank, ")");
-
-  // PadOp_C2
-  if (interiorPadding.getType().getNumElements() != rank)
-    return emitOptionalError(location, "interior_padding length (",
-                             interiorPadding.getType().getNumElements(),
                              ") must match operand rank (", rank, ")");
 
   auto inputShape = inputType.getShape();
@@ -2465,7 +2459,7 @@ LogicalResult inferPadOp(std::optional<Location> location, Value operand,
         edgePaddingHigh.getValues<APInt>()[i].getSExtValue();
     int64_t paddingInteriorVal =
         interiorPadding.getValues<APInt>()[i].getSExtValue();
-    // PadOp_C3
+    // pad_c3
     if (paddingInteriorVal < 0)
       return emitOptionalError(
           location,
@@ -2480,7 +2474,7 @@ LogicalResult inferPadOp(std::optional<Location> location, Value operand,
           operandSizeOrBound + paddingLowVal + paddingHighVal +
           std::max<int64_t>(operandSizeOrBound - 1, 0LL) * paddingInteriorVal;
 
-      // PadOp_C4
+      // pad_c4
       if (resultSizeOrBound < 0) {
         auto sizeOrBound = isStaticDim ? "size" : "bound";
         return emitOptionalError(location, "Padding result in negative ",
@@ -2490,7 +2484,7 @@ LogicalResult inferPadOp(std::optional<Location> location, Value operand,
     }
   }
 
-  // PadOp_C1
+  // pad_c1
   inferredReturnTypes.push_back(RankedTensorType::get(
       resultShape, inputType.getElementType(),
       boundsToEncoding(inputType.getEncoding(), resultBounds)));
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -2435,17 +2435,20 @@ LogicalResult inferPadOp(std::optional<Location> location, Value operand,
                              "tensor, is rank ",
                              padType.getRank());
 
+  // PadOp_C2
   int64_t rank = inputType.getRank();
   if (edgePaddingLow.getType().getNumElements() != rank)
     return emitOptionalError(location, "edge_padding_low length (",
                              edgePaddingLow.getType().getNumElements(),
                              ") must match operand rank (", rank, ")");
 
+  // PadOp_C2
   if (edgePaddingHigh.getType().getNumElements() != rank)
     return emitOptionalError(location, "edge_padding_high length (",
                              edgePaddingHigh.getType().getNumElements(),
                              ") must match operand rank (", rank, ")");
 
+  // PadOp_C2
   if (interiorPadding.getType().getNumElements() != rank)
     return emitOptionalError(location, "interior_padding length (",
                              interiorPadding.getType().getNumElements(),
@@ -2462,6 +2465,7 @@ LogicalResult inferPadOp(std::optional<Location> location, Value operand,
         edgePaddingHigh.getValues<APInt>()[i].getSExtValue();
     int64_t paddingInteriorVal =
         interiorPadding.getValues<APInt>()[i].getSExtValue();
+    // PadOp_C3
     if (paddingInteriorVal < 0)
       return emitOptionalError(
           location,
@@ -2476,6 +2480,7 @@ LogicalResult inferPadOp(std::optional<Location> location, Value operand,
           operandSizeOrBound + paddingLowVal + paddingHighVal +
           std::max<int64_t>(operandSizeOrBound - 1, 0LL) * paddingInteriorVal;
 
+      // PadOp_C4
       if (resultSizeOrBound < 0) {
         auto sizeOrBound = isStaticDim ? "size" : "bound";
         return emitOptionalError(location, "Padding result in negative ",
@@ -2484,6 +2489,8 @@ LogicalResult inferPadOp(std::optional<Location> location, Value operand,
       (isStaticDim ? resultShape : resultBounds)[i] = resultSizeOrBound;
     }
   }
+
+  // PadOp_C1
   inferredReturnTypes.push_back(RankedTensorType::get(
       resultShape, inputType.getElementType(),
       boundsToEncoding(inputType.getEncoding(), resultBounds)));
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -2429,20 +2429,30 @@ LogicalResult inferPadOp(std::optional<Location> location, Value operand,
   auto inputType = operand.getType().cast<RankedTensorType>();
   auto padType = paddingValue.getType().cast<RankedTensorType>();
 
+  // pad_i2
   if (padType.getRank() != 0)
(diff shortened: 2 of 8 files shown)


subhankarshah (approved):
LGTM

stablehlo/reference/Ops.cpp
```diff
@@ -157,6 +157,25 @@ Tensor evalOrOp(const Tensor &lhs, const Tensor &rhs, Type resultType) {
...
+       ++resItr)
+    result.set(*resItr, paddingValue.get({}));
+  for (auto opItr = operand.index_begin(); opItr != operand.index_end();
+       ++opItr) {
+    SmallVector<int64_t> adjustedIdx(result.getType().getRank());
+    for (int64_t i = 0; i < operand.getType().getRank(); ++i)
+      adjustedIdx[i] =
+          edgePaddingLow[i] + (*opItr)[i] * (interiorPadding[i] + 1);
+    if (failed(verifyIndex(result.getType().getShape(), adjustedIdx))) continue;
+    result.set(adjustedIdx, operand.get(*opItr));
```
sdasgup3, line 174: 1. No initialization of result Tensor. Ideally a `Tensor` has to be immutatble (https://github.com/openxla/stablehlo/issues/233).
2. IMO this looks more closer to the spec. 

```
#include<numeric>
 // not initialization of result
 SmallVector<int64_t> dims(operand.getType().getRank());
  std::iota(dims.begin(), dims.end(), 0);
  if (llvm::any_of(dims, [](int64_t dim)
        return (*resItr)[dim] !=
               edgePaddingLow[dim] + (*opItr)[dim] * (interiorPadding[i] + 1);
      }))
    result.set(*resItr, paddingValue.get({}));
  else 
    result.set(*resItr, operand.get(*opItr));
```

Please let me know if you have other concerns

stablehlo/tests/interpret_pad.mlir
```diff
@@ -0,0 +1,86 @@
+// RUN: stablehlo-interpreter --interpret -split-input-file %s | FileCheck %s
```
sdasgup3, line 1: A test for creating zero dimension size result would be interesting.

stablehlo/tests/interpret_pad.mlir
```diff
@@ -0,0 +1,86 @@
...
+  // CHECK-NEXT: -1 : i64
+  // CHECK-NEXT: -1 : i64
+  // CHECK-NEXT: -1 : i64
+}
+
+// -----
+
+// CHECK-LABEL: Evaluated results of function: pad_negative_padding
+func.func @pad_negative_padding() -> tensor<2x7xi64> {
+  %operand = stablehlo.constant dense<[[0, 1, 2, 3, 0],
```
sdasgup3, line 63: Can we keep a test with a mix of positive and negative padding. That way we can keep the essence of both the tests with out writing two tests.

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -5499,25 +5462,26 @@ func.func @pad(%arg0: tensor<1x2x3xf16>, %arg1: tensor<f16>) -> tensor<8x8x8xf16
...
+    interior_padding = dense<0> : tensor<4xi64>
+  } : (tensor<?x48x48x32xf32>, tensor<f32>) -> tensor<?x48x48x48xf32>
+  func.return %1 : tensor<?x48x48x48xf32>
 }
 
 // -----
 
-func.func @pad(%arg0: tensor<1x2x3xf16>, %arg1: tensor<f16>) -> tensor<2x4x7xf16> {
-  // expected-error@+1 {{Padding result in negative size for dimension 2}}
+func.func @pad_i1(%arg0: tensor<1x2x3xf16>, %arg1: tensor<2xf16>) -> tensor<2x4x7xf16> {
```
sdasgup3, line 5478: same comment as above

stablehlo/dialect/StablehloOps.td
```diff
@@ -2669,9 +2669,10 @@ def StableHLO_ReverseOp: StableHLO_Op<"reverse",
 }
 
 def StableHLO_PadOp: StableHLO_ShapedInterfaceOp<"pad",
-      [Pure, SameOperandsAndResultElementType,
+      [Pure, SameOperandsAndResultElementType /*pad_c1*/,
+      AllTypesMatch<["edge_padding_low", "edge_padding_high", "interior_padding"]> /*pad_c2_i3_i4_i5*/,
```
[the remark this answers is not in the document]
ghpvnist: Resolved offline.

stablehlo/reference/Ops.cpp
```diff
@@ -157,6 +157,25 @@ Tensor evalOrOp(const Tensor &lhs, const Tensor &rhs, Type resultType) {
...
+Tensor evalPadOp(const Tensor &operand, const Tensor &paddingValue,
+                 ArrayRef<int64_t> edgePaddingLow,
+                 ArrayRef<int64_t> interiorPadding, Type resultType) {
+  Tensor result(resultType);
+  for (auto resItr = result.index_begin(); resItr != result.index_end();
+       ++resItr)
+    result.set(*resItr, paddingValue.get({}));
+  for (auto opItr = operand.index_begin(); opItr != operand.index_end();
+       ++opItr) {
+    SmallVector<int64_t> adjustedIdx(result.getType().getRank());
```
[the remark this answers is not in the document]
ghpvnist: I'll rename it `resultIdx` since at the end, it is used as an index to set the element in the result tensor.

stablehlo/reference/Ops.cpp
ghpvnist: Though the current code is initializing the padding values first, I think it still stays close to the spec. I've considered writing without initializing the tensor, but the way the spec is written, though concise, needs to loop through both resultIdx and operandIdx resulting in quadratic complexity. If we initialize the result early, we run at O(max(operand, result)) linear time. I've also looked into your suggestion, but it has the problem of which tensor to iterate to make it linear: result or operand or both. I haven't considered how to do this in linear time in depth, so I may be missing something, so let me know if you have suggestions in mind.

stablehlo/tests/interpret_pad.mlir
ghpvnist: Per the guidelines, let's not have this test since there is no special handling of empty result tensors needed (we end up iterating through the operand with no action, but that is okay since we don't plan on doing any sort of optimizations at this point).

sdasgup3: Makes sense!

stablehlo/reference/Ops.cpp
sdasgup3: I agree with your observation. Checking if a result index matches with an expression of operand index and padding attributes requires iterating over all the operands. That would unnecessarily complicates the code.  Computing the result index from an operand index (which is what you are doing) seems right and concise. Please drop my suggestion.

sdasgup3 (approved):

stablehlo/reference/Ops.cpp
```diff
@@ -157,6 +157,25 @@ Tensor evalOrOp(const Tensor &lhs, const Tensor &rhs, Type resultType) {
...
+       ++resultItr)
+    result.set(*resultItr, paddingValue.get({}));
+  for (auto operandItr = operand.index_begin();
+       operandItr != operand.index_end(); ++operandItr) {
+    SmallVector<int64_t> resultIdx(result.getType().getRank());
+    for (auto i = 0; i < operand.getType().getRank(); ++i)
+      resultIdx[i] =
+          edgePaddingLow[i] + (*operandItr)[i] * (interiorPadding[i] + 1);
+    if (succeeded(verifyIndex(result.getType().getShape(), resultIdx)))
+      result.set(resultIdx, operand.get(*operandItr));
```
sdasgup3, line 174: One more comment:

Rather than initializing the `result`, let us add pad value to result in the else block.

sdasgup3 (commented):
one minor comment.

stablehlo/reference/Ops.cpp
ghpvnist: Actually, we cannot do that here since the `verifyIndex` functions tells us whether `resultIdx` is out of bounds, so if this expression is false, then the `resultIdx` is not a valid index in the `result` tensor.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -2464,28 +2464,35 @@ LogicalResult inferPadOp(std::optional<Location> location, Value operand,
...
                              padType.getRank());
 
+  // pad_c2, pad_i3, pad_i4, pad_i5
+  if (edgePaddingLow.getType() != edgePaddingHigh.getType() ||
+      edgePaddingLow.getType() != interiorPadding.getType())
+    return emitOptionalError(
+        location, "edge_padding_low, edge_padding_high, ",
+        "and interior_padding must have the same type but got: ",
+        edgePaddingLow.getType(), ", ", edgePaddingHigh.getType(), ", and ",
+        interiorPadding.getType());
```
burmako, line 2481: Since we're refactoring this, looks like this check can be expressed via `AllTypesMatch` in the ODS? It'll reduce the amount of C++ code and will allow us to avoid writing some tests.

stablehlo/reference/Ops.cpp
```diff
@@ -157,6 +157,25 @@ Tensor evalOrOp(const Tensor &lhs, const Tensor &rhs, Type resultType) {
   return result;
 }
 
+Tensor evalPadOp(const Tensor &operand, const Tensor &paddingValue,
+                 ArrayRef<int64_t> edgePaddingLow,
+                 ArrayRef<int64_t> interiorPadding, Type resultType) {
+  Tensor result(resultType);
+  for (auto resultItr = result.index_begin(); resultItr != result.index_end();
```
burmako, line 164: I think that in nearby functions we're calling this variable `resultIt`? Same for `operandIt` below.

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -5339,125 +5339,138 @@ func.func @conv_i4(%arg0: tensor<64x8x8x8xi4>, %arg1: tensor<4x4x8x32xi4>) -> te
...
-  func.return %1 : tensor<?x48x48x48xf32>
+func.func @pad_c2(%arg0: tensor<1x2x3xf16>, %arg1: tensor<f16>) -> tensor<2x4x7xf16> {
+  // expected-error@+1 {{edge_padding_low length (2) must match operand rank (3)}}
+  %0 = "stablehlo.pad"(%arg0, %arg1) {
+    edge_padding_low = dense<[0, 1]> : tensor<2xi64>,
+    edge_padding_high = dense<[1, 1]> : tensor<2xi64>,
+    interior_padding = dense<[0, 0]> : tensor<2xi64>
+  } : (tensor<1x2x3xf16>, tensor<f16>) -> tensor<2x4x7xf16>
+  func.return %0 : tensor<2x4x7xf16>
 }
```
burmako, line 5363: Based on the PR description, we also need `pad_c2` tests for edge_padding_high and interior_padding?

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -5339,125 +5339,138 @@ func.func @conv_i4(%arg0: tensor<64x8x8x8xi4>, %arg1: tensor<4x4x8x32xi4>) -> te
...
+    interior_padding = dense<1> : tensor<i64>
   } : (tensor<1x2x3xf16>, tensor<f16>) -> tensor<2x4x7xf16>
   func.return %0 : tensor<2x4x7xf16>
 }
 
 // -----
 
-func.func @pad(%arg0: tensor<1x2x3xf16>, %arg1: tensor<f16>) -> tensor<8x8x8xf16> {
-  // expected-error@+1 {{'stablehlo.pad' op inferred type(s) 'tensor<2x4x7xf16>' are incompatible with return type(s) of operation 'tensor<8x8x8xf16>'}}
+func.func @pad_i3_i4(%arg0: tensor<1x2x3xf16>, %arg1: tensor<f16>) -> tensor<2x4x7xf16> {
```
burmako, line 5439: This test looks like `pad_i4` to me, to be honest, given that it's testing edge_padding_high.

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -5339,125 +5339,138 @@ func.func @conv_i4(%arg0: tensor<64x8x8x8xi4>, %arg1: tensor<4x4x8x32xi4>) -> te
...
+    interior_padding = dense<1> : tensor<1xi64>
+  } : (tensor<1x2x3xf16>, tensor<f16>) -> tensor<2x4x7xf16>
+  func.return %0 : tensor<2x4x7xf16>
 }
 
 // -----
 
-func.func @pad(%arg0: tensor<1x2x3xf16>, %arg1: tensor<f16>) -> tensor<2x4x3xf16> {
-  // expected-error@+1 {{Interior padding cannot be negative: -1}}
+func.func @pad_i3_i5(%arg0: tensor<1x2x3xf16>, %arg1: tensor<f16>) -> tensor<2x4x7xf16> {
```
burmako, line 5451: Same as above, but for `pad_i5`.

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -5339,125 +5339,138 @@ func.func @conv_i4(%arg0: tensor<64x8x8x8xi4>, %arg1: tensor<4x4x8x32xi4>) -> te
...
+    interior_padding = dense<1> : tensor<2xi64>
+  } : (tensor<1x2x3xf16>, tensor<f16>) -> tensor<2x4x7xf16>
+  func.return %0 : tensor<2x4x7xf16>
 }
 
 // -----
 
-func.func @pad(%arg0: tensor<1x2x3xf16>, %arg1: tensor<f16>) -> tensor<2x4x7xf16> {
-  // expected-error@+1 {{Padding result in negative size for dimension 2}}
+func.func @pad_i4(%arg0: tensor<1x2x3xf16>, %arg1: tensor<f16>) -> tensor<2x4x7xf16> {
```
burmako, line 5463: I think this test is redundant, if we compare it with the test plan in the PR description.

burmako (changes requested):

stablehlo/reference/Ops.cpp
```diff
@@ -157,6 +157,25 @@ Tensor evalOrOp(const Tensor &lhs, const Tensor &rhs, Type resultType) {
...
+       ++resultItr)
+    result.set(*resultItr, paddingValue.get({}));
+  for (auto operandItr = operand.index_begin();
+       operandItr != operand.index_end(); ++operandItr) {
+    SmallVector<int64_t> resultIdx(result.getType().getRank());
+    for (auto i = 0; i < operand.getType().getRank(); ++i)
+      resultIdx[i] =
+          edgePaddingLow[i] + (*operandItr)[i] * (interiorPadding[i] + 1);
+    if (succeeded(verifyIndex(result.getType().getShape(), resultIdx)))
```
burmako, line 173: I don't think we need to call `verifyIndex` here: 1) in these functions, we can assume that ops have passed the verifier, 2) verifyIndex is called in Tensor::set anyway.

burmako (changes requested):

stablehlo/reference/Ops.cpp
```diff
@@ -157,6 +157,25 @@ Tensor evalOrOp(const Tensor &lhs, const Tensor &rhs, Type resultType) {
...
+  Tensor result(resultType);
+  for (auto resultItr = result.index_begin(); resultItr != result.index_end();
+       ++resultItr)
+    result.set(*resultItr, paddingValue.get({}));
+  for (auto operandItr = operand.index_begin();
+       operandItr != operand.index_end(); ++operandItr) {
+    SmallVector<int64_t> resultIdx(result.getType().getRank());
+    for (auto i = 0; i < operand.getType().getRank(); ++i)
+      resultIdx[i] =
+          edgePaddingLow[i] + (*operandItr)[i] * (interiorPadding[i] + 1);
```
burmako, line 172: Something to consider in the future - I think it might be useful to have `IndexSpaceIterator` return `Index` instead of `ArrayRef<int64_t>`, and then have this `Index` class overload operators like `*` or `+`. cc @sdasgup3.

stablehlo/dialect/TypeInference.cpp
ghpvnist: I originally added the `AllTypesMatch` Trait, but @zhouxin913 suggested not to use this since dynamism work will not be compatible with this Trait.

stablehlo/reference/Ops.cpp
ghpvnist: `verifyIndex` does need to be called since the `resultIdx` is constructed by blindly applying padding to the `operandItr`, and there is not guarantee that the `resultIdx` is in bounds. This means the `resultIndex` can be out of bounds for some `operandItr` and calling `Tensor::set` on this would error out, which is the case when we have negative padding and not all `operandItr` maps to `resultIdx`.

stablehlo/dialect/TypeInference.cpp
burmako: In this specific case, it'll be fine because `edge_padding_low`, `edge_padding_high` and `interior_padding` are a special case - they are not going to have dynamic types.

stablehlo/tests/ops_stablehlo.mlir
ghpvnist: I've added the `AllTypesMatch` trait and this test is not necessary as noted in the PR description.

stablehlo/tests/ops_stablehlo.mlir
ghpvnist: https://github.com/openxla/stablehlo/pull/1005#discussion_r1104954411

stablehlo/tests/ops_stablehlo.mlir
ghpvnist: https://github.com/openxla/stablehlo/pull/1005#discussion_r1104954411

stablehlo/tests/ops_stablehlo.mlir
ghpvnist: https://github.com/openxla/stablehlo/pull/1005#discussion_r1104954411

burmako (approved):

PR #1006, opened 2023-01-28
Add interpreter for ReverseOp

### LINKED ISSUE #985: Add interpreter for reverse

ghpvnist:
closes #985

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -4143,16 +4143,6 @@ and produces a `result` tensor. More formally,
 #### Examples
 
 ```mlir
-// Reverse along dimension 0
-
-// %operand = [[1, 2], [3, 4], [5, 6]]
-%result = "stablehlo.reverse"(%operand) {
-  dimensions = dense<0> : tensor<i64>
-} : (tensor<3x2xi32>) -> tensor<3x2xi32>
-// %result: [[5, 6], [3, 4], [1, 2]]
-
-// Reverse along dimension 1
-
 // %operand = [[1, 2], [3, 4], [5, 6]]
 %result = "stablehlo.reverse"(%operand) {
   dimensions = dense<1> : tensor<i64>
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -4150,6 +4150,8 @@ and produces a `result` tensor. More formally,
 // %result: [[2, 1], [4, 3], [6, 5]]
 ```
 
+&nbsp;[More Examples](../stablehlo/tests/interpret_reverse.mlir)
+
 ### rng
 
 #### Semantics
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -3725,20 +3725,24 @@ LogicalResult verifyReshapeOp(std::optional<Location> location, Value operand,
 
 LogicalResult verifyReverseOp(std::optional<Location> location, Value operand,
                               DenseIntElementsAttr dimensions) {
+  // reverse_i2
+  if (dimensions.getType().getRank() != 1)
+    return emitOptionalError(location, "dimensions has rank ",
+                             dimensions.getType().getRank(),
+                             " instead of required rank 1.");
   auto dims = dimensions.getValues<int64_t>();
   llvm::SmallDenseSet<int64_t> uniqueDims(dims.begin(), dims.end());
-  // ReverseOp_C2
+  // reverse_c2
   if (uniqueDims.size() != dims.size())
     return emitOptionalError(location,
                              "dimensions should be unique. Got: ", dims);
   auto operandTy = operand.getType().dyn_cast<RankedTensorType>();
   for (int64_t dim : uniqueDims) {
-    // ReverseOp_C3
+    // reverse_c3
     if (dim < 0)
       return emitOptionalError(
           location,
           "all dimensions should be non-negative. Got dimension: ", dim, ".");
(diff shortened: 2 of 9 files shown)


subhankarshah (approved):
LGTM, neat.

stablehlo/reference/Ops.cpp
```diff
@@ -178,6 +178,22 @@ Tensor eval(ReshapeOp op, const Tensor &operand) {
...
+       resultIt != result.index_end(); ++resultIt, ++operandIt) {
+    auto operandIdx = (*operandIt).vec();
+    for (auto dim : dimensions) {
+      operandIdx[dim] = -(operandIdx[dim] - (shape[dim] - 1));
+    }
+    result.set(*resultIt, operand.get(operandIdx));
+  }
+  return result;
+}
+
```
sdasgup3, line 196: Suggestion:
1. Avoid loop in collecting the dimensions in a `SmallVector`.
2. As per the spec, given an index of result we can derive the index of the operand. With that said, 
  -  we can avoid the second iterator. 
  -  make the implementation follow more closely to the spec
3. Prefer `SmallVector` over `std::vector`

```
Tensor eval(ReverseOp op, const Tensor &operand) {
  Tensor result(op.getType());
  auto shape = op.getType().getShape();
  auto dimensions = op.getDimensions().getValues<int64_t>();

       ++resultIt) {
    SmallVector<int64_t> operandIdx(*resultIt);
    for (auto dim : dimensions) {
      operandIdx[dim] = (shape[dim] - 1) - operandIdx[dim];
    }
    result.set(*resultIt, operand.get(operandIdx));
  }
  return result;
}

```

sdasgup3:
Other than the suggested change, the PR LGTM.

sdasgup3 (approved):

docs/spec.md
```diff
@@ -4143,16 +4143,6 @@ and produces a `result` tensor. More formally,
 #### Examples
 
 ```mlir
```
sdasgup3, line 4145: Link to test file `interpreter_reverse.mlir` is missing

stablehlo/reference/Ops.cpp
```diff
@@ -178,6 +178,20 @@ Tensor eval(ReshapeOp op, const Tensor &operand) {
   return result;
 }
 
+Tensor eval(ReverseOp op, const Tensor &operand) {
```
burmako, line 181: This will need to be rebased on top of #1007.

stablehlo/reference/Ops.cpp
```diff
@@ -178,6 +178,20 @@ Tensor eval(ReshapeOp op, const Tensor &operand) {
...
+  Tensor result(op.getType());
+  auto shape = op.getType().getShape();
```
burmako, line 183: `resultShape` to make it clear what shape exactly this is referring to?

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -3046,7 +3056,7 @@ func.func @reverse_duplicate_dimensions(%operand: tensor<3x2xi32>) -> tensor<3x2
 
 // -----
 
-func.func @reverse_invalid_dimensions_unranked(%operand: tensor<*xi32>) -> tensor<*xi32> {
+func.func @reverse_C3(%operand: tensor<*xi32>) -> tensor<*xi32> {
```
burmako, line 3059: How about calling this `@reverse_c3` to align with the snake case naming convention used in the file?

stablehlo/tests/interpret_reverse.mlir
```diff
@@ -0,0 +1,53 @@
+// RUN: stablehlo-interpreter --interpret -split-input-file %s | FileCheck %s
```
burmako, line 1: Do we need all the three tests here or just one is sufficient, similarly to `ops_stablehlo.mlir` which has just one positive test?

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3728,15 +3728,18 @@ LogicalResult verifyReverseOp(std::optional<Location> location, Value operand,
...
     return emitOptionalError(location,
                              "dimensions should be unique. Got: ", dims);
   auto operandTy = operand.getType().dyn_cast<RankedTensorType>();
   for (int64_t dim : uniqueDims) {
+    // ReverseOp_C3
     if (dim < 0)
       return emitOptionalError(
           location,
           "all dimensions should be non-negative. Got dimension: ", dim, ".");
+    // ReverseOp_C3
```
burmako, line 3742: Let's perhaps drop this line, given that we're already under `ReverseOp_C3`?

burmako (changes requested):

stablehlo/tests/ops_stablehlo.mlir
ghpvnist: I followed the guideline in the [checklist](https://github.com/openxla/stablehlo/blob/main/docs/reference_checklist.md). Perhaps tests (ops_stablehlo.mlir and interpret_foo.mlir), comments, test names should all be snake lowercase without appending "op"?
```
// slice_c1

func.func @slice_c2
```

burmako: Sounds reasonable to me. cc @sdasgup3

sdasgup3: LGTM

burmako (approved):

Issue #454, opened 2022-11-05
Make sure that examples in the spec/ODS conform to spec_checklist.md


burmako:
See #453 for context and for initial work in this area.

sdasgup3:
Just to document:

These are one still having multiple examples:
``` 
stablehlo.divide
stablehlo.or
stablehlo.xor
stablehlo.not
stablehlo.exponential
stablehlo.iota
stablehlo.log
stablehlo.logistic
stablehlo.negate
stablehlo.reverse
stablehlo.slice
stablehlo.sort
stablehlo.sqrt
```

burmako:
"Performs element-wise bitwise AND of two tensors..." => "Performs element-wise bitwise or logical AND of two tensors" should also apply to not/or/xor.

PR #1008, opened 2023-01-28
Add interpreter for DivideOp

### LINKED ISSUE #971: Add interpreter for divide

subhankarshah:
closes #971 

```
(I1) lhs is a tensor of integer, floating-point, or complex type.
(I2) rhs is a tensor of integer, floating-point, or complex type.
(C1) lhs, rhs, and result have the same type.
```

```
(I1) lhs is not a tensor of integer, floating-point, or complex type. (covered by ODS)
(I2) rhs is not a tensor of integer, floating-point, or complex type. (covered by ODS)
(C1) lhs, rhs, and result do not have the same type. (covered by ODS)
```

Add a binary op version of mapWithUpcastToDouble

### DIFF
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -739,7 +739,8 @@ def StableHLO_ComplexOp: StableHLO_BinaryElementwiseOp<"complex", [Pure,
 }
 
 def StableHLO_DivOp : StableHLO_BinaryElementwiseOp<"divide", [Pure,
-    HLO_CompatibleOperandsAndResultType /*div_c1*/], HLO_IntFpOrComplexTensor> {
+    HLO_CompatibleOperandsAndResultType /* div_c1 */],
+    HLO_IntFpOrComplexTensor /* div_i1, div_i2 */> {
   let summary = "Divide operation";
   let description = [{
     Performs element-wise division of dividend `lhs` and divisor `rhs` tensors
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -738,8 +738,8 @@ def StableHLO_ComplexOp: StableHLO_BinaryElementwiseOp<"complex", [Pure,
   }];
 }
 
-def StableHLO_DivOp : StableHLO_BinaryElementwiseOp<"divide",
-      [Pure, HLO_CompatibleOperandsAndResultType], HLO_IntFpOrComplexTensor> {
+def StableHLO_DivOp : StableHLO_BinaryElementwiseOp<"divide", [Pure,
+    HLO_CompatibleOperandsAndResultType /*div_c1*/], HLO_IntFpOrComplexTensor> {
   let summary = "Division operation";
   let description = [{
     Performs element-wise division of dividend `lhs` and divisor `rhs` tensors
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -740,7 +740,7 @@ def StableHLO_ComplexOp: StableHLO_BinaryElementwiseOp<"complex", [Pure,
 
 def StableHLO_DivOp : StableHLO_BinaryElementwiseOp<"divide", [Pure,
     HLO_CompatibleOperandsAndResultType /*div_c1*/], HLO_IntFpOrComplexTensor> {
-  let summary = "Division operation";
+  let summary = "Divide operation";
   let description = [{
     Performs element-wise division of dividend `lhs` and divisor `rhs` tensors
     and produces a `result` tensor.
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -740,7 +740,7 @@ def StableHLO_ComplexOp: StableHLO_BinaryElementwiseOp<"complex", [Pure,
 
 def StableHLO_DivOp : StableHLO_BinaryElementwiseOp<"divide",
       [Pure, HLO_CompatibleOperandsAndResultType], HLO_IntFpOrComplexTensor> {
-  let summary = "Division operator";
+  let summary = "Division operation";
   let description = [{
     Performs element-wise division of dividend `lhs` and divisor `rhs` tensors
     and produces a `result` tensor.
@@ -750,7 +750,7 @@ def StableHLO_DivOp : StableHLO_BinaryElementwiseOp<"divide",
 
     Example:
     ```mlir
-    %result = stablehlo.divide %lhs, %rhs : tensor<2xi32>
+    %result = stablehlo.divide %lhs, %rhs : tensor<4xf32>
     ```
   }];
 }
--- a/stablehlo/reference/Element.cpp
+++ b/stablehlo/reference/Element.cpp
@@ -317,6 +317,19 @@ Element Element::operator+(const Element &other) const {
       });
 }
 
+Element Element::operator/(const Element &other) const {
+  return mapWithUpcastToDouble(
+      *this, other, [&](APInt lhs, APInt rhs) { return lhs.sdiv(rhs); },
+      [&](APInt lhs, APInt rhs) { return lhs.udiv(rhs); },
+      [](bool lhs, bool rhs) -> bool {
+        llvm::report_fatal_error("-bool is unsupported");
+      },
+      [](double lhs, double rhs) { return lhs / rhs; },
+      [](std::complex<double> lhs, std::complex<double> rhs) {
+        return lhs / rhs;
+      });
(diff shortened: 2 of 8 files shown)


stablehlo/reference/Element.cpp
```diff
@@ -30,15 +30,21 @@ namespace stablehlo {
 
 namespace {
```
sdasgup3, line 31: Thanks for the suggestion!
What is the motivation behind the splitting the `integerFn` into its variants? Is there an op which might need this in future.  Currently we use a common`integerFn` function which handles the variants differently  or uniformly based on the semantics of a particular op.  So far, none of the ops, except for`minimum` and `maximum` need to use different treatment for signed/unsigned variants. The problems with adding the change without many ops using it is that all the ops now has to provide two callbacks which in most of the cases will be the same or throw error for the type being unsupported.

Having said that, `convert` would be a potential op which might need to add additional types in `mapWithUpcastToDouble`.

My proposal is to revisit augmenting these map* functions in the context of an op on a need basis.

Please let me know your opinion.

subhankarshah: I already have an a pr which uses this, I just split the change.

subhankarshah: https://github.com/subhankarshah/stablehlo/pull/1

subhankarshah: I wanted to stack the changes in different prs. This is a pr in my fork.

sdasgup3: Thanks for sharing. I am afraid that I am still unsure if the handling of  `divOp` is any different from [maxOp](https://github.com/openxla/stablehlo/blob/e28fac48833374bd7686407718953d3801f8dd79/stablehlo/reference/Element.cpp#L293) and we need the split for integer variants. 
But at the same time my call might be biased. 

How about seeking opinion from @burmako  and we can continue with the feedback once we get the green light.

stablehlo/reference/Element.cpp
```diff
@@ -190,6 +257,22 @@ Element Element::operator+(const Element &other) const {
...
 
+Element Element::operator/(const Element &other) const {
+  return mapWithUpcastToDouble(
+      *this, other,
+      [&](APInt lhs, APInt rhs) {
+        return isSupportedSignedIntegerType(other.getType()) ? lhs.sdiv(rhs)
+                                                             : lhs.udiv(rhs);
+      },
+      [](bool lhs, bool rhs) -> bool {
+        llvm::report_fatal_error("-bool is unsupported");
```
ghpvnist, line 268: Artifact from refactoring? Here and below.
```
llvm::report_fatal_error("bool / bool is unsupported");
```

stablehlo/tests/interpret_divide.mlir
```diff
@@ -0,0 +1,57 @@
+// RUN: stablehlo-interpreter --interpret -split-input-file %s | FileCheck %s
+
+// CHECK-LABEL: Evaluated results of function: div_op_test_ui64
+func.func @div_op_test_ui64() -> tensor<4xui64> {
+  %0 = stablehlo.constant dense<[17, 18, 19, 20]> : tensor<4xui64>
```
ghpvnist, line 5: SSA values should be renamed i.e. %operand, %result, etc.

stablehlo/tests/interpret_divide.mlir
```diff
@@ -0,0 +1,57 @@
+// RUN: stablehlo-interpreter --interpret -split-input-file %s | FileCheck %s
+
+// CHECK-LABEL: Evaluated results of function: div_op_test_ui64
+func.func @div_op_test_ui64() -> tensor<4xui64> {
```
ghpvnist, line 4: I believe there's no guideline for ordering tests other than the generic types (bool, integer, floating-point, and complex), so let's order the rest alphabetically: i.e., `si64`, `ui64`, `f64`, and `c128`.

ghpvnist (changes requested):
Refer to our [reference_checklist](https://github.com/openxla/stablehlo/blob/main/docs/reference_checklist.md) to ensure that there's no missing files.

ghpvnist (approved):
Can you also update the PR description to talk about the additional overloaded function added.

sdasgup3 (commented):
It would be better if we update the description with why we need to introduce the function (Just to help the next review get aware of the context)?
 
Given that `APFLoat`  have a `operator/` for float, we could have used the just the map function for non complex types. The downside is to inline the upcast-to-double logic for the complex type. 

I am assuming that the reason for updating the existing function is that the complex types  would be trivial to implement using up-cast to double. But using update-to-double for non-float or on-complex does not seem right mainly because contrary to the function name we are not doing any upcasts.

Please let know if I am missing anything. 
cc @burmako

sdasgup3:
Also we can add a simple testing strategy similar to https://github.com/openxla/stablehlo/pull/1079#issue-1574895499

ghpvnist:

stablehlo/dialect/StablehloOps.td
```diff
@@ -733,9 +733,9 @@ def StableHLO_ComplexOp: StableHLO_BinaryElementwiseOp<"complex", [Pure,
   }];
 }
 
-def StableHLO_DivOp : StableHLO_BinaryElementwiseOp<"divide",
-      [Pure, HLO_CompatibleOperandsAndResultType], HLO_IntFpOrComplexTensor> {
-  let summary = "Division operator";
+def StableHLO_DivOp : StableHLO_BinaryElementwiseOp<"divide", [Pure,
+    HLO_CompatibleOperandsAndResultType /*div_c1*/], HLO_IntFpOrComplexTensor> {
+  let summary = "Divide operation";
```
sdasgup3, line 738: Add input constraints as per  the checklist.

sdasgup3 (approved):

stablehlo/reference/Element.cpp
```diff
@@ -100,11 +100,23 @@ Element map(const Element &lhs, const Element &rhs, IntegerFn integerFn,
                                      debugString(type).c_str()));
 }
 
-template <typename FloatFn, typename ComplexFn>
-Element mapWithUpcastToDouble(const Element &el, FloatFn floatFn,
+template <typename IntegerFn, typename BooleanFn, typename FloatFn,
+          typename ComplexFn>
+Element mapWithUpcastToDouble(const Element &el, IntegerFn integerFn,
+                              BooleanFn boolFn, FloatFn floatFn,
```
burmako, line 106: "using update-to-double for non-float or on-complex does not seem right mainly because contrary to the function name we are not doing any upcasts". Yes I agree. It's not evident how is the "upcast to double" part supposed to work for e.g. boolean types. I think it would be better to implement support for division without using map.

burmako (changes requested):

stablehlo/reference/Element.cpp
subhankarshah: I needed the upcast function, but they do not return optional result instead throw errors and failures, so I could not use them as is.

stablehlo/reference/Element.cpp
```diff
@@ -190,6 +231,54 @@ Element Element::operator+(const Element &other) const {
...
+  }
+
+  if (isSupportedFloatType(type)) {
+    APFloat lhsVal = lhs.getFloatValue();
+    APFloat rhsVal = rhs.getFloatValue();
+    const llvm::fltSemantics &elSemantics = lhsVal.getSemantics();
+    APFloat resultVal(lhsVal.convertToDouble() / rhsVal.convertToDouble());
+    bool roundingErr;
+    resultVal.convert(elSemantics, APFloat::rmNearestTiesToEven, &roundingErr);
+    return Element(type, resultVal);
```
burmako, line 259: Can this be simplified to `return Element(type, lhsVal / rhsVal)`?

stablehlo/reference/Element.cpp
```diff
@@ -131,6 +131,47 @@ Element mapWithUpcastToDouble(const Element &el, FloatFn floatFn,
...
+    APFloat resultReal(resultVal.real());
+    resultReal.convert(elSemantics, APFloat::rmNearestTiesToEven, &roundingErr);
+    APFloat resultImag(resultVal.imag());
+    resultImag.convert(elSemantics, APFloat::rmNearestTiesToEven, &roundingErr);
+    return Element(type, std::complex<APFloat>(resultReal, resultImag));
+  }
+
+  report_fatal_error(invalidArgument("Unsupported element type: %s",
+                                     debugString(type).c_str()));
+}
```
burmako, line 173: It doesn't look like this function is used anywhere?

burmako (changes requested):

burmako (approved):

Issue #863, opened 2023-01-02
Implement type inference for RealDynamicSliceOp


burmako:

burmako:
This month, I've been thinking about implementing a shape function for RealDynamicSliceOp, as well as shape functions for its friends - DynamicConvOp (#861), DynamicGatherOp (#690) and DynamicPadOp (#862).

Such a shape function would need to compute the result shape from runtime values (`start_indices`, `limit_indices` and `strides` operands), which is an inherently imprecise process. We can make things more precise by using various static analysis techniques, e.g. partial evaluation, but we don't have to do this immediately. Simply starting with two cases: 1) all values are static => infer fully static shape, 2) otherwise => infer fully dynamic shape - would be a good step forward from the state of the art which doesn't support even that.

However, there's a logistical problem with this. Whenever we add a shape function to a StableHLO op, its autogenerated Python bindings change to no longer include the result type in the signature of the `FooOp.__init__` function. This is not a problem for precise shape functions, but not so much for imprecise shape functions like this one. If the producer knows a more precise type than what the shape function can infer, then `FooOp.__init__` won't work for them, and they will need to use the much less convenient `FooOp.build_generic`. Here's an example from JAX ([click here for full code](https://github.com/google/jax/blob/ef4c2a38d65f4111b54f08f0c2e065174dde9b40/jax/_src/lax/slicing.py#L1403-L1415)):

```Python
  # TODO(burmako): Fix overly conservative type inference of DynamicGatherOp.
  # For now use the build_generic so that we can specify the result type.
  # return hlo.DynamicGatherOp(
  #     operand, indices, mlir.shape_tensor(slice_sizes),
  #     dnums, indices_are_sorted=ir.BoolAttr.get(indices_are_sorted)).results
  results = [mlir.aval_to_ir_type(aval_out)]
  operands = [operand, indices, mlir.shape_tensor(slice_sizes)]
  attributes = {
      "dimension_numbers": dnums,
      "indices_are_sorted": ir.BoolAttr.get(indices_are_sorted)
  }
  return hlo.DynamicGatherOp.build_generic(
      results=results, operands=operands, attributes=attributes).results
```

This problem prompted me to re-evaluate the path forward. Originally, I decided to work on this ticket because of #622. My plan was to add shape functions for as many StableHLO ops as possible and then use these shape functions to specialize the ops from dynamic shapes to static shapes where possible. 

[...]

PR #1012, opened 2023-01-28
Improve type inference for DynamicGatherOp

burmako:
If slice_sizes are constant, then we can infer static dimension sizes for the offset dims parts of the result shape.

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -48,6 +48,7 @@ limitations under the License.
 #include "mlir/IR/BuiltinTypes.h"
 #include "mlir/IR/Diagnostics.h"
 #include "mlir/IR/Location.h"
+#include "mlir/IR/Matchers.h"
 #include "mlir/IR/Operation.h"
 #include "mlir/IR/OperationSupport.h"
 #include "mlir/IR/TypeUtilities.h"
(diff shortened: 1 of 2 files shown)


stablehlo/dialect/TypeInference.cpp
```diff
@@ -2046,7 +2047,13 @@ LogicalResult inferDynamicGatherOp(
                           collapsedSliceDims, startIndexMap, indexVectorDim)))
     return failure();
 
-  auto getSliceDim = [](int64_t index) { return ShapedType::kDynamic; };
+  auto getSliceDim = [&](int64_t index) {
+    DenseIntElementsAttr sliceSizesAttr;
+    if (!matchPattern(sliceSizes, m_Constant(&sliceSizesAttr))) {
+      return ShapedType::kDynamic;
+    }
+    return sliceSizesAttr.getValues<APInt>()[index].getSExtValue();
```
zhouxin913, line 2055: LGTM but with this comment:
However, now we have the power of use `getSExtValue()` for a Value if it's constant. Is there a plan to apply this refinement to many op, or, for this gather op only?

zhouxin913 (approved):

stablehlo/dialect/TypeInference.cpp
burmako: Good question! In principle, this idea is applicable to all ops whose result shapes are computed based on operands: `dynamic_broadcast_in_dim`, `dynamic_conv`, `dynamic_gather`, `dynamic_iota`, `dynamic_pad`, `dynamic_reshape`, `real_dynamic_slice`, `rng`.

However, in practice we may want to postpone implementing this idea until the Python API is figured out, as discussed in https://github.com/openxla/stablehlo/issues/863#issuecomment-1407562475. I'm even thinking to propose removing the shape function for DynamicGatherOp in the meanwhile, although specifically for this op it doesn't matter much logistically, because JAX is already using a workaround.

PR #1018, opened 2023-01-29
Document TorchIndexSelectOp

burmako:
MLIR-HLO commit: tensorflow/mlir-hlo@6353126.


sdasgup3:
Added https://github.com/openxla/stablehlo/issues/1022 to track introducing the spec for the op.

sdasgup3 (approved):

Issue #1000, opened 2023-01-27
Consider refactoring eval functions in Ops.cpp


ghpvnist:
Ops for whose implementations not written in [Element.cpp](https://github.com/openxla/stablehlo/blob/main/stablehlo/reference/Element.cpp) cannot be reused by other ops because the eval functions have to take in the op as an argument. This would pose problems in the future with Batch*ops which require using a combination of other ops.

Recently in DynamicUpdateSliceOp's interpreter #998, I also faced an issue where I'm not able to use ClampOp since I need to construct a tensor for min and max, which cannot be created with ease. Introducing an API to create tensors out of primitive types and/or with shape (defaulting to scalar) would also be useful.

sdasgup3:
One option could be convert the `eval(AddOp, Tensor lhs, ...)` to `eval_add(Type returnType, Tensor ..)`. Anyways the `eval` is currently not overloaded, so we are not adding more declaration by doing that. 

Let us keep the alternative thoughts coming.

sdasgup3:
Following is the list of those 17 ops whose spec has with dependencies written in sub-bullets. 
The dependencies are mostly elementwise ops and others (all_reduce, broadcast_in_dim, clamp, concatenate, constant, dot_general, pad, reduce, reshape, slice, transpose) 


* all_gather
  - concatenate
* all_reduce
  - reduce
* all_to_all
  - slice, concatenate
* batch_norm_grad
  - reduce, constant, broadcast_in_dim, divide, subtract, sqrt, add, multiply
* batch_norm_inference
  - broadcast_in_dim, subtract, sqrt, add, divide, add, multiply
* Batch_norm_training
  - reduce, constant, broadcast_in_dim, divide, subtract, multiply
* convolution
  - pad, dot_general, Slice, Concatenate
* dot_general
  - transpose, slice, reshape, reduce
* dynamic_Slice
  - clamp
* dynamic_update_slice
  - clamp
* log_plus_one
  - log
* logistic
  - divide, add, exponentiation
* maximum
  - or
* minimum
  - and
* reduceScatter
  - all_reduce, slice
* reduce_window
  - pad, slice
* select_andScatter
  - reduce_window, reduce

sdasgup3:
The proposal is to make the changes in two steps:
1. To replace `eval(AddOp, ...)` with `eval_add(Type returnType, ...)`.
[...]

PR #1019, opened 2023-01-29
Update signature of `infer<Op>` function for re-usability

sdasgup3:
Based on the proposal discussed @ https://github.com/openxla/stablehlo/issues/1000, this PR implements the second step: "Sharing the shape inference code from TypeInference.cpp"

Some of the salient  features of the PR:
1. Addresses updating the signature of the type-inference-methods of 4 four ops (`concatenate, dot_general, reduce, slice`) which, according to #1000, could be shared for type-inference and reference implementation.

2. The shared code (e.g. `inferDotGeneralOp, inferConcatenateOp, inferSliceOp, inferReduceOp`)  do have code related to verification of input/output constraints. That means, the reference implementation will execute that verification code
whenever it wants to re-use the eval for those ops. Those verification checks are  redundant, as they must have already been executed while parsing the code, and affect the performance of the interpreter. Despite of that I decided to keep the code as is based on following reasons: 
    - Performance is a non-goal for the "readable" reference. After all it affects only 4 ops. 
    - Some part of type inference is tied closely with the verification code (e.g. usage of [newDimensions at inferReduceOp](https://github.com/openxla/stablehlo/blob/211bd058c88a0c1164b1e77c0dd1e067c6a1bb63/stablehlo/dialect/TypeInference.cpp#L2567)). Isolating just type inference counterparts  involve duplication of code affecting the performance of the verification/inference  which is critical. I note that this observation is true only for `inferReuceOp`. For the remaining ops, the split seems possible.

@burmako @zhouxin913 Please let me know your opinion on this.

### DIFF
--- a/stablehlo/dialect/TypeInference.h
+++ b/stablehlo/dialect/TypeInference.h
@@ -157,7 +157,7 @@ LogicalResult inferComplexOp(std::optional<Location> location, Value lhs,
                              SmallVectorImpl<Type>& inferredReturnTypes);
 
 LogicalResult inferConcatenateOp(std::optional<Location> location,
-                                 ValueRange inputs, int64_t dimension,
+                                 TypeRange inputTypes, int64_t dimension,
                                  SmallVectorImpl<Type>& inferredReturnTypes);
 
 LogicalResult inferConstantOp(std::optional<Location>, ElementsAttr value,
@@ -193,7 +193,7 @@ LogicalResult inferDotOp(
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes);
 
 LogicalResult inferDotGeneralOp(
-    std::optional<Location> location, Value lhs, Value rhs,
+    std::optional<Location> location, Type lhsType, Type rhsType,
     ArrayRef<int64_t> lhsBatchingDimensions,
     ArrayRef<int64_t> rhsBatchingDimensions,
     ArrayRef<int64_t> lhsContractingDimensions,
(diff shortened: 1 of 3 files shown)


zhouxin913 (approved):
This this was exactly I'm worried: the type inference code is tied closed with the many verifications. Yes I agree that the performance is not an big issue here.

stablehlo/dialect/TypeInference.h
```diff
@@ -157,7 +157,7 @@ LogicalResult inferComplexOp(std::optional<Location> location, Value lhs,
                              SmallVectorImpl<Type>& inferredReturnTypes);
 
 LogicalResult inferConcatenateOp(std::optional<Location> location,
-                                 ValueRange inputs, int64_t dimension,
+                                 TypeRange inputTypes, int64_t dimension,
```
burmako, line 160: I like the idea of simplifying signatures in TypeInference.h to use the least possible amount of information. Should we also have a ticket to downgrade Value/ValueRange to Type/TypeRange as much as possible in other functions in this file? cc @zhouxin913

burmako (approved):

stablehlo/dialect/TypeInference.h
zhouxin913: "as much as possible" I looked into this yesterday: I think most of them can be done. There could be very few of cases where we need the Value itself: https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/StablehloOps.cpp#L133
(I'm a little bit surprised this function is not moved)
But it can stay in `hlo_ops.cc` or we look for other ways to work around.

@sdasgup3 could you open a bug for this refactor? and add link of this PR? Thanks!

sdasgup3: Thanks for the insights.

Here is the ticket: https://github.com/openxla/stablehlo/issues/1028

Issue #1002, opened 2023-01-27
Check compliance of already implemented interpreters with `interpreter_checklist.md`


sdasgup3:
The following list of ops where implemented before the existence [interpreter_checklist.md(https://github.com/openxla/stablehlo/blob/main/docs/interpreter_checklist.md) and hence missed the compliance check. The goal of this ticket is to catch up with the pending checks. 

- [ ] add
- [ ] and
- [ ] ceil
- [ ] constant
- [ ] cosine
- [ ] floor
- [ ] iota
- [ ] maximum
- [ ] minimum
- [ ] multiply
- [ ] negate
- [ ] not
- [ ] or
- [ ] reshape
- [ ] sine
- [ ] subtract
- [ ] tanh
- [ ] transpose
- [ ] xor

ghpvnist:
The updated guideline now suggests to have comments related to shape inference in StablehloOps.td and move shape inference tests from `ops_stablehlo.mlir` to `infer_stablehlo.mlir`.

sdasgup3:
From https://github.com/openxla/stablehlo/pull/1371#discussion_r1158894737
make sure that we consistently mark the InferTensorType and similar traits in ODS with constraint labels

sdasgup3:
While exploring the powerOp interpreter implementation, it occurred that the available tests  may be redundant from the POV of code coverage. Just as an example: https://github.com/openxla/stablehlo/blob/65ce9e60ae6c6c8bf9dd04274d06c29f589b675f/stablehlo/tests/interpret_power.mlir#L4 the test vectors (lhs, rhs) = (-1, 1), (-3, 3), (0, 2) all seem to exercise the  loop at https://github.com/openxla/stablehlo/blob/65ce9e60ae6c6c8bf9dd04274d06c29f589b675f/stablehlo/reference/Element.cpp#L720 1, 3 or 2 times, without improving the lcov coverage. 

For now, we can skip any such clean up, and focus on meeting the desired test coverage. In future we might have to revisit the vectors based on code coverage or something else (e.g. the fact that the vectors are exploring some conceptually interesting scenarios).

PR #1026, opened 2023-01-31
Replace `FileCheck` based testing with Check dialect

### LINKED ISSUE #268: Fix lit checks with floating-point ops having implementation defined precision
### Request description

We have floating-point arithmetic ops which might produce results with implementation defined precision. The current lit-based interpreter testing fails if the result is computed with a different precision that the one mentioned in the CHECK directives. As a quickfix, we allow the lit checks to measure the accuracy up to  an arbitrary places after the decimal point. But the solution is far from ideal.

Instead we explore to have  a custom FileCheck utility to address the floating-point comparison.

sdasgup3:
Fixes #268 

The PR does the following:
1. Introduces a dialect `CHECK` with opset (e.g. `stablehlo::check::expect_eq`, `stablehlo::check::expect_eq_const`, `stablehlo::check::expect_almost_eq`, and `stablehlo::check::expect_almost_eq_const`) to be used for implementing test harness for  StableHLO interpreter. The opset is inspired from [iree::CheckOps.td](https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Modules/Check/IR/CheckOps.td).
2. Implement the operations of the dialect. This will enable setting test assertions on the results of the interpreter. 
3. Modify all the `interpret_*.mlir` files to accommodate assertion based testing. Also, removes the dependency on [FileCheck](https://llvm.org/docs/CommandGuide/FileCheck.html).

Note:  
1. The dialect has a dependency on [StablehloBase](https://github.com/openxla/stablehlo/blob/7be911a1a8468248efdc16e2a51ed5ff30c1cfbb/stablehlo/dialect/CMakeLists.txt#L22) because it is sharing the types like `HLO_Tensor`. Even though it is possible to avoid any dependency (by using generic types like `AnyTensor`), but the decision is deliberate in order to brand dialect as a test facilitator for stableHLO only.

2. The core check for checking two floating points are equal modulo some tolerance is inspired from https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/ and https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon  

3. The dialect get registered only when running the interpreter: not registered to be used with `stablehlo-opt` tool.

@GleasonK As the proposed opset is not an extension of StableHLO dialect and meant to be used only for testing, I do not think we need to make any compatibility guarantees. Let me know your opinion. 

Added the comment https://github.com/openxla/stablehlo/pull/1026/files#r1115404431 to explain the rationale behind the implementation of  `areAproximatelyEqual(ApFloat, APFloat)`.


stablehlo/dialect/CheckOps.td
```diff
@@ -0,0 +1,132 @@
...
+#ifndef STABLEHLO_DIALECT_CHECK_OPS
+#define STABLEHLO_DIALECT_CHECK_OPS
+
+include "stablehlo/dialect/Base.td"
+
+//===----------------------------------------------------------------------===//
+// CHECK dialect.
+//===----------------------------------------------------------------------===//
+
+def CHECK_Dialect : Dialect {
```
burmako, line 26: Given that this dialect is specific to tests, I think it should live under `tests/`.

stablehlo/dialect/CheckOps.td
```diff
@@ -0,0 +1,132 @@
...
+  let useDefaultAttributePrinterParser = 0;
+  let useDefaultTypePrinterParser = 0;
+  let useFoldAPI = kEmitFoldAdaptorFolder;
+}
+
+//===----------------------------------------------------------------------===//
+// CHECK op definitions.
+//===----------------------------------------------------------------------===//
+
+def CHECK_ExpectEqOp : Op<CHECK_Dialect, "expect_eq", [SameTypeOperands]> {
```
burmako, line 43: Is this op used anywhere? So far, it looks like only `check.expect_almost_eq` is used?

stablehlo/dialect/CheckOps.td
```diff
@@ -0,0 +1,132 @@
...
+    Op<CHECK_Dialect, "expect_eq_const", [AllTypesMatch<["lhs", "value"]>]> {
+  let summary = [{Checks that the tensor operand is equal to some constant}];
+  let description =  [{
+    Verifies that the tensor operand is exactly equal to a constant attribute.
+    For floating-point and complex types, uses bitwise comparison for equality.
+
+    Issues a non-fatal failure if the verification fails.
+
+    ```mlir
+    check.expect_eq_const(%arg0, dense<[1, 2]> : tensor<2xi32>) : tensor<2xi32>
```
burmako, line 74: The syntax looks kind of weird at the moment, with the same type duplicated twice. Could this be something like `check.expect_eq %arg0, dense<[1, 2]> : tensor<2xi32>`? (Without `_const` to make things more brief).

stablehlo/reference/Ops.h
```diff
@@ -44,6 +45,12 @@ Tensor evalTransposeOp(Type resultType, const Tensor &operand,
                        DenseIntElementsAttr permutation);
 Tensor evalXorOp(const Tensor &lhs, const Tensor &rhs, Type resultType);
 
+// The eval functions for the following ops are used only for test harness.
+llvm::Error evalExpectAlmostEqConstOp(const Tensor &lhs, ElementsAttr value);
+llvm::Error evalExpectAlmostEqOp(const Tensor &lhs, const Tensor &rhs);
+llvm::Error evalExpectEqConstOp(const Tensor &lhs, ElementsAttr value);
+llvm::Error evalExpectEqOp(const Tensor &lhs, const Tensor &rhs);
```
burmako, line 52: Can we move these functions into a separate file in `tests/`?

stablehlo/tests/interpret_tanh.mlir
```diff
@@ -1,107 +1,53 @@
...
-  // CHECK-NEXT: 7.617180e-01 : bf16
-  // CHECK-NEXT: 1.245120e-01 : bf16
-  // CHECK-NEXT: 9.960930e-02 : bf16
-  // CHECK-NEXT: 9.960930e-01 : bf16
-  // CHECK-NEXT: 1.000000e+00 : bf16
-  // CHECK-NEXT: -1.000000e+00 : bf16
-  // CHECK-NEXT: 0x7FFF : bf16
-  // CHECK-NEXT: 9.183550e-41 : bf16
-  // CHECK-NEXT: -9.183550e-41 : bf16
+  check.expect_almost_eq_const(%1, dense<[0.000000e+00, -0.000000e+00, 7.617180e-01, 1.245120e-01, 9.960930e-02, 9.960930e-01, 1.000000e+00, -1.000000e+00, 0x7FFF, 9.183550e-41, -9.183550e-41]> : tensor<11xbf16>) : tensor<11xbf16>
```
burmako, line 6: These lines are getting very long and not super easy to read.

stablehlo/dialect/CheckOps.td
```diff
@@ -0,0 +1,132 @@
...
+
+  let arguments = (ins
+      HLO_Tensor:$lhs,
+      HLO_Tensor:$rhs
+  );
+
+  let assemblyFormat = "`(` $lhs `,` $rhs `)` attr-dict `:` type($lhs)";
+}
+
+def CHECK_ExpectEqConstOp :
```
burmako, line 64: Can we shorten the name to make this easier to read? E.g. `check.eq` or `expect.eq`?

burmako (changes requested):
Only reviewed the API, will wait for an approving review before reviewing everything else.

stablehlo/dialect/CheckOps.td
sdasgup3: Yes, the op is not used anywhere at the moment. Minimized the opset with what is currently in use.

stablehlo/dialect/CheckOps.td
sdasgup3: done.

stablehlo/dialect/CheckOps.td
sdasgup3: done.

stablehlo/dialect/CheckOps.td
sdasgup3: done

stablehlo/reference/Ops.h
sdasgup3: done.

stablehlo/tests/interpret_tanh.mlir
sdasgup3: Some of these checks will be removed as part of [Testing Guideline G1](https://github.com/openxla/stablehlo/blob/main/docs/reference.md#testing-guidelines )
driven by https://github.com/openxla/stablehlo/issues/1002

For now we can keep one constant per line.

stablehlo/dialect/CheckOps.td
GleasonK: Do you think `tests/` even though interpreter needs to support these ops? Currently only testing utils and tools depend on any files in our `tests/` folder. Something feels a bit off about `reference/` depending on something from  `test/`

stablehlo/tests/CheckOps.td
```diff
@@ -0,0 +1,91 @@
...
+    check.eq %arg0, dense<[1, 2]> : tensor<2xi32>
+    ```
+  }];
+
+  let arguments = (ins
+    HLO_Tensor:$lhs,
+    ElementsAttr:$value
+  );
+
+  // let assemblyFormat = " $lhs `,` $value attr-dict `:` type($lhs)";
```
GleasonK, line 62: clean up

stablehlo/tests/CheckOps.td
```diff
@@ -0,0 +1,91 @@
...
+    check.expect_almost_eq_const(%arg0, dense<[0.999999, 2.0]> : tensor<5xf32>) : tensor<5xf32>
+    ```
+  }];
+
+  let arguments = (ins
+    TensorOf<[AnyFloat, AnyComplex]>:$lhs,
+    ElementsAttr:$value
+  );
+
+  // let assemblyFormat = " $lhs `,` $value attr-dict `:` type($lhs)";
```
GleasonK, line 87: clean up

stablehlo/tests/CheckOps.td
```diff
@@ -0,0 +1,91 @@
...
+  let summary = [{Checks that the tensor operand is almost equal to some constant}];
+  let description =  [{
+    Verifies that the tensor operand with floating-point or complex element
+    type is almost equal to the constant attribute within an
+    implementation-defined tolerance.
+
+    Issues a non-fatal failure if the verification fails.
+
+    ```mlir
+    check.expect_almost_eq_const(%arg0, dense<[0.999999, 2.0]> : tensor<5xf32>) : tensor<5xf32>
```
GleasonK, line 78: update with new syntax

stablehlo/tests/CheckOps.td
```diff
@@ -0,0 +1,91 @@
...
+  );
+
+  // let assemblyFormat = " $lhs `,` $value attr-dict `:` type($lhs)";
+  let assemblyFormat = " $lhs `,` $value attr-dict";
+}
+
+def CHECK_AlmostEqOp :
+    Op<CHECK_Dialect,
+       "almost_eq", [AllTypesMatch<["lhs", "value"]>]> {
+  let summary = [{Checks that the tensor operand is almost equal to some constant}];
```
GleasonK, line 69: If single line use `"..."`, else put this on its own line, probably own line since longer than 80 chars, but fine with either:

```
let summary = [{
  Checks that the tensor operand is almost equal to some constant
}];
```

stablehlo/tests/CheckOps.td
GleasonK: Also op is now called "almost_eq"

stablehlo/tests/CheckOps.td
```diff
@@ -0,0 +1,91 @@
...
+    ```
+  }];
+
+  let arguments = (ins
+    TensorOf<[AnyFloat, AnyComplex]>:$lhs,
+    ElementsAttr:$value
+  );
+
+  // let assemblyFormat = " $lhs `,` $value attr-dict `:` type($lhs)";
+  let assemblyFormat = " $lhs `,` $value attr-dict";
```
GleasonK, line 88: Remove leading space?

stablehlo/tests/CheckOps.td
```diff
@@ -0,0 +1,91 @@
...
+    ```
+  }];
+
+  let arguments = (ins
+    HLO_Tensor:$lhs,
+    ElementsAttr:$value
+  );
+
+  // let assemblyFormat = " $lhs `,` $value attr-dict `:` type($lhs)";
+  let assemblyFormat = " $lhs `,` $value attr-dict";
```
GleasonK, line 63: Remove leading space?

stablehlo/dialect/CheckOps.td
burmako: "Something feels a bit off about reference/ depending on something from tests/". That's a good point. Continuing that logic, would you agree that it's odd for `reference/` to include stuff related to testing, e.g. for Ops.h to include `evalEqOp`?

stablehlo/tests/CheckOps.td
```diff
@@ -0,0 +1,91 @@
...
+//===----------------------------------------------------------------------===//
+
+def CHECK_EqOp :
+    Op<CHECK_Dialect, "eq", [AllTypesMatch<["lhs", "value"]>]> {
+  let summary = [{Checks that the tensor operand is equal to some constant}];
+  let description =  [{
+    Verifies that the tensor operand is exactly equal to a constant attribute.
+    For floating-point and complex types, uses bitwise comparison for equality.
+
+    Issues a non-fatal failure if the verification fails.
```
GleasonK, line 50: Can we get some negative test cases for both the new ops? With only positive test cases I don't think we test that the "approximately equal" works, and to what tolerance - in the past I had implemented a float comparison function that unintentionally always returned true, which went uncaught for awhile because of inadequate negative testing.

GleasonK (changes requested):
For compatibility - Check doesn't need compatibility guarantees in the VHLO sense, but StableHLO backward compatibility is based on the reference interpreter results, so I would think that once we settle in on these checks, we probably shouldn't be changing the implementation details too much.

Left a bunch of nitpicks. This is a super cool PR, excited to have this feature!

stablehlo/tests/CheckOps.td
GleasonK: Also I only skimmed the tests and searched for what the error would be, if there are negative tests that I missed let me know and feel free to resolve the comment

stablehlo/tests/CheckOps.td
sdasgup3: rephrased and fitted into one line

stablehlo/dialect/CheckOps.td
GleasonK: From a user perspective - do we ever intend another project to take a dependency on a reference library? If so would we expect them to want check ops? If yes, I would think that verification (especially the machinery which determines an acceptable threshold for FP equality) is a feature of a reference interpreter.

If not, then yeah I agree this is test machinery specific to StableHLO testing and a perfect design probably would allow plugging custom dialects into the reference interpreter, like these test ops. That feels like overkill, so the current design is probably good enough for now. I think that's what IREE does? It has some VM Native implementation declaration.

burmako: "a perfect design probably would allow plugging custom dialects into the reference interpreter, like these test ops". A low-tech alternative to decouple the check dialect, if we decide this is desirable, would involve splitting `--interpret` into `--interpret` (reference-specific) and `--check` (tests-specific).

sdasgup3: Hi All, Thanks for all the comments. 

>> do we ever intend another project to take a dependency on a reference library?

Thanks for the feedback! 
I do not think the consumers of StableHLO op `eval`uators should be interested in `Check` specific ops. Using  `checkDialect` is one way for testing the evaluators and it is not unlikely for a consumer to have different testing strategies which may not reply on those. One such need could be to cross validate their implementation of StableHLO ops  against the interpreter.


Having said that this is my proposal:
## Proposal 
1. Move the definition of `Check` ops in test/.
2. Move the reference implementation of `Check` ops in test/
3. Move `Interpreter.cpp` in tools/.

With the above, the binary for `stablehlo-interpreter` will evolved as follows:
B1. dialect/ and refernece/ respectively provide the libraries related to  StableHLO ops and  the implementation of those ops.
B2. test/ provides the the libraries related to  both CheckOps ops and  the implementation of those ops.
B3. (1) and (2) contribute to create a library w.r.t tools/Interpreter.cpp (which is the interpreter driver for different dialects; more about this later)
B4. Finally (3) is linked to create the executable at tool/.

### Rationale for 1 and 2
Both the dialect and its implementation are need for testing purposes only.  

### rationale for 3
The  `Interpreter.cpp` provides the evaluator for `func::func` op. Specifically, it provides a an interpreter drive loop to evaluate op from different dialects (like `func::retrunOp, Stablehlo::*, check::*`). IMO, it should not be part of reference/ which host the evaluator of StableHLO ops only.

To me "..nterpreter drive loop to evaluate op from different dialects.." seems very close to tooling. making  
`Interpreter.cpp` part of test/ has the problem that consumer has to depend on test/ unless they are willing to write their own driver.

### Outcome
reference/ has now **no** dependency on tests.  This enables writing an interpreter driver depending solely on reference/.

## About the added dependency from from test/ to tool/  (because of B4)

Given that we already have similar  dependency w.r.t `StablehloTestUtils`, at least we are no introducing a new dependency. However, we can provide additional machinery: For example two interpreter tools like `check-stablehlo-interpreter` (depending on check/,  dialect/ and reference), and  `stablehlo-interpreter` (depending on only dialect/ and reference). For our use case only the former makes sense.



>>  A low-tech alternative to decouple the check dialect, if we decide this is desirable, would involve splitting --interpret into --interpret (reference-specific) and --check (tests-specific).
This sound great! Given the above proposal, do we need to explore this? @burmako  

Let me know your feedback.

stablehlo/tests/interpret_tanh.mlir
```diff
@@ -1,107 +1,55 @@
...
-  // CHECK-NEXT: 9.960930e-02 : bf16
-  // CHECK-NEXT: 9.960930e-01 : bf16
-  // CHECK-NEXT: 1.000000e+00 : bf16
-  // CHECK-NEXT: -1.000000e+00 : bf16
-  // CHECK-NEXT: 0x7FFF : bf16
-  // CHECK-NEXT: 9.183550e-41 : bf16
-  // CHECK-NEXT: -9.183550e-41 : bf16
+
+  check.almost_eq %1, dense<[0.000000e+00, -0.000000e+00, 7.617180e-01, 1.245120e-01, 9.960930e-02, 9.960930e-01, 1.000000e+00, -1.000000e+00, 0x7FFF, 9.183550e-41, -9.183550e-41]> : tensor<11xbf16>
+
```
burmako, line 8: Extraneous newlines on lines 6 and 8?

stablehlo/tests/interpret_tanh.mlir
burmako: We don't have to solve this right now, but let's think how to make our tests more readable in the long term.

stablehlo/dialect/CheckOps.td
burmako: > IMO, it should not be part of reference/ which host the evaluator of StableHLO ops only 

I think that the reference implementation should cover all ops that constitute StableHLO programs, because that's the fundamental use case for the reference implementation - take a StableHLO program and run it.

Today, some of these ops come from other dialects (e.g. builtin/func for modularity, arith/shape/tensor for dynamism). In the future, we are planning to fork most of these ops into StableHLO. But regardless of where these ops live, I think that we want to handle them in the interpreter (i.e. under `reference/`).

> Move Interpreter.cpp in tools/ (or tests/ with trade offs discussed later 

If we abstract away from naming a little bit (whether the file is called Interpreter.h or something else) and view this from the perspective of the functionality that we want `reference/` to provide, I think this functionality would look roughly as follows:

```C++
EvaluationResult evaluate(StablehloProgram program, ArrayRef<Argument> args);
```

"making Interpreter.cpp part of test/ has the problem that consumer has to depend on test/ unless they are willing to write their own driver". I would reframe this as "has the problem that consumer will need to implement the functionality that they need on their own, using the individual `evalFoo` functions provided by `reference/`". When viewed from this angle, I think this would be a considerable downside - it would be quite undesirable for us to leave parts of the implementation of our core functionality to users.

Given that, I would recommend to find a way to keep Interpreter.h under `reference/`, e.g. by changing `eval` to take an optional parameter callback that looks along the lines of `llvm::function_ref<void(Operation*, Scope*)>`. Then regular users won't even notice this, and the `stablehlo-interpreter` tool would be able to pass a callback that handles `check` ops.

burmako (changes requested):

stablehlo/dialect/CheckOps.td
sdasgup3: Thanks for the suggestion. It looks great to me. Implemented, PTAL.

stablehlo/tests/interpret_tanh.mlir
sdasgup3: I agree. We anyway need to trim most of the test cases to abide by the testing guidelines. We can work of readability of tests  as part of that.

stablehlo/tests/CheckOps.td
sdasgup3: The following example demonstrates the tolerance in action. I have updated the reference.md with this information. Please have a look.

```mlir
func.func @check_tolerance() {
  %0 = stablehlo.constant dense<0.20000000000000001110> : tensor<f32>
  check.almost_eq %0, dense<0.199999988> : tensor<f32> // aseert true
  check.eq %0, dense<0.199999988> : tensor<f32> // asset false
  func.return
}
```

>> negative test

That is a good point. 
In the current implementation we giving  fatal error when the mismatch happens. With that I am not sure if I am unsure if I can add a negative test in the test suite.

burmako (changes requested):

stablehlo/reference/Element.cpp
```diff
@@ -131,6 +131,36 @@ Element mapWithUpcastToDouble(const Element &el, FloatFn floatFn,
                                      debugString(type).c_str()));
 }
 
+template <class T>
+typename std::enable_if<!std::numeric_limits<T>::is_integer, bool>::type
+almost_equal(T x, T y, int ulp) {
```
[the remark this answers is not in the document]
sdasgup3: Sorry for the confusion. No there is no differences. Renamed all of them to `areApproximatelyEqual`.

docs/reference.md
```diff
@@ -201,13 +198,39 @@ generalized. Tests exercising undefined behavior do not contribute towards
...
+  std::fabs(a - r) < std::numeric_limits<T>::min();
+```
+
+The following example demonstrates the above tolerance in action.
+
+```mlir
+func.func @check_tolerance() {
+  %0 = stablehlo.constant dense<0.20000000000000001110> : tensor<f32>
+  check.almost_eq %0, dense<0.199999988> : tensor<f32> // assert true
+  check.eq %0, dense<0.199999988> : tensor<f32> // assert false
```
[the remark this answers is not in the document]
sdasgup3: reworded. PTAL

stablehlo/reference/Element.h
```diff
@@ -121,6 +128,10 @@ Element sine(const Element &e);
 /// Returns element-wise tanh of Element object.
 Element tanh(const Element &e);
 
+/// Checks if the underlying float or complex value stored in Element objects,
+/// e1 and e2, are almost equal within an implementation-defined tolerance.
```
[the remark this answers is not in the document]
sdasgup3: Clarified the comment.

stablehlo/tests/CheckOps.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+llvm::Error evalEqOp(const Tensor &lhs, ElementsAttr value) {
+  auto rhs = makeTensor(value.cast<DenseElementsAttr>());
+  for (auto lhsIt = lhs.index_begin(), rhsIt = rhs.index_begin();
+       lhsIt != lhs.index_end(); ++lhsIt, ++rhsIt) {
+    if (lhs.get(*lhsIt) != rhs.get(*rhsIt)) {
+      return invalidArgument(
+          "Element value don't match: %s (actual) vs %s (expected)\n",
+          debugString(lhs.get(*lhsIt)).c_str(),
+          debugString(rhs.get(*rhsIt)).c_str());
```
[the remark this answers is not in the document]
sdasgup3: >> How is this going to be reported to the user?

The errors you pointed out will be propagated from the callback function to the caller `eval(Region..` (in Ops.cpp), where it is handled as a fatal error which prints the following message followed by a crash report.
```                                                                                                                                   
LLVM ERROR: Error evaluating function: add_op_test_si4.                                                                                                                                                                                
        Assersion eq failed: Element value don't match: 7 : i4 (actual) vs 5 : i4 (expected) 
```

As mentioned earlier I am not sure how to add negative tests given the fatal error behavior, which IMO, is there to maintain uniform behavior across all the op-evaluations  in Ops.cpp has the same behavior.

stablehlo/reference/Ops.h
```diff
@@ -73,10 +73,13 @@ Tensor evalXorOp(const Tensor &lhs, const Tensor &rhs, Type resultType);
 /// Evaluates an mlir::Region `region` using the runtime values `args`
 /// corresponding to the arguments of the entry block of the region.
 /// Interprets the operations within the entry block and returns the runtime
-/// values for the terminator's arguments.
-/// Assumes that the region has only one block.
-llvm::SmallVector<Tensor> eval(Region &region, llvm::ArrayRef<Tensor> args,
-                               Scope *parentScope = nullptr);
+/// values for the terminator's arguments. The optional callback `fallback` is
+/// used for evaluating ops, other than `func::returnOp` and the ones in
+/// stablehlo dialect, within the `region`. Assumes that `region` has only one
```
burmako, line 78: "for evaluating ops, other than `func::returnOp` and the ones in stablehlo dialect, within the `region`". How about just "for evaluating ops which are not supported by the interpreter"? Rationale:
  1) There's also `func::CallOp` that we'll need to support eventually.
  2) These ops can be outside the region, e.g. if we call into another function.

stablehlo/tests/CheckOps.td
```diff
@@ -0,0 +1,84 @@
...
+//===----------------------------------------------------------------------===//
+// CHECK dialect.
+//===----------------------------------------------------------------------===//
+
+def CHECK_Dialect : Dialect {
+  let name = "check";
+  let cppNamespace = "::mlir::stablehlo::check";
+
+  let description = [{
+    Opset to implement test assertions for StableHLO interpreter.
```
burmako, line 30: "to implement checks for testing the StableHLO interpreter"? Since we're moving away from the terminology involving assertions.

stablehlo/tests/CheckOps.cpp
```diff
@@ -0,0 +1,73 @@
...
+      >();
+}
+
+llvm::Error evalEqOp(const Tensor &lhs, ElementsAttr value) {
+  auto rhs = makeTensor(value.cast<DenseElementsAttr>());
+  for (auto lhsIt = lhs.index_begin(), rhsIt = rhs.index_begin();
+       lhsIt != lhs.index_end(); ++lhsIt, ++rhsIt) {
+    if (lhs.get(*lhsIt) != rhs.get(*rhsIt)) {
+      return invalidArgument(
+          "Element value don't match: %s (actual) vs %s (expected)\n",
```
burmako, line 47: Typo: "Element values".

stablehlo/tests/CheckOps.cpp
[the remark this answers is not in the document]
burmako: Let's open a ticket for improving error reporting. That would involve: 1) somehow adding negative tests, 2) improving the error message which is currently going to pretty hard to debug (it just prints mismatching values, without printing the whole tensors or the indices of the mismatching position). No need to block this PR on addressing this right away for the sake of moving faster.

docs/reference.md
```diff
@@ -99,39 +99,36 @@ folding in MHLO, at which point we'll improve ergonomics of the code snippet
...
 be fed to the program, and generates output data values, which are matched
-against the user-provided expected data values.
-
-In the current implementation, we package the inputs (MLIR program + input data
-values) and outputs in a
-[lit-based](https://llvm.org/docs/CommandGuide/lit.html) test as follows:
+against the user-provided expected data values. The data values (B) are
+hard-coded in the program itself using `stablehlo.constant` operations. The
+interpreter evaluates the input program. The output(s) of the op under test
+is checked via assertions (e.g. `check.eq`, `check.almost_eq`), as shown below.
```
burmako, line 109: Another use of assertions which would be good to migrate away. Replace with "checks"? (Same below in the document).

docs/reference.md
```diff
@@ -99,39 +99,36 @@ folding in MHLO, at which point we'll improve ergonomics of the code snippet
...
+`check.eq` checks for bitwise equality for any supported type and
+`check.almost_eq` checks for nearly equality within an implementation-defined
```
burmako, line 111: "within a tolerance described below".

docs/reference.md
```diff
@@ -201,13 +198,49 @@ generalized. Tests exercising undefined behavior do not contribute towards
...
+```
 
-The current lit-based interpreter testing fails if the result is computed with a
-different precision than what is mentioned in the lit CHECK directives. As a
-quick-fix, we allow the lit checks to measure the accuracy up to an arbitrary
-places after the decimal point. But the solution is far from ideal. We plan to
-resolve it using [ticket](https://github.com/openxla/stablehlo/issues/268).
+This is just the first step in testing numerical accuracy of StableHLO ops. At
+the moment, this is an underspecced area of the StableHLO spec, and there is
+ongoing work to figure it out (#1156) based on our experience with using
```
burmako, line 241: Turn this into a link?

stablehlo/tools/StablehloInterpreterMain.cpp
```diff
@@ -15,22 +15,53 @@ limitations under the License.
...
-           << "\n";
+        auto evalCheckOps = [&](Operation &op,
+                                stablehlo::Scope &scope) -> llvm::Error {
+          if (auto almostEqOp = dyn_cast<stablehlo::check::AlmostEqOp>(op)) {
+            stablehlo::Tensor runtimeOperand = scope.find(almostEqOp.getLhs());
+            auto status = stablehlo::check::evalAlmostEqOp(
+                runtimeOperand, almostEqOp.getValue());
+            if (status)
+              return stablehlo::invalidArgument(
+                  "Error evaluating function: %s. \n\tAssersion almost_eq "
```
burmako, line 42: "Check"? Here and below.

burmako (changes requested):
I didn't have the time to review everything in detail, but I reviewed the architecture/APIs and the comments. While there are open questions (e.g. error reporting), I would propose that we merge and iterate in follow-up PRs (after the current feedback is addressed). This is a significant improvement to readability and reproducibility of the tests.

stablehlo/tests/CheckOps.cpp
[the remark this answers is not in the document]
sdasgup3: Sure. Here is the ticket https://github.com/openxla/stablehlo/issues/1196

burmako (approved):
LGTM, but I didn't review the implementation details, so this will need another review.

stablehlo/tests/CheckOps.h
```diff
@@ -0,0 +1,50 @@
...
+namespace check {
+
+class CheckDialect : public Dialect {
+ public:
+  explicit CheckDialect(MLIRContext *context);
+  static StringRef getDialectNamespace() { return "check"; }
+};
+
+// The eval functions for the following ops are used only for test harness.
+llvm::Error evalEqOp(const Tensor &lhs, ElementsAttr value);
```
ghpvnist, line 40: Nit: can you alphabetically order these definitions, implementations, and the tablegen file?

stablehlo/tests/CheckOps.td
```diff
@@ -0,0 +1,84 @@
...
+    Verifies that the tensor operand is exactly equal to a constant attribute.
+    For floating-point and complex types, uses bitwise comparison for equality.
+
+    ```mlir
+    check.eq %arg0, dense<[1, 2]> : tensor<2xi32>
+    ```
+  }];
+
+  let arguments = (ins
+    HLO_Tensor:$lhs,
```
ghpvnist, line 55: Would we have a use case for using this particular op other than `HLO_INT` and `HLO_PRED`? In other words, should we restrict floats and complex element types from using `EqOp`?

stablehlo/tests/interpret_abs.mlir
```diff
@@ -1,36 +1,26 @@
...
+func.func @abs_op_test_si64() {
   %operand = stablehlo.constant dense<[-2, 0, 2]> : tensor<3xi64>
   %result = stablehlo.abs %operand : tensor<3xi64>
-  func.return %result : tensor<3xi64>
-  // CHECK-NEXT: tensor<3xi64>
-  // CHECK-NEXT: 2 : i64
-  // CHECK-NEXT: 0 : i64
-  // CHECK-NEXT: 2 : i64
+  check.eq %result, dense<[2, 0, 2]> : tensor<3xi64>
+  func.return
```
ghpvnist, line 7: Just curious, is `func.return` required?

stablehlo/tests/interpret_ceil.mlir
```diff
@@ -1,75 +1,35 @@
-// RUN: stablehlo-interpreter --interpret -split-input-file %s | FileCheck %s
+// RUN: stablehlo-interpreter --interpret -split-input-file %s
 
-// CHECK-LABEL: Evaluated results of function: ceil_op_test_bf16
-func.func @ceil_op_test_bf16() -> tensor<9xbf16> {
-  %0 = stablehlo.constant dense<[0xFF80, -2.5, 0x8001, -0.0, 0.0, 0x0001, 2.5, 0x7F80, 0x7FC0]>  : tensor<9xbf16>
+func.func @ceil_op_test_bf16() {
+  %0 = stablehlo.constant dense<[0xFF80, -2.5, 0x8001, -0.0, 0.0, 0x0001, 2.5, 0x7F80, 0x7FC0]> : tensor<9xbf16>
```
ghpvnist, line 4: Can you provide a comment for what the hex values represent? It will make reading the tests much easier.
```
// -Inf, -2.5, -Denormal, -0.0, 0.0, +Denormal, 2.5, +Inf, +NaN
```
As for `0x7FC0`, can we use `0x7F81` instead? Both are +NaNs, but `0x7F81` is the smallest representable +NaN while `0x7Fc0` falls somewhere in the middle, and the rest of the special values all follow the same convention. I haven't gone through other test values, but they all seem to default to `0x7F80...` depending on the bit length, which is an incorrect assumption for smallest +NaNs depending on the number of exponent bits.

docs/reference.md
```diff
@@ -201,13 +198,50 @@ generalized. Tests exercising undefined behavior do not contribute towards
...
 
 **(G6) While writing tests for floating-point type, to what precision the
-results need to be specified in llvm lit checks?**
+expected result need to be specified in checks?**
+
+For elementary operations (addition, subtraction, multiplication, division, and
+square), an implementation following IEEE specification is expected to provide a
+rounded result within 0.5 ULP of the mathematically exact result. That said, we
+can safely imagine the expected result coming out of these operations to be
+atmost 1 ULP apart. However, this may not work for transcendental functions
```
ghpvnist, line 207: Typo: `at most`

docs/reference.md
```diff
@@ -201,13 +198,50 @@ generalized. Tests exercising undefined behavior do not contribute towards
...
+(`sine`, `cosine`) for which the precision guarantees are implementation-defined
```
ghpvnist, line 208: Let's add the `etc` part to show that ```(`sine`, `cosine`)``` is not the exhaustive list.
```
(`sine`, `cosine`, etc.)
```

docs/reference.md
```diff
@@ -201,13 +198,50 @@ generalized. Tests exercising undefined behavior do not contribute towards
...
+[rationale](https://github.com/openxla/stablehlo/issues/96).
```
ghpvnist, line 209: Nit: ([rationale](https://github.com/openxla/stablehlo/issues/96)).

stablehlo/reference/Element.cpp
```diff
@@ -296,6 +366,28 @@ Element abs(const Element &el) {
...
 }
 
+bool areApproximatelyEqual(const Element &e1, const Element &e2) {
+  Type type = e1.getType();
+  if (type != e2.getType())
+    report_fatal_error(invalidArgument("Element types don't match: %s vs %s",
+                                       debugString(type).c_str(),
+                                       debugString(e2.getType()).c_str()));
+
+  if (isSupportedFloatType(type)) {
```
ghpvnist, line 376:
```
  if (isSupportedFloatType(type))
    return areApproximatelyEqual(e1.getFloatValue(), e2.getFloatValue());
```

ghpvnist (changes requested):

stablehlo/tests/interpret_ceil.mlir
sdasgup3: As part of https://github.com/openxla/stablehlo/issues/1002, the plan to to revisit the test files interpret_foo.mlir and make them complaint with the checklist. I believe that revision will trim most of these test inputs to something simpler  (most probably to the example test we have in the specification).  

Having said that we should wait till then.

stablehlo/tests/interpret_abs.mlir
sdasgup3: I do not think we can have a block without terminator. 
Besides that, from POV of evaluation of a StableHLO program, `func.return` plays an important role in breaking the  interpreter loop [here](https://github.com/openxla/stablehlo/blob/7e9ce9af667a65e2755c1fffec2ddd201f92d98d/stablehlo/reference/Ops.cpp#L537)

stablehlo/tests/CheckOps.td
sdasgup3: >> Would we have a use case for using this particular op other than HLO_INT and HLO_PRED

We do not have a usecase for now. The support of  float and complex types  are provided for completeness of `check.eq` which checks for bitwise equality of any type.

ghpvnist (approved):

stablehlo/reference/Element.cpp
```diff
@@ -162,6 +192,46 @@ std::complex<APFloat> Element::getComplexValue() const {
...
 
+bool Element::operator==(const Element &other) const {
+  Type type = other.getType();
+  if (type_ != type)
+    report_fatal_error(invalidArgument("Element types don't match: %s vs %s",
+                                       debugString(type_).c_str(),
+                                       debugString(type).c_str()));
+
+  if (isSupportedIntegerType(type)) {
+    auto intLhs = (*this).getIntegerValue();
```
GleasonK, line 203: Any reason to not use `this->` ?

GleasonK: Same below in this file

GleasonK: On second thought, is `this` even needed? Can we just call:

```
auto intLhs = getIntegerValue();
```

docs/reference.md
sdasgup3: I missed pushing these changes. Just pushed now.

docs/reference.md
```diff
@@ -201,13 +198,50 @@ generalized. Tests exercising undefined behavior do not contribute towards
 the understanding of the op's behavior.
 
 **(G6) While writing tests for floating-point type, to what precision the
-results need to be specified in llvm lit checks?**
+expected result need to be specified in checks?**
+
+For elementary operations (addition, subtraction, multiplication, division, and
+square), an implementation following IEEE specification is expected to provide a
+rounded result within 0.5 ULP of the mathematically exact result. That said, we
+can safely imagine the expected result coming out of these operations to be
```
ghpvnist, line 206: Nit with 80 line character:
```
can safely imagine the expected result coming out of these operations to be at
most 1 ULP apart. However, this may not work for transcendental functions
```

stablehlo/tests/CheckOps.cpp
```diff
@@ -0,0 +1,73 @@
...
+#define GET_OP_LIST
+#include "stablehlo/tests/CheckOps.cpp.inc"
+      >();
+}
+
+llvm::Error evalAlmostEqOp(const Tensor &lhs, ElementsAttr value) {
+  auto rhs = makeTensor(value.cast<DenseElementsAttr>());
+  for (auto lhsIt = lhs.index_begin(), rhsIt = rhs.index_begin();
+       lhsIt != lhs.index_end(); ++lhsIt, ++rhsIt) {
+    if (!areApproximatelyEqual(lhs.get(*lhsIt), rhs.get(*rhsIt))) {
```
ghpvnist, line 45: Remove inner braces here and below.

ghpvnist (approved):
LGTM with minor comments

stablehlo/reference/Element.cpp
```diff
@@ -131,6 +131,35 @@ Element mapWithUpcastToDouble(const Element &el, FloatFn floatFn,
...
+}
+
+// Checks if two APFloat values, f and g, are almost equal.
+bool areApproximatelyEqual(APFloat f, APFloat g) {
+  if (&f.getSemantics() != &g.getSemantics()) return false;
+
+  llvm::APFloatBase::cmpResult cmpResult = f.compare(g);
+  if (cmpResult == APFloat::cmpEqual) return true;
+  if (cmpResult == APFloat::cmpUnordered) return f.isNaN() == g.isNaN();
+  if (!f.isFiniteNonZero() || !g.isFiniteNonZero()) return false;
```
sdasgup3, line 151: Let me explain the rationale behind the way the function is implemented. The implementation is inspired from [APFLoat::compare](https://github.com/llvm/llvm-project/blob/1c2280264058c26aee2bcd8fe4ca90d0f843cc5c/llvm/lib/Support/APFloat.cpp#L2327)

I wanted to handle normal values separately for "almost equality" by calling  `areApproximatelyEqual<double> or areApproximatelyEqual<float>`, as shown in the code starting with comment "Both f and g are normal values.". Before doing that  I would like to handle all the non-normal cases. This is the same approach as how `APFloat::compare` is implemented. For example, before the [program point](https://github.com/llvm/llvm-project/blob/1c2280264058c26aee2bcd8fe4ca90d0f843cc5c/llvm/lib/Support/APFloat.cpp#L2376) , it handles all the  non-normal cases before handling the normal values. 

From the   `APFloat::compare`  implementation,  `cmpResult` equals to `APFloat::cmpEqual` or `APFloat::cmpUnordered` is returned for most of the non-normal cases, except when either of `f` or `g` is  not-[finite](https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/ADT/APFloat.h#L380) or `Zero` for which the `APFloat::compare`  returns compare results other than `cmpEqual` and `cmpUnordered`. 

We handled all the non normal cases using 3 checks which are 
```
  if (cmpResult == APFloat::cmpEqual) return true;

  // For NaNs, one option was to return false, when compare result is `cmpUnordered`. We chose the path as
  // shown below to avoid false alarms. 
  if (cmpResult == APFloat::cmpUnordered) return f.isNaN() == g.isNaN();   


  if (!f.isFiniteNonZero() || !g.isFiniteNonZero()) return false;
```

GleasonK (approved):
LGTM

PR #1029, opened 2023-02-01
Move and merge verifier of Set/GetDimensionSizeOp

zhouxin913:
SetDimensionSizeOp is very special, compared to other ops when involving bounded dynamism: it is possible that the output have bounds while the inputs has no bounds at all. Thus we can't use this util function to generate the encoding at the end of shape function: 
```
Attribute boundsToEncoding(Attribute prototype, ArrayRef<int64_t> bounds)
```
Because the `prototype`, which comes from `input.getEncoding()` maybe null. So in this PR it create bounds attr from dialect directly. 

Change `inferReturnTypes` to `inferReturnTypeComponents` and remove the TODO in comment as well.


stablehlo/dialect/TypeInference.cpp
```diff
@@ -2689,6 +2703,55 @@ LogicalResult inferSendOp(Dialect* dialect, std::optional<Location> location,
   return success();
 }
 
+LogicalResult inferSetDimensionSizeOp(
+    Dialect* dialect, std::optional<Location> location, Type operandType,
+    Value size, int64_t dim,
+    SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
+  if (auto sizeType = size.getType().dyn_cast<RankedTensorType>()) {
+    if (sizeType && sizeType.getRank() != 0)
```
sdasgup3, line 2711: no need to check `sizeType` as it has already been checked above.

sdasgup3: Or maybe
```
auto sizeType = size.getType().dyn_cast<RankedTensorType>();
if (sizeType && sizeType.getRank() != 0)
  emitOptionalError(location, "size operand should be of rank-0");
```

sdasgup3 (approved):

stablehlo/dialect/TypeInference.cpp
zhouxin913: Sure updated as your suggested.

stablehlo/dialect/TypeInference.h
```diff
@@ -231,8 +231,8 @@ LogicalResult inferGatherOp(
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes);
 
 LogicalResult inferGetDimensionSizeOp(
-    MLIRContext* context, std::optional<Location> location,
-    SmallVectorImpl<Type>& inferredReturnTypes);
+    MLIRContext* context, std::optional<Location> location, Type operandType,
+    int64_t dim, SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes);
```
burmako, line 235: `dimension`, according to how this is called in the ODS? (Same below).

stablehlo/dialect/TypeInference.h
```diff
@@ -231,8 +231,8 @@ LogicalResult inferGatherOp(
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes);
 
 LogicalResult inferGetDimensionSizeOp(
-    MLIRContext* context, std::optional<Location> location,
-    SmallVectorImpl<Type>& inferredReturnTypes);
+    MLIRContext* context, std::optional<Location> location, Type operandType,
```
burmako, line 234: Now that we're passing in `operandType`, we can drop the `context` parameter and obtain it from the type - `operandType.getContext()`.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -2689,6 +2703,54 @@ LogicalResult inferSendOp(Dialect* dialect, std::optional<Location> location,
   return success();
 }
 
+LogicalResult inferSetDimensionSizeOp(
+    Dialect* dialect, std::optional<Location> location, Type operandType,
+    Value size, int64_t dim,
+    SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
+  auto sizeType = size.getType().dyn_cast<RankedTensorType>();
+  if (sizeType && sizeType.getRank() != 0)
+    emitOptionalError(location, "size operand should be of rank-0");
```
burmako, line 2712: `return emitOptionalError(...)`?

burmako (changes requested):

stablehlo/dialect/TypeInference.cpp
zhouxin913: Good catch! Involved when addressing comments.

stablehlo/dialect/TypeInference.h
zhouxin913: Sure

stablehlo/dialect/TypeInference.h
zhouxin913: Sure.

burmako (approved):

PR #1034, opened 2023-02-02
Migrate away from using MLIR Attributes in `evalTransposeOp`

### LINKED ISSUE #1031: Refactor inferFooOp functions to use ArrayRef<int64_t> instead of DenseIntElementsAttr
This will: 1) make things more consistent (we're already using `ArrayRef<int64_t>` for some parameters of inferFooOp functions), 2) make things easier to use (it's easier to whip up an ArrayRef than a DenseIntElementsAttr).

ghpvnist:
This PR migrates away from using `DenseIntElementsAttr` in favor of `SmallVector<int64_t>`. The rationale is that it is harder to construct MLIR wrapper classes, and we want to avoid having a mix of custom classes (Tensor) and MLIR classes (DenseIntElementsAttr, with ConstantOp being an exception for now) so that we can use the `evalFoo` functions when a more complex op can benefit from reusing them.

closes #1031

### DIFF
--- a/stablehlo/reference/Interpreter.cpp
+++ b/stablehlo/reference/Interpreter.cpp
@@ -29,6 +29,16 @@ limitations under the License.
 namespace mlir {
 namespace stablehlo {
 
+namespace {
+
+SmallVector<int64_t> getSExtValues(DenseIntElementsAttr attr) {
+  SmallVector<int64_t> values;
+  for (auto i : attr) values.push_back(i.getSExtValue());
+  return values;
+}
+
+}  // namespace
+
 llvm::Expected<SmallVector<Tensor>> eval(func::FuncOp func,
                                          ArrayRef<Tensor> args) {
   if (func->getNumRegions() != 1) {
@@ -150,8 +160,9 @@ llvm::Expected<SmallVector<Tensor>> eval(func::FuncOp func,
       populateResults({runtimeResult});
     } else if (auto transposeOp = dyn_cast<TransposeOp>(op)) {
       Tensor runtimeOperand = fetchOperand(transposeOp.getOperand());
-      Tensor runtimeResult = evalTransposeOp(
-          transposeOp.getType(), runtimeOperand, transposeOp.getPermutation());
+      auto permutation = getSExtValues(transposeOp.getPermutation());
+      Tensor runtimeResult =
+          evalTransposeOp(runtimeOperand, permutation, transposeOp.getType());
       populateResults({runtimeResult});
     } else if (auto xorOp = dyn_cast<XorOp>(op)) {
       Tensor runtimeLhs = fetchOperand(xorOp.getLhs());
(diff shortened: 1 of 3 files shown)


stablehlo/reference/Interpreter.cpp
```diff
@@ -29,6 +29,16 @@ limitations under the License.
 namespace mlir {
 namespace stablehlo {
 
+namespace {
+
+SmallVector<int64_t> getSExtValues(DenseIntElementsAttr attr) {
```
sdasgup3, line 34: No need for this.

stablehlo/reference/Interpreter.cpp
```diff
@@ -150,8 +160,9 @@ llvm::Expected<SmallVector<Tensor>> eval(func::FuncOp func,
       populateResults({runtimeResult});
     } else if (auto transposeOp = dyn_cast<TransposeOp>(op)) {
       Tensor runtimeOperand = fetchOperand(transposeOp.getOperand());
-      Tensor runtimeResult = evalTransposeOp(
-          transposeOp.getType(), runtimeOperand, transposeOp.getPermutation());
+      auto permutation = getSExtValues(transposeOp.getPermutation());
```
sdasgup3, line 163:
```  
auto permutation =
          llvm::to_vector(transposeOp.getPermutation().getValues<int64_t>());
```

stablehlo/reference/Ops.h
```diff
@@ -29,7 +29,7 @@ Tensor evalCeilOp(const Tensor &operand, Type resultType);
 Tensor evalConstantOp(ElementsAttr value);
 Tensor evalCosineOp(const Tensor &operand, Type resultType);
 Tensor evalFloorOp(const Tensor &operand, Type resultType);
-Tensor evalIotaOp(int64_t iotaDimension, Type resultType);
+Tensor evalIotaOp(const int64_t iotaDimension, Type resultType);
```
sdasgup3, line 32: I generally avoid const on primitive type when passed by value as  it does not provide any meaningful signal to the user: Just the fact that the argument is passed by value, provides guarantee that the actual argument is not going to change.

sdasgup3 (approved):

stablehlo/reference/Ops.cpp
```diff
@@ -42,25 +42,22 @@ std::vector<int64_t> permute(ArrayRef<int64_t> array, ArrayRef<int64_t> perm) {
 
 Tensor evalAddOp(const Tensor &lhs, const Tensor &rhs, Type resultType) {
   Tensor result(resultType);
-  for (auto it = result.index_begin(); it != result.index_end(); ++it) {
+  for (auto it = result.index_begin(); it != result.index_end(); ++it)
     result.set(*it, lhs.get(*it) + rhs.get(*it));
-  }
```
burmako, line 47: We don't have a specific code style guide around whether or not to use braces for single-statement ifs and fors, but I think this change is fine. However, I would recommend to factor it out into a separate PR given that it ended up dominating the pull request.

burmako (approved):

PR #1039, opened 2023-02-02
Standardize formatting in Ops.cpp and Interpreter.cpp

ghpvnist:
Removes braces around functions which only have one line for conciseness.

### DIFF
--- a/stablehlo/reference/Interpreter.cpp
+++ b/stablehlo/reference/Interpreter.cpp
@@ -31,26 +31,24 @@ namespace stablehlo {
 
 llvm::Expected<SmallVector<Tensor>> eval(func::FuncOp func,
                                          ArrayRef<Tensor> args) {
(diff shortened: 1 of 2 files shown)


sdasgup3 (approved):

stablehlo/reference/Interpreter.cpp
```diff
@@ -31,26 +31,24 @@ namespace stablehlo {
 
 llvm::Expected<SmallVector<Tensor>> eval(func::FuncOp func,
                                          ArrayRef<Tensor> args) {
-  if (func->getNumRegions() != 1) {
```
burmako, line 34: If we're standardizing on this style guideline (which is not a bad idea), why limit ourselves to just these files?

ghpvnist: You're right. We should do this for all our C++ files. I've created a separate PR #1042 to address this.

PR #1040, opened 2023-02-02
Fix the notations for dynamic_update_slice and dynamic_slice

sdasgup3:
`adjusted_start_indices` seems like a 1-d tensor from the type of `shape(operand)` in `clamp(0, start_indices, shape(operand) -slice_sizes)`.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -2237,7 +2237,7 @@ contain the sizes of the slice for each dimension.
 
 More formally, `result[i0, ..., iR-1] = operand[j0, ..., jR-1]` where:
 
-* `jd = adjusted_start_indices[d][] + id`.
+* `jd = adjusted_start_indices[d] + id`.
 * `adjusted_start_indices = clamp(0, start_indices, shape(operand) -`
   `slice_sizes)`.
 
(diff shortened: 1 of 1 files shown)


docs/spec.md
```diff
@@ -2237,7 +2237,7 @@ contain the sizes of the slice for each dimension.
 
 More formally, `result[i0, ..., iR-1] = operand[j0, ..., jR-1]` where:
 
-* `jd = adjusted_start_indices[d][] + id`.
+* `jd = adjusted_start_indices[d] + id`.
```
burmako, line 2240: Given that `start_indices` is a vararg of 0-dimensional tensors, I imagine that `adjusted_start_indices` has the same type? If that's the case, then we need the second `[]` to turn a 0-dimensional tensor into a scalar, right?

burmako (changes requested):

docs/spec.md
sdasgup3: @burmako 
Thanks! I agree with why we should need.

Having noted that and assuming that `clamp` is same as `stablehlo.clamp`, there is still some confusion on the usage of `clamp` op  as  `stablehlo.clamp` which takes  arguments of the same shape, where as we have a var arg of `tensor<T>` (for `start_indices`) and `tensor<1xT>` (for `shape`). 


If you agree to that then we can add it to be done as part of #289.

ghpvnist, line 2240: `id = adjusted_start_indices[d] + jd` like below?

ghpvnist (changes requested):

docs/spec.md
sdasgup3: I also doubted the same before, but fortunately this formulation is correct. So we should not need this.

sdasgup3: The way I look at it is: 
index of larger window (the operand in this case) = start indices + the indices of the slice (the result here)

sdasgup3: Please let me know if this makes sense.

burmako (approved):

docs/spec.md
burmako: Sounds like a plan!

ghpvnist (approved):

sdasgup3:
@burmako Can you help merge it?

Issue #1028, opened 2023-01-31
Explore simplifying `infer<Op>` interfaces with minimal but sufficient arguments


sdasgup3:
Originated from the [comment](https://github.com/openxla/stablehlo/pull/1019#discussion_r1092362462) on #1019. 

This will simplify the infer<Op> interface by using the least possible amount of information which is sufficient to drive the type inference and thereby enable reuse of these methods.

zhouxin913:
This can be done for "almost all" ops, with the exceptions including (but not limit to):
1. SetDimensionSizeOp: https://github.com/openxla/stablehlo/pull/1029
2. (by Eugene) groups C and D in https://github.com/openxla/stablehlo/issues/867

burmako:
Assigning to Xin, since this falls into the overall theme of providing a public API for StableHLO.

PR #1041, opened 2023-02-02
Enable specializing dynamically-shaped StableHLO programs to static shapes

burmako:
Dynamic shapes in StableHLO are an appealing feature for ML frameworks, but not all ML compilers implement them. As a result, there is a practical need for infrastructure that would bridge this gap by specializing dynamically-shaped StableHLO programs to static shapes.

For example, we may want to take the following StableHLO program that performs a broadcasting addition of two dynamically-shaped tensors and then specialize it to static shapes based on ahead-of-time or just-in-time knowledge of the static shapes of its arguments.

```
func.func @main(
    %arg0: tensor<i32>,
    %arg1: tensor<?x4xf32>,
    %arg2: tensor<2x?x4xf32>) -> tensor<2x?x4xf32> {
  %0 = stablehlo.constant dense<2> : tensor<1xi32>
  %1 = stablehlo.reshape %arg0 : (tensor<i32>) -> tensor<1xi32>
  %2 = stablehlo.constant dense<4> : tensor<1xi32>
  %3 = stablehlo.concatenate %0, %1, %2, dim = 0 : (tensor<1xi32>, tensor<1xi32>, tensor<1xi32>) -> tensor<3xi32>
  %4 = stablehlo.dynamic_broadcast_in_dim %arg1, %3, dims = [1, 2] : (tensor<?x4xf32>, tensor<3xi32>) -> tensor<2x?x4xf32>
  %5 = stablehlo.add %4, %arg2 : tensor<2x?x4xf32>
  return %5 : tensor<2x?x4xf32>
}
```

Let's say that we know that %arg0 (which carries the size of the unbounded dimensions of %arg1 and %arg2) is equal to 3, then the corresponding statically-shaped program might look as follows.

```
func.func @main(
    %arg1: tensor<3x4xf32>,
    %arg2: tensor<2x3x4xf32>) -> tensor<2x3x4xf32> {
  %0 = stablehlo.broadcast_in_dim %arg1, dims = [1, 2] : (tensor<3x4xf32>) -> tensor<2x3x4xf32>
  %1 = stablehlo.add %0, %arg2 : tensor<2x3x4xf32>
  return %1 : tensor<2x3x4xf32>
}
```

This PR implements an initial version of --stablehlo-refine-shapes, a pass that propagates shape information across StableHLO programs. In this initial version, we're making the following simplifying assumptions:

(A1) Current design for StableHLO dynamism. We are using the current design for StableHLO dynamism, but we're also planning to update this pass to account for the changes in the upcoming dynamism RFC when it's ready. We found it useful to implement the first version of shape refinement prior to implementing the new design for dynamism and not the other way around in order to speed up the delivery.

(A2) Shape polymorphism. We are limiting ourselves to StableHLO programs that support shape polymorphism. In such programs, only data can depend on shapes (e.g. a broadcast can produce data of a dynamically-computed shape), but shapes cannot depend on data. Therefore, the shapes of all intermediate values can be computed from the shapes of the inputs.

(A3) Full specialization. We are only interested in full specialization to statically-shaped StableHLO programs. In alignment with A2, this means that we want the resulting programs to have statically-shaped types and constant shape values. Specialization to bounded dynamism might be an interesting avenue for future work.

(A4) Inlined programs. We assume that these programs consist of a single StableHLO function, i.e. that it's fine to ask users to inline their programs into one main function before running shape refinement.


stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -0,0 +1,937 @@
...
+// users of this op if any updates to its results have happened during execution
+// of the function.
+LogicalResult refineReturnTypes(PatternRewriter& rewriter, Operation* op,
+                                ArrayRef<Type> types) {
+  if (failed(refineValues(rewriter, op, op->getResults(), types)))
+    return failure();
+
+  // This `replaceOpWithIf` call doesn't actually change the IR, but
+  // it does ask the rewriter to visit all the users of this op. I haven't
+  // found a direct way of doing that.
```
zhouxin913, line 454: > "I haven't found a direct way of doing that."

How about we say "It could be optimized if there's more direct way of doing this."

stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -0,0 +1,937 @@
...
+//      and leaves that to the callers.
+LogicalResult refineValues(PatternRewriter& rewriter, Operation* op,
+                           ValueRange values, TypeRange types) {
+  if (values.size() != types.size())
+    return rewriter.notifyMatchFailure(op, [&](Diagnostic& diag) {
+      diag << "refineValues failed for " << types << ": expected "
+           << values.size() << " types, got " << types.size();
+    });
+
+  // Check whether `maybeRefinedReturnTypes` contain any new information with
```
zhouxin913, line 364: what is `maybeRefinedReturnTypes`, maybe an old var name of `TypeRange types`?

stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -0,0 +1,937 @@
...
+    auto unrefinedType = value.getType();
+    value.setType(refinedType);
+
+    // Special case: for `func.return`, guard the refinement with a
+    // `tensor.cast` and leave propagation of the refined return type to a
+    // dedicated pattern.
+    auto isFuncReturn = [](OpOperand& use) -> bool {
+      return isa<func::ReturnOp>(use.getOwner());
+    };
+    if (llvm::none_of(value.getUses(), isFuncReturn)) continue;
```
zhouxin913, line 433: Question: does this mean that we only do the refinement until a "return", i.e., the end of a function? I think so because the plan in the design doc says that the refinement only works within a mlir function. Just ask to confirm.

zhouxin913 (changes requested):
The PR is comprehensive and looks good to me. Leave a few comments for discussion. Will review the tests later.

stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -0,0 +1,937 @@
...
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(ConcatenateOp op,
+                                PatternRewriter& rewriter) const override {
+    auto resultType = op.getResult().getType().dyn_cast<RankedTensorType>();
+    if (!resultType || resultType.getRank() != 1)
+      return rewriter.notifyMatchFailure(op, "expected 1-dimensional type");
+
+    SmallVector<APInt> result;
+    for (Value operand : op->getOperands()) {
+      if (failed(matchInts(operand, result)))
```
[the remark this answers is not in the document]
burmako: The name and behavior matches the name and behavior of upstream `matchPattern` which fills in the provided arguments, e.g. `DenseIntElementsAttr`. Given the precedent, I think it might be acceptable to leave things as they are. What do you think?

stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -0,0 +1,937 @@
...
+    });
+  }
+};
+
+struct EvalConcatenateOpPattern : public OpRewritePattern<ConcatenateOp> {
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(ConcatenateOp op,
+                                PatternRewriter& rewriter) const override {
+    auto resultType = op.getResult().getType().dyn_cast<RankedTensorType>();
+    if (!resultType || resultType.getRank() != 1)
```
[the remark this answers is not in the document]
burmako: Given that this pattern is only used to constant fold shape computations, I don't anticipate that we'll want to extend it beyond 1-D tensors (given that shapes are modelled either as scalars or as 1-D tensors).

stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -0,0 +1,937 @@
...
+    auto operandType = op.getOperand().getType().dyn_cast<RankedTensorType>();
+    if (!operandType)
+      return rewriter.notifyMatchFailure(op, "expected ranked operand");
+    if (operandType.isDynamicDim(op.getDimension()))
+      return rewriter.notifyMatchFailure(op, "expected static dimension");
+
+    auto resultType = op.getResult().getType().cast<RankedTensorType>();
+    auto result = operandType.getDimSize(op.getDimension());
+    rewriter.replaceOpWithNewOp<ConstantOp>(
+        op, DenseIntElementsAttr::get<int32_t>(resultType, result));
```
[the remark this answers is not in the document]
burmako: Yes, we have a test for this pattern. It will start failing when we update the ODS to disallow `i32`.

stablehlo/transforms/StablehloRefineShapes.cpp
burmako: Reworded the comment.

stablehlo/transforms/StablehloRefineShapes.cpp
burmako: Good catch! Updated the comment.

stablehlo/transforms/StablehloRefineShapes.cpp
burmako: `applyPatternsAndFoldGreedily` goes through all the ops in the module. For each op, if appropriate it refines its type and propagates it to its uses. The supported uses include CHLO dialect, StableHLO dialect as well as `func.return` because of the fundamental role that it plays. However, because `func.return` is outside of our control, it needs special treatment.

Overall, yes, `func.return` does need special treatment, but not because of our traversal strategy - it's because it is the only op outside of our dialects that we support. Please let me know if this answers the question!

stablehlo/transforms/StablehloRefineShapes.cpp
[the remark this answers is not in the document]
zhouxin913: I see. That reminds me of the arg `inferReturnTypes` which is also appended inside the shape function. Looks like this is an MLIR coding pattern. Let's keep it as is then!

stablehlo/transforms/StablehloRefineShapes.cpp
zhouxin913: Yes, and thanks!

stablehlo/tests/stablehlo_refine_shapes.mlir
```diff
@@ -0,0 +1,521 @@
...
+  return %0 : tensor<*xf32>
+}
+
+// -----
+
+// CHECK-LABEL: func @update_region_type
+func.func @update_region_type(%arg0: tensor<i32>, %arg1: tensor<4xf32>) -> tensor<*xf32> {
+  // CHECK: "stablehlo.case"
+  // CHECK: -> tensor<4xf32>
+  // CHECK: stablehlo.abs{{.*}} : tensor<4xf32>
```
zhouxin913, line 513: I put this question in the "review cell" in last review and no "resolve" button related to that. Let me copy the question here again:

After looking at the tests, I'm wondering, will this PR breaks a lot of MHLO test? I mean we propagate the refined types in the later part of program (`abs` here) so the CHECKs will start to fail after the line of refined types because the output become different than what it was (tensor<*xf32> VS tensor<4xf32>? Does it need a switch to turn on/off these new passes for tests purpose?

zhouxin913 (approved):

stablehlo/tests/stablehlo_refine_shapes.mlir
burmako: Given that this PR is limited to a single StableHLO pass (changes to `dialect/` are NFC along the lines of #1028), I don't expect that it will affect MHLO at all. I think we should be good.

PR #1042, opened 2023-02-02
Standardize the usage of braces in control flow statements

ghpvnist:
Bootstrapping from #1039, this PR removes braces around functions which only have one line for conciseness across all cpp files.

### DIFF
--- a/stablehlo/dialect/Base.cpp
+++ b/stablehlo/dialect/Base.cpp
@@ -86,11 +86,12 @@ bool isCompatibleForHloTypeInference(Type tp1, Type tp2) {
   // Individual ops may introduce additional constraints.
   auto qtp1 = tp1.dyn_cast<quant::QuantizedType>();
   auto qtp2 = tp2.dyn_cast<quant::QuantizedType>();
-  if (qtp1 && qtp2)
+  if (qtp1 && qtp2) {
     if (qtp1.getStorageType() != qtp2.getStorageType() ||
         qtp1.getStorageTypeMin() != qtp2.getStorageTypeMin() ||
         qtp1.getStorageTypeMax() != qtp2.getStorageTypeMax())
       return false;
+  }
   auto etp1 = getExpressedTypeOrSelf(tp1);
   auto etp2 = getExpressedTypeOrSelf(tp2);
 
--- a/stablehlo/dialect/Base.cpp
+++ b/stablehlo/dialect/Base.cpp
@@ -75,11 +75,10 @@ bool isCompatibleForHloTypeInference(Type tp1, Type tp2) {
   // ops with partially inferred types to pass verification.
   auto stp1 = tp1.dyn_cast<ShapedType>();
   auto stp2 = tp2.dyn_cast<ShapedType>();
-  if (stp1 && stp2) {
+  if (stp1 && stp2)
     return succeeded(verifyCompatibleShapeWithBounds(stp1, stp2)) &&
            isCompatibleForHloTypeInference(stp1.getElementType(),
                                            stp2.getElementType());
-  }
 
   // Quantization: In the most general case, we allow any combination of
   // quantized/non-quantized across any combination of operands/results,
@@ -87,12 +86,11 @@ bool isCompatibleForHloTypeInference(Type tp1, Type tp2) {
   // Individual ops may introduce additional constraints.
   auto qtp1 = tp1.dyn_cast<quant::QuantizedType>();
   auto qtp2 = tp2.dyn_cast<quant::QuantizedType>();
-  if (qtp1 && qtp2) {
+  if (qtp1 && qtp2)
     if (qtp1.getStorageType() != qtp2.getStorageType() ||
         qtp1.getStorageTypeMin() != qtp2.getStorageTypeMin() ||
         qtp1.getStorageTypeMax() != qtp2.getStorageTypeMax())
       return false;
-  }
   auto etp1 = getExpressedTypeOrSelf(tp1);
   auto etp2 = getExpressedTypeOrSelf(tp2);
 
@@ -129,13 +127,11 @@ LogicalResult deriveShapeFromOperand(
 }
 
 TensorType getSameShapeTensorType(TensorType tensorType, Type elementType) {
-  if (auto rankedTensorTy = tensorType.dyn_cast<RankedTensorType>()) {
+  if (auto rankedTensorTy = tensorType.dyn_cast<RankedTensorType>())
     return RankedTensorType::get(rankedTensorTy.getShape(), elementType,
                                  rankedTensorTy.getEncoding());
-  }
-  if (auto unrankedTensorTy = tensorType.dyn_cast<UnrankedTensorType>()) {
+  if (auto unrankedTensorTy = tensorType.dyn_cast<UnrankedTensorType>())
     return UnrankedTensorType::get(elementType);
-  }
   llvm_unreachable("unhandled type");
 }
 
@@ -144,9 +140,8 @@ TensorType getSameShapeTensorType(TensorType tensorType, Type elementType) {
 //   Ex: tensor<4xcomplex<f32>>  -->  tensor<4xf32>
 Type createRealType(TensorType type) {
   auto elementTy = type.getElementType();
-  if (auto complexTy = elementTy.dyn_cast<ComplexType>()) {
+  if (auto complexTy = elementTy.dyn_cast<ComplexType>())
     elementTy = complexTy.getElementType();
-  }
   return hlo::getSameShapeTensorType(type, elementTy);
 }
 
@@ -158,20 +153,18 @@ LogicalResult verifyBounds(ArrayRef<int64_t> bounds, RankedTensorType type,
                            function_ref<InFlightDiagnostic()> emitError) {
   int64_t boundsLen = bounds.size();
   int64_t rank = type.getRank();
-  if (boundsLen != rank) {
+  if (boundsLen != rank)
     return emitError() << "Bounds length is " << boundsLen
                        << ", expected to be equal to rank(" << rank
                        << ") of the tensor";
-  }
 
   for (int64_t dim = 0; dim < rank; ++dim) {
     int64_t bound = bounds[dim];
     int64_t dimSize = type.getDimSize(dim);
-    if (bound != ShapedType::kDynamic && dimSize != ShapedType::kDynamic) {
+    if (bound != ShapedType::kDynamic && dimSize != ShapedType::kDynamic)
       return emitError() << "Static dimension " << dim
                          << " cannot have a bound, use ShapedType::kDynamic to "
                             "indicate a missing bound";
-    }
   }
 
   return success();
--- a/stablehlo/dialect/ChloOps.cpp
+++ b/stablehlo/dialect/ChloOps.cpp
@@ -84,9 +84,7 @@ ShapedTypeComponents getBroadcastType(
     DenseIntElementsAttr broadcastDimensionsAttr) {
   auto xRanked = x.dyn_cast<RankedTensorType>();
   auto yRanked = y.dyn_cast<RankedTensorType>();
-  if (!xRanked || !yRanked) {
-    return {elementType};
-  }
+  if (!xRanked || !yRanked) return {elementType};
 
   auto shapeX = xRanked.getShape();
   auto shapeY = yRanked.getShape();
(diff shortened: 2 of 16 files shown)


stablehlo/dialect/ChloOps.cpp
```diff
@@ -84,31 +84,27 @@ ShapedTypeComponents getBroadcastType(
...
+  if (!xRanked || !yRanked) return {elementType};
 
   auto shapeX = xRanked.getShape();
   auto shapeY = yRanked.getShape();
 
   // If no broadcast dimensions, assume "numpy" broadcasting.
   if (shapeX.size() == shapeY.size() || !broadcastDimensionsAttr) {
     llvm::SmallVector<int64_t, 4> outShape;
-    if (!mlir::OpTrait::util::getBroadcastedShape(shapeX, shapeY, outShape)) {
+    if (!mlir::OpTrait::util::getBroadcastedShape(shapeX, shapeY, outShape))
```
GleasonK, line 95: Don't omit when comment is inside of if

stablehlo/dialect/ChloOps.cpp
```diff
@@ -84,31 +84,27 @@ ShapedTypeComponents getBroadcastType(
...
-    }
     return {outShape, elementType};
   }
 
   auto shapeLarge = shapeX.size() > shapeY.size() ? shapeX : shapeY;
   auto shapeSmall = shapeX.size() <= shapeY.size() ? shapeX : shapeY;
 
   auto broadcastDimensions = broadcastDimensionsAttr.getValues<APInt>();
-  if (broadcastDimensions.size() != shapeSmall.size()) {
+  if (broadcastDimensions.size() != shapeSmall.size())
```
GleasonK, line 105: Don't omit when comment inside of if

stablehlo/transforms/StablehloLegalizeToVhlo.cpp
```diff
@@ -40,52 +40,45 @@ Attribute convertAttrToVhlo(Attribute stablehloAttr) {
   // Handle StableHLO attributes.
   // The logic that handles attributes from other dialects (e.g. builtin
   // attributes) lives below.
-  if (auto attr = stablehloAttr.dyn_cast<stablehlo::ChannelHandleAttr>()) {
+  if (auto attr = stablehloAttr.dyn_cast<stablehlo::ChannelHandleAttr>())
```
GleasonK, line 43: Let's keep these for consistency within the file. Strange for some of the `if`s in a block to have braces while others don't.

stablehlo/transforms/VhloLegalizeToStablehlo.cpp
```diff
@@ -174,20 +167,19 @@ class VhloToStablehloOpConverter : public OpConversionPattern<VhloOpTy> {
     // vhlo.case that uses a variadic number of regions which means an
     // additional argument for the generic builder.
     VhloToStablehloOp<VhloOpTy> stablehloOp;
-    if constexpr (std::is_same<VhloOpTy, vhlo::CaseOpV1>::value) {
+    if constexpr (std::is_same<VhloOpTy, vhlo::CaseOpV1>::value)
```
GleasonK, line 170: if with else shouldn't drop braces

GleasonK (changes requested):
Personally not a fan of omitting braces, a lot of modern languages (golang for example) require braces on all if/for statements to avoid the [pitfalls](https://softwareengineering.stackexchange.com/questions/16528/single-statement-if-block-braces-or-no) of omitting braces, such as [dangling else](https://www.cs.drexel.edu/~jpopyack/Courses/GovSchool/2005/Wi04/lectures/08.2_nested_conditionals/DanglingElse.html?CurrentSlide=3#:~:text=The%20dangling%20else%20problem%20results,the%20last%20if%20statement%20used.), etc.

That said, our codebase is a mix of Google and LLVM style guides (LLVM in the places where Google style guide is lenient):
- [Google style guide on braces](https://google.github.io/styleguide/cppguide.html#Conditionals)
   + _if an if statement has no else or else if clauses, then the curly braces for the controlled statement or the line breaks inside the curly braces may be omitted if as a result the entire if statement appears on either a single line_
- [LLVM style guide on braces](https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements)
  + _When writing the body of an if, else, or for/while loop statement, we prefer to omit the braces to avoid unnecessary line noise. However, braces should be used in cases where the omission of braces harm the readability and maintainability of the code._

I'd suggest we stay on the side of the (slightly more strict) Google style guide, and lets only do this when the entire statement fits on one line. I'd be ok with if it fits on one line after as well, but otherwise the more complicated ones we should probably wrap. That said, interested in others feedback as well. At a bare minimum we should be as strict as LLVM style guide.

stablehlo/transforms/VhloLegalizeToStablehlo.cpp
ghpvnist: The LLVM guide suggests dropping if the if/else has a single statement.

stablehlo/dialect/Base.cpp
```diff
@@ -75,24 +75,22 @@ bool isCompatibleForHloTypeInference(Type tp1, Type tp2) {
...
-  }
 
   // Quantization: In the most general case, we allow any combination of
   // quantized/non-quantized across any combination of operands/results,
   // and some differences in quantization parameters across operands/results.
   // Individual ops may introduce additional constraints.
   auto qtp1 = tp1.dyn_cast<quant::QuantizedType>();
   auto qtp2 = tp2.dyn_cast<quant::QuantizedType>();
-  if (qtp1 && qtp2) {
+  if (qtp1 && qtp2)
```
GleasonK, line 89: LLVM style guide:

```
// Use braces on the outer `if` to avoid a potential dangling `else`
// situation.
if (isa<VarDecl>(D)) {
  if (shouldProcessAttr(A))
    handleAttr(A);
}
```

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -1935,10 +1909,9 @@ LogicalResult SelectOp::reifyReturnTypeShapes(
 //===----------------------------------------------------------------------===//
 
 LogicalResult SetDimensionSizeOp::verify() {
-  if (auto size = this->getSize().getType().dyn_cast<RankedTensorType>()) {
+  if (auto size = this->getSize().getType().dyn_cast<RankedTensorType>())
     if (size.getRank() != 0)
       return emitOpError() << "size operand should be of rank-0";
```
GleasonK, line 1914:
```
// Use braces on the outer `if` to avoid a potential dangling `else`
// situation.
if (isa<VarDecl>(D)) {
  if (shouldProcessAttr(A))
    handleAttr(A);
}
```

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2477,10 +2450,9 @@ ParseResult WhileOp::parse(OpAsmParser& parser, OperationState& result) {
     if (succeeded(parser.parseOptionalRParen())) break;
     if (failed(parser.parseComma())) return failure();
   } while (true);
-  if (!operands.empty()) {
+  if (!operands.empty())
     if (parser.parseColon() || parser.parseTypeList(result.types))
       return failure();
```
GleasonK, line 2455:
```
// Use braces on the outer `if` to avoid a potential dangling `else`
// situation.
if (isa<VarDecl>(D)) {
  if (shouldProcessAttr(A))
    handleAttr(A);
}
```

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2679,19 +2649,16 @@ static ParseResult parseStruct(
...
-        if (failed(parser.parseOptionalComma())) return parser.parseGreater();
-        seen[index] = true;
-        foundOne = true;
-      }
+      if (failed(parser.parseOptionalKeyword(keyword))) continue;
+      if (seen[index])
+        return parser.emitError(parser.getCurrentLocation())
+               << "duplicated `" << keyword << "` entry";
+      if (parseEqual.empty() || parseEqual[index])
+        if (failed(parser.parseEqual())) return failure();
```
GleasonK, line 2657:
```
// Use braces on the outer `if` to avoid a potential dangling `else`
// situation.
if (isa<VarDecl>(D)) {
  if (shouldProcessAttr(A))
    handleAttr(A);
}
```

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3802,15 +3759,14 @@ LogicalResult verifyScatterOp(std::optional<Location> location,
...
   // P1.
-  if (scatterIndicesTypeRanked) {
+  if (scatterIndicesTypeRanked)
     if (indexVectorDim > scatterIndicesType.getRank() || indexVectorDim < 0)
       return emitOptionalError(
           location,
           "expects scatter index leaf dimension to be within [0, "
           "rank(scatter_indices) + 1. rank(scatter_indices) is ",
           scatterIndicesType.getRank(), " and scatter index leaf dimension is ",
           indexVectorDim, ".");
```
GleasonK, line 3769:
```
// Use braces on the outer `if` to avoid a potential dangling `else`
// situation.
if (isa<VarDecl>(D)) {
  if (shouldProcessAttr(A))
    handleAttr(A);
}
```

stablehlo/dialect/TypeInference.cpp
```diff
@@ -4004,7 +3959,7 @@ LogicalResult verifySelectAndScatterOp(
   auto windowDimsOrErr =
       convert1DAttribute(windowDimensions, location, "window_dimensions");
   if (failed(windowDimsOrErr)) return failure();
-  if (operandType.hasRank()) {
+  if (operandType.hasRank())
     if (operandType.getRank() !=
         static_cast<int64_t>((*windowDimsOrErr).size()))
       return emitOptionalError(
```
GleasonK, line 3965: braces on outer if

stablehlo/transforms/VhloLegalizeToStablehlo.cpp
```diff
@@ -41,47 +41,42 @@ namespace {
...
-  if (auto attr = vhloAttr.dyn_cast<vhlo::DotDimensionNumbersAttr>()) {
+  if (auto attr = vhloAttr.dyn_cast<vhlo::DotDimensionNumbersAttr>())
     return stablehlo::DotDimensionNumbersAttr::get(
         attr.getContext(), attr.getLhsBatchingDimensions(),
         attr.getRhsBatchingDimensions(), attr.getLhsContractingDimensions(),
         attr.getRhsContractingDimensions());
-  }
   if (auto attr = vhloAttr.dyn_cast<vhlo::FftTypeAttr>()) {
     RETURN_CONVERTED_ENUM_ATTR(FftType);
   }
```
GleasonK, line 71: Still not sure what's the right thing to do for a chain of if's where some have braces and others do not. This feels strange with a mix.

stablehlo/transforms/VhloLegalizeToStablehlo.cpp
GleasonK: Looked in LLVM codebase, and the rule applies to `if constexpr` as well.

stablehlo/dialect/Version.cpp
```diff
@@ -27,9 +27,8 @@ namespace {
 // Precondition that numRef is a valid decimal digit.
 static int64_t parseNumber(llvm::StringRef numRef) {
   int64_t num;
-  if (numRef.getAsInteger(/*radix=*/10, num)) {
+  if (numRef.getAsInteger(/*radix=*/10, num))
```
GleasonK, line 30: I'd probably leave braces around macro definitions like `llvm_unreachable`. The macro expansion of `llvm_unreachable` is single line, but I'm not sure there's a guarantee that this won't change. This is similar to what you did for 

```
  if (auto attr = vhloAttr.dyn_cast<vhlo::ComparisonDirectionAttr>()) {
    RETURN_CONVERTED_ENUM_ATTR(ComparisonDirection);
  }
```

In terms of the LLVM Style Guide, I would categorize this under _"Maintainability is harmed if..."_

stablehlo/transforms/VhloLegalizeToStablehlo.cpp
ghpvnist: Since this one is not clear, let's just leave this file unchanged for consistency in the file. Wdyt?

stablehlo/dialect/Version.cpp
ghpvnist: We actually have multiple instances of `llvm_unreachable` in our codebase without braces, notably `ChloAttributes.cpp` and `StablehloAttributes.cpp`

GleasonK: This is less than ideal, but I suppose it would be bad for upstream to change a function-like macro to be more than a single statement. I found several instances of where this macro would cause issues upstream if it were changed as well. Given that it is treated as a function (and uses function capitalization) it is likely ok to have this.

GleasonK (approved):

burmako (approved):

PR #1044, opened 2023-02-03
Update llvm lint tool with flag to disable SHA256 check

GleasonK:
The SHA256 check takes awhile, and really only impacts llvm revision bump changelists. Given that I run all lint scripts on pre-commit, I'd prefer a way to only verify commit, if commit changes, then I can worry about sha256 validation. In CI we will always check both.

Also I refactored the script a bit, there was so much global state that it was difficult to make modifications, modified to have minimal global state and functions that echo their results. We lose some logging with this, but the error messages we see are still pretty good.

Mismatch error messages:

```bash
# ... commit mismatch

$ ./build_tools/github_actions/lint_llvm_commit.sh -s .
Validating LLVM commit hash...
Commit mismatch:
1c1
< f58de2125caf75ec0d40bc3e094a93c0b314a66a
---
> 858de2125caf75ec0d40bc3e094a93c0b314a66a

Auto-fix using:
  $ lint_llvm_commit.sh -f <path/to/stablehlo/root>

# ... SHA256 mismatch

$ ./build_tools/github_actions/lint_llvm_commit.sh .
Validating LLVM commit hash...
Commit hashes match.
Validating LLVM SHA256 hash...
...SHA256 mismatch:
1c1
< 896047c57addc2a349defab0c7fb84979aeac039ed9ada2901cd2e0ca88fffb6
---
> a96047c57addc2a349defab0c7fb84979aeac039ed9ada2901cd2e0ca88fffb6

Auto-fix using:
  $ lint_llvm_commit.sh -f <path/to/stablehlo/root>

# ...  404

$ ./build_tools/github_actions/lint_llvm_commit.sh .
Validating LLVM commit hash...
Commit hashes match.
Validating LLVM SHA256 hash...
...SHA256 mismatch:
1c1
< a96047c57addc2a349defab0c7fb84979aeac039ed9ada2901cd2e0ca88fffb6
---
> Error 404 downloading LLVM at commit '858de2125caf75ec0d40bc3e094a93c0b314a66a'.

Auto-fix using:
  $ lint_llvm_commit.sh -f <path/to/stablehlo/root>
```

### DIFF
--- a/build_tools/github_actions/lint_llvm_commit.sh
+++ b/build_tools/github_actions/lint_llvm_commit.sh
@@ -14,12 +14,15 @@
 print_usage() {
   echo "Usage: $0 [-f] <path/to/stablehlo/root>"
   echo "    -f           Auto-fix LLVM commit mismatch."
+  echo "    -s           Skip sha256 hash validation."
 }
 
 FORMAT_MODE='validate'
-while getopts 'f' flag; do
+VALIDATE_SHA256='true'
+while getopts 'fs' flag; do
   case "${flag}" in
     f) FORMAT_MODE="fix" ;;
+    s) VALIDATE_SHA256="false" ;;
     *) print_usage
        exit 1 ;;
   esac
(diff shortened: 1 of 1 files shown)


ghpvnist (approved):
LGTM Thanks for making the script tidy :)

burmako (approved):

PR #1050, opened 2023-02-03
Update interpreter testing guideline with coverage criterion

sdasgup3:
The PR updates the interpreter testing guidelines  based on https://github.com/openxla/stablehlo/pull/996#discussion_r1093590287.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -160,7 +160,21 @@ We can use a combination of following rules to decide it:
        choose the one with the following primitive type, in decreasing order of
        preference: integer, floating-point, boolean, complex.
 
-**(G2) How about adding tests dedicated for testing the interpreter
+**(G2) How much of test coverage we should aim for?**
+
+$\ge$ 90%.
+
+**(G3) How do we decide on the number of tests needed to cover an op's
+behavior?**
+
+Ideally, it is desirable to have minimal tests which meets the set coverage
+criteria. We should write tests with best effort" to cover multiple behaviors
+of an op under test. By doing that most of the simpler ops will end up having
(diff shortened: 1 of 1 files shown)


docs/reference.md
```diff
@@ -160,7 +160,21 @@ We can use a combination of following rules to decide it:
...
+
+**(G3) How do we decide on the number of tests needed to cover an op's
+behavior?**
+
+Ideally, it is desirable to have minimal tests which meets the set coverage
+criteria. We should write tests with best effort" to cover multiple behaviors
+of an op under test. By doing that most of the simpler ops will end up having
+one test. From collective "best efforts" from reviewer and author, we should be
+able to get a near close to the ideal goal. Taking assistance from the coverage
+tools will help us when to stop.
```
burmako, line 175: I agree with the spirit of the proposal, but I think the wording should be more prescriptive. Please find a proposal below:

```
The goal is to comprehensively cover the logic of the interpreter for the op 
(i.e. all corner cases of the implementation) with a minimal number of tests. 

Minimizing the number of tests is important for maintainability. The fewer tests 
we have, the easier it is to review them and to make sure that they 
comprehensively cover the op. As a result, we expect that most of the simpler 
ops will end up having just one test.

If due to some good reason comprehensive coverage is impractical, then it is 
fine to stop at >= 90%. This will be decided on case-by-case basis during pull 
request review.
```

burmako (changes requested):

docs/reference.md
sdasgup3: Thanks!

burmako (approved):

PR #1053, opened 2023-02-04
Move VhloTypeDefs.cpp.inc into VhloTypes.cpp

burmako:
As discussed in #899, it is desirable to move the following includes from VhloOps.cpp to VhloTypes.cpp:

  #include "stablehlo/dialect/VhloTypeInterfaces.cpp.inc"
  #define GET_TYPEDEF_CLASSES
  #include "stablehlo/dialect/VhloTypeDefs.cpp.inc"

This way, vhlo_types becomes self-contained, i.e. we will no longer have to link its users with vhlo_ops. Given the fairly involved dependency structure around the VHLO dialect, this is a desirable design property.


GleasonK (approved):

stablehlo/dialect/VhloOps.cpp
```diff
@@ -176,20 +146,39 @@ VhloDialect::VhloDialect(MLIRContext* context)
...
+  //         >();
+  //
+  // However, Dialect::addTypes doesn't work for our situation where we want
+  // to decouple the vhlo_ops and vhlo_types targets. VhloTypeDefs.h.inc only
+  // includes forward declarations of TypeStorage structs, and that's not enough
+  // for Dialect::addTypes to compile.
+  //
+  // Therefore, we work around by introducing this function and then
+  // reimplementing Dialect::addTypes as shown below.
+  addTypesWithoutRegistering<
```
GleasonK, line 171: Any concerns with this longer term? Is this in-model enough where upstream changes aren't too likely to break it?

burmako: Time will tell, I guess. If upstream changes break this, we can always revert to the previous state of `vhlo_types` not being self-contained.

PR #1054, opened 2023-02-04
Change return type for shape function of bounds

zhouxin913:
This is a quick fix revert back the return type of inferMost/LeastType methods from ShapeType to Type, to accommodate the tuple type in MHLO tests.

### DIFF
--- a/stablehlo/dialect/Base.cpp
+++ b/stablehlo/dialect/Base.cpp
@@ -368,7 +368,9 @@ LogicalResult inferMostSpecificTypeComponents(
 
   auto rankedResultType = (*inferredTypeOrErr).dyn_cast<RankedTensorType>();
   if (!rankedResultType) {
-    inferredReturnShapes.emplace_back((*inferredTypeOrErr).cast<ShapedType>());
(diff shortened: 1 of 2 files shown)


stablehlo/dialect/Base.cpp
```diff
@@ -368,7 +368,7 @@ LogicalResult inferMostSpecificTypeComponents(
 
   auto rankedResultType = (*inferredTypeOrErr).dyn_cast<RankedTensorType>();
   if (!rankedResultType) {
-    inferredReturnShapes.emplace_back(*inferredTypeOrErr);
+    inferredReturnShapes.emplace_back((*inferredTypeOrErr).cast<ShapedType>());
```
sdasgup3, line 371: Can you also add the fallback code for the if case:

```
auto inferredShapeType = (*inferredTypeOrErr).dyn_cast<ShapedType>();
if(!inferedShapeType) return failure();
inferredReturnShapes.emplace_back(inferredShapeType);
```

sdasgup3:
The change is to accommodate both tuple types and unranked tensor types.

sdasgup3 (approved):

burmako (approved):

PR #1058, opened 2023-02-04
Add interpreter for GatherOp

### LINKED ISSUE #976: Add interpreter for gather

ghpvnist:

```
(I1) operand tensor.
(I2) start_indices tensor of integer type.
(I3) offset_dims 1-dimensional tensor constant of type `si64`.
(I4) collapsed_slice_dims 1-dimensional tensor constant of type `si64`.
(I5) start_index_map 1-dimensional tensor constant of type `si64`.
(I6) index_vector_dim constant of type `si64`.
(I7) slice_sizes 1-dimensional tensor constant of type `si64`.
(I8) indices_are_sorted constant of type `i1`.
(C1) rank(`operand`) $=$ size(`offset_dims`) $+$
     size(`collapsed_slice_dims`).
(C2) $0 \le$ `index_vector_dim` $\le$ rank(`start_indices`).
(C3) size(`start_index_map`) $=$
     `index_vector_dim` $\lt$ rank(`start_indices`) ?
     dim(`start_indices`, `index_vector_dim`) : 1.
(C4) All dimensions in `offset_dims` are unique and sorted in ascending
     order.
(C5) $0 \le$ `offset_dims`[i] $\lt$ rank(`result`) $\forall i$
     such that $0 \le$ i $\lt$ size(`offset_dims`).
(C6) All dimensions in `collapsed_slice_dims` are unique and sorted in
     ascending order.
(C7) $0 \le$ `collapsed_slice_dims`[i] $\lt$ size(`slice_sizes`)
      $\forall i$ such that $0 \le$ i $\lt$ size(`collapsed_slice_dims`).
(C8) `slice_sizes`[i] $\le$ 1 $\forall i \in$ `collapsed_slice_dims`.
(C9) All dimensions in `start_index_map` are unique.
(C10) $0 \le$ `start_index_map`[i] $\lt$ rank(`operand`) $\forall i$
     such that $0 \le$ i $\lt$ size(`start_index_map`).
(C11) size(`slice_sizes`) $=$ rank(`operand`).
(C12) $0 \le$ `slice_sizes`[i] $\le$ dim(`operand`, i) $\forall i$
      such that $0 \le$ i $\lt$ size(`slice_sizes`).
(C13) `shape(result)` $=$ `combine(batch_dim_sizes, offset_dim_sizes)`
      where:
      * `batch_dim_sizes` = `shape(start_indices)` except that the dimension size
        of `start_indices` corresponding to `index_vector_dim` is not included.
      * `offset_dim_sizes` = `shape(slice_sizes)` except that the dimension sizes
        in `slice_sizes` corresponding to `collapsed_slice_dims` are not included.
      * `combine` puts `batch_dim_sizes` at axes corresponding to `batch_dims` and
        `offset_dim_sizes` at axes corresponding to `offset_dims`.
(C14) `operand` and `result` have the same element type.
```

```
I2: a) start_indices is not a tensor of integer type. (Covered by ODS).
I3: a) offset_dims is not a 1-dimensional tensor. (Covered by ODS).
    b) element_type(offset_dims) != `si64`. (Covered by ODS).
I4: a) collapsed_slice_dims is not a 1-dimensional tensor. (Covered by ODS).
    b) element_type(collapsed_slice_dims) != `si64`. (Covered by ODS).
I5: a) start_index_map is not a 1-dimensional tensor. (Covered by ODS).
    b) element_type(start_index_map) != `si64`. (Covered by ODS).
I6: a) type(index_vector_dim) != `si64`. (Covered by ODS).
I7: a) slice_sizes is not a 1-dimensional tensor.
    b) element_type(slice_sizes) != `si64`. (Covered by ODS).
I8: a) element_type(indices_are_sorted) != `i1`. (Covered by ODS).
(C1) a) rank(operand) != size(offset_dims) + size(collapsed_slice_dims).
(C2) a) index_vector_dim < 0.
     b)  index_vector_dim > rank(start_indices).
(C3) a) size(start_index_map) != dim(start_indices, index_vector_dim) if index_vector_dim < rank(start_indices).
     b) size(start_index_map) != 1 if index_vector_dim >= rank(start_indices).
(C4) a) offset_dims values are not unique.
     b) offset_dims values are not sorted in ascending order.
(C5) a) offset_dims[i] < 0 for any i.
     b) offset_dims[i] >= rank(result) for any i.
(C6) a) collapsed_slice_dims values are not unique.
     b) collapsed_slice_dims are not sorted in ascending order.
(C7) a) collapsed_slice_dims[i] < 0 for any i.
     b) collapsed_slice_dims[i] >= size(slice_sizes) for any i.
(C8) a) slice_sizes[i] > 1 for any i in collapsed_slice_dims.
(C9) a) start_index_map values are not unique.
(C10) a) start_index_map[i] < 0 for any i.
      b) start_index_map[i] >= rank(operand) for any i.
(C11) a) size(slice_sizes) != rank(operand).
(C12) a) slice_sizes[i] < 0 for any i.
      b) slice_sizes[i] > dim(operand, i) for any i.
(C13) no negative test needed since it's just inferring the shape.
(C14) element_type(operand) !=  element_type(result). (Covered by ODS).
```

```
I7a: slice_sizes is not a 1-dimensional tensor.
C1a: rank(operand) != size(offset_dims) + size(collapsed_slice_dims).
C2a: index_vector_dim < 0.
C2b: index_vector_dim > rank(start_indices).
C3a: size(start_index_map) != dim(start_indices, index_vector_dim) if index_vector_dim < rank(start_indices).
C3b: size(start_index_map) != 1 if index_vector_dim >= rank(start_indices).
C4a: offset_dims values are not unique.
C4b: offset_dims values are not sorted in ascending order.
C5a: offset_dims[i] < 0 for any i.
C5b: offset_dims[i] >= rank(result) for any i.
C6a: collapsed_slice_dims values are not unique.
C6b: collapsed_slice_dims are not sorted in ascending order.
C7a: collapsed_slice_dims[i] < 0 for any i.
C7b: collapsed_slice_dims[i] >= size(slice_sizes) for any i.
C8a: slice_sizes[i] > 1 for any i in collapsed_slice_dims.
C9a: start_index_map values are not unique.
C10a: start_index_map[i] < 0 for any i.
C10b: start_index_map[i] >= rank(operand) for any i.
C11a: size(slice_sizes) != rank(operand).
C12a: slice_sizes[i] < 0 for any i.
C12b: slice_sizes[i] > dim(operand, i) for any i.
```

Also fixed typo (C15) -> (C14)

closes #976

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -2803,15 +2803,14 @@ More formally, `result[result_index] = operand[operand_index]` where:
 * `batch_dims` = [`d` for `d` in `axes(result)` and `d` not in `offset_dims`].
 * `batch_index` = [`result_index[d]` for `d` in `batch_dims`].
 * `start_index` =
-  * `start_indices[bi0, ..., :, ..., biN]` where `bi` are individual
-      elements in `batch_index` and `:` is inserted at the `index_vector_dim`
-      index, if `index_vector_dim` < `rank(start_indices)`.
+  * `start_indices[bi0, ..., :, ..., biN]` where `bi` are individual elements in
+    `batch_index` and `:` is inserted at the `index_vector_dim` index, if
+    `index_vector_dim` < `rank(start_indices)`.
   * `[start_indices[batch_index]]` otherwise.
-* For `di` in `axes(operand)`,
-  * `full_start_index[di]` =
-    `clamp(start_index[dj], 0, dim(operand, di) - slice_sizes[di])` if
-    `di = start_index_map[dj]`.
-  * `full_start_index[di]` = `0` otherwise.
+* For `d_operand` in `axes(operand)`,
+  * `full_start_index[d_operand]` = `start_index[d_start]` if
+    `d_operand = start_index_map[d_start]`.
+  * `full_start_index[d_operand]` = `0` otherwise.
 * `offset_index` = [`result_index[d]` for `d` in `offset_dims`].
 * `full_offset_index` = `[oi0, ..., 0, ..., oiN]` where `oi` are individual
   elements in `offset_index`, and `0` is inserted at indices from
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -2810,7 +2810,7 @@ More formally, `result[result_index] = operand[operand_index]` where:
 * For `di` in `axes(operand)`,
   * `full_start_index[di]` =
     `clamp(start_index[dj], 0, dim(operand, di) - slice_sizes[di])` if
-    `do = start_index_map[dj]`.
+    `di = start_index_map[dj]`.
   * `full_start_index[di]` = `0` otherwise.
 * `offset_index` = [`result_index[d]` for `d` in `offset_dims`].
 * `full_offset_index` = `[oi0, ..., 0, ..., oiN]` where `oi` are individual
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -2873,7 +2873,7 @@ behavior is undefined. More formally, for all `id < jd` from `indices(result)`,
     in `slice_sizes` corresponding to `collapsed_slice_dims` are not included.
   * `combine` puts `batch_dim_sizes` at axes corresponding to `batch_dims` and
    `offset_dim_sizes` at axes corresponding to `offset_dims`.
-* (C15) `operand` and `result` have the same element type.
+* (C14) `operand` and `result` have the same element type.
 
 #### Examples
 
@@ -2909,6 +2909,8 @@ behavior is undefined. More formally, for all `id < jd` from `indices(result)`,
 //            ]
 //          ]
 ```
+
+&nbsp;[More Examples](../stablehlo/tests/interpret_gather.mlir)
 
 ### get_dimension_size
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -2828,7 +2828,7 @@ behavior is undefined. More formally, for all `id < jd` from `indices(result)`,
 
 | Label | Name                   | Type                                         | Constraints                   |
 |-------|------------------------|----------------------------------------------|-------------------------------|
-| (I1)  | `operand`              | tensor                                       | (C1), (C10-C12), (C15)        |
+| (I1)  | `operand`              | tensor                                       | (C1), (C10-C12), (C14)        |
 | (I2)  | `start_indices`        | tensor of integer type                       | (C2), (C3), (C13)             |
 | (I3)  | `offset_dims`          | 1-dimensional tensor constant of type `si64` | (C1), (C4), (C5), (C13)       |
 | (I4)  | `collapsed_slice_dims` | 1-dimensional tensor constant of type `si64` | (C1), (C6), (C7), (C8), (C13) |
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -2807,9 +2807,11 @@ More formally, `result[result_index] = operand[operand_index]` where:
       elements in `batch_index` and `:` is inserted at the `index_vector_dim`
       index, if `index_vector_dim` < `rank(start_indices)`.
   * `[start_indices[batch_index]]` otherwise.
-* For `do` in `axes(operand)`,
-  * `full_start_index[do]` = `start_index[ds]` if `do = start_index_map[ds]`.
-  * `full_start_index[do]` = `0` otherwise.
+* For `di` in `axes(operand)`,
+  * `full_start_index[di]` =
+    `clamp(start_index[dj], 0, dim(operand, di) - slice_sizes[di])` if
+    `do = start_index_map[dj]`.
+  * `full_start_index[di]` = `0` otherwise.
 * `offset_index` = [`result_index[d]` for `d` in `offset_dims`].
 * `full_offset_index` = `[oi0, ..., 0, ..., oiN]` where `oi` are individual
   elements in `offset_index`, and `0` is inserted at indices from
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -1073,6 +1073,7 @@ static LogicalResult verifyGather(
       return emitOptionalError(location, "index_vector_dim ", indexVectorDim,
                                " is out of bounds for start indices with rank ",
                                startIndicesShape.getRank());
+
     // gather_c3
     bool impliedTrailingDim = indexVectorDim == startIndicesShape.getRank();
     if (impliedTrailingDim || !startIndicesShape.isDynamicDim(indexVectorDim)) {
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -1046,30 +1046,18 @@ LogicalResult validateScatterDimensionNumbers(
   return success();
 }
 
-// Verify the following properties:
-//  P1. Verify no repeat in start_index_map.
-//  P2. Verify 0 <= start_index_map[i] < rank(operand), for every i.
-//  P3. Verify 0 <= index_vector_dim <= rank(start_indices).
-//  P4. Verify size(start_index_map) == shape(start_indices)[index_vector_dim].
-//  P5. Verify offset_dims is_sorted and not repeated.
-//  P6. Verify collapsed_slice_dims is_sorted and not repeated.
-//  P7. Verify rank(operand) == size(offset_dims) + size(collapsed_slice_dims).
-//  P8. Verify slice_sizes has rank of 1.
-//  P9. Verify size(slice_sizes) == rank(operand).
-//  P10. Verify 0 <= collapsed_slice_dims[i] < size(slice_sizes) for all items.
 static LogicalResult verifyGather(
     std::optional<Location> location, ShapeAdaptor operandShape,
     ShapeAdaptor startIndicesShape, ShapeAdaptor sliceSizesShape,
     ArrayRef<int64_t> offsetDims, ArrayRef<int64_t> collapsedSliceDims,
     ArrayRef<int64_t> startIndexMap, int64_t indexVectorDim) {
-  // P1.
-  // Check startIndexMap
+  // gather_c9
   if (hasDuplicates(startIndexMap))
     return emitOptionalError(location,
                              "expects start_index_map to not repeat, got: [",
                              startIndexMap, "]");
 
-  // P2.
+  // gather_c10
   for (int64_t i = 0; i < static_cast<int64_t>(startIndexMap.size()); ++i)
     if (startIndexMap[i] < 0 ||
         (operandShape.hasRank() && startIndexMap[i] >= operandShape.getRank()))
@@ -1078,14 +1066,14 @@ static LogicalResult verifyGather(
           " is out of bounds for ", "operand rank ", operandShape.getRank());
 
   if (startIndicesShape.hasRank()) {
-    // P3.
+    // gather_c2
     // index_vector_dim == start_indices.rank implies a trailing 1 on the shape
     // of start_indices.
     if (indexVectorDim > startIndicesShape.getRank() || indexVectorDim < 0)
       return emitOptionalError(location, "index_vector_dim ", indexVectorDim,
                                " is out of bounds for start indices with rank ",
                                startIndicesShape.getRank());
-
+    // gather_c3
     bool impliedTrailingDim = indexVectorDim == startIndicesShape.getRank();
     if (impliedTrailingDim || !startIndicesShape.isDynamicDim(indexVectorDim)) {
       int64_t effectiveDimSize;
@@ -1093,7 +1081,6 @@ static LogicalResult verifyGather(
         effectiveDimSize = 1;
       else
         effectiveDimSize = startIndicesShape.getDimSize(indexVectorDim);
-      // P4.
       if (effectiveDimSize != static_cast<int64_t>(startIndexMap.size()))
         return emitOptionalError(
             location, "start_index_map size (", startIndexMap.size(),
@@ -1102,7 +1089,7 @@ static LogicalResult verifyGather(
     }
   }
 
-  // P5.
+  // gather_c4
   if (!llvm::is_sorted(offsetDims))
     return emitOptionalError(
         location, "expects offset_dims to be sorted, got: [", offsetDims, "]");
@@ -1110,7 +1097,7 @@ static LogicalResult verifyGather(
     return emitOptionalError(
         location, "expects offset_dims to not repeat, got: [", offsetDims, "]");
 
-  // P6.
+  // gather_c6
   if (!llvm::is_sorted(collapsedSliceDims))
     return emitOptionalError(
         location, "expects collapsed_slice_dims to be sorted, got: [",
@@ -1120,7 +1107,7 @@ static LogicalResult verifyGather(
         location, "expects collapsed_slice_dims to not repeat, got: [",
         collapsedSliceDims, "]");
 
-  // P7.
+  // gather_c1
   int64_t impliedOperandRank = offsetDims.size() + collapsedSliceDims.size();
   if (operandShape.hasRank() && operandShape.getRank() != impliedOperandRank)
     return emitOptionalError(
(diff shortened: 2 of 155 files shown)


stablehlo/reference/Ops.cpp
```diff
@@ -165,6 +165,67 @@ Tensor evalFloorOp(const Tensor &operand, TensorType resultType) {
...
+                          .getIntegerValue()
+                          .getSExtValue(),
+                      operand.getType().getShape()[dim] - sliceSizes[dim]));
+    }
+    Index sliceSizesBeginIdx(sliceSizes.size());
+    auto sliceSizesIt = IndexSpaceIterator(sliceSizes, sliceSizesBeginIdx);
+    auto sliceSizesItEnd = IndexSpaceIterator(sliceSizes, {});
+    for (; sliceSizesIt != sliceSizesItEnd; ++sliceSizesIt) {
+      Index operandIdx;
+      for (auto i = 0; i < operand.getType().getRank(); ++i)
```
sdasgup3, line 210:
```   
   Index operandIdx = operandBaseIdx + (*sliceSizesIt);
```

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -3591,387 +3591,421 @@ func.func @gather(%operand : tensor<2x4x9xi32>, %start_indices : tensor<*xi32>)
...
+    indices_are_sorted = false
+  } : (tensor<?x?x2xi32>, tensor<*xi32>) -> tensor<*xi32>
+  func.return %res : tensor<*xi32>
 }
 
 // -----
 
-func.func @gather(%operand : tensor<2x4x9xi32>, %start_indices : tensor<1x5x2xi32>) -> tensor<1x5x8xi32> {
-  // expected-error@+1 {{start_index_map[0]: -2 is out of bounds for operand rank 3}}
+func.func @gather_c15(%operand : tensor<2x4x9xi32>, %start_indices : tensor<1x5x2xi32>) -> tensor<3xi32> {
```
sdasgup3, line 3964: c14 here and below?

stablehlo/tests/interpret_gather.mlir
```diff
@@ -0,0 +1,21 @@
...
+                                       [[17, 18], [19, 20], [21, 22], [23, 24]]]> : tensor<3x4x2xi64>
+  %start_indices = stablehlo.constant dense<[[[0, 0], [1, 0], [2, 1]],
+                                             [[0, 1], [1, 1], [0, 2]]]> : tensor<2x3x2xi64>
+  %result = "stablehlo.gather"(%operand, %start_indices) {
+    dimension_numbers = #stablehlo.gather<
+      offset_dims = [1, 3],
+      collapsed_slice_dims = [2],
+      start_index_map = [1, 0],
+      index_vector_dim = 2>,
+    slice_sizes = dense<[2, 2, 1]> : tensor<3xi64>,
```
sdasgup3, line 15: Can we have the same example as the spec.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1089,22 +1077,21 @@ static LogicalResult verifyGather(
...
-    // P3.
+    // gather_c2
     // index_vector_dim == start_indices.rank implies a trailing 1 on the shape
     // of start_indices.
     if (indexVectorDim > startIndicesShape.getRank() || indexVectorDim < 0)
       return emitOptionalError(location, "index_vector_dim ", indexVectorDim,
                                " is out of bounds for start indices with rank ",
                                startIndicesShape.getRank());
-
+    // gather_c3
```
sdasgup3, line 1087: Is this a space which we can remove?

sdasgup3 (commented):
Reviewed as per the checklist . Everything looks good except that the implementation of the  gather can be written to be aligned more with the spec. I will suggest soon.

stablehlo/reference/Ops.cpp
```diff
@@ -233,6 +233,57 @@ Tensor evalFloorOp(const Tensor &operand, TensorType resultType) {
   return result;
 }
 
+Tensor evalGatherOp(const Tensor &operand, const Tensor &startIndices,
```
sdasgup3, line 236: The current implementation does not seem to match the spec. I can roughly understand what the current implementation is doing, but it may not be obvious to users familiar with the spec and expecting strong correlation between the two artifacts. Here are some of the non-obvious differences:
1.   `nonCollapsedSliceDims` to store the axes of operand which are not mentioned in the `collapsedSliceDims`. 
2. Use of clamp (which IMO we should not need if follow the spec).
3. Starting with startIndices to derive the input/output indices.   

IMO, implementing the spec, as it is, should be possible in this case. One non-trivial aspect could be 
```
start_index = start_indices[bi0, ..., :, ..., biN]
``` 

but can be implemented using a slice op with start index as the  `batch_indices` and limit index should be `[bi0, ..., , bik + dim(start_indices, index_vector_dims), ... biN] `, wherte  `k =  index_vector_dims`

ghpvnist: Thanks for the suggestion updated PR.

stablehlo/reference/Index.h
```diff
@@ -34,10 +34,11 @@ class IndexSpaceIterator {
...
   IndexSpaceIterator(Sizes shape, std::optional<Index> index)
       : shape_(shape), index_(index) {
-    if (index && !index->inBounds(shape))
-      llvm::report_fatal_error(
-          "Incompatible index and shape found while creating "
-          "an IndexSpaceIterator");
+    if (index && !index->inBounds(shape)) index_ = std::nullopt;
+  }
+
+  IndexSpaceIterator(Sizes shape) : shape_(shape) {
```
sdasgup3, line 40: Not sure why we need this. 

The earlier constructor has the provision of a `std::nullopt` to index param, which would initialize `index_` the way this constructor is doing. To me it seems the constructor is introduce to avoid explicitly passing that `std::nullopt`.

stablehlo/reference/Ops.cpp
```diff
@@ -769,6 +781,60 @@ Tensor evalFloorOp(const Tensor &operand, ShapedType resultType) {
...
+    Index batchIndex;
+    for (auto batchDim : batchDims) batchIndex.push_back((*resultIt)[batchDim]);
+
+    Index startIndex;
+    Sizes batchIndexOffset(startIndices.getRank(), 1);
+    if (indexVectorDim < startIndices.getRank()) {
+      batchIndex.insert(batchIndex.begin() + indexVectorDim, 0);
+      batchIndexOffset[indexVectorDim] =
+          startIndices.getShape()[indexVectorDim];
+    }
```
sdasgup3, line 805: I am not sure if i get this. 

from the POV of slice op, the relation between `batch_index` and `limit_indices` (for the slice) is:
```
If batchIndex = [bi0, ..., bik, ..., biN] for k = index_vector_dims
then limit_indices = [bi0, ..., bik + dim(start_indices, index_vector_dims) -1, ..., biN]
```

A few concerns:
1.  why are we making the index value at `indexVectorDim` zero.
2.  Can we avoid the update of `batchIndex`?

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -3540,32 +3540,32 @@ func.func @reduce_precision_invalid_mantissa(%arg: tensor<2x4xf32>) -> tensor<2x
 func.func @gather(%operand : tensor<2x4x9xi32>, %start_indices : tensor<1x5x2xi32>) -> tensor<1x5x8xi32> {
```
sdasgup3, line 3540: Do we have a test to check `c13`?

stablehlo/reference/Ops.cpp
```diff
@@ -769,6 +781,60 @@ Tensor evalFloorOp(const Tensor &operand, ShapedType resultType) {
...
+    if (!llvm::is_contained(offsetDims, i)) batchDims.push_back(i);
+
+  for (auto resultIt = result.index_begin(); resultIt != result.index_end();
+       ++resultIt) {
+    Index batchIndex;
+    for (auto batchDim : batchDims) batchIndex.push_back((*resultIt)[batchDim]);
+
+    Index startIndex;
+    Sizes batchIndexOffset(startIndices.getRank(), 1);
+    if (indexVectorDim < startIndices.getRank()) {
```
sdasgup3, line 801: Can we add one more test when this `if` check is false.

docs/spec.md
```diff
@@ -2802,9 +2802,11 @@ More formally, `result[result_index] = operand[operand_index]` where:
       elements in `batch_index` and `:` is inserted at the `index_vector_dim`
       index, if `index_vector_dim` < `rank(start_indices)`.
   * `[start_indices[batch_index]]` otherwise.
-* For `do` in `axes(operand)`,
-  * `full_start_index[do]` = `start_index[ds]` if `do = start_index_map[ds]`.
-  * `full_start_index[do]` = `0` otherwise.
+* For `di` in `axes(operand)`,
+  * `full_start_index[di]` =
+    `clamp(start_index[dj], 0, dim(operand, di) - slice_sizes[di])` if
```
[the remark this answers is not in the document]
ghpvnist: I believe the way the spec is formulated and with the constraints put, it is not possible to construct an out-of-bound `operand_index` when start_index is clamped. The spec says the behavior is implementation defined, but in our case, the behavior would be undefined.

stablehlo/reference/Index.h
ghpvnist: Thanks for pointing it out, I meant to remove this.

stablehlo/tests/ops_stablehlo.mlir
ghpvnist: It was an oversight to not label this one. Updated the function name in `infer_stablehlo.mlir` which covers this.

stablehlo/reference/Ops.cpp
```diff
@@ -850,6 +862,61 @@ Tensor evalFloorOp(const Tensor &operand, ShapedType resultType) {
...
+  for (auto i = 0; i < result.getRank(); ++i)
+    if (!llvm::is_contained(offsetDims, i)) batchDims.push_back(i);
+
+  for (auto resultIt = result.index_begin(); resultIt != result.index_end();
+       ++resultIt) {
+    Index batchIndex;
+    for (auto batchDim : batchDims) batchIndex.push_back((*resultIt)[batchDim]);
+
+    Index sliceStartIndices(batchIndex);
+    Index sliceLimitIndices(Sizes(sliceStartIndices) + 1);
```
sdasgup3, line 881: why `+ 1` ? It should the same batch index as sliceStartIndices?

stablehlo/reference/Ops.cpp
sdasgup3: Ah, sorry I got confused here. Thanks!

sdasgup3 (approved):

docs/spec.md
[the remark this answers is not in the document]
sdasgup3: Sounds reasonable to me. Let us keep it open for seeking opinion from @burmako.

stablehlo/reference/Ops.cpp
ghpvnist: The slice index is from [startIndex, limitIndex). Without +1, it would be [batchIndex, batchIndex) which would be nothing.

stablehlo/reference/Ops.cpp
```diff
@@ -850,6 +862,61 @@ Tensor evalFloorOp(const Tensor &operand, ShapedType resultType) {
   return result;
 }
 
+Tensor evalGatherOp(const Tensor &operand, const Tensor &startIndices,
```
burmako, line 865: Very nice to see the implementation so close to the spec. In https://github.com/ghpvnist/stablehlo/pull/5, I made it even closer. PTAL.

burmako (changes requested):
Only reviewed evalGatherOp.

stablehlo/reference/Ops.cpp
ghpvnist: > Why do we need a clamp here if we can just do if (operandIndex < operand.getShape()) below? The alternative is obviously correct, whereas it wasn't obvious to me what to make of this code.

You're right that we can just do if (operandIndex < operand.getShape()) and it will still be a correct reference implementation as the spec says that the behavior is implementation defined if operandIndex is OOB. However, JAX assumes the clamping behavior, which I chose to implement (and to enable these tests).

burmako: I would say that we shouldn't enable tests that depend on implementation-defined behavior.

How about in this PR we do `if (operandIndex < operand.getShape())`, and then in a follow-up PR we update spec.md to clamp and enable the rest of the tests? In principle, this all can be done in this PR, but I think it would be better to compartmentalize this change.

burmako: For example, we did the same for SignOp - first, implemented the spec as is, and then clarified it, updated the interpreter and enabled more tests: https://github.com/openxla/stablehlo/pull/1395.

burmako: "is also implementation specific right". Agreed - personally, I prefer it because it's probably the simplest way to handle the out of bounds situation.

ghpvnist: The reason why we need clamping though is that that `Index operandIndex = fullStartIndex + fullOffsetIndex;` could result in correct (jax behavior) bounds when clamped as they can get "too" OOB otherwise. There is a maximum limit to how large `fullOffsetIndex` can get (limited by `resultIndex`), but it is arbitrarily large for fullStartIndex which derives from `startIndices` operand.

burmako: Huh this was totally non-obvious to me. Looks like this behavior is not covered by the spec at all then.

ghpvnist: Updated the PR now. One of the interpreter test also fails, so I've removed it for now.

docs/spec.md
```diff
@@ -2873,7 +2874,7 @@ behavior is undefined. More formally, for all `id < jd` from `indices(result)`,
     in `slice_sizes` corresponding to `collapsed_slice_dims` are not included.
   * `combine` puts `batch_dim_sizes` at axes corresponding to `batch_dims` and
    `offset_dim_sizes` at axes corresponding to `offset_dims`.
-* (C15) `operand` and `result` have the same element type.
+* (C14) `operand` and `result` have the same element type.
```
burmako, line 2877: Do we also need to update the `Inputs` section now that this constraint has been renamed?

stablehlo/reference/Ops.cpp
```diff
@@ -897,6 +933,49 @@ Tensor evalFloorOp(const Tensor &operand, ShapedType resultType) {
...
+    for (auto d : offsetDims) offsetIndex.push_back((*resultIt)[d]);
+
+    Index fullOffsetIndex(offsetIndex.size() + collapsedSliceDims.size(), 0);
+    for (size_t i = 0, oi = 0; i < fullOffsetIndex.size(); ++i) {
+      if (llvm::is_contained(collapsedSliceDims, i)) continue;
+      fullOffsetIndex[i] = offsetIndex[oi++];
+    }
+
+    Index operandIndex = fullStartIndex + fullOffsetIndex;
+    result.set(*resultIt, operand.get(operandIndex));
```
burmako, line 974: Let's guard with `if (operandIndex.inBounds(operand.getShape())` to avoid outright crashes?

stablehlo/reference/Ops.cpp
```diff
@@ -1248,10 +1325,9 @@ SmallVector<Tensor> evalSortOp(ArrayRef<Tensor> inputs, Axis dimension,
 
     // Instead of literally putting the slices together into a vector of tuples,
     // we're representing these tuples with integer handles, with each handle
-    // being an index within the slice.
-    // Then, instead of sorting a vector of tuples, we're sorting a vector of
-    // handles, and the comparator knows how to use these handles to fetch
-    // the actual input elements being compared.
+    // being an index within the slice. Then, instead of sorting a vector of
+    // tuples, we're sorting a vector of handles, and the comparator knows how
+    // to use these handles to fetch the actual input elements being compared.
```
burmako, line 1330: Is there a particular reason why the formatting needs to be changed here? I think the previous version is a bit easier to read.

burmako (changes requested):

docs/spec.md
ghpvnist: Thanks for the catch.

stablehlo/reference/Ops.cpp
ghpvnist: Agreed, changed back.

burmako (approved):

docs/spec.md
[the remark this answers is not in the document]
ghpvnist: I agree. As discussed below, I'll create a separate PR to address this.

PR #1071, opened 2023-02-06
Revert #958: Verify compare_type for compare_op

sdasgup3:
Reverting https://github.com/openxla/stablehlo/pull/958 to unblock downstream integration.

## Details
The tests which are failing downstream falls into two categories:
(1) Verifier failing on compareOps with  wrong combination of element-type and compare_type. 
(2) Verifier  failing on creating `compareOp` with `compare_type == NOTYPE`. Note that creating a compareOp with `NOTYPE` is still allowed. 
(3) Type inference failures: As the implementation of verification and type inference is unified for this op, the above mentioned categories also contributed to type inference failures.


burmako (approved):
Before merging, can you also update the PR description to provide more details about what specifically fails downstream?

PR #1079, opened 2023-02-07
Add interpreter for ExpOp

zhouxin913:
(I1) operand: tensor of floating-point or complex type (covered by ODS)
(C1) operand and result have the same type. (covered by ODS)

Note that there is only one constrain of this op and covered by parent `StableHLO_UnaryElementwiseOp`, so no tests in `ops_stablehlo.mlir`.


stablehlo/tests/interpret_exponential.mlir
```diff
@@ -0,0 +1,13 @@
+// RUN: stablehlo-interpreter --interpret -split-input-file %s | FileCheck %s
+
+// CHECK-LABEL: Evaluated results of function: exponential_op_test_i64
+func.func @exponential_op_test_i64() -> tensor<4xf64> {
```
ghpvnist, line 4: Can you add a test for `complex<f64>` element type?

ghpvnist (approved):
Minor comment

stablehlo/tests/interpret_exponential.mlir
zhouxin913: Sure, also changed names from %0 %1 to %operand %result

sdasgup3 (approved):

ghpvnist:

stablehlo/dialect/StablehloOps.td
```diff
@@ -304,8 +304,9 @@ def StableHLO_CosineOp: StableHLO_UnaryElementwiseOp<"cosine",
 }
 
 def StableHLO_ExpOp: StableHLO_UnaryElementwiseOp<"exponential",
-    [Pure, HLO_CompatibleOperandsAndResultType], HLO_FpOrComplexTensor> {
-  let summary = "Exponential operator";
+    [Pure, HLO_CompatibleOperandsAndResultType /* exponential_c1 */],
+     HLO_FpOrComplexTensor> {
+  let summary = "Exponential operation";
```
sdasgup3, line 309: Add input constraints as per  the checklist.

stablehlo/dialect/StablehloOps.td
```diff
@@ -304,8 +304,9 @@ def StableHLO_CosineOp: StableHLO_UnaryElementwiseOp<"cosine",
 }
 
 def StableHLO_ExpOp: StableHLO_UnaryElementwiseOp<"exponential",
-    [Pure, HLO_CompatibleOperandsAndResultType], HLO_FpOrComplexTensor> {
-  let summary = "Exponential operator";
+    [Pure, HLO_CompatibleOperandsAndResultType /* exponential_c1 */],
```
burmako, line 307: Are we using the `/*foo_c1*/` or `/* foo_c1 */` convention? I think the checklist proposes the former.

stablehlo/reference/Element.h
```diff
@@ -109,6 +109,9 @@ Element cosine(const Element &e);
 /// Returns element-wise floor of Element object.
 Element floor(const Element &e);
 
+/// Returns element-wise exponential of Element object.
+Element exponential(const Element &el);
+
```
burmako, line 114: Let's order alphabetically - here and in the .cpp file.

stablehlo/tests/interpret_exponential.mlir
```diff
@@ -0,0 +1,28 @@
+// RUN: stablehlo-interpreter --interpret -split-input-file %s | FileCheck %s
+
+// CHECK-LABEL: Evaluated results of function: exponential_op_test_i64
+func.func @exponential_op_test_i64() -> tensor<5xf64> {
+  %operand = stablehlo.constant dense<[0.0, 1.0, -2.3, 3.1415926535897931, 0x7FF0000000000000]> : tensor<5xf64>
```
burmako, line 5: What's our convention for the number of values in these tests? The spec says "As an exception to this rule, floating-point exceptions in StableHLO programs have well-defined behavior", so does this mean that we need to test all floating-point corner cases of numerical ops?

docs/status.md
```diff
@@ -87,7 +87,7 @@ one of the following tracking labels.
 | dynamic_slice            | yes           | yes          | yes            | yes             | no          |
 | dynamic_update_slice     | yes           | yes          | yes            | yes             | no          |
 | einsum                   | no            | revisit      | no             | yes             | no          |
-| exponential              | yes           | yes          | yes            | yes             | no          |
+| exponential              | yes           | yes          | yes            | yes             | yes         |
```
burmako, line 90: Should we also update TypeInference.cpp with `/*exponential_c1*/` next to `INFER_RETURN_TYPE_COMPONENTS_FROM_OPERANDS(ExpOp)`?

burmako (changes requested):

stablehlo/dialect/StablehloOps.td
zhouxin913: Ah sure, updated in this PR and also other PRs as well.

stablehlo/reference/Element.h
zhouxin913: I'm mislead by the `cosine()` in Element.cpp which is out of order. 
Fixed `exponential` here. Not touch `cosine` in case conflict when merge.

docs/status.md
zhouxin913: Good catch! I was thinking of this but not sure how to add a comment for all INFER_RETURN_TYPE_COMPONENTS_FROM_OPERANDS macros.
Take your suggestion that add this for ExpOp only in this specific PR.

stablehlo/dialect/StablehloOps.td
zhouxin913: Exp's all inputs are all from StableHLO_UnaryElementwiseOp, so no clear to add exp_i1 in StableHLO_UnaryElementwiseOp. Same issue for other ops. Please let me know if you think otherwise.

sdasgup3:
```
def StableHLO_ExpOp: StableHLO_UnaryElementwiseOp<"exponential",
    [Pure, HLO_CompatibleOperandsAndResultType /*exponential_c1*/],
     HLO_FpOrComplexTensor /*exponential_i1*/> {
```

stablehlo/tests/interpret_exponential.mlir
sdasgup3: The interpreter test input selection is solely based on the coverage of the implementation.  In this case the example we have in the spec should be good enough to fully cover our implementation. 

I agree that there are many interesting cases, but the current scope is to stick to basic ones providing good coverage.

zhouxin913: Sure, resolved by remove the 0x7FF0000000000000 and keep the rest.

docs/status.md
zhouxin913: I changed my mind: I noticed that all ops with `INFER_RETURN_TYPE_COMPONENTS_FROM_OPERANDS` has `HLO_CompatibleOperandsAndResultType` in ODS thus will have xyz_c1 here. So I believe there is no need to add another /*exponential_c1*/ after the micro. 
What do you think of it?

sdasgup3: I also think that adding comment on `HLO_CompatibleOperandsAndResultType` is sufficient?

stablehlo/tests/interpret_exponential.mlir
```diff
@@ -0,0 +1,27 @@
+// RUN: stablehlo-interpreter --interpret -split-input-file %s | FileCheck %s
+
+// CHECK-LABEL: Evaluated results of function: exponential_op_test_i64
+func.func @exponential_op_test_i64() -> tensor<4xf64> {
+  %operand = stablehlo.constant dense<[0.0, 1.0, -2.3, 3.1415926535897931]> : tensor<4xf64>
+  %result = stablehlo.exponential %operand : tensor<4xf64>
+  func.return %result : tensor<4xf64>
```
sdasgup3, line 7: Can we keep the test in the spec? That would affect the coverage but is more familiar to the user of the spec.

sdasgup3 (approved):

docs/spec.md
```diff
@@ -2464,6 +2464,8 @@ Performs element-wise exponential operation on `operand` tensor and produces a
 // %result: (-1.13120438, 2.47172667)
```
sdasgup3, line 2464: I think we should have one test per op spec. Make sure that is in sync with the Example in the ODS.

stablehlo/dialect/StablehloOps.td
```diff
@@ -306,8 +306,9 @@ def StableHLO_CosineOp: StableHLO_UnaryElementwiseOp<"cosine",
 }
 
 def StableHLO_ExpOp: StableHLO_UnaryElementwiseOp<"exponential",
-    [Pure, HLO_CompatibleOperandsAndResultType], HLO_FpOrComplexTensor> {
-  let summary = "Exponential operator";
+    [Pure, HLO_CompatibleOperandsAndResultType /*exponential_c1*/],
+     HLO_FpOrComplexTensor /*exponential_i1*/> {
+  let summary = "Exponential operation";
```
sdasgup3, line 311: The ODS example seems off.

sdasgup3 (approved):

stablehlo/tests/interpret_exponential.mlir
zhouxin913: Sure, aligned Spec with test. And also update them with f64: the highest precision of float type.

stablehlo/dialect/StablehloOps.td
zhouxin913: Aligned.

docs/spec.md
zhouxin913: Now Spec and tests are aligned for the 2 tests. ODS now has one example and is aligned with one example of them.

docs/status.md
zhouxin913: Agreed. Let Eugene continue the review.

burmako: Agreed that it's sufficient.

docs/spec.md
```diff
@@ -2456,14 +2456,16 @@ Performs element-wise exponential operation on `operand` tensor and produces a
...
-%result = "stablehlo.exponential"(%operand) : (tensor<2x2xf32>) -> tensor<2x2xf32>
-// %result: [[1.0, 2.71828183], [7.38905610, 20.08553692]]
+%result = "stablehlo.exponential"(%operand) : (tensor<2x2xf64>) -> tensor<2x2xf64>
+// %result: [[1.0, 2.7182818284590451], [7.3890560989306504, 20.085536923187668]]
 
 // %operand: (1.0, 2.0)
-%result = "stablehlo.exponential"(%operand) : (tensor<complex<f32>>) -> tensor<complex<f32>>
-// %result: (-1.13120438, 2.47172667)
+%result = "stablehlo.exponential"(%operand) : (tensor<complex<f64>>) -> tensor<complex<f64>>
+// %result: (-1.1312043837568135, 2.4717266720048188)
```
burmako, line 2464: Drop the second example? The checklist asks to keep just one example in the spec.

burmako (changes requested):
One more minor change, and we're good to go.

docs/spec.md
zhouxin913: Sure, fixed.

burmako (approved):

PR #1086, opened 2023-02-08
Add interpreter for LogOp

zhouxin913:
(I1) operand: tensor of floating-point or complex type (coved by ODS)
(C1) operand and result have the same type. (coved by ODS)

Note that there is only one constrain of this op and coved by ODS, so no tests in `ops_stablehlo.mlir`.

Corner cases are implemented in this way: See code in `Element.cpp`:
// These cases are implementation-defined, and the behavior of this interpreter:
// For floats: if the input are invalid (<= 0), the result is 0xFFF0000000000000
// For complex numbers: if the input are invalid, the result is 
//   [0xFFF0000000000000 : f64, 0.000000e+00 : f64]
Element log(const Element &el){}


ghpvnist:

stablehlo/dialect/StablehloOps.td
```diff
@@ -396,8 +396,9 @@ def StableHLO_IsFiniteOp: StableHLO_UnaryElementwiseOp<"is_finite", [Pure,
 }
 
 def StableHLO_LogOp: StableHLO_UnaryElementwiseOp<"log",
-    [Pure, HLO_CompatibleOperandsAndResultType], HLO_FpOrComplexTensor> {
-  let summary = "Logarithm operator";
+    [Pure, HLO_CompatibleOperandsAndResultType /* log_c1 */],
+     HLO_FpOrComplexTensor> {
+  let summary = "Logarithm operation";
```
sdasgup3, line 401: Add input constraints as per  the checklist.

stablehlo/reference/Element.cpp
```diff
@@ -314,6 +314,16 @@ Element cosine(const Element &el) {
       [](std::complex<double> e) { return std::cos(e); });
 }
 
+// These cases are implementation-defined, and the behavior of this interpreter:
+// For floats: if the input are invalid (<= 0), the result is 0xFFF0000000000000
+// For complex numbers: if the input are invalid, the result is
+//   [0xFFF0000000000000 : f64, 0.000000e+00 : f64]
+Element log(const Element &el) {
```
sdasgup3, line 321: IMO we do not have to call it out the case for `<=0`. Because the spec for `std::log` already subsumed that

```
If a domain error occurs, an implementation-defined value is returned
```

How about we remove the comment.

stablehlo/reference/Element.h
```diff
@@ -112,6 +112,9 @@ Element cosine(const Element &e);
 /// Returns floor of Element object.
 Element floor(const Element &e);
 
+/// Returns element-wise log of Element object.
```
sdasgup3, line 115: remove `element-wise`?

docs/spec.md
```diff
@@ -3083,6 +3083,8 @@ Performs element-wise logarithm operation on `operand` tensor and produces a
 // %result: (0.80471896, 1.10714871)
```
sdasgup3, line 3083: Keep just one  example in the spec (first one in this case)

stablehlo/dialect/StablehloOps.td
```diff
@@ -398,8 +398,9 @@ def StableHLO_IsFiniteOp: StableHLO_UnaryElementwiseOp<"is_finite", [Pure,
 }
 
 def StableHLO_LogOp: StableHLO_UnaryElementwiseOp<"log",
```
sdasgup3, line 400: The example used in the spec is not the same used in the ODS example. We should make them equal.

stablehlo/tests/interpret_log.mlir
```diff
@@ -0,0 +1,25 @@
+// RUN: stablehlo-interpreter --interpret -split-input-file %s | FileCheck %s
+
+// CHECK-LABEL: Evaluated results of function: log_op_test_i64
+func.func @log_op_test_i64() -> tensor<3xf64> {
+  %operand = stablehlo.constant dense<[1.0, 3.1415926535897931, 0x7FF0000000000000]> : tensor<3xf64>
+  %result = stablehlo.log %operand : tensor<3xf64>
+  func.return %result : tensor<3xf64>
+  // CHECK-NEXT: tensor<3xf64>
+  // CHECK-NEXT: 0.000000e+00 : f64
+  // CHECK-NEXT: 1.1447298858494002 : f64
```
sdasgup3, line 10: Use the same example as the one used in spec?

stablehlo/reference/Element.cpp
zhouxin913: Sure, as the description already has it.

sdasgup3 (approved):

stablehlo/dialect/StablehloOps.td
```diff
@@ -398,8 +398,9 @@ def StableHLO_IsFiniteOp: StableHLO_UnaryElementwiseOp<"is_finite", [Pure,
...
-    [Pure, HLO_CompatibleOperandsAndResultType], HLO_FpOrComplexTensor> {
-  let summary = "Logarithm operator";
+    [Pure, HLO_CompatibleOperandsAndResultType /*log_c1*/],
+     HLO_FpOrComplexTensor /*log_i1*/> {
+  let summary = "Logarithm operation";
```
burmako, line 403: "Log operation".

burmako, line 403: Can you also change the summary for AbsOp to say "Abs operation"?

burmako (changes requested):

stablehlo/dialect/StablehloOps.td
zhouxin913: Looks like it is already "Abs operation"? https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/StablehloOps.td#L207
Please let me know if I didn't get it.

burmako: Apologies! I meant ExpOp to say "Exp operation".

zhouxin913: I see. The summary "Exponential" is align with the op name "exponential", instead of the class name "ExpOp": I think it's better because the readers of Spec will use "stablehlo.exponential" (not "exp") in their StableHLO program.
For log op, yes "Log operation" is good because the op name is indeed "log".
What do you think of it?

burmako: Got it, understood that this is intentional, and that there are multiple possibilities for this. The checklist recommends "Exp operation", but we can discuss this further after merging this PR. cc @sdasgup3 to drive to a conclusion and update the checklist or the ODS as needed.

burmako (approved):

PR #1088, opened 2023-02-08
Don't load Tensor dialect in StableHLO dialect constructor

burmako:
I recently noticed this code when reviewing #849, and I'm not sure why we need it there.

This seems like a pretty strong statement about a fundamental role of the Tensor dialect in the workings of the StableHLO dialect, and I don't think we have established that yet.

It would seem that we've inherited this from MHLO when bootstrapping StableHLO (#1), but I don't think I understand the reasoning on the MHLO side either. This change was introduced as part of an LLVM integrate (https://github.com/tensorflow/mlir-hlo/commit/ba0346b071f4212008943c50f47766e677992b56), and the commit description doesn't go into detail about motivation.

Given that, I propose to revert this in the StableHLO dialect and see what happens. All tests in this repository are passing, but maybe we'll learn more after downstream integrations.

### DIFF
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -2546,7 +2546,6 @@ StablehloDialect::StablehloDialect(MLIRContext* context)
 #define GET_ATTRDEF_LIST
 #include "stablehlo/dialect/StablehloAttrs.cpp.inc"
       >();
-  context->loadDialect<tensor::TensorDialect>();
 }
 
 Type StablehloDialect::parseType(DialectAsmParser& parser) const {


GleasonK (changes requested):
LGTM

Thoughts on removing TensorDialect from `registerAllDialects` as well? Seems we have an equal reliance (maybe even less of a reliance) on TensorDialect as we do quant and sparse, and we removed those from `registerAllDialects`.

GleasonK:
Edit, I misspoke, its `registry.insert<mlir::sparse_tensor::SparseTensorDialect>();` in register all. Any idea if that should be there?

GleasonK (approved):

burmako:
That's a good question! Opened #1089 to follow up on that.

GleasonK:
This change caused failures in the following testpoint in TF: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/tests/xla_call_module_test.py#L211

We likely need to load this dialect _somewhere_ in the xla pipeline.

burmako:
Oof I think this shows that removing the Tensor dialect from the StableHLO dialect constructor was the right thing to do. Looks like this test started depending on this unintended behavior, instead of explicitly declaring the dependency on the Tensor dialect in its context initialization.

PR #1090, opened 2023-02-08
Modify build file so StableHLO can be included in other projects

GleasonK:
This should allow us to use one build file for mlir-hlo, tensorflow, and stablehlo repositories.

Boils down to: tablegen targets need `include = ["."]` and most other targets need `skip_include_prefix = "."`, inspired by https://github.com/tensorflow/mlir-hlo/blob/master/BUILD

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -37,6 +37,7 @@ cc_library(
     hdrs = [
         "stablehlo/dialect/Base.h",
     ],
+    strip_include_prefix = ".",
     deps = [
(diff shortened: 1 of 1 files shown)


burmako (approved):
Thank you for figuring this out! I can't say that I have sufficient Bazel knowledge to comment on this, but I like that this allows us to use the same build across multiple projects. If down the line we learn of a better way of approaching this, we'll be able to refactor as needed while keeping this desirable property.

PR #1092, opened 2023-02-09
Add interpreter for RemOp

zhouxin913:
(I1) lhs tensor of integer, floating-point or complex type (covered by ODS)
(I2) rhs tensor of integer, floating-point or complex type (covered by ODS)
(C1) lhs, rhs and result have the same type. (covered by ODS)

Note that all constraints are covered by ODS, so no tests in ops_stablehlo.mlir.

It is expected that `hlo_evaluator` and `mlir_interpreter` handle corner case differently ([hlo_evaluator](https://github.com/tensorflow/tensorflow/blob/c4c3c066d48983f57afd9e8fbe3734b03ee98018/tensorflow/compiler/xla/hlo/evaluator/hlo_evaluator_typed_visitor.h#L686) vs [mlir_interpreter](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/mlir_hlo/tools/mlir_interpreter/dialects/cwise_math.h#L83)). Here is our way (same as comment in code):

// These cases are implementation-defined, and the behavior of this interpreter:
// For integers: if rhs is 0, the result is the same as `lhs`.
//   For signed integers, if lhs is INT_MAX and rhs is -1, the result is 0.
// For floats: use APFloat.mod(). If rhs is 0, the result is 0x7FF8000000000000

There are tests for corner cases and NOT included in test file (on purpose).
```
// Results from hlo interpreter

// CHECK-LABEL: Evaluated results of function: remainder_op_test_si64
func.func @remainder_op_test_si64() -> tensor<2xi64> {
  %lhs = stablehlo.constant dense<[17, 0x7FFFFFFFFFFFFFFF]> : tensor<2xi64>
  %rhs = stablehlo.constant dense<[0, -1]> : tensor<2xi64>
  %result = stablehlo.remainder %lhs, %rhs : tensor<2xi64>
  func.return %result : tensor<2xi64>
}

  17 : i64
  0 : i64

// Results from mlir interpreter

// CHECK-LABEL: Evaluated results of function: remainder_op_test_f64
func.func @remainder_op_test_f64() -> tensor<1xf64> {
  %lhs = stablehlo.constant dense<[1.0]> : tensor<1xf64>
  %rhs = stablehlo.constant dense<[0.0]> : tensor<1xf64>
  %result = stablehlo.remainder %lhs, %rhs : tensor<1xf64>
  func.return %result : tensor<1xf64>
}

0x7FF8000000000000 : f64
```

### DIFF
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -827,7 +827,7 @@ def StableHLO_PowOp : StableHLO_BinaryElementwiseOp<"power",
 
 def StableHLO_RemOp : StableHLO_BinaryElementwiseOp<"remainder",
       [Pure, HLO_CompatibleOperandsAndResultType /*remainder_c1*/],
-       HLO_IntFpOrComplexTensor /*remainder_i1*/> {
+       HLO_IntFpOrComplexTensor /*remainder_i1, remainder_i2*/> {
   let summary = "Remainder operation";
   let description = [{
     Performs element-wise remainder of dividend `lhs` and divisor `rhs` tensors
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -826,8 +826,9 @@ def StableHLO_PowOp : StableHLO_BinaryElementwiseOp<"power",
 }
 
 def StableHLO_RemOp : StableHLO_BinaryElementwiseOp<"remainder",
-      [Pure, HLO_CompatibleOperandsAndResultType], HLO_IntFpOrComplexTensor> {
-  let summary = "Remainder operator";
+      [Pure, HLO_CompatibleOperandsAndResultType /* remainder_c1 */],
+       HLO_IntFpOrComplexTensor> {
+  let summary = "Remainder operation";
   let description = [{
     Performs element-wise remainder of dividend `lhs` and divisor `rhs` tensors
     and produces a `result` tensor.
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -826,8 +826,8 @@ def StableHLO_PowOp : StableHLO_BinaryElementwiseOp<"power",
 }
 
 def StableHLO_RemOp : StableHLO_BinaryElementwiseOp<"remainder",
-      [Pure, HLO_CompatibleOperandsAndResultType /* remainder_c1 */],
-       HLO_IntFpOrComplexTensor> {
+      [Pure, HLO_CompatibleOperandsAndResultType /*remainder_c1*/],
+       HLO_IntFpOrComplexTensor /*remainder_i1*/> {
   let summary = "Remainder operation";
   let description = [{
     Performs element-wise remainder of dividend `lhs` and divisor `rhs` tensors
--- a/stablehlo/reference/Element.cpp
+++ b/stablehlo/reference/Element.cpp
@@ -533,7 +533,7 @@ Element rem(const Element &e1, const Element &e2) {
                                                           : lhs.urem(rhs);
       },
       [](bool lhs, bool rhs) -> bool {
-        llvm::report_fatal_error("bool \% bool is unsupported");
+        llvm::report_fatal_error("reminder of two bool numbers is unsupported");
       },
       [](APFloat lhs, APFloat rhs) {
         lhs.mod(rhs);
@@ -542,7 +542,8 @@ Element rem(const Element &e1, const Element &e2) {
       [](std::complex<APFloat> lhs,
          std::complex<APFloat> rhs) -> std::complex<APFloat> {
         // TODO(#997): remove support for complex
-        llvm::report_fatal_error("complex \% complex is unsupported");
+        llvm::report_fatal_error(
+            "reminder of two complex numbers is unsupported");
       });
 }
 
--- a/stablehlo/reference/Element.cpp
+++ b/stablehlo/reference/Element.cpp
@@ -520,10 +520,6 @@ Element real(const Element &el) {
                                      debugString(el.getType()).c_str()));
 }
 
-// These cases are implementation-defined, and the behavior of this interpreter:
-// For integers: if rhs is 0, the result is the same as `lhs`.
-//   For signed integers, if lhs is INT_MAX and rhs is -1, the result is 0.
-// For floats: use APFloat.mod(). If rhs is 0, the result is 0x7FF8000000000000
 Element rem(const Element &e1, const Element &e2) {
   return map(
       e1, e2,
@@ -535,9 +531,9 @@ Element rem(const Element &e1, const Element &e2) {
         llvm::report_fatal_error("rem(bool, bool) is unsupported");
       },
       [](APFloat lhs, APFloat rhs) {
-        // std::fmod VS std:remainder: the returned value of the latter is not 
-        // guaranteed to have the same sign as lhs. So mod() is preferred here.
-        // The returned "APFloat::opStatus" is ignored.
+        // APFloat::fmod VS APFloat:remainder: the returned value of the latter
+        // is not guaranteed to have the same sign as lhs. So mod() is preferred
+        // here. The returned "APFloat::opStatus" is ignored.
         (void)lhs.mod(rhs);
         return lhs;
       },
--- a/stablehlo/reference/Element.cpp
+++ b/stablehlo/reference/Element.cpp
@@ -528,12 +528,11 @@ Element rem(const Element &e1, const Element &e2) {
   return map(
       e1, e2,
       [&](APInt lhs, APInt rhs) {
-        if (rhs == 0) return lhs;
         return isSupportedSignedIntegerType(e1.getType()) ? lhs.srem(rhs)
                                                           : lhs.urem(rhs);
       },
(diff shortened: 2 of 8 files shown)


stablehlo/dialect/StablehloOps.td
```diff
@@ -819,8 +819,9 @@ def StableHLO_PowOp : StableHLO_BinaryElementwiseOp<"power",
 }
 
 def StableHLO_RemOp : StableHLO_BinaryElementwiseOp<"remainder",
-      [Pure, HLO_CompatibleOperandsAndResultType], HLO_IntFpOrComplexTensor> {
-  let summary = "Remainder operator";
+      [Pure, HLO_CompatibleOperandsAndResultType /* remainder_c1 */],
+       HLO_IntFpOrComplexTensor> {
+  let summary = "Remainder operation";
```
sdasgup3, line 824: Add input constraints as per  the checklist.

stablehlo/reference/Element.cpp
```diff
@@ -350,6 +350,33 @@ Element min(const Element &e1, const Element &e2) {
       });
 }
 
+// These cases are implementation-defined, and the behavior of this interpreter:
+// For integers: if rhs is 0, the result is the same as `lhs`.
+//   For signed integers, if lhs is INT_MAX and rhs is -1, the result is 0.
+// For floats: use APFloat.mod(). If rhs is 0, the result is 0x7FF8000000000000
```
sdasgup3, line 356: Not sure if we should add comment about what behavior we choose for domain errors, given that we are not doing this religiously for other ops like `add`. 
 
https://github.com/llvm/llvm-project/blob/b309bc04eebc9c736b6c34d73d520a6ef7baf302/llvm/lib/Support/APFloat.cpp#L2167 seems to be equivanet to std::fmod

and `std::fmod` already calls out that the domain errors are implementation defined. 

IMO, any behavior is fine for us to have. Having said that maybe we can remove the special case for "divide by zero". As we know that we added it for parity with HLO, but having parity with HLO on cases which are free to vary across  implementations, does not seem worth following unless we have a definitive use-case in mind (a naive  use case could be minimizing false positives while  cross validating results with hlo interpreter)  

We can seek opinion from @burmako on this.

sdasgup3 (approved):

stablehlo/reference/Element.cpp
zhouxin913: Add @burmako as reviewer for more comments on this.

stablehlo/reference/Element.cpp
```diff
@@ -350,6 +350,33 @@ Element min(const Element &e1, const Element &e2) {
...
+Element rem(const Element &e1, const Element &e2) {
+  return map(
+      e1, e2,
+      [&](APInt lhs, APInt rhs) {
+        if (rhs == 0) return lhs;
+        return isSupportedSignedIntegerType(e1.getType()) ? lhs.srem(rhs)
+                                                          : lhs.urem(rhs);
+      },
+      [](bool lhs, bool rhs) -> bool {
+        llvm::report_fatal_error("reminder of two bool numbers is unsupported");
```
burmako, line 366: Let's use the wording similar to other functions above: "rem(bool, bool) is unsupported"?

stablehlo/reference/Element.cpp
burmako: Let's drop the comment. G5 in https://github.com/openxla/stablehlo/blob/main/docs/reference.md#testing-guidelines explicitly discourages writing tests for implementation-defined behavior, so I think we should extrapolate this to documenting implementation-defined behavior.

stablehlo/reference/Element.cpp
```diff
@@ -350,6 +350,33 @@ Element min(const Element &e1, const Element &e2) {
...
 
+// These cases are implementation-defined, and the behavior of this interpreter:
+// For integers: if rhs is 0, the result is the same as `lhs`.
+//   For signed integers, if lhs is INT_MAX and rhs is -1, the result is 0.
+// For floats: use APFloat.mod(). If rhs is 0, the result is 0x7FF8000000000000
+Element rem(const Element &e1, const Element &e2) {
+  return map(
+      e1, e2,
+      [&](APInt lhs, APInt rhs) {
+        if (rhs == 0) return lhs;
```
burmako, line 361: Why do we need this check at all?

stablehlo/reference/Element.cpp
```diff
@@ -350,6 +350,33 @@ Element min(const Element &e1, const Element &e2) {
...
+        return isSupportedSignedIntegerType(e1.getType()) ? lhs.srem(rhs)
+                                                          : lhs.urem(rhs);
+      },
+      [](bool lhs, bool rhs) -> bool {
+        llvm::report_fatal_error("reminder of two bool numbers is unsupported");
+      },
+      [](APFloat lhs, APFloat rhs) {
+        lhs.mod(rhs);
```
burmako, line 369: Can we directly return the result, instead of having the return appear on the subsequent line?

burmako, line 369: Why did we choose `APFloat::mod` over `APFloat::remainder`?

burmako (changes requested):

stablehlo/reference/Element.cpp
zhouxin913: Agreed with you guys, removed.

stablehlo/reference/Element.cpp
zhouxin913: Without this, there will be an crash with error message from
```
Evaluated results of function: remainder_op_test_si64
stablehlo-interpreter: .../stablehlo/llvm-project/llvm/lib/Support/APInt.cpp:1667:
llvm::APInt llvm::APInt::urem(const llvm::APInt&) const: 
Assertion `RHS.U.VAL != 0 && "Remainder by zero?"' failed.
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
Stack dump:
...
```

I thought this was unacceptable. But after rethink, this seems good enough.
Removed that line of code. Please "resolve this comment" if it looks good.

stablehlo/reference/Element.cpp
zhouxin913: Note that: APFloat::remainder(17.0, 3.0) = -1.000000e+00

Let's see the code: 
APFloat::mod is using [IEEEFloat::mod](https://llvm.org/doxygen/classllvm_1_1detail_1_1IEEEFloat.html#af98a9f1655f7320272c11d555e7e6846) while APFloat::remainder is using [IEEEFloat::remainder](https://llvm.org/doxygen/classllvm_1_1detail_1_1IEEEFloat.html#a878514b4bc5399e8ba0b104aa81d2fdf)

Our spec is: *For floating-point element types, this operation is in contrast with the remainder operation from IEEE-754 specification where d is an integral value nearest to the exact value of lhs/rhs with ties to even.*

I think the spec does not have a preference? (Please correct me if wrong). So considering the consistence that rem(17, 3) is 2, we should have rem(17.0, 3.0) as 2.0 instead of -1.0. 

Btw, HLO is using C mod, same as `APFloat::mod` above. https://github.com/tensorflow/tensorflow/blob/c4c3c066d48983f57afd9e8fbe3734b03ee98018/tensorflow/compiler/xla/hlo/evaluator/hlo_evaluator_typed_visitor.h#L700

What do you think of this?

zhouxin913: APFloat.mod() return status instead of a value, so current code can catch the error or return lhs. 
Please advise if there's a better way to do this?

burmako: Oh I missed that. Let's leave a comment that we're deliberately ignoring this return value, and change this line to say `(void)lhs.mod(rhs)` to emphasize this.

burmako: Thank you for explaining! Can you call out in a comment that using mod and not remainder is deliberate?

burmako (changes requested):

stablehlo/reference/Element.cpp
zhouxin913: Sure! Added.

zhouxin913: Yes.

stablehlo/reference/Element.cpp
```diff
@@ -350,6 +350,30 @@ Element min(const Element &e1, const Element &e2) {
...
+      e1, e2,
+      [&](APInt lhs, APInt rhs) {
+        return isSupportedSignedIntegerType(e1.getType()) ? lhs.srem(rhs)
+                                                          : lhs.urem(rhs);
+      },
+      [](bool lhs, bool rhs) -> bool {
+        llvm::report_fatal_error("rem(bool, bool) is unsupported");
+      },
+      [](APFloat lhs, APFloat rhs) {
+        // std::fmod VS std:remainder: the returned value of the latter is not
```
burmako, line 364: "APFloat::mod vs APFloat::remainder"?

burmako (approved):
One minor comment, and we're good to go.

stablehlo/reference/Element.cpp
zhouxin913: I meant the latter, but seems APFloat is more clear here. Changed.

PR #1093, opened 2023-02-09
Implement Region evaluation

### LINKED ISSUE #995: Augment interpreter framework to `eval` regions
With the current interpreter support for stableHLO ops, having no regions, the management of runtime stack frame was simple. For example, with all the ops scoped inside a function body, a single runtime stack  [here](https://github.com/openxla/stablehlo/blob/2636119c7c61b1e1814cdb7e30a38d341c05e1e8/stablehlo/reference/Interpreter.cpp#L50)  was enough. 

With general mlir ops, the associated [scoping rules](https://mlir.llvm.org/docs/LangRef/#value-scoping) are as follows:

1. A value defined in a region can be used by an operation which has a parent in the same region, if and only if the parent could use the value. 
2. A value defined by an argument to a region can always be used by any operation deeply contained in the region. 
3. A value defined in a region can never be used outside of the region.

The runtime stack used in the interpreter framework can accommodate the above rules using the following changes (in pseudo code):

1. Define scope
```c++
struct Scope {
  llvm::DenseMap<Value, Tensor> stackFrame;
  Scope *parentScope;
   
  Tensor findInScope(value value) {
       auto it = stackFrame.find(value);
      if (it != stackFrame.end()) return it->second;
      assert(parentScope && "value not found");
      return parentScope->findInScope(value);
  }

 void addToScope(Operation &op, ArrayRef<Tensor> runtimeValues)
      assert(op.getNumResults() == runtimeValues.size());
      for (auto [ssaResult, runtimeResult] : llvm::zip(op.getResults(), runtimeValues)) {
        stackFrame[ssaResult] = runtimeResult;
      }
};
```
2.  Evaluation of a region can start with adding a new scope with `parentScope` set to the scope of the parent of the region. A function body, being an example of a region with additional semantic restriction `IsolatedFromAbove`, we can add the scope for the function body with `parentScope` set to null.
3. The SSA value(s) defined, as result(s) of an `op` within a region, and their associated runtime value(s), are tied together in the runtime stack frame using `scope.addToScope(op, {runtimeValues})`, where `scope` corresponds to the one created during the evaluation of the immediate enclosing region.
4. The runtime `Tensor` value corresponding to an mlir `value` can be queried using `scope.findInScope(val)`, where `scope` 
corresponds to the one created during the evaluation of the immediate enclosing region.
5. The formal arguments of a region along with the associated runtime values `runtimeValues` are tied together in the runtime stack frame using the following, where `scope` is associated with the region where the arguments are defined.
```
auto& block = region.getBlocks().front();
for (auto [ssaArg, runtimeValue] : llvm::zip(block.getArguments(), runtimeValues)) {
   stackFrame[ssaArg] = runtimeValue;
} 
```
6. Once the evaluation of a region ends, make sure the associated scope is deleted.  

A few side comments:
1. No special treatment for non-loop control flow structures, like If/Case, where we do not use any arguments and instead rely on implicit capture. 
2. The evaluation of a region can be realized using
```
llvm::Expected<SmallVector<Tensor>> eval(Region& region, ArrayRef<Tensor> args, Scope* parentScope) {
   // Create a new scope    
   ...
}
```
3. Include tests covering the scoping rules using nesting of regions [e.g. nested_if](https://github.com/openxla/stablehlo/blob/2636119c7c61b1e1814cdb7e30a38d341c05e1e8/stablehlo/tests/ops_stablehlo.mlir#L997) to test the implementation.
4. Given that none of the StableHLO ops with regions are `IsolatedFromAbove` the above management of runtime stack should work.

\cc @ghpvnist @burmako

sdasgup3:
Closes #995 

The implementation follows the discussion in #995 in the following ways:

## Class `InterpreterScope`

To represent runtime stack corresponding to a region of a program under evaluation. The idea is to use a "parent-child" hierarchy of the class objects to implement the scoping rules.
 1. Name : `InterpreterScope`. Refrained fro using just `Scope` because that seems more like a static program concept.
 2. Members variables:
     -   `stackFrame`: mapping from SSA values, defined in the current region, to their evaluated runtime `Tensor` values.
     -   `parentScope`: a handle to `InterpreterScope` object corresponding to the (syntactically) enclosing region.
 3. Member functions:
     -  `void add(Value ssaValue, Tensor runtimeValue)`: Add the mapping SSA value (`ssaValue`), defined in region, to its runtime value (`runtimeValue`) in the current scope.
    -  `Tensor find(Value ssaValue) const;`: Find the runtime value mapped to SSA value `ssaValue`. The search starts
       with the current scope and then recursively continues over to the scope defined by `parentScope`.
 
A few points to note:
1. A `InterpreterScope` object is instantiated every time a region is evaluated. This is just a simplification over the following alternative
    - Store a map from a ` region` to corresponding `InterpreterScope` object.
    - Whenever a region is evaluated used the mapped `InterpreterScope` object for scope resolution. 

2. A `InterpreterScope` object treats the `parentScope` as immutable to align with the fact that a StableHLO program, in pure SSA form (without memory allocation/load/store ops), disallows mutating the `parentScope` object from within a region.

3. Copy prohibited for `InterpreterScope` objects.

## Region evaluation

The evaluation of a region is realized using

```
/// Evaluates an mlir::Region `region` using the runtime values `runtimeArgs`
/// corresponding to the arguments of the containing block, assuming that the
/// region has only one block. Interprets the operations within the block and
/// returns the runtime values for the terminator's arguments.
llvm::Expected<llvm::SmallVector<Tensor>> eval(Region &region, llvm::ArrayRef<Tensor> runtimeArgs,
  const InterpreterScope *const parentScope);
```

The above function is used to evaluated the region for any op including the `func::FuncOp`. Its definition provides a sole place when a `InterpreterScope` is stack-allocated and its `parentScope` is initialed. 

### How the `eval(Region..` is invoked

The evaluator of  `func::FuncOp` invokes the function as
```
eval(func.getBody(), args, nullptr); // This is because IsIsolatedFromAbove trait for func::FucnOp is true
``` 

The evaluator for other StableHLO ops, say `stableHLO::whileOp`, invokes the function as 
```
bool isIsolatedFromAbove = cond.getParentOp()->hasTrait<OpTrait::IsIsolatedFromAbove>();
eval(cond, runtimeInputs, isIsolatedFromAbove ? nullptr : &scope);
```
Note: 
  -  `scope` is the `InterpreterScope` object corresponding to the region containing the `whileOp`
  -  As the evaluator for an op expected to know the semantics, one might get tempted to skip the check for [OpTrait::IsolatedFromAbove](https://mlir.llvm.org/docs/Traits/#isolatedfromabove) and directly pass the appropriate scope. Please note that these checks make the code adjust to any future changes in the behavior of the op w.r.t its visibility to outer scope.
  -  The way the `parentScope` is initialized makes sure the following relationship to hold between a  `InterpreterScope` object and its stored `parentScope`
      - `InterpreterScope Scope`: Defined whenever a `region` is evaluated
      -  `Scope.parentScope`: `InterpreterScope` object corresponding to the syntactically enclosing region  

## [Scoping rules](https://mlir.llvm.org/docs/LangRef/#value-scoping) to cover

1. **A value defined by an argument to a region can always be used by any operation deeply contained in the region:** Covered by the design of `InterpreterScope`.
2.  **A value defined in a region can never be used outside of the region**:  For a value defined in a region `RI` to be used outside of the region `RO`, then `RO`, during its evaluation, should receive the `parentScope` as the `InterpreterScope` object corresponding to `RI` , which will never happen per how `parentScope` is always set to  `InterpreterScope` object  corresponding to the syntactically enclosing region. For example,
 
For the evaluation of  `R1` and `R2`, `parentScope` is set to `InterpreterScope` object  corresponding to the `R`.
```
  { // enclosing Region, R
    { // region R1
      %value_defined_in_a_region = ...
    }, { // region R0
      %operation_wants_to_use_value_defined_in_a_region = xyz(%value_defined_in_a_region
    }
   // 
```

3. **A value defined in a region can be used by an operation which has a parent in the same region, if and only if the parent could use the value**:

Here is the example to demonstrate the above rule
```
   { // Region R
       %value_defined_in_region = ...

       // CASE: I
       %parent_defined_in_same_region_as_value_and_can_use_value = ...({ '' Region R1
            %operation_uses_value = ...(%value_defined_in_region, ...
       }) 

      // CASE : II
      %parent_defined_in_same_region_as_value_and_cannot_use_value_as_isolated_from_above = ...({ // Region R2
            %operation_uses_value = ...(%value_defined_in_region, ...
       })
   }
 
```

Case I: For evaluation `R1`, the  `InterpreterScope` object  corresponding to the `R` is provided as `parentScope`. Hence the value `value_defined_in_region` will be visible.

Case II:  For evaluation `R2`, the `nullptr` provided as `parentScope` depending on the op trait [OpTrait::IsolatedFromAbove](https://mlir.llvm.org/docs/Traits/#isolatedfromabove). Hence the value `value_defined_in_region` will not be visible.

## Testing 

As there is no region op currently implemented to test the PR, we have implemented a simplified version of `whileOp`. The simplification is that if the runtime value corresponding to `cond` block is `true`, then we iterate the loop for fixed (2) times.The simplification has to be made because there is no `compreOp` implemented at time time. 

We used the `whileOp` to implement a program, `interpret_while.mlir`, which captures values from various allowable scope. The illegal reference to scopes will be flagged by the versifier only. So the interpreter can assume the input program to always obey the scoping rules and thereby avoid checking for illegal cases. 

update: Changed name `InterpretScope` to `Scope` [Lexical_scope_vs._dynamic_scope](https://en.wikipedia.org/wiki/Scope_(computer_science)#Lexical_scope_vs._dynamic_scope)


stablehlo/reference/InterpreterScope.h
```diff
@@ -0,0 +1,64 @@
...
+class InterpreterScope {
+ public:
+  InterpreterScope(const InterpreterScope *const pScope)
+      : parentScope(pScope) {}
+
+  InterpreterScope(const InterpreterScope &Scope) = delete;
+
+  /// Add the mapping from SSA value (`ssaValue`), defined in a region, to
+  /// its evaluated runtime value (`runtimeValue`).
+  void add(Value ssaValue, Tensor runtimeValue);
```
burmako, line 46: Now that we have a dedicated class for this, let's add another overload for `add` which takes `ArrayRef<Value>` and `ArrayRef<Tensor>`.

stablehlo/reference/InterpreterScope.h
```diff
@@ -0,0 +1,64 @@
...
+
+  /// Find the runtime value mapped to SSA value `ssaValue`. The search starts
+  /// with the current scope and then recursively continues over to the scope
+  /// defined by `parentScope`.
+  Tensor find(Value ssaValue) const;
```
burmako, line 51: Similarly, it looks like another overload which takes `ArrayRef<Value>` and produces `SmallVector<Tensor>` would be useful.

stablehlo/reference/InterpreterScope.h
```diff
@@ -0,0 +1,64 @@
...
+/// Note:
+/// 1. A `InterpreterScope` object is instantiated every time a region is
+///    evaluated.
+/// 2. A `InterpreterScope` object treats the `parentScope` as immutable to
+///    align with the fact that a StableHLO program, in pure SSA form (without
+///    memory allocation/load/store ops), disallows mutating the `parentScope`
+///    object from within a region.
+class InterpreterScope {
+ public:
+  InterpreterScope(const InterpreterScope *const pScope)
```
burmako, line 39: Let's call this parameter `parent`? Both [the LLVM style guide](https://llvm.org/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly) and [the Google style guide](https://google.github.io/styleguide/cppguide.html) discourage the use of abbreviations unless they are well known. I would say that abbreviating "parent" to "p" isn't particularly well known. Same for the constructor below.

stablehlo/reference/InterpreterScope.h
```diff
@@ -0,0 +1,64 @@
...
+      : parentScope(pScope) {}
+
+  InterpreterScope(const InterpreterScope &Scope) = delete;
```
burmako, line 42: Is there a reason to not make this class copyable?

burmako, line 42: Also, let's follow [the Google style guide](https://google.github.io/styleguide/cppguide.html#Copyable_Movable_Types) around copyable/movable types. Let's decide which type the InterpreterScope is, and implement the needed boilerplate correspondingly.

stablehlo/reference/InterpreterScope.h
```diff
@@ -0,0 +1,64 @@
...
+  /// with the current scope and then recursively continues over to the scope
+  /// defined by `parentScope`.
+  Tensor find(Value ssaValue) const;
+
+ private:
+  /// Internal store for mapping from SSA values to runtime `Tensor` values.
+  llvm::DenseMap<Value, Tensor> stackFrame;
+
+  /// A handle to the parent's scope.
+  const InterpreterScope *const parentScope = nullptr;
```
burmako, line 58: Just `parent`? I think it's unambiguous that a scope's parent is also a scope.

stablehlo/reference/Interpreter.cpp
```diff
@@ -35,142 +36,174 @@ llvm::Expected<SmallVector<Tensor>> eval(func::FuncOp func,
...
-           llvm::zip(op.getResults(), runtimeValues))
-        stackFrame[ssaResult] = runtimeResult;
+           llvm::zip(op.getResults(), runtimeResults)) {
+        if (ssaResult.getType() != runtimeResult.getType()) {
+          llvm::report_fatal_error(
+              "Expected same value for an SSA register and its evalted value");
+        }
+        scope.add(ssaResult, runtimeResult);
+      }
     };
```
burmako, line 87: I think we should delete all the lambdas above and then introduce and use `InterpreterScope` methods instead (see comments to `InterpreterScope.h`).

stablehlo/reference/Interpreter.cpp
```diff
@@ -35,142 +36,174 @@ llvm::Expected<SmallVector<Tensor>> eval(func::FuncOp func,
...
+      addOpResultsToScope({runtimeResult});
     } else if (auto returnOp = dyn_cast<func::ReturnOp>(op)) {
       SmallVector<Tensor> runtimeOperands;
       for (Value ssaOperand : returnOp.getOperands())
         runtimeOperands.push_back(fetchOperand(ssaOperand));
       return runtimeOperands;
+    } else if (auto returnOp = dyn_cast<ReturnOp>(op)) {
+      SmallVector<Tensor> runtimeOperands;
+      for (Value ssaOperand : returnOp.getOperands())
+        runtimeOperands.push_back(fetchOperand(ssaOperand));
```
burmako, line 171: I think this should also become a helper method in `InterpreterScope`.

stablehlo/reference/InterpreterScope.cpp
```diff
@@ -0,0 +1,46 @@
...
+}
+
+Tensor InterpreterScope::find(Value ssaValue) const {
+  auto it = stackFrame.find(ssaValue);
+
+  if (it != stackFrame.end()) return it->second;
+
+  if (!parentScope)
+    llvm::report_fatal_error(
+        llvm::formatv("Expected the value {0} to be already evaluated",
```
burmako, line 39: How about making the error message more straightforward, e.g. `"value {0} not found in scope"`?

stablehlo/reference/Ops.cpp
```diff
@@ -224,6 +225,47 @@ Tensor evalTransposeOp(const Tensor &operand, ArrayRef<int64_t> permutation,
...
+// simplification is based on iterating the loop `dummy_limit` number of times
+// if the loop condition is `true`. The simplifiction is done as compare op is
+// not supported. To be corrected and improved as part of #967 and #992
+// respectively.
+// ***********************************************************************
+SmallVector<Tensor> evalWhileOp(ArrayRef<Tensor> runtimeInputs, Region &cond,
+                                Region &body, const InterpreterScope &scope) {
+  SmallVector<Tensor> runtimeResults(runtimeInputs);
+  bool isIsolatedFromAbove =
+      cond.getParentOp()->hasTrait<OpTrait::IsIsolatedFromAbove>();
```
burmako, line 239: Don't we know this statically?

Re: "Please note that these checks make the code adjust to any future changes in the behavior of the op w.r.t its visibility to outer scope". I would imagine that we'll have this covered when/if we actually introduce these changes?

stablehlo/reference/Ops.cpp
```diff
@@ -224,6 +225,47 @@ Tensor evalTransposeOp(const Tensor &operand, ArrayRef<int64_t> permutation,
...
 }
 
+// ***********************************************************************
+// This is an simplified implementation of while op semantics. The
+// simplification is based on iterating the loop `dummy_limit` number of times
+// if the loop condition is `true`. The simplifiction is done as compare op is
+// not supported. To be corrected and improved as part of #967 and #992
+// respectively.
+// ***********************************************************************
+SmallVector<Tensor> evalWhileOp(ArrayRef<Tensor> runtimeInputs, Region &cond,
```
burmako, line 235: Do we have to say "runtime" here and below? In nearby functions, we simply say stuff like "lhs" and "result".

stablehlo/reference/Ops.cpp
```diff
@@ -224,6 +225,47 @@ Tensor evalTransposeOp(const Tensor &operand, ArrayRef<int64_t> permutation,
...
+                                Region &body, const InterpreterScope &scope) {
+  SmallVector<Tensor> runtimeResults(runtimeInputs);
+  bool isIsolatedFromAbove =
+      cond.getParentOp()->hasTrait<OpTrait::IsIsolatedFromAbove>();
+
+  auto runtimeCondResultsOrErr =
+      eval(cond, runtimeInputs, isIsolatedFromAbove ? nullptr : &scope);
+  if (!runtimeCondResultsOrErr)
+    llvm::report_fatal_error("Error in while op evaluation");
```
burmako, line 244: Error reporting is pretty underdeveloped in the interpreter at the moment, but let's maybe make these messages more specific? E.g. here we could say "Failed to evaluate cond", immediately below we could say "Failed to evaluate body".

stablehlo/reference/Ops.cpp
```diff
@@ -224,6 +225,47 @@ Tensor evalTransposeOp(const Tensor &operand, ArrayRef<int64_t> permutation,
...
+// ***********************************************************************
+SmallVector<Tensor> evalWhileOp(ArrayRef<Tensor> runtimeInputs, Region &cond,
+                                Region &body, const InterpreterScope &scope) {
+  SmallVector<Tensor> runtimeResults(runtimeInputs);
+  bool isIsolatedFromAbove =
+      cond.getParentOp()->hasTrait<OpTrait::IsIsolatedFromAbove>();
+
+  auto runtimeCondResultsOrErr =
+      eval(cond, runtimeInputs, isIsolatedFromAbove ? nullptr : &scope);
+  if (!runtimeCondResultsOrErr)
```
burmako, line 243: `|| runtimeCondResultsOrErr->size() != 1`? (Same below).

stablehlo/reference/Ops.cpp
```diff
@@ -224,6 +225,47 @@ Tensor evalTransposeOp(const Tensor &operand, ArrayRef<int64_t> permutation,
   return result;
 }
 
+// ***********************************************************************
+// This is an simplified implementation of while op semantics. The
+// simplification is based on iterating the loop `dummy_limit` number of times
+// if the loop condition is `true`. The simplifiction is done as compare op is
+// not supported. To be corrected and improved as part of #967 and #992
+// respectively.
+// ***********************************************************************
```
burmako, line 234: I think we should just hardcode support for ConvertOp directly in this function. That would probably take a similar amount of effort to implementing the notion of dummy_limit and explaining why we need it :)

stablehlo/tests/interpret_while.mlir
```diff
@@ -0,0 +1,35 @@
...
+          stablehlo.return %v4, %arg7 : tensor<i64>, tensor<i64>
+      }) : (tensor<i64>, tensor<i64>) -> (tensor<i64>, tensor<i64>)
+
+      stablehlo.return %results#0, %arg3 : tensor<i64>, tensor<i64>
+  }) : (tensor<i64>, tensor<i64>) -> (tensor<i64>, tensor<i64>)
+
+  func.return %results0 : tensor<i64>
+}
+// CHECK-NEXT: tensor<i64>
+// CHECK-NEXT: 420 : i64
```
burmako, line 35: Woah that's a pretty intense test. Can we make it simpler without hurting coverage?

burmako (changes requested):

stablehlo/reference/Interpreter.cpp
```diff
@@ -35,142 +36,174 @@ llvm::Expected<SmallVector<Tensor>> eval(func::FuncOp func,
     return invalidArgument("Expected one region in func %s",
                            func.getName().str().c_str());
 
-  if (!func.getBody().hasOneBlock())
-    return invalidArgument("Expected one block in func %s",
-                           func.getName().str().c_str());
```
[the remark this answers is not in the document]
sdasgup3: Actually, I moved this check to `evalRegion` so that the same check applied to both func::funcOp and StableHLO ops.

stablehlo/reference/Interpreter.cpp
```diff
@@ -35,142 +36,174 @@ llvm::Expected<SmallVector<Tensor>> eval(func::FuncOp func,
...
-
   Block &block = func.front();
   if (block.getNumArguments() != args.size())
     return invalidArgument(
         "Expected same amount of func arguments in %s "
         "and runtime arguments (%d)",
         func.getName().str().c_str(), args.size());
```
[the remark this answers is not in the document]
sdasgup3: I kept this check separate from the similar check inside `evalRegion` mainly because 

1. The check in function is a input error (as the input `Tensors` are provided by the user), whereas the check in `evalRegion` is a fatal error as it signifies something is wrong which the implementation of the interpreter. 

2. The diagnostic message for this check is more informative than what we have in `evalRegion`, which is

```
  Block &block = region.front();
  if (block.getArguments().size() != runtimeArgs.size())
    report_fatal_error(invalidArgument(
        "Expected same amount of block arguments and runtime arguments (%d)",
        runtimeArgs.size()));
  ...
```

In order to implement the above differences, having two copies seems a less overhead solution to me.
 
Let me know if the above justifies keeping those checks are separate. Otherwise happy to de-duplicate.

stablehlo/reference/InterpreterScope.h
sdasgup3: I am calling this `parentScope` just to differentiate it with a member variable with the same name. Just for clarity.

stablehlo/reference/InterpreterScope.h
GleasonK: My worry would be expensive copies, of data members in `llvm::DenseMap<Value, Tensor> stackFrame;`

That said, it looks like Tensor uses a reference counted pointer, so massive tensor copies aren't an issue. Some programs do have a lot of ssa values, so maybe that map will get big and we'd want to avoid copying without a need?

stablehlo/reference/Ops.cpp
GleasonK: Apologies if this conversation has already happened - is there any reason to not sync these with (either ODS or spec) names -`operands` in this case?

stablehlo/reference/Ops.cpp
sdasgup3: makes sense.

stablehlo/tests/interpret_while.mlir
sdasgup3: PTAL.

stablehlo/reference/InterpreterScope.h
sdasgup3: Agreed.

sdasgup3: My take was as a `InterpreterScope` object is tied to a `Region` and it does not seem right to allow a copy of that.

stablehlo/reference/Interpreter.h
```diff
@@ -32,6 +33,14 @@ namespace stablehlo {
 llvm::Expected<llvm::SmallVector<Tensor>> eval(func::FuncOp func,
                                                llvm::ArrayRef<Tensor> args);
 
+/// Evaluates an mlir::Region `region` using the runtime values `runtimeArgs`
+/// corresponding to the arguments of the containing block, assuming that the
+/// region has only one block. Interprets the operations within the block and
+/// returns the runtime values for the terminator's arguments.
+llvm::Expected<llvm::SmallVector<Tensor>> eval(
+    Region &region, llvm::ArrayRef<Tensor> runtimeArgs,
+    const InterpreterScope *const parentScope);
```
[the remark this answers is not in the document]
sdasgup3: Eve though we are not allowing copy of `InterpreterScope` objects, we are passing the parent Scope to function `evalRegion` via pointer.
 

It would be ideal if `evalRegion` can take the parent pointer as `const InterpreterScope &parent`. But we could not use that because 
1. `evalRegion` is called either by `evalFunc` or `eval<Op>` and `evalFunc` needs to pass some value denoting "no scope". Since "optional reference" is not a thing, the choice I made is `nullptr`.
The `const InterpreterScope * const` signature is to mimic the affect of a const reference to disallow mutation of the parent pointer.

stablehlo/reference/Interpreter.h
```diff
@@ -32,6 +33,14 @@ namespace stablehlo {
 llvm::Expected<llvm::SmallVector<Tensor>> eval(func::FuncOp func,
                                                llvm::ArrayRef<Tensor> args);
```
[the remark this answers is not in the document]
sdasgup3: Yes, I think we should remove Interpreter.* files and the driver `stablehlo-intepreter` to directly call the evalRegion.

stablehlo/reference/Interpreter.cpp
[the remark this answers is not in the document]
sdasgup3: After moving `evalRegion` to `Ops.*`, chose to removed the   `Interpreter.* files`  (as part of comment https://github.com/openxla/stablehlo/pull/1093#discussion_r1103221122). With that the the above check and call to `evalRegion` are hoisted to the interpreter [driver](https://github.com/openxla/stablehlo/blob/main/stablehlo/tools/StablehloInterpreterMain.cpp#L32). As the `arg` passed on to `eval(func::FuncOp)`  were empty, the above check will be trivially  folded.

stablehlo/reference/InterpreterScope.h
```diff
@@ -0,0 +1,64 @@
...
+/// region.
+///
+/// Note:
+/// 1. A `InterpreterScope` object is instantiated every time a region is
+///    evaluated.
+/// 2. A `InterpreterScope` object treats the `parentScope` as immutable to
+///    align with the fact that a StableHLO program, in pure SSA form (without
+///    memory allocation/load/store ops), disallows mutating the `parentScope`
+///    object from within a region.
+class InterpreterScope {
```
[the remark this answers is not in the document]
sdasgup3: The name Scope seems to me as a lexical scope which no connection with the execution context. [Lexical_scope_vs._dynamic_scope](https://en.wikipedia.org/wiki/Scope_(computer_science)#Lexical_scope_vs._dynamic_scope) explains a wider view of it.

Renamed to Scope.

stablehlo/reference/Ops.cpp
sdasgup3: That is a good point. The idea was very much  to keep them in sync with ODS. But I admit that it is not documented anywhere.  Point noted.

stablehlo/reference/Ops.cpp
```diff
@@ -224,12 +236,187 @@ Tensor evalTransposeOp(const Tensor &operand, ArrayRef<int64_t> permutation,
...
+  if (block.getArguments().size() != args.size())
+    report_fatal_error(invalidArgument(
+        "Expected same amount of block arguments and runtime arguments (%d)",
+        args.size()));
+
+  Scope scope(parentScope);
+  for (auto [ssaArg, runtimeArg] : llvm::zip(block.getArguments(), args)) {
+    assert(ssaArg.getType() == runtimeArg.getType());
+    scope.add(ssaArg, runtimeArg);
+  }
```
burmako, line 287: `scope.add(block.getArguments(), args)`?

stablehlo/reference/Scope.cpp
```diff
@@ -0,0 +1,60 @@
...
+#include "mlir/Support/DebugStringHelper.h"
+
+namespace mlir {
+namespace stablehlo {
+
+void Scope::add(Value ssaValue, Tensor runtimeValue) {
+  // We are instantiating a new `Scope` object every time the
+  // interpreter evaluates a region. With that, the `stackFrame` should not have
+  // any duplicates.
+  assert(!stackFrame.count(ssaValue));  // TODO
```
burmako, line 28: What does this TODO mean?

stablehlo/reference/InterpreterScope.h
burmako: The current convention that we're using for this follows Google style: call parameters `foo` and corresponding fields `foo_`.

stablehlo/reference/InterpreterScope.h
burmako: Sure, sounds good.

burmako: Looks like we're choosing the `MoveOnly` strategy, in which case here's how the boilerplate should look like:

```C++
class MoveOnly {
 public:
  MoveOnly(MoveOnly&& other) = default;
  MoveOnly& operator=(MoveOnly&& other) = default;

  // The copy operations are implicitly deleted, but you can
  // spell that out explicitly if you want:
  MoveOnly(const MoveOnly&) = delete;
  MoveOnly& operator=(const MoveOnly&) = delete;
};
```

stablehlo/reference/Ops.cpp
burmako: Even if this is not documented, let's follow the style of the nearby code which doesn't use "runtime" prefixes in `evalFooOp` functions?

stablehlo/reference/Ops.h
```diff
@@ -46,8 +49,18 @@ Tensor evalSubtractOp(const Tensor &lhs, const Tensor &rhs, Type resultType);
 Tensor evalTanhOp(const Tensor &operand, Type resultType);
 Tensor evalTransposeOp(const Tensor &operand, ArrayRef<int64_t> permutation,
                        Type resultType);
+SmallVector<Tensor> evalWhileOp(ArrayRef<Tensor> inputs, Region &cond,
+                                Region &body, const Scope &scope);
```
burmako, line 53: Why does the scope have to be non-null here? Theoretically, nothing should stop us from calling this function with a nullptr scope if the condition and the body of the loop don't capture anything from above?

stablehlo/reference/Ops.cpp
```diff
@@ -224,12 +236,187 @@ Tensor evalTransposeOp(const Tensor &operand, ArrayRef<int64_t> permutation,
...
+
+  for (Operation &op : block) {
+    SmallVector<Value> results(op.getResults().begin(), op.getResults().end());
+    if (auto addOp = dyn_cast<AddOp>(op)) {
+      Tensor runtimeLhs = scope.find(addOp.getLhs());
+      Tensor runtimeRhs = scope.find(addOp.getRhs());
+      Tensor runtimeResult = evalAddOp(runtimeLhs, runtimeRhs, addOp.getType());
+
```
burmako, line 295: Extraneous newline?

stablehlo/reference/Ops.cpp
```diff
@@ -224,12 +236,187 @@ Tensor evalTransposeOp(const Tensor &operand, ArrayRef<int64_t> permutation,
...
+        args.size()));
+
+  Scope scope(parentScope);
+  for (auto [ssaArg, runtimeArg] : llvm::zip(block.getArguments(), args)) {
+    assert(ssaArg.getType() == runtimeArg.getType());
+    scope.add(ssaArg, runtimeArg);
+  }
+
+  for (Operation &op : block) {
+    SmallVector<Value> results(op.getResults().begin(), op.getResults().end());
```
burmako, line 290: Can we avoid the need to create a temporary vector? The `add` overload doesn't necessarily have to an `ArrayRef`. Something zippable would do as well.

stablehlo/tests/interpret_while.mlir
```diff
@@ -0,0 +1,27 @@
...
+func.func @region_test() -> tensor<i64> {
+  %iter = stablehlo.constant dense<2> : tensor<i64>
+  %step = stablehlo.constant dense<1> : tensor<i64>
+  %v0 = stablehlo.constant dense<10> : tensor<i64>
+
+  %results0, %results1 = "stablehlo.while"(%v0, %iter) ({
+    ^bb0(%arg0: tensor<i64>, %arg1: tensor<i64>):
+      %cond = stablehlo.convert %arg1 :  (tensor<i64>) -> tensor<i1>
+      "stablehlo.return"(%cond) : (tensor<i1>) -> ()
+
```
burmako, line 13: Extraneous newline?

stablehlo/tests/interpret_while.mlir
burmako: Can we have something even simpler, like:

```mlir
// RUN: stablehlo-interpreter --interpret -split-input-file %s | FileCheck %s

// CHECK-LABEL: Evaluated results of function: while
func.func @while() -> tensor<i64> {
  // int i = 10;
  // int state = 0;
  // while (i > 0) {
  //   i -= 1;
  //   state += 1;
  // }
  %ten = stablehlo.constant dense<10> : tensor<i64>
  %zero = stablehlo.constant dense<0> : tensor<i64>
  %one = stablehlo.constant dense<1> : tensor<i64>
  %result_i, %result_state = "stablehlo.while"(%ten, %zero) ({
    ^bb0(%i: tensor<i64>, %state: tensor<i64>):
      %cond = stablehlo.convert %i : (tensor<i64>) -> tensor<i1>
      stablehlo.return %cond : tensor<i1>
  }, {
    ^bb0(%i: tensor<i64>, %state: tensor<i64>):
      %new_i = stablehlo.subtract %i, %one :  tensor<i64>
      %new_state = stablehlo.add %state, %one :  tensor<i64>
      stablehlo.return %new_i, %new_state : tensor<i64>, tensor<i64>
  }) : (tensor<i64>, tensor<i64>) -> (tensor<i64>, tensor<i64>)

  func.return %result_state : tensor<i64>
}
// CHECK-NEXT: tensor<i64>
// CHECK-NEXT: 10 : i64
```

burmako (changes requested):

stablehlo/reference/Interpreter.h
```diff
@@ -32,6 +33,14 @@ namespace stablehlo {
...
 
+/// Evaluates an mlir::Region `region` using the runtime values `runtimeArgs`
+/// corresponding to the arguments of the containing block, assuming that the
+/// region has only one block. Interprets the operations within the block and
+/// returns the runtime values for the terminator's arguments.
+llvm::Expected<llvm::SmallVector<Tensor>> eval(
+    Region &region, llvm::ArrayRef<Tensor> runtimeArgs,
```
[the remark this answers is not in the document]
sdasgup3: Done.

stablehlo/reference/Ops.h
sdasgup3: >> Why does the scope have to be non-null here?


This `scope`, created at `eval(Region..` and  passed into the `evalWhileOp`,   refers to the scope where the while operation is defined and IMO it should never be null.

>>  Theoretically, nothing should stop us from calling this function with a nullptr scope if the condition and the body of the loop don't capture anything from above?

If the  `cond` and `body` regions do not capture any thing from above, that would be realized using the scope object for those regions to bear a parent scope = `nullptr`. 

 Sorry if I misunderstood this question. Please let me know.

stablehlo/reference/Ops.cpp
sdasgup3: done. Used `ValueRange` instead.

stablehlo/reference/Scope.cpp
sdasgup3: The Todo is related to  adding a  report_fatal_error error. Done.

stablehlo/reference/InterpreterScope.h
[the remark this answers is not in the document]
sdasgup3: >> What I meant is that if something is called "InterpreterScope", why not just simplify it to "Scope". The same arguments against "Scope" similarly apply to "InterpreterScope", right?
That is right. I agree that the choice I made was not indicative of my intention (mentioned below).

>> how is this class different from lexical scope? .. At any given point of time, the current Scope contains the fullness of the lexical scope for the operation that is being executed.
I agree we are using the lexical scope with name-binding to runtime values. I was assuming the lexical scope to provide name binding to static entities (like variables) only. But the binding can be anything.

stablehlo/reference/Ops.h
```diff
@@ -46,8 +49,17 @@ Tensor evalSubtractOp(const Tensor &lhs, const Tensor &rhs, Type resultType);
...
 Tensor evalTransposeOp(const Tensor &operand, ArrayRef<int64_t> permutation,
                        Type resultType);
+SmallVector<Tensor> evalWhileOp(ArrayRef<Tensor> operand, Region &cond,
+                                Region &body, Scope &scope);
 Tensor evalXorOp(const Tensor &lhs, const Tensor &rhs, Type resultType);
 
+/// Evaluates an mlir::Region `region` using the runtime values `runtimeArgs`
+/// corresponding to the arguments of the containing block, assuming that the
+/// region has only one block. Interprets the operations within the block and
+/// returns the runtime values for the terminator's arguments.
```
burmako, line 59: Proposed rewording:

```
/// Evaluates an mlir::Region `region` using the runtime values `args`
/// corresponding to the arguments of the entry block of the region.
/// Interprets the operations within the entry block and returns the runtime 
/// values for the terminator's arguments.
/// Assumes that the region has only one block.
```

Feedback:
  1) Rename "runtimeArgs" to "args".
  2) Use the more idiomatic "entry block" to refer to the region's first block.

stablehlo/reference/Scope.h
```diff
@@ -0,0 +1,66 @@
...
+namespace mlir {
+namespace stablehlo {
+
+/// Represents the scope corresponding to a region of a program under
+/// evaluation. Holds (1) mapping from SSA values, defined in the current
+/// region, to their evaluated runtime `Tensor` values, and (2) handle to
+/// `Scope` object corresponding to the syntactically enclosing region.
+class Scope {
+ public:
+  Scope(Scope *parentScope) : parent_(parentScope) {}
```
burmako, line 30: Now this can be called `parent`, right?

stablehlo/reference/Scope.h
```diff
@@ -0,0 +1,66 @@
...
+  /// Find the runtime values mapped to SSA values `ssaValues`.
+  // SmallVector<Tensor> find(ArrayRef<Value> ssaValues) const;
+  SmallVector<Tensor> find(ValueRange ssaValues) const;
+
+ private:
+  /// Internal store for mapping from SSA values to runtime `Tensor` values.
+  llvm::DenseMap<Value, Tensor> stack_frame_;
+
+  /// A handle to the parent's scope.
+  Scope *parent_ = nullptr;
```
burmako, line 60: Do we need to explicitly specify the default value here?

stablehlo/reference/Ops.cpp
```diff
@@ -224,12 +236,172 @@ Tensor evalTransposeOp(const Tensor &operand, ArrayRef<int64_t> permutation,
...
   Tensor result(resultType);
   for (auto it = lhs.index_begin(); it != lhs.index_end(); ++it)
     result.set(*it, lhs.get(*it) ^ rhs.get(*it));
   return result;
 }
 
+SmallVector<Tensor> eval(Region &region, ArrayRef<Tensor> args,
+                         Scope *parentScope) {
+  if (!region.hasOneBlock())
+    llvm::report_fatal_error("Expected single block region");
```
burmako, line 267: I'm realizing that we probably don't have to explicitly check this. If there are multiple blocks, then the execution path either doesn't encounter them (good) or the execution path will encounter a cf.br operation or similar and crash (also good).

stablehlo/reference/Ops.cpp
```diff
@@ -224,12 +236,172 @@ Tensor evalTransposeOp(const Tensor &operand, ArrayRef<int64_t> permutation,
...
+      Tensor runtimeRhs = scope.find(xorOp.getRhs());
+      Tensor runtimeResult = evalXorOp(runtimeLhs, runtimeRhs, xorOp.getType());
+      scope.add(op.getResults(), {runtimeResult});
+    } else {
+      report_fatal_error(
+          invalidArgument("Unsupported op: %s", debugString(op).c_str()));
+    }
+  }
+
+  llvm::report_fatal_error("Expected a terminator when evaluating func");
```
burmako, line 403: "when evaluating a region"?

stablehlo/reference/Ops.h
burmako: I agree that we will not encounter a non-null `scope` during normal execution of the `eval` function. From the perspective of that use case, the current signature is just fine.

However, I also noticed that the non-nullness of the `scope` is not used in `evalWhileOp`, and this got me thinking about alternative use cases. Imagine if someone wants to just evaluate a WhileOp, without worrying about the enclosing function. If the scope is nullable, they can easily do that, but if the scope is not nullable, they will have to create a dummy scope.

Personally, I don't have a strong preference either way - just wanted to elaborate on where I was coming from. Please feel free to resolve how you see fit.

burmako (approved):
LGTM except for a few minor comments above.

stablehlo/reference/Ops.cpp
sdasgup3: makes sense.

PR #1147, opened 2023-02-10
Allow -f fix for lint_markdown.sh

GleasonK:
This change allows for 2 arguments to be specified, and forward them to the `markdownlint-cli`. This allows us to use the `-f` flag to autofix markdown issues.

Modifying contribution from @scottamain, if there is any reason to avoid this flag let me know!

### DIFF
--- a/build_tools/github_actions/lint_markdown.sh
+++ b/build_tools/github_actions/lint_markdown.sh
@@ -15,8 +15,9 @@
 # If passing the files as a glob, be sure to wrap in quotes. For more info,
 # see https://github.com/igorshubovych/markdownlint-cli#globbing
 
-if [[ $# -ne 1 ]] ; then
-  echo "Usage: $0 <files|directories|globs>"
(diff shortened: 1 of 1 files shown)


build_tools/github_actions/lint_markdown.sh
```diff
@@ -15,8 +15,9 @@
 # If passing the files as a glob, be sure to wrap in quotes. For more info,
 # see https://github.com/igorshubovych/markdownlint-cli#globbing
 
-if [[ $# -ne 1 ]] ; then
-  echo "Usage: $0 <files|directories|globs>"
+if [[ $# -gt 2 ]] ; then
+  echo "Usage: $0 [-f] <files|directories|globs>"
+  echo "  -f  Autofix markdown issues."
```
GleasonK, line 20: @scottamain is there any reason to avoid using this feature?

burmako (approved):

PR #1149, opened 2023-02-10
RFC for aligning StableHLO and TOSA arithmetic

eric-k256:
Initial version of an RFC to discuss aligning StableHLO and TOSA arithmetic operation.

Signed-off-by: Eric Kunze <eric.kunze@arm.com>

### DIFF
--- a/rfcs/20230210-stablehlo-tosa.md
+++ b/rfcs/20230210-stablehlo-tosa.md
@@ -1,0 +1,180 @@
+# RFC: Align StableHLO and TOSA arithmetic
+
+| Status        | Proposed                                                |
+| :------------ | :------------------------------------------------------ |
+| **RFC #**     | |
+| **Author(s)** | Eric Kunze (eric.kunze@arm.com), Dominic Symes (dominic.symes@arm.com), Tom Cooksey (tom.cooksey@arm.com) |
+| **Updated**   |                                               |
+
+## Objective
+
+Align the operations of StableHLO and TOSA.
+
+## Proposal
+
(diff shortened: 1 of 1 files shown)


burmako (commented):
Hi folks! Apologies for the late reply - there have been out of band conversations about this RFC, and now I'd like to summarize them here on GitHub.

As the RFC mentions, StableHLO dialect currently has support for quantization via:
  1) Supporting `quant.uniform` element types.
  2) Having dedicated ops like `uniform_quantize` / `uniform_dequantize`.
  3) Allowing regular ops like `add` / `convolution` to take quantized tensors.

This support was inherited from MHLO when StableHLO was bootstrapped, and MHLO's support was motivated by mobile use cases and inherited from TFLite. TFLite quantization has [a specification](https://www.tensorflow.org/lite/performance/quantization_spec), but StableHLO quantization does not.

One key aspect of TFLite quantization spec is that it uses a floating-point scale and an integer zero point as quantization parameters. In comparison, the quantization parameters proposed in this RFC involve an integer multiplier and shift. Harmonizing this or coming up with some kind of a compromise solution looks like the main open question at the moment.

Towards that end, I would like to propose for us to collaborate on pull requests to StableHLO specification. @sdasgup3 has created an initial PR #1352 that drafts a specification for QuantizedType and for semantics of quantized add, modelled after TFLite quantization semantics. Let's get together as a community and discuss the details, with the plan to progress to more involved ops like `convolution` in the future pull requests.

GleasonK:
I believe this proposal is covered by the RescaleOp work we had completed, closing as such. If that is not the case let me know and we can reopen!

PR #1154, opened 2023-02-11
Create Element.h constructors to get Element with primitive value

ghpvnist:
Currently, there is not an easy way to initialize Element according to the element types and each op needs to implement their own logic to initialize them. The proposed API allows easily creating Element object by specifying the type and its primitive value.

### DIFF
--- a/stablehlo/reference/Element.cpp
+++ b/stablehlo/reference/Element.cpp
@@ -21,6 +21,9 @@ limitations under the License.
 #include "llvm/Support/Error.h"
 #include "mlir/Dialect/Complex/IR/Complex.h"
 #include "mlir/IR/BuiltinAttributes.h"
+#include "mlir/Support/DebugStringHelper.h"
+#include "stablehlo/reference/Errors.h"
+#include "stablehlo/reference/Types.h"
 
 namespace mlir {
 namespace stablehlo {
--- a/stablehlo/reference/Element.cpp
+++ b/stablehlo/reference/Element.cpp
@@ -121,12 +121,11 @@ Element mapWithUpcastToDouble(const Element &el, FloatFn floatFn,
 Element::Element(Type type, APInt value) : type_(type), value_(value) {}
 
 Element::Element(Type type, int64_t value) {
+  type_ = type;
   if (isSupportedSignedIntegerType(type)) {
-    *this = Element(
-        type, APInt(type.getIntOrFloatBitWidth(), value, /*isSigned=*/true));
+    value_ = APInt(type.getIntOrFloatBitWidth(), value, /*isSigned=*/true);
   } else if (isSupportedUnsignedIntegerType(type)) {
-    *this = Element(
-        type, APInt(type.getIntOrFloatBitWidth(), value, /*isSigned=*/false));
+    value_ = APInt(type.getIntOrFloatBitWidth(), value, /*isSigned=*/false);
   } else {
     report_fatal_error(invalidArgument("Unsupported element type: %s",
                                        debugString(type).c_str()));
@@ -138,12 +137,13 @@ Element::Element(Type type, bool value) : type_(type), value_(value) {}
 Element::Element(Type type, APFloat value) : type_(type), value_(value) {}
 
 Element::Element(Type type, double value) {
+  type_ = type;
   if (isSupportedFloatType(type)) {
     APFloat floatVal(static_cast<double>(value));
     bool roundingErr;
     floatVal.convert(type.cast<FloatType>().getFloatSemantics(),
                      APFloat::rmNearestTiesToEven, &roundingErr);
-    *this = Element(type, floatVal);
+    value_ = floatVal;
   } else {
     report_fatal_error(invalidArgument("Unsupported element type: %s",
                                        debugString(type).c_str()));
@@ -154,6 +154,7 @@ Element::Element(Type type, std::complex<APFloat> value)
     : type_(type), value_(std::make_pair(value.real(), value.imag())) {}
 
 Element::Element(Type type, std::complex<double> value) {
+  type_ = type;
   if (isSupportedComplexType(type)) {
     APFloat real(value.real());
     APFloat imag(value.imag());
@@ -163,7 +164,7 @@ Element::Element(Type type, std::complex<double> value) {
                  &roundingErr);
     imag.convert(floatTy.getFloatSemantics(), APFloat::rmNearestTiesToEven,
                  &roundingErr);
-    *this = Element(type, std::complex<APFloat>(real, imag));
+    value_ = std::make_pair(real, imag);
   } else {
     report_fatal_error(invalidArgument("Unsupported element type: %s",
                                        debugString(type).c_str()));
--- a/stablehlo/reference/Element.cpp
+++ b/stablehlo/reference/Element.cpp
@@ -118,6 +118,58 @@ Element mapWithUpcastToDouble(const Element &el, FloatFn floatFn,
 
 }  // namespace
 
+Element::Element(Type type, APInt value) : type_(type), value_(value) {}
+
+Element::Element(Type type, int64_t value) {
+  if (isSupportedSignedIntegerType(type)) {
+    *this = Element(
+        type, APInt(type.getIntOrFloatBitWidth(), value, /*isSigned=*/true));
+  } else if (isSupportedUnsignedIntegerType(type)) {
+    *this = Element(
+        type, APInt(type.getIntOrFloatBitWidth(), value, /*isSigned=*/false));
+  } else {
+    report_fatal_error(invalidArgument("Unsupported element type: %s",
+                                       debugString(type).c_str()));
+  }
+}
+
+Element::Element(Type type, bool value) : type_(type), value_(value) {}
+
+Element::Element(Type type, APFloat value) : type_(type), value_(value) {}
+
+Element::Element(Type type, double value) {
+  if (isSupportedFloatType(type)) {
+    APFloat floatVal(static_cast<double>(value));
+    bool roundingErr;
+    floatVal.convert(type.cast<FloatType>().getFloatSemantics(),
+                     APFloat::rmNearestTiesToEven, &roundingErr);
+    *this = Element(type, floatVal);
+  } else {
+    report_fatal_error(invalidArgument("Unsupported element type: %s",
+                                       debugString(type).c_str()));
+  }
+}
+
+Element::Element(Type type, std::complex<APFloat> value)
+    : type_(type), value_(std::make_pair(value.real(), value.imag())) {}
+
+Element::Element(Type type, std::complex<double> value) {
+  if (isSupportedComplexType(type)) {
+    APFloat real(value.real());
+    APFloat imag(value.imag());
+    auto floatTy = type.cast<ComplexType>().getElementType().cast<FloatType>();
+    bool roundingErr;
+    real.convert(floatTy.getFloatSemantics(), APFloat::rmNearestTiesToEven,
+                 &roundingErr);
+    imag.convert(floatTy.getFloatSemantics(), APFloat::rmNearestTiesToEven,
+                 &roundingErr);
+    *this = Element(type, std::complex<APFloat>(real, imag));
+  } else {
+    report_fatal_error(invalidArgument("Unsupported element type: %s",
+                                       debugString(type).c_str()));
+  }
+}
+
 APInt Element::getIntegerValue() const {
   if (!isSupportedIntegerType(type_))
     llvm::report_fatal_error("Element is not an integer");
--- a/stablehlo/reference/Element.cpp
+++ b/stablehlo/reference/Element.cpp
@@ -21,9 +21,6 @@ limitations under the License.
 #include "llvm/Support/Error.h"
 #include "mlir/Dialect/Complex/IR/Complex.h"
 #include "mlir/IR/BuiltinAttributes.h"
-#include "mlir/Support/DebugStringHelper.h"
-#include "stablehlo/reference/Errors.h"
-#include "stablehlo/reference/Types.h"
 
 namespace mlir {
 namespace stablehlo {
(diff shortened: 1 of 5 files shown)


stablehlo/reference/Element.h
```diff
@@ -41,6 +44,38 @@ class Element {
       : type_(type), value_(std::make_pair(value.real(), value.imag())) {}
 
   Element(const Element &other) = default;
+  Element() = default;
+
+  template <class V>
+  static Element getValue(Type type, V value) {
```
sdasgup3, line 50: Can we use `T` for the type instead? IMO seems more natural for type.

stablehlo/reference/Element.h
```diff
@@ -41,6 +44,38 @@ class Element {
       : type_(type), value_(std::make_pair(value.real(), value.imag())) {}
 
   Element(const Element &other) = default;
+  Element() = default;
+
```
sdasgup3, line 48: Can we add a  comment comment? In the comment, we need to highlight the behavior of when type == complex as it is non trivial. 

Also, it does not seem idiomatic to include this static method as part of constructor comment group. We can use "Value Generator"  along the line of https://github.com/llvm/llvm-project/blob/31c42f8f35887a78301bb5ac640f77727fdd7d97/llvm/include/llvm/ADT/APInt.h#L177.

stablehlo/reference/Element.h
ghpvnist: I was considering T as well, but it felt weird to have `(Type type, T value)` when we already have a `Type` class as a variable.

stablehlo/reference/Element.cpp
```diff
@@ -21,9 +21,6 @@ limitations under the License.
 #include "llvm/Support/Error.h"
 #include "mlir/Dialect/Complex/IR/Complex.h"
 #include "mlir/IR/BuiltinAttributes.h"
-#include "mlir/Support/DebugStringHelper.h"
```
sdasgup3, line 24: Keep them following https://google.github.io/styleguide/cppguide.html#Include_What_You_Use

stablehlo/reference/Element.h
```diff
@@ -23,6 +23,9 @@ limitations under the License.
 #include "llvm/Support/raw_ostream.h"
 #include "mlir/IR/BuiltinTypes.h"
 #include "mlir/IR/Types.h"
+#include "mlir/Support/DebugStringHelper.h"
```
sdasgup3, line 26: Keep them based on https://google.github.io/styleguide/cppguide.html#Include_What_You_Use

stablehlo/reference/Element.h
```diff
@@ -41,8 +44,45 @@ class Element {
...
+    } else if (isSupportedUnsignedIntegerType(type)) {
+      return Element(
+          type, APInt(type.getIntOrFloatBitWidth(), value, /*isSigned=*/false));
+    } else if (isSupportedFloatType(type)) {
+      APFloat floatVal((double)value);
+      bool roundingErr;
+      floatVal.convert(type.cast<FloatType>().getFloatSemantics(),
+                       APFloat::rmNearestTiesToEven, &roundingErr);
+      return Element(type, floatVal);
+    } else if (isSupportedComplexType(type)) {
```
sdasgup3, line 69: For complex type, use a `std::complex<V, V>` instead of `double`. This rhymes better with what an `Element` can internally store (`APInt, bool, APFloat, or std::pair<APFloat, APFloat>`).

stablehlo/reference/Element.h
```diff
@@ -41,8 +44,45 @@ class Element {
...
+  Element() = default;
   /// @}
 
+  /// @}
+  /// \name Value Generators
+  /// @{
+
+  /// Get the specified value for the specified type.
+  template <typename V>
+  static Element getValue(Type type, V value) {
```
sdasgup3, line 56: As we are not checking what are the allowable types for `V`.
Let us split this function into
```
Static Element getValue(Type, APInt);
Static Element getValue(Type, int64_t);
Static Element getValue(Type, double);
Static Element getValue(Type, APFloat);
Static Element getValue(Type, std::complex<double, double>);
Static Element getValue(Type, std::complex<APFLoat,APFloat>);
``` 

Another alternative is to use the template d version as is and specializing the template using checks (`std::is_integral`, `std::is_floating_point etc). However, splitting to me seems simpler for this use case.

ghpvnist: I agree, we can have a separate functions for each type since our current usage is pretty specific. However, since most usage will be providing integers to get the Element with specified type,
```
Static Element getValue(Type, int64_t);
```
would be the only option that is currently needed. For example, we don't really need
```
Static Element getValue(Type, APInt);
```
since this is the equivalent of
```
Element(Type, APInt)
```
constructor, which does not provide more utility. We can implement the rest (`double` and `std::complex<double, double>`) if needed in the future.

stablehlo/reference/Element.h
```diff
@@ -41,8 +44,44 @@ class Element {
...
   Element(const Element &other) = default;
+  Element() = default;
   /// @}
 
+  /// @}
+  /// \name Value Generators
+  /// @{
+
+  /// Get the specified value for the specified type.
+  static Element getValue(Type type, int64_t value) {
```
sdasgup3, line 55: If we are keeping this variant of `getValue`, let us add a comment along the lines 

```
The function `getValue(Type type, int64_t value)` produces an `Element` object using `type` and `value` in the following way

1. If `type` is an integer types, `value` is used with proper interpretation based on integer signedness.
2. If `type` is a floating point type, the `Element` object represents a floating point value equivalent to  `value`.
3. If `type` is a complex type,  the `Element` object represents a complex value with real part equivalent `value` and imaginary part as zero.
```
A sidenote on the inclusion of this function as Element`s public API:

In it's current form the function is very much customized for the use cases like
1. IotaOp: Need to generate Element equivalents for  index (which is an integer)
2. Get Element equivalents of One  to be used in `Log1pOp`: Need to generate Element equivalents for specific value of integer.

From the above two usecases, there does not seems to be an immediate need for 
```
Element getValue(double);
Element getValue(stad::complex<double,double>);
```

However, to me it might be good for completeness. The downside is the usecase (1) will become verbose as it has to deal with types separately. But that seems OK to me given `getValue(.., int64_t)` is similarly load bearing with the current implementation.

IMO, the design of public API of `Element` class seems non intuitive with the only  version`getValue(Type type, int64_t value)` included. Maybe the comment might help deal with the non intuitiveness. 

Feel free to seek for additional feedback from next reviewer.

ghpvnist: Makes sense. I've added APIs for double and complex as well. I have not added support for integers for the double and complex versions as it requires complicated semantics which can be covered by implementing the `ConvertOp` in the future to convert double Element to an integer Element (similarly with complex Element).

stablehlo/reference/Element.h
```diff
@@ -41,8 +44,62 @@ class Element {
...
 
   Element(const Element &other) = default;
+  Element() = default;
   /// @}
 
+  /// @}
+  /// \name Value Generators
+  /// @{
+
+  /// The function `getValue(Type type, int64_t value)` produces an `Element`
```
sdasgup3, line 54: remove the function name and signature? here and below. Instead just use
`The function produces ...`

stablehlo/reference/Ops.cpp
```diff
@@ -81,35 +81,20 @@ Tensor evalFloorOp(const Tensor &operand, Type resultType) {
...
-                   &roundingErr);
-      result.set(*it, Element(elType, std::complex<APFloat>(real, imag)));
+    if (isSupportedIntegerType(elementTy)) {
+      result.set(*it, Element::getValue(elementTy, (*it)[iotaDimension]));
+    } else if (isSupportedFloatType(elementTy)) {
+      result.set(*it, Element::getValue(elementTy, static_cast<double>(
+                                                       (*it)[iotaDimension])));
+    } else if (isSupportedComplexType(elementTy)) {
+      result.set(*it, Element::getValue(
+                          elementTy, std::complex<double>(static_cast<double>(
```
sdasgup3, line 93: Use `std::complex<double>(static_cast<double>(...), 0.0)` instead? Juts to be transparent to the choice of imaginary value and explicitly maintaining parity with spec.

stablehlo/reference/Element.h
sdasgup3: Thanks!

sdasgup3 (approved):
LGTM with a few nits.

stablehlo/reference/Element.h
```diff
@@ -41,8 +44,60 @@ class Element {
       : type_(type), value_(std::make_pair(value.real(), value.imag())) {}
 
   Element(const Element &other) = default;
+  Element() = default;
```
burmako, line 47: Why do we need to add a no-parameter constructor?

stablehlo/reference/Element.h
```diff
@@ -41,8 +44,60 @@ class Element {
...
   /// @}
 
+  /// @}
+  /// \name Value Generators
+  /// @{
```
burmako, line 52: What are the benefits of having this functionality as methods vs constructors? Given that we already have constructors which take Type and something else, why not just add more constructors?

stablehlo/reference/Element.h
```diff
@@ -41,8 +44,60 @@ class Element {
...
+      bool roundingErr;
+      real.convert(floatTy.getFloatSemantics(), APFloat::rmNearestTiesToEven,
+                   &roundingErr);
+      imag.convert(floatTy.getFloatSemantics(), APFloat::rmNearestTiesToEven,
+                   &roundingErr);
+      return Element(type, std::complex<APFloat>(real, imag));
+    }
+    report_fatal_error(invalidArgument("Unsupported element type: %s",
+                                       debugString(type).c_str()));
+  }
```
burmako, line 99: Let's move the definitions of these methods (or whatever we decide to refactor them to) into the .cpp file. I don't think we need them in the header file.

burmako (changes requested):
Looks like a good idea overall. Left a few logistical comments.

stablehlo/reference/Element.h
ghpvnist: This was is in DotGeneralOp to declare the Element object, but now with this PR, it is no longer needed. Will remove.

stablehlo/reference/Element.h
burmako: Hmm this doesn't look resolved.

burmako (changes requested):

stablehlo/reference/Element.h
ghpvnist: Ah my bad, I just addressed it.

stablehlo/reference/Element.cpp
```diff
@@ -105,34 +104,72 @@ Element mapWithUpcastToDouble(const Element &el, FloatFn floatFn,
...
-    APFloat resultReal(resultVal.real());
-    resultReal.convert(elSemantics, APFloat::rmNearestTiesToEven, &roundingErr);
-    APFloat resultImag(resultVal.imag());
-    resultImag.convert(elSemantics, APFloat::rmNearestTiesToEven, &roundingErr);
-    return Element(type, std::complex<APFloat>(resultReal, resultImag));
+    real.convert(floatTy.getFloatSemantics(), APFloat::rmNearestTiesToEven,
+                 &roundingErr);
+    imag.convert(floatTy.getFloatSemantics(), APFloat::rmNearestTiesToEven,
+                 &roundingErr);
+    *this = Element(type, std::complex<APFloat>(real, imag));
```
sdasgup3, line 166: Same here.

stablehlo/reference/Element.cpp
```diff
@@ -105,34 +104,72 @@ Element mapWithUpcastToDouble(const Element &el, FloatFn floatFn,
...
+
+  report_fatal_error(invalidArgument("Unsupported element type: %s",
+                                     debugString(type).c_str()));
+}
+
+}  // namespace
+
+Element::Element(Type type, APInt value) : type_(type), value_(value) {}
+
+Element::Element(Type type, int64_t value) {
```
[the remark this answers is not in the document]
sdasgup3: Do we need to handle separate cases or can just do:
```
if(!isSupportedIntegerType())
  report_fatal_error()

value_ = APInt(type.getIntOrFloatBitWidth(), value, isSupportedSignedIntegerType(type));
```

stablehlo/reference/Element.cpp
```diff
@@ -105,34 +104,73 @@ Element mapWithUpcastToDouble(const Element &el, FloatFn floatFn,
...
+
+Element::Element(Type type, APFloat value) : type_(type), value_(value) {}
+
+Element::Element(Type type, double value) {
+  type_ = type;
   if (isSupportedFloatType(type)) {
-    APFloat elVal = el.getFloatValue();
-    const llvm::fltSemantics &elSemantics = elVal.getSemantics();
-    APFloat resultVal(floatFn(elVal.convertToDouble()));
+    APFloat floatVal(static_cast<double>(value));
```
sdasgup3, line 142: This is more of a preference question, so feel free to keep it open.

How  about we avoid if/else if possible. To me (and here comes the preference) it seems more cleaner than `if {} else {}`. 

```
if(!isSupportedFloatType())
  report_fatal_error

APFloat floatVal(static_cast<double>(value));
 bool roundingErr;
floatVal.convert(type.cast<FloatType>().getFloatSemantics(),
                     APFloat::rmNearestTiesToEven, &roundingErr);
 value_ = floatVal;
```

same below.

sdasgup3 (approved):

stablehlo/reference/Ops.cpp
```diff
@@ -165,35 +165,21 @@ SmallVector<Tensor> evalIfOp(const Tensor &pred, Region &trueBranch,
 
 Tensor evalIotaOp(int64_t iotaDimension, TensorType resultType) {
   Tensor result(resultType);
-  Type elType = result.getElementType();
+  Type elementTy = result.getElementType();
```
burmako, line 168: Let's go even further and call this `elementType`? I don't think the `Ty` suffix is used anywhere in the file.

burmako (approved):

PR #1159, opened 2023-02-12
Update ODS documentation for yet unspecced ops

### LINKED ISSUE #739: Clearly document the plans for "Not in HLO" ops
At the moment, we aren't planning to spec "Not in HLO" ops, because the plan of record is to move them to CHLO (#3), and CHLO is very likely going to have its own spec (#602). We should clearly document these plans in: 1) the spec, 2) the ODS.

burmako:
For "Not in HLO" ops, the ODS now indicates that they are on their way out of StableHLO and informally explains their semantics.

For "Dynamism" ops, the ODS now says that they are being worked on and also informally explains their semantics.

Closes #739.

### DIFF
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -125,13 +125,14 @@ def StableHLO_IotaOp : StableHLO_Op<"iota", [Pure]> {
 }
 
 def StableHLO_DynamicIotaOp: StableHLO_ShapedInterfaceOp<"dynamic_iota", [Pure]> {
-  let summary = "Create linear increasing values from 0 to length -1.";
+  let summary = "DynamicIota operation";
   let description = [{
-    Produces an HLO Tensor of the specified shape, with an incremental set of
-    values along the specified dimension starting at 0.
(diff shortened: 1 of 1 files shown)


stablehlo/dialect/StablehloOps.td
```diff
@@ -3219,21 +3303,19 @@ def StableHLO_ComputeReshapeShapeOp :
...
-    %1 = stablehlo.cstr_reshapable 13, [2, -1] -> failure
-    ```
+    Informally, this operation creates a witness on the constraint that
+    ComputeReshapeShape would succeed with the provided operands.
 
     Example:
 
     ```mlir
-    %0 = stablehlo.cstr_reshapable %arg0, %arg1 : (index, tensor<3xi32>) -> !shape.witness
+    %0 = stablehlo.cstr_reshapable %num_elements, %dynamic_shape
```
sdasgup3, line 3317: `%result = `?

sdasgup3 (approved):
with one minor comment.

stablehlo/dialect/StablehloOps.td
burmako: Done.

PR #1161, opened 2023-02-13
Add spec for Quantized Types and Constants

### LINKED ISSUE #588: Spec quantization
This will involve documenting: 1) a representation for quantized tensors, 2) UniformQuantizeOp, 3) UniformDequantizeOp, 4) which of the existing ops can take quantized tensors and how their semantics changes from that.

subhankarshah:
closes #588

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -140,6 +140,7 @@ which may allow us to remove tuple types from StableHLO
 
 ```ebnf
 ElementType ::= BooleanType | IntegerType | FloatType | ComplexType
+                | QuantizedType
 BooleanType ::= 'i1'
 IntegerType ::= 'si4' | 'si8' | 'si16' | 'si32' | 'si64'
               | 'ui4' | 'ui8' | 'ui16' | 'ui32' | 'ui64'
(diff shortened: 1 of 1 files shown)


burmako:
@subhankarshah Thank you for your contribution! This is a valuable step towards documenting quantization exactly as it is informally implemented in the StableHLO dialect at the moment (although this PR alone won't be able to close #588, because subsequent steps will be needed to specify which operations allow quantized inputs and output and then define their behavior).

However, given that there's an ongoing RFC that touches quantization (#1149), let's put this pull request on hold until we complete the discussion of that RFC.

docs/spec.md
```diff
@@ -173,10 +174,66 @@ values of type `tensor<T>`).
...
+dequantized_value = (quantized_value − zero_point) * scale
+```
+
+  Given `alpha` and `beta` are maximum and minimum values (per-tensor or
+  per-axis) needed to be represented in `expressed_type`, the values of `scale`
+  and `zero_point` can be computed as follows -
+
+```
+scale = (alpha -  beta) / (max_val - min_val)
+zero_point = min_val - round(beta * scale)
```
sdasgup3, line 233: `beta/scale`?

burmako:
Closing this PR for now while we're figuring out the next steps for speccing quantization. Thanks again for your work - we will use it as a reference one way or another!

docs/spec.md
subhankarshah: ah yes

Issue #581, opened 2022-11-23
Consider updating case, if and while to take/produce tuples


burmako:
Given that `custom_call` can produce tuples, I think that control flow ops should be able to do that as well, unless I'm missing something.

burmako:
The same argument applies to OptimizationBarrierOp.

Issue #365, opened 2022-10-19
Should `if`, `case` and `while` ops allow 0 outputs.


subhankarshah:
### Request description

Note: Since `while` inputs and outputs have same type and size, this means while also has 0 inputs.

PR #1164, opened 2023-02-14
Add interpreter for IfOp

### LINKED ISSUE #977: Add interpreter for IfOp

ghpvnist:

```
(I1) pred 0-dimensional tensor constant of type `i1`.
(I2) true_branch function.
(I3) false_branch function.
(C1) `true_branch` and `false_branch` have 0 inputs.
(C2) `true_branch` and `false_branch` have the same output types.
(C3) For all `i`, `type(results[i]) = type(true_branch.outputs[i])`.
```

```
I1: a) pred is not 0-dimensional.
    b) element_type(pred) != `i1`. (Covered by ODS).
I2: a) true_branch is not a function. (Covered by ODS).
I3: a) false_branch is not a function. (Covered by ODS).
C1: a) `true_branch` does not have 0 inputs.
    b) `false_branch` does not have 0 inputs.
C2: a) `true_branch` and `false_branch` does not have the same output types.
C3: no additional tests are added since this it's just inferring the type.
```

```
I1a: pred is not 0-dimensional.
C1a: `true_branch` does not have 0 inputs.
C1b: `false_branch` does not have 0 inputs.
C2a: `true_branch` and `false_branch` does not have the same output types.
```

closes #977


ghpvnist:
Tagging issues related to IfOp for future reference: #365, #581

stablehlo/reference/Ops.cpp
```diff
@@ -91,6 +91,14 @@ Tensor evalFloorOp(const Tensor &operand, Type resultType) {
   return result;
 }
 
+SmallVector<Tensor> evalIfOp(const Tensor &pred, Region &trueBranch,
+                             Region &falseBranch, Scope &scope) {
+  SmallVector<Tensor> runtimeResults;
```
sdasgup3, line 96: No need to use this vector for empty args. Pass empty args as `{}` to the evals instead.

stablehlo/reference/Ops.cpp
```diff
@@ -301,6 +309,11 @@ SmallVector<Tensor> eval(Region &region, ArrayRef<Tensor> args, Scope *parent) {
       Tensor runtimeOperand = scope.find(floorOp.getOperand());
       Tensor runtimeResult = evalFloorOp(runtimeOperand, floorOp.getType());
       scope.add(op.getResults(), {runtimeResult});
+    } else if (auto ifOp = dyn_cast<IfOp>(op)) {
+      Tensor runtimePred = scope.find(ifOp.getPred());
+      auto runtimeResults = evalIfOp(runtimePred, ifOp.getTrueBranch(),
+                                     ifOp.getFalseBranch(), scope);
+      scope.add(op.getResults(), {runtimeResults});
```
sdasgup3, line 316: `{runtimeResults}` -> `runtimeResults` as it is already a vector.

stablehlo/reference/Scope.cpp
```diff
@@ -28,14 +28,15 @@ void Scope::add(Value ssaValue, Tensor runtimeValue) {
   if (stack_frame_.count(ssaValue))
     llvm::report_fatal_error("Duplicate SSA register found in scope");
 
-  if (ssaValue.getType() != runtimeValue.getType())
-    llvm::report_fatal_error(
-        "Expected same type for an SSA register and its evaluated value");
+  // if (ssaValue.getType() != runtimeValue.getType())
```
sdasgup3, line 31: Do we need to comment them out?

stablehlo/reference/Scope.cpp
```diff
@@ -28,14 +28,15 @@ void Scope::add(Value ssaValue, Tensor runtimeValue) {
...
+  //   llvm::report_fatal_error(
+  //       "Expected same type for an SSA register and its evaluated value");
 
   stack_frame_[ssaValue] = runtimeValue;
 }
 
 void Scope::add(ValueRange ssaValues, ArrayRef<Tensor> runtimeValues) {
+  llvm::errs() << ssaValues.size() << ", " << runtimeValues.size() << "\n";
```
sdasgup3, line 39: remove?

stablehlo/reference/Scope.cpp
ghpvnist: Artifact from debugging :) thanks for catching!

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -945,128 +945,105 @@ func.func @broadcast_in_dim_unranked_operand(%arg0 : tensor<*xf32>) -> tensor<2x
...
-    }) : (tensor<i1>) -> tensor<f32>
-  func.return
+      "stablehlo.return"(%false_branch_eperand) : (tensor<?xf32>) -> ()
+    }) : (tensor<i1>) -> tensor<2xf32>
+  func.return %0 : tensor<2xf32>
 }
+
 // -----
 
-func.func @if_mismatch_return_type(%pred : tensor<i1>, %branch_operand : tensor<f32>) {
```
sdasgup3, line 1062: Did we miss this out?

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3391,6 +3394,16 @@ LogicalResult verifyDynamicReshapeOp(std::optional<Location> location,
   return success();
 }
 
+LogicalResult verifyIfOp(std::optional<Location> location, Value pred) {
+  auto predType = pred.getType().dyn_cast<RankedTensorType>();
```
sdasgup3, line 3398: Good catch on `pred` shape!

sdasgup3 (approved):
LGTM with a minor comment.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1301,9 +1301,16 @@ static LogicalResult inferGatherReturnTypeComponents(
...
+LogicalResult inferConditionalOp(Optional<Location> location, Value operand,
                                  RegionRange branches,
                                  SmallVectorImpl<Type>& inferredReturnTypes) {
+  // if_i1
+  auto operandRankedTy = operand.getType().dyn_cast<RankedTensorType>();
+  if (operandRankedTy && operandRankedTy.getRank() != 0)
+    return emitOptionalError(location,
+                             "operand should be rank 0 tensor but got rank ",
+                             operandRankedTy.getRank());
+  // if_i2, if_i3
```
burmako, line 1313: I don't think that `if_i2` and `if_i3` apply here. The fact that branches are functions is already validated by the ODS.

stablehlo/tests/interpret_if.mlir
```diff
@@ -0,0 +1,27 @@
...
+    %false_branch_result = "stablehlo.if"(%false_branch_pred) ({
+      %1 = stablehlo.constant dense<1> : tensor<2xi64>
+      "stablehlo.return"(%1) : (tensor<2xi64>) -> ()
+    }, {
+      %2 = stablehlo.constant dense<2> : tensor<2xi64>
+      "stablehlo.return"(%2) : (tensor<2xi64>) -> ()
+    }) : (tensor<i1>) -> tensor<2xi64>
+    "stablehlo.return"(%false_branch_result, %false_branch_result) : (tensor<2xi64>, tensor<2xi64>) -> ()
+  }) : (tensor<i1>) -> (tensor<2xi64>, tensor<2xi64>)
+  func.return %result0, %result1 : tensor<2xi64>, tensor<2xi64>
```
burmako, line 20: Can we perhaps have two tests - one for `if_true` and another one for `if_false`? That'll be easier to read than a nested test.

stablehlo/tests/interpret_if.mlir
```diff
@@ -0,0 +1,27 @@
+// RUN: stablehlo-interpreter --interpret -split-input-file %s | FileCheck %s
+
+// CHECK-LABEL: Evaluated results of function: if
+func.func @if() -> (tensor<2xi64>, tensor<2xi64>) {
+  %pred = stablehlo.constant dense<false> : tensor<i1>
+  %result0, %result1 = "stablehlo.if"(%pred) ({
+    %0 = stablehlo.constant dense<0> : tensor<2xi64>
+    "stablehlo.return"(%0, %0) : (tensor<2xi64>, tensor<2xi64>) -> ()
```
burmako, line 8: How about using pretty syntax for stablehlo.return? This will make things easier to write and read.

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -945,128 +945,117 @@ func.func @broadcast_in_dim_unranked_operand(%arg0 : tensor<*xf32>) -> tensor<2x
...
+      "stablehlo.return"(%branch_operand, %branch_operand) : (tensor<f32>, tensor<f32>) -> ()
+  }, {
+      %2 = "stablehlo.if"(%pred) ({
+          "stablehlo.return"(%pred) : (tensor<i1>) -> ()
+      }, {
+          "stablehlo.return"(%pred) : (tensor<i1>) -> ()
+      }) : (tensor<i1>) -> tensor<i1>
+      "stablehlo.return"(%branch_operand, %branch_operand) : (tensor<f32>, tensor<f32>) -> ()
+  }) : (tensor<i1>) -> (tensor<f32>, tensor<f32>)
+  func.return %0, %1 : tensor<f32>, tensor<f32>
```
burmako, line 960: Why do we need a more complicated test here? Does it test something that the simpler test didn't test?

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -945,128 +945,117 @@ func.func @broadcast_in_dim_unranked_operand(%arg0 : tensor<*xf32>) -> tensor<2x
...
+    }) : (tensor<i1>) -> tensor<i32>
+  func.return %0 : tensor<i32>
 }
 
 // -----
 
-func.func @if_unexpected_arguments_in_region_of_false_branch(%pred : tensor<i1>, %branch_operand : tensor<f32>) {
-  // @expected-error@+1 {{branch 1 must have 0 arguments, but found 1}}
+// CHECK-LABEL: if_dynamic_branch_result
+func.func @if_dynamic_branch_result(%pred : tensor<i1>, %true_branch_eperand: tensor<2xf32>, %false_branch_eperand : tensor<?xf32>) -> tensor<2xf32> {
```
burmako, line 1016: Typo: "operand". (Same below).

docs/spec.md
```diff
@@ -2861,7 +2861,7 @@ output of `true_branch` is returned, else if pred is `false`, output of
 
 * (C1) `true_branch` and `false_branch` have 0 inputs.
 * (C2) `true_branch` and `false_branch` have the same output types.
-* (C3) For all `i`, `type(results[i]) = type(true_branch).outputs[i]`.
+* (C3) For all `i`, `type(results[i]) = type(true_branch.outputs[i])`.
```
burmako, line 2864: I think both notations can plausibly work, so how about we make a note about this in #289, and leave the current syntax unchanged (otherwise, we'd want to revisit all places where similar syntax is used).

stablehlo/tests/ops_stablehlo.mlir
ghpvnist: There were two positive tests, one with and another without nested `if` operations. I decided to use the more complex one to remove "duplicate" positive tests. If the purpose of the positive tests in this file is for coherence check that the op works, then I think it's fine to include a simpler one. Another option is to have both tests, and I don't have preference for either option. What are your thoughts?

stablehlo/tests/interpret_if.mlir
ghpvnist: I intentionally included this one to comprehensively test that:
1) Both if_true and if_false works
2) Nested if works
3) Can return variadic outputs

I agree we should have two tests since the code has two logical paths as per the testing guideline. However, we would still need to test 2) and 3), which we need to include in both paths to have minimal tests (unless such positive tests are in ops_stablehlo.mlir, but still debatable). Thoughts?

burmako: Agreed by 1 and 3, but why do we need to test 2? If has two regions, but these regions aren't special, right?

stablehlo/tests/ops_stablehlo.mlir
burmako: I think that minimizing the number and the composition of tests is important for maintainability. We have only captured this in https://github.com/openxla/stablehlo/blob/main/docs/reference.md#testing-guidelines, but I think this also applies to ops_stablehlo.mlir.

stablehlo/tests/interpret_if.mlir
ghpvnist: Agreed with not testing nested behaviors. This behavior should be tested separately from the ops.

burmako (approved):

PR #1167, opened 2023-02-14
Add support for DynamicConvOp to --stablehlo-refine-shapes

burmako:
This PR adds a RefineConvolutionOpPattern pattern that reuses inferConvolutionOp to do shape refinement for DynamicConvOp.

Additionally, it makes some changes to constant folding patterns to support related shape computations.

### DIFF
--- a/stablehlo/transforms/StablehloRefineShapes.cpp
+++ b/stablehlo/transforms/StablehloRefineShapes.cpp
@@ -172,8 +172,8 @@ struct EvalConcatenateOpPattern : public OpRewritePattern<ConcatenateOp> {
   LogicalResult matchAndRewrite(ConcatenateOp op,
                                 PatternRewriter& rewriter) const override {
     auto resultType = op.getResult().getType().dyn_cast<RankedTensorType>();
-    if (!resultType || resultType.getRank() != 1)
-      return rewriter.notifyMatchFailure(op, "expected 1-dimensional type");
+    if (!resultType || op.getDimension() != 0)
+      return rewriter.notifyMatchFailure(op, "expected dimension = 0");
 
     SmallVector<APInt> result;
     for (Value operand : op->getOperands()) {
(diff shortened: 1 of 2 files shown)


stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -215,6 +215,18 @@ struct EvalConvertOpPattern : public OpRewritePattern<ConvertOp> {
...
 };
 
+struct EvalDivOpPattern : public OpRewritePattern<DivOp> {
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(DivOp op,
+                                PatternRewriter& rewriter) const override {
+    auto resultType = op.getResult().getType().cast<ShapedType>();
+    auto isResultUnsigned = resultType.getElementType().isUnsignedInteger();
+    return evalBinary(rewriter, op, [&](APInt lhs, APInt rhs) {
+      return isResultUnsigned ? lhs.udiv(rhs) : lhs.sdiv(rhs);
```
GleasonK, line 225: Is this a safe optimization to make always? It looks like we're using the same functions in the reference interpreter, not sure if the "upcast to double" is important or an implementation detail:
https://github.com/openxla/stablehlo/pull/1008/files#diff-c2f81cc25e3bff18b8f8ab80deabf5e1c02ee728c3902b5e99ff0505620d27c7R260

GleasonK (approved):

stablehlo/transforms/StablehloRefineShapes.cpp
sdasgup3: Just to add one detail: 

For integer types, the reference implementation PR uses  `udiv` or `sdiv` without "upcast to double". The mapper function name "mapWithUpcastToDouble" is a bit misleading in that sense.

burmako: I think so, yes, given that it conforms to the documented semantics of DivideOp. The "upcast to double" part in the reference implementation only applies to the floating-point part of interpreting the op.

PR #1169, opened 2023-02-14
Prototype a mechanism to customize shape refinement for CustomCallOp

burmako:
This is a prototype of the CustomCallOp shape refinement mechanism to enable specializing dynamically-shaped programs with custom calls to static shapes (#851).

Based on the results of prototyping, I'm planning to propose an RFC to extend CustomCallOp, so that the logic can migrate from using an unregistered attribute to something more solid.

### DIFF
--- a/stablehlo/transforms/StablehloRefineShapes.cpp
+++ b/stablehlo/transforms/StablehloRefineShapes.cpp
@@ -603,6 +603,8 @@ struct RefineCustomCallOpPattern : public OpRewritePattern<CustomCallOp> {
       SmallVector<int64_t> refinement;
       if (failed(matchInts(op->getOperand(operandIndex), refinement)))
         return rewriter.notifyMatchFailure(op, "expected constant operand");
+      if (llvm::any_of(refinement, [&](int64_t x) { return x < 0; }))
+        return rewriter.notifyMatchFailure(op, "expected non-negative sizes");
       refinements.emplace_back(refinement);
     }
     return refineReturnTypes(rewriter, op, refinements);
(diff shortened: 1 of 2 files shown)


stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -585,6 +585,30 @@ struct RefineConvolutionOpPattern : public OpRewritePattern<ConvolutionOp> {
...
+
+    SmallVector<ShapedTypeComponents> refinements;
+    for (auto operandIndexElt : operandIndicesAttr.getValues<APInt>()) {
+      int64_t operandIndex = operandIndexElt.getSExtValue();
+      if (operandIndex < 0 || operandIndex >= op->getNumOperands())
+        return rewriter.notifyMatchFailure(op, "expected valid operand index");
+      SmallVector<int64_t> refinement;
+      if (failed(matchInts(op->getOperand(operandIndex), refinement)))
+        return rewriter.notifyMatchFailure(op, "expected constant operand");
+      refinements.emplace_back(refinement);
```
GleasonK, line 606: Do we need any validation for "all values must be positive" ?

GleasonK (approved):

stablehlo/transforms/StablehloRefineShapes.cpp
burmako: Good catch! I've updated the code here and opened a similar ticket for verification in general (#1170).

GleasonK: lgtm

PR #1171, opened 2023-02-15
Add interpreter for CaseOp

### LINKED ISSUE #965: Add interpreter for case

ghpvnist:

```
(I1) index 0-dimensional tensor constant of type `si32`.
(I2) branches variadic number of functions.
(C1) `branches` have at least one function.
(C2) All functions in `branches` have 0 inputs.
(C3) All functions in `branches` have the same output types.
(C4) For all `i`, `type(results[i]) = type(branches[0].outputs[i])`.
```

```
I1: a) index is not 0-dimensional tensor.
    b) element_type(index) != `si32`. (Covered by ODS).
I2: a) branches is not variadic functions. (Covered by ODS).
C1: a) size(branches) < 1.
C2: a) not all functions in `branches` have 0 inputs.
C3: a) not all functions in `branches` have the same output types.
C4: a) type(results[i]) != type(branches[0].outputs[i]) for any i.
```

```
I1a: index is not 0-dimensional tensor.
C1a: size(branches) < 1.
C2a: not all functions in `branches` have 0 inputs.
C3a: not all functions in `branches` have the same output types.
C4a: type(results[i]) != type(branches[0].outputs[i]) for any i.
```

Notes:
* Fixed typo:
```
`index` | 1-dimensional tensor of type `si32`
```
to
```
`index` | 0-dimensional tensor constant of type `si32`
```

* The status for verification remains `revisit` due to #365 and #581.

closes #965

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -1320,10 +1320,10 @@ returned.
 
 #### Inputs
 
-| Label | Name       | Type                                | Constraints |
-|-------|------------|-------------------------------------|-------------|
-| (I1)  | `index`    | 1-dimensional tensor of type `si32` |             |
-| (I2)  | `branches` | variadic number of functions        | (C1-C4)     |
+| Label | Name       | Type                                         | Constraints |
+|-------|------------|----------------------------------------------|-------------|
+| (I1)  | `index`    | 0-dimensional tensor constant of type `si32` |             |
+| (I2)  | `branches` | variadic number of functions                 | (C1-C4)     |
 
 #### Outputs
 
@@ -1336,7 +1336,7 @@ returned.
 * (C1) `branches` have at least one function.
 * (C2) All functions in `branches` have 0 inputs.
 * (C3) All functions in `branches` have the same output types.
-* (C4) For all `i`, `type(results[i]) = type(branches[0]).outputs[i]`.
+* (C4) For all `i`, `type(results[i]) = type(branches[0].outputs[i])`.
 
 #### Examples
 
@@ -1351,6 +1351,8 @@ returned.
 }) : (tensor<i32>) -> tensor<i32>
 // %result: 11
 ```
+
+&nbsp;[More Examples](../stablehlo/tests/interpret_case.mlir)
 
 ### cbrt
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -1342,14 +1342,15 @@ returned.
 
 ```mlir
 // %index: -1
-// %result_branch0: 0
-// %result_branch1: 1
-%result = "stablehlo.case"(%index) ({
-  "stablehlo.return"(%result_branch0) : (tensor<2xi64>) -> ()
+// %result_branch0: [0, 0]
+// %result_branch1: [1, 1]
+%result0, %result1 = "stablehlo.case"(%index) ({
+  "stablehlo.return"(%result_branch0, %result_branch0) : (tensor<2xi64>, tensor<2xi64>) -> ()
 }, {
-  "stablehlo.return"(%result_branch1) : (tensor<2xi64>) -> ()
-}) : (tensor<i32>) -> tensor<2xi64>
-// %result: [1, 1]
+  "stablehlo.return"(%result_branch1, %result_branch1) : (tensor<2xi64>, tensor<2xi64>) -> ()
+}) : (tensor<i32>) -> (tensor<2xi64>, tensor<2xi64>)
+// %result0: [1, 1]
+// %result1: [1, 1]
 ```
 
 &nbsp;[More Examples](../stablehlo/tests/interpret_case.mlir)
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -1341,15 +1341,15 @@ returned.
 #### Examples
 
 ```mlir
-// %result_branch0: 10
-// %result_branch1: 11
-// %index: 1
+// %index: -1
+// %result_branch0: 0
+// %result_branch1: 1
 %result = "stablehlo.case"(%index) ({
-  "stablehlo.return"(%result_branch0) : (tensor<i32>) -> ()
+  "stablehlo.return"(%result_branch0) : (tensor<2xi64>) -> ()
 }, {
-  "stablehlo.return"(%result_branch1) : (tensor<i32>) -> ()
-}) : (tensor<i32>) -> tensor<i32>
-// %result: 11
+  "stablehlo.return"(%result_branch1) : (tensor<2xi64>) -> ()
+}) : (tensor<i32>) -> tensor<2xi64>
+// %result: [1, 1]
 ```
 
 &nbsp;[More Examples](../stablehlo/tests/interpret_case.mlir)
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -1320,10 +1320,10 @@ returned.
 
 #### Inputs
 
-| Label | Name       | Type                                         | Constraints |
-|-------|------------|----------------------------------------------|-------------|
-| (I1)  | `index`    | 0-dimensional tensor constant of type `si32` |             |
-| (I2)  | `branches` | variadic number of functions                 | (C1-C4)     |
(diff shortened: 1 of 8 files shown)


stablehlo/reference/Ops.cpp
```diff
@@ -78,6 +78,12 @@ Tensor evalBroadcastInDimOp(const Tensor &operand,
   return result;
 }
 
+SmallVector<Tensor> evalCaseOp(const Tensor &index,
+                               MutableArrayRef<Region> branches, Scope &scope) {
+  return eval(branches[index.get({}).getIntegerValue().getSExtValue()], {},
```
sdasgup3, line 83: maybe we should check for index out of bound (>= branches.size()). If yes, then the `branches[branches.size()-1]` is chosen.

stablehlo/reference/Ops.cpp
```diff
@@ -78,6 +78,14 @@ Tensor evalBroadcastInDimOp(const Tensor &operand,
   return result;
 }
 
+SmallVector<Tensor> evalCaseOp(const Tensor &index,
+                               MutableArrayRef<Region> branches, Scope &scope) {
```
sdasgup3, line 82: Can we use `RegionRange` instead?

https://github.com/llvm/llvm-project/blob/c2e248c6aea6396dbfe3bd3ed08ce587b5d63283/mlir/include/mlir/IR/Region.h#L321
 
The subsequent code will look like. 
```
return eval(*branches[idx], {}, &scope);
```

Use of `MutatbleArrayRef` gives the impression that we are going to change the regions in someway. `RegionRange` avoids that.

stablehlo/tests/interpret_case.mlir
```diff
@@ -0,0 +1,46 @@
...
+func.func @case() -> (tensor<2xi64>, tensor<2xi64>) {
+  %index = stablehlo.constant dense<1> : tensor<i32>
+  %result_branch0 = stablehlo.constant dense<0> : tensor<2xi64>
+  %result_branch1 = stablehlo.constant dense<1> : tensor<2xi64>
+  %result0, %result1 = "stablehlo.case"(%index) ({
+    "stablehlo.return"(%result_branch0, %result_branch0) : (tensor<2xi64>, tensor<2xi64>) -> ()
+  }, {
+    "stablehlo.return"(%result_branch1, %result_branch1) : (tensor<2xi64>, tensor<2xi64>) -> ()
+  }, {
+    "stablehlo.return"(%result_branch0, %result_branch0) : (tensor<2xi64>, tensor<2xi64>) -> ()
```
sdasgup3, line 13: We can one two two regions one default and other non-default. This avoids the mix of values.

sdasgup3: And below we can use the same example using an OOB index value to pick the default branch.

docs/spec.md
```diff
@@ -1336,22 +1336,24 @@ returned.
...
+* (C4) For all `i`, `type(results[i]) = type(branches[0].outputs[i])`.
 
 #### Examples
 
 ```mlir
-// %result_branch0: 10
-// %result_branch1: 11
-// %index: 1
+// %index: -1
+// %result_branch0: 0
```
sdasgup3, line 1345: `%result_branch0: [0, 0]` and below

stablehlo/tests/interpret_case.mlir
```diff
@@ -0,0 +1,38 @@
...
+// CHECK-NEXT: 0 : i64
+// CHECK-NEXT: 0 : i64
+// CHECK-NEXT: tensor<2xi64>
+// CHECK-NEXT: 0 : i64
+// CHECK-NEXT: 0 : i64
+
+// -----
+
+// CHECK-LABEL: Evaluated results of function: case_default
+func.func @case_default() -> (tensor<2xi64>) {
```
sdasgup3, line 25: Can we use the same example above?

sdasgup3 (approved):

stablehlo/tests/interpret_case.mlir
ghpvnist: This example comes from the spec (that tests default case) whereas the other test is more generic (e.g. returns multiple values). Actually, let me merge the two and edit the spec example.

docs/spec.md
```diff
@@ -1336,22 +1336,25 @@ returned.
 * (C1) `branches` have at least one function.
 * (C2) All functions in `branches` have 0 inputs.
 * (C3) All functions in `branches` have the same output types.
-* (C4) For all `i`, `type(results[i]) = type(branches[0]).outputs[i]`.
+* (C4) For all `i`, `type(results[i]) = type(branches[0].outputs[i])`.
```
burmako, line 1339: Similar comment to #1164. Let's revert this change and leave this to #289.

stablehlo/tests/interpret_case.mlir
```diff
@@ -0,0 +1,15 @@
+// RUN: stablehlo-interpreter --interpret -split-input-file %s
+
+func.func @case() {
+  %index = stablehlo.constant dense<-1> : tensor<i32>
```
burmako, line 4: Sounds like we need three tests here: 1) less than zero, 2) in bounds, 3) greater or equal than the number of branches.

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -1048,115 +1048,115 @@ func.func @if_unranked(%pred : tensor<i1>, %true_branch_operand: tensor<2xf32>,
...
   }, {
-      "stablehlo.return"(%branch_operand) : (tensor<2xf32>) -> ()
-  }) : (tensor<i32>) -> tensor<2xf32>
-  func.return
+      %2 = "stablehlo.case"(%index) ({
+          "stablehlo.return"(%index) : (tensor<i32>) -> ()
+      }) : (tensor<i32>) -> tensor<i32>
+      "stablehlo.return"(%branch_operand, %branch_operand) : (tensor<f32>, tensor<f32>) -> ()
+  }) : (tensor<i32>) -> (tensor<f32>, tensor<f32>)
+  func.return %0, %1 : tensor<f32>, tensor<f32>
```
burmako, line 1060: Similarly to #1174, let's simplify the test case by dropping the nested case.

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -1048,115 +1048,115 @@ func.func @if_unranked(%pred : tensor<i1>, %true_branch_operand: tensor<2xf32>,
...
-  func.return
+      %1 = "stablehlo.negate"(%operand_1) : (tensor<f32>) -> tensor<f32>
+      "stablehlo.return"(%1) : (tensor<f32>) -> ()
+    },  {
+      %1 = stablehlo.constant dense<2> : tensor<i32>
+      "stablehlo.return"(%1) : (tensor<i32>) -> ()
+    },  {
+      %1 = "stablehlo.floor"(%operand_3) : (tensor<f32>) -> tensor<f32>
+      "stablehlo.return"(%1) : (tensor<f32>) -> ()
+    }
```
burmako, line 1097: Can we trim this to two branches? The third one seems irrelevant for the test.

stablehlo/reference/Ops.cpp
```diff
@@ -75,6 +75,14 @@ Tensor evalBroadcastInDimOp(const Tensor &operand, Axes broadcastDimensions,
   return result;
 }
 
+SmallVector<Tensor> evalCaseOp(const Tensor &index, RegionRange branches,
+                               Scope &scope) {
+  int64_t idx = index.get({}).getIntegerValue().getSExtValue();
+  if (idx < 0 || idx >= static_cast<int64_t>(branches.size()))
+    return eval(*branches[branches.size() - 1], {}, &scope);
```
burmako, line 82: `idx = branches.size() - 1`? This will avoid a duplicate call to `eval`.

burmako (approved):

Issue #1163, opened 2023-02-14
Consider changing IndexSpaceIterator to return Index


burmako:
At the moment, it returns `ArrayRef<int64_t>` but I think this is starting to become insufficient - in several recent PRs, I saw loops that perform elementwise operations on these things. Given that we're focusing on readability, I think it would be better to encapsulate that in a dedicated class `Index` and its overloaded operators / methods.

PR #1173, opened 2023-02-15
Introduce Index class

sdasgup3:
closes https://github.com/openxla/stablehlo/issues/1163

The PR:
1. Introduces a class `Index` to represent the index vector of a `Tensor`. That is to replace the explicit use of `ArrayRef<int64>` for the purpose. 
2. `Tensor::begin_index` and `Tensor::end_index` to return `Index` objects instead.
3. Promote the use of `Index` in various `evalFooOp`. The use of overload `Index::operator+` is handy for `DynamicSlice`  and `DynamicUpdateSlice` op.

### DIFF
--- a/stablehlo/reference/Index.cpp
+++ b/stablehlo/reference/Index.cpp
@@ -9,7 +9,7 @@ You may obtain a copy of the License at
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permutationissions and
+See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 
--- a/stablehlo/reference/Index.cpp
+++ b/stablehlo/reference/Index.cpp
@@ -9,7 +9,7 @@ You may obtain a copy of the License at
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
+See the License for the specific language governing permutationissions and
 limitations under the License.
 ==============================================================================*/
 
@@ -21,17 +21,50 @@ namespace mlir {
 namespace stablehlo {
 
 Index Index::operator+(const Index &other) const {
-  SmallVector<int64_t> combined;
+  if (size() != other.size())
+    llvm::report_fatal_error("Index add expects operands of same size.");
+
+  SmallVector<int64_t> result;
   for (auto [lhsIdx, rhsIdx] : llvm::zip(index_, other.index_))
-    combined.push_back(lhsIdx + rhsIdx);
-  return Index(combined);
+    result.push_back(lhsIdx + rhsIdx);
+  return Index(result);
+}
+
+Index Index::operator+(ArrayRef<int64_t> array) const {
+  if (size() != array.size())
+    llvm::report_fatal_error("Index add expects operands of same size.");
+
+  SmallVector<int64_t> result;
+  for (auto [lhsIdx, rhsIdx] : llvm::zip(index_, array))
+    result.push_back(lhsIdx + rhsIdx);
+  return Index(result);
+}
+
+Index Index::operator*(const Index &other) const {
+  if (size() != other.size())
+    llvm::report_fatal_error("Index product expects operands of same size.");
+
+  SmallVector<int64_t> result;
+  for (auto [lhsIdx, rhsIdx] : llvm::zip(index_, other.index_))
+    result.push_back(lhsIdx * rhsIdx);
+  return Index(result);
+}
+
+Index Index::operator*(ArrayRef<int64_t> array) const {
+  if (size() != array.size())
+    llvm::report_fatal_error("Index product expects operands of same size.");
+
+  SmallVector<int64_t> result;
+  for (auto [lhsIdx, rhsIdx] : llvm::zip(index_, array))
+    result.push_back(lhsIdx * rhsIdx);
+  return Index(result);
 }
 
 LogicalResult verifyIndex(llvm::ArrayRef<int64_t> shape, const Index &index) {
   if (shape.size() != index.size()) return failure();
 
-  for (auto [dim, shapeDim] : llvm::enumerate(shape))
-    if (index[dim] < 0 || index[dim] >= shapeDim) return failure();
+  for (auto [shapeDim, indexDim] : llvm::zip(shape, index.getIndexArray()))
+    if (indexDim < 0 || indexDim >= shapeDim) return failure();
 
   return success();
 }
@@ -68,6 +101,25 @@ IndexSpaceIterator IndexSpaceIterator::operator++(int) {
   IndexSpaceIterator tempIter = *this;
   ++*this;
   return tempIter;
+}
+
+Index Index::permute(ArrayRef<int64_t> permutation) {
+  if (size() != permutation.size())
+    llvm::report_fatal_error(
+        "Index permutationute expects permutation of same size as the index.");
+
+  Index result(size());
+  for (size_t i = 0; i < permutation.size(); i++)
+    result[i] = (*this)[permutation[i]];
+  return result;
+}
+
+Index operator+(ArrayRef<int64_t> array, const Index &index) {
+  return index + array;
+}
+
+Index operator*(ArrayRef<int64_t> array, const Index &index) {
+  return index * array;
 }
 
 }  // namespace stablehlo
--- a/stablehlo/reference/Index.cpp
+++ b/stablehlo/reference/Index.cpp
@@ -18,13 +18,13 @@ limitations under the License.
 namespace mlir {
 namespace stablehlo {
 
-const Sizes &IndexSpaceIterator::operator*() const {
+const Index &IndexSpaceIterator::operator*() const {
   if (!index_)
     llvm::report_fatal_error("Dereferencing a past-the-end iterator.");
   return *index_;
 }
 
-const Sizes *IndexSpaceIterator::operator->() const { return &(*index_); }
+const Index *IndexSpaceIterator::operator->() const { return &(*index_); }
 
 IndexSpaceIterator &IndexSpaceIterator::operator++() {
   if (!index_)
--- a/stablehlo/reference/Index.cpp
+++ b/stablehlo/reference/Index.cpp
@@ -15,65 +15,16 @@ limitations under the License.
 
 #include "stablehlo/reference/Index.h"
 
-#include "llvm/Support/Error.h"
-
 namespace mlir {
 namespace stablehlo {
 
-Index Index::operator+(const Index &other) const {
-  if (size() != other.size())
-    llvm::report_fatal_error("Index add expects operands of same size.");
-
-  SmallVector<int64_t> result;
-  for (auto [lhsIdx, rhsIdx] : llvm::zip(index_, other.index_))
-    result.push_back(lhsIdx + rhsIdx);
-  return Index(result);
-}
-
-Index Index::operator+(ArrayRef<int64_t> array) const {
-  if (size() != array.size())
-    llvm::report_fatal_error("Index add expects operands of same size.");
-
-  SmallVector<int64_t> result;
-  for (auto [lhsIdx, rhsIdx] : llvm::zip(index_, array))
-    result.push_back(lhsIdx + rhsIdx);
-  return Index(result);
-}
-
-Index Index::operator*(const Index &other) const {
-  if (size() != other.size())
-    llvm::report_fatal_error("Index product expects operands of same size.");
-
-  SmallVector<int64_t> result;
-  for (auto [lhsIdx, rhsIdx] : llvm::zip(index_, other.index_))
-    result.push_back(lhsIdx * rhsIdx);
-  return Index(result);
-}
-
-Index Index::operator*(ArrayRef<int64_t> array) const {
-  if (size() != array.size())
-    llvm::report_fatal_error("Index product expects operands of same size.");
-
-  SmallVector<int64_t> result;
-  for (auto [lhsIdx, rhsIdx] : llvm::zip(index_, array))
-    result.push_back(lhsIdx * rhsIdx);
-  return Index(result);
-}
-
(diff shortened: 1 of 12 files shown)


ghpvnist (approved):

stablehlo/reference/Ops.cpp
```diff
@@ -31,21 +31,14 @@ namespace stablehlo {
...
-// Applies the permutation `perm` to an array `array` where perm[i] indicates
+// Applies the permutation `perm` to an index `array` where perm[i] indicates
 // the location where the current array[i] goes.
-SmallVector<int64_t> permute(ArrayRef<int64_t> array, ArrayRef<int64_t> perm) {
-  SmallVector<int64_t> result(array.size());
+Index permute(const Index &array, ArrayRef<int64_t> perm) {
+  Index result(array.size());
   for (size_t i = 0; i < array.size(); i++) result[i] = array[perm[i]];
   return result;
 }
```
burmako, line 40: Looks like this also could be a method on `Index`?

stablehlo/reference/Ops.cpp
```diff
@@ -282,9 +278,9 @@ Tensor evalSliceOp(const Tensor &operand, ArrayRef<int64_t> startIndices,
   Tensor result(resultType);
   for (auto resultIt = result.index_begin(); resultIt != result.index_end();
        ++resultIt) {
-    SmallVector<int64_t> operandIdx;
+    Index operandIdx(operand.getType().getRank());
     for (auto dim = 0; dim < operand.getType().getRank(); ++dim)
-      operandIdx.push_back(startIndices[dim] + (*resultIt)[dim] * strides[dim]);
+      operandIdx[dim] = startIndices[dim] + (*resultIt)[dim] * strides[dim];
```
burmako, line 283: Can we rewrite this as something like `auto operandIdx = startIndices + *resultIt * strides`?

stablehlo/reference/Index.cpp
```diff
@@ -20,17 +20,23 @@ limitations under the License.
 namespace mlir {
 namespace stablehlo {
 
-LogicalResult verifyIndex(llvm::ArrayRef<int64_t> shape,
-                          llvm::ArrayRef<int64_t> index) {
+Index Index::operator+(const Index &other) const {
+  SmallVector<int64_t> combined;
```
burmako, line 24: `result`?

stablehlo/reference/Index.cpp
```diff
@@ -20,17 +20,23 @@ limitations under the License.
...
+  return Index(combined);
+}
+
+LogicalResult verifyIndex(llvm::ArrayRef<int64_t> shape, const Index &index) {
   if (shape.size() != index.size()) return failure();
 
-  for (auto [shapeDim, indexDim] : llvm::zip(shape, index))
-    if (indexDim < 0 || indexDim >= shapeDim) return failure();
+  for (auto [dim, shapeDim] : llvm::enumerate(shape))
+    if (index[dim] < 0 || index[dim] >= shapeDim) return failure();
```
burmako, line 34: Can we make it possible to keep using `zip` here?

stablehlo/reference/Index.h
```diff
@@ -54,13 +86,13 @@ class IndexSpaceIterator {
...
   /// Dereferencing a past-the-end iterator will result in a fatal error.
-  llvm::ArrayRef<int64_t> operator*() const;
+  Index operator*() const;
 
   /// Compare the iterator to another iterator.
   /// Two iterators are equal if they have the same underlying shape and
   /// reference the same element in the index space.
   bool operator==(const IndexSpaceIterator &it) {
-    return shape_ == it.shape_ && index_ == it.index_;
+    return (shape_ == it.shape_) && (index_ == it.index_);
```
burmako, line 95: Why do we need parentheses here?

stablehlo/reference/Ops.cpp
```diff
@@ -110,14 +103,16 @@ Tensor evalCosineOp(const Tensor &operand, Type resultType) {
...
   Tensor result(resultType);
-  SmallVector<int64_t> adjustedStartIndices;
+  Index adjustedStartIndices(startIndices.size());
   for (size_t i = 0; i < startIndices.size(); ++i)
-    adjustedStartIndices.push_back(std::min(
-        std::max(startIndices[i].get({}).getIntegerValue().getSExtValue(), 0l),
-        operand.getType().getShape()[i] - sliceSizes[i]));
+    adjustedStartIndices[i] = std::min(
+        std::max(
+            startIndices[i].get(Index({})).getIntegerValue().getSExtValue(),
```
burmako, line 110: It's unfortunate that we need additional boilerplate in the form of `Index({})`. Let's keep the `ArrayRef<int64_t>` overload as well?

burmako (changes requested):

stablehlo/reference/Index.h
sdasgup3: No not needed. Removed.

stablehlo/reference/Index.cpp
sdasgup3: Added the following to achieve this.
```
  /// Get the index array.
  ArrayRef<int64_t> getIndexArray() const { return index_; }
```

stablehlo/reference/Ops.cpp
```diff
@@ -265,10 +246,8 @@ Tensor evalPadOp(const Tensor &operand, const Tensor &paddingValue,
     result.set(*resultIt, paddingValue.get({}));
   for (auto operandIt = operand.index_begin(); operandIt != operand.index_end();
        ++operandIt) {
-    SmallVector<int64_t> resultIdx(result.getType().getRank());
-    for (auto i = 0; i < operand.getType().getRank(); ++i)
-      resultIdx[i] =
-          edgePaddingLow[i] + (*operandIt)[i] * (interiorPadding[i] + 1);
+    Index resultIdx =
+        edgePaddingLow + *operandIt * interiorPadding + *operandIt;
```
sdasgup3, line 250: If use the overload`Index::operator+(int64_t n)`, we can achieve the following.

```
 Index resultIdx =
        edgePaddingLow + (*operandIt) * (Index(interiorPadding) + 1);
```
I prefer the implemented one with distributed `*operandIt` over explicit `Index` casts. Please let me know what you think?

stablehlo/reference/Index.cpp
```diff
@@ -20,17 +20,56 @@ limitations under the License.
...
+}
+
+Index Index::operator+(ArrayRef<int64_t> array) const {
+  if (size() != array.size())
+    llvm::report_fatal_error("Index add expects operands of same size.");
+
+  SmallVector<int64_t> result;
+  for (auto [lhsIdx, rhsIdx] : llvm::zip(index_, array))
+    result.push_back(lhsIdx + rhsIdx);
+  return Index(result);
```
burmako, line 40: This can be deduplicated to `return *this + Index(array)`. Same below for `operator*`.

stablehlo/reference/Ops.cpp
burmako: The current code looks good to me. 

For what it's worth, I think it would be possible to enable `... * (interiorPadding + 1)` by replacing uses of `ArrayRef<int64_t>` in Ops.cpp with domain-specific types: 
  * BroadcastInDimOp::broadcastDimensions, ReverseOp::dimensions and TransposeOp::permutation would become `Axes`.
  * Various startIndices would become `Indices`.
  * Other ArrayRef<Int64_t> parameters would become `Sizes`.
  * We'd also introduce `Tensor::getShape` which would return `Sizes`.
  * We could also have a helper function that turns `ArrayRef<Tensor> startIndices` into `Indices`. 

Then we could define the appropriate operators and helpers. E.g. we shouldn't do anything with Axes, but we can add, multiply etc Sizes and Indices and mix them with each other. This way we'd be able to replace code like this:

```C++
Index adjustedStartIndices(startIndices.size());
for (size_t i = 0; i < startIndices.size(); ++i)
  adjustedStartIndices[i] = std::min(
      std::max(startIndices[i].get({}).getIntegerValue().getSExtValue(), 0l),
      operand.getType().getShape()[i] - sliceSizes[i]);
```

With code like this:

```C++
auto adjustedStartIndices = clamp(0, evalIndices(startIndices), operand.getShape() - sliceSizes);
```

burmako (changes requested):

stablehlo/reference/Ops.cpp
burmako: I couldn't resist and prototyped how this could look like: https://github.com/sdasgup3/stablehlo/pull/4.

burmako: The prototype is finished. Please review the PR to your branch, and then we could potentially merge it to make it part of this PR?

stablehlo/reference/Ops.cpp
```diff
@@ -289,7 +268,7 @@ Tensor evalReverseOp(const Tensor &operand, ArrayRef<int64_t> dimensions,
   auto resultShape = result.getType().getShape();
   for (auto resultIt = result.index_begin(); resultIt != result.index_end();
        ++resultIt) {
-    SmallVector<int64_t> operandIdx(*resultIt);
+    Index operandIdx(*resultIt);
     for (auto dim : dimensions)
       operandIdx[dim] = (resultShape[dim] - 1) - operandIdx[dim];
```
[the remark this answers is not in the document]
sdasgup3: done.

stablehlo/reference/Ops.cpp
sdasgup3: Thanks for the suggestion. Looks great to me. Merged and resolved the TODO to remove `Tensor(Type type)` constructor. 

Regarding the prototype:

>> E.g. we shouldn't do anything with Axes, but we can add, multiply etc Sizes and Indices and mix them with each other. 

Currently we are using `Sizes` in places like
```
... evalBroadcastInDimOp(.. {
   ...
Sizes operandIdx(operandShape.size());
  ...
```

Shouldn't we be using an `Index` type in the following places to improve readability. 
1. in places like the above.
2. `evalIndices` to return `Index`.
3. `IndexSpaceIterator::operator*` and `IndexSpaceIterator::operator->` to return `Index`.
4. The private member of `IndexSpaceIterator` to be `std::optional<Index> index_`.

 `Index` could be an alias of `Sizes` (i.e. `using Index = Sizes;`).

What do you think?
@burmako

stablehlo/reference/Ops.cpp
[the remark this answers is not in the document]
burmako: Actually I was wrong - we cannot use helper methods here because this logic is only meant to affect a subset of axes. Our tests didn't catch this because we only have one test which reverses all axes :)

stablehlo/reference/Tensor.h
```diff
@@ -70,8 +70,6 @@ class Tensor {
   /// \name Constructors
   /// @{
   Tensor();
-  // TODO: Remove the need in having this constructor.
-  explicit Tensor(Type type) : Tensor(type.cast<RankedTensorType>()) {}
```
burmako: Hmm that's a lot of casts. How about we replace `RankedTensorType` with `TensorType` for now? This way, we will be able to remove all the casts.

stablehlo/reference/Ops.cpp
burmako: Yeah, I also don't love the current situation. This is why I initially had `Index` and `Sizes` as two different types (you can see in commit history), but that was a lot of boilerplate.

Eventually, I kind of convinced myself to follow LLVM's (and C++'s) convention to use `size_t` to represent both sizes and indices, which generalizes to `Sizes` and sizes/indices in StableHLO, so I merged `Index` and `Sizes` into just `Sizes`.

Let's give `use Index = Sizes;` a try and see how well this works out. I'd also add "6) Signatures of evalFooOp functions where appropriate" as another place to use `Index`.

burmako (changes requested):

stablehlo/reference/Ops.cpp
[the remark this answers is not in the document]
sdasgup3: That is right. I missed that.
Reverted that change.

stablehlo/reference/Ops.cpp
sdasgup3: PTAL.
Added one more
7. Called `evalIndices` to prepare an `Index` to be passed into the following call to `evalDynamicSliceOp` or `evalDynamicUpdateSliceOp`. This allows  those `evalFoOp` to have `Index startIndices` as formal arguments which is more readable.

stablehlo/reference/Ops.cpp
```diff
@@ -221,82 +198,82 @@ Tensor evalIotaOp(int64_t iotaDimension, Type resultType) {
...
-                     Type resultType) {
+Tensor evalReverseOp(const Tensor &operand, Axes dimensions,
+                     TensorType resultType) {
   Tensor result(resultType);
-  auto resultShape = result.getType().getShape();
+  auto resultShape = result.getShape();
   for (auto resultIt = result.index_begin(); resultIt != result.index_end();
        ++resultIt) {
-    SmallVector<int64_t> operandIdx(*resultIt);
+    Sizes operandIdx(*resultIt);
```
burmako, line 276: `Index`?

burmako (changes requested):

stablehlo/reference/Ops.cpp
```diff
@@ -456,18 +430,18 @@ SmallVector<Tensor> eval(Region &region, ArrayRef<Tensor> args, Scope *parent) {
       scope.add(op.getResults(), {runtimeResult});
     } else if (auto dynamicSliceOp = dyn_cast<DynamicSliceOp>(op)) {
       Tensor runtimeOperand = scope.find(dynamicSliceOp.getOperand());
-      auto runtimeStartIndices = scope.find(dynamicSliceOp.getStartIndices());
-      auto runtimeSliceSizes =
-          llvm::to_vector(dynamicSliceOp.getSliceSizes().getValues<int64_t>());
+      Index runtimeStartIndices =
+          evalIndices(scope.find(dynamicSliceOp.getStartIndices()));
```
burmako, line 434: I noticed that this changes the API of `evalDynamicSliceOp` and `evalDynamicUpdateSliceOp`. In the StableHLO opset, the corresponding ops are taking tensors not tensor constants, so how about we stick to the existing API which also takes tensors and call evalIndices inside the corresponding functions?

burmako (approved):
One final comment, and we're good to go.

stablehlo/reference/Ops.cpp
sdasgup3: Sure.

PR #1176, opened 2023-02-16
Remove support for complex types in various ops

### LINKED ISSUE #560: Removing complex type supported for Min/Max ops

### LINKED ISSUE #997: Remove Complex type from RemainderOp
Explore the removal of complex type from the supported types of `RemainderOp` as it is not supported in XLA and in JAX.

ghpvnist:
This PR removes support for complex element types for ClampOp, MaxOp, MinOp, and RemainderOp.

closes #560 
closes #997

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -1481,23 +1481,19 @@ value and produces a `result` tensor. More formally, `result[i0, ..., iR-1]` =
 where `min_val = rank(min) == 0 ? min : min[i0, ..., iR-1]`,
 `max_val = rank(max) == 0 ? max : max[i0, ..., iR-1]`.
 
-Imposing an ordering on complex numbers involves surprising semantics,
-so in the future we are planning to remove support for complex numbers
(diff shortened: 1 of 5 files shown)


burmako:
Closing this for now to prioritize on reference implementation work (P0 workstream) over specification compliance / new features work (P1 workstreams).

PR #1178, opened 2023-02-16
Allow zero dimensional slice_size in DynamicSliceOp

subhankarshah:
1 dimensional tensors with 1 element are often represented as scalar tensors (0 dimensional), to support this kind of lowering allow 0 dimensional tensor as slice_size.

  * 0-dimensional slice_size `tensor<i64>` is treated same as `tensor<1xi64>`.
  * Updated constraint `dynamic_slice_i3` combined with `dynamic_slice_c2` should fix this.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -2316,11 +2316,11 @@ More formally, `result[i0, ..., iR-1] = operand[j0, ..., jR-1]` where:
 
 #### Inputs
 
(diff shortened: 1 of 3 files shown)


burmako:
Hi @subhankarshah! Thank you for your contribution. Can you talk a bit more about the motivation for this change?

PR #1180, opened 2023-02-17
Add support for collective ops in --stablehlo-refine-shapes

burmako:
At the moment, AllGatherOp and ReduceScatterOp don't have shape functions because they rely on `dim(process_groups, 1)` which in its full generality requires knowledge of `num_partitions`.

(Per https://github.com/openxla/stablehlo/blob/main/docs/spec.md: For cross_replica and flattened_ids process grouping strategies, `dim(process_groups, 1)` is equal to `dim(replica_groups, 1)`. However, for cross_replica_and_partition `dim(process_groups, 1)` is equal to `dim(replica_groups, 1) * num_partitions`).

Even though we cannot have shape functions for these two ops without changing the StableHLO spec, we can still make best effort at refining their shapes for cross_replica and flattened_ids cases. This is what this PR implements.

### DIFF
--- a/stablehlo/transforms/StablehloRefineShapes.cpp
+++ b/stablehlo/transforms/StablehloRefineShapes.cpp
@@ -526,6 +526,28 @@ LogicalResult refineReturnShape(PatternRewriter& rewriter, OpType op,
   return refineReturnShape(rewriter, op, shape);
 }
 
+struct RefineAllGatherOpPattern : public OpRewritePattern<AllGatherOp> {
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AllGatherOp op,
+                                PatternRewriter& rewriter) const override {
+    auto operandType = op.getOperand().getType().cast<ShapedType>();
+    if (!operandType.hasRank())
+      return rewriter.notifyMatchFailure(op, "expected ranked operand type");
+
+    if (op.getChannelHandle() && !op.getUseGlobalDeviceIds())
+      return rewriter.notifyMatchFailure(op, "unsupported strategy");
+    DenseIntElementsAttr replicaGroups = op.getReplicaGroups();
+    if (!replicaGroups || replicaGroups.getType().getRank() != 2)
+      return rewriter.notifyMatchFailure(op, "unsupported replica_groups");
+    auto shardCount = replicaGroups.getType().getDimSize(1);
+
+    SmallVector<int64_t> refinement(operandType.getShape());
+    if (!operandType.isDynamicDim(op.getAllGatherDim()))
+      refinement[op.getAllGatherDim()] *= shardCount;
+    return refineReturnShape(rewriter, op, refinement);
+  }
+};
+
 struct RefineBitcastConvertOpPattern
     : public OpRewritePattern<BitcastConvertOp> {
   using OpRewritePattern::OpRewritePattern;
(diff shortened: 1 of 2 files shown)


stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -526,6 +526,28 @@ LogicalResult refineReturnShape(PatternRewriter& rewriter, OpType op,
...
+                                PatternRewriter& rewriter) const override {
+    auto operandType = op.getOperand().getType().cast<ShapedType>();
+    if (!operandType.hasRank())
+      return rewriter.notifyMatchFailure(op, "expected ranked operand type");
+
+    if (op.getChannelHandle() && !op.getUseGlobalDeviceIds())
+      return rewriter.notifyMatchFailure(op, "unsupported strategy");
+    DenseIntElementsAttr replicaGroups = op.getReplicaGroups();
+    if (!replicaGroups || replicaGroups.getType().getRank() != 2)
+      return rewriter.notifyMatchFailure(op, "unsupported replica_groups");
```
GleasonK, line 541: Not many of the eval functions have comments.. but I don't understand these constraints fully, so comments may be useful for maintenance.

I.e. Why is channel handle with global IDs supported?

GleasonK: https://github.com/GleasonK/stablehlo/blob/main/docs/spec.md#all_gather

Found it in spec. Will review shortly.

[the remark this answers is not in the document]
GleasonK: Both might be verified? Can replica groups be empty?

stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -526,6 +526,28 @@ LogicalResult refineReturnShape(PatternRewriter& rewriter, OpType op,
...
 
+struct RefineAllGatherOpPattern : public OpRewritePattern<AllGatherOp> {
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AllGatherOp op,
+                                PatternRewriter& rewriter) const override {
+    auto operandType = op.getOperand().getType().cast<ShapedType>();
+    if (!operandType.hasRank())
+      return rewriter.notifyMatchFailure(op, "expected ranked operand type");
+
+    if (op.getChannelHandle() && !op.getUseGlobalDeviceIds())
```
GleasonK, line 537: A comment like "This combination of attributes requires knowing `num_partitions` and cannot be optimized". Would be helpful in capturing the rationale of this change. I find myself flipping between spec and PR description several times to understand this code.

stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -811,6 +833,28 @@ struct RefineRealDynamicSliceOpPattern
...
+      return rewriter.notifyMatchFailure(op, "unsupported strategy");
+    DenseIntElementsAttr replicaGroups = op.getReplicaGroups();
+    if (!replicaGroups || replicaGroups.getType().getRank() != 2)
+      return rewriter.notifyMatchFailure(op, "unsupported replica_groups");
```
GleasonK, line 848: Same question regarding verifiers

GleasonK (approved):

stablehlo/transforms/StablehloRefineShapes.cpp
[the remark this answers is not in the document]
burmako: You're right - I got my wires crossed. Somehow I got an impression that we're still allowing empty replica groups, but we no longer do that.

stablehlo/transforms/StablehloRefineShapes.cpp
burmako: Added a comment. Thank you for the feedback!

PR #1189, opened 2023-02-19
Audit VhloTypes.td

burmako:
This pull request completes the audit of VhloTypes.td, including refactorings aimed to align the file with the spec.
  1) Order alphabetically, merging native and forked types.
     (Within the vision that types must correspond 1:1 to the
     specification, there's no difference between where these types
     are originally coming from, so there's no need to separate the
     two categories).
  2) Introduce BooleanType to match the spec.
  3) Rename FunctionType::results to FunctionType::outputs to match
     the spec.
  4) Rename IntegerI...Type to IntegerSI...Type to match the spec.
  5) Rename floating-point types to FloatBF...Type or FloatF...Type
     to match the naming style used for integer types.
  6) Add comments to each attribute, either identifying the
     corresponding program element from the spec or linking a ticket
     that tracks the alignment effort.
  7) Rename kFooType to kFooV1Type to be consistent with the TableGen
     file that defines the attributes.
  8) Renumber kFooAttr in increasing order to match the alphabetical
     order of the names. (This is a breaking change for VHLO, but I
     think it's fine because we haven't officially released it yet).
  9) A few more minor renamings and cleanups.

### DIFF
--- a/stablehlo/dialect/VhloBytecode.cpp
+++ b/stablehlo/dialect/VhloBytecode.cpp
@@ -240,105 +240,112 @@ enum AttributeCode {
 enum TypeCode {
   // TO ADD TYPE: Add an enum value with doc string for new type.
 
-  ///   TokenType {
+  ///   BooleanV1Type {
   ///   }
-  kTokenType = 0,
+  kBooleanV1Type = 0,
 
-  ///   BFloat16Type {
-  ///   }
-  kBFloat16Type = 1,
-
-  ///   ComplexType {
+  ///   ComplexV1Type {
   ///     elementType: Type
   ///   }
-  kComplexType = 2,
+  kComplexV1Type = 1,
 
-  ///   Float16Type {
+  ///   FloatBF16V1Type {
   ///   }
-  kFloat16Type = 3,
+  kFloatBF16V1Type = 2,
 
-  ///   Float32Type {
+  ///   FloatF16V1Type {
   ///   }
-  kFloat32Type = 4,
+  kFloatF16V1Type = 3,
 
-  ///   Float64Type {
+  ///   FloatF32V1Type {
   ///   }
-  kFloat64Type = 5,
+  kFloatF32V1Type = 4,
 
-  ///   FunctionType {
+  ///   FloatF64V1Type {
+  ///   }
+  kFloatF64V1Type = 5,
+
+  ///   FloatF8E4M3FNV1Type {
+  ///   }
+  kFloatF8E4M3FNV1Type = 6,
+
+  ///   FloatF8E5M2V1Type {
+  ///   }
+  kFloatF8E5M2V1Type = 7,
+
+  ///   FunctionV1Type {
   ///     inputs: Type[]
-  ///     results: Type[]
+  ///     outputs: Type[]
   ///   }
-  kFunctionType = 6,
+  kFunctionV1Type = 8,
 
-  ///   IntegerI1Type {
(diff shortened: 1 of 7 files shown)


GleasonK:
> 8. ... (This is a breaking change for VHLO, but I think it's fine because we haven't officially released it yet).

Agree. Breaking changes are still ok for now.

stablehlo/dialect/VhloBytecode.cpp
```diff
@@ -1059,12 +1054,12 @@ void VhloBytecodeInterface::write(DictionaryV1Attr attr,
...
-  if (type.isa<Float16V1Type>()) return APFloat::IEEEhalf();
-  if (type.isa<Float32V1Type>()) return APFloat::IEEEsingle();
-  if (type.isa<Float64V1Type>()) return APFloat::IEEEdouble();
+  if (type.isa<FloatBF16V1Type>()) return APFloat::BFloat();
+  if (type.isa<FloatF16V1Type>()) return APFloat::IEEEhalf();
+  if (type.isa<FloatF32V1Type>()) return APFloat::IEEEsingle();
+  if (type.isa<FloatF64V1Type>()) return APFloat::IEEEdouble();
+  if (type.isa<FloatF8E4M3FNV1Type>()) return APFloat::Float8E4M3FN();
+  if (type.isa<FloatF8E5M2V1Type>()) return APFloat::Float8E5M2();
   llvm_unreachable("non-floating point type used");
```
GleasonK, line 1063: Should we make this `report_fatal_error`? Just noticing this.

stablehlo/dialect/VhloBytecode.cpp
```diff
@@ -1111,12 +1106,13 @@ void VhloBytecodeInterface::write(FlatSymbolRefV1Attr attr,
...
-  if (type.isa<IntegerI32V1Type>() || type.isa<IntegerUI32V1Type>()) return 32;
-  if (type.isa<IntegerI64V1Type>() || type.isa<IntegerUI64V1Type>()) return 64;
+  // TODO: Remove BooleanV1Type from here once/if BooleanAttr lands.
+  if (type.isa<BooleanV1Type>()) return 1;
+  if (type.isa<IntegerSI4V1Type>() || type.isa<IntegerUI4V1Type>()) return 4;
+  if (type.isa<IntegerSI8V1Type>() || type.isa<IntegerUI8V1Type>()) return 8;
+  if (type.isa<IntegerSI16V1Type>() || type.isa<IntegerUI16V1Type>()) return 16;
+  if (type.isa<IntegerSI32V1Type>() || type.isa<IntegerUI32V1Type>()) return 32;
+  if (type.isa<IntegerSI64V1Type>() || type.isa<IntegerUI64V1Type>()) return 64;
   llvm_unreachable("unsupported integer type used in IntegerV1Attr");
```
GleasonK, line 1116: Same question

GleasonK (approved):
I have a feeling you've memorized the spec word-for-word after all these PRs. This looks great!

stablehlo/dialect/VhloBytecode.cpp
burmako: Yes, this has been addressed in https://github.com/openxla/stablehlo/pull/1185.

stablehlo/dialect/VhloBytecode.cpp
burmako: Same answer :)

Issue #1047, opened 2023-02-03
Consider aligning VhloAttrs.td with the spec


burmako:
Having completed another round of reviews of #849, I was thinking about the overall design of VHLO. We've been working on it since October - it went through several iterations of design, and several big pull requests - so it felt like it's a good time to reflect.

Even though it's a lot of code, I'm quite pleased with how things are going. Our high-level design is fairly approachable - we've got a dedicated dialect, three crisp passes that operate on it and the rest is implementation details. E.g. if needed, each pass can be rewritten from scratch with little effect on the rest of the system.

Then I started thinking about the VHLO dialect aka our serialization schema:
  * Ops look fine - in some areas, the opset is all over the place (especially around the Dynamism and Not in HLO categories), which creates essential complexity of the domain, so it's fine that things are messy right now. We have tickets to clean them up.
  * Types look fine too - in 849, they mostly correspond 1:1 to the spec (except for encodings, index and quantization, but we have tickets to clean all that up as well through dynamism, quantization and sparsity work).
  * However, I am somewhat concerned about attrs. In 849, they represent a mixture of custom attributes and forked upstream attributes, and I find it rather hard to wrap my head around that.

Just like with types, I think that we should aim at 1:1 correspondence with the spec. Concretely, towards that end I'd like to propose to:
  1. Get rid of all `VHLO_ConvDimensionNumbersAttr` and friends. I think that grouping of the elementary attrs into bigger attrs is pretty incidental as far as serialization format goes, so I would propose to break all of them down, just like the spec does.
  2. Rename `VHLO_DenseIntOrFPElementsAttr` to `VHLO_TensorAttr` to be aligned with the spec. Keeping the forked name is understandable, but I think that alignment with the spec is more important.
  3. Drop `VHLO_FlatSymbolRefAttr` and use `VHLO_StringAttr` instead.
  4. Drop `VHLO_UnitAttr` and use `VHLO_IntegerAttr` instead.

GleasonK:
Also consider scoping 
https://github.com/openxla/stablehlo/pull/849/files/0ccd3220810030d7e968fcf5b2cd419d009c10fd#r1096258167

PR #1190, opened 2023-02-19
Rename VHLO_DenseIntOrFPElementsAttr to VHLO_TensorAttr

burmako:
This is one of the several steps towards aligning VhloAttrs.td with the StableHLO spec.

Keeping the forked name is understandable, but I think that alignment with the spec is more important. See #1047.

### DIFF
--- a/stablehlo/dialect/VhloBytecode.cpp
+++ b/stablehlo/dialect/VhloBytecode.cpp
@@ -186,12 +186,6 @@ enum AttributeCode {
   ///   }
   kArrayAttr = 16,
 
-  ///   DenseIntOrFPElementsAttr {
-  ///     type: Type
-  ///     data: blob
-  ///   }
-  kDenseIntOrFPElementsAttr = 17,
-
   ///   ArrayAttr {
   ///   DictionaryAttr {
   ///     attrs: <Attribute, Attribute>[]
@@ -219,6 +213,12 @@ enum AttributeCode {
   ///     value: string
   ///   }
   kStringAttr = 22,
+
+  ///   TensorAttr {
+  ///     type: Type
+  ///     data: blob
+  ///   }
+  kTensorAttr = 17,
 
   ///   TypeAttr {
   ///     value: Type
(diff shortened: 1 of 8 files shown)


stablehlo/dialect/VhloBytecode.cpp
```diff
@@ -220,6 +214,12 @@ enum AttributeCode {
   ///   }
   kStringAttr = 22,
 
+  ///   TensorAttr {
+  ///     type: Type
+  ///     data: blob
+  ///   }
+  kTensorAttr = 17,
```
GleasonK, line 221: Merge the other bytecode modifications then lets renumber since we can break still. Could open an issue to follow up.

stablehlo/dialect/VhloBytecode.cpp
```diff
@@ -597,13 +595,13 @@ LogicalResult VhloBytecodeInterface::writeAttribute(
...
             return success();
           })
-      .Case<ArrayV1Attr, DenseIntOrFPElementsV1Attr, DictionaryV1Attr,
-            FlatSymbolRefV1Attr, FloatV1Attr, IntegerV1Attr, StringV1Attr,
-            TypeV1Attr>([&](auto attr) {
-        LOG_WRITE_CALL;  // Forked attrs
-        write(attr, writer);
-        return success();
-      })
+      .Case<ArrayV1Attr, DictionaryV1Attr, FlatSymbolRefV1Attr, FloatV1Attr,
```
GleasonK, line 598: This split for `.Case` above and `.Case` here is that some are upstream and some are StableHLO. We can merge these two case statements since this distinction no longer matters (maybe this was done in a different PR)

GleasonK (approved):

stablehlo/dialect/VhloOps.cpp
```diff
@@ -178,23 +178,23 @@ ParseResult parseFunctionBody(OpAsmParser& parser, Attribute& name,
   return success();
 }
 
-void DenseIntOrFPElementsV1Attr::print(mlir::AsmPrinter& p) const {
```
GleasonK, line 181: This file could use some sorting eventually, but nothing blocking.

stablehlo/dialect/VhloBytecode.cpp
burmako: Yes, this is done in the final PR of this chain ("Audit StablehloAttrs.td").

stablehlo/dialect/VhloBytecode.cpp
burmako: Yeah, same here :)

stablehlo/dialect/VhloOps.cpp
burmako: Agreed! I was planning to finish externalizing as much of VhloOps.cpp as possible first (all verifiers and shape functions have been moved, then I wanted to move reifyReturnTypeShapes and prettyprinters/parsers), and then alphabetically sort everything.

PR #1192, opened 2023-02-19
Remove VHLO_FlatSymbolRefAttr

burmako:
This is one of the several steps towards aligning VhloAttrs.td with
the StableHLO spec.

As far as representational power goes, FlatSymbolRefAttr is equivalent
to SymbolAttr, so in the spirit of VHLO being a shallow dialect, I think
that we should remove the former and use the latter instead.


stablehlo/transforms/StablehloLegalizeToVhlo.cpp
```diff
@@ -247,18 +267,30 @@ class StablehloToVhloOpConverter : public OpConversionPattern<StablehloOpTy> {
...
+        }
+        if (stablehloAttr.getName() == "called_computations") {
+          vhloAttr = convertCustomCallCalledComputations(
+              stablehloAttr.getValue(), this->getTypeConverter());
+          if (!vhloAttr) return failure();
+        }
+      } else if constexpr (std::is_same<StablehloOpTy, func::CallOp>::value) {
+        if (stablehloAttr.getName() == "callee") {
+          vhloAttr = convertFuncCallee(stablehloAttr.getValue(),
+                                       this->getTypeConverter());
```
GleasonK, line 285: For the StableHLO --> VHLO direction we don't need to be this precise and can map all FlatStymbolRefAttrs to StringAttr I believe, the other direction will require this though.

GleasonK (approved):

stablehlo/transforms/StablehloLegalizeToVhlo.cpp
GleasonK: Probably safe to keep these in sync actually. We will only support flat symbol refs in these ops which is probably best.

GleasonK:
> As far as representational power goes, FlatSymbolRefAttr is equivalent
to SymbolAttr

SymbolAttr or StringAttr? Should we add a SymbolAttr?

The spec for custom call says:

> variadic number of functions

Should this say number of constant `strings` instead? I'm wondering if it is useful to keep a distinction between 'string' and 'string that refers to a function name' in StableHLO spec / VHLO. We could make both stored as a string with a different mnemonic.

stablehlo/dialect/VhloBytecode.cpp
```diff
@@ -192,11 +192,6 @@ enum AttributeCode {
   ///   }
   kDictionaryAttr = 18,
 
-  ///   FlatSymbolRefAttr {
-  ///     rootReference: StringAttr
-  ///   }
-  kFlatSymbolRefAttr = 19,
```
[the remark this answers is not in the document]
burmako: Done in the top PR of this PR chain.

burmako:
"SymbolAttr or StringAttr? Should we add a SymbolAttr?". Oops, I meant StringAttr. I'll fix the wording.

"variadic number of functions => variadic number of strings". Great idea! Indeed, this conflates the notions of functions and references to functions, and I agree that we should say "variadic number of strings" in the spec. Will update that in this PR.

stablehlo/transforms/StablehloLegalizeToVhlo.cpp
burmako: Alright, will keep the code unchanged then.

PR #1193, opened 2023-02-19
Replace VHLO_UnitAttr with VHLO_BooleanAttr

burmako:
This is one of the several steps towards aligning VhloAttrs.td with
the StableHLO spec.

As far as representational power goes, UnitAttr is equivalent
to BooleanAttr, so in the spirit of VHLO being a shallow dialect, I
think that we should remove the former and use the latter instead.

When making this change, I noticed that BooleanAttr attributes from
StableHLO get translated to IntegerAttr attributes from VHLO. This has
been changed to use the newly introduced BooleanAttr from VHLO.

### DIFF
--- a/stablehlo/transforms/StablehloLegalizeToVhlo.cpp
+++ b/stablehlo/transforms/StablehloLegalizeToVhlo.cpp
@@ -184,11 +184,16 @@ Attribute convertAttrToVhlo(Attribute stablehloAttr,
     return vhlo::FloatV1Attr::get(attr.getContext(), vhloFloatType,
                                   attr.getValue());
   }
-  if (auto attr = stablehloAttr.dyn_cast<IntegerAttr>()) {
-    auto vhloIntegerType = typeConverter->convertType(attr.getType());
-    if (!vhloIntegerType) return {};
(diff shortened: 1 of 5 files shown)


stablehlo/transforms/StablehloLegalizeToVhlo.cpp
```diff
@@ -268,8 +270,10 @@ class StablehloToVhloOpConverter : public OpConversionPattern<StablehloOpTy> {
     SmallVector<NamedAttribute> vhloAttrs;
     for (NamedAttribute stablehloAttr : stablehloOp->getAttrs()) {
       Attribute vhloAttr;
-      if constexpr (std::is_same<StablehloOpTy,
-                                 stablehlo::CustomCallOp>::value) {
+      if (stablehloAttr.getName() == "use_global_device_ids") {
```
GleasonK, line 273: Do we want to check if it is a UnitAttr?

GleasonK (changes requested):

stablehlo/transforms/StablehloLegalizeToVhlo.cpp
burmako: Done.

PR #1194, opened 2023-02-19
Inline VHLO_ConvDimensionNumbersAttr and friends

burmako:
I think that grouping of the elementary attrs into bigger attrs is
pretty incidental as far as serialization format goes, so I would
propose to break all of them down, just like the spec does.

### DIFF
--- a/stablehlo/transforms/VhloLegalizeToStablehlo.cpp
+++ b/stablehlo/transforms/VhloLegalizeToStablehlo.cpp
@@ -69,41 +69,17 @@ class VhloToStablehloTypeConverter : public vhlo::VhloTypeConverter {
 Attribute convertAttrToStablehlo(Attribute vhloAttr,
                                  TypeConverter* typeConverter) {
   LLVM_DEBUG(llvm::dbgs() << "Converting attr " << vhloAttr);
-  if (auto attr = vhloAttr.dyn_cast<vhlo::ChannelHandleV1Attr>()) {
-    return stablehlo::ChannelHandleAttr::get(attr.getContext(),
-                                             attr.getHandle(), attr.getType());
-  }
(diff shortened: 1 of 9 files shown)


stablehlo/transforms/VhloLegalizeToStablehlo.cpp
```diff
@@ -284,6 +441,51 @@ class VhloToStablehloOpConverter : public OpConversionPattern<VhloOpTy> {
               vhloAttr.getValue(), this->getTypeConverter());
           if (!stablehloAttr) return failure();
         }
+      } else if constexpr (std::is_same<VhloOpTy,
+                                        vhlo::DotGeneralOpV1>::value) {
```
GleasonK, line 445: As this list gets longer we may want to consider using a template function if it help organization at all `convertOpAttr<VhloOpTy>(attr)` or something similar.

GleasonK: Could also consider merging the above conversions code into this loop as well. For each op take the alphabetically first attribute name and generate the (conv dim attr / channel handle / etc), and skip for the other names. Might make it more clear why we are continuing without adding for these attrs.

GleasonK (approved):

stablehlo/transforms/VhloLegalizeToStablehlo.cpp
burmako: I was also thinking that it might be nice to have dedicated conversion patterns for these ops. We could keep the current generic pattern for most of the ops and then specialize for particularly intense cases. I think it might even be less code overall.

Given that you've already approved this PR, let's leave figuring out to a later point. I'll resolve this conversation in the meanwhile.

Issue #584, opened 2022-11-23
Consider removing compare_type from CompareOp


burmako:
This is how things have been working [in HLO](https://github.com/tensorflow/tensorflow/blob/3451f5d39ea8add4c254acc9615d62ff589259b8/tensorflow/compiler/xla/comparison_util.h#L71-L79) since recently.

GleasonK:
This op is used in production, and serialized for on-device consumption. As such (And given that the HLO representation is entirely different than this single enum), this op will be changed on an as-needed basis via RFC.

PR #1199, opened 2023-02-22
Add interpreter for CompareOp

### LINKED ISSUE #967: Add interpreter for CompareOp

ghpvnist:

```
(I1) lhs tensor.
(I2) rhs tensor.
(I3) comparison_direction enum of `EQ`, `NE`, `GE`, `GT`, `LE`, or `LT`.
(I4) compare_type enum of `FLOAT`, `TOTALORDER`, `SIGNED`, or `UNSIGNED`.
(C1) `lhs` and `rhs` have the same element type.
(C2) `lhs`, `rhs`, and `result` have the same shape.
(C3) Given `E` is the `lhs` element type, the following are legal values of
     `compare_type`:
     * If `E` is signed integer type, `compare_type` = `SIGNED`.
     * If `E` is unsigned integer or boolean type, `compare_type` = `UNSIGNED`.
     * If `E` is floating-point type,
       `compare_type` $\in$ {`FLOAT`, `TOTALORDER`}.
     * If `E` is complex type, `compare_type` = `FLOAT`.
```

```
I1: a) lhs is not a tensor. (Covered by ODS).
I2: a) rhs is not a tensor. (Covered by ODS).
I3: a) comparison_direction is not enum of `EQ`, `NE`, `GE`, `GT`, `LE`, or `LT`. (Covered by ODS).
I4: a) compare_type is not enum of `FLOAT`, `TOTALORDER`, `SIGNED`, or `UNSIGNED`. (Covered by ODS).
C1: a) element_type(lhs) != element_type(rhs). (Covered by the ODS).
C2: a) lhs, rhs, and result do not have the same shape. (Covered by the ODS).
C3: a) `E` is signed integer type and `compare_type` != `SIGNED`.
    b) `E` is unsigned integer type and `compare_type` != `UNSIGNED`.
    c) `E` is boolean type and `compare_type` != `UNSIGNED`.
    d) `E` is floating-point type and `compare_type` $\notin$ {`FLOAT`, `TOTALORDER`}.
    e) `E` is complex type and `compare_type` != `FLOAT`.
```

```
C3a: `E` is signed integer type and `compare_type` != `SIGNED`.
C3b: `E` is unsigned integer type and `compare_type` != `UNSIGNED`.
C3c: `E` is boolean type and `compare_type` != `UNSIGNED`.
C3d: `E` is floating-point type and `compare_type` $\notin$ {`FLOAT`, `TOTALORDER`}.
C3e: `E` is complex type and `compare_type` != `FLOAT`.
```

Notes:
* However, due to the issue noted in #1071 and given the higher priority work of reference implementation, this spec compliance work (i.e. verifications for (C3)) will be deferred.
* Updated the spec wording to be clear about special behaviors.
* Also reorders Element.cpp and Element.h `operatorFoo` functions to be in lexicographical order.
* We plan to drop complex element type support for comparison directions {GE, GT, LE, LT} in #560, so these were not implemented.

closes #967


stablehlo/tests/interpret_compare.mlir
```diff
@@ -0,0 +1,395 @@
...
+}
+
+// -----
+
+func.func @compare() {
+  // -NaN, -NaN, -Inf, -Inf, -2.0, -2.0, -0.0, -0.0, +0.0, 1.0, 2.0, +Inf, +NaN, +minNaN
+  // -NaN, +NaN, -Inf, +Inf, -2.0, -1.0, -0.0, +0.0, +0.0, 2.0, 2.0, +Inf, +NaN, +maxNaN
+  %lhs = stablehlo.constant dense<[0xFFF0000000000001, 0xFFF0000000000001, 0xFFF0000000000000, 0xFFF0000000000000, -2.0, -2.0, 0x8000000000000000, 0x8000000000000000, 0x0000000000000000, 1.0, 2.0, 0x7FF0000000000000, 0x7FF0000000000001, 0x7FF0000000000001]> : tensor<14xf64>
+  %rhs = stablehlo.constant dense<[0xFFF0000000000001, 0x7FF0000000000001, 0xFFF0000000000000, 0x7FF0000000000000, -2.0, -1.0, 0x8000000000000000, 0x0000000000000000, 0x0000000000000000, 2.0, 2.0, 0x7FF0000000000000, 0x7FF0000000000001, 0x7FFFFFFFFFFFFFFF]> : tensor<14xf64>
+  %result = stablehlo.compare GE, %lhs, %rhs, TOTALORDER : (tensor<14xf64>, tensor<14xf64>) -> tensor<14xi1>
```
sdasgup3, line 356: lhs[5]: -2.0 or 0xC000000000000000 or 13835058055282163712 (unsigned)
rhs[5]: -1.0 or 0xBFF0000000000000 or 13830554455654793216 (unsigned)

`lhs >= rhs == true` right?

stablehlo/reference/Ops.cpp
```diff
@@ -94,6 +94,34 @@ Tensor evalClampOp(const Tensor &min, const Tensor &operand, const Tensor &max,
   return result;
 }
 
+Tensor evalCompareOp(const Tensor &lhs, const Tensor &rhs,
```
sdasgup3, line 97: IMO the implementation can be simplified a lot if we define operator overloads for `<=` , `<`, `>=` and `>` for reuse. 


Here is the suggestion
```
Tensor evalCompareOp(const Tensor &lhs, const Tensor &rhs,
                     ComparisonDirection comparisonDirection,
                     std::optional<ComparisonType> compareType,
                     TensorType resultType) {
  Tensor result(resultType);
  auto elementTy = result.getType().dyn_cast<TensorType>().getElementType();
  auto comparisonType = ComparisonType::NOTYPE;
  if (compareType && *compareType != ComparisonType::NOTYPE) {
    comparisonType = *compareType;
  } else if (isSupportedUnsignedIntegerType(elementTy) ||
             isSupportedBooleanType(elementTy)) {
    comparisonType = ComparisonType::UNSIGNED;
  } else if (isSupportedSignedIntegerType(elementTy)) {
    comparisonType = ComparisonType::SIGNED;
  } else if (isSupportedFloatType(elementTy) ||
             isSupportedComplexType(elementTy)) {
    comparisonType = ComparisonType::FLOAT;
  } else {
    report_fatal_error(invalidArgument("Unsupported element type: %s",
                                       debugString(elementTy).c_str()));
  }

  for (auto it = result.index_begin(); it != result.index_end(); ++it) {
    bool cmpResult = false;
    switch (comparisonDirection) {
      case ComparisonDirection::EQ: {
        if (comparisonType == ComparisonType::TOTALORDER)
          cmpResult = lhs.get(*it).getFloatValue().bitcastToAPInt() ==
                      rhs.get(*it).getFloatValue().bitcastToAPInt();
        else
          cmpResult = lhs.get(*it) == rhs.get(*it);
        break;
      }
      case ComparisonDirection::NE: {
        if (comparisonType == ComparisonType::TOTALORDER)
          cmpResult = lhs.get(*it).getFloatValue().bitcastToAPInt() !=
                      rhs.get(*it).getFloatValue().bitcastToAPInt();
        else
          cmpResult = lhs.get(*it) != rhs.get(*it);
        break;
      }
      case ComparisonDirection::GE: {
        if (comparisonType == ComparisonType::TOTALORDER) {
          auto lhsFloat = lhs.get(*it).getFloatValue();
          auto rhsFloat = rhs.get(*it).getFloatValue();
          if (lhsFloat.isNegative() ^ rhsFloat.isNegative())
            cmpResult = rhsFloat.isNegative();
          else
            cmpResult =
                lhsFloat.bitcastToAPInt().uge(rhsFloat.bitcastToAPInt());
        } else {
          cmpResult = lhs.get(*it) >= rhs.get(*it);
        }
        break;
      }
      case ComparisonDirection::GT: {
        if (comparisonType == ComparisonType::TOTALORDER) {
          auto lhsFloat = lhs.get(*it).getFloatValue();
          auto rhsFloat = rhs.get(*it).getFloatValue();
          if (lhsFloat.isNegative() ^ rhsFloat.isNegative())
            cmpResult = rhsFloat.isNegative();
          else
            cmpResult =
                lhsFloat.bitcastToAPInt().ugt(rhsFloat.bitcastToAPInt());
        } else {
          cmpResult = lhs.get(*it) > rhs.get(*it);
        }
        break;
      }
      case ComparisonDirection::LE: {
        if (comparisonType == ComparisonType::TOTALORDER) {
          auto lhsFloat = lhs.get(*it).getFloatValue();
          auto rhsFloat = rhs.get(*it).getFloatValue();
          if (lhsFloat.isNegative() ^ rhsFloat.isNegative())
            cmpResult = lhsFloat.isNegative();
          else
            cmpResult =
                lhsFloat.bitcastToAPInt().ule(rhsFloat.bitcastToAPInt());
        } else {
          cmpResult = lhs.get(*it) <= rhs.get(*it);
        }
        break;
      }
      case ComparisonDirection::LT: {
        if (comparisonType == ComparisonType::TOTALORDER) {
          auto lhsFloat = lhs.get(*it).getFloatValue();
          auto rhsFloat = rhs.get(*it).getFloatValue();
          if (lhsFloat.isNegative() ^ rhsFloat.isNegative())
            cmpResult = lhsFloat.isNegative();
          else
            cmpResult =
                lhsFloat.bitcastToAPInt().ult(rhsFloat.bitcastToAPInt());
        } else {
          cmpResult = lhs.get(*it) < rhs.get(*it);
        }
        break;
      }
    }
    result.set(*it, Element(elementTy, cmpResult));
  }
  return result;
}
```

and in `Element.cpp` we can have the overloads for `Element::operator<`. 

```
bool Element::operator<(const Element &other) const {
  Type type = other.getType();
  if (type_ != type)
    report_fatal_error(invalidArgument("Element types don't match: %s vs %s",
                                       debugString(type_).c_str(),
                                       debugString(type).c_str()));

  if (isSupportedIntegerType(type)) {
    auto intLhs = getIntegerValue();
    auto intRhs = other.getIntegerValue();
    return isSupportedSignedIntegerType(type) ? intLhs.slt(intRhs)
                                              : intLhs.ult(intRhs);
  }

  if (isSupportedBooleanType(type)) {
    auto boolLhs = getBooleanValue();
    auto boolRhs = other.getBooleanValue();
    return boolLhs < boolRhs;
  }

    auto floatLhs = getFloatValue();
    auto floatRhs = other.getFloatValue();
    return floatLhs.compare(floatRhs) == APFloat::cmpLessThan;
  }

  if (isSupportedComplexType(type)) {
    // TODO
  }

  report_fatal_error(invalidArgument("Unsupported element type: %s",
                                     debugString(type).c_str()));
}

bool Element::operator<=(const Element &other) const {
  return !(*this < other) || (*this == other);
}

bool Element::operator>(const Element &other) const {
  return !(*this <= other);
}

bool Element::operator>=(const Element &other) const {
  return !(*this < other);
}
```

stablehlo/reference/Element.h
```diff
@@ -23,6 +23,7 @@ limitations under the License.
 #include "llvm/Support/raw_ostream.h"
 #include "mlir/IR/BuiltinTypes.h"
 #include "mlir/IR/Types.h"
+#include "stablehlo/dialect/StablehloOps.h"
```
sdasgup3, line 26: `Element.*` does not have to depend on `#include "stablehlo/dialect/StablehloOps.h"` for `ComparisonType` or `ComparisonDirection`. PTAL below for suggestion.

stablehlo/reference/Ops.cpp
ghpvnist: Thanks for the suggestion, I've borrowed majority of this, except that we still need to define our own function for operator> to avoid accidentally negating NaNs (e.g. !(NaN <= NaN) is true when it should be false).

stablehlo/reference/Element.cpp
```diff
@@ -436,6 +437,199 @@ Element ceil(const Element &el) {
...
+  }
+  if (compareType == ComparisonType::TOTALORDER) {
+    switch (comparisonDirection) {
+      case ComparisonDirection::EQ:
+        return lhs.bitwiseIsEqual(rhs);
+      case ComparisonDirection::NE:
+        return !lhs.bitwiseIsEqual(rhs);
+      case ComparisonDirection::GE:
+        if (lhs.isNegative() ^ rhs.isNegative()) return rhs.isNegative();
+        return lhs.isNegative() ? lhs.bitcastToAPInt().getZExtValue() <=
```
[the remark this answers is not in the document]
ghpvnist: If both signs are negative, then we need to flip the comparison. I've modified some code on top of your suggestion. PTAL.

stablehlo/tests/interpret_compare.mlir
ghpvnist: -2 >= -1 is false. In your previous suggestion, there is a small bug which does not account for same negative signs, in which case the boolean should be flipped.

stablehlo/reference/Element.cpp
```diff
@@ -334,44 +338,132 @@ Element Element::operator/(const Element &other) const {
...
+  if (isSupportedFloatType(type)) {
+    auto floatLhs = getFloatValue();
+    auto floatRhs = other.getFloatValue();
+    return floatLhs < floatRhs;
+  }
+
+  if (isSupportedComplexType(type)) {
+    auto complexLhs = getComplexValue();
+    auto complexRhs = other.getComplexValue();
+    if (complexLhs.real().isNaN() || complexLhs.imag().isNaN() ||
```
sdasgup3, line 370: These checks of `NaN` values are redundant. The logic below should help returning the desired false?

sdasgup3: same for `operator>`

stablehlo/reference/Ops.cpp
sdasgup3: >> need to define our own function for operator> to avoid accidentally negating NaNs
That is right. Thanks!

stablehlo/reference/Ops.cpp
```diff
@@ -102,6 +102,125 @@ Tensor evalClampOp(const Tensor &min, const Tensor &operand, const Tensor &max,
...
+        }
+        break;
+      }
+      case ComparisonDirection::GE: {
+        if (comparisonType == ComparisonType::TOTALORDER) {
+          auto lhsFloat = lhs.get(*it).getFloatValue();
+          auto rhsFloat = rhs.get(*it).getFloatValue();
+          if (lhsFloat.isNegative() ^ rhsFloat.isNegative())
+            cmpResult = rhsFloat.isNegative();
+          else
```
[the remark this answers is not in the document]
ghpvnist: Nice!

stablehlo/reference/Ops.cpp
```diff
@@ -102,6 +102,125 @@ Tensor evalClampOp(const Tensor &min, const Tensor &operand, const Tensor &max,
...
+  }
+
+  for (auto it = result.index_begin(); it != result.index_end(); ++it) {
+    bool cmpResult = false;
+    switch (comparisonDirection) {
+      case ComparisonDirection::EQ: {
+        if (comparisonType == ComparisonType::TOTALORDER) {
+          auto lhsFloat = lhs.get(*it).getFloatValue();
+          auto rhsFloat = rhs.get(*it).getFloatValue();
+          cmpResult = lhsFloat.bitwiseIsEqual(rhsFloat);
```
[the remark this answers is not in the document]
ghpvnist: Thanks, this was an oversight.

sdasgup3 (approved):

stablehlo/reference/Ops.cpp
```diff
@@ -102,6 +102,123 @@ Tensor evalClampOp(const Tensor &min, const Tensor &operand, const Tensor &max,
...
+        if (comparisonType == ComparisonType::TOTALORDER) {
+          auto lhsFloat = lhs.get(*it).getFloatValue();
+          auto rhsFloat = rhs.get(*it).getFloatValue();
+          if (lhsFloat.isNegative() ^ rhsFloat.isNegative())
+            cmpResult = rhsFloat.isNegative();
+          else
+            cmpResult =
+                lhsFloat.isNegative()
+                    ? lhsFloat.bitcastToAPInt().ule(rhsFloat.bitcastToAPInt())
+                    : lhsFloat.bitcastToAPInt().uge(rhsFloat.bitcastToAPInt());
```
burmako, line 162: Also, is anyone using `TOTALORDER`? I skimmed the usual suspects, and I don't think they do. 

Given that, I think that we should just skip implementing it in the interpreter and (in this PR) update the spec along the lines of "For floating-point element types and `compare_type = TOTALORDER`, the op uses the combination of `totalOrder` and `compareQuietEqual` operations from IEEE-754. This feature appears to be unused, so in the future, we are planning to remove it ([#584](https://github.com/openxla/stablehlo/issues/584))"?

Also, I've just moved the corresponding ticket from New Features (P1) to Frontend Contract (P0) because the presence of this feature seems to have non-trivial poitential for confusion, so I think we should resolve it by v1.0.

burmako (changes requested):

stablehlo/reference/Element.cpp
```diff
@@ -362,44 +366,124 @@ Element Element::operator/(const Element &other) const {
...
+    auto complexLhs = getComplexValue();
+    auto complexRhs = other.getComplexValue();
+    if (complexLhs.real() < complexRhs.real()) return true;
+    if (!(complexLhs.real() == complexRhs.real())) return false;
+    return complexLhs.imag() < complexRhs.imag();
+  }
+
+  report_fatal_error(invalidArgument("Unsupported element type: %s",
+                                     debugString(type).c_str()));
 }
```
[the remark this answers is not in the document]
ghpvnist: The current `map` function assumes the element types do not change, but regardless of the input element type, these functions return a `bool`, and the current map implementations do not allow specifying the input type to check and the result type to return. Also, `map` does not differentiate signed vs unsigned integers, so it will require some refactoring to use it. I've opened a ticket for this #1303.

stablehlo/reference/Ops.cpp
```diff
@@ -102,6 +102,123 @@ Tensor evalClampOp(const Tensor &min, const Tensor &operand, const Tensor &max,
...
+  if (compareType && *compareType != ComparisonType::NOTYPE) {
+    comparisonType = *compareType;
+  } else if (isSupportedUnsignedIntegerType(elementTy) ||
+             isSupportedBooleanType(elementTy)) {
+    comparisonType = ComparisonType::UNSIGNED;
+  } else if (isSupportedSignedIntegerType(elementTy)) {
+    comparisonType = ComparisonType::SIGNED;
+  } else if (isSupportedFloatType(elementTy) ||
+             isSupportedComplexType(elementTy)) {
+    comparisonType = ComparisonType::FLOAT;
```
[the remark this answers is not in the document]
ghpvnist: Hmm, my opinion is that the comment is not necessary. There are 5 enum types, namely `NOTYPE`, `UNSIGNED`, `SIGNED`, `FLOAT`, and `TOTALORDER`. Obviously, `NOTYPE`, `SIGNED`, and `UNSIGNED` are out of the question. It can't be `TOTALORDER` since this has to do with the absolute ordering of the elements and unrelated to the element type of what is being compared. Complex types shouldn't be treated differently, and under the hood, it uses a pair of floating-points. The spec also does state it in (C3):
```
* (C3) Given `E` is the `lhs` element type, the following are legal values of
       `compare_type`:
  * ...
  * ...
  * ...
  * If `E` is complex type, `compare_type` = `FLOAT`.
```

stablehlo/reference/Ops.cpp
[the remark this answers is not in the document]
ghpvnist: Good question! I was also weary of using it. IEEE section 5.10 c.3 says:
```
If x and y represent the same floating-point datum:

i) If x and y have negative sign,
totalOrder(x, y) is true if and only if the exponent of x ≥ the exponent of y

ii) otherwise,
totalOrder(x, y) is true if and only if the exponent of x ≤ the exponent of
```
This behavior can be captured by interpreting the bits as an integer and comparing it like normal integer comparison.

stablehlo/reference/Element.cpp
```diff
@@ -362,44 +366,124 @@ Element Element::operator/(const Element &other) const {
...
+    return floatLhs < floatRhs;
+  }
+
+  if (isSupportedComplexType(type)) {
+    auto complexLhs = getComplexValue();
+    auto complexRhs = other.getComplexValue();
+    if (complexLhs.real() < complexRhs.real()) return true;
+    if (!(complexLhs.real() == complexRhs.real())) return false;
+    return complexLhs.imag() < complexRhs.imag();
+  }
```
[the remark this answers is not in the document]
ghpvnist: I've updated the PR description. Do you want me to also comment in the code? I think it'd be unnecessary as there would be no code changes when we drop support for this.

stablehlo/reference/Ops.cpp
ghpvnist: #584 links to HLO code that says this is deprecated: https://github.com/tensorflow/tensorflow/blob/3451f5d39ea8add4c254acc9615d62ff589259b8/tensorflow/compiler/xla/comparison_util.h#L71-L79
```
  // (DEPRECATED) Represents the type of comparison. Prefer xla::PrimitiveType
  // and Comparison::Order, since there are multiple floating point
  // representations that support total ordering.
  enum class [[deprecated("Use PrimitiveType and Order")]] Type : uint8_t{
      kFloat,
      kFloatTotalOrder,
      kSigned,
      kUnsigned,
  };
```
but a few lines above that, there is another enum class introduced to replace this, and indeed it is being used in [HLO](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/comparison_util.cc#L55).

https://github.com/tensorflow/tensorflow/blob/3451f5d39ea8add4c254acc9615d62ff589259b8/tensorflow/compiler/xla/comparison_util.h#L54-L59
```
  // Represents the ordering of the comparison.
  enum class Order : uint8_t {
    // https://en.wikipedia.org/wiki/Total_order
    kTotal,
    // https://en.wikipedia.org/wiki/Partially_ordered_set
    kPartial,
  };
```

docs/spec.md
```diff
@@ -1603,8 +1603,10 @@ For boolean and integer element types:
 * `LE`: `lhs` $\le$ `rhs`.
 * `LT`: `lhs` $\lt$ `rhs`.
 
-For floating-point element types and `compare_type = FLOAT`, the op implements
-the following IEEE-754 operations:
+For floating-point element types with `compare_type = FLOAT`, the op implements
+the following IEEE-754 operations. If any comparison is unordered,
```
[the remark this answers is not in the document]
ghpvnist: I'm fine with keeping either versions:
a)
```
If any comparison is unordered, `comparison_direction = NE` compares unordered
`NE` and the rest uses ordered comparison.
```
b)
```
If any of `lhs` or `rhs` is `NaN`, the op returns `false` except for
`comparison_direction = NE` for which it returns `true` to make `lhs != rhs`
equivalent to `!(lhs == rhs)`.
```

Any preference?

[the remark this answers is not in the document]
burmako: Why do we need to make any changes to the spec? 

`compareQuietNotEqual` is already unordered, in the sense that it is true when the result of comparison is `unordered` (which arises when at least one operand is NaN).

Other comparison predicates called out in the spec are already ordered, in the sense that they are false when the result of comparison is `unordered`.

stablehlo/tests/print_stablehlo.mlir
```diff
@@ -241,9 +241,9 @@ func.func @pairwise_ops(%arg0 : tensor<4xf32>) -> () {
 // CHECK-LABEL: func @compare_op
 func.func @compare_op(%arg0 : tensor<3xi32>) -> () {
   // CHECK:      %0 = stablehlo.compare LT, %arg0, %arg0 : (tensor<3xi32>, tensor<3xi32>) -> tensor<3xi1>
-  // CHECK-NEXT: %1 = stablehlo.compare LT, %arg0, %arg0, TOTALORDER : (tensor<3xi32>, tensor<3xi32>) -> tensor<3xi1>
+  // CHECK-NEXT: %1 = stablehlo.compare LT, %arg0, %arg0, SIGNED : (tensor<3xi32>, tensor<3xi32>) -> tensor<3xi1>
    %0 = "stablehlo.compare"(%arg0, %arg0) {comparison_direction = #stablehlo<comparison_direction LT>} : (tensor<3xi32>, tensor<3xi32>) -> tensor<3xi1>
-   %1 = "stablehlo.compare"(%arg0, %arg0) {compare_type = #stablehlo<comparison_type TOTALORDER>, comparison_direction = #stablehlo<comparison_direction LT>} : (tensor<3xi32>, tensor<3xi32>) -> tensor<3xi1>
+   %1 = "stablehlo.compare"(%arg0, %arg0) {compare_type = #stablehlo<comparison_type SIGNED>, comparison_direction = #stablehlo<comparison_direction LT>} : (tensor<3xi32>, tensor<3xi32>) -> tensor<3xi1>
   "stablehlo.return"() : () -> ()
 }
```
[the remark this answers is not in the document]
burmako: Not sure why this comment got resolved. It looks like `stablehlo/tests/stablehlo_legalize_to_vhlo.mlir` is still modified.

stablehlo/reference/Ops.cpp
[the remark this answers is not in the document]
burmako: Agreed that the spec is clear about what comparison type should be used for complex types. It is still not obvious to me why we should be using `FLOAT` over `TOTALORDER` for floating-point types, since the spec says that `TOTALORDER` is (only) legal for them.

Also, now that I'm thinking further about this, the spec doesn't say anything about NOTYPE, so we need a comment anyway because even computing a default value of ComparisonType is deviating from the spec. How about this?

```C++
  // TODO(#1311): Unlike `stablehlo.compare` from the spec, the StableHLO 
  // dialect defines `compare_type` as an optional attribute, so here we 
  // explicitly compute its default value.
  // For non-floating point element types, the default value is unambiguously
  // defined from the C3 constraint in the spec. For floating-point element
  // types, we're picking FLOAT because that's the conventional semantics
  // across the users of CompareOp.
  auto comparisonType = ComparisonType::NOTYPE;
  if (compareType) {
    ...
```

stablehlo/reference/Ops.cpp
burmako: It is evident that `Order::kTotal` seems to be used in HLO. However, is it used by frontends? That's what would matter for including or not including this in StableHLO.

burmako (changes requested):
A couple more comments somehow got posted outside of this review - they will be visible as unresolved conversations.

docs/spec.md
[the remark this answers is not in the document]
ghpvnist: Oh, I misunderstood "everything stated in the addition to the StableHLO spec in this PR" part. I see your point that original spec already explains this behavior.

stablehlo/tests/print_stablehlo.mlir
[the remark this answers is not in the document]
ghpvnist: Oops not sure how I missed that either. Both files should be addressed now. Thanks.

stablehlo/reference/Ops.cpp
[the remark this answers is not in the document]
ghpvnist: I see your point, your suggestion makes sense.

stablehlo/reference/Ops.cpp
ghpvnist: I looked into this recently, and it looks like none of the frontends are using it specifically in CompareOp (e.g. they may use total order in, say, sort op by default) or is a backend implementation detail.

docs/spec.md
```diff
@@ -1614,9 +1614,9 @@ the following IEEE-754 operations:
 * `LE`: `compareQuietLessEqual`.
 * `LT`: `compareQuietLess`.
 
-For floating-point element types and `compare_type = TOTALORDER`, the op
+For floating-point element types with `compare_type = TOTALORDER`, the op
```
burmako, line 1617: If we are updating "and" to "with" here, then we'll want to do the same for the preceding paragraph for consistency.

stablehlo/reference/Ops.cpp
```diff
@@ -102,6 +102,50 @@ Tensor evalClampOp(const Tensor &min, const Tensor &operand, const Tensor &max,
...
+        break;
+      case ComparisonDirection::GT:
+        cmpResult = lhs.get(*it) > rhs.get(*it);
+        break;
+      case ComparisonDirection::LE:
+        cmpResult = lhs.get(*it) <= rhs.get(*it);
+        break;
+      case ComparisonDirection::LT:
+        cmpResult = lhs.get(*it) < rhs.get(*it);
+        break;
```
burmako, line 142: Let's insert a `default` case here, so that this doesn't start to silently misbehave if we add another comparison direction.

stablehlo/reference/Ops.cpp
```diff
@@ -102,6 +102,50 @@ Tensor evalClampOp(const Tensor &min, const Tensor &operand, const Tensor &max,
...
+       comparisonDirection == ComparisonDirection::LE ||
+       comparisonDirection == ComparisonDirection::LT)) {
+    report_fatal_error(invalidArgument(
+        "Unsupported element type %s for comparison direction %s",
+        debugString(elementTy).c_str(),
+        debugString(comparisonDirection).c_str()));
+  }
+
+  for (auto it = result.index_begin(); it != result.index_end(); ++it) {
+    bool cmpResult = false;
```
burmako, line 123: `bool resultElement`: 
  1) E.g. evalClampOp above uses the same convention to distinguish `result` vs an element of a result.
  2) Let's make this variable uninitialized to make sure that it's initialized with a reasonable value before it's used. (Rather than potentially enable cases when `false` keeps being used because of a logic error).

burmako (approved):
LGTM with minor changes.

burmako (approved):

PR #1207, opened 2023-02-24
Add interpreter for PowOp

### LINKED ISSUE #981: Add interpreter for power

subhankarshah:
closes #981 

```
(I1) lhs is a tensor of integer, floating-point, or complex type.
(I2) rhs is a tensor of integer, floating-point, or complex type.
(C1) lhs, rhs, and result have the same type.
```

```
(I1) lhs is not a tensor of integer, floating-point, or complex type. (covered by ODS)
(I2) rhs is not a tensor of integer, floating-point, or complex type. (covered by ODS)
(C1) lhs, rhs, and result do not have the same type. (covered by ODS)
```

### DIFF
--- a/stablehlo/reference/Element.cpp
+++ b/stablehlo/reference/Element.cpp
@@ -538,6 +538,67 @@ Element min(const Element &e1, const Element &e2) {
       });
 }
 
+Element power(const Element &e1, const Element &e2) {
+  Type type = e1.getType();
+  if (e1.getType() != e2.getType())
+    report_fatal_error(invalidArgument("Element types don't match: %s vs %s",
+                                       debugString(e1.getType()).c_str(),
+                                       debugString(e2.getType()).c_str()));
+
+  if (isSupportedIntegerType(type)) {
+    APInt x = e1.getIntegerValue();
+    APInt y = e2.getIntegerValue();
+    bool isSigned = x.isSignedIntN(x.getBitWidth());
+    if (isSigned && y.isNegative()) {
+      if (x.isZero())
+        llvm::report_fatal_error("divide by zero unsupported");
+      else if (x.abs().isOne())
+        y = y.abs();
+      else
+        return Element(type, APInt(x.getBitWidth(), 0, isSigned));
+    }
+    APInt constant1(x.getBitWidth(), 1, isSigned);
+    APInt result(constant1);
+    while (!y.isZero()) {
+      if ((y & constant1).getBoolValue()) result *= x;
+      x *= x;
+      y = y.lshr(1);
+    }
+    return Element(type, result);
+  }
+
+  if (isSupportedFloatType(type)) {
+    APFloat lhsVal = e1.getFloatValue();
+    APFloat rhsVal = e2.getFloatValue();
+    const llvm::fltSemantics &elSemantics = lhsVal.getSemantics();
+    APFloat resultVal(
+        std::pow(lhsVal.convertToDouble(), rhsVal.convertToDouble()));
+    bool roundingErr;
+    resultVal.convert(elSemantics, APFloat::rmNearestTiesToEven, &roundingErr);
+    return Element(type, resultVal);
+  }
+
+  if (isSupportedComplexType(type)) {
+    std::complex<APFloat> lhsVal = e1.getComplexValue();
+    std::complex<APFloat> rhsVal = e2.getComplexValue();
+    const llvm::fltSemantics &elSemantics = lhsVal.real().getSemantics();
+    auto resultVal =
+        std::pow(std::complex<double>(lhsVal.real().convertToDouble(),
+                                      lhsVal.imag().convertToDouble()),
+                 std::complex<double>(rhsVal.real().convertToDouble(),
+                                      rhsVal.imag().convertToDouble()));
+    bool roundingErr;
+    APFloat resultReal(resultVal.real());
+    resultReal.convert(elSemantics, APFloat::rmNearestTiesToEven, &roundingErr);
+    APFloat resultImag(resultVal.imag());
+    resultImag.convert(elSemantics, APFloat::rmNearestTiesToEven, &roundingErr);
+    return Element(type, std::complex<APFloat>(resultReal, resultImag));
+  }
+
+  report_fatal_error(invalidArgument("Unsupported element type: %s",
+                                     debugString(type).c_str()));
+}
+
 Element real(const Element &el) {
   if (isSupportedFloatType(el.getType())) return el;
   if (isSupportedComplexType(el.getType()))
--- a/stablehlo/reference/Element.cpp
+++ b/stablehlo/reference/Element.cpp
@@ -126,31 +126,18 @@ Element mapWithUpcastToDouble(const Element &lhs, const Element &rhs,
                                        debugString(rhs.getType()).c_str()));
 
   if (isSupportedFloatType(type)) {
-    APFloat lhsVal = lhs.getFloatValue();
-    APFloat rhsVal = rhs.getFloatValue();
-    const llvm::fltSemantics &elSemantics = lhsVal.getSemantics();
-    APFloat resultVal(
-        floatFn(lhsVal.convertToDouble(), rhsVal.convertToDouble()));
-    bool roundingErr;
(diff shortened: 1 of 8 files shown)


stablehlo/tests/interpret_power.mlir
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// -----
+
+// CHECK-LABEL: Evaluated results of function: power_op_test_f64
+func.func @power_op_test_f64() {
+  %lhs = stablehlo.constant dense<[-2.0, -0.0, -36.0, 5.0, 3.0, 10000.0]> : tensor<6xf64>
+  %rhs = stablehlo.constant dense<[2.0, 2.0, 1.1, 2.0, -1.0, 10.0]> : tensor<6xf64>
+  %result = stablehlo.power %lhs, %rhs : tensor<6xf64>
+  check.almost_eq %result, dense<[4.000000e+00, 0.000000e+00, 0xFFF8000000000000,
```
ghpvnist, line 30: Is `0xFFF8000000000000` the value produced by the APFloat API or arbitrary?

ghpvnist (changes requested):

stablehlo/tests/interpret_power.mlir
subhankarshah: It is the actual result.

stablehlo/reference/Element.cpp
```diff
@@ -131,6 +131,47 @@ Element mapWithUpcastToDouble(const Element &el, FloatFn floatFn,
                                      debugString(type).c_str()));
 }
 
+template <typename FloatFn, typename ComplexFn>
+Element mapWithUpcastToDouble(const Element &lhs, const Element &rhs,
+                              FloatFn floatFn, ComplexFn complexFn) {
+  Type type = lhs.getType();
+  if (lhs.getType() != rhs.getType())
```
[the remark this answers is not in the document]
subhankarshah: This follows the binary map function convention, it is generic, does not have to do with power op specifically.

stablehlo/reference/Element.cpp
```diff
@@ -511,6 +552,38 @@ Element min(const Element &e1, const Element &e2) {
       });
 }
 
+Element power(const Element &e1, const Element &e2) {
+  Type type = e1.getType();
+
+  if (isSupportedIntegerType(type)) {
+    APInt x = e1.getIntegerValue();
```
ghpvnist, line 559: Can you rename the variables to something like `base` and `exponent`?

stablehlo/reference/Element.cpp
```diff
@@ -131,6 +131,47 @@ Element mapWithUpcastToDouble(const Element &el, FloatFn floatFn,
...
+  Type type = lhs.getType();
+  if (lhs.getType() != rhs.getType())
+    report_fatal_error(invalidArgument("Element types don't match: %s vs %s",
+                                       debugString(lhs.getType()).c_str(),
+                                       debugString(rhs.getType()).c_str()));
+
+  if (isSupportedFloatType(type)) {
+    APFloat lhsVal = lhs.getFloatValue();
+    APFloat rhsVal = rhs.getFloatValue();
+    const llvm::fltSemantics &elSemantics = lhsVal.getSemantics();
```
ghpvnist, line 146: Can you rebase this to #1154, it will simplify the logic in this map function.

ghpvnist (changes requested):

stablehlo/reference/Element.cpp
```diff
@@ -116,6 +116,34 @@ Element mapWithUpcastToDouble(const Element &el, FloatFn floatFn,
...
+template <typename FloatFn, typename ComplexFn>
+Element mapWithUpcastToDouble(const Element &lhs, const Element &rhs,
+                              FloatFn floatFn, ComplexFn complexFn) {
+  Type type = lhs.getType();
+  if (lhs.getType() != rhs.getType())
+    report_fatal_error(invalidArgument("Element types don't match: %s vs %s",
+                                       debugString(lhs.getType()).c_str(),
+                                       debugString(rhs.getType()).c_str()));
+
+  if (isSupportedFloatType(type)) {
```
ghpvnist, line 128: Remove braces here and below.

ghpvnist (approved):
LGTM with minor comment.

stablehlo/reference/Element.cpp
```diff
@@ -538,6 +564,38 @@ Element min(const Element &e1, const Element &e2) {
       });
 }
 
+Element power(const Element &e1, const Element &e2) {
+  Type type = e1.getType();
+
+  if (isSupportedIntegerType(type)) {
+    APInt base = e1.getIntegerValue();
+    APInt exponent = e2.getIntegerValue();
+    bool isSigned = base.isSignedIntN(base.getBitWidth());
```
burmako, line 573: This is not going to do what you expect, because `APInt::isSignedIntN` already assumes that the number is signed. `isSupportedSignedIntegerType(type)` is the right way of checking this.

stablehlo/reference/Element.cpp
```diff
@@ -538,6 +564,38 @@ Element min(const Element &e1, const Element &e2) {
...
+    if (isSigned && exponent.isNegative()) {
+      if (base.isZero())
+        llvm::report_fatal_error("divide by zero unsupported");
```
burmako, line 576: Let's drop these two lines of code to simplify the interpreter? Per the spec, corner cases like this one are implementation-defined, so we might as well define our implementation to be as simple as possible.

stablehlo/reference/Element.cpp
```diff
@@ -538,6 +564,38 @@ Element min(const Element &e1, const Element &e2) {
...
+      else if (base.abs().isOne())
+        exponent = exponent.abs();
+      else
+        return Element(type, APInt(base.getBitWidth(), 0, isSigned));
```
burmako, line 580: `return Element(type, 0)`?

stablehlo/reference/Element.cpp
```diff
@@ -538,6 +564,38 @@ Element min(const Element &e1, const Element &e2) {
...
+    }
+    const APInt kOne(base.getBitWidth(), 1, isSigned);
+    APInt result(kOne);
+    while (!exponent.isZero()) {
+      if ((exponent & kOne).getBoolValue()) result *= base;
+      base *= base;
+      exponent = exponent.lshr(1);
+    }
```
burmako, line 588: Proposed refactoring:

```
APInt result(base.getBitWidth(), 1, isSigned);
while (!exponent.isZero()) {
  if (!(exponent & 1).isZero()) result *= base;
  base *= base;
  exponent >>= 1;
}
```

Feedback:
  1) We don't need kOne because `&` takes regular integers as well.
  2) `isZero` check of the result of `&` represents the intent more directly - check whether the bit is set by comparing the output of AND with zero.
  3) `exponent >>= 1` is a thing, so let's leverage it.

burmako (changes requested):

stablehlo/reference/Element.cpp
subhankarshah: This would mean 0 raised to negative power would evaluate to 0.

stablehlo/reference/Element.cpp
subhankarshah: Regarding 3
error: no match for ‘operator>>=’ (operand types are ‘llvm::APInt’ and ‘int’)
error: no match for ‘operator>>’ (operand types are ‘llvm::APInt’ and ‘int’)
error: no match for ‘operator>>’ (operand types are ‘llvm::APInt’ and ‘llvm::APInt’)

stablehlo/reference/Element.cpp
burmako: I think it would be best to produce some value rather than crashing. If we have two big tensors and just one of the elementwise pairs is zero base and negative exponent, it is useful to continue evaluation rather than crash the entire thing.

stablehlo/reference/Element.cpp
burmako: Apologies - I'm realizing that the proposed code doesn't compile. You'll want to say `return Element(type, 0l)` to disambiguate.

burmako (approved):

Issue #1196, opened 2023-02-20
Improve error reporting for the interpreter testing


sdasgup3:
This would involve: 1) somehow adding negative tests, 2) improving the error message which is currently going to pretty hard to debug (it just prints mismatching values, without printing the whole tensors or the indices of the mismatching position).

PR #1210, opened 2023-02-26
Add index info in interpreter-mismatch error message

sdasgup3:
Even though we have a dedicated issue https://github.com/openxla/stablehlo/issues/1196 to help improving such cases, this change will have some immediate benefits during debugging.

### DIFF
--- a/stablehlo/tests/CheckOps.cpp
+++ b/stablehlo/tests/CheckOps.cpp
@@ -57,9 +57,10 @@ llvm::Error evalEqOp(const Tensor &lhs, ElementsAttr value) {
        lhsIt != lhs.index_end(); ++lhsIt, ++rhsIt)
     if (lhs.get(*lhsIt) != rhs.get(*rhsIt))
       return invalidArgument(
-          "Element values don't match: %s (actual) vs %s (expected)\n",
+          "Element values don't match: %s (actual) vs %s (expected) at index %s\n",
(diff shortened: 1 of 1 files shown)


stablehlo/tests/CheckOps.cpp
```diff
@@ -57,9 +57,10 @@ llvm::Error evalEqOp(const Tensor &lhs, ElementsAttr value) {
        lhsIt != lhs.index_end(); ++lhsIt, ++rhsIt)
     if (lhs.get(*lhsIt) != rhs.get(*rhsIt))
       return invalidArgument(
-          "Element values don't match: %s (actual) vs %s (expected)\n",
+          "Element values don't match: %s (actual) vs %s (expected) at index %s\n",
           debugString(lhs.get(*lhsIt)).c_str(),
-          debugString(rhs.get(*rhsIt)).c_str());
+          debugString(rhs.get(*rhsIt)).c_str(),
+          debugString((*lhsIt)).c_str());
```
burmako, line 63: What about almost_eq?

burmako (changes requested):

stablehlo/tests/CheckOps.cpp
sdasgup3: Yes, I missed that. Added

burmako (approved):

PR #1213, opened 2023-02-27
Refactor StableHLO <=> VHLO

burmako:
In the last few pull requests, I crammed a lot of adhoc logic into
the corresponding conversion patterns.

When working on removing optional attrs from VHLO, I needed to further
update the conversion patterns, and I found it to be unnecessarily
complicated with all the stuff in there.

This PR refactors the conversion patterns to encapsulate the logic
that bridges the gaps between StableHLO and VHLO. I think that modifying
the conversion patterns will become considerably easier now.

### DIFF
--- a/stablehlo/transforms/StablehloLegalizeToVhlo.cpp
+++ b/stablehlo/transforms/StablehloLegalizeToVhlo.cpp
@@ -500,55 +500,45 @@ SpecialResult convertSpecial(const OpConversionPattern<StablehloOpTy>& pattern,
                              stablehlo::CollectivePermuteOp>::value ||
                 std::is_same<StablehloOpTy,
                              stablehlo::ReduceScatterOp>::value) {
-    if (stablehloName == "channel_handle") {
+    if (stablehloName == "channel_handle")
       return convertChannelId(pattern, stablehloAttr, vhloAttrs);
-    }
-    if (stablehloName == "use_global_device_ids") {
+    if (stablehloName == "use_global_device_ids")
       return convertUseGlobalDeviceIds(pattern, stablehloAttr, vhloAttrs);
-    }
   }
   if constexpr (std::is_same<StablehloOpTy, stablehlo::ConvolutionOp>::value ||
                 std::is_same<StablehloOpTy, stablehlo::DynamicConvOp>::value) {
-    if (stablehloName == "dimension_numbers") {
+    if (stablehloName == "dimension_numbers")
       return convertConvDimensionNumbers(pattern, stablehloAttr, vhloAttrs);
-    }
   }
   if constexpr (std::is_same<StablehloOpTy, stablehlo::CustomCallOp>::value) {
-    if (stablehloName == "api_version") {
+    if (stablehloName == "api_version")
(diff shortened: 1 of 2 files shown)


stablehlo/transforms/StablehloLegalizeToVhlo.cpp
```diff
@@ -182,261 +188,377 @@ Attribute convertAttrToVhlo(Attribute stablehloAttr,
...
+                             StringRef stablehloName, Attribute stablehloAttr,
+                             SmallVector<NamedAttribute>& vhloAttrs) {
+  if constexpr (std::is_same<StablehloOpTy, stablehlo::AllGatherOp>::value ||
+                std::is_same<StablehloOpTy, stablehlo::AllReduceOp>::value ||
+                std::is_same<StablehloOpTy, stablehlo::AllToAllOp>::value ||
+                std::is_same<StablehloOpTy,
+                             stablehlo::CollectivePermuteOp>::value ||
+                std::is_same<StablehloOpTy,
+                             stablehlo::ReduceScatterOp>::value) {
+    if (stablehloName == "channel_handle") {
```
GleasonK, line 503: How about no curly braces on the `if (stablehloName == "...")` but keep them on the constexpr ones for dangling if protection

stablehlo/transforms/VhloLegalizeToStablehlo.cpp
```diff
@@ -324,31 +350,123 @@ Attribute convertScatterDimensionNumbers(vhlo::ScatterOpV1 vhloOp,
...
-    : public impl::VhloLegalizeToStablehloPassBase<
-          VhloLegalizeToStablehloPass> {
-  void runOnOperation() override {
-    ConversionTarget target(getContext());
-    target.addIllegalDialect<vhlo::VhloDialect>();
-    target.addLegalDialect<stablehlo::StablehloDialect>();
-    target.addLegalDialect<func::FuncDialect>();
+template <typename... StringTy>
+void eraseAttrs(SmallVector<NamedAttribute>& attrs, StringTy... names) {
+  StringSet<llvm::MallocAllocator> nameSet({names...});
```
GleasonK, line 355: If we use a set type that takes list initialization we could get away without a template function. From a callsite perspective the grouping makes sense:
`eraseAttrs(attrs, {things, to, erase})`

GleasonK (approved):
Took me a minute to understand the setup, but I think it is good. My worry with separating the logic was having to modify several functions in a few different places, but I think this design does a good job as single-sourcing things.

stablehlo/transforms/StablehloLegalizeToVhlo.cpp
burmako: Done.

stablehlo/transforms/VhloLegalizeToStablehlo.cpp
burmako: Yeah, I considered that too, but I think it's nice to type fewer characters, and the template function doesn't look awfully complicated. Given that you approved the PR, I'll merge and then we can follow up on this if needed.

Issue #1220, opened 2023-02-27
Block optimizations in stablehlo/conformance


burmako:
It would be useful to explicitly block optimizations in the conformance suite, so that constant folding doesn't fold the entire suite away. IREE [has](https://github.com/openxla/iree/blob/96b61ba6fe20e6cf9a647cd4019612940fc879ec/compiler/src/iree/compiler/Dialect/Util/IR/UtilOps.td#L304-L306) `util.optimization_barrier` and `util.unfoldable_constant`, but we could also consider using `stablehlo.optimization_barrier` for this purpose.

GleasonK:
Ideally we could do this in a pass to keep tests concise (wrap all constants with a barrier, update uses). StableHLO doesn't have optimizations, so any consumer of StableHLO directly wouldn't need this feature.

Issue #1219, opened 2023-02-27
Align our Check dialect with IREE's Check dialect


burmako:
To follow up on #1218, it would be great to align [our Check dialect](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/CheckOps.td) with [IREE's Check dialect](https://github.com/openxla/iree/blob/main/compiler/src/iree/compiler/Modules/Check/IR/CheckOps.td), so that the conformance suite could also be used verbatim as a test suite for IREE.

GleasonK:
This is a great idea. Consideration should be given to single-sourcing the check dialect if possible (if dependencies allow).

I want to also note that I like IREE's distinction between `expect*` and `expect*_const`. It would be useful in StableHLO to have a variant that can accept two operands, especially for our generated test data which stores expected results separately for readability.

Issue #1218, opened 2023-02-27
Convert stablehlo/testdata to use the Check dialect


burmako:
It would be good to update the conformance suite to use the Check dialect, so that the tests are more easily aligned with IREE.

sdasgup3:
@GleasonK 

Following the suggestion here in this issue and the fact that VHLO does not support the check ops, does it make sense add the `custom_call` wrapper around the check ops just before serialization and according unwrap after de-serialization.

GleasonK:
This is the reason custom_call is currently used. I think a lot of these questions require a design document and discussions with stakeholders that we won't have a chance to do in the near future (unless we reshuffle a bit).

There is a tradeoff no matter how we go about it:

**Option 1**, Test files use custom_call, and provide passes to meet consumer requirements. Under this design we require consumers to call these passes in their test infra. This has the benefit that not all consumers _need_ any of these transforms. Ex1: Any backend that converts these tests into an MLIR dialect that may optimize away constants will need to call these passes before running their test.

**Option 2**, Test files use check dialect and have optimization_barriers (#1220) embedded. Under this design, some consumers may need to write transformations to remove these bits they don't want. Ex1: StableHLO has no folders so if a backend consumes StableHLO they now need to support optimization_barrier just for these tests. Ex2: VHLO doesn't support check dialect so we need to convert them away into custom calls for serialization as you mention.

Issue #1217, opened 2023-02-27
Turn stablehlo/testdata into a compatibility suite


burmako:
As highlighted in #1215, the newly checked in files in stablehlo/conformance can be turned into a compatibility suite in two ways: 1) static tests (roundtrip through VHLO and compare syntax), 2) dynamic tests (roundtrip through VHLO and compare semantics via reference implementation). Let's start with static tests.

PR #1221, opened 2023-02-27
Move generated tests to testdata until conformance requirements are fully spec'ed

GleasonK:
A formal spec will be needed for conformance suite, covering topics like:
- Test structure, ensure that all clients are able to consume and use the test data
- Test content, criteria for a test to be included in the conformance suite
- Test cleanup, remove redundancy in the test files
- Op spec, only fully spec'ed ops should be covered in a conformance suite
- Floating point imprecision, what to do for these ops, and what threshold to use?

In the meanwhile we'll move `conformance/` to `testdata/` to avoid giving an impression that the tests in there can already be used as a conformance suite. These tests are still valuable and can be used for validating a lot of aspects of an implementation, but they are not yet a full conformance suite.

Once the requirements are ready and we have a roadmap for implementing them, one of the items on the roadmap will be moving the tests back to `conformance/`.

Related issues: #1220, #1219, #1218, #1217, #1216


burmako (approved):
LGTM. Let's also document the intent a bit further in the PR description.

PR #1239, opened 2023-03-01
Add interpreter for LogisticOp

### LINKED ISSUE #978: Add interpreter for logistic

ghpvnist:
```
(I1) operand is a tensor of floating-point or complex type. (Covered by ODS).
(C1) `operand` and `result` have the same type. (Covered by ODS).
```

No additional tests are needed as all constraints and shape inference is covered by ODS.

closes #978

### DIFF
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -431,8 +431,9 @@ def StableHLO_Log1pOp: StableHLO_UnaryElementwiseOp<"log_plus_one",
 }
 
 def StableHLO_LogisticOp: StableHLO_UnaryElementwiseOp<"logistic",
-    [Pure, HLO_CompatibleOperandsAndResultType], HLO_FpOrComplexTensor> {
-  let summary = "Logistic operation";
+    [Pure, HLO_CompatibleOperandsAndResultType /*logistic_c1*/],
+    HLO_FpOrComplexTensor /*logistic_i1*/> { /*logistic_c1*/
+  let summary = "Logistic operator";
   let description = [{
     Performs element-wise logistic operation on `operand` tensor and produces a
     `result` tensor.
@@ -442,7 +443,7 @@ def StableHLO_LogisticOp: StableHLO_UnaryElementwiseOp<"logistic",
 
     Example:
     ```mlir
-    %result = stablehlo.logistic %operand : tensor<2x2xf32>
+    %result = stablehlo.logistic %operand : tensor<2x2xf64>
     ```
   }];
 }
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -433,7 +433,7 @@ def StableHLO_Log1pOp: StableHLO_UnaryElementwiseOp<"log_plus_one",
 def StableHLO_LogisticOp: StableHLO_UnaryElementwiseOp<"logistic",
     [Pure, HLO_CompatibleOperandsAndResultType /*logistic_c1*/],
     HLO_FpOrComplexTensor /*logistic_i1*/> { /*logistic_c1*/
-  let summary = "Logistic operator";
+  let summary = "Logistic operation";
   let description = [{
     Performs element-wise logistic operation on `operand` tensor and produces a
     `result` tensor.
--- a/stablehlo/reference/Element.cpp
+++ b/stablehlo/reference/Element.cpp
@@ -530,6 +530,14 @@ Element log(const Element &el) {
       [](std::complex<double> e) { return std::log(e); });
 }
 
+Element logistic(const Element &el) {
+  Element one;
+  if (isSupportedFloatType(el.getType())) one = Element(el.getType(), 1.0);
+  if (isSupportedComplexType(el.getType()))
+    one = Element(el.getType(), std::complex<double>(1.0));
+  return one / (one + exponential(-el));
+}
+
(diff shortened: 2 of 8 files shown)


stablehlo/reference/Element.cpp
```diff
@@ -502,6 +502,14 @@ Element log(const Element &el) {
       [](std::complex<double> e) { return std::log(e); });
 }
 
+Element logistic(const Element &el) {
+  Element one;
```
sdasgup3, line 506: Given that 

(1) `Element` equivalents of ones/zeros are something we might need in  other places as well
(2) For creating an `Element` obj out of a arbitrary real, imag pair, we use 
```
  Element(Type type, std::complex<double> value);
```
This constructor is mostly used with creating `Element` equivalent of a complex value with imag part == 0. For e.g its usage in the `iota`. IMO, this function will be rarely used to get Element equiv of an arbitrary complex values with imag != 0. 

With that, how about we  use 
```
  Element(Type type, double value);
```
for creating both complex and float types, with proper comments explaining the API.

With that we can get:
```
Element one(el.getType(), 1.0);
return one / (one + exponential(-el));
```

Additionally,
We can chose to remove the following:
```

Element(Type type, std::complex<double> value); // no compelling use
Element(Type type, bool); // Removing it should not affect any usages as  
                                            // Element(Type type, int64_t) will be used 
                                             // as alternative 
```

Given that 
The logistic function turns out to be trivially implementable (by which I meant without any explicit code to handle type differences), we can also chose to inline it in the Ops.cpp.

stablehlo/reference/Element.h
```diff
@@ -42,6 +42,7 @@ class Element {
   Element(Type type, std::complex<APFloat> value);
   Element(Type type, std::complex<double> value);
   Element(const Element &other) = default;
+  Element() = default;
```
sdasgup3, line 45: If we chose to go the route suggested in the comment above, we will no longer need to add this.

stablehlo/dialect/StablehloOps.td
```diff
@@ -433,7 +433,8 @@ def StableHLO_Log1pOp: StableHLO_UnaryElementwiseOp<"log_plus_one",
 }
 
 def StableHLO_LogisticOp: StableHLO_UnaryElementwiseOp<"logistic",
-    [Pure, HLO_CompatibleOperandsAndResultType], HLO_FpOrComplexTensor> {
+    [Pure, HLO_CompatibleOperandsAndResultType /*logistic_c1*/],
+    HLO_FpOrComplexTensor /*logistic_i1*/> { /*logistic_c1*/
```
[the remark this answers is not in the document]
ghpvnist: The updated guideline has to account for writing comments related to shape inference. Since this is hidden by the `StableHLO_UnaryElementwiseOp` class, we have another comment on line 437 as having them inside StableHLO_UnaryElementwiseOp would make it messy. The ops you mentioned (`LogOp` and `AbsOp`) were checked in before the updated guideline, so there is a discrepancy which would be addressed in #1002.

stablehlo/reference/Element.cpp
ghpvnist: Thanks for the suggestion. I agree with most of your suggestion but let's keep the implementation in Ops.cpp to be consistent with how we provide implementations for LogOp in the same file.

sdasgup3 (approved):

stablehlo/dialect/StablehloOps.td
```diff
@@ -431,8 +431,9 @@ def StableHLO_Log1pOp: StableHLO_UnaryElementwiseOp<"log_plus_one",
 }
 
 def StableHLO_LogisticOp: StableHLO_UnaryElementwiseOp<"logistic",
-    [Pure, HLO_CompatibleOperandsAndResultType], HLO_FpOrComplexTensor> {
-  let summary = "Logistic operation";
+    [Pure, HLO_CompatibleOperandsAndResultType /*logistic_c1*/],
+    HLO_FpOrComplexTensor /*logistic_i1*/> { /*logistic_c1*/
+  let summary = "Logistic operator";
```
burmako, line 436: This should be "Logistic operation" per the checklist.

stablehlo/tests/interpret_logistic.mlir
```diff
@@ -0,0 +1,20 @@
+// RUN: stablehlo-interpreter --interpret -split-input-file %s
+
+func.func @logistic_op_test_i64() {
```
burmako, line 3: Typo: f64.

stablehlo/reference/Element.cpp
```diff
@@ -190,16 +190,32 @@ Element::Element(Type type, bool value) : type_(type), value_(value) {}
 
 Element::Element(Type type, APFloat value) : type_(type), value_(value) {}
 
+// The double `value` can be used to construct two types: floating-point or
+// complex element types. By specifying a double value for complex element type,
+// the real part is set to `value`, and the imaginary part is zero.
```
burmako, line 195: Let's move this comment to Element.h.

burmako (changes requested):

stablehlo/dialect/StablehloOps.td
ghpvnist: Thanks for pointing it out! This is a common mistake in reviews, so I've updated the checklist example to explicitly show this.

burmako (approved):

PR #1243, opened 2023-03-01
Fix prettyprint bug with default optional attr in CholeskyOp

GleasonK:
Currently printing CholeskyOp with default value will cause a crash since the attribute is not assigned.

This disables the diff test in `legalize` since there are a few diffs that were hidden due to this crashing behavior. These will be addressed in a followup.


stablehlo/tests/stablehlo_legalize_to_vhlo.mlir
```diff
@@ -1,5 +1,5 @@
 // RUN: stablehlo-opt --stablehlo-legalize-to-vhlo --mlir-print-op-generic --split-input-file %s | FileCheck %s
-// RUN: diff <(stablehlo-opt --stablehlo-legalize-to-vhlo -vhlo-to-version='target=current' %s | stablehlo-opt --vhlo-legalize-to-stablehlo) <(stablehlo-opt %s)
+// RUN-diasbled: diff <(stablehlo-opt --stablehlo-legalize-to-vhlo -vhlo-to-version='target=current' %s | stablehlo-opt --vhlo-legalize-to-stablehlo) <(stablehlo-opt %s)
```
burmako, line 2: Typo: "disabled".

burmako (approved):

PR #1253, opened 2023-03-02
[CHLO] Add erf_inv

atondwal:


atondwal:
This ports over the change from tf https://github.com/tensorflow/tensorflow/blob/master/third_party/stablehlo/temporary.patch

stablehlo/dialect/ChloOps.td
```diff
@@ -673,6 +673,15 @@ def CHLO_ErfOp : CHLO_UnaryElementwiseOp<"erf",
...
 
+def CHLO_ErfInvOp : CHLO_UnaryElementwiseOp<"erf_inv",
+    [HLO_CompatibleOperandsAndResultType], HLO_FpTensor, HLO_FpTensor> {
+  let summary = "Inverse Erf";
+  let description = [{
+    Returns `ErfInv(operand)` element-wise.
+  }];
+}
+
+
```
GleasonK, line 684: Remove extra space

stablehlo/dialect/ChloOps.td
```diff
@@ -673,6 +673,15 @@ def CHLO_ErfOp : CHLO_UnaryElementwiseOp<"erf",
   }];
 }
 
+def CHLO_ErfInvOp : CHLO_UnaryElementwiseOp<"erf_inv",
+    [HLO_CompatibleOperandsAndResultType], HLO_FpTensor, HLO_FpTensor> {
```
GleasonK, line 677: Does this op support anything other than f32/f64? Should we verify type at all?

GleasonK (approved):

stablehlo/dialect/ChloOps.td
atondwal: In practice the lowering in mlir_hlo upcasts less wide types, but in this repo I don't think it makes sense to restrict it. Also, leaving it like this is consistent with the other CHLO ops

stablehlo/dialect/ChloOps.td
atondwal: Done

PR #1259, opened 2023-03-02
Treat VHLO as an implementation detail in compatibility.md

burmako:
This PR reformulates compatibility.md solely in terms of StableHLO to avoid exposing users to the implementation details of the serialization format.

There are dedicated document about those details - bytecode.md and vhlo.md - but the entire target audience of compatibility.md doesn't have to know about them.


GleasonK (approved):
Once we have serialization APIs we should mention them here

PR #1263, opened 2023-03-02
Update compatibility.md with a link to the VHLO cookbook

burmako:
Making final adjustments to the docs to make sure that they clearly state how to obtain portable artifacts.


GleasonK (approved):
Much better. Portable artifacts is better word

PR #1269, opened 2023-03-03
Use std::optional instead of llvm::Optional

atondwal:
Note that llvm::Optional is just an alias for std::optional these days and have since been deprecated upstream in favor of std::optional.

based on tensorflow@515b5d867bb901fd1c37911d7dfe6118ee3cc925


atondwal:
duplicate

PR #1271, opened 2023-03-03
Fix crash in typechecker

### LINKED ISSUE #1264: Typechecker assertion failure on mismatched ranks
### What happened?

Not totally sure how this runs into the stablehlo codebase, but if I try to build an op like `mhlo.and(tensor<2xi1>,tensor<i1>)`, lldb backtrace gives me a crash at https://github.com/openxla/stablehlo/blob/52a9462f42c53795dbbf2c16c10e80db02221b94/stablehlo/dialect/Base.cpp#L315

It looks like we're iterating over the rank of the first argument without checking the rank of the second.

I'm not sure what the invariants of this function are supposed to be.

atondwal:
fixes #1264

### DIFF
--- a/stablehlo/dialect/Base.cpp
+++ b/stablehlo/dialect/Base.cpp
@@ -300,6 +300,13 @@ FailureOr<TensorType> inferTypeWithCustomFn(
         Optional<Location>, int64_t, int64_t, int64_t, int64_t, int64_t)>
         inferDimAndBoundFn) {
   auto rank = rankedTypes[0].getRank();
+  for (auto& type : rankedTypes) {
+    if (type.getRank() != rank) {
+      return emitOptionalError(location, "Mismatched ranks of types",
+                               rankedTypes[0].getRank(), " vs ",
+                               type.getRank());
+    }
+  }
   SmallVector<int64_t> inferredSizes = to_vector(rankedTypes[0].getShape());
   SmallVector<int64_t> inferredBounds(rank, ShapedType::kDynamic);
   ArrayRef<int64_t> bounds = encodingToBounds(rankedTypes[0].getEncoding());


atondwal:
(Do we have negative tests for this sort of thing?)

stablehlo/dialect/Base.cpp
```diff
@@ -300,6 +300,7 @@ FailureOr<TensorType> inferTypeWithCustomFn(
         Optional<Location>, int64_t, int64_t, int64_t, int64_t, int64_t)>
         inferDimAndBoundFn) {
   auto rank = rankedTypes[0].getRank();
+  if (rankedTypes[1].getRank() != rank) return failure();
```
sdasgup3, line 303: Should we check for all  i `rankedTypes[i].getRank() != rank`

atondwal: Yep.

stablehlo/dialect/Base.cpp
```diff
@@ -300,6 +300,11 @@ FailureOr<TensorType> inferTypeWithCustomFn(
         Optional<Location>, int64_t, int64_t, int64_t, int64_t, int64_t)>
         inferDimAndBoundFn) {
   auto rank = rankedTypes[0].getRank();
+  for (auto& type : rankedTypes) {
+    if (type.getRank() != rank) {
+      return failure();
+    }
+  }
```
sdasgup3, line 307: Thanks!
Can we use the loop below to add the check and on failure, return with a diagnostic message. Something like (feel free to use other messages)
```
  return emitOptionalError(location, "Mismatched ranks of types", rankedTypes[0].getRank(), " vs ", rankedTypes[i].getRank());
```

stablehlo/dialect/Base.cpp
```diff
@@ -300,6 +300,13 @@ FailureOr<TensorType> inferTypeWithCustomFn(
         Optional<Location>, int64_t, int64_t, int64_t, int64_t, int64_t)>
         inferDimAndBoundFn) {
   auto rank = rankedTypes[0].getRank();
+  for (auto& type : rankedTypes) {
```
sdasgup3, line 303: Checking for such assertions early  makes sense (as opposed to my earlier suggestion about merging with the for loop)

A trade-off  for more conciseness at the cost of slight less verbose messaging. Feel free to pass on to the next reviewer in case we disagree. 

```
  if (llvm::any_of(rankedTypes, [&](RankedTensorType type) {
        return rank != type.getRank();
      }))
    return emitOptionalError(location, "Expected input types to have same ranks");
```

sdasgup3 (approved):

stablehlo/dialect/Base.cpp
atondwal: I think printing the ranks is helpful for fault localization

PR #1272, opened 2023-03-03
Add StableHLO portable artifact serialization APIs

### LINKED ISSUE #1262: Introduce dedicated APIs for serializing and deserializing portable artifacts
Let's encapsulate dealing with VHLO behind the APIs along the lines of `... serializePortableArtifact(mlir::ModuleOp module)` and `mlir::Module deserializePortableArtifact(...)`.

GleasonK:
Add serialization APIs:

```c++
// Write a StableHLO program to a portable artifact
LogicalResult serializePortableArtifact(ModuleOp module,
                                        StringRef targetVersion,
                                        raw_ostream& os);

// Read StableHLO portable artifact
OwningOpRef<ModuleOp> deserializePortableArtifact(StringRef sourceStr,
                                                  MLIRContext* context);
```

This is exposed via our translator tool `stablehlo-interpret`. I'll rename this tool in a followup to `stablehlo-translate` since that is more in-model, wanted to separate that fix to keep this PR concise.

```
stablehlo-translate --serialize --target=0.9.0 file.mlir > file.mlir.bc
stablehlo-translate --deserialize file.mlir.bc
```

This creates an MLIR Bytecode File with the header `VHLO_serializerVerion_targetVersion`. I.e. for a `0.11.0` libStablehlo targeting `0.9.0`, the bytecode header reads: `VHLO_0.11.0_0.9.0`.

Still todo:
- Rename translate tool
- Update serialization tests to use `stablehlo-translate --deserialize`
- Expose these API's via Python

Closes #1262


stablehlo/tools/StablehloInterpreterMain.cpp
```diff
@@ -76,6 +81,30 @@ TranslateFromMLIRRegistration stablehlo_interpreter(
       registry.insert<stablehlo::StablehloDialect>();
     });
 
+llvm::cl::opt<std::string> _target(
+    "target", llvm::cl::desc("Target version for serialization"),
+    llvm::cl::init(""));
+
+TranslateFromMLIRRegistration stablehlo_serialize(
+    "serialize", "Serialize StableHLO program into portable artifact",
```
burmako, line 89: "a portable artifact", to be consistent with deserialize?

stablehlo/tools/StablehloInterpreterMain.cpp
```diff
@@ -76,6 +81,30 @@ TranslateFromMLIRRegistration stablehlo_interpreter(
       registry.insert<stablehlo::StablehloDialect>();
     });
 
+llvm::cl::opt<std::string> _target(
```
burmako, line 84: "targetOption"?

stablehlo/tools/StablehloInterpreterMain.cpp
```diff
@@ -76,6 +81,30 @@ TranslateFromMLIRRegistration stablehlo_interpreter(
...
+    "target", llvm::cl::desc("Target version for serialization"),
+    llvm::cl::init(""));
+
+TranslateFromMLIRRegistration stablehlo_serialize(
```
burmako, line 88: How about "serializeRegistration"? Similarly, for "interpretRegistration" and "deserializeRegistration".

stablehlo/tools/CMakeLists.txt
```diff
@@ -49,6 +50,8 @@ target_link_libraries(stablehlo-interpreter PRIVATE
   StablehloReferenceOps
   StablehloReferenceScope
   StablehloReferenceTensor
+  StablehloRegister
+  StablehloSerialization
```
burmako, line 54: `VhloOps` as well?

stablehlo/dialect/Serialization.h
```diff
@@ -0,0 +1,38 @@
+/* Copyright 2022 The StableHLO Authors.
```
burmako, line 1: 2023

stablehlo/dialect/Serialization.h
```diff
@@ -0,0 +1,38 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#ifndef STABLEHLO_DIALECT_ASSEMBLYFORMAT_H
+#define STABLEHLO_DIALECT_ASSEMBLYFORMAT_H
```
burmako, line 17: `STABLEHLO_DIALECT_SERIALIZATION_H`.

stablehlo/dialect/Serialization.h
```diff
@@ -0,0 +1,38 @@
...
+#include "mlir/IR/BuiltinOps.h"
+#include "mlir/IR/MLIRContext.h"
+#include "mlir/Support/LogicalResult.h"
+
+namespace mlir {
+namespace stablehlo {
+
+// Write a StableHLO program to a portable artifact
+LogicalResult serializePortableArtifact(ModuleOp module,
+                                        std::string const& targetVersion,
```
burmako, line 28: `StringRef` would be more idiomatic.

stablehlo/dialect/Serialization.h
```diff
@@ -0,0 +1,38 @@
...
+                                        raw_ostream& os);
+
+// Read StableHLO portable artifact
+mlir::OwningOpRef<mlir::ModuleOp> deserializePortableArtifact(
+    llvm::StringRef sourceStr, MLIRContext* context);
```
burmako, line 33: The usage of namespaces in this file is fairly inconsistent. How about we drop `mlir::` and `llvm::` everywhere in this file?

BUILD.bazel
```diff
@@ -722,6 +722,29 @@ cc_library(
...
+cc_library(
+    name = "stablehlo_serialization",
+    srcs = [
+        "stablehlo/dialect/Serialization.cpp",
+    ],
+    hdrs = [
+        "stablehlo/dialect/Serialization.h",
+    ],
+    deps = [
+        ":chlo_ops",
```
burmako, line 734: Not sure why CHLO is used here (and in the CMake build).

stablehlo/dialect/Serialization.cpp
```diff
@@ -0,0 +1,89 @@
...
+  headerOs << "VHLO_" << vhlo::Version::getCurrentVersion() << "_"
+           << vhlo::Version::fromString(targetVersion);
+  BytecodeWriterConfig config(artifactHeader);
+  writeBytecodeToFile(module, os, config);
+  return success();
+}
+
+mlir::OwningOpRef<mlir::ModuleOp> deserializePortableArtifact(
+    llvm::StringRef sourceStr, MLIRContext* context) {
+  // FIXME: Not sure if the lifetime is correct here.
```
burmako, line 71: What are you not sure about? Let's discuss.

stablehlo/dialect/Serialization.cpp
```diff
@@ -0,0 +1,89 @@
...
+  // target version failures.
+  {
+    mlir::PassManager pm(context);
+    pm.addPass(mlir::stablehlo::createVhloToVersionPass({targetVersion}));
+    if (!mlir::succeeded(pm.run(module))) {
+      return failure();
+    }
+  }
+
+  // Writes the header VHLO_<serializerVersion>_<targetVersion>
```
burmako, line 59: What is the use case for storing this information, especially since it's not part of the deserialization API?

stablehlo/tests/stablehlo_legalize_to_vhlo.0_9_0.mlir
```diff
@@ -1,6 +1,6 @@
 // RUN: stablehlo-opt --mlir-print-op-generic %s.bc | FileCheck %s
-// RUN: stablehlo-opt --vhlo-to-version='target=current' --vhlo-legalize-to-stablehlo --stablehlo-legalize-to-vhlo --vhlo-to-version='target=0.9.0' --mlir-print-op-generic %s.bc | FileCheck %s
-// RUN: diff <(stablehlo-opt --vhlo-to-version='target=current' --vhlo-legalize-to-stablehlo %s.bc) <(stablehlo-opt %s)
+// RUN: stablehlo-interpreter --deserialize %s.bc | stablehlo-interpreter --serialize --target=0.9.0 | stablehlo-opt --mlir-print-op-generic | FileCheck %s
+// RUN: diff <(stablehlo-interpreter --deserialize %s.bc | stablehlo-opt ) <(stablehlo-opt %s)
```
burmako, line 3: Extraneous whitespace after `| stablehlo-opt`.

stablehlo/dialect/Serialization.h
```diff
@@ -0,0 +1,38 @@
...
+
+}  // namespace stablehlo
+}  // namespace mlir
+
+#endif  // STABLEHLO_DIALECT_ASSEMBLYFORMAT_H
```
burmako, line 38: Let's also update documentation?
  1) compatibility.md to point to this file instead of the VHLO cookbook (so that VHLO stays an implementation detail from the point of view of compatibility.md).
  2) vhlo.md to use `stablehlo-interpreter` instead of `stablehlo-opt`.

burmako (changes requested):
This is great! A satisfying conclusion to the work on hiding VHLO from the majority of users.

stablehlo/dialect/Serialization.cpp
GleasonK: It will help with resolving bugs if we know the serializer version. Target version is helpful as a repro aid.

Some email feedback I received recently:

> One other thought I had is how easy will be it to debug compatibility problems. Wouldn't it be useful for the serialized format to contain the date and code version of the serializer? Alternatively, we can have a table with the dates at which different VHLO versions were introduced, to make it easy to see if something was serialized too far in the past.

While I don't like storing the date (two artifacts serialized the same way by the same version should probably create a deterministic file), I think we should store some plaintext readable information for debugging.

stablehlo/dialect/Serialization.cpp
GleasonK: Figured this out, forgot to delete the fixme. Wasn't sure on lifetime of an owning ref back when this API didn't return the owning ref. Now this is in-line with some upstream APIs.

stablehlo/dialect/Serialization.h
GleasonK: Lets talk more about this. I updated (2), but given the suggested update to (2) I think that updating (1) would
result in providing less useful information, or redundant information.

burmako: Sounds good.

burmako (changes requested):
How about we factor out writing a header into a separate pull request? This has compatibility implications that I'd like to think about and discuss further. In contrast, everything else in the PR is already mergeable in my opinion.

stablehlo/dialect/Serialization.cpp
GleasonK: Removed header for now

burmako (approved):

PR #1273, opened 2023-03-03
Fix typo in ReverseOp's spec example

ghpvnist:


docs/spec.md
```diff
@@ -4237,7 +4237,7 @@ and produces a `result` tensor. More formally,
 ```mlir
 // %operand = [[1, 2], [3, 4], [5, 6]]
 %result = "stablehlo.reverse"(%operand) {
-  dimensions = dense<1> : tensor<i64>
+  dimensions = dense<1> : tensor<1xi64>
```
sdasgup3, line 4240: This is correct right? 
`<[1]>` would be of type `tensor<1xi64>`

ghpvnist: `dimensions` should be a rank-1 attribute, and `dense<1> : tensor<1xi64>` is a syntax sugar for dense tensors. E.g. `dense<1> : tensor<2xi64>` is equivalent to `dense<[1, 1]> : tensor<2xi64>`

sdasgup3: Ah! I didn't read into the fact that it is reverse op and surprised why we want to change the type. Anyway, never mind then.

sdasgup3 (approved):

burmako (approved):

PR #1274, opened 2023-03-04
Improve newcomer experience

burmako:
I've tried to follow our build instructions on my personal macbook to see how smooth the experience is and how onerous it is to run the testdata/ suite on a reasonably mainstream machine.

Turns out that our instructions don't really work on macOS, and predictably checking the entire testdata/ takes a considerable amount of time (20+ seconds on a 2021 M1 Pro macbook).

This PR fixes the build instructions to work on macOS and updates the CMake target called out in the readme to only check the tests/ suite.

### DIFF
--- a/build_tools/build_mlir.sh
+++ b/build_tools/build_mlir.sh
@@ -37,11 +37,12 @@ mkdir -p "$build_dir"
 echo "Beginning build (commands will echo)"
 set -x
 
+[[ uname != "Darwin" ]] && LLVM_ENABLE_LLD="ON" || LLVM_ENABLE_LLD="OFF"
 cmake -GNinja \
   "-H$LLVM_SRC_DIR/llvm" \
   "-B$build_dir" \
   -DLLVM_INSTALL_UTILS=ON \
-  -DLLVM_ENABLE_LLD=ON \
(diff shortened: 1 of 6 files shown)


build_tools/build_mlir.sh
```diff
@@ -51,4 +52,4 @@ cmake -GNinja \
   -DCMAKE_BUILD_TYPE=RelWithDebInfo \
   -DLLVM_ENABLE_ASSERTIONS=On
 
-cmake --build "$build_dir" --target all --target mlir-cpu-runner
```
GleasonK, line 54: Any clue why this was here? Are we able to target mlir-cpu-runner? Will that build a more specific targets that improves build speed?

GleasonK (approved):
This is much better. Very happy to see the more specific check command in our docs.

build_tools/build_mlir.sh
burmako: No idea. I think we inherited this from MLIR-HLO build where it was perhaps load-bearing, but we aren't using it in our repo. Also, due to some reason, this target didn't work on my macbook, so I figured that it would be useful to remove it.

PR #1277, opened 2023-03-05
Significantly compress syntax for bounded dynamism

burmako:
Before:

```
  tensor<?x4xf32, #stablehlo.type_extensions<bounds = [4, ?]>>
```

After:

```
  tensor<?x4xf32, #stablehlo.bounds<4, ?>>
```

Old syntax is still accepted for compatibility reasons.

Additionally, this PR makes a few cleanups which I noticed when working on it:
  1) TypeInference.h now takes HloDialectInterface* instead of Dialect*
     to increase type safety, similarly to newly introduced functions
     in AssemblyFormat.h.
  2) parseAttribute/parseType error messages were changed to say
     "StableHLO" instead of "stablehlo", to use the canonical name
     of the StableHLO dialect.


stablehlo/tests/print_stablehlo.mlir
```diff
@@ -281,10 +281,10 @@ func.func @fft_op(%arg0: tensor<16xcomplex<f32>>) -> tensor<16xcomplex<f32>> {
 }
 
 // CHECK-LABEL: func @extensions
-func.func @extensions(%arg0 : tensor<?x?xf32, #stablehlo.type_extensions<bounds = [3, ?]>>,
+func.func @extensions(%arg0 : tensor<?x?xf32, #stablehlo.bounds<3, ?>>,
```
burmako, line 284: We could also consider `tensor<?x?xf32, #stablehlo.bounds<3x?>>` to be consistent with how dimension sizes are prettyprinted. I'm on the fence - really don't know which one I prefer.

stablehlo/dialect/AssemblyFormat.cpp
```diff
@@ -371,5 +368,28 @@ ParseResult parseCustomCallTarget(AsmParser& parser, StringAttr& target) {
...
+  auto printBound = [&](int64_t bound) { os << dimSizeToString(bound); };
+  llvm::interleaveComma(attr.getBounds(), os, printBound);
+  os << ">";
+}
+
+Attribute parseTypeExtensions(HloDialectInterface* dialect,
+                              DialectAsmParser& parser) {
+  SmallVector<int64_t> bounds;
+  auto parseBound = [&]() -> ParseResult {
+    if (!parser.parseOptionalQuestion()) {
```
GleasonK, line 382: `succeeded(parser.parseOptionalQuestion())`?

stablehlo/tests/print_stablehlo.mlir
```diff
@@ -281,10 +281,10 @@ func.func @fft_op(%arg0: tensor<16xcomplex<f32>>) -> tensor<16xcomplex<f32>> {
...
                 %arg1 : tensor<i32>) -> () {
-  // CHECK:      %0 = stablehlo.set_dimension_size %arg0, %arg1, dim = 1 : (tensor<?x?xf32, #stablehlo.type_extensions<bounds = [3, ?]>>, tensor<i32>) -> tensor<*xf32>
-  %0 = "stablehlo.set_dimension_size"(%arg0, %arg1) {dimension = 1 : i64} : (tensor<?x?xf32, #stablehlo.type_extensions<bounds = [3, ?]>>, tensor<i32>) -> tensor<*xf32>
+  // CHECK:      %0 = stablehlo.set_dimension_size %arg0, %arg1, dim = 1 : (tensor<?x?xf32, #stablehlo.bounds<3, ?>>, tensor<i32>) -> tensor<*xf32>
```
GleasonK, line 286: Could you add a test that generic syntax gets parsed properly?

```
// CHECK: %0 = stablehlo.set_dimension_size %arg0, %arg1, dim = 1 : (tensor<?x?xf32, #stablehlo.bounds<3, ?>>, tensor<i32>) -> tensor<*xf32>
// CHECK: %1 = stablehlo.set_dimension_size %arg0, %arg1, dim = 1 : (tensor<?x?xf32, #stablehlo.bounds<3, ?>>, tensor<i32>) -> tensor<*xf32>
%0 = "stablehlo.set_dimension_size"(%arg0, %arg1) {dimension = 1 : i64} : (tensor<?x?xf32, #stablehlo.bounds<3, ?>>, tensor<i32>) -> tensor<*xf32>
%1 = "stablehlo.set_dimension_size"(%arg0, %arg1) {dimension = 1 : i64} : (tensor<?x?xf32, #stablehlo.type_extensions<bounds = [3, ?]>>, tensor<i32>) -> tensor<*xf32>
```

stablehlo/tests/print_stablehlo.mlir
```diff
@@ -281,10 +281,10 @@ func.func @fft_op(%arg0: tensor<16xcomplex<f32>>) -> tensor<16xcomplex<f32>> {
...
+  %0 = "stablehlo.set_dimension_size"(%arg0, %arg1) {dimension = 1 : i64} : (tensor<?x?xf32, #stablehlo.bounds<3, ?>>, tensor<i32>) -> tensor<*xf32>
   "stablehlo.return"() : () -> ()
```
GleasonK, line 288: I tend to find bugs in similar list-printing functions when they are tested with empty input. Given that the following is a StableHLO program that passes verification, it may be worth a test that this doesn't crash or gives reasonable output (or adding a verifier check if this is invalid):

```
module {
  func.func @test(%arg0: tensor<f32, #stablehlo.type_extensions<bounds = []>>) {
    return
  }
}
```

This may fail with new syntax:

```
module {
  func.func @test(%arg0: tensor<f32, #stablehlo.bounds<>>) {
    return
  }
}
```

GleasonK (changes requested):
I understand the benefit, but not sure I love the different printing formats for type extensions. I.e. if we ever add something to `type_extensions`, the current format doesn't make sense, as it seems like we have an attribute named `bounds`. A more scalable syntax would be:

```
type_extensions<[bounds] (, otherFeature)*>
```

We would still omit `bounds =` since it is largely unambiguous. Given that there is no stability in assembly we can fix this at any time, so this is ok for now and we can revisit this in the future.

stablehlo/dialect/AssemblyFormat.cpp
burmako: Done, also refactored the code that this was copypasted from.

GleasonK (approved):

PR #1283, opened 2023-03-07
Add interpreter for SortOp

### LINKED ISSUE #991: Add interpreter for sort

ghpvnist:

```
(I1) inputs: variadic number of tensors.
(I2) dimension: constant of type `si64`.
(I3) is_stable: constant of type `i1`.
(I4) comparator: function.
(C1) `inputs` have at least 1 tensor.
(C2) For all `i`, `type(inputs[i])` = `type(results[i])`.
(C3) All tensors in `inputs` and `results` have the same shape.
(C4) `-R` $\le$ `dimension` $\lt$ `R`, where `R` is rank of `inputs[0]`.
(C5) `comparator` has type
`(tensor<E1>, tensor<E1>, ..., tensor<EN-1>, tensor<EN-1>) -> tensor<i1>`,
where `Ei` is element type of `inputs[i]`.
```

```
I1: a) inputs is not variadic tensor. (Covered by ODS).
I2: a) element_type(dimension) != `si64`. (Covered by ODS).
I3: a) is_stable is not a constant of type `i1`. (Covered by ODS).
I4: a) comparator is not a function. (Covered by ODS).
C1: a) size(inputs) < 1. (Covered by ODS).
C2: a) For any `i`, `type(inputs[i])` != `type(results[i])`.
C3: a) Any tensors in `inputs` and `results` do not have the same shape. (Covered by ODS).
C4: a) `dimension` < `-R` where `R` is rank of `inputs[0]`.
    b) `dimension` >= `R`, where `R` is rank of `inputs[0]`.
C5: a) `comparator` does not have type
`(tensor<E1>, tensor<E1>, ..., tensor<EN-1>, tensor<EN-1>) -> tensor<i1>`,
where `Ei` is element type of `inputs[i]`.
```

```
C2a: For any `i`, `type(inputs[i])` != `type(results[i])`.
C4a: `dimension` < `-R` where `R` is rank of `inputs[0]`.
C4b: `dimension` >= `R`, where `R` is rank of `inputs[0]`.
C5a: `comparator` does not have type
`(tensor<E1>, tensor<E1>, ..., tensor<EN-1>, tensor<EN-1>) -> tensor<i1>`,
where `Ei` is element type of `inputs[i]`.
```

Notes:
* Simplified spec wording from "a variadic number of tensors in `foo`" to just "`foo`" for brevity.
* Clarified the semantics to describe the op's behavior in more detail.

closes #991


ghpvnist:
This PR also contains temporary commit rebased from #1199. You may ignore the following files for now, and partially review the changes related to SortOp for others:
root:
* BUILD.bazel

`stablehlo/reference/`:
* Element.cpp
* Element.h
* Ops.h

`stablehlo/tests/`:
* interpret_compare.mlir
* print_stablehlo.mlir
* stablehlo_legalize_to_vhlo.mlir

stablehlo/reference/Ops.cpp
```diff
@@ -372,6 +403,56 @@ Tensor evalSliceOp(const Tensor &operand, Index startIndices, Sizes strides,
...
+SmallVector<Tensor> evalSortOp(ArrayRef<Tensor> inputs, Axis dimension,
+                               bool isStable, Region &comparator,
+                               Scope &scope) {
+  SmallVector<Tensor> runtimeResults;
+  for (auto input : inputs) runtimeResults.push_back(Tensor(input.getType()));
+  if (dimension < 0) dimension += inputs[0].getRank();
+  SmallVector<int64_t> indices(inputs.front().getShape()[dimension]);
+  auto inputShape = inputs.front().getShape();
+  inputShape[dimension] = 1;
+  auto inputSizes = Sizes(inputShape);
```
sdasgup3, line 415: `inputShape` is sufficient, no need for `inputSizes`. 
We can re-write it as:

```
  Sizes inputShape(inputs.front().getShape());
  inputShape[dimension] = 1;
  for (auto offsetIt = IndexSpaceIterator(inputShape, Index(inputShape.size()));
       offsetIt != IndexSpaceIterator(inputShape, std::nullopt); ++offsetIt) {
```

stablehlo/reference/Ops.cpp
```diff
@@ -372,6 +403,56 @@ Tensor evalSliceOp(const Tensor &operand, Index startIndices, Sizes strides,
...
+  for (auto resultIt = IndexSpaceIterator(inputSizes, Index(inputShape.size()));
```
sdasgup3, line 416: Do no call the iterator as `resultIt` as this is not exactly result index at this point but a partial index with the dimension axis clipped to 0. We can use the `offsetIt` instead.

stablehlo/reference/Ops.cpp
```diff
@@ -372,6 +403,56 @@ Tensor evalSliceOp(const Tensor &operand, Index startIndices, Sizes strides,
...
+    std::iota(indices.begin(), indices.end(), 0);
+    auto compare = [&](int64_t lhs, int64_t rhs) {
+      SmallVector<Tensor> args;
+      auto lhsIndex = *resultIt;
+      auto rhsIndex = *resultIt;
+      lhsIndex[dimension] = lhs;
+      rhsIndex[dimension] = rhs;
+      for (auto input : inputs) {
+        auto type =
+            comparator.front().getArgumentTypes()[0].dyn_cast<TensorType>();
```
sdasgup3, line 427: The arg type of an inputs[i] is not always equal to the type of the first bock argument but depends on which input[i] we are referring to. Note that the argument type for inputs[i] and inputs[i+1] could be different.

In this case type should be equal to something like  `comparator.front().getArgumentTypes()[2*input_index].dyn_cast<TensorType>();`

stablehlo/reference/Ops.cpp
```diff
@@ -372,6 +403,56 @@ Tensor evalSliceOp(const Tensor &operand, Index startIndices, Sizes strides,
...
+      }
+      auto cmpResult = eval(comparator, args, &scope);
+      return cmpResult[0].get({}).getBooleanValue();
+    };
+
+    if (isStable)
+      std::stable_sort(indices.begin(), indices.end(), compare);
+    else
+      std::sort(indices.begin(), indices.end(), compare);
+    auto resultIdx = *resultIt;
```
sdasgup3, line 443: We can be simplified and improve readability if we keep separate copy for resultIdx and operandIdx. 
PTAL suggestion below

stablehlo/reference/Ops.cpp
```diff
@@ -372,6 +403,56 @@ Tensor evalSliceOp(const Tensor &operand, Index startIndices, Sizes strides,
...
+        resultIdx[dimension] = src;
+        Element resultElement = input.get(resultIdx);
+        resultIdx[dimension] = static_cast<int64_t>(dst);
+        result.set(resultIdx, resultElement);
+      }
+    }
+  }
+  return runtimeResults;
+}
+
```
sdasgup3, line 455: Here is your code with the suggestion mentioned above included.

```
SmallVector<Tensor> evalSortOp(ArrayRef<Tensor> inputs, Axis dimension,
                               bool isStable, Region &comparator,
                               Scope &scope) {
  SmallVector<Tensor> runtimeResults;
  for (auto input : inputs) runtimeResults.push_back(Tensor(input.getType()));

  if (dimension < 0) dimension += inputs[0].getRank();

  Index indices(inputs.front().getShape()[dimension]);

  Sizes inputShape(inputs.front().getShape());
  inputShape[dimension] = 1;
  for (auto offsetIt = IndexSpaceIterator(inputShape, Index(inputShape.size()));
       offsetIt != IndexSpaceIterator(inputShape, std::nullopt); ++offsetIt) {
    std::iota(indices.begin(), indices.end(), 0);

    auto compare = [&](int64_t lhs, int64_t rhs) {
      SmallVector<Tensor> args;
      auto lhsIndex = *offsetIt;
      auto rhsIndex = *offsetIt;
      lhsIndex[dimension] = lhs;
      rhsIndex[dimension] = rhs;
      for (const auto &indexedInput : llvm::enumerate(inputs)) {
        auto type =
            comparator.front().getArgumentTypes()[2*indexedInput.index()].dyn_cast<TensorType>();
        auto lhs = Tensor(type);
        auto rhs = Tensor(type);
        lhs.set({}, indexedInput.value().get(lhsIndex));
        rhs.set({}, indexedInput.value().get(rhsIndex));
        args.push_back(lhs);
        args.push_back(rhs);
      }
      auto cmpResult = eval(comparator, args, &scope);
      return cmpResult[0].get({}).getBooleanValue();
    };

    if (isStable)
      std::stable_sort(indices.begin(), indices.end(), compare);
    else
      std::sort(indices.begin(), indices.end(), compare);

    auto resultIdx = *offsetIt;
    auto operandIdx = *offsetIt;
    for (auto [dst, src] : llvm::enumerate(indices)) {
      for (auto [input, result] : llvm::zip(inputs, runtimeResults)) {
        operandIdx[dimension] = src;
        resultIdx[dimension] = static_cast<int64_t>(dst);
        result.set(resultIdx, input.get(operandIdx));
      }
    }
  }
  return runtimeResults;
}
```

ghpvnist: Thanks for the summary!

stablehlo/tests/interpret_sort.mlir
```diff
@@ -0,0 +1,35 @@
+// RUN: stablehlo-translate --interpret -split-input-file %s
```
sdasgup3, line 1: Can we re-write the example in such a way that sort and stable-sort would differ? With our offline discussion it seem you have a plan. Let's try that.

sdasgup3 (approved):
lgtm other than one comment about testing.

stablehlo/tests/interpret_sort.mlir
sdasgup3: As I am thinking more about this:
If the solution to this involves sorting different representations of NaN (using partial order), then we might avoid (also ignore my comment above) that mainly because seems more of a testing-interesting-feature to me where we have to rely on comparison semantics of  NaN. Code-coverage wise we will not gain anything.

ghpvnist: There is no way to sort primitive types to tell the difference. Since we didn't implement `TOTALORDER` in CompareOp, we also cannot sort between different representations of NaNs. Sorting should (rightfully) be broken with NaNs as we are using unordered comparisons.

stablehlo/tests/infer_stablehlo.mlir
```diff
@@ -418,6 +418,18 @@ func.func @sort(%input0: tensor<16x16xf32>, %input1: tensor<16x16xi32>) -> tenso
 
 // -----
 
+func.func @sort_c2(%input0: tensor<16x16xf32>, %input1: tensor<16x16xi32>) {
```
burmako, line 421: Similarly to while, I think we should drop this test case from this file.

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -2641,106 +2641,58 @@ func.func @sort(%input0: tensor<16x16xf32>, %input1: tensor<16x16xi32>) {
...
+  }) {dimension = 1 : i64, is_stable = true} : (tensor<*xf32>, tensor<16x16xi32>) -> (tensor<16x16xf32>, tensor<16x16xi32>)
   func.return
 }
 
 // -----
 
-func.func @sort_wrong_block_arg_type(%input0: tensor<16x16xf32>, %input1: tensor<16x16xi32>) {
-  // expected-error @+1 {{comparator block argument #3 should be of type 'tensor<i32>' but got 'tensor<f32>'}}
+func.func @sort_c5(%input0: tensor<16x16xf32>, %input1: tensor<16x16xi32>) {
+  // expected-error @+1 {{comparator block argument #2 should be of type 'tensor<i32>' but got 'tensor<f32>'}}
```
burmako, line 2693: This test seems to be testing the same thing as the test above? "Argument #N should be of type Foo".

stablehlo/reference/Ops.h
```diff
@@ -83,6 +83,8 @@ Tensor evalSelectOp(const Tensor &pred, const Tensor &onTrue,
 Tensor evalSineOp(const Tensor &operand, TensorType resultType);
 Tensor evalSliceOp(const Tensor &operand, Index startIndices, Sizes strides,
                    TensorType resultType);
+SmallVector<Tensor> evalSortOp(ArrayRef<Tensor> inputs, Axis dimension,
+                               bool isStable, Region &comparator, Scope &scope);
```
burmako, line 87: Should the signature also include `resultTypes`, similarly to all other eval functions in this file? Same question for `evalWhileOp`.

burmako (changes requested):
Will review Ops.cpp later today.

stablehlo/reference/Ops.cpp
```diff
@@ -423,6 +426,55 @@ Tensor evalSliceOp(const Tensor &operand, Index startIndices, Sizes strides,
   return result;
 }
 
+SmallVector<Tensor> evalSortOp(ArrayRef<Tensor> inputs, Axis dimension,
+                               bool isStable, Region &comparator,
+                               Scope &scope) {
```
burmako, line 431: Provided feedback for this function in a pull request: https://github.com/ghpvnist/stablehlo/pull/2. Overall, the logic looks reasonable, but I had some comments on framing, based on what confused me when I saw the code for the first time.

burmako (changes requested):

stablehlo/reference/Ops.h
ghpvnist: I've been thinking about this some time back, but decided it's not necessary to include it since there is no immediate need for this parameter. Though it does not align perfectly with the rest, but for example, IfOp, CastOp, and ConstantOp don't have `resultType` either, and we haven't run into issues so far.

stablehlo/tests/interpret_sort.mlir
```diff
@@ -0,0 +1,35 @@
...
+      %predicate = stablehlo.compare GT, %arg0, %arg1 : (tensor<i64>, tensor<i64>) -> tensor<i1>
+      stablehlo.return %predicate : tensor<i1>
+  }) {
+    dimension = 0 : i64,
+    is_stable = false
+  } : (tensor<2x3xi64>, tensor<2x3xi64>) -> (tensor<2x3xi64>, tensor<2x3xi64>)
+  check.expect_eq_const %result0, dense<[[3, 2, 3], [1, 2, 1]]> : tensor<2x3xi64>
+  check.expect_eq_const %result1, dense<[[1, 2, 1], [3, 2, 3]]> : tensor<2x3xi64>
+  func.return
+}
```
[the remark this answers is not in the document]
ghpvnist: No tests can be enabled as of now.

stablehlo/tests/ops_stablehlo.mlir
ghpvnist: Yes, you're right. First test is checking that input is of correct type, and the second test was meant to check that init_value is also of correct type, but the code checks for both so this test is redundant. Removed.

stablehlo/tests/interpret_sort.mlir
```diff
@@ -0,0 +1,35 @@
...
+    is_stable = true
+  } : (tensor<2x3xi64>, tensor<2x3xi64>) -> (tensor<2x3xi64>, tensor<2x3xi64>)
+  check.expect_eq_const %result0, dense<[[3, 2, 3], [1, 2, 1]]> : tensor<2x3xi64>
+  check.expect_eq_const %result1, dense<[[1, 2, 1], [3, 2, 3]]> : tensor<2x3xi64>
+  func.return
+}
+
+// -----
+
+func.func @sort_unstable() {
```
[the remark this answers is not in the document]
ghpvnist: Opened #1412 to track this.

burmako (approved):
LGTM, but let's double-check with @cheshire on the wording of the spec given the discussion from earlier today.

PR #1290, opened 2023-03-07
Add reference interpreter tests to compatibility suite

GleasonK:
Spoke with @sdasgup3 / @ghpvnist about submitting these tests. Decided integer interpreter tests are safe to submit now, will keep FP tests filtered until we figure out proper FP thresholds.

- V0: Filter all tests using `RUN-DISABLED:`
- V1(now): Unfilter integer and boolean tests for supported ops. Filter supported op FP tests using `RUN-DISABLED(#1278):`
- V2(after #1278): Unfilter floating point tests for supported ops.
- Vfuture: Spec workstream will decide on process for unfiltering TPs for newly supported ops.

Logfiles with existing failures can be found at commit: https://github.com/openxla/stablehlo/commit/b8875e16dace1cfac8691f8a223974ce785da973 (these have been removed before submit, but commit should live on)

PR #1292, opened 2023-03-07
Refactor CustomCallOp evaluator

sdasgup3:
This PR is to refactor the custom-call eval function which has been recently added to improve interpreter testing (#1290).

### DIFF
--- a/stablehlo/tools/StablehloTranslateMain.cpp
+++ b/stablehlo/tools/StablehloTranslateMain.cpp
@@ -42,8 +42,7 @@ stablehlo::Tensor makeBooleanTensor(MLIRContext *context, bool value) {
 }
 
 llvm::Error evalCustomCallCheckEq(stablehlo::CustomCallOp op,
-                                  stablehlo::Scope &scope,
-                                  llvm::StringRef funcName) {
+                                  stablehlo::Scope &scope) {
   if (op->getNumOperands() != 2) {
     return stablehlo::invalidArgument("Unsupported op: %s",
                                       debugString(op).c_str());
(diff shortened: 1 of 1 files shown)


ghpvnist (approved):

stablehlo/tools/StablehloTranslateMain.cpp
```diff
@@ -56,49 +55,87 @@ llvm::Error evalCustomCallCheckEq(stablehlo::CustomCallOp op,
...
   if (auto customCall = dyn_cast<stablehlo::CustomCallOp>(op)) {
-    if (customCall.getCallTargetName() == "check.eq")
-      return evalCustomCallCheckEq(customCall, scope, funcName);
+    if (customCall.getCallTargetName() == "check.eq") {
+      auto status = evalCustomCallCheckEq(customCall, scope);
+      if (status)
+        return stablehlo::invalidArgument(
+            "Error evaluating function: %s. \n\tCheck eq failed: %s",
+            funcName.data(), toString(std::move(status)).c_str());
+      return llvm::Error::success();
```
burmako, line 75: Can we create a helper for this block of code?

```C++
if (status)
    return stablehlo::invalidArgument(
        "Error evaluating function: %s. \n\tCheck eq failed: %s",
        funcName.data(), toString(std::move(status)).c_str());
  return llvm::Error::success();
```

It seems to be repeated 5 times in this file, which is a bit too much I think.

```C++
llvm::Error wrapStatus(llvm::Error status, llvm::StringRef funcName,
                       llvm::StringRef fallbackName) {
  if (status)
    return stablehlo::invalidArgument(
        "Error evaluating function: %s. \n\tFallback for %s failed: %s",
        funcName.data(), fallbackName.data(),
        toString(std::move(status)).c_str());
  return llvm::Error::success();
}
```

burmako (changes requested):

stablehlo/tools/StablehloTranslateMain.cpp
GleasonK: +1

stablehlo/tools/StablehloTranslateMain.cpp
```diff
@@ -56,49 +55,87 @@ llvm::Error evalCustomCallCheckEq(stablehlo::CustomCallOp op,
...
 
-  scope.add(op.getResults(), makeBooleanTensor(op->getContext(), true));
-  return llvm::Error::success();
+  return status;
 }
 
 llvm::Error interpreterFallback(Operation &op, stablehlo::Scope &scope,
                                 llvm::StringRef funcName) {
-  // TODO: Write a pass to inline this custom call into a check op.
+  // custom_call(...) implementation
```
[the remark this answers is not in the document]
GleasonK: This also is a fallback for more ops than just custom_call, correct? Unimplemented ops as well?

PR #1297, opened 2023-03-08
Update how pad is used in spec

sdasgup3:
Fix a bug in how pad is used in specification.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -1884,7 +1884,7 @@ If `feature_group_count = 1` and `batch_group_count = 1`, then for all
 `output_spatial_index` in the index space of `dim(result, output_spatial_dimensions)`,
 `result[result_shape(:, output_spatial_index, :)] = dot_product` where:
 
-* `padded_lhs = pad(lhs, 0, lhs_padding[:, 0], lhs_padding[:, 1], lhs_base_dilations - 1)`.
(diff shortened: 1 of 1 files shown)


docs/spec.md
```diff
@@ -1884,7 +1884,7 @@ If `feature_group_count = 1` and `batch_group_count = 1`, then for all
 `output_spatial_index` in the index space of `dim(result, output_spatial_dimensions)`,
 `result[result_shape(:, output_spatial_index, :)] = dot_product` where:
 
-* `padded_lhs = pad(lhs, 0, lhs_padding[:, 0], lhs_padding[:, 1], lhs_base_dilations)`.
+* `padded_lhs = pad(lhs, 0, lhs_padding[:, 0], lhs_padding[:, 1], lhs_base_dilations - 1)`.
```
ghpvnist, line 1887: How about `lhs_base_dilations[:] - 1` to explicitly call out all elements are being subtracted by 1.

ghpvnist (changes requested):

docs/spec.md
sdasgup3: Thanks! It is better.

ghpvnist (approved):

docs/spec.md
burmako: Fwiw my preference lies with `lhs_base_dilations - 1` because it's shorter and looks exactly how this would idiomatically look like in NumPy. I'm fine either way for now, and we'll sort this all out in #289.

burmako (approved):

PR #1306, opened 2023-03-10
[RFC] Increase backward compatibility guarantees to 6 months

burmako:
We have recently had the first official release of StableHLO, which fully implemented the compatibility guarantees established by the compatibility RFC accepted in December.

These guarantees ended up being more compelling than we anticipated in the RFC discussions, and there are stakeholders who are interested in using them in production right away. To enable that, we have received a request to bump up the backward compatibility guarantees to 6 months (from 1 month for the 0.x.x series, as established by the original RFC).

I would like to proposes to fulfil this request to further strengthen the practical usefulness of the StableHLO opset, given that the additional maintenance cost looks acceptable (5 extra months of maintaining backward-compatible VHLO ops).

This RFC does not affect the long-term compatibility guarantees established for the 1.x.x series and onwards, which remain at 5 years of forward and backward compatibility.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -16,10 +16,10 @@ Future work section for more details.
 
 ## Guarantees
 
-**1 month of backward compatibility:** Portable artifacts serialized by an old
+**6 months of backward compatibility:** Portable artifacts serialized by an old
 version of libStablehlo have the same semantics* when deserialized by a new
 version of libStablehlo if these versions are built from openxla/stablehlo
-commits which are less than 1 month apart.
+commits which are less than 6 months apart.
 
 **1 month of forward compatibility:** Portable artifacts serialized by a new
 version of libStablehlo have the same semantics* when deserialized by an old
--- a/rfcs/20220912-compatibility.md
+++ b/rfcs/20220912-compatibility.md
@@ -5,6 +5,7 @@ Initial version: 9/12/2022<br/>
 Last updated: 3/9/2022
 
 Version log
+
 * 9/12/2022: Initial version.
 * 11/9/2022: Major updates based on a prototype implementation and conversations
              with OpenXLA and MLIR communities.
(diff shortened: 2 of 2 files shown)


stellaraccident:
The cost of an increased compatibility window will fall on your team, so if you say you are ready, I think it unlikely that others will object.

stellaraccident (approved):

bhack:
Just for curiosity, have you ever evaluated a differentiated compatibility policy for a subset of operators to modulate the effort? For example an LTS classification on a "qualified" subset.

burmako:
"Just for curiosity, have you ever evaluated a differentiated compatibility policy for a subset of operators to modulate the effort?". This hasn't come up directly, but we discussed something related - potentially creating "profiles" within the StableHLO opset, e.g. based on hardware capabilities or on the level of official support (including having tiers that ops can graduate from/to). There is no burning need for this feature at the moment, but I expect we'll need it in the near future - most likely before v1.0.

For what it's worth, we kind of already have infrastructure for establishing a DIY differentiated compatibility policy based on CustomCallOp. Semantics for this op are barely mandated by the spec, so implementations can DIY however they see fit - in fact, this is what XLA [is already doing](https://github.com/tensorflow/mlir-hlo/blob/376c9548b0a54b467f4df8ada87726d9aefbd72d/mhlo/transforms/hlo_legalize_to_stablehlo/hlo_legalize_to_stablehlo.cc#L122-L135). This design is not necessarily ideal, but it can get the job done and allows building experience with how a more principled feature could look like.

bhack:
Yes I understand that this could be at the crossroads between custom calls and supported ops.

And it is why I also find this in the tension between the more ore less long term support overhead on scale and what I call a ["custom call" escape/shortcut](https://github.com/openxla/openxla-pjrt-plugin/issues/6#issuecomment-1434802123).

Custom call escape is a sort of adversarial to RFC. An equilibrium to be found for the whole OpenXLA stack to express Tensor programs.

burmako:
[As discussed](https://github.com/openxla/community/blob/main/meetings/2023-03-21-community-meeting/AGENDA.md) at the community meeting on Tuesday 3/21, the review period for this RFC closes today. There is support for increasing backward compatibility guarantees to 6 months, and we haven't received any blocking feedback, so the RFC is now approved.

PR #1307, opened 2023-03-10
Fix formatting in the rfcs/ directory

burmako:
Looks like one of our RFCs has formatting which is not compliant with our linters. This PR fixes that.


sdasgup3 (approved):

subhankarshah (approved):
LGTM

PR #1308, opened 2023-03-10
[RFC] Add a new type to StableHLO: E4M3B11FNUZ

majnemer:
This is a proposal to add a new floating point type to StableHLO, please see rfcs/20230309-e4m3b11.md for more details.

### DIFF
--- a/rfcs/20230309-e4m3b11.md
+++ b/rfcs/20230309-e4m3b11.md
@@ -1,0 +1,50 @@
+# RFC: E4M3B11FNUZ in XLA
+
+## Summary
+
+Google has hardware which features a floating-point format similar to the one
+recommended in Sun et al.[^1]:
+
+- 1 sign bit
+- 4 exponent bits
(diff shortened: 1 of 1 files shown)


majnemer:
APFloat and related changes are posted for review [here](https://reviews.llvm.org/D146441).

sherhut:
Adding this type makes sense to me and I expect that we will see more requests of this kind as the use of different floating point formats proliferates across hardware. We need to be able to target these different floating point types from stablehlo.

At some point we will need to start thinking about optionality/portability of fp8 formats and whether we want to have some sort of hierarchy so that compilers can chose a specific implementation type for a specified type at the stablehlo level. We already have the concept of a `precision_config` but that is fairly coarse grained and not well specified. However, we should not gate landing this on the larger design question, as I expect that we'll need a way to target specific types anyway, e.g., for numeric reasons where that matters.

burmako (approved):

Issue #362, opened 2022-10-19
Add support for Bazel build/tests


ghpvnist:
Currently we are not using Bazel, and do not have support for Python tests. Need to consider adding them in the future.

burmako:
We don't have a Bazel build in this repository yet (#14), but we do have one in MLIR-HLO (see [BUILD](https://github.com/tensorflow/mlir-hlo/blob/master/stablehlo/BUILD)). One thing that this build doesn't support is tests - neither FileCheck tests nor Python tests. In case the work on a native Bazel build takes a while, we might want to fix the MLIR-HLO build in the meanwhile.

burmako:
We now have a separate ticket for running FileCheck tests in Bazel (#1315), so that this ticket can focus on its original intent to keep track of work needed for running Python tests in Bazel.

PR #1321, opened 2023-03-12
Add support for running FileCheck tests in the Bazel build

### LINKED ISSUE #1315: Add support for running FileCheck tests in the Bazel build
At the moment, we're only running tests in the CMake build, because there wasn't a good way to run FileCheck tests in Bazel when the repository was created. I heard that the situation has improved since then, so we should enable try to these tests.

burmako:
Add support for running FileCheck tests in the Bazel build

Before this PR, only CMake build could actually ran tests, and Bazel
build could only build things.

With this PR, Bazel build can now test everything except for a few
tests that involve Python bindings. This is relatively acceptable for
now, and we have #362 to keep track of specifically Python tests.

While learning about FileCheck support in Bazel and applying this
knowledge to our repositoru, I discovered that our CMake build for
FileCheck tests was vastly overcomplicated, so I simplified it
considerably.

Finally, I noticed that ci_build_llvm.sh builds mlir-cpu-runner.
This is something that we got rid of in README.md recently, and I don't
think we need this in CI either.

Fixes #1315


stablehlo/testdata/BUILD.bazel
```diff
@@ -0,0 +1,51 @@
...
+    lit_test(
+        name = "%s.test" % src,
+        srcs = [src],
+        data = [
+            "lit.cfg.py",
+            "lit.site.cfg.py",
+            "//:stablehlo-opt",
+            "//:stablehlo-translate",
+            "@llvm-project//llvm:FileCheck",
+        ] + glob(["%s.0_9_0.bc" % src]),
```
GleasonK, line 47: Any way to wildcard this like `"%s.*.bc`? Want to avoid adding a new line to this for every version of libStablehlo, until we figure out where these tests should live.

GleasonK (approved):
Our integration testing puts this all in the top-level build file, is there any reason to avoid doing this here as well?

burmako:
"Our integration testing puts this all in the top-level build file, is there any reason to avoid doing this here as well?". As it currently stands, `lit_test` in the Bazel build uses `lit.cfg.py` and `lit.site.cfg.py.in` files from the CMake build. Therefore, deduplicating `lit_test` in the Bazel build  would involve deduplicating these files between tests/ and testdata/, and I don't know how to do this in CMake (yet). Opened #1322 to track this work.

PR #1327, opened 2023-03-15
Fix bug with eval slice in refine shapes

GleasonK:
Converting to uint was asserting when passing a `SmallVector<int64_t>`.


GleasonK:
Closing in favor of #1328

Issue #1319, opened 2023-03-11
Update Version::getCurrentVersion now that we've released v0.9.0


burmako:
Now that v0.9.0 has been released, we'll want to update Version::getCurrentVersion. Since we haven't yet made any changes to VHLO, the current state is not incorrect, but it will be once we do.

We don't yet have a precedent for how to label prerelease builds. In my past life, when I was working on the Scala compiler, we used the Maven convention and appended `-SNAPSHOT` to version numbers to indicate that a build is prerelease (e.g. see [a StackOverflow answer](https://stackoverflow.com/questions/5901378/what-exactly-is-a-maven-snapshot-and-why-do-we-need-it) for background). Within that convention, our current version would be `1.0.0-SNAPSHOT`.

Additionally, given that we are providing time-based compatibility guarantees, to be compatible with HEAD-oriented development environments, we'll likely want to think how to tie this to version numbers as well. E.g. are we going to say that `0.9.0` is compatible with all kinds of `1.0.0-SNAPSHOT` builds? [SemVer](https://semver.org/) says "A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version", so we don't have to provide these guarantees but we can if we want.

One thing that we could do to improve debuggability of these snapshot builds including build metadata in their versions, e.g. `1.0.0-SNAPSHOT+3722fce`. In Scala, the build tool that we used made it very easy to generate this metadata at build time. Not sure yet how to do this consistently across CMake and Bazel (if we decide to do this at all). Maybe compute this in Bash and then pass to the build via an environment variable?

burmako:
To capture a brainstorming discussion that we just had in the context of #1306 and #1330:

1\) Framing things in terms of "earliest forward compatible version" (#1330) can be confusing. From one perspective, such a version can be called "earliest compatible" (earliest in the compatibility window centered at today). From another perspective, such a version can be called "latest compatible" (latest among the versions which can be used with compatibility guarantees).

2\) Instead, there is an alternative idea to provide users with a lower-level API that lets them directly work with dates. Here's how such a hypothetical API could look like in Python:

```Python
# E.g. if we want to put StableHLO into SavedModel, we want artifacts to comply with TF compatibility policy.
# Per https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/compat/compat.py
# this means a 3-week forward compatibility window.
latest_compatible_date = datetime.now() - timedelta(weeks=3)
[...]

PR #1330, opened 2023-03-16
Get earliest forward compatible version API

GleasonK:


stablehlo/integrations/python/mlir/dialects/stablehlo.py
```diff
@@ -17,3 +17,11 @@
...
 from ._stablehlo_ops_gen import *
 from .._mlir_libs._stablehlo import *
+
+
+def get_earliest_forward_compatible_version():
+  """Return the earliest StableHLO version that the current StableHLO version
+
+  is still forward compatible with.
+  """
+  return "0.9.0"
```
GleasonK, line 27: Any thoughts on single-sourcing this number? Our python build confuses me a little bit, we could add this as a method in StablehloModule.cpp which calls Serialization.h.. but how does building / including that file work? Would we be expecting all users of stablehlo python API's to include serialization dependencies in their build dependencies?

stablehlo/dialect/Serialization.cpp
```diff
@@ -79,5 +79,7 @@ OwningOpRef<ModuleOp> deserializePortableArtifact(StringRef sourceStr,
   return module;
 }
 
+StringRef getEarliestForwardCompatibleVersion() { return "0.9.0"; }
```
burmako, line 82: Should we also add `getEarliestBackwardCompatibleVersion()`, given that these two don't have to be the same?

stablehlo/integrations/python/mlir/dialects/stablehlo.py
```diff
@@ -17,3 +17,11 @@
 # pylint: disable=wildcard-import,relative-beyond-top-level,g-import-not-at-top
 from ._stablehlo_ops_gen import *
 from .._mlir_libs._stablehlo import *
+
+
+def get_earliest_forward_compatible_version():
+  """Return the earliest StableHLO version that the current StableHLO version
+
```
burmako, line 24: Looks like extraneous newline.

stablehlo/integrations/python/mlir/dialects/stablehlo.py
burmako, line 27: Should we also expose this API through `stablehlo-translate`?

burmako: I think it's a good idea to have a single source of truth. 

The concern about incurring additional dependency on StablehloModule.cpp is a good point, but since we aren't yet guaranteeing much about our public APIs, I think it's fine to have this dependency. We can refactor later if this turns out to be a problem.

burmako (changes requested):

stablehlo/integrations/python/mlir/dialects/stablehlo.py
```diff
@@ -17,3 +17,11 @@
 # pylint: disable=wildcard-import,relative-beyond-top-level,g-import-not-at-top
```
sdasgup3, line 17: nit:
 copyright notice?

stablehlo/integrations/python/mlir/dialects/stablehlo.py
GleasonK: removed method in favor of StablehloModule.cpp impl

stablehlo/integrations/python/mlir/dialects/stablehlo.py
GleasonK: updated all copyrights

stablehlo/integrations/python/mlir/dialects/stablehlo.py
GleasonK: Is there any reason for that? We don't expose any other language bindings or C/Python APIs via `stablehlo-translate`? If we were to expose it, I would think we should do it the same way we do `--target=current`. Something like `--target=earliest` or `minimum`

stablehlo/dialect/Serialization.cpp
GleasonK: Should we wait for a use case for that? Given that there is no way to tell what version a payload is currently, I'm not sure this method is very useful.

burmako (changes requested):
As discussed offline and summarized in https://github.com/openxla/stablehlo/issues/1319#issuecomment-1478770397, I think a better way of expressing this would be to expose a `std::optional<Version> lookupKnownVersion(Date date)` API that would make user intent straightforward.

```Python
# E.g. if we want to put StableHLO into SavedModel, we want artifacts to comply with TF compatibility policy.
# Per https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/compat/compat.py
# this means a 3-week forward compatibility window.
latest_compatible_date = datetime.now() - timedelta(weeks=3)
latest_compatible_version = stablehlo.lookup_known_version(latest_compatible_date)
stablehlo.serialize_portable_artifact(stablehlo_module, latest_compatible_version)
```

"I want to serialize my StableHLO module so that it can be deserialized by a tool from 3 weeks ago". This way we sidestep the notion of compatibility windows, forward/backward compatibility, earliest/latest, etc - the user doesn't need to know any of that. And if they do something funny, e.g. try to serialize stuff outside the compatibility window, they will get an exception that will tell them what they did wrong.

GleasonK:
Updated this to just be for Serialization APIs and current version. I still think design work is needed for the details of the `lookupKnownVersion` API.

stablehlo/dialect/Serialization.h
```diff
@@ -16,9 +16,12 @@ limitations under the License.
 #ifndef STABLEHLO_DIALECT_SERIALIZATION_H
 #define STABLEHLO_DIALECT_SERIALIZATION_H
 
+#include <cstdint>
```
burmako: Looks like this include is no longer needed.

stablehlo/dialect/Serialization.h
```diff
@@ -32,6 +35,9 @@ LogicalResult serializePortableArtifact(ModuleOp module,
 OwningOpRef<ModuleOp> deserializePortableArtifact(StringRef sourceStr,
                                                   MLIRContext* context);
 
+// Get current StableHLO version number
```
burmako, line 36: Just "version"?

stablehlo/integrations/python/StablehloModule.cpp
```diff
@@ -474,4 +476,39 @@ PYBIND11_MODULE(_stablehlo, m) {
                                        stablehloTypeExtensionsGetBoundsSize,
                                        stablehloTypeExtensionsGetBoundsElem);
       });
+
+  //
+  // Serialization APIs.
+  //
+
+  m.def("get_current_version",
+        []() { return mlir::stablehlo::getCurrentVersion(); });
```
burmako, line 485: Let's align the order of these functions between Serialization.h and StablehloModule.cpp? I don't have a preference for the order, just for it being the same across files.

burmako (changes requested):

stablehlo/dialect/Serialization.cpp
```diff
@@ -79,5 +79,9 @@ OwningOpRef<ModuleOp> deserializePortableArtifact(StringRef sourceStr,
   return module;
 }
 
+std::string getCurrentVersion() {
+  return mlir::vhlo::Version::getMinimumVersion().toString();
```
burmako, line 83: Should be getCurrentVersion I think.

burmako (approved):

PR #1335, opened 2023-03-17
Fix tests by updating STABLEHLO_TOOLS_DIR

burmako:
`set(STABLEHLO_TOOLS_DIR ${CMAKE_BINARY_DIR}/bin)` is the right way of obtaining a path to the directory that contains stablehlo-opt and stablehlo-translate.

CMAKE_CURRENT_BINARY_DIR works as well if we're building standalone, but that breaks down when building as part of another project, which we discovered when integrating into MLIR-HLO. This pull request fixes this oversight.


sdasgup3 (approved):

CMakeLists.txt
```diff
@@ -114,7 +114,7 @@ endif()
 
 set(STABLEHLO_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 set(STABLEHLO_BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR})
-set(STABLEHLO_TOOLS_DIR ${CMAKE_CURRENT_BINARY_DIR}/bin)
+set(STABLEHLO_TOOLS_DIR ${CMAKE_BINARY_DIR}/bin)
```
GleasonK, line 117: Let's make sure this works in projects including StableHLO before submit

Issue #1334, opened 2023-03-17
`SameVariadicOperandSize` does not verify operand sizes


ghpvnist:
ReduceOp, ReduceWindowOp, and other ops with `SameVariadicOperandSize` trait that have consecutive variadic operands do not verify that these variadic operands have the same shape at the MLIR level. More investigation is needed to fix this. It seems that this trait has been removed at some point and verification got lost.

Here is an example of a stablehlo program where we have 3 operands/init_values but passing the trait:
```
func.func @reduce(%arg0: tensor<4x4xf32>, %arg1 : tensor<f32>) -> (tensor<f32>) {
  %0 = "stablehlo.reduce"(%arg0, %arg1, %arg1) ({
    ^bb0(%arg2: tensor<f32>, %arg3: tensor<f32> ):
      %1 = stablehlo.add %arg2, %arg3 : tensor<f32>
      stablehlo.return %1 : tensor<f32>
    }) {
      dimensions = dense<[0, 1]> : tensor<2xi64>
    } : (tensor<4x4xf32>, tensor<f32>, tensor<f32>) -> tensor<f32>
  func.return %0: tensor<f32>
}
```

PR #1280, opened 2023-03-06
Add interpreter for ReduceOp

ghpvnist:
We have the following constraints in the spec:

```
(I1) inputs: variadic number of tensors
(I2) init_values: variadic number of 0-dimensional tensors
(I3) dimensions: 1-dimensional tensor constant of type `si64`
(I4) body: function
(C1) All `inputs` have the same shape.
(C2) element_type(`inputs[k]`) = element_type(`init_values[k]`) =
element_type(`results[k]`) for all `k` $\in$ [0, N).
(C3) size(`inputs`) = size(`init_values`) $=$ size(`results`) $=$ N where
N >= 1.
(C4) 0 <= `dimensions[d]` < rank(`inputs[0][d]`) for all dimension `d`.
(C5) All dimensions in `dimensions` are unique.
(C6) `body` has type `(tensor<E0>, ..., tensor<EN-1>, tensor<E0>, ...,`
`tensor<EN-1>) -> (tensor<E0>, ..., tensor<EN-1>)` where
`Ek = element_type(inputs[k])`.
(C7) shape(`results[k]`) = shape(`inputs[k]`) except that the dimension
sizes of `inputs[k]` corresponding to `dimensions` are not included.
```

These constraints will be comprehensively covered by the following tests:

```
I1: a) inputs is not a variadic tensor. (Covered by ODS).
I2: a) init_values is not a variadic 0-dimensional tensors.
I3: a) dimensions is not a 1-dimensional tensor.
    b) element_type(dimensions) != si64. (Covered by ODS).
I4: a) body is not a function. (Covered by ODS).
C1: a) Not all `inputs` have the same shape.
C2: a) element_type(`inputs[k]`) != element_type(`init_values[k]`) for any `k` $\in$ [0, N).
    b) element_type(`inputs[k]`) != element_type(`results[k]`) for any `k` $\in$ [0, N).
C3: a) size(`inputs`) != size(`init_values`). (Covered by ODS).
    b) size(`inputs`) != size(`results`). (Covered by ODS).
    c) size(`inputs`) < 1. (Covered by ODS).
C4: a) 0 > `dimensions[d]` for any dimension `d`.
    b) `dimensions[d]` >= rank(`inputs[0][d]`) for any dimension `d`.
C5: a) Dimensions in `dimensions` are not unique.
C6: a) `body` does not have type `(tensor<E0>, ..., tensor<EN-1>, tensor<E0>, ...,`
[...]

Issue #704, opened 2022-12-06
Decide on ranks of reducer arguments/results


burmako:
```mlir
func.func @main() -> tensor<4xi32> {
  %input = stablehlo.constant dense<1> : tensor<4x4xi32>
  %init_value = stablehlo.constant dense<0> : tensor<4xi32>
  %result = "stablehlo.reduce"(%input, %init_value) ({
    ^bb0(%arg2: tensor<4xi32>, %arg3: tensor<4xi32>):
      %1 = "stablehlo.add"(%arg2, %arg3) : (tensor<4xi32>, tensor<4xi32>) -> tensor<4xi32>
      "stablehlo.return"(%1) : (tensor<4xi32>) -> ()
  }) {
    dimensions = dense<[0]> : tensor<1xi64>
  } : (tensor<4x4xi32>, tensor<4xi32>) -> tensor<4xi32>
  func.return %result: tensor<4xi32>
}
```

This is a valid program which goes contrary to the spec that says that reducers can only have scalar arguments/results. Let's figure out why this is a valid program in MHLO and decide on the next steps.

sdasgup3:
As part of fixing this, we will be adding verification tests to ops missing tjose (e.g.  https://github.com/openxla/stablehlo/pull/1336#discussion_r1144027053)

PR #1336, opened 2023-03-18
Add interpreter for ReduceWindowOp

### LINKED ISSUE #983: Add interpreter for reduce_window

ghpvnist:

```
(I1) `inputs`: variadic number of tensors.
(I2) `init_values`: variadic number of 0-dimensional tensors.
(I3) `window_dimensions`: 1-dimensional tensor constant of type `si64`.
(I4) `window_strides`: 1-dimensional tensor constant of type `si64`.
(I5) `base_dilations`: 1-dimensional tensor constant of type `si64`.
(I6) `window_dilations`: 1-dimensional tensor constant of type `si64`.
(I7) `padding`: 2-dimensional tensor constant of type `si64`.
(I8) `body`: function.
(C1) size(`inputs`) = size(`init_values`) = size(`results`) = N and
  N >= 1.
(C2) All `inputs` have the same shape.
(C3) `element_type(inputs[k]) = element_type(init_values[k])` for all k
  in [0, N).
(C4) size(`window_dimensions`) = rank(`inputs[0]`).
(C5) `window_dimensions[i]` > 0 for all i in [0, size(`window_dimensions`)).
(C6) size(`window_strides`) = rank(`inputs[0]`).
(C7) `window_strides[i]` > 0 for all i in [0, size(`window_strides`)).
(C8) size(`base_dilations`) = rank(`inputs[0]`).
(C9) `base_dilations[i]` > 0 for all i in [0, size(`base_dilations`)).
(C10) size(`window_dilations`) = rank(`inputs[0]`).
(C11) `window_dilations[i]` > 0 for all i in [0, size(`window_dilations`)).
(C12) dim(`padding`, 0) = rank(`inputs[0]`) and dim(`padding`, 1) = 2.
(C13) `body` has type `(tensor<E0>, ..., tensor<EN-1>, tensor<E0>, ..., tensor<EN-1>) -> (tensor<E0>, ..., tensor<EN-1>)`
  where `Ek = element_type(inputs[0])`.
(C14) All `results` have the same shape.
(C15) `shape(results[0]) = num_windows`
  * `dilated_input_shape = shape(inputs[0]) == 0 ? 0 : (shape(inputs[0]) - 1) * base_dilations + 1`.
  * `padded_input_shape = padding[:, 0] + dilated_input_shape + padding[:, 1]`.
  * `dilated_window_shape = window_dimensions == 0 ? 0 : (window_dimensions - 1) * window_dilations + 1`.
  * `num_windows = (padded_input_shape == 0 || dilated_window_shape > padded_input_shape) ? 0 : floor((padded_input_shape - dilated_window_shape) / window_strides) + 1`.
(C16) `element_type(results[k]) = element_type(init_values[k])` for all k
  in [0, N).
```

```
I1: a) `inputs` is not variadic number of tensors. (Covered by ODS).
I2: a) `init_values` is not variadic number of tensors. (Covered by ODS).
    b) `init_values` is not 0-dimensional tensors.
I3: a) `window_dimensions` is not a 1-dimensional tensor constant.
    b) element_type(`window_dimensions`) != `si64`. (Covered by ODS).
I4: a) `window_strides` is not a 1-dimensional tensor constant.
    b) element_type(`window_strides`) != `si64`. (Covered by ODS).
I5: a) `base_dilations` is not a  1-dimensional tensor constant.
    b) element_type(`base_dilations`) != `si64`. (Covered by ODS).
I6: a) `window_dilations` is not a 1-dimensional tensor constant.
    b) element_type(`window_dilations`) != `si64`. (Covered by ODS).
I7: a) `padding` is not a 2-dimensional tensor constant.
    b) element_type(`padding`) != `si64`. (Covered by ODS).
I8: a) `body` is not a function. (Covered by ODS).
C1: a) size(`inputs`) != size(`init_values`)
    b) size(`inputs`) != size(`results`)
    c) size(`inputs`) < 1.
C2: a) Any `inputs` does not have the same shape.
C3: a) `element_type(inputs[k]) != element_type(init_values[k])` for any k
  in [0, N).
C4: a) size(`window_dimensions`) != rank(`inputs[0]`).
C5: a) `window_dimensions[i]` <= 0 for any i in [0, size(`window_dimensions`)).
C6: a) size(`window_strides`) != rank(`inputs[0]`).
C7: a) `window_strides[i]` <= 0 for any i in [0, size(`window_strides`)).
C8: a) size(`base_dilations`) != rank(`inputs[0]`).
C9: a) `base_dilations[i]` <= 0 for any i in [0, size(`base_dilations`)).
C10: a) size(`window_dilations`) != rank(`inputs[0]`).
C11: a) `window_dilations[i]` <= 0 for any i in [0, size(`window_dilations`)).
C12: a) dim(`padding`, 0) != rank(`inputs[0]`)
     b) dim(`padding`, 1) != 2.
C13: a) `body` does not have type `(tensor<E0>, ..., tensor<EN-1>, tensor<E0>, ..., tensor<EN-1>) -> (tensor<E0>, ..., tensor<EN-1>)`
  where `Ek = element_type(inputs[0])`.
C14: a) Any `results` do not have the same shape.
C15: a) `shape(results[0]) != num_windows`
  * `dilated_input_shape = shape(inputs[0]) == 0 ? 0 : (shape(inputs[0]) - 1) * base_dilations + 1`.
  * `padded_input_shape = padding[:, 0] + dilated_input_shape + padding[:, 1]`.
  * `dilated_window_shape = window_dimensions == 0 ? 0 : (window_dimensions - 1) * window_dilations + 1`.
  * `num_windows = (padded_input_shape == 0 || dilated_window_shape > padded_input_shape) ? 0 : floor((padded_input_shape - dilated_window_shape) / window_strides) + 1`.
C16: a) `element_type(results[k]) != element_type(init_values[k])` for any k
  in [0, N).
```

```
I2b: `init_values` is not 0-dimensional tensors.
I3a: `window_dimensions` is not a 1-dimensional tensor constant.
I4a: `window_strides` is not a 1-dimensional tensor constant.
I5a: `base_dilations` is not a  1-dimensional tensor constant.
I6a: `window_dilations` is not a 1-dimensional tensor constant.
I7a: `padding` is not a 2-dimensional tensor constant.
C1a: size(`inputs`) != size(`init_values`)
C1b: size(`inputs`) != size(`results`)
C1c: size(`inputs`) < 1.
C2a: Any `inputs` does not have the same shape.
C3a: `element_type(inputs[k]) != element_type(init_values[k])` for any k in [0, N).
C4a: size(`window_dimensions`) != rank(`inputs[0]`).
C5a: `window_dimensions[i]` <= 0 for any i in [0, size(`window_dimensions`)).
C6a: size(`window_strides`) != rank(`inputs[0]`).
C7a: `window_strides[i]` <= 0 for any i in [0, size(`window_strides`)).
C8a: size(`base_dilations`) != rank(`inputs[0]`).
C9a: `base_dilations[i]` <= 0 for any i in [0, size(`base_dilations`)).
C10a: size(`window_dilations`) != rank(`inputs[0]`).
C11a: `window_dilations[i]` <= 0 for any i in [0, size(`window_dilations`)).
C12a: dim(`padding`, 0) != rank(`inputs[0]`)
C12b: dim(`padding`, 1) != 2.
C13a: `body` does not have type `(tensor<E0>, ..., tensor<EN-1>, tensor<E0>, ..., tensor<EN-1>) -> (tensor<E0>, ..., tensor<EN-1>)`
  where `Ek = element_type(inputs[0])`.
C14a: Any `results` do not have the same shape.
C15a: `shape(results[0]) != num_windows`
  * `dilated_input_shape = shape(inputs[0]) == 0 ? 0 : (shape(inputs[0]) - 1) * base_dilations + 1`.
  * `padded_input_shape = padding[:, 0] + dilated_input_shape + padding[:, 1]`.
  * `dilated_window_shape = window_dimensions == 0 ? 0 : (window_dimensions - 1) * window_dilations + 1`.
  * `num_windows = (padded_input_shape == 0 || dilated_window_shape > padded_input_shape) ? 0 : floor((padded_input_shape - dilated_window_shape) / window_strides) + 1`.
C16a: `element_type(results[k]) != element_type(init_values[k])` for any k in [0, N).
```

Notes:
* Minor wording change in the spec.
* We cannot verify C1a: size(`inputs`) != size(`init_values`) as noted in #1334.
* Removed some duplicate verification tests.

closes #983


ghpvnist:
You may review the latest commit [Add interpreter for ReduceWindowOp](https://github.com/openxla/stablehlo/pull/1336/commits/15ffab9545b08c07a6a2928ff8df927f31447924) for easier review instead of the entire PR until ReduceOp is merged.

stablehlo/tests/verify_reduce_window.mlir
```diff
@@ -742,39 +544,127 @@ func.func @reduce_window_invalid_reducer(%arg0: tensor<4x2xf32>, %init0: tensor<
 
 // -----
 
-func.func @reduce_window_invalid_reducer(%arg0: tensor<4x2xf32>, %init0: tensor<4x2xf32>)
-  -> tensor<2x1xf32> {
-  %0 = "stablehlo.reduce_window"(%arg0, %init0) ({
-         ^bb0(%a0: tensor<4x2xf32>, %b0: tensor<4x2xf32>):
-              %2 = stablehlo.add %a0, %b0 : tensor<4x2xf32>
-              "stablehlo.return"(%2) : (tensor<4x2xf32>) -> ()
+func.func @reduce_window_i2(%arg0: tensor<4x2xf32>, %arg1: tensor<4x2xi32>,
```
sdasgup3, line 547: This seems more like c13 test. The correspondence between result and element type of input (which equal to init_value as per c3) is testing in c13

stablehlo/tests/verify_reduce_window.mlir
```diff
@@ -742,39 +544,127 @@ func.func @reduce_window_invalid_reducer(%arg0: tensor<4x2xf32>, %init0: tensor<
...
+           base_dilations = dense<[1, 1]> : tensor<2xi64>,
+           window_dilations = dense<[[1, 1]]> : tensor<1x2xi64> }
+         : (tensor<*xf32>, tensor<4x?xi32>, tensor<f32>, tensor<i32>) ->
+              (tensor<?x?xf32>, tensor<*xi32>)
+  func.return %0#0, %0#1 : tensor<?x?xf32>, tensor<*xi32>
+}
+
+// -----
+
+func.func @reduce_window_i7(%arg0: tensor<4x2xf32>, %arg1: tensor<4x2xi32>,
```
sdasgup3, line 656: `i7_c12`

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3753,6 +3755,7 @@ LogicalResult verifyReduceWindowOp(
   }
   bool allInputsUnranked = (rankedInputIdx == -1);
   Block& block = body.front();
+  // reduce_window_c3, reduce_window_c13, reduce_window_i2
```
sdasgup3, line 3758: As per https://github.com/openxla/stablehlo/pull/1280#issuecomment-1476730861
we have a missing check for i2. We can fix it is reduce op to be percolated in other ops.

stablehlo/tests/verify_reduce_window.mlir
sdasgup3: We can have a dedicated test once we have the check for i2.

stablehlo/reference/Ops.cpp
```diff
@@ -618,6 +770,58 @@ SmallVector<Tensor> eval(
...
+          scope.find(reduceOp.getInitValues());
+      SmallVector<TensorType> resultTypes;
+      for (auto resultType : reduceOp.getResultTypes())
+        resultTypes.push_back(resultType.cast<TensorType>());
+      SmallVector<Tensor> runtimeResults = evalReduceOp(
+          runtimeInputs, runtimeInitValues, Axes(reduceOp.getDimensions()),
+          reduceOp.getBody(), scope, resultTypes);
+      scope.add(op.getResults(), {runtimeResults});
+    } else if (auto reduceWindowOp = dyn_cast<ReduceWindowOp>(op)) {
+      SmallVector<Tensor> runtimeInputs =
```
sdasgup3, line 785: Added some space and keep the related things together. 
```
 SmallVector<Tensor> runtimeInputs =
          scope.find(reduceWindowOp.getInputs());
      SmallVector<Tensor> runtimeInitValues =
          scope.find(reduceWindowOp.getInitValues());

      SmallVector<TensorType> resultTypes;
      for (auto resultType : reduceWindowOp.getResultTypes())
        resultTypes.push_back(resultType.cast<TensorType>());

      SmallVector<int64_t> windowStrides =
          SmallVector<int64_t>(runtimeInputs[0].getRank(), 1);
      if (reduceWindowOp.getWindowStrides().has_value())
        windowStrides = llvm::to_vector(
            (*reduceWindowOp.getWindowStrides()).getValues<int64_t>());

      SmallVector<int64_t> baseDilations =
          SmallVector<int64_t>(runtimeInputs[0].getRank(), 1);
      if (reduceWindowOp.getBaseDilations().has_value())
        baseDilations = llvm::to_vector(
            (*reduceWindowOp.getBaseDilations()).getValues<int64_t>());

      SmallVector<int64_t> windowDilations =
          SmallVector<int64_t>(runtimeInputs[0].getRank(), 1);
     if (reduceWindowOp.getWindowDilations().has_value())
        windowDilations = llvm::to_vector(
            (*reduceWindowOp.getWindowDilations()).getValues<int64_t>());

      SmallVector<SmallVector<int64_t>> padding =
          SmallVector<SmallVector<int64_t>>(runtimeInputs[0].getRank(), {0, 0});
      if (reduceWindowOp.getPadding().has_value()) {
        auto paddingOrErr =
            hlo::convertPaddingAttribute(reduceWindowOp.getPadding(), {});
        if (failed(paddingOrErr))
          report_fatal_error(invalidArgument("Invalid padding format found."));
        for (auto i = 0; i < runtimeInputs[0].getRank(); ++i) {
          padding[i][0] = (*paddingOrErr)[i].first;
          padding[i][1] = (*paddingOrErr)[i].second;
        }
      }
```

stablehlo/reference/Ops.cpp
```diff
@@ -314,6 +325,147 @@ Tensor evalRealOp(const Tensor &operand, TensorType resultType) {
...
+}
+
+SmallVector<Tensor> evalReduceWindowOp(
+    ArrayRef<Tensor> inputs, ArrayRef<Tensor> initValues,
+    const Axes &windowDimensions, const Axes &windowStrides,
+    const Axes &baseDilations, const Axes &windowDilations,
+    const SmallVector<SmallVector<int64_t>> &padding, Region &body,
+    Scope &scope, ArrayRef<TensorType> resultTypes) {
+  SmallVector<Tensor> runtimeResults;
+  Builder builder = mlir::Builder(resultTypes[0].getContext());
```
sdasgup3, line 371: IMO we can create a shared function for getting specific mlir types given how often they are used elsewhere.

Something like

```
static Type getI64Type(MLIRContext * context) {
...
} 
```

stablehlo/reference/Ops.cpp
```diff
@@ -314,6 +325,147 @@ Tensor evalRealOp(const Tensor &operand, TensorType resultType) {
...
+  Type i64Type = builder.getI64Type();
+
+  for (auto [resultType, initValue] : llvm::zip(resultTypes, initValues)) {
```
sdasgup3, line 374: It seems that initializing a Tensor with some value is a frequent operation as well. How about we add a constructor:
```
explicit Tensor(TensorType type, const Element &element);
```

stablehlo/reference/Ops.cpp
```diff
@@ -314,6 +325,147 @@ Tensor evalRealOp(const Tensor &operand, TensorType resultType) {
...
+  }
+
+  SmallVector<int64_t> paddingLow;
+  SmallVector<int64_t> paddingHigh;
+  for (auto paddingPair : padding) {
+    paddingLow.push_back(paddingPair[0]);
+    paddingHigh.push_back(paddingPair[1]);
+  }
+
+  SmallVector<Tensor> paddedInputs;
```
sdasgup3, line 389: Can we use the same helper function we used in ConvolutionOp PR 
```
TensorType inferPadOpType(ArrayRef<std::pair<int64_t, int64_t>> lhsPadding,
                          Type operandType, Type resultElementType,
                          ArrayRef<int64_t> interiorPadding);
```


to encapsulate all the intricacies about inferring padOp return type.

stablehlo/reference/Ops.cpp
```diff
@@ -314,6 +325,147 @@ Tensor evalRealOp(const Tensor &operand, TensorType resultType) {
...
+    SmallVector<Tensor> windows;
+    auto windowStart = (*resultIt) * Sizes(windowStrides);
+    for (auto paddedInput : paddedInputs) {
+      SmallVector<int64_t> limitIndices;
+      for (size_t i = 0; i < windowStart.size(); ++i)
+        limitIndices.push_back(
+            std::min(windowStart[i] + windowDimensions[i] * baseDilations[i],
+                     paddedInput.getShape()[i]));
+      SmallVector<Type> inferredSliceType;
+      auto sliceStatus = hlo::inferSliceOp(
```
sdasgup3, line 421: Same with inferring sliceOp type
```
TensorType inferSliceOpType(Type operandType,
                            SmallVector<int64_t> lhsWindowStart,
                            SmallVector<int64_t> limitIndices,
                            SmallVector<int64_t> lhsWindowDilations) 
```

stablehlo/reference/Ops.cpp
```diff
@@ -314,6 +325,147 @@ Tensor evalRealOp(const Tensor &operand, TensorType resultType) {
...
+    SmallVector<Type> inputTypes;
+    for (auto input : inputs) inputTypes.push_back(input.getType());
+    SmallVector<Type> initValueTypes;
+    for (auto initValue : initValues)
+      initValueTypes.push_back(initValue.getType());
+    SmallVector<int64_t> dimensions(inputs[0].getRank());
+    std::iota(dimensions.begin(), dimensions.end(), 0);
+
+    SmallVector<ShapedTypeComponents> inferredReduceTypes;
+    auto reduceStatus = hlo::inferReduceOp(
```
sdasgup3, line 448: Same with inferring reduce op.

stablehlo/tests/verify_reduce_window.mlir
ghpvnist: Nice catch!

stablehlo/reference/Ops.cpp
```diff
@@ -618,6 +770,58 @@ SmallVector<Tensor> eval(
...
+      if (reduceWindowOp.getWindowStrides().has_value())
+        windowStrides = llvm::to_vector(
+            (*reduceWindowOp.getWindowStrides()).getValues<int64_t>());
+      if (reduceWindowOp.getBaseDilations().has_value())
+        baseDilations = llvm::to_vector(
+            (*reduceWindowOp.getBaseDilations()).getValues<int64_t>());
+      if (reduceWindowOp.getWindowDilations().has_value())
+        windowDilations = llvm::to_vector(
+            (*reduceWindowOp.getWindowDilations()).getValues<int64_t>());
+      if (reduceWindowOp.getPadding().has_value()) {
```
sdasgup3, line 809: padding can use the same type as it is returned from `hlo::convertPaddingAttribute` so that we do not need any conversion code:

```
  SmallVector<std::pair<int64_t, int64_t>> padding(
          runtimeInputs[0].getRank(), std::pair<int64_t, int64_t>(0, 0));
      if (reduceWindowOp.getPadding().has_value()) {
        auto paddingOrErr =
            hlo::convertPaddingAttribute(reduceWindowOp.getPadding(), {});
        if (failed(paddingOrErr))
          report_fatal_error(invalidArgument("Invalid padding format found."));
        padding = *paddingOrErr;
      }

```

stablehlo/reference/Ops.cpp
```diff
@@ -618,6 +770,58 @@ SmallVector<Tensor> eval(
...
+      scope.add(op.getResults(), {runtimeResults});
+    } else if (auto reduceWindowOp = dyn_cast<ReduceWindowOp>(op)) {
+      SmallVector<Tensor> runtimeInputs =
+          scope.find(reduceWindowOp.getInputs());
+      SmallVector<Tensor> runtimeInitValues =
+          scope.find(reduceWindowOp.getInitValues());
+      SmallVector<TensorType> resultTypes;
+      for (auto resultType : reduceWindowOp.getResultTypes())
+        resultTypes.push_back(resultType.cast<TensorType>());
+      SmallVector<int64_t> windowStrides =
```
sdasgup3, line 792: Use the constructor of `SmallVector` instead of assignment.

```
    SmallVector<int64_t> windowStrides(runtimeInputs[0].getRank(), 1);
      if (reduceWindowOp.getWindowStrides().has_value())
        windowStrides = llvm::to_vector(
            (*reduceWindowOp.getWindowStrides()).getValues<int64_t>());
```

sdasgup3 (commented):
A few more comments before I conclude my first review iteration.

stablehlo/dialect/TypeInference.cpp
ghpvnist: Let's leave this work to #704 as this requires some non-trivial changes.

stablehlo/tests/verify_reduce_window.mlir
ghpvnist: > We can have a dedicated test once we have the check for i2.
See below comment on #704

stablehlo/reference/Ops.cpp
```diff
@@ -314,6 +325,147 @@ Tensor evalRealOp(const Tensor &operand, TensorType resultType) {
...
+
+  for (auto resultIt = runtimeResults[0].index_begin();
+       resultIt != runtimeResults[0].index_end(); ++resultIt) {
+    SmallVector<Tensor> windows;
+    auto windowStart = (*resultIt) * Sizes(windowStrides);
+    for (auto paddedInput : paddedInputs) {
+      SmallVector<int64_t> limitIndices;
+      for (size_t i = 0; i < windowStart.size(); ++i)
+        limitIndices.push_back(
+            std::min(windowStart[i] + windowDimensions[i] * baseDilations[i],
```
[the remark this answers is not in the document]
ghpvnist:
```
(C5) dim(result, d) =  (limit_indices[d]-start_indices[d])/stride[d] for all dimension d in operand.
 ```
I think we do need it. The above is the constraint in SliceOp, and if limit_indices do not account for strides, the shape inference would be wrong. Here is an example slice op:
```
func.func @slice_op() {
  %operand = stablehlo.constant dense<[[1, 1, 1],
                                       [1, 1, 1],
                                       [1, 1, 1]]> : tensor<3x6xi64>
  %result = "stablehlo.slice"(%operand) {
    start_indices = dense<[0, 0]> : tensor<2xi64>,
    limit_indices = dense<[3, 3]> : tensor<2xi64>,
    strides = dense<[10, 10]> : tensor<2xi64>
  } : (tensor<3x6xi64>) -> tensor<1x1xi64>
  check.expect_eq_const %result, dense<[[1]]> : tensor<1x1xi64>
  func.return
}
```

File: stablehlo/reference/Tensor.h [hunk 13]
```diff
@@ -72,6 +72,9 @@ class Tensor {
   Tensor();
   explicit Tensor(TensorType type);
   explicit Tensor(TensorType type, AsmResourceBlob blob);
+  /// This constructor initializes the tensor populated with provided initial
+  /// values. This constructor is O(n) with respect to the tensor size.
```
sdasgup3, line 76: value singular?

stablehlo/reference/Ops.cpp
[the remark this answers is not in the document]
sdasgup3: I had the same concern you said, but here is the thing. 

While we are interpreting a StableHLO program, we are assuming that the program is valid. The purpose of calling inferSlice from a `evalOp` is not to check the validity of the result type but to just get the result type so that we can invoke `evalSliceOp`  and while doing that we can take liberty on the assumption that the program is valid.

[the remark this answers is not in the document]
ghpvnist: I see what you mean, thanks for the clarification!

sdasgup3 (approved):

stablehlo/reference/Ops.cpp
```diff
@@ -665,6 +796,58 @@ SmallVector<Tensor> evalReduceOp(ArrayRef<Tensor> inputs,
...
+       resultIt != runtimeResults[0].index_end(); ++resultIt) {
+    SmallVector<Tensor> windows;
+    auto windowStart = (*resultIt) * windowStrides;
+    for (auto paddedInput : paddedInputs) {
+      auto limitIndices = llvm::to_vector(windowStart + windowDimensions);
+      auto inferredSliceType =
+          inferSliceOpType(paddedInput.getType(), llvm::to_vector(windowStart),
+                           limitIndices, llvm::to_vector(windowDilations));
+      auto window = evalSliceOp(paddedInput, windowStart, windowDilations,
+                                inferredSliceType);
```
burmako, line 833: We should change the type of evalSliceOp's second parameter from `Index` to `Sizes`.

stablehlo/reference/Ops.cpp
```diff
@@ -665,6 +796,58 @@ SmallVector<Tensor> evalReduceOp(ArrayRef<Tensor> inputs,
...
+    SmallVector<TensorType> inferredReduceTypes =
+        inferReduceOpType(inputs, initValues, dimensions);
+    auto reducedValues = evalReduceOp(windows, initValues, Axes(dimensions),
+                                      body, scope, inferredReduceTypes);
+
+    for (auto [result, value] : llvm::zip(runtimeResults, reducedValues))
+      result.set(*resultIt, value.get({}));
+  }
+  return runtimeResults;
+}
```
burmako, line 849: Thank you! It must've been a lot of work to figure this out, but it is satisfying that the implementation ended up being quite close to the spec. It is an open question whether this implementation will be sufficiently fast though, but time will tell.

Most of my feedback for this pull request involves simplifying the notation of `evalReduceWindowOp`, so that it more clearly maps on the spec. Here's what I ended up with (see https://github.com/ghpvnist/stablehlo/pull/3 for the implementation):

```C++
SmallVector<Tensor> evalReduceWindowOp(
    ArrayRef<Tensor> inputs, ArrayRef<Tensor> initValues,
    const Sizes &windowDimensions, const Sizes &windowStrides,
    const Sizes &baseDilations, const Sizes &windowDilations,
    const Sizes &paddingLow, const Sizes &paddingHigh, Region &body,
    Scope &scope, ArrayRef<TensorType> resultTypes) {
  SmallVector<Tensor> results;
  for (auto [resultType, initValue] : llvm::zip(resultTypes, initValues)) {
    results.push_back(Tensor(resultType, initValue.get({})));
  }

  SmallVector<Tensor> paddedInputs;
  for (auto [input, initValue] : llvm::zip(inputs, initValues)) {
    paddedInputs.push_back(evalPadOp(input, initValue, paddingLow, paddingHigh,
                                     baseDilations - 1));
  }

  for (auto resultIt = results[0].index_begin();
       resultIt != results[0].index_end(); ++resultIt) {
    SmallVector<Tensor> windows;
    auto windowStart = (*resultIt) * windowStrides;
    for (auto paddedInput : paddedInputs) {
      windows.push_back(evalSliceOp(paddedInput, windowStart,
                                    windowStart + windowDimensions,
                                    windowDilations));
    }

    Axes dimensions(inputs[0].getRank());
    std::iota(dimensions.begin(), dimensions.end(), 0);
    auto reducedValues =
        evalReduceOp(windows, initValues, dimensions, body, scope);
    for (auto [result, value] : llvm::zip(results, reducedValues))
      result.set(*resultIt, value.get({}));
  }
  return results;
}
```

Integrating inferFooOpType functions into evalFooOp overloads removes clutter and provides a cleaner look for evalReduceWindowOp, but that will have performance implications since each invocation of these overloads will perform type inference over and over. I would propose to move forward with a simpler implementation and then optimize later if it becomes a problem.

burmako (changes requested):

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3751,6 +3756,7 @@ LogicalResult verifyReduceWindowOp(
...
```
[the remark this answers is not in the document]
ghpvnist: It is checked in combination to C13 as it is not possible to test them individually.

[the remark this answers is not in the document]
ghpvnist: I've added `reduce_window_i2` in the corresponding place now.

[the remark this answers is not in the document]
ghpvnist: Looking back at the checklist, these comments should be present in both places. Added them in `verifyReducerShape`.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3730,8 +3734,10 @@ LogicalResult verifyReduceWindowOp(
...
       [](Type t) -> TensorType { return t.cast<TensorType>(); })};
   uint64_t numInputs = inputs.size();
+  // reduce_window_c1
+  if (inputs.size() < 1)
+    return emitOptionalError(location, "requires at least 1 input value");
 
-  // P1. ~ P3.
   SmallVector<int64_t> windowDims;
   SmallVector<WindowDimension> inferredWindow;
   if (failed(verifyReduceWindowOpInputsAndInferWindow(
```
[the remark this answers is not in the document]
ghpvnist: Yes that was the intention, but looking back at the checklist, we should. It's added now.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -329,24 +331,28 @@ verifyWindowAttributesAndInferWindowDimensions(
     WindowDimension& dim = window[i];
 
     dim.size = windowDimensions[i];
+    // reduce_window_c5
     if (!isDynamicDimSize(dim.size) && dim.size <= 0)
```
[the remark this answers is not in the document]
ghpvnist: You're right, this should be a constant so it should never be dynamic.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -679,27 +688,29 @@ LogicalResult verifyReduceWindowOpInputsAndInferWindow(
...
   if (failed(windowDilationsOrErr)) return failure();
-  auto windowReversalOrErr = convertWindowReversalAttribute(
-      windowReversal, location, "window_reversal");
-  if (failed(windowReversalOrErr)) return failure();
+  // reduce_window_c12, reduce_window_i7
+  auto paddingOrErr = convertPaddingAttribute(padding, location);
+  if (failed(paddingOrErr)) return failure();
 
+  // reduce_window_c5, reduce_window_c6, reduce_window_c7, reduce_window_c8,
+  // reduce_window_c9, reduce_window_c10, reduce_window_c11, reduce_window_c12
```
burmako, line 708: Can we perhaps simplify this to something along the lines of "reduce_window_c5 ... reduce_window_c12"?

stablehlo/dialect/TypeInference.cpp
```diff
@@ -2567,12 +2578,16 @@ LogicalResult inferReduceWindowOp(
 
   SmallVector<int64_t> windowDims;
   SmallVector<WindowDimension> inferredWindow;
+  // reduce_window_c1, reduce_window_c2, reduce_window_c4, reduce_window_c5,
+  // reduce_window_c7, reduce_window_c9, reduce_window_c11, reduce_window_c12,
+  // reduce_window_i4, reduce_window_i5, reduce_window_i6, reduce_window_i7
```
burmako, line 2583: Is this a comprehensive list? I think the full list is c1..c13 and i3..i7.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -666,10 +674,11 @@ LogicalResult verifyReduceWindowOpInputsAndInferWindow(
             " is not compatible with shape at input-index ", rankedInputIdx);
   }
 
-  // P2.
+  // reduce_window_i3
   auto windowDimsOrErr =
       convert1DAttribute(windowDimensions, location, "window_dimensions");
   if (failed(windowDimsOrErr)) return failure();
+  // reduce_window_c4
```
burmako, line 681: What do you think about moving the reduce_window_c4 code block after all reduce_window_ix code blocks, so that we do i3..i7, then c4, and then c5..c12 (via verifyWindowAttributesAndInferWindowDimensions)?

stablehlo/reference/Ops.cpp
```diff
@@ -33,13 +34,71 @@ namespace mlir {
...
+      /*location=*/{}, inputTypes, initValueTypes,
+      builder.getI64TensorAttr(dimensions), inferredReduceTypes);
+  if (failed(reduceStatus))
+    report_fatal_error(
+        invalidArgument("Could not infer ReduceOp's return type"));
+
+  SmallVector<ShapedType> resultTypes;
+  for (auto inferredType : inferredReduceTypes)
+    resultTypes.push_back(RankedTensorType::get(inferredType.getDims(),
+                                                inferredType.getElementType()));
```
burmako, line 83: Now that I'm looking at this again, I'm realizing that this logic is incomplete, e.g. it doesn't account for encoding, and it doesn't account for missing pieces.

How about we add a function to Base.h, next to unflattenTupleTypes, with the following signature `ShapedType createShapedType(ShapedTypeComponents components)` and the following behavior:
  * If elementType is null, then return null.
  * If hasRank, then created a RankedTensorType from dims, elementType and attribute.
  * If !hasRank, then created an UnrankedTensorType from elementType.

Then here we would call this function, check if it returned null (and report_fatal_error if that's the case) and if not, push back its result.

stablehlo/dialect/TypeInference.cpp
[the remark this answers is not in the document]
burmako: In that case, let's have a ReduceWindow-specific check that makes sure that its windowDimensions cannot be dynamic?

burmako (changes requested):
Thank you for addressing all these comments quickly! I have just a few more.

stablehlo/dialect/TypeInference.cpp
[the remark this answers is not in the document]
ghpvnist: Or not? since attributes can't have dynamic sizes to begin with. Otherwise, we would need to add this check for all ops with attributes, and there isn't a way to encode `?` in an attribute.

stablehlo/dialect/TypeInference.cpp
ghpvnist: c3 is not covered in this function but it covers the remaining c1..c13. Updated the comments accordingly.

stablehlo/dialect/TypeInference.cpp
ghpvnist: The purpose of explicitly stating them was to make the constraints easier to find, but condensing them in the caller and expanding them in the callee doesn't lose this information.

stablehlo/reference/Ops.cpp
ghpvnist: Ended up having a slightly different signature. PTAL.

stablehlo/dialect/TypeInference.cpp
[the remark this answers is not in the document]
burmako: "there isn't a way to encode ? in an attribute to test it". It is possible to use `-9223372036854775808` as a value of the window_dimensions attribute. For example, this passes verification right now:

```mlir
func.func @main(%arg0: tensor<4x6xi16>, %arg1: tensor<i16>) -> tensor<3x5xi16> {
  %0 = "stablehlo.reduce_window"(%arg0, %arg1) ({
  ^bb0(%arg2: tensor<i16>, %arg3: tensor<i16>):
    %5 = stablehlo.add %arg2, %arg3 : tensor<i16>
    stablehlo.return %5 : tensor<i16>
  }) {
    window_dimensions = dense<-9223372036854775808> : tensor<2xi64>
  } : (tensor<4x6xi16>, tensor<i16>) -> tensor<3x5xi16>
  func.return %0 : tensor<3x5xi16>
}
```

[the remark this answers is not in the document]
ghpvnist: That's what I wanted to avoid. IIRC we were using -1 as `kDynamic` previously, so I wanted to avoid using that number.   But upon further look, it looks like we do have instances of using `-9223372036854775808` elsewhere. I guess in the unlikely chance this value changes, it will be handled in the future.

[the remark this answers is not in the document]
ghpvnist: Since we don't have this check for all ops just yet, I've added #1472 to address this as part of spec compliance. I'll add this check for this op for now.

stablehlo/dialect/TypeInference.cpp
ghpvnist: Added #1474 to update the checklist.

burmako: I don't think this specific comment has been updated.

stablehlo/dialect/TypeInference.cpp
burmako: Yeah, you're right c3 and c13 are not covered - they fall under verifyReducerShape which is not called by this function.

stablehlo/dialect/Base.h
```diff
@@ -187,6 +187,9 @@ void flattenTupleTypes(TypeRange types, SmallVector<Type> &result);
 LogicalResult unflattenTupleTypes(TypeRange prototype, TypeRange types,
                                   SmallVector<Type> &result);
 
+LogicalResult createShapedType(ShapedTypeComponents components,
+                               ShapedType &type);
```
burmako, line 191: Why not return ShapedType directly? It is idiomatic to use empty values for stuff like this - e.g. dyn_cast<T> returns T, and not LogicalResult. The alternative involves more code and doesn't provide any benefits.

Now, if let's say our function returned a SmallVector and could fail, then we wouldn't be able to get away without LogicalResult / FailureOr / std::optional. That's because an empty SmallVector is a valid return value on its own, so it cannot be used to signal failure.

stablehlo/reference/Ops.cpp
```diff
@@ -33,13 +34,74 @@ namespace mlir {
...
+      builder.getI64TensorAttr(dimensions), inferredReduceTypes);
+  if (failed(reduceStatus))
+    report_fatal_error(
+        invalidArgument("Could not infer ReduceOp's return type"));
+
+  SmallVector<ShapedType> resultTypes;
+  for (auto inferredType : inferredReduceTypes) {
+    ShapedType shapedType;
+    if (failed(hlo::createShapedType(inferredType, shapedType)))
+      llvm::report_fatal_error("Could not retrieve type.");
```
burmako, line 84: This error message seems a bit too cryptic. Perhaps something along the lines of "Could not infer ReduceOp's return type"?

stablehlo/dialect/TypeInference.cpp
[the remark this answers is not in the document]
burmako: If the fix is intended to focus on just ReduceWindowOp, then I think it's premature to change `convert1DAttribute` and `convertPaddingAttribute` because they are used in other ops as well. How about we revert the changes to these functions in this PR and leave everything to #1472?

burmako (changes requested):

stablehlo/dialect/TypeInference.cpp
ghpvnist: It should be updated now.

burmako (approved):

Issue #755, opened 2022-12-13
Spec the behavior of precision_config in DotGeneralOp


burmako:
See https://github.com/openxla/stablehlo/pull/307#discussion_r1045135311.

ghpvnist:
Here are the results of my investigation:
On the [JAX](https://github.com/google/jax/blob/main/jax/_src/lax/lax.py#L657-L670) side, this is what the enum values mean:
* DEFAULT: uses `bf16` for computation with accumulation done in at least the precision of the output type, for which `preferred_element_type` (#600) can override, in which case depending on whether the precision is larger or smaller, the accumulation can have more/less precision, respectively.
* HIGH: uses 3 `bf16` passes or `tf32` (8 exponent, 10 mantissa bits) type if available for computation. The accumulation is done in a similar fashion.
* HIGHEST: uses `f32` or `f64` as applicable for computation. The accumulation is done in a similar fashion.

For StableHLO, we currently do not have `preferred_element_type` as an (optional) attribute and rely on the result element type. This means that we always have a defined `preferred_element_type`, so accumulation would always happen using the result element type. Also, we do not have `tf32` or model passes to calculate multiplication by splitting the bits (as this is more hardware specific). Therefore, I propose we use the following to do computation and accumulation based on the `precison_config` values:
* DEFAULT: use `bf16` for computation, accumulating values using the result element type as `preferred_element_type`.
* HIGH: use `f32` for computation, accumulating values using the result element type as `preferred_element_type`.
* HIGHEST: use `f64` for computation, accumulating values using the result element type as `preferred_element_type`.

In other words, the multiplication/sum (intermediate results) is calculated using `bf16`, `f32`, or `f64`, and the resulting sum (final value) is casted to the result element type.

ghpvnist:
(C1) lhs and rhs have the same element type.

Currently, (C1) does not allow lhs and rhs element types to be different. This would prevent specifying distinct preferred element types for lhs and rhs which means that having two config values does not provide extra information since this behavior can be captured with just one config value. This constraint should also be removed while this issue is addressed.

zichuan-wei:
Given that the inputs and outputs of an op already contain information about the precision of the expected input and output, I think it makes not so much sense to have a separate precision config, especially for on-device use cases. And DEFAULT, HIGH, & HIGHEST is kind of vague in terms of what they represent.

sdasgup3:
Thanks @zichuan-wei for the input! 
[...]

PR #1342, opened 2023-03-21
[RFC] Add Float8E4M3FNUZ and Float8E5M2FNUZ to StableHLO.

jakeh-gc:
This is a proposal to add Float8E4M3FNUZ and Float8E5M2FNUZ floating point types to StableHLO.
Feedback welcome, see `rfcs/20230321-fp8_fnuz.md` for more details.

### DIFF
--- a/rfcs/20230321-fp8_fnuz.md
+++ b/rfcs/20230321-fp8_fnuz.md
@@ -18,9 +18,9 @@ infinities), `N` for with special NaN encoding, `UZ` for unsigned zero. I
 propose keeping this naming scheme in StableHLO, matching LLVM/MLIR.
 
 These changes mean there's an additional exponent value available. To keep
-the same dynamic range as an IEEE-like FP8 type, we bias the exponent one more
-than would be expected given the number of exponent bits (8 for Float8E4M3FNUZ
-and 16 for Float8E5M2FNUZ).
+the same dynamic range as an IEEE-like FP8 type, the exponent is biased one
+more than would be expected given the number of exponent bits (8 for
+Float8E4M3FNUZ and 16 for Float8E5M2FNUZ).
 
 ### Float8E4M3FNUZ
 
--- a/rfcs/20230321-fp8_fnuz.md
+++ b/rfcs/20230321-fp8_fnuz.md
@@ -1,81 +1,115 @@
 # RFC: Float8E4M3FNUZ and Float8E5M2FNUZ
 
 ## Summary
-Graphcore, AMD, and Qualcomm have proposed two new FP8 types, Float8E4M3FNUZ and Float8E5M2FNUZ[^1].
-These types are implemented in commercially available hardware[^2], and added to MLIR builtin types[^4] and LLVM APFloat[^5].
 
-These two types appear similar to the existing types Float8E4M3FN and Float8E5M2[^3], but differ in important ways.
+Graphcore, AMD, and Qualcomm have proposed two new FP8 types, Float8E4M3FNUZ
+and Float8E5M2FNUZ[^1]. These types are implemented in commercially available
+hardware[^2], and added to MLIR builtin types[^4] and LLVM APFloat[^5].
+
+These two types appear similar to the existing types Float8E4M3FN and
+Float8E5M2[^3], but differ in important ways.
 
 ## Details
-Both Float8E4M3FNUZ and Float8E5M2FNUZ differ from typical floating point types in their support for NaN, infinities, negative zero, and exponent bias.
-The suffix "FNUZ" is derived from these differences. `F` is for "finite" (no infinities), `N` for with special NaN encoding, `UZ` for unsigned zero.
+
+Both Float8E4M3FNUZ and Float8E5M2FNUZ differ from typical floating point types
+in their support for NaN, infinities, negative zero, and exponent bias. The
+suffix "FNUZ" is derived from these differences. `F` is for "finite" (no
+infinities), `N` for with special NaN encoding, `UZ` for unsigned zero.
 I propose keeping this naming scheme in StableHLO.
 
 ### Float8E4M3FNUZ
+
 8-bit floating point with 3 bit mantissa.
 
-An 8-bit floating point type with 1 sign bit, 4 bits exponent and 3 bits mantissa. This is not a standard type as defined by IEEE-754, but it follows similar conventions, with the exception that there are no infinity values, no negative zero, and only one NaN representation. This type has the following characteristics:
+An 8-bit floating point type with 1 sign bit, 4 bits exponent and 3 bits
+mantissa. This is not a standard type as defined by IEEE-754, but it follows
+similar conventions, with the exception that there are no infinity values, no
+negative zero, and only one NaN representation. This type has the following
+characteristics:
+
 * bit encoding: S1E4M3 - `0bSEEEEMMM`
 * exponent bias: 8
 * infinities: Not supported
-* NaNs: Supported with sign bit set to 1, exponent bits and mantissa bits set to all 0s
+* NaNs: Supported with sign bit set to 1, exponent bits and mantissa bits set
+to all 0s
 * denormals when exponent is 0
 
 ### Comparison of Float8E4M3FN and Float8E4M3FNUZ
-|                   |Float8E4M3FN                                              |Float8E4M3FNUZ                                             |
-|-------------------|----------------------------------------------------------|-----------------------------------------------------------|
-|Bias               |7                                                         |8                                                          |
-|Min Normal Value   |`0bS0001000` = -1<sup>S</sup> * 1.0 * 2<sup>-6</sup>      |`0bS0001000` = -1<sup>S</sup> * 1.0 * 2<sup>-7</sup>       |
-|Max Normal Value   |`0bS1111110` = -1<sup>S</sup> * 1.75 * 2<sup>8</sup> = 448|`0bS1111111` = -1<sup>S</sup> * 1.875 * 2<sup>7</sup> = 240|
-|Min Subnormal Value|`0bS0000001` = -1<sup>S</sup> * 0.125 * 2<sup>-6</sup>    |`0bS0000001` = -1<sup>S</sup> * 0.125 * 2<sup>-7</sup>     |
-|Max Subnormal Value|`0bS0000111` = -1<sup>S</sup> * 0.875 * 2<sup>-6</sup>    |`0bS0000111` = -1<sup>S</sup> * 0.875 * 2<sup>-7</sup>     |
-|NaN                |`0bS1111111`                                              |`0b10000000`                                               |
-|Infinity           |N/A                                                       |N/A                                                        |
-|-0                 |`0b10000000`                                              |N/A                                                        |
+
+|                   |Float8E4M3FN                                                            |Float8E4M3FNUZ                                                           |
+|-------------------|------------------------------------------------------------------------|-------------------------------------------------------------------------|
+|Bias               |7                                                                       |8                                                                        |
+|Min Normal Value   |`0bS0001000` = -1<sup>S</sup> $\times$ 1.0 $\times$ 2<sup>-6</sup>      |`0bS0001000` = -1<sup>S</sup> $\times$ 1.0 $\times$ 2<sup>-7</sup>       |
+|Max Normal Value   |`0bS1111110` = -1<sup>S</sup> $\times$ 1.75 $\times$ 2<sup>8</sup> = 448|`0bS1111111` = -1<sup>S</sup> $\times$ 1.875 $\times$ 2<sup>7</sup> = 240|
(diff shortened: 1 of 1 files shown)


burmako:
@jakeh-gc Thank you for your RFC! Given its similarity to [the F8E4M3FN and F8E5M2 RFC](https://github.com/openxla/stablehlo/blob/main/rfcs/20221031-fp8.md) as well as [the F8E4M3B11FNUZ RFC](https://github.com/openxla/stablehlo/blob/main/rfcs/20230309-e4m3b11.md), I think that we have a strong case for the dtypes proposed in this RFC as well.

Similarly, to the latter RFC which we have just approved, let's keep this one open for two weeks and then approve as well if no blocking feedback from the community.

sherhut:
As stated in https://github.com/openxla/stablehlo/pull/1308#issuecomment-1477641665, adding fp8 formats supported by hardware makes sense to me. The question about longer term remains. @burmako how would we organize work to explore address this? @jakeh-gc have you thought about `precision_config`?

Just to be clear: I am not asking for this longer-term design questions to block addition of the type itself.

rfcs/20230321-fp8_fnuz.md
```diff
@@ -0,0 +1,165 @@
...
+
+These two types appear similar to the existing types Float8E4M3FN and
+Float8E5M2[^3], but differ in important ways.
+
+## Details
+
+Both Float8E4M3FNUZ and Float8E5M2FNUZ differ from typical floating point types
+in their support for NaN, infinities, negative zero, and exponent bias. The
+suffix "FNUZ" is derived from these differences. `F` is for "finite" (no
+infinities), `N` for with special NaN encoding, `UZ` for unsigned zero.
```
cantonios, line 17: Can you be a bit more specific here about the bias?  i.e. that it is one higher than typical IEEE-754?  Is this part of the `UZ` naming convention - as in all types labelled with `UZ` would have this assumption?  Otherwise I would have expected the non-standard bias to also be reflected in the name.

jakeh-gc: Thanks for taking a look. I'll clear it up in the RFC, but also explain here.

The change to the bias is a result of having an additional exponent encoding freed up by not having `inf` and only one `NaN` encoding. We decide to use this "extra exponent" at the bottom of the range, so we bias by one extra value.

Because of the "extra exponent" we have the same max value of a IEEE-754-like FP8 type. This can be seen in the comparison of `Float8E5M2` and `Float8E5M2FNUZ`. They both have maximum value 57344, but `Float8E5M2FNUZ` can go down to 0.25 $\times$ 2<sup>-15</sup> = 2<sup>-17</sup>.

The use of the additional encoding space is a free design choice.

jakeh-gc: Perhaps another way to say it is these proposed types have the same absolute maximum normal number values as a hypothetical IEEE FP8 type with the same number of exponent bits, with extra precision near zero. Given that users interact with floating point numbers as decimal values and not their binary encodings, that seems more intuitively aligned with the IEEE types.

cantonios: Yes, I understand the reasoning - and maybe this discussion was already had for the MLIR/LLVM types - if it was a free choice, I would have expected it to be reflected in the name - just as you describe for the other differences F,N,UZ.

My question was more about the choice of name.  I do think we should keep the current naming consistent with MLIR/LLVM, so I'm not proposing to change it now. 

My request for the RFC though it just to explicitly call out the exact difference in bias here (i.e. increased by one), and maybe add a short comment similar to above, like
> ... in order to keep a similar dynamic range

jakeh-gc:
@sherhut
> @jakeh-gc have you thought about `precision_config`?

I'm not sure how `precision_config` should be dealt with. I'm not personally a fan of how it's defined at the moment:
```markdown
* `DEFAULT`: Fastest calculation, but least accurate approximation to the
  original number.
* `HIGH`: Slower calculation, but more accurate approximation to the
  original number.
* `HIGHEST`: Slowest calculation, but most accurate approximation to the
  original number.
```
For an implementation, `DEFAULT` and `HIGH` seem like a license to do anything. I hope #755 resolves this 🤞.

My approach has been to **at least** use the precision of the input types, but that's certainly not endorsed by all of my colleagues who want "benchmarks" to go fast.

rfcs/20230321-fp8_fnuz.md
jakeh-gc: I've update the doc. Let me know if you prefer a different wording.

cantonios: sg, this is exactly what I meant, thanks.

cantonios (approved):

rfcs/20230321-fp8_fnuz.md
```diff
@@ -0,0 +1,170 @@
...
+backends freedom to exploit any hardware support.
+
+Here's an example of a real JAX program (logging the MLIR) computing a simple
+dot product in Float8E5M2FNUZ. Note the answer is slightly "wrong", as expected
+due to the lower precision.
+
+```python
+>>> import jax
+>>> import jax.numpy as jnp
+>>> x = jnp.arange(16, dtype=jnp.float8_e5m2fnuz)
```
amberhuffman, line 120: There are a lot of FP8 variants across hardware these days and discussion about standards (e.g., IEEE P3109). Should we have a discussion as an OpenXLA Community on the principles for approving FP8 dtypes? Having such principles will hopefully, in the end, allow us to simplify and have fewer than 5 dtypes for representing FP8.

burmako: Thank you for the feedback! I agree that it would be a useful discussion to have as a community.

In [the first LLVM FP8 RFC](https://discourse.llvm.org/t/rfc-add-apfloat-and-mlir-type-support-for-fp8-e5m2/65279) from September 2022, @stellaraccident wrote "There are also a bounded number of floating point types in use for current and upcoming hardware, and we can just implement them like this", and this is the convention that has been followed to date - e.g. [MLIR has 5 dtypes](https://github.com/llvm/llvm-project/blob/87b5807d3802b932c06d83c4287014872aa2caab/mlir/include/mlir/IR/BuiltinTypes.h#L48-L52).

I think this is a reasonable convention for StableHLO to follow as well, but I also think there is a StableHLO-specific part of this discussion that would be good to talk about. There is an OpenXLA community meeting coming up on 4/18, but we could also consider meeting sooner than that. I will create an openxla-discuss topic to align on the date and time of the meeting.

More specifically, for the agenda of this meeting I propose that we arrive at an agreed upon OpenXLA-specific convention for how we should add FP8 dtypes to OpenXLA. This convention would then apply to both the already supported types and the types under review in this RFC. Please let me know what you think.

jakeh-gc: As additional context, this is also the approach being taken by onnx https://github.com/onnx/onnx/pull/4805.

HaiShaw: From equality point of view, adding this types and Google's [the F8E4M3B11FNUZ RFC](https://github.com/openxla/stablehlo/blob/main/rfcs/20230309-e4m3b11.md) all look fine.

From an end user (model developer), all these variants do not look clear but only confusing more (thanks to lack of standard and fast-pace in the space). 

A natural question arises is - Can our ML system software (framework, libraries and compiler) manage to expose minimum set of FP8 dtypes to ML users by abstracting away hardware or vendor specifics from many variants. If we choose an exposable type or two as open exchange format, we can enforce it be used in the debug/console/chkpt/etc. viewers (those only matters to external users); and internally we can enable operations in vendor specific or native formats of different current or future hardware.

Again, this is not an explicit opinion on this RFC, but a suggestion to further discussions.

awf: > ... have fewer than 5 dtypes for representing FP8. 

The premise here appears to be that there is a downside to having many FP8 types.  This is not perceived to be an issue by LLVM, ONNX, and MLIR, as evidenced by their adoption of the multiple types.  I know from personal communication that the Tensorflow team also see no problem with multiple (six was mentioned) dtypes.  There are so many dtypes today because hardware manufacturers are experimenting with different designs, balancing, for example, dynamic range with the number of special codes.  These are not theoretical designs, but hardware available in market, and running commercial AI workloads.   To insist on a smaller number of dtypes would be to suppress innovation (and competition) in an emerging area.  When this argument was presented to IEEE Working Group P3109 on "Arithmetic Formats for Machine Learning", it was also rejected.

Now this is not in contradiction with @HaiShaw's question
> Can [we] manage to expose minimum set of FP8 dtypes to ML users by abstracting away hardware or vendor specifics from many variants

That is indeed a valuable line of investigation (and might ultimately involve adding two more FP8 types) but the current types--that represent what underlying hardware actually does--will still be needed at the framework level, even if future users generally use the software-defined abstractions rather than the hardware dtypes.   But of course this line of investigation should not delay the incorporation of the types that allow us to talk to hardware today, not least because it's not at all obvious how long such a design will take to invent and converge on.

cantonios: > I know from personal communication that the Tensorflow team also see no problem with multiple (six was mentioned) dtypes

This is not true, speaking from the TensorFlow team.  We _do_ see a problem.  From a usability perspective, having so many types makes model development challenging, and will likely tie your model to specific hardware.  There are important questions about how to make a general API at the framework level to allow for model portability and to simplify usage for non-experts.  Though we do want power users to still have direct control if they do know what they're doing.

To allow any of this though, we do need these types to exist at the StableHLO level.

awf: > > I know from personal communication that the Tensorflow team also see no problem with multiple (six was mentioned) dtypes
> 
> 
> 
> This is not true, speaking from the TensorFlow team.

Apologies, I should maybe have said 'the tensorflow team members with whom I last discussed this'.

It does sound, though, as if we are in agreement that these types are needed now.

reedwm: > Apologies, I should maybe have said 'the tensorflow team members with whom I last discussed this'.

Yeah I think it's fine adding the 5 dtypes (is there a 6th?), both to TF and XLA. But I only work on XLA now, not TF, so @cantonios opinion matters more here for the TF side. If XLA has less than 5 dtypes, it's infeasible for TF to have all 5. The inverse is not true though: If XLA has 5 dtypes, it's possible for TF to have less than 5 (but I think this would be very difficult to implement in practice, as, among other reasons, the [DType object](https://www.tensorflow.org/api_docs/python/tf/dtypes/DType) has attributes like `max` which differ among the FP8 types).

jakeh-gc: I'm going to tentatively mark this thread as "resolved" unless there's any objection, at least in the scope of this RFC.

I think we all mostly agree on the wider topic, and we happen to be talking about different parts of the software stack. We must keep in mind XLA's users (as @reedwm points out, TF authors, JAX authors, Torch-XLA authors, and others) might make different decisions about exposing FP8 (1, 2, or many types), and we should support all of that.

If we consolidated the FP8 types (or any types) at **this** level of the stack, we would be making a value judgement that might not align with our users and would be difficult for them to work around.

amberhuffman: Sounds good. The impetus for my comment was that the OpenXLA Community should discuss principles (like what emerged in this thread). Related, another vendor was advocating emulation of their FP8 variant due to their concern over user confusion, etc. It would be good to have principles documented over time (e.g., when to include something in StableHLO, when to deprecate, etc). However, that does not need to slow progress, versus something to do as the OpenXLA Community builds over time.

burmako:
Thank you, everyone, for the lively discussion! It's been two weeks, and we have broad support for adding the 2 new dtypes, with all the feedback addressed. The RFC is now approved.

A recurring theme in the discussions of the last few weeks has been working out a long-term story for the fp8 dtypes. I think this is important to follow up on, and I'll work with @theadactyl on including this on the agenda of our community meetings.

burmako (approved):

rfcs/20230321-fp8_fnuz.md
wenchenvincent: > > Apologies, I should maybe have said 'the tensorflow team members with whom I last discussed this'.
> 
> Yeah I think it's fine adding the 5 dtypes (is there a 6th?), both to TF and XLA. But I only work on XLA now, not TF, so @cantonios opinion matters more here for the TF side. If XLA has less than 5 dtypes, it's infeasible for TF to have all 5. The inverse is not true though: If XLA has 5 dtypes, it's possible for TF to have less than 5 (but I think this would be very difficult to implement in practice, as, among other reasons, the [DType object](https://www.tensorflow.org/api_docs/python/tf/dtypes/DType) has attributes like `max` which differ among the FP8 types).

Very much agreed. I think it would be very beneficial to consider holistically what to expose at each level of abstraction. Different fp8 data types would be needed at lower levels (e.g. llvm, MLIR), but as we go up the level of abstraction, it might be possible to hide the difference of these details. It would be a good exercise to clarify the set of users for each level of abstraction and what's necessary for them at each level in the community discussion.

PR #1343, opened 2023-03-22
Fix CompareOp interpreter interface inconsistency

anakinxc:
It looks like the decl of evalCompareOp has an extra `compareType` parameter.


sdasgup3:
Thanks for the fix which LGTM.

@ghpvnist Maybe we should make sure the use of evalOp* in `eval` function (https://github.com/openxla/stablehlo/blob/311f14ce78c7fe35d304ee91007b58c335cf821e/stablehlo/reference/Ops.cpp#L484) to rely on the declarations in `Ops.h` for type checking. One solution might be to put that function at the beginning of `Ops.cpp`.

sdasgup3 (approved):

PR #1352, opened 2023-03-24
Introduce QuantizedType

sdasgup3:
StableHLO dialect currently supports quantization via:
  1) Supporting `quant.uniform` element types.
  2) Having dedicated ops like `uniform_quantize` / `uniform_dequantize`.
  3) Allowing regular ops like `add` / `convolution` to take quantized tensors.

This support was inherited from MHLO when StableHLO was bootstrapped, and
MHLO support was motivated by mobile use cases and inherited from TFLite.

As pointed out in #1149, StableHLO specification doesn't support quantization
at the moment, and this is an important gap that we would like to fix before
StableHLO v1.0 (see #588).

To continue the discussion started in #1149 and to make progress towards v1.0,
this pull request:
  A) Adds QuantizedType to the StableHLO specification, modelled after
     [TFLite quantization spec](https://www.tensorflow.org/lite/performance/quantization_spec).
  B) To start a conversation about the applications of QuantizedType and the
     semantics of quantized ops, proposes semantics for quantized `add`.

TFLite quantization spec doesn't cover everything. It specs constraints on
types (which we captured accordingly in this pull request), but it doesn't go
into describing semantics of quantized ops.

As a result, the proposed semantics for quantized `add` is intentionally naive,
as compared with the much more involved implementations in the TensorFlow
repository, e.g.:
  * [tfl.add](https://www.google.com/url?sa=D&q=https%3A%2F%2Fgithub.com%2Ftensorflow%2Ftensorflow%2Fblob%2Fmaster%2Ftensorflow%2Flite%2Fkernels%2Fadd.cc).
  * [tf.UniformQuantizedAdd](https://www.google.com/url?sa=D&q=https%3A%2F%2Fgithub.com%2Ftensorflow%2Ftensorflow%2Fblob%2Fmaster%2Ftensorflow%2Fcore%2Fkernels%2Funiform_quant_ops%2Funiform_quantized_add_op.cc).

In this pull request, we are looking for feedback from the community to
determine the right level of detail for speccing quantized types and quantized
ops. Please let us know!

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -575,13 +575,13 @@ Performs element-wise addition of two tensors `lhs` and `rhs` and produces a
 | Label | Name  | Type   | Constraints |
 |-------|-------|--------|-------------|
 | (I1)  | `lhs` | tensor | (C1)-(C6)   |
-| (I2)  | `rhs` | tensor | (C1)-(C6)   |
+| (I2)  | `rhs` | tensor | (C1)-(C3)   |
 
 #### Outputs
 
 | Name     | Type   | Constraints |
 |----------|--------|-------------|
-| `result` | tensor | (C1)-(C6)   |
+| `result` | tensor | (C1)-(C3)   |
 
 #### Constraints
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -213,19 +213,24 @@ Quantized types satisfy the following constraints:
 * (C1) `num_bits(storage_type) < num_bits(expressed_type)`.
 * (C2) `type(storage_min) = storage_type`.
 * (C3) `type(storage_max) = storage_type`.
-* (C4) `storage_min` and `storage_max` have the following values:
-  * If `is_signed(storage_type)` and `d = num_bits(storage_type)`:
-    * `storage_min = -2^(d-1)`, `storage_max = 2^(d-1)-1` OR
-    * `storage_min = -2^(d-1)+1`, `storage_max = 2^(d-1)-1`.
-  * If `is_unsigned(storage_type)`:
-    * `storage_min = 0`, `storage_max = 2^d - 1`.
-* (C5) For all `i`, `type(scales[i]) = expressed_type`.
-* (C6) For all `i`, `scales[i] > 0`.
-* (C7) For all `i`, `storage_min <= zero_points[i] <= storage_max`.
-* (C8) For all `i`, `type(zero_points[i]) = storage_type`.
-* (C9) `size(scales) = size(zero_points)`.
-* (C10) If `quantization_dimension` is empty, then `size(scales) = 1`.
-* (C11) If `quantization_dimension` is not empty, then
+* (C4) If `storage_min` is empty and `d = num_bits(storage_type)`:
+  * `storage_min = is_signed(storage_type) ? -2^(d-1) : 0`.
+* (C5) If `storage_max` is empty:
+  * `storage_max = is_signed(storage_type) ? 2^(d-1) : 2^d - 1`.
+* (C6) If `storage_min` is not empty:
+  * If `is_signed(storage_type)`, `storage_min >= -2^(d-1)`.
+  * If `is_unsigned(storage_type)`, `storage_min >= 0`.
+* (C7) If `storage_max` is not empty:
+  * If `is_signed(storage_type)`, `storage_max <= 2^(d-1)-1`
+  * If `is_unsigned(storage_type)`, `storage_max <= 2^d - 1`.
+* (C8) `storage_max - storage_min > 0`.
+* (C9) For all `i`, `type(scales[i]) = expressed_type`.
+* (C10) For all `i`, `scales[i] > 0`.
+* (C11) For all `i`, `storage_min <= zero_points[i] <= storage_max`.
+* (C12) For all `i`, `type(zero_points[i]) = storage_type`.
+* (C13) `size(scales) = size(zero_points)`.
+* (C14) If `quantization_dimension` is empty, then `size(scales) = 1`.
+* (C15) If `quantization_dimension` is not empty, then
   `0 <= quantization_dimension`.
 
 Furthermore, tensors of quantized types satisfy the following constraints:
@@ -233,8 +238,8 @@ Furthermore, tensors of quantized types satisfy the following constraints:
 * For per-tensor quantization:
   * No additional constraints.
 * For per-axis quantization:
-  * (C12) `quantization_dimension < size(shape)`.
-  * (C13) `size(scales) = shape[quantization_dimension]`.
+  * (C16) `quantization_dimension < size(shape)`.
+  * (C17) `size(scales) = shape[quantization_dimension]`.
 
 ```ebnf
 FunctionType ::= '(' [ValueType {',' ValueType}] ')' '->' '(' [ValueType {',' ValueType}] ')'
@@ -592,8 +597,9 @@ Performs element-wise addition of two tensors `lhs` and `rhs` and produces a
 * If the operation uses quantized types:
   * (C3) `element_type(lhs) = element_type(rhs) = element_type(result)`,
     except for scales and zero points which may differ.
-  * (C4) `abs(storage_min(lhs)) != abs(storage_max(lhs))`.
-  * (C5) `quantization_dimension(lhs)` is empty.
+  * (C4) `storage_min = is_signed(storage_type) ? -2^(d-1) : 0, where d = num_bits(storage_type)`.
+  * (C5) `storage_max = is_signed(storage_type) ? 2^(d-1) : 2^d - 1`.
+  * (C6) `quantization_dimension(lhs)` is empty.
 
 #### Examples
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -228,8 +228,8 @@ Furthermore, tensors of quantized types satisfy the following constraints:
 * For per-tensor quantization:
   * No additional constraints.
 * For per-axis quantization:
-  * (C10) `size(scales) = shape[quantization_dimension]`.
   * (C11) `quantization_dimension < size(shape)`.
+  * (C12) `size(scales) = shape[quantization_dimension]`.
 
 ```ebnf
 FunctionType ::= '(' [ValueType {',' ValueType}] ')' '->' '(' [ValueType {',' ValueType}] ')'
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -146,20 +146,23 @@ IntegerType ::= 'si4' | 'si8' | 'si16' | 'si32' | 'si64'
               | 'ui4' | 'ui8' | 'ui16' | 'ui32' | 'ui64'
 FloatType   ::= 'f8E4M3FN' | 'f8E5M2' | 'bf16' | 'f16' | 'f32' | 'f64'
 ComplexType ::= 'complex' '<' ('f32' | 'f64') '>'
-QuantizedType ::= '!quant.uniform' '<' QuantizationStorageType
-                  ':' QuantizationExpressedType
-                  ':' QuantizationStorageMin
-                  ':' QuantizationStorageMax
-                  ':' QuantizationScales
-                  ':' QuantizationZeroPoints
-                  {':' QuantizationDimension} '>'
+
+QuantizedType ::= '!quant.uniform' '<' QuantizationStorageTypeSpec
+                ':' QuantizationExpressedType
+                [':' QuantizationDimension]
+                ',' QuantizationScalesAndZeroPoints '>'
+QuantizationStorageTypeSpec ::= QuantizationStorageType
+                              ['<' QuantizationStorageMin ':' QuantizationStorageMax '>']
 QuantizationStorageType ::= IntegerType
-QuantizationExpressedType ::= FloatType
 QuantizationStorageMin ::= IntegerConstant
 QuantizationStorageMax ::= IntegerConstant
-QuantizationScales ::= '[' FloatConstant {',' FloatConstant} ']'
-QuantizationZeroPoints ::= '[' IntegerConstant {',' IntegerConstant} ']'
+QuantizationExpressedType ::= FloatType
 QuantizationDimension ::= IntegerConstant
+QuantizationScalesAndZeroPoints ::= (QuantizationScaleAndZero
+                                  | '{' QuantizationScaleAndZero {',' QuantizationScaleAndZero} '}')
+QuantizationScaleAndZero ::= QuantizationScale ':' QuantizationZeroPoint
+QuantizationScale ::=  FloatConstant
+QuantizationZeroPoint ::=  IntegerConstant
 ```
 
 **Element types** represent elements of tensor types. Unlike in many programming
@@ -215,12 +218,12 @@ Quantized types satisfy the following constraints:
     * `storage_min = -2^(d-1)+1`, `storage_max = 2^(d-1)-1`.
   * If `is_unsigned(storage_type)`:
     * `storage_min = 0`, `storage_max = 2^d - 1`.
-* (C5) For all `i`, `type(scales[i]) = expressed_type`.
-* (C6) For all `i`, `scales[i] > 0`.
-* (C7) For all `i`, `type(zero_points[i]) = i64`.
-* (C8) `size(scales) = size(zero_points)`.
-* (C9) If `quantization_dimension` is empty, then `size(scales) = 1`.
-* (C10) If `quantization_dimension` is not empty, then
+* (C5) `type(scale) = expressed_type`.
+* (C6) `scale > 0`.
+* (C7) `type(zero_point) = i64`.
+* (C8) If `quantization_dimension` is empty, then
+  `size(scales_and_zero_points) = 1`.
+* (C9) If `quantization_dimension` is not empty, then
   `0 <= quantization_dimension`.
 
 Furthermore, tensors of quantized types satisfy the following constraints:
@@ -228,8 +231,8 @@ Furthermore, tensors of quantized types satisfy the following constraints:
 * For per-tensor quantization:
   * No additional constraints.
 * For per-axis quantization:
-  * (C11) `quantization_dimension < size(shape)`.
-  * (C12) `size(scales) = shape[quantization_dimension]`.
+  * (C10) `quantization_dimension < size(shape)`.
+  * (C11) `size(scales_and_zero_points) = shape[quantization_dimension]`.
 
 ```ebnf
 FunctionType ::= '(' [ValueType {',' ValueType}] ')' '->' '(' [ValueType {',' ValueType}] ')'
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -566,36 +566,23 @@ Performs element-wise addition of two tensors `lhs` and `rhs` and produces a
 * For integers: integer addition.
 * For floats: `addition` from IEEE-754.
 * For complex numbers: complex addition.
-* For quantized types:
-  * `float_result = (lhs - zero_point(lhs)) * scale(lhs) +
-    (rhs - zero_point(rhs)) * scale(rhs)`.
-  * `rounded_result = round_nearest_even(float_result / scale(result))`.
-  * `result = clamp(storage_min(result), rounded_result + zero_point(result), storage_max(result))`.
 
 #### Inputs
 
-| Label | Name  | Type                       | Constraints |
-|-------|-------|----------------------------|-------------|
-| (I1)  | `lhs` | tensor or quantized tensor | (C1)-(C6)   |
-| (I2)  | `rhs` | tensor or quantized tensor | (C1)-(C3)   |
+| Label | Name  | Type   | Constraints |
+|-------|-------|--------|-------------|
+| (I1)  | `lhs` | tensor | (C1)        |
+| (I2)  | `rhs` | tensor | (C1)        |
 
 #### Outputs
 
-| Name     | Type                       | Constraints |
-|----------|----------------------------|-------------|
-| `result` | tensor or quantized tensor | (C1)-(C3)   |
+| Name     | Type   | Constraints |
+|----------|--------|-------------|
+| `result` | tensor | (C1)        |
 
 #### Constraints
 
-* (C1) `shape(lhs) = shape(rhs) = shape(result)`.
-* If the operation uses non-quantized tensors:
-  * (C2) `element_type(lhs) = element_type(rhs) = element_type(result)`.
-* If the operation uses quantized tensors:
-  * (C3) `element_type(lhs) = element_type(rhs) = element_type(result)`,
-    except for scales and zero points which may differ.
-  * (C4) `storage_min(lhs) = min_value(storage_type)`.
-  * (C5) `storage_max(lhs) = max_value(storage_type)`.
-  * (C6) `quantization_dimension(lhs)` is empty.
+* (C1) `lhs`, `rhs` and `result` have the same type.
 
 #### Examples
 
@@ -2189,11 +2176,9 @@ tensor and produces a `result` tensor.
 #### Semantics
 
 Encapsulates an implementation-defined operation `call_target_name` that takes
-`inputs` and `called_computations` and produces `results`. The metadata
-`has_side_effect` is used to prevent optimizations like speculation (e.g
-loop-invariant code motion) and dead-code elimination. `backend_config`
-and `api_version` may be used to provide additional implementation-defined
-metadata.
+`inputs` and `called_computations` and produces `results`. `has_side_effect`,
+`backend_config` and `api_version` may be used to provide additional
+implementation-defined metadata.
 
 At the moment, this operation contains a fairly disorganized collection of
 metadata which reflects organic evolution of its counterpart operation in
@@ -2869,10 +2854,10 @@ Produces the size of the given `dimension` of the `operand`.
 
 #### Inputs
 
-| Label | Name        | Type                         | Constraints |
-|-------|-------------|------------------------------|-------------|
-| (I1)  | `operand`   | tensor                       | (C1)        |
-| (I2)  | `dimension` | constant of type `si64`      | (C1)        |
+| Label | Name        | Type                    | Constraints |
+|-------|-------------|-------------------------|-------------|
+| (I1)  | `operand`   | tensor                  | (C1)        |
+| (I2)  | `dimension` | constant of type `si64` | (C1)        |
 
 #### Outputs
 
@@ -4718,11 +4703,11 @@ where `pred_val = rank(pred) == 0 ? pred : pred[i0, ..., iR-1]`.
 
 #### Inputs
 
-| Label | Name       | Type                         | Constraints |
-|-------|------------|------------------------------|-------------|
-| (I1)  | `pred`     | tensor of type `i1`          | (C1)        |
-| (I2)  | `on_true`  | tensor                       | (C1), (C2)  |
-| (I3)  | `on_false` | tensor                       | (C2)        |
+| Label | Name       | Type                | Constraints |
+|-------|------------|---------------------|-------------|
+| (I1)  | `pred`     | tensor of type `i1` | (C1)        |
+| (I2)  | `on_true`  | tensor              | (C1), (C2)  |
+| (I3)  | `on_false` | tensor              | (C2)        |
 
 #### Outputs
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -197,7 +197,9 @@ values of type `tensor<T>`).
   `f = (i - zero_point) * scale`, where `scale` and `zero_point` are called
   **quantization parameters**. The `storage_min` and `storage_max`, being
   optional, have default values which should satisfy the associated constraints
-  as laid out in this document.
+  as laid out in this document. The `storage_min` and `storage_max` are optional
+  in the grammar, but have default values of `min_value(storage_type)` and
+  `max_value(storage_type)` respectively.
 
 Quantization can be **per-tensor**, meaning, having one `scale` and/or
 `zero_point` for the entire tensor or can be **per-axis**, meaning, having
@@ -213,24 +215,16 @@ Quantized types satisfy the following constraints:
 * (C1) `num_bits(storage_type) < num_bits(expressed_type)`.
 * (C2) `type(storage_min) = storage_type`.
 * (C3) `type(storage_max) = storage_type`.
-* (C4) If `storage_min` is empty and `d = num_bits(storage_type)`:
-  * `storage_min = is_signed(storage_type) ? -2^(d-1) : 0`.
-* (C5) If `storage_max` is empty:
-  * `storage_max = is_signed(storage_type) ? 2^(d-1) : 2^d - 1`.
-* (C6) If `storage_min` is not empty:
-  * If `is_signed(storage_type)`, `storage_min >= -2^(d-1)`.
-  * If `is_unsigned(storage_type)`, `storage_min >= 0`.
-* (C7) If `storage_max` is not empty:
-  * If `is_signed(storage_type)`, `storage_max <= 2^(d-1)-1`
-  * If `is_unsigned(storage_type)`, `storage_max <= 2^d - 1`.
-* (C8) `storage_max - storage_min > 0`.
-* (C9) For all `i`, `type(scales[i]) = expressed_type`.
-* (C10) For all `i`, `scales[i] > 0`.
-* (C11) For all `i`, `storage_min <= zero_points[i] <= storage_max`.
-* (C12) For all `i`, `type(zero_points[i]) = storage_type`.
-* (C13) `size(scales) = size(zero_points)`.
-* (C14) If `quantization_dimension` is empty, then `size(scales) = 1`.
-* (C15) If `quantization_dimension` is not empty, then
+* (C4) `storage_min >= min_value(storage_type)`.
+* (C5) `storage_max <= max_value(storage_type)`.
+* (C6) `storage_max - storage_min > 0`.
+* (C7) For all `i`, `type(scales[i]) = expressed_type`.
+* (C8) For all `i`, `scales[i] > 0`.
+* (C9) For all `i`, `storage_min <= zero_points[i] <= storage_max`.
+* (C10) For all `i`, `type(zero_points[i]) = storage_type`.
+* (C11) `size(scales) = size(zero_points)`.
+* (C12) If `quantization_dimension` is empty, then `size(scales) = 1`.
+* (C13) If `quantization_dimension` is not empty, then
   `0 <= quantization_dimension`.
 
 Furthermore, tensors of quantized types satisfy the following constraints:
@@ -238,8 +232,8 @@ Furthermore, tensors of quantized types satisfy the following constraints:
 * For per-tensor quantization:
   * No additional constraints.
 * For per-axis quantization:
-  * (C16) `quantization_dimension < size(shape)`.
-  * (C17) `size(scales) = shape[quantization_dimension]`.
+  * (C14) `quantization_dimension < size(shape)`.
+  * (C15) `size(scales) = shape[quantization_dimension]`.
 
 ```ebnf
 FunctionType ::= '(' [ValueType {',' ValueType}] ')' '->' '(' [ValueType {',' ValueType}] ')'
@@ -597,8 +591,8 @@ Performs element-wise addition of two tensors `lhs` and `rhs` and produces a
 * If the operation uses quantized types:
   * (C3) `element_type(lhs) = element_type(rhs) = element_type(result)`,
     except for scales and zero points which may differ.
-  * (C4) `storage_min = is_signed(storage_type) ? -2^(d-1) : 0, where d = num_bits(storage_type)`.
-  * (C5) `storage_max = is_signed(storage_type) ? 2^(d-1) : 2^d - 1`.
+  * (C4) `storage_min = min_value(storage_type)`.
+  * (C5) `storage_max = max_value(storage_type)`.
   * (C6) `quantization_dimension(lhs)` is empty.
 
 #### Examples
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -146,22 +146,21 @@ IntegerType ::= 'si4' | 'si8' | 'si16' | 'si32' | 'si64'
               | 'ui4' | 'ui8' | 'ui16' | 'ui32' | 'ui64'
 FloatType   ::= 'f8E4M3FN' | 'f8E5M2' | 'bf16' | 'f16' | 'f32' | 'f64'
 ComplexType ::= 'complex' '<' ('f32' | 'f64') '>'
-
-QuantizedType ::= '!quant.uniform' '<' QuantizationStorageTypeSpec
-                ':' QuantizationExpressedType
-                [':' QuantizationDimension]
-                ',' QuantizationScalesAndZeroPoints '>'
-QuantizationStorageTypeSpec ::= QuantizationStorageType
-                              ['<' QuantizationStorageMin ':' QuantizationStorageMax '>']
+QuantizedType ::= '!quant.uniform' '<'
+                  QuantizationStorageType
+                  ['<' QuantizationStorageMin ':' QuantizationStorageMax '>']
+                  ':' QuantizationExpressedType
+                  [':' QuantizationDimension]
+                  ',' QuantizationParameters '>'
 QuantizationStorageType ::= IntegerType
 QuantizationStorageMin ::= IntegerConstant
 QuantizationStorageMax ::= IntegerConstant
 QuantizationExpressedType ::= FloatType
 QuantizationDimension ::= IntegerConstant
-QuantizationScalesAndZeroPoints ::= (QuantizationScaleAndZero
-                                  | '{' QuantizationScaleAndZero {',' QuantizationScaleAndZero} '}')
-QuantizationScaleAndZero ::= QuantizationScale ':' QuantizationZeroPoint
-QuantizationScale ::=  FloatConstant
+QuantizationParameters ::= QuantizationParameter
+                         | '{' QuantizationParameter {',' QuantizationParameter} '}'
+QuantizationParameter ::= QuantizationScale ':' QuantizationZeroPoint
+QuantizationScale ::= FloatConstant
 QuantizationZeroPoint ::=  IntegerConstant
 ```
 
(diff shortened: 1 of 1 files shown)


docs/spec.md
```diff
@@ -173,10 +188,48 @@ values of type `tensor<T>`).
...
+* (C10) If `quantization_dimension` is not empty, then
+  `0 <= quantization_dimension`.
+
+Furthermore, tensors of quantized types satisfy the following constraints:
+
+* For per-tensor quantization:
+  * No additional constraints.
+* For per-axis quantization:
+  * (C10) `size(scales) = shape[quantization_dimension]`.
+  * (C11) `quantization_dimension < size(shape)`.
```
loganchien, line 232: Maybe swap the order of C10 & C11?

docs/spec.md
```diff
@@ -140,11 +140,26 @@ which may allow us to remove tuple types from StableHLO
 
 ```ebnf
 ElementType ::= BooleanType | IntegerType | FloatType | ComplexType
+              | QuantizedType
 BooleanType ::= 'i1'
 IntegerType ::= 'si4' | 'si8' | 'si16' | 'si32' | 'si64'
               | 'ui4' | 'ui8' | 'ui16' | 'ui32' | 'ui64'
 FloatType   ::= 'f8E4M3FN' | 'f8E5M2' | 'bf16' | 'f16' | 'f32' | 'f64'
 ComplexType ::= 'complex' '<' ('f32' | 'f64') '>'
+QuantizedType ::= '!quant.uniform' '<' QuantizationStorageType
```
loganchien, line 149: If I read this correctly, we are using different syntax w.r.t. TFLite.  Is this intended?

For example:

```
TFL: !quant.uniform<u8<0,255>:f32, 1.0:0>
HLO: !quant.uniform<ui8:f32:0:255:[1.0]:[0]>

TFL: !quant.uniform<u8<0,255>:f32:3, {1.0:0, 2.0:2}>
HLO: !quant.uniform<ui8:f32:0:255:[1.0, 2.0]:[0, 2]:3>
```

Where `storage_min` and `storage_max` are not optional, scales are grouped together, and `zero_points` are grouped together.

BTW, the unification of signed flag and storage type is good IMO.

docs/spec.md
```diff
@@ -173,10 +188,48 @@ values of type `tensor<T>`).
...
+
+* (C1) `num_bits(storage_type) < num_bits(expressed_type)`.
+* (C2) `type(storage_min) = storage_type`.
+* (C3) `type(storage_max) = storage_type`.
+* (C4) `storage_min` and `storage_max` have the following values:
+  * If `is_signed(storage_type)` and `d = num_bits(storage_type)`:
+    * `storage_min = -2^(d-1)`, `storage_max = 2^(d-1)-1` OR
+    * `storage_min = -2^(d-1)+1`, `storage_max = 2^(d-1)-1`.
+  * If `is_unsigned(storage_type)`:
+    * `storage_min = 0`, `storage_max = 2^d - 1`.
```
loganchien, line 217: Are `storage_min` and `storage_max` user-specified values or derived values from the storage type?

docs/spec.md
sdasgup3: done.

docs/spec.md
sdasgup3: >> If I read this correctly, we are using different syntax w.r.t. TFLite. Is this intended?


Yes, you are right in pointing the differences.
 
While we were speccing the StableHLO opset, we made a conscious decision about not to use MLIR or XLA as the notational vehicle for the spec. I think the reason is to keep the specification immune to any future formatting changes. As an example, in the specification, MLIR attributes are called constants.

I believe the same decision has been extended here to TFL specific notations.

sdasgup3: >> we made a conscious decision about not to use MLIR or XLA as the notational vehicle for the spec

Realized that the decision is more specific to using MLIR notations like Traits, Attributes etc in the specification.  A grammar that follows MLIR closely could benefit from familiarity.

Updated the ebnf grammar together with relevant changes in constraints.
PTAL.

docs/spec.md
sdasgup3: With the above change of grammar, the `storage_min` and `storage_min` can be optional in which case the value can be inferred using a mix of 

(1) The general constraints 
```
* (C4) `storage_min` and `storage_max` have the following values:
  * If `is_signed(storage_type)` and `d = num_bits(storage_type)`:
    * `storage_min = -2^(d-1)`, `storage_max = 2^(d-1)-1` OR
    * `storage_min = -2^(d-1)+1`, `storage_max = 2^(d-1)-1`.
  * If `is_unsigned(storage_type)`:
    * `storage_min = 0`, `storage_max = 2^d - 1`.
```

(2) Op specific constraints, like for AddOp we have
```
  * (C4) `abs(storage_min(lhs)) != abs(storage_max(lhs))`.
```

docs/spec.md
```diff
@@ -140,11 +140,29 @@ which may allow us to remove tuple types from StableHLO
...
+QuantizationStorageType ::= IntegerType
+QuantizationStorageMin ::= IntegerConstant
+QuantizationStorageMax ::= IntegerConstant
+QuantizationExpressedType ::= FloatType
+QuantizationDimension ::= IntegerConstant
+QuantizationScalesAndZeroPoints ::= (QuantizationScaleAndZero
+                                  | '{' QuantizationScaleAndZero {',' QuantizationScaleAndZero} '}')
+QuantizationScaleAndZero ::= QuantizationScale ':' QuantizationZeroPoint
+QuantizationScale ::=  FloatConstant
+QuantizationZeroPoint ::=  IntegerConstant
```
burmako, line 165: Proposed rewording:

```ebnf
QuantizedType ::= '!quant.uniform' '<'
                  QuantizationStorageType
                  ['<' QuantizationStorageMin ':' QuantizationStorageMax '>']
                  ':' QuantizationExpressedType
                  [':' QuantizationDimension]
                  ',' QuantizationParameters '>'
QuantizationStorageType ::= IntegerType
QuantizationStorageMin ::= IntegerConstant
QuantizationStorageMax ::= IntegerConstant
QuantizationExpressedType ::= FloatType
QuantizationDimension ::= IntegerConstant
QuantizationParameters ::= QuantizationParameter
                         | '{' QuantizationParameter {',' QuantizationParameter} '}'
QuantizationParameter ::= QuantizationScale ':' QuantizationZeroPoint
QuantizationScale ::= FloatConstant
QuantizationZeroPoint ::=  IntegerConstant
```

Feedback:
  1) No newline needed between ComplexType and QuantizationType.
  2) Aligned some lines similarly to other places in the spec.
  3) Inlined QuantizationStorageTypeSpec - it's just a regular production used only once.
  4) Renamed QuantizationScalesAndZeroPoints to QuantizationParameters.

docs/spec.md
burmako: Given that storage_min and storage_max are optional in the new grammar, where do we specify their default values?

docs/spec.md
```diff
@@ -215,21 +218,21 @@ Quantized types satisfy the following constraints:
...
     * `storage_min = 0`, `storage_max = 2^d - 1`.
-* (C5) For all `i`, `type(scales[i]) = expressed_type`.
-* (C6) For all `i`, `scales[i] > 0`.
-* (C7) For all `i`, `type(zero_points[i]) = i64`.
-* (C8) `size(scales) = size(zero_points)`.
-* (C9) If `quantization_dimension` is empty, then `size(scales) = 1`.
-* (C10) If `quantization_dimension` is not empty, then
+* (C5) `type(scale) = expressed_type`.
+* (C6) `scale > 0`.
+* (C7) `type(zero_point) = i64`.
```
burmako, line 223: I think we should still keep `scales` and `zero_points` here. I understand that in the new grammar they don't exist as top-level non-terminals, but so do `scale` and `zero_point`.

burmako (changes requested):

docs/spec.md
```diff
@@ -173,10 +188,48 @@ values of type `tensor<T>`).
...
+* (C3) `type(storage_max) = storage_type`.
+* (C4) `storage_min` and `storage_max` have the following values:
+  * If `is_signed(storage_type)` and `d = num_bits(storage_type)`:
+    * `storage_min = -2^(d-1)`, `storage_max = 2^(d-1)-1` OR
+    * `storage_min = -2^(d-1)+1`, `storage_max = 2^(d-1)-1`.
+  * If `is_unsigned(storage_type)`:
+    * `storage_min = 0`, `storage_max = 2^d - 1`.
+* (C5) For all `i`, `type(scales[i]) = expressed_type`.
+* (C6) For all `i`, `scales[i] > 0`.
+* (C7) For all `i`, `type(zero_points[i]) = i64`.
```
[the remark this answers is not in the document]
sdasgup3: Thanks @dominicsymes for your valuable feedback.

Since `zero_point` can get out of the  storage value range, the proposal is using a wider type `i64` than the storage type, with max limit as [32](https://github.com/llvm/llvm-project/blob/0676c6d91fef302e0ae17f1a4571f28b1be1b27c/mlir/include/mlir/Dialect/Quant/QuantTypes.h#L58).


>> Additionally, I think zero points are most used for small data types – such 8-bit and below.

That is pretty useful information and can help in restricting the `zero_point` type. I will explore it further and welcome feedback from other reviewers.

docs/spec.md
```diff
@@ -507,23 +560,35 @@ Performs element-wise addition of two tensors `lhs` and `rhs` and produces a
 * For integers: integer addition.
 * For floats: `addition` from IEEE-754.
 * For complex numbers: complex addition.
+* For quantized types:
```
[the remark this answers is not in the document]
burmako: In the current version of the PR, we started with the TFLite quantization spec and were looking to get feedback from users on what they would be looking for. Thank you for providing yours! We've also pinged a few other groups to share their thoughts, so how about we keep this thread open for now and then decide?

docs/spec.md
```diff
@@ -140,11 +140,29 @@ which may allow us to remove tuple types from StableHLO
...
+                              ['<' QuantizationStorageMin ':' QuantizationStorageMax '>']
+QuantizationStorageType ::= IntegerType
+QuantizationStorageMin ::= IntegerConstant
+QuantizationStorageMax ::= IntegerConstant
+QuantizationExpressedType ::= FloatType
+QuantizationDimension ::= IntegerConstant
+QuantizationScalesAndZeroPoints ::= (QuantizationScaleAndZero
+                                  | '{' QuantizationScaleAndZero {',' QuantizationScaleAndZero} '}')
+QuantizationScaleAndZero ::= QuantizationScale ':' QuantizationZeroPoint
+QuantizationScale ::=  FloatConstant
```
burmako, line 164: @dominicsymes Following up on integer only processing, what if we had two different versions of QuantizedType - one as proposed in the PR, and another one where a floating-point scale is replaced with integer multiplier and shift. Would that work for you with respect to #1149?

This idea was proposed by @sngyhan in an internal discussion, and I think that it's pretty neat. I think that it allows implementations to more easily detect different styles of quantization, which is nice. Also, through FP8, we already have a precedent for providing multiple ways of doing similar things through similar types, and this proposal goes along the same lines. Please let us know what you think!

docs/spec.md
```diff
@@ -173,10 +191,48 @@ values of type `tensor<T>`).
...
+* (C3) `type(storage_max) = storage_type`.
+* (C4) `storage_min` and `storage_max` have the following values:
+  * If `is_signed(storage_type)` and `d = num_bits(storage_type)`:
+    * `storage_min = -2^(d-1)`, `storage_max = 2^(d-1)-1` OR
+    * `storage_min = -2^(d-1)+1`, `storage_max = 2^(d-1)-1`.
+  * If `is_unsigned(storage_type)`:
+    * `storage_min = 0`, `storage_max = 2^d - 1`.
+* (C5) `type(scale) = expressed_type`.
+* (C6) `scale > 0`.
+* (C7) `type(zero_point) = i64`.
```
loganchien, line 223: In addition to the type constraint, we need another value range constraint on `zero_point`s:

```
(C?) `storage_min <= zero_point <= storage_max`
```

We need this constraint because we must ensure that floating point zero `0.0` can always be quantized into a quantized value so that we can pad the activation properly (e.g. convolution padding).

docs/spec.md
sdasgup3: change will be accepted in the next checkin.

docs/spec.md
sdasgup3: will be restored in the next commit.

docs/spec.md
loganchien: Yes.  "The floating point `0.0` will always lie in the range" will be implied if we add this constraint.

loganchien: Even though I can't find a good source of documentation, many math don't work if the floating point `0.0` can't be exactly quantized.  For example, in TFLite optimizer, they need some adjustment to ensure `0.0` is included:

https://github.com/tensorflow/tensorflow/blob/f93185001ff1d7b75e9966b80a144601d7d6e696/tensorflow/lite/tools/optimize/quantization_utils.cc#L74

Also, mentioned in [Quantization and Training of Neural Networks for Efficient Integer-Arithmetic-Only Inference (Jacob et al, 2017)](https://arxiv.org/pdf/1712.05877.pdf) p.3.

docs/spec.md
eric-k256: @dominicsymes is out of the office at the moment. I won't speak for him, but I'll add a bit here, and he can add more when he's back.
In my view, it's not a completely bad option. If we implemented this, there should be a pass that was able to convert between the two formats with a fully defined result. If there is agreement on that pass, then all integer based implementations would agree, and an integer based quantized tensor could be converted to a floating-point version.

docs/spec.md
sdasgup3: I agree that we should clarify that. 
This is my proposal: 
```
**Quantized types** represent ... are called  **quantization parameters**. The `storage_min` and `storage_max`, being optional, have default values which should satisfy the associated constraints
 as laid out in this document.
```

Please take a look.

docs/spec.md
sngyhan: Hi. Maybe in future we can extend QuantizationScale to accept a type for multiplier/shift to support the case with the defined conversion format between the two.

sngyhan (commented):
LGTM in overall.

docs/spec.md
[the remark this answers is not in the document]
dominicsymes: Thanks @sdasgup3 and @sngyhan  for your reply.

I think the first point is resolved by the comment from @loganchien who points out that zero-point range must be restricted to storage range so the value 0 can be represented. From this I understand that zero-point will be limited to storage range?

On the second point, I’d like to expand a bit more the problems of having zero points on larger data types (greater than 8 bit). For example, an si16 data type with an si16 zero point can give data range of -65535 to +65535. Multiplying two such values will overflow a 32-bit result type. This then has an implementation cost. Since zero_point seems most effective on small data types which have limited range and precision, I’d suggest not to allow on larger data types unless there is a clear use case. I think also in TFLite the zero-point value for integer 16-bit quantized type is set to zero in usage.

docs/spec.md
dominicsymes: Hi @burmako. @sngyhan

Thanks for your comments. Having a quantization type with integer multiplier and shift does get part of the way there provided (as @eric-k256  mentions) that there is a defined conversion between the two formats.
For implementations to agree it also needs to be defined how to convert these into actual scales to be applied.
In an element-wise multiply operation the scale applied after the multiply is the ratio of the product of the input scales to the output scale. Similarly, in the element-wise add in this PR it needs be defined how the integer scales apply as input and output scales.
An advantage of separating out these scale changes is that they can then be handled in a standard way rather than being defined as part of each operation.

docs/spec.md
sdasgup3: Incorporated the proposal in the specification. Feel free to review.

docs/spec.md
sdasgup3: Thanks @loganchien!  As per your feedback proposed edits as mentioned in https://github.com/openxla/stablehlo/pull/1352#discussion_r1157559676

docs/spec.md
```diff
@@ -507,23 +565,35 @@ Performs element-wise addition of two tensors `lhs` and `rhs` and produces a
...
 
 #### Constraints
 
-* (C1) `lhs`, `rhs` and `result` have the same type.
+* (C1) `shape(lhs) = shape(rhs) = shape(result)`.
+* If the operation doesn't use quantized types:
+  * (C2) `element_type(lhs) = element_type(rhs) = element_type(result)`.
+* If the operation uses quantized types:
+  * (C3) `element_type(lhs) = element_type(rhs) = element_type(result)`,
+    except for scales and zero points which may differ.
```
loganchien, line 594: Instead of using `some constraints except ...` explicitly list the actual constraint.

docs/spec.md
loganchien: Instead of:

> `storage_min` and `storage_max` have the following values:

Suggestion:

> If `storage_min` and `storage_max` are unspecified, the default values are defined as:

docs/spec.md
```diff
@@ -507,23 +565,35 @@ Performs element-wise addition of two tensors `lhs` and `rhs` and produces a
...
+  * (C4) `abs(storage_min(lhs)) != abs(storage_max(lhs))`.
```
loganchien, line 595: Why do we need this constraint?

I thought `!quant.uniform<i8<-127,127>:f32, scale:zp>` has been used by some models.

docs/spec.md
burmako: We tried this too, but QuantizedType has a lot of components, and that expanded into a lot of constraints.

docs/spec.md
```diff
@@ -173,10 +190,51 @@ values of type `tensor<T>`).
...
+  `f = (i - zero_point) * scale`, where `scale` and `zero_point` are called
+  **quantization parameters**. The `storage_min` and `storage_max`, being
+  optional, have default values which should satisfy the associated constraints
+  as laid out in this document.
+
+Quantization can be **per-tensor**, meaning, having one `scale` and/or
+`zero_point` for the entire tensor or can be **per-axis**, meaning, having
+multiple `scales` and/or `zero_points`, one pair per slice of a particular
+dimension `quantized_dimension`. More formally, in a tensor `t` of per-axis
+quantized element type, there are `dim(t, quantized_dimension)` slices of the
```
burmako, line 206: Typo: quantization_dimension. (Here and below).

docs/spec.md
burmako: Thinking about this some more, do we have to explicitly specify the default values for storage_min and storage_max? We can use the "is empty" predicate that we used for quantization_dimension below to account for the case when they are missing.

Also, are storage_min and storage_max fully customizable by the user or not? I.e. can they take just one of the 1-2 support values and this is it, or they can be arbitrary, as long as C7 is satisfied?

burmako (changes requested):

docs/spec.md
sdasgup3: will be reflected in the next commit.

docs/spec.md
sdasgup3: The constraint, specifically for AddOp,  is inspired from the TFLite quantization spec for AddOp [ref](https://www.tensorflow.org/lite/performance/quantization_spec#int8_quantized_operator_specifications).

Please let us know more on the use-cases.

docs/spec.md
loganchien: Nvm.

docs/spec.md
loganchien: May you quote the line you are referring to as well?  I am not sure which line translates to `abs(storage_min(lhs)) != abs(storage_max(lhs))`.

IIUC, the asymmetric vs. symmetric comparison in the link is only on the "ability to use whatever min/max convenient" vs. "keep both end symmetric".  Asymmetric doesn't translate to "input can't be symmetric".

docs/spec.md
sdasgup3: Thanks for the suggestion!

>> Also, are storage_min and storage_max fully customization by the user or not? I.e. can they take just one of the 1-2 support values and this is it, or they can be arbitrary, as long as C7 is satisfied?

Per https://github.com/llvm/llvm-project/blob/215f36f7917d87f81eeeee0de8ec55d11d83a164/mlir/lib/Dialect/Quant/IR/QuantTypes.cpp#L54, they can be somewhat arbitrary as long as the the following constraints are satisfied.

```
*  If is_signed(storage_type) and d = num_bits(storage_type):
      * storage_min >= -2^(d-1), storage_max <= 2^(d-1)-1 OR
* If is_unsigned(storage_type):
      * storage_min >= 0, storage_max <= 2^d - 1.
* `storage_max - storage_min > 0`.
``` 
 
>> We can use the "is empty" predicate that we used for quantization_dimension below to account for the case when they are missing.

I agree with that.

We can specify the semantics, of missing `storage_min` and `storage_max`, using the predicate as follows ([ref](https://github.com/llvm/llvm-project/blob/215f36f7917d87f81eeeee0de8ec55d11d83a164/mlir/lib/Dialect/Quant/IR/TypeParser.cpp#L67))

```
* If `storage_min` is empty, then `storage_min = is_signed(storage_type) ? -2^(d-1) : 0`.
* If `storage_max` is empty, then `storage_max = is_signed(storage_type) ? 2^(d-1) : 2^d - 1`.
  where d = num_bits(storage_type)
```

>> do we have to explicitly specify the default values for storage_min and storage_max? 

IMO, we do not have to specify the above default values explicitly, but add them as constraints. With that, as you can imagine, we can have multiple constraints base on the predicates "is_empty" and "is_signed".
Else, we can keep the constraints separate from the specification of default values. As you can see, the constraints will be equally applicable when min/max is empty or not.

docs/spec.md
loganchien: IMO, the quoted line only says that they choose `[-128, 127]` as storage type min/max instead of imposing `abs(storage_min(lhs)) != abs(storage_max(lhs))`.

sdasgup3: The spec has some general constraints for the components `storage_min` and `storage_max` of quantized element type like
```
* (C4) `storage_min` and `storage_max` have the following values:
  * If `is_signed(storage_type)` and `d = num_bits(storage_type)`:
    * `storage_min = -2^(d-1)`, `storage_max = 2^(d-1)-1` OR
    * `storage_min = -2^(d-1)+1`, `storage_max = 2^(d-1)-1`.
  * If `is_unsigned(storage_type)`:
    * `storage_min = 0`, `storage_max = 2^d - 1`.
```

and some AddOp specific constraints (under section `add`)
```
  * (C3) element_type(lhs) = element_type(rhs) = element_type(result), except for scales and zero points which may differ.
  * (C4) `abs(storage_min(lhs)) != abs(storage_max(lhs))`.
```

Together it suggests that the min/max storage type for both the `lhs` and `rhs` for AddOp is -128/127.

With the new formulation proposed in https://github.com/openxla/stablehlo/pull/1352#discussion_r1157942416
AddOp should specify specific values -128/127 for the `storage_min`/`storage_max`.

loganchien: The latest revision LGTM (w.r.t. this comment).  This can be resolved.

docs/spec.md
```diff
@@ -507,23 +570,35 @@ Performs element-wise addition of two tensors `lhs` and `rhs` and produces a
...
 #### Constraints
 
-* (C1) `lhs`, `rhs` and `result` have the same type.
+* (C1) `shape(lhs) = shape(rhs) = shape(result)`.
+* If the operation doesn't use quantized types:
+  * (C2) `element_type(lhs) = element_type(rhs) = element_type(result)`.
+* If the operation uses quantized types:
+  * (C3) `element_type(lhs) = element_type(rhs) = element_type(result)`,
+    except for scales and zero points which may differ.
+  * (C4) `storage_min = -128` and `storage_max = 127`.
```
burmako, line 600: This makes sense from the perspective of int8-based TFLite quantization spec, but how will this generalize to different num_bits values? E.g. to int4?

docs/spec.md
```diff
@@ -173,10 +190,56 @@ values of type `tensor<T>`).
...
+  * `storage_min = is_signed(storage_type) ? -2^(d-1) : 0`.
+* (C5) If `storage_max` is empty:
+  * `storage_max = is_signed(storage_type) ? 2^(d-1) : 2^d - 1`.
+* (C6) If `storage_min` is not empty:
+  * If `is_signed(storage_type)`, `storage_min >= -2^(d-1)`.
+  * If `is_unsigned(storage_type)`, `storage_min >= 0`.
+* (C7) If `storage_max` is not empty:
+  * If `is_signed(storage_type)`, `storage_max <= 2^(d-1)-1`
+  * If `is_unsigned(storage_type)`, `storage_max <= 2^d - 1`.
+* (C8) `storage_max - storage_min > 0`.
```
burmako, line 226: This is subsumed by C11, right?

docs/spec.md
```diff
@@ -173,10 +190,56 @@ values of type `tensor<T>`).
...
+
+Quantized types satisfy the following constraints:
+
+* (C1) `num_bits(storage_type) < num_bits(expressed_type)`.
+* (C2) `type(storage_min) = storage_type`.
+* (C3) `type(storage_max) = storage_type`.
+* (C4) If `storage_min` is empty and `d = num_bits(storage_type)`:
+  * `storage_min = is_signed(storage_type) ? -2^(d-1) : 0`.
+* (C5) If `storage_max` is empty:
+  * `storage_max = is_signed(storage_type) ? 2^(d-1) : 2^d - 1`.
```
burmako, line 219: I don't think we have a precedent for constraints mutating the components of ops, attributes or types, and I'm not sure that it has intuitive semantics. For example, do these mutations apply before or after other constraints? Only those below? Or all of them?

I would propose that we model this in one of the two ways:
  1) On lines 198-200, say "The `storage_min` and `storage_max` are optional in the grammar, but have default values of `min_value(storage_type)` and `max_value(storage_type)` respectively" and then treat these values as non-optional everywhere, i.e. drop "if is empty".
  2) Treat `storage_min` and `storage_max` as optional everywhere, i.e. check if is empty on each access.

I think that option 1 is preferable, because option 2 is going to be really annoying.

docs/spec.md
```diff
@@ -173,10 +190,56 @@ values of type `tensor<T>`).
...
+* (C6) If `storage_min` is not empty:
+  * If `is_signed(storage_type)`, `storage_min >= -2^(d-1)`.
+  * If `is_unsigned(storage_type)`, `storage_min >= 0`.
```
burmako, line 222: Can we abstract this away by using `min_value(type)` and `max_value(type)` metafunctions?

burmako (changes requested):

docs/spec.md
burmako: Left a few follow-up comments on the updated text. Let's resolve this thread and proceed in individual new threads.

docs/spec.md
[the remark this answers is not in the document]
loganchien: There are some implementations with efficient different-scale quantized add as well.  Imposing same-scale requirement and adding extra rescale ops would actually mean slower code in those implementations.  Also, such rescale op either requires a larger immediate type or risk losing some LSBs for the operand with smaller scale (i.e. rounding happens twice).

nutsiepully (approved):
LGTM, thanks.

docs/spec.md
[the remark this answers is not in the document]
nutsiepully: It would be preferable if the semantics of the computation here are compatible with TFLite, since that is widely used across Mobile hardware. The existing floating-point type allows for the compatibility. Breaking out the rescaling limits optimizations.

Also, some of our implementations indeed use int based computation for scale as an implementation detail to avoid Float computations. But I don't think it needs to show up in the spec.

docs/spec.md
sdasgup3: That was an overlook. PLAT at the updated spec.

docs/spec.md
sdasgup3: Maybe I am missing something.
I am not sure if we can subsume `storage_max - storage_min > 0` using the followings
```
* (C4) `storage_min >= min_value(storage_type)`.
* (C5) `storage_max <= max_value(storage_type)`.
* (C9) For all `i`, `storage_min <= zero_points[i] <= storage_max`.
```

(I have updated the constraints with newly suggested macros max_value etc.)

loganchien (approved):

docs/spec.md
```diff
@@ -507,23 +564,36 @@ Performs element-wise addition of two tensors `lhs` and `rhs` and produces a
...
+  * `rounded_result = round_nearest_even(float_result / scale(result))`.
+  * `result = clamp(storage_min(result), rounded_result + zero_point(result), storage_max(result))`.
 
 #### Inputs
 
 | Label | Name  | Type   | Constraints |
 |-------|-------|--------|-------------|
-| (I1)  | `lhs` | tensor | (C1)        |
-| (I2)  | `rhs` | tensor | (C1)        |
+| (I1)  | `lhs` | tensor | (C1)-(C5)   |
```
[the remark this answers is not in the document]

docs/spec.md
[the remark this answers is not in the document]
sdasgup3: >> Since zero_point seems most effective on small data types which have limited range and precision, I’d suggest not to allow on larger data types unless there is a clear use case.

@dominicsymes I can see you point on restricting the type of zero point to 8-bits. Thanks for explaining!

@loganchien @sngyhan Do you mind providing your thoughts here?  

For unsigned quantized types (> 8-bits), the zero point  cannot be zero and need to match the storage type. Can this be a use-case?

docs/spec.md
```diff
@@ -140,11 +140,28 @@ which may allow us to remove tuple types from StableHLO
 
 ```ebnf
 ElementType ::= BooleanType | IntegerType | FloatType | ComplexType
+              | QuantizedType
```
burmako, line 143: Having thought about this some more, I think it would be best to express quantization in types via `TensorType | QuantizedType` rather than `ElementType ::= ... | QuantizedType`. See https://github.com/sdasgup3/stablehlo/pull/5 for details.

burmako (changes requested):

docs/spec.md
sdasgup3: Thanks for the proposal! Incorporated.

burmako (changes requested):
LGTM once we figure out the type of zero_point. Should it be 8-bit or lower, or should it be the same as storage_type as currently specified? 

For another open discussion thread at the moment of writing - about an integer-based QuantizedType - this is something that I think we should explore in a follow-up pull request. Based on the review feedback this looks like a promising idea that will help meet requirements from several groups, and I think we should nail down the details in the very near future.

docs/spec.md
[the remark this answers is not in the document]
loganchien: I prefer not adding such constraint at this moment.  IMO, StableHLO is more of the compiler IR that aims to support different targets instead of a least common denominator that everyone must support.  So, for this particular case, I prefer to keep the symmetry of IR design rather than adding a constraint just because some targets can't efficiently implement it.

I understand that spec has bidirectional effects, i.e. the implementors want to know what are the features must be implemented and the users (quantizers) want to know what are the features that can be used.  But, I think we can defer that later into the concept of profiles or compatibility matrix.

For concrete usages, I only find a usage when the signed integer type is emulated with the quantized type.  However, that is only an internal usage and can be implemented differently (thus does not necessary coupling with StableHLO spec).  As of writing, I don't find quantize models with >8bit zero point.  Maybe @sngyhan will have a better idea because quantizer is not my full-day work.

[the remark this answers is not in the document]
loganchien: IMO, an authoring tool/framework doesn't have to use a feature just because the spec can support/express such feature.  The authoring tool (producer) can still limit themselves to use `zp in [-128, 127]` for `i8`/`i16` if the narrower zero point range can accommodate their need.  Interpreters/validators may have a valid case because they are on the implementor side (consumer).  But, like the other parts of the spec, I don't think an interpreter can efficiently cover all features.  And, we will still have per-op constraints which can be used to narrow down the scope as needed (all multiplicative ops may need a narrower range, which I agree).  For visualization, I don't think it will add constraints to the value range.

My point is not to enact a constraint too early.  Maybe someday we will evolve to a stage with real need in real model.

My proposed wording is to limit zero point to storage type, i.e. `i8` gets `[-128, 127]`, `u8` gets `[0, 255]`, `i16` gets `[-32768, 32767]`, etc.  Thus, only `i64` storage type can get `i64` zero point.

But I am fine with adding constraints (albeit not preferring).  Actually, the implementation I am working on probably can't support all zero points in all operations either (in case the product overflows as pointed out by @dominicsymes earlier).  My comments is more on the design symmetry or my personal view on aesthetic.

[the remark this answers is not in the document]
sdasgup3: Thanks everybody for the comments! 
@sngyhan Can you please add your opinion as well.

docs/spec.md
```diff
@@ -116,6 +116,68 @@ types, for example, to include layouts
...
+`max_value(storage_type)` respectively. Quantized element types have the
+following constraints:
+
+* (C1) `num_bits(storage_type) < num_bits(expressed_type)`.
+* (C2) `type(storage_min) = storage_type`.
+* (C3) `type(storage_max) = storage_type`.
+* (C4) `min_value(storage_type) <= storage_min < storage_max <= max_value(storage_type)`.
+* (C5) For all `i`, `type(scales[i]) = expressed_type`.
+* (C6) For all `i`, `scales[i] > 0`.
+* (C7) For all `i`, `storage_min <= zero_points[i] <= storage_max`.
```
Maratyszcza, line 155: In TFLite `zero_point` is per-tensor, even when per-channel quantization is used ("per-channel" affects only scales, and BTW per-channel quantization is supported only for signed quantization schema, where zero point must be 0).

docs/spec.md
```diff
@@ -116,6 +116,68 @@ types, for example, to include layouts
...
+* (C8) For all `i`, `type(zero_points[i]) = storage_type`.
+* (C9) `size(scales) = size(zero_points)`.
```
Maratyszcza, line 157: Same, is there a good reason to have per-channel zero point?

docs/spec.md
sdasgup3: Hi @Maratyszcza 
Thanks for the comments! 

We already have (1). I have added (2)
```
* (C6) For all `i`, `scales[i] > 0`.
* (C7) For all `i`, `is_finite(scales[i]) = true`.
```

>> About (3): .... assures compatibility with systems which don't support subnormal numbers where they will treated as 0.

Maybe I did not get this. 
Isn't the constraint will be restrictive to the systems supporting sub-normals? 
If this check is relevant when converting a double multiplier (e.g. `S1*S2/S3` in eq (5) in https://arxiv.org/pdf/1712.05877.pdf to a int multiplier/shift), then does it matter if we add constraints to the individual scales? That is, isn't the double multiplier can still be sub-normal is the individual scales are not. 

On a side note: For (3), Can we explore this constraint when we deal with the "integer based quantized type" in a separate PR (as suggested in https://github.com/openxla/stablehlo/pull/1352#pullrequestreview-1378179457) ?

docs/spec.md
```diff
@@ -116,6 +116,68 @@ types, for example, to include layouts
...
+
+**Quantized element types** represent integer values of a **storage type** in
+the range from `storage_min` to `storage_max` (inclusive) that correspond to
+floating-point values of an **expressed type**. For a given integer value `i`,
+the corresponding floating-point value `f` can be computed as
+`f = (i - zero_point) * scale`, where `scale` and `zero_point` are called
+**quantization parameters**. The `storage_min` and `storage_max` are optional
+in the grammar, but have default values of `min_value(storage_type)` and
+`max_value(storage_type)` respectively. Quantized element types have the
+following constraints:
```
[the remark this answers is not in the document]
sdasgup3: >> What is the semantics of storage_min and storage_max?

The semantics is to make sure that the quantized values are within this range.  This semantics is enforced by the individual ops. 

>> Are operators which produce quantized values supposed to clamp them to [storage_min, storage_max]?

Yes.

>> Are StableHLO consumers supposed to validate that quantized tensors that are inputs to the model don't have values outside of [storage_min, storage_max]?

The spec in its current form, uses the individual ops to apply the semantics of the `[storage_min, storage_max]`, which is to clamp the quantized values. There is no constraint in the input data and IMO the consumers can still be compliant to the spec even if they do not have the validation.

docs/spec.md
sdasgup3: Thanks @Maratyszcza for the valuable information! 
The constraint we have is straight from text-book definition of per-axis zero points and hence is more general. But I am open to improvement.  IMO, this is also somewhat related to our prior discussion about the trade-off between generality of spec vs practical values of zero points.  [here](https://github.com/openxla/stablehlo/pull/1352#discussion_r1147883143). 


@sngyhan Do you mind commenting on this in the light of the following:

For dynamic range quantization, for say convolution op, when the activation is quantized on the fly, I can see from [here](https://github.com/tensorflow/tensorflow/blob/3a838bf19754eccfc7a9d324ba0e874735a77383/tensorflow/core/kernels/uniform_quant_ops/uniform_quantized_convolution_ops.cc#L717) and [here](https://github.com/tensorflow/tensorflow/blob/3a838bf19754eccfc7a9d324ba0e874735a77383/tensorflow/core/kernels/uniform_quant_ops/math_utils.h#L165), that the activation quantization is per-channel (on batch dimension) but is asymmetric with zero point  != 0. Moreover, IIUC, in most of the kernel implementations in https://github.com/tensorflow/tensorflow/tree/3a838bf19754eccfc7a9d324ba0e874735a77383/tensorflow/core/kernels/uniform_quant_ops are using per-channel zero-points and they does not seem to have special value constraint on them.

Can you please let me know if  I missing something.

docs/spec.md
```diff
@@ -81,7 +81,7 @@ completely numeric to simplify generation of StableHLO programs.
 
 ```ebnf
 Type         ::= ValueType | NonValueType
-ValueType    ::= TensorType | TokenType | TupleType
+ValueType    ::= TensorType | QuantizedTensorType | TokenType | TupleType
```
abattery, line 84: Should the QuantizedTensorType be a special case of TensorType instead of ValueType?

For example, quantized tensor constants can not be declared since it is a part of TensorType in the spec.

docs/spec.md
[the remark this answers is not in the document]
abattery: For some cases, we also use the [-7, 8] range with eight bit integer storage to mimick four bit quantization or other bit scheme quantization with wider integer storage.

For popular eight bit quantization cases, as you said, the typical ranges, [-128, 128] [-127, 128], are popular. However, IMO, for various reasons (e.g, special quantization treatments, research purpose) it might be better not to limit allowed ranges with one or two ranges only in the language specification.

docs/spec.md
abattery: +1 for opening the possibility of both scaling mechanisms.

docs/spec.md
```diff
@@ -116,6 +116,69 @@ types, for example, to include layouts
...
+                  ',' QuantizationParameters '>'
+QuantizationStorageType ::= IntegerType
+QuantizationStorageMin ::= IntegerConstant
+QuantizationStorageMax ::= IntegerConstant
+QuantizationExpressedType ::= FloatType
+QuantizationDimension ::= IntegerConstant
+QuantizationParameters ::= QuantizationParameter
+                         | '{' QuantizationParameter {',' QuantizationParameter} '}'
+QuantizationParameter ::= QuantizationScale ':' QuantizationZeroPoint
+QuantizationScale ::= FloatConstant
```
abattery, line 135: Can we open the possibility to have unknown values for scale and zero point for quantized training, which requires the training programs to provide scale and zero point values on the fly?

docs/spec.md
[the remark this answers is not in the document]
abattery: I am also not preferring specific constraints based on we've observed from TFLite product. IMO, it potentially might limit the adoptions of new quantization schemes/kernels in the future, e.g., experimental quantization schemes.

Numerically, to cover all the possible cases, "original value range ⊂ zp range" might be the right constraint.

docs/spec.md
burmako: Incidentally, this is what the initial version of the pull request did, but we decided to split TensorType and QuantizedTensorType recently, for the reasons described in https://github.com/sdasgup3/stablehlo/pull/5#discussion_r1161141397.

docs/spec.md
burmako: I think that's an interesting idea, and #1149 is highlighting this as an open issue as well. 

However, given that we don't have a precedent for this functionality (quant.uniform doesn't support this yet, and therefore StableHLO dialect doesn't support this either), I think that this functionality would need to be introduced in a dedicated RFC once we define the baseline StableHLO spec.

This is similar to how we've forked StableHLO from MHLO and specced baseline HLO semantics, and now are starting to make changes to it along the lines of #1143, #1308, #1342, etc.

burmako (approved):
Thank you, everyone, for the discussion! Looks like we have alignment on quite a few topics already, so I think it would be beneficial to merge this pull request and continue working out the rest in follow up PRs.

More specifically, let's keep the definition of QuantizedTensorType, remove the definition of quantized AddOp for now and open tickets to follow up on the topics where we haven't reached consensus yet - #1404, #1405, #1406 and #1407.

From here, we'll be able to explore multiple directions simultaneously, including speccing integer-based QuantizedType, as well as discussing the semantics of quantized elementwise ops and convolutions.

docs/spec.md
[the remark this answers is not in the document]
burmako: Thank you for the feedback! It looks like there's broad consensus among reviewers that exploring an integer-based QuantizedTensorType makes sense, so we opened #1404 and will follow up on it shortly in a separate PR.

docs/spec.md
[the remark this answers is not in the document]
burmako: This is a very lively discussion! Given the variety of proposals, I think that this conversation would benefit from a video meeting. I'll work with @theadactyl to set up a community meeting about quantization, and this will be one of the topics. We've opened #1405 to keep track.

docs/spec.md
burmako: As discussed above, let's move the work on integer-based QuantizedTensorType to a separate PR, given that there's broad interest in exploring this. We opened #1404 to track this.

docs/spec.md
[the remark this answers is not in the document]
burmako: This is a good point! Let's discuss more in #1406. In the meanwhile, we removed the spec for quantized AddOp to sidestep having an opinion on this in this specific PR.

docs/spec.md
burmako: Similarly to other unresolved conversation, let's follow up on this in a dedicated ticket: #1405.

docs/spec.md
burmako: Let's discuss above: https://github.com/openxla/stablehlo/pull/1352#discussion_r1163130568.

PR #1353, opened 2023-03-24
Add Python API version number

GleasonK:
Adds an arbitrary version number which gets incremented on all changes to public APIs.

StableHLO API's do not offer any stability. This change may be useful for filtering tests that require bug fixes, etc. The goal is to help downstream StableHLO integration run more smoothly as the API's/features evolve.

### DIFF
--- a/stablehlo/dialect/Version.h
+++ b/stablehlo/dialect/Version.h
@@ -43,10 +43,6 @@ class Version {
   /// Return a Version representing the minimum supported VHLO dialect version.
   static Version getMinimumVersion() { return Version(0, 9, 0); }
 
-  /// Return the current version for StableHLO APIs.
-  /// Increments on all C++/C/Python API changes
-  static int64_t getApiVersion() { return 1; }
-
   /// Construct Version from major, minor, patch integers.
   Version(int64_t major, int64_t minor, int64_t patch)
       : majorMinorPatch({major, minor, patch}) {}
--- a/stablehlo/dialect/Version.h
+++ b/stablehlo/dialect/Version.h
@@ -79,6 +79,14 @@ mlir::Diagnostic& operator<<(mlir::Diagnostic& diag, const Version& version);
 llvm::raw_ostream& operator<<(llvm::raw_ostream& os, const Version& version);
 
 }  // namespace vhlo
+
+namespace stablehlo {
+
+/// Return the current version for StableHLO APIs.
+/// Increments on all C++/C/Python API changes
+inline int64_t getApiVersion() { return 1; }
+
+}  // namespace stablehlo
 }  // namespace mlir
 
 #endif  // STABLEHLO_DIALECT_VERSION_H
--- a/stablehlo/dialect/Version.h
+++ b/stablehlo/dialect/Version.h
@@ -43,6 +43,10 @@ class Version {
   /// Return a Version representing the minimum supported VHLO dialect version.
   static Version getMinimumVersion() { return Version(0, 9, 0); }
 
+  /// Return the current version for StableHLO APIs.
+  /// Increments on all C++/C/Python API changes
+  static int64_t getApiVersion() { return 1; }
+
   /// Construct Version from major, minor, patch integers.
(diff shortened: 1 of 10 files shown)


stablehlo/integrations/python/mlir/dialects/stablehlo.py
```diff
@@ -17,3 +17,7 @@
 # pylint: disable=wildcard-import,relative-beyond-top-level,g-import-not-at-top
 from ._stablehlo_ops_gen import *
 from .._mlir_libs._stablehlo import *
+
+
+# Version number for StableHLO Python APIs. Increment on all Python API changes.
+api_version = 1
```
burmako, line 23: As discussed offline, let's turn this into a StablehloModule function that returns a counter which is also incremented on all C++ API changes.

burmako (changes requested):

stablehlo/dialect/Version.h
```diff
@@ -43,6 +43,10 @@ class Version {
   /// Return a Version representing the minimum supported VHLO dialect version.
   static Version getMinimumVersion() { return Version(0, 9, 0); }
 
+  /// Return the current version for StableHLO APIs.
+  /// Increments on all C++/C/Python API changes
+  static int64_t getApiVersion() { return 1; }
```
burmako, line 48: How about moving this to a free function at the bottom of this file, under the `mlir::stablehlo` namespace? It feels a bit weird to have StableHLO API version under `mlir::vhlo`.

stablehlo/integrations/c/StablehloDialect.h
```diff
@@ -22,6 +22,9 @@ extern "C" {
 
 MLIR_DECLARE_CAPI_DIALECT_REGISTRATION(Stablehlo, stablehlo);
 
+// Get the current version of StableHLO API's. Incremented with all changes.
+MLIR_CAPI_EXPORTED int getStablehloApiVersion();
```
burmako, line 26: Do we need to expose this as part of our C API? Fwiw, we don't do this for `getCurrentVersion`.

burmako (changes requested):

stablehlo/integrations/c/StablehloDialect.h
GleasonK: This would remove the need for users of StablehloModule.cpp from taking another dependency on `:version`. We got around this for those other API's by placing them in `Serialization.h`, which may not be the proper location for them.

We could move all "get*version" functions to `Version.h` and require the additional dependency from users of StablehloModule. Perhaps that's a more proper design.

stablehlo/dialect/Version.h
GleasonK: Given our recent discussions on VHLO as an implementation detail, would it make sense to move this entire file into `stablehlo` namespace. Then VHLO would be querying the current version of StableHLO.

burmako (approved):

stablehlo/dialect/Version.h
GleasonK: I guess the question is "What is the version source of truth" - Is VHLO targeting a StableHLO version, or is StableHLO's version the VHLO version.

If the former, then probably `stablehlo::Version` makes sense. The latter, `vhlo::Version`. This can be done later. I just find it odd that we expose `stablehlo.get_current_version` via `Serialization.h`, which requests the current VHLO version.

stablehlo/dialect/Version.h
```diff
@@ -75,6 +75,14 @@ mlir::Diagnostic& operator<<(mlir::Diagnostic& diag, const Version& version);
 llvm::raw_ostream& operator<<(llvm::raw_ostream& os, const Version& version);
 
 }  // namespace vhlo
+
+namespace stablehlo {
+
+/// Return the current version for StableHLO APIs.
+/// Increments on all C++/C/Python API changes
```
burmako, line 82: As just discussed, we are planning to limit this to just changes to APIs under `integrations/` for now because this is the original use case. Can expand in the future.

PR #1363, opened 2023-03-31
StableHLO doc review

pcoet:
I reviewed the StableHLO docs and did some light copy editing. I had to timebox this work, and I was only able to skim the spec. Overall, content is well written, thoughtful, has a friendly tone -- good stuff!


pcoet:
@burmako

docs/spec.md
```diff
@@ -10,12 +10,14 @@ interoperability between various ML frameworks (such as TensorFlow, JAX and
...
-section provides semantics for all these ops executing together within
-a program.
+This specification contains three major sections. First, the
+[Programs](#programs) section describes the structure of StableHLO programs
+which consist of StableHLO functions which themselves consist of StableHLO ops.
+Within that structure, the [Ops](#ops) section specifies the semantics of
+individual ops. Finally, the [Execution](#execution) section provides semantics
+for all these ops executing together within a program.
+
+<a id="programs"></a>
```
sdasgup3, line 20: Very nice touch!

sdasgup3:
LGTM for reference.md, reference_checklist.md, spec.md, spec_checklist.md, type_inference.md

docs/spec.md
pcoet: Thanks for the review!

docs/compatibility.md
```diff
@@ -48,6 +49,8 @@ request, we are testing that deserialization of the compatibility suite at HEAD
 produces StableHLO programs which are syntactically identical to the original
 StableHLO programs that were serialized by the latest release of libStablehlo.
 
+<a id="future-work"></a>
```
burmako, line 52: I don't think we need explicit anchor tags, given that GitHub autogenerates them anyway? For example, this already works at HEAD: https://github.com/openxla/stablehlo/blob/main/docs/compatibility.md#future-work. (Same below).

docs/roadmap.md
```diff
@@ -41,57 +41,62 @@ and tied these workstreams to the milestones. A limited number of tickets
 (less than 10%) are not assigned to any specific workstream and aren't part of
 any specific milestone.
 
-(P0) [Compatibility implementation](https://github.com/orgs/openxla/projects/4)
+(P0) The
+[Compatibility implementation](https://github.com/orgs/openxla/projects/4)
```
burmako, line 45: Should we capitalize both words, i.e. "Compatibility Implementation"? Same question for other workstream names.

burmako (changes requested):

docs/compatibility.md
pcoet: This is true. There are pros and cons for the explicit anchor tags in a doc like this, they do add clutter. Removed.

docs/roadmap.md
pcoet: Maybe so. Workstream links updated.

pcoet:
@burmako All the proposed changes should be made now. Thanks!

PR #1373, opened 2023-04-05
Add interpreter for MapOp

### LINKED ISSUE #1106: Add  interpreter for MapOp

ghpvnist:

```
(I1) inputs: variadic number of tensors.
(I2) dimensions: 1-dimensional tensor constant of type `si64`.
(I3) computation: function.
(C1) All `inputs` and `result` have the same shape.
(C2) size(`inputs`) = N >= 1.
(C3) `dimensions = [0, ..., R-1]`, where `R` = rank(`inputs[0]`).
(C4) `computation` has type `(tensor<E0>, ..., tensor<EN-1>) -> tensor<E'>`
where `Ek` = element_type(`inputs[k]`) and `E'` = element_type(`result`).
```

```
I1: a) inputs is not a variadic tensor. (Covered by ODS).
I2: a) dimensions is not a 1-dimensional tensor.
    b) element_type(dimensions) != si64. (Covered by ODS).
I3: a) computation is not a function. (Covered by ODS).
C1: a) Not all `inputs` have the same shape. (Covered by ODS).
    b) `inputs` and `result` do not have the same shape. (Covered by ODS).
C2: size(`inputs`) < 1. (Covered by ODS).
C3: a) `dimensions != [0, ..., R-1]`, where `R` = rank(`inputs[0]`).
C4: a) `computation` does not have type `(tensor<E0>, ..., tensor<EN-1>) -> tensor<E'>`
where `Ek` = element_type(`inputs[k]`) and `E'` = element_type(`result`).
```

```
I2a: dimensions is not a 1-dimensional tensor.
C3a: `dimensions != [0, ..., R-1]`, where `R` = rank(`inputs[0]`).
C4a: `computation` does not have type `(tensor<E0>, ..., tensor<EN-1>) -> tensor<E'>`
where `Ek` = element_type(`inputs[k]`) and `E'` = element_type(`result`).
```

closes #1106

### DIFF
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -2375,8 +2375,7 @@ def StableHLO_GetDimensionSizeOp: StableHLO_Op<"get_dimension_size",
 }
 
 def StableHLO_MapOp: StableHLO_ShapedInterfaceOp<"map",
-      [RecursiveMemoryEffects, SameVariadicOperandSize /*map_c2*/,
-       SameOperandsAndResultShape /*map_c1*/,
+      [RecursiveMemoryEffects, SameOperandsAndResultShape /*map_c1, map_c2*/,
        SingleBlockImplicitTerminator<"ReturnOp">, InferTensorTypeWithReify]> {
   let summary = "Map operation";
   let description = [{
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -2375,7 +2375,8 @@ def StableHLO_GetDimensionSizeOp: StableHLO_Op<"get_dimension_size",
 }
 
 def StableHLO_MapOp: StableHLO_ShapedInterfaceOp<"map",
-      [RecursiveMemoryEffects, SameOperandsAndResultShape,
+      [RecursiveMemoryEffects, SameVariadicOperandSize /*map_c2*/,
+       SameOperandsAndResultShape /*map_c1*/,
        SingleBlockImplicitTerminator<"ReturnOp">, InferTensorTypeWithReify]> {
   let summary = "Map operation";
   let description = [{
@@ -2388,19 +2389,19 @@ def StableHLO_MapOp: StableHLO_ShapedInterfaceOp<"map",
     Example:
     ```mlir
     %result = "stablehlo.map"(%input0, %input1) ({
-      ^bb0(%arg0: tensor<i32>, %arg1: tensor<i32>):
-        %0 = stablehlo.multiply %arg0, %arg1 : tensor<i32>
-        stablehlo.return %0 : tensor<i32>
+      ^bb0(%arg0: tensor<i64>, %arg1: tensor<i64>):
+        %0 = stablehlo.multiply %arg0, %arg1 : tensor<i64>
+        stablehlo.return %0 : tensor<i64>
     }) {
       dimensions = dense<[0, 1]> : tensor<2xi64>
-    } : (tensor<2x2xi32>, tensor<2x2xi32>) -> tensor<2x2xi32>
+    } : (tensor<2x2xi64>, tensor<2x2xi64>) -> tensor<2x2xi64>
     ```
   }];
   let arguments = (ins
-    Variadic<HLO_Tensor>:$inputs,
-    I64ElementsAttr:$dimensions
+    Variadic<HLO_Tensor>:$inputs /*map_i1*/,
+    I64ElementsAttr:$dimensions /*map_i2*/
   );
-  let regions = (region SizedRegion<1>:$computation);
+  let regions = (region SizedRegion<1>:$computation /*map_i3*/);
   let results = (outs HLO_Tensor);
 }
 
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -2353,6 +2353,12 @@ LogicalResult inferMapOp(
     std::optional<Location> location, ValueRange inputs,
     DenseIntElementsAttr dimensions, Region& computation,
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
+  // map_i2
+  if (dimensions.getType().getRank() != 1)
+    return emitOptionalError(location,
+                             "dimensions should be rank 1 but got rank ",
+                             dimensions.getType().getRank());
+
   if (failed(verifyRegionNotEmpty(location, computation))) return failure();
 
   // Checks if the number of `operands` match the arity of the map `computation`
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -2434,6 +2434,7 @@ LogicalResult inferMapOp(
     }
   }
 
+  // map_c4
   if (allInputsUnranked)
     inferredReturnShapes.emplace_back(computationOutputType.getElementType());
   else
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -2361,8 +2361,7 @@ LogicalResult inferMapOp(
 
   if (failed(verifyRegionNotEmpty(location, computation))) return failure();
 
-  // Checks if the number of `operands` match the arity of the map `computation`
-  // region.
+  // map_c4
   auto& computationBlock = computation.front();
   auto computationArgs = computationBlock.getArguments();
   if (inputs.size() != computationArgs.size())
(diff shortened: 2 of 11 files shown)


stablehlo/dialect/TypeInference.cpp
```diff
@@ -2354,6 +2354,12 @@ LogicalResult inferMapOp(
     std::optional<Location> location, ValueRange inputs,
```
sdasgup3, line 2354: Missing labels for constraints like C3a and c4a.

stablehlo/dialect/StablehloOps.td
```diff
@@ -2371,7 +2371,8 @@ def StableHLO_GetDimensionSizeOp: StableHLO_Op<"get_dimension_size",
 }
 
 def StableHLO_MapOp: StableHLO_ShapedInterfaceOp<"map",
-      [RecursiveMemoryEffects, SameOperandsAndResultShape,
+      [RecursiveMemoryEffects, SameVariadicOperandSize /*map_c2*/,
```
sdasgup3, line 2374: No need to add `SameVariadicOperandSize` if we do not have multiple variadic operands?

stablehlo/dialect/TypeInference.cpp
```diff
@@ -2354,6 +2354,12 @@ LogicalResult inferMapOp(
...
     DenseIntElementsAttr dimensions, Region& computation,
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
+  // map_i2
+  if (dimensions.getType().getRank() != 1)
+    return emitOptionalError(location,
+                             "dimensions should be rank 1 but got rank ",
+                             dimensions.getType().getRank());
+
   if (failed(verifyRegionNotEmpty(location, computation))) return failure();
```
sdasgup3, line 2363: Also add a check for c2 similar to what HLO has
https://github.com/tensorflow/tensorflow/blob/4f1dd6d5123f4eb6afc85fac36df09b4a8b49c83/tensorflow/compiler/xla/service/shape_inference.cc#L1161

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -1610,54 +1610,43 @@ func.func @map(%arg0: tensor<4x5xf32>, %arg1: tensor<4x5xf32>) -> tensor<4x5xf32
...
-    "stablehlo.return"(%arg2) : (tensor<f32>) -> ()
-  }) {dimensions = dense<0> : tensor<1xi64>} : (tensor<2xf32>, tensor<2xi32>) -> tensor<2xf32>
-  func.return %0 : tensor<2xf32>
-}
-
-// -----
-
-// CHECK-LABEL: func @map_scalar_operands
-func.func @map_scalar_operands(%arg0: tensor<f32>, %arg1: tensor<f32>) -> tensor<f32> {
+func.func @map_c3(%arg0: tensor<4x5xf32>, %arg1: tensor<4x5xf32>) -> tensor<4x5xf32> {
```
sdasgup3, line 1613: add a test for c2

stablehlo/reference/Ops.cpp
```diff
@@ -541,6 +546,23 @@ Tensor evalLogisticOp(const Tensor &operand, TensorType resultType) {
   return result;
 }
 
+Tensor evalMapOp(ArrayRef<Tensor> inputs, Region &computation, Scope &scope,
+                 TensorType resultType) {
+  Tensor result(resultType);
+  for (auto inputIt = inputs[0].index_begin(); inputIt != inputs[0].index_end();
```
sdasgup3, line 552: IMO this top level iterator is used only to get the result index. In that case, let it be a iterator on result.

stablehlo/reference/Ops.cpp
```diff
@@ -541,6 +546,23 @@ Tensor evalLogisticOp(const Tensor &operand, TensorType resultType) {
...
+       ++inputIt) {
+    SmallVector<Tensor> args;
+    for (size_t i = 0; i < inputs.size(); ++i) {
+      auto tensor =
```
sdasgup3, line 556: No need to extract an tensor element to get the element type of the tensor.
```
Tensor arg(RankedTensorType::get({}, inputs[i].getElementType());
```

Even better: IIRC there is a PR where we talked about introducing a constructor like 
```
 explicit Tensor(TensorType, const Element &);
```

We should use it here as well.

ghpvnist: Thanks for the suggestion. I also recall discussing this somewhere but haven't find which PR it was (unfortunately GH doesn't have a great search engine).

stablehlo/dialect/StablehloOps.td
ghpvnist: `SameVariadicOperandSize` should be renamed to `VariadicOperandsHaveSameSize`. The singular `Operand` is misleading :)

stablehlo/dialect/TypeInference.cpp
ghpvnist: The check is not needed as this is still caught by the ODS with the `verifyAtLeastNOperands(op, 1)` check.
Reference: https://github.com/llvm/llvm-project/blob/main/mlir/lib/IR/Operation.cpp#L844-L845

stablehlo/tests/ops_stablehlo.mlir
ghpvnist: Since the ODS covers this, a test won't be needed.

sdasgup3 (approved):

stablehlo/dialect/TypeInference.cpp
```diff
@@ -2387,24 +2391,22 @@ LogicalResult inferMapOp(
...
+  // map_c4
   auto computationOutputs = computationBlock.getTerminator()->getOperands();
   if (computationOutputs.size() != 1)
     return emitOptionalError(location,
                              "computation must return single output, but got: ",
                              computationOutputs.size());
 
-  // The output of computation must be scalar and have the same element type
-  // as op result.
+  // map_c4
```
burmako, line 2401: Looks like lines 2438-2442 are also `map_c4`?

stablehlo/tests/infer_stablehlo.mlir
```diff
@@ -366,6 +366,18 @@ func.func @map(%arg0: tensor<4x5xf32>, %arg1: tensor<4x5xf32>) -> tensor<4x5xind
...
+  // expected-error@+1 {{inferred type(s) 'tensor<4x5xi32>' are incompatible with return type(s) of operation 'tensor<4x5xf32>'}}
+  %0 = "stablehlo.map"(%arg0, %arg1) ({
+    ^bb0(%arg2: tensor<f32>, %arg3: tensor<f32>):
+    %1 = stablehlo.constant dense<2> : tensor<i32>
+    "stablehlo.return"(%1) : (tensor<i32>) -> ()
+  }) {dimensions = dense<[0, 1]> : tensor<2xi64>} : (tensor<4x5xf32>, tensor<4x5xf32>) -> tensor<4x5xf32>
+  func.return %0 : tensor<4x5xf32>
+}
+
+// -----
```
burmako, line 379: Let's drop this test, similarly to other PRs that we discussed recently?

stablehlo/reference/Ops.cpp
```diff
@@ -541,6 +546,21 @@ Tensor evalLogisticOp(const Tensor &operand, TensorType resultType) {
...
 
+Tensor evalMapOp(ArrayRef<Tensor> inputs, Region &computation, Scope &scope,
+                 TensorType resultType) {
+  Tensor result(resultType);
+  for (auto resultIt = inputs[0].index_begin();
+       resultIt != inputs[0].index_end(); ++resultIt) {
+    SmallVector<Tensor> args;
+    for (size_t i = 0; i < inputs.size(); ++i)
+      args.push_back(
+          Tensor(RankedTensorType::get({}, inputs[i].get(*resultIt).getType()),
```
burmako, line 557: Can we use `computation.getArgument(i).getType()` instead? Shorter to type and doesn't require creating types, i.e. avoids potential logic errors.

stablehlo/reference/Tensor.cpp
```diff
@@ -93,6 +93,11 @@ Tensor::Tensor(TensorType type)
 Tensor::Tensor(TensorType type, AsmResourceBlob blob)
     : impl_(llvm::makeIntrusiveRefCnt<detail::Buffer>(type, std::move(blob))) {}
 
+Tensor::Tensor(TensorType type, const Element &element)
+    : impl_(llvm::makeIntrusiveRefCnt<detail::Buffer>(type)) {
+  this->set({}, element);
+}
```
burmako, line 99: I see how this constructor is more convenient that first creating a tensor, and then doing a set. 

However, I think that it's currently not pulling its weight. On the one hand, it's used only once, and its convenience is marginal (compresses a 2-liner into a 1-liner). On the other hand, it does introduce a new function and complicate the API of `Tensor`, so the cost is non-trivial.

Let's inline the constructor into the callsite?

burmako (changes requested):
LGTM modulo a few minor comments.

burmako (approved):

Issue #1367, opened 2023-04-02
A generic mechanism for shape refinement for custom calls


gnecula:
### Request description

Currently I am trying to use the `indices_of_shape_operands` attribute for the custom_call op to specify the shapes of the outputs as 1D tensor operands.

There are a couple of problems with this mechanism. First, shape refinement does not work today when the custom call returns a tuple. This could be fixed easily.

The real problem is that this mechanism forces the custom call target to take an extra operand, which likely has no use during the computation. Some of these custom call targets do take operands that are used in the computation and determine the output shape, but for the `indices_of_shape_operands` I am forced to add another operand just for shape refinement. Furthermore, this operand will be transferred to the device, which is a waste.

In some important use cases, I am now forced to change the custom call target in order to take this operand, even though after shape refinement it has no use. Changing the C++ targets is not easy because we need to keep the old code for backwards compatibility. We could write a pass that runs after shape refinement to rewrite dynamic_custom_call_target into custom_call_target, if the output shape is a constant, but this does not work in general because one would have to change the `backend_config` in some cases.

I propose a different, less intrusive mechanism. We can add a RefineShapeOp: `refine_shape(%a, %shape_operand)` that can be used for an operand `%a` with dynamic shapes to refine the shape with a 1d tensor `%shape_operand`. The rank of `%a` must be equal to the length of `%shape_operand`. During shape refinement the `shape_operand` will be a constant, or will have some constant elements, which will refine the shape of `%a`. 

The front-end will add this to the outputs of a custom call, and the shape refinement will have to flow the contraints to the producer of its operand.

This mechanism could be used for other ops that today need to take a shape operand only for the purpose of shape refinement, e.g., DynamicGatherOp (#690).

gnecula:
One simpler solution is to agree that the operands marked by `indices_of_operand_shapes` are not actual arguments, they are removed from the argument list when the C++ call target is called. If the call target needs them, then they can be passed twice. They can be removed when legalizing to MHLO.

burmako:
Started working on this in #1378. Once/if that PR is approved, it will be very easy to extend it to do what George is proposing in the second comment.

PR #1378, opened 2023-04-06
Introduce --stablehlo-canonicalize-dynamism

burmako:
This pull request achieves two goals:

1) Makes specialization of dynamic StableHLO programs completely independent from MHLO. Before this PR, we were depending on StableHLO => MHLO and then MHLO canonicalization to actually specialize away dynamic_foo ops, but not anymore.

2) Provides a platform for addressing #1367. As a follow-up to this PR, we will add another pattern to this pass that transforms CustomCallOp to erase the indices_of_shape_operands attributes as well as the operands called out in indices_of_shape_operands.

The patterns in this PR are quite straightforward, since almost all of the work is done in --stablehlo-refine-shapes.
CanonicalizeRealDynamicSliceOpToDynamicSliceOpPattern is the only exception because it isn't just about "the result has static shape, let's go" or "the attributes are all static, let's go".

Furthermore, this PR extends some existing verifiers and adds some new verifiers for dynamic ops. Since dynamic ops aren't specced yet, it is expected that they have implementation gaps, but in this case, these gaps would've allowed invalid dynamic ops to get transformed into valid static ops, so I fixed them right away.


stablehlo/dialect/TypeInference.cpp
```diff
@@ -3471,13 +3499,22 @@ LogicalResult verifyDynamicPadOp(std::optional<Location> location,
 
 LogicalResult verifyDynamicReshapeOp(std::optional<Location> location,
                                      Value outputShape, Value result) {
-  auto resultType = result.getType().dyn_cast<RankedTensorType>();
-  auto outputShapeType = outputShape.getType().dyn_cast<RankedTensorType>();
-  if (resultType && outputShapeType && outputShapeType.hasStaticShape() &&
+  auto resultType = result.getType().dyn_cast<ShapedType>();
+  auto outputShapeType = outputShape.getType().dyn_cast<ShapedType>();
+  if (resultType.hasRank() && outputShapeType.hasStaticShape() &&
```
GleasonK, line 3504: Deleted the `outputShapeType` and `resultType` null check? Should `dyn_cast` be `cast` if this is guaranteed?

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3471,13 +3499,22 @@ LogicalResult verifyDynamicPadOp(std::optional<Location> location,
...
                              "elements in output_shape");
+  SmallVector<int64_t> outputShapeVec;
+  if (succeeded(matchInts(outputShape, outputShapeVec))) {
+    auto inferredResultType =
+        RankedTensorType::get(outputShapeVec, resultType.getElementType());
+    if (!isCompatibleForHloTypeInference(inferredResultType, resultType))
+      return emitOptionalError(
+          location, "inferred type ", inferredResultType,
+          " is incompatible with return type of operation ", resultType);
+  }
```
GleasonK, line 3517: This block of code appears 3 times in this PR, is it worth making this generic?

GleasonK (approved):

stablehlo/dialect/TypeInference.cpp
burmako: Yes, this is guaranteed. Good catch that this should be a `cast` - we actually have quite a few places in our codebase which are in a similar situation which I have all fixed now.

stablehlo/dialect/TypeInference.cpp
burmako: Done.

PR #1379, opened 2023-04-06
Add Float8E4M3FNUZ and Float8E5M2FNUZ type support.

jakeh-gc:
As proposed in [RFC: Float8E4M3FNUZ and Float8E5M2FNUZ](https://github.com/openxla/stablehlo/blob/main/rfcs/20230321-fp8_fnuz.md) (#1342), this change adds support for these types to StableHLO.

This includes the type definitions, vhlo, and interpreter support. The testing approach mirrors the BFloat16 tests, since it is also a "non-standard" floating point type supported by StableHLO.


stablehlo/dialect/VhloBytecode.cpp
```diff
@@ -316,6 +316,14 @@ enum TypeCode {
...
   ///   }
   kWitnessV1Type = 26,
+
+  ///   FloatF8E4M3FNUZV1Type {
+  ///   }
+  kFloatF8E4M3FNUZV1Type = 27,
+
+  ///   FloatF8E5M2FNUZV1Type {
+  ///   }
+  kFloatF8E5M2FNUZV1Type = 28,
```
burmako, line 326: How about we keep this list ordered alphabetically, even if this would mean that the enum values are no longer sorted in ascending order?

stablehlo/tests/interpret_abs.mlir
```diff
@@ -9,6 +9,24 @@ func.func @abs_op_test_si64() {
 
 // -----
 
+func.func @abs_op_test_f8_e4m3_fnuz() {
+  %operand = stablehlo.constant dense<[24.0, -24.0, -0.0, 0x80]> : tensor<4xf8E4M3FNUZ>
+  %result = stablehlo.abs %operand : tensor<4xf8E4M3FNUZ>
+  check.expect_almost_eq_const %result, dense<[24.0, 24.0, 0.0, 0x80]> : tensor<4xf8E4M3FNUZ>
+  func.return
+}
```
burmako, line 17: @sdasgup3 What's our policy for updating interpreter tests for new types? If I recall correctly from the [Q&A](https://github.com/openxla/stablehlo/blob/main/docs/reference.md#testing-guidelines), we only add test for highest bitwidth types, and rely on APInt/APFloat to handle the general case?

stablehlo/dialect/VhloTypes.td
```diff
@@ -86,6 +86,12 @@ def VHLO_FloatF8E4M3FNV1 : VHLO_TypeDef<"FloatF8E4M3FNV1", "f8E4M3FN_v1", "0.9.0
 // Corresponds to the 'f8E5M2' FloatType from the StableHLO spec.
 def VHLO_FloatF8E5M2V1 : VHLO_TypeDef<"FloatF8E5M2V1", "f8E5M2_v1", "0.9.0", "current">;
 
+// Corresponds to the 'f8E4M3FNUZ' FloatType from the StableHLO spec.
+def VHLO_FloatF8E4M3FNUZV1 : VHLO_TypeDef<"FloatF8E4M3FNUZV1", "f8E4M3FNUZ_v1", "0.9.0", "current">;
+
+// Corresponds to the 'f8E5M2FNUZ' FloatType from the StableHLO spec.
+def VHLO_FloatF8E5M2FNUZV1 : VHLO_TypeDef<"FloatF8E5M2FNUZV1", "f8E5M2FNUZ_v1", "0.9.0", "current">;
+
```
burmako, line 94: @GleasonK Looks like the time has come to figure out the versioning scheme moving forward (similarly to what we discussed in #1319). What would be your recommendation?

burmako (changes requested):
Thanks a lot for the super quick follow-up! I understand that this is a draft at the moment, but I left a few quick comments to prefetch some discussions for when the PR is ready for review.

stablehlo/dialect/VhloTypes.td
GleasonK, line 94: A few notes on the compatibility side, since adding a new type is not forward compatible and requires a new version number:

- These should be `"0.10.0", "current"`
- `stablehlo/dialect/Version.h` function `getCurrentVersion` should get bumped to `(0,10,0)`.
- `stablehlo/tests/stablehlo_legalize_to_vhlo.mlir` needs a test (looks like you have this covered already)
- Add an invalid downgrade test. Unfortunately this is the first opset change since our public release so there isn't a dedicated test file yet, but look at [vhlo_to_version_downgrade_invalid.mlir](https://github.com/openxla/stablehlo/blob/6886b59f6cd4369674e7e3beff61301c145176e2/stablehlo/tests/vhlo_to_version_downgrade_invalid.mlir#L102) from before the public release cleanup for examples of what this test looked like for fp8_E5M2. The new test should specify `0.9.0` as the target version and error if a new FP8 type is used.

jakeh-gc: This predicts exactly one of my questions I had for you guys 👍

stablehlo/dialect/VhloBytecode.cpp
jakeh-gc: I got the opposite impression from the comment at the top of this enum
```C
/// This enum contains marker codes used to indicate which type is
/// currently being decoded, and how it should be decoded. The order of these
/// codes must not be changed, as any changes will break compatibility
/// with older bytecode.
```

jakeh-gc: Oh, I see what you mean. Order the definitions, but preserve the values.

GleasonK: This may make it difficult to figure out "what value should the next added be?" Personally in favor of leaving it in numeric order, sorting was nice as a starting point, but I think will add to confusion and risk gaps in the enum values longer term.

stablehlo/dialect/VhloTypes.td
GleasonK: Just seeing this. Added a separate comment with a few steps

stablehlo/dialect/VhloBytecode.cpp
burmako: Sounds good. Let's stick with sorting values in ascending order.

stablehlo/dialect/VhloTypes.td
GleasonK: Moved to https://github.com/openxla/stablehlo/pull/1379#discussion_r1159924774

stablehlo/tests/interpret_abs.mlir
sdasgup3: @jakeh-gc Thanks for the interpreter support and tests.  

As the newly introduced types are handled uniformly just like other floating point types in the context of add op, I would recommend to stick to the guideline [G1](https://github.com/openxla/stablehlo/blob/main/docs/reference.md#testing-guidelines), that Eugene pointed out. With that we can skip adding extra tests here and for other ops.

>> I'm happy to drop them.
Thanks!

>>  There are code paths only exercised by these types.

You are absolutely right in pointing that! 
For every op we have a code path, in Tensor::set and Tensor::get (exactly as you pointed), which basically packs and unpacks the tensor data and by doing so  "covers" all the types. As the interpreter tests are written based on meeting code coverage, this means for every op we should write tests for every type, which is not valuable and hard to maintain. 

>> Clearly, 2 could be covered by just a few cases.

That is exactly the path we followed!
 
To avoid the problem, mentioned above, we decided, via [G3](https://github.com/openxla/stablehlo/blob/main/docs/reference.md#testing-guidelines) to write dedicated tests  just for this packing/unpacking code and thereby reduce the burden on op level tests.

stablehlo/dialect/VhloTypes.td
jakeh-gc: @GleasonK I assume point 2 also means updating places that refer to `0.9.0` in the present tense?
Examples of what I mean:
https://github.com/openxla/stablehlo/blob/7f6976e15595d3a37ad177a58b4a3c2024d9ae9c/docs/bytecode.md?plain=1#L24
https://github.com/openxla/stablehlo/blob/7f6976e15595d3a37ad177a58b4a3c2024d9ae9c/docs/compatibility.md?plain=1#L10

stablehlo/tests/interpret_abs.mlir
jakeh-gc: With the exception of `constant`, I've removed the interpreter test cases.

stablehlo/dialect/VhloTypes.td
jakeh-gc: I think I've covered these items. Let me know if there's anything I've missed wrt. versioning.

GleasonK (approved):
Apologies for the delay, we were making sure we had a process for integrating these version bumping changes. I'll be adding document to the repo in the near future to outline this proposed process, but I that shouldn't block this submission since we have means of detecting and testing forward/backward incompatibilities that are consistent with the current state of the repo, that should be enough for now.

Changes LGTM! Thanks!

Reviewed everything except interpreter, will let @sdasgup3 give the final word there.

GleasonK:
Forgot to updated assignees. No further flags from my end! Let's get feedback/approvals from the other reviewers then merge this in

burmako (approved):
LGTM. Kevin and I discussed some follow-ups last week (#1409 and #1410), but you don't need to worry about them - we'll take care of them separately.

GleasonK:
Thanks again for the contribution! I'll get the finishing touches that Eugene mentioned going so we can tag a commit this week as the `0.10.0` release containing this feature!

PR #1382, opened 2023-04-08
Add interpreter for SignOp.

### LINKED ISSUE #1115: Add interpreter for SignOp

karthikrangasai:
closes #1115 

```
(I1) operand is a tensor of signed integer, floating-point, or complex type.
(C1) operand and result have the same type.
```

```
I1: a) operand is not a tensor of signed integer, floating-point, or complex type. (Covered by ODS).
C1: a) type(operand) != type(result). (Covered by ODS).
```

### DIFF
--- a/stablehlo/reference/Element.cpp
+++ b/stablehlo/reference/Element.cpp
@@ -774,18 +774,6 @@ Element sign(const Element &el) {
       return Element(type, std::complex<APFloat>(APFloat::getNaN(elSemantics),
                                                  APFloat::getNaN(elSemantics)));
 
-    // auto absElVal = abs(el).getFloatValue();
-    // if (absElVal.isZero()) return Element(type, (double)0.0);
-
-    // bool roundingErr;
-    // APFloat resultReal(elVal.real().convertToDouble() /
-    //                    absElVal.convertToDouble());
-    // resultReal.convert(elSemantics, APFloat::rmNearestTiesToEven,
-    // &roundingErr); APFloat resultImag(elVal.imag().convertToDouble() /
-    //                    absElVal.convertToDouble());
-    // resultImag.convert(elSemantics, APFloat::rmNearestTiesToEven,
-    // &roundingErr);
-
     return el / Element(type, abs(el).getFloatValue().convertToDouble());
   }
 
--- a/stablehlo/reference/Element.cpp
+++ b/stablehlo/reference/Element.cpp
@@ -747,6 +747,72 @@ Element rsqrt(const Element &el) {
       [](std::complex<double> e) { return 1.0 / std::sqrt(e); });
 }
 
+Element sign(const Element &el) {
+  Type type = el.getType();
+
+  if (isSupportedIntegerType(type)) {
+    auto intEl = el.getIntegerValue();
+    if (intEl.isZero()) {
+      return Element(type, (int64_t)0);
+    } else if (intEl.isNegative()) {
+      return Element(type, (int64_t)-1);
+    } else {
+      return Element(type, (int64_t)1);
+    }
+  }
+
+  if (isSupportedFloatType(type)) {
+    auto elVal = el.getFloatValue();
+
+    if (elVal.isNaN()) {
+      return Element(el);
+    } else {
+      if (elVal.isNegZero()) {
+        return Element(type, (double)-0.0);
+      } else if (elVal.isPosZero()) {
+        return Element(type, (double)+0.0);
+      } else if (elVal.isNegInfinity()) {
+        return Element(type, (double)-1.0);
+      } else if (elVal.isPosInfinity()) {
+        return Element(type, (double)+1.0);
+      } else if (elVal.isNegative()) {
+        return Element(type, (double)-1.0);
+      } else if (!elVal.isNegative()) {
+        return Element(type, (double)+1.0);
+      }
+    }
+  }
+
+  if (isSupportedComplexType(type)) {
+    auto elVal = el.getComplexValue();
+    const llvm::fltSemantics &elSemantics = elVal.real().getSemantics();
+    if (elVal.real().isNaN() || elVal.imag().isNaN()) {
+      return Element(type,
+                     std::complex<APFloat>(APFloat::getNaN(elSemantics),
+                                           APFloat::getZero(elSemantics)));
+    } else {
+      auto absElValDouble = abs(el).getFloatValue().convertToDouble();
+      auto signVal =
+          std::complex<double>(elVal.real().convertToDouble() / absElValDouble,
+                               elVal.imag().convertToDouble() / absElValDouble);
+
+      bool roundingErr;
+      APFloat resultReal(signVal.real());
+      resultReal.convert(elSemantics, APFloat::rmNearestTiesToEven,
+                         &roundingErr);
+      APFloat resultImag(signVal.imag());
+      resultImag.convert(elSemantics, APFloat::rmNearestTiesToEven,
+                         &roundingErr);
+      auto retVal =
+          Element(type, std::complex<APFloat>(resultReal, resultImag));
+      return retVal;
+    }
+  }
+
+  report_fatal_error(invalidArgument("Unsupported element type: %s",
+                                     debugString(type).c_str()));
+}
+
 Element sine(const Element &el) {
   return mapWithUpcastToDouble(
       el, [](double e) { return std::sin(e); },
--- a/stablehlo/reference/Element.cpp
+++ b/stablehlo/reference/Element.cpp
@@ -752,13 +752,12 @@ Element sign(const Element &el) {
 
   if (isSupportedIntegerType(type)) {
     auto intEl = el.getIntegerValue();
-    if (intEl.isZero()) {
+    if (intEl.isZero())
       return Element(type, (int64_t)0);
-    } else if (intEl.isNegative()) {
+    else if (intEl.isNegative())
       return Element(type, (int64_t)-1);
-    } else {
+    else
       return Element(type, (int64_t)1);
-    }
   }
 
   if (isSupportedFloatType(type)) {
@@ -766,20 +765,19 @@ Element sign(const Element &el) {
 
     if (elVal.isNaN()) {
       return Element(el);
-    } else {
-      if (elVal.isNegZero()) {
-        return Element(type, (double)-0.0);
-      } else if (elVal.isPosZero()) {
-        return Element(type, (double)+0.0);
-      } else if (elVal.isNegInfinity()) {
-        return Element(type, (double)-1.0);
-      } else if (elVal.isPosInfinity()) {
-        return Element(type, (double)+1.0);
-      } else if (elVal.isNegative()) {
-        return Element(type, (double)-1.0);
-      } else if (!elVal.isNegative()) {
-        return Element(type, (double)+1.0);
-      }
+    }
+    if (elVal.isNegZero()) {
+      return Element(type, (double)-0.0);
+    } else if (elVal.isPosZero()) {
+      return Element(type, (double)+0.0);
+    } else if (elVal.isNegInfinity()) {
+      return Element(type, (double)-1.0);
+    } else if (elVal.isPosInfinity()) {
+      return Element(type, (double)+1.0);
+    } else if (elVal.isNegative()) {
+      return Element(type, (double)-1.0);
+    } else if (!elVal.isNegative()) {
+      return Element(type, (double)+1.0);
     }
   }
 
@@ -790,23 +788,19 @@ Element sign(const Element &el) {
       return Element(type,
                      std::complex<APFloat>(APFloat::getNaN(elSemantics),
                                            APFloat::getZero(elSemantics)));
-    } else {
-      auto absElValDouble = abs(el).getFloatValue().convertToDouble();
-      auto signVal =
-          std::complex<double>(elVal.real().convertToDouble() / absElValDouble,
-                               elVal.imag().convertToDouble() / absElValDouble);
-
-      bool roundingErr;
-      APFloat resultReal(signVal.real());
-      resultReal.convert(elSemantics, APFloat::rmNearestTiesToEven,
-                         &roundingErr);
-      APFloat resultImag(signVal.imag());
-      resultImag.convert(elSemantics, APFloat::rmNearestTiesToEven,
-                         &roundingErr);
-      auto retVal =
-          Element(type, std::complex<APFloat>(resultReal, resultImag));
-      return retVal;
     }
+    auto absElValDouble = abs(el).getFloatValue().convertToDouble();
+    auto signVal =
+        std::complex<double>(elVal.real().convertToDouble() / absElValDouble,
+                             elVal.imag().convertToDouble() / absElValDouble);
+
+    bool roundingErr;
+    APFloat resultReal(signVal.real());
+    resultReal.convert(elSemantics, APFloat::rmNearestTiesToEven, &roundingErr);
+    APFloat resultImag(signVal.imag());
+    resultImag.convert(elSemantics, APFloat::rmNearestTiesToEven, &roundingErr);
+    auto retVal = Element(type, std::complex<APFloat>(resultReal, resultImag));
+    return retVal;
   }
 
   report_fatal_error(invalidArgument("Unsupported element type: %s",
--- a/stablehlo/reference/Element.cpp
+++ b/stablehlo/reference/Element.cpp
@@ -751,22 +751,19 @@ Element sign(const Element &el) {
   Type type = el.getType();
 
   if (isSupportedIntegerType(type)) {
-    auto intEl = el.getIntegerValue();
-    if (intEl.isZero()) return Element(type, (int64_t)0);
-    if (intEl.isNegative())
-      return Element(type, (int64_t)-1);
-    else
-      return Element(type, (int64_t)1);
+    auto elVal = el.getIntegerValue();
+    if (elVal.isNegative()) return Element(type, (int64_t)-1);
+    if (elVal.isZero()) return Element(type, (int64_t)0);
+    return Element(type, (int64_t)1);
   }
 
   if (isSupportedFloatType(type)) {
     auto elVal = el.getFloatValue();
-
-    if (elVal.isNaN()) return Element(el);
+    if (elVal.isNaN()) return el;
     if (elVal.isNegZero()) return Element(type, (double)-0.0);
-    if (elVal.isPosZero()) return Element(type, (double)0.0);
+    if (elVal.isPosZero()) return Element(type, (double)+0.0);
     if (elVal.isNegative()) return Element(type, (double)-1.0);
-    if (!elVal.isNegative()) return Element(type, (double)1.0);
+    return Element(type, (double)1.0);
   }
 
   if (isSupportedComplexType(type)) {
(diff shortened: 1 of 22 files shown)


stablehlo/tests/interpret_sign.mlir
```diff
@@ -0,0 +1,168 @@
+// RUN: stablehlo-translate --interpret -split-input-file %s
+
+// CHECK-LABEL: Evaluated results of function: sign_op_test_si4
```
ghpvnist, line 3: Per this [checklist](https://github.com/openxla/stablehlo/blob/main/docs/reference.md?plain=1#L155-L157), we only need to test `si64`, `f64`, and `c128`.

stablehlo/tests/interpret_sign.mlir
```diff
@@ -0,0 +1,168 @@
...
+  // CHECK-NEXT: -1 : i32
+  // CHECK-NEXT: 0 : i32
+  // CHECK-NEXT: 1 : i32
+}
+
+// -----
+
+// CHECK-LABEL: Evaluated results of function: sign_op_test_si64
+func.func @sign_op_test_si64() -> tensor<3xi64> {
+  %operand = stablehlo.constant dense<[-9223372036854775807, 0, 9223372036854775807]> : tensor<3xi64>
```
ghpvnist, line 57: I think you meant the negative number to be `-9223372036854775808`. In general, let's use more readable numbers as whether we test maximal/minimal representable values does not give us any additional information and avoids typo. Proposed numbers:
```
%operand = stablehlo.constant dense<[-1, 0, 1]> : tensor<3xi64>
```

stablehlo/tests/interpret_sign.mlir
```diff
@@ -0,0 +1,168 @@
...
+  // CHECK-NEXT: -1.000000e+00 : f32
+  // CHECK-NEXT: 1.000000e+00 : f32
+}
+
+// -----
+
+// CHECK-LABEL: Evaluated results of function: sign_op_test_f64
+func.func @sign_op_test_f64() -> tensor<8xf64> {
+  // -0.0, +0.0, -3.14, +3.14, -NaN, +NaN, -Inf, +Inf
+  %operand = stablehlo.constant dense<[-0.0, 0.0, -3.14159265358979323846, 3.14159265358979323846, 0xFFFFFFFFFFFFFFFF, 0x7FFFFFFFFFFFFFFF, 0xFFF0000000000000, 0x7FF0000000000000]> : tensor<8xf64>
```
ghpvnist, line 126: Let's test numbers which are relevant to the logic of the code without duplicating logical paths:
```
// +NaN, -1.0, -0.0, +0.0, 1.0
%operand = stablehlo.constant dense<[0x7FFFFFFFFFFFFFFF, -1.0, -0.0, 0.0, 1.0]> : tensor<5xf64>
```

stablehlo/reference/Element.cpp
```diff
@@ -720,6 +720,72 @@ Element rsqrt(const Element &el) {
       [](std::complex<double> e) { return 1.0 / std::sqrt(e); });
 }
 
+Element sign(const Element &el) {
+  Type type = el.getType();
+
+  if (isSupportedIntegerType(type)) {
+    auto intEl = el.getIntegerValue();
+    if (intEl.isZero()) {
```
ghpvnist, line 728: Let's remove the curly braces following the [llvm style guide](https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements) across the changes in this file.

stablehlo/reference/Element.cpp
```diff
@@ -720,6 +720,72 @@ Element rsqrt(const Element &el) {
...
+      return Element(type, (int64_t)1);
+    }
+  }
+
+  if (isSupportedFloatType(type)) {
+    auto elVal = el.getFloatValue();
+
+    if (elVal.isNaN()) {
+      return Element(el);
+    } else {
```
ghpvnist, line 742: No need for else block here.

ghpvnist (changes requested):
@karthikrangasai Thank you for the valuable contribution! For your reference, we are using these two checklists on our PRs: [reference_checklist](https://github.com/openxla/stablehlo/blob/main/docs/reference_checklist.md) and [testing_guideline](https://github.com/openxla/stablehlo/blob/main/docs/reference.md#testing-guidelines).

Can you also negate (C1) in your PR description:
```
I1: a) operand is not a tensor of integer, floating-point, or complex type. (Covered by ODS).
C1: a) shape(operand) != shape(result). (Covered by ODS).
    b) element_type(operand) != element_type(result). (Covered by ODS).
```

ghpvnist (changes requested):
Few more comments and it lgtm!

stablehlo/reference/Element.cpp
```diff
@@ -720,6 +720,57 @@ Element rsqrt(const Element &el) {
       [](std::complex<double> e) { return 1.0 / std::sqrt(e); });
 }
 
+Element sign(const Element &el) {
```
ghpvnist, line 723: Sorry for the repeated reviews! A few more issues popped up after enabling the CI. Here is the comprehensive fixes:
```
Element sign(const Element &el) {
  Type type = el.getType();

  if (isSupportedIntegerType(type)) {
    auto intEl = el.getIntegerValue();

    auto isNonZero = Element(type, (int64_t)!intEl.isZero());
    return intEl.isNegative() ? -isNonZero : isNonZero;
  }

    auto elVal = el.getFloatValue();

    if (elVal.isNaN()) return el;
    auto isNonZero = Element(type, (double)!elVal.isZero());
    return elVal.isNegative() ? -isNonZero : isNonZero;
  }

  if (isSupportedComplexType(type)) {
    auto elVal = el.getComplexValue();
    const llvm::fltSemantics &elSemantics = elVal.real().getSemantics();

    if (elVal.real().isNaN() || elVal.imag().isNaN())
      return Element(type, std::complex<APFloat>(APFloat::getNaN(elSemantics),
                                                 APFloat::getNaN(elSemantics)));

    auto absElVal = abs(el).getFloatValue();
    if (absElVal.isZero()) return Element(type, (double)0.0);

    bool roundingErr;
    APFloat resultReal(elVal.real().convertToDouble() /
                       absElVal.convertToDouble());
    resultReal.convert(elSemantics, APFloat::rmNearestTiesToEven, &roundingErr);
    APFloat resultImag(elVal.imag().convertToDouble() /
                       absElVal.convertToDouble());
    resultImag.convert(elSemantics, APFloat::rmNearestTiesToEven, &roundingErr);

    return Element(type, std::complex<APFloat>(resultReal, resultImag));
  }

  report_fatal_error(invalidArgument("Unsupported element type: %s",
                                     debugString(type).c_str()));
}
```
Feedback:
* Reduced logic in handling integer and float type.
* Unnecessary wrapper around `el`: `if (elVal.isNaN()) return Element(el);` -> `if (elVal.isNaN()) return el;`.
* HLO returns NaNs for both real and imaginary parts, which makes sense since `abs(NaN) = NaN` and divide by NaN should also be NaN. We should also follow that.

stablehlo/tests/interpret_sign.mlir
```diff
@@ -0,0 +1,43 @@
+// RUN: stablehlo-translate --interpret -split-input-file %s
```
ghpvnist, line 1: We recently migrated our testing infrastructure from lit-based testing. Here is the file following our current format (also removed duplicate comment under complex test):
```
// RUN: stablehlo-translate --interpret -split-input-file %s

// CHECK-LABEL: Evaluated results of function: sign_op_test_si64
func.func @sign_op_test_si64() {
  %operand = stablehlo.constant dense<[-1, 0, 1]> : tensor<3xi64>
  %result = stablehlo.sign %operand : tensor<3xi64>
  check.expect_eq_const %result, dense<[-1, 0, 1]> : tensor<3xi64>
  func.return
}

// -----

// CHECK-LABEL: Evaluated results of function: sign_op_test_f64
func.func @sign_op_test_f64() {
  // +NaN, -1.0, -0.0, +0.0, 1.0
  %operand = stablehlo.constant dense<[0x7FFFFFFFFFFFFFFF, -1.0, -0.0, 0.0, 1.0]> : tensor<5xf64>
  %result = stablehlo.sign %operand : tensor<5xf64>
  check.expect_almost_eq_const %result, dense<[0x7FFFFFFFFFFFFFFF, -1.0, -0.0, 0.0, 1.0]> : tensor<5xf64>
  func.return
}

// -----

// CHECK-LABEL: Evaluated results of function: sign_op_test_c128
func.func @sign_op_test_c128() {
  // (+NaN, +0.0), (+0.0, +NaN), (0.0, 1.0)
  %operand = stablehlo.constant dense<[(0x7FF0000000000001, 0x0000000000000000),
                                       (0x0000000000000000, 0x7FF0000000000001),
                                       (0.0, 1.0)]> : tensor<3xcomplex<f64>>
  %result = stablehlo.sign %operand : tensor<3xcomplex<f64>>
  check.expect_almost_eq_const %result, dense<[(0x7FF0000000000001, 0x7FF0000000000001),
                                               (0x7FF0000000000001, 0x7FF0000000000001),
                                               (0.0, 1.0)]> : tensor<3xcomplex<f64>>
  func.return
}
```

ghpvnist (changes requested):
Few more comments :)

stablehlo/reference/Element.cpp
burmako: "Reduced logic in handling integer and float type". I would propose that we aim to have 1:1 correspondence with the code provided in the spec. 

The implementation proposed in the comment is clever, but a non-zero mental effort is needed to correlate it with the spec. Instead, how about this (I slightly reordered the implementation provided in the current version of the PR):

```C++
  if (isSupportedIntegerType(type)) {
    auto elVal = el.getIntegerValue();
    if (intEl.isNegative()) return Element(type, (int64_t)-1);
    if (intEl.isZero()) return Element(type, (int64_t)0);
    return Element(type, (int64_t)1);
  }

    auto elVal = el.getFloatValue();
    if (elVal.isNaN()) return Element(el);
    if (elVal.isNegative()) return Element(type, (double)-1.0);
    if (elVal.isNegZero()) return Element(type, (double)-0.0);
    if (elVal.isPosZero()) return Element(type, (double)+0.0);
    return Element(type, (double)1.0);
  }
```

stablehlo/reference/Element.cpp
```diff
@@ -720,6 +720,57 @@ Element rsqrt(const Element &el) {
...
+
+    auto absElValDouble = abs(el).getFloatValue().convertToDouble();
+    bool roundingErr;
+
+    APFloat resultReal(elVal.real().convertToDouble() / absElValDouble);
+    resultReal.convert(elSemantics, APFloat::rmNearestTiesToEven, &roundingErr);
+    APFloat resultImag(elVal.imag().convertToDouble() / absElValDouble);
+    resultImag.convert(elSemantics, APFloat::rmNearestTiesToEven, &roundingErr);
+
+    return Element(type, std::complex<APFloat>(resultReal, resultImag));
```
burmako, line 767: Can we reformulate the complex case in terms of already existing functions, i.e. `el / abs(el)`? (While still keeping the special logic that handles NaNs).

burmako (changes requested):

stablehlo/reference/Element.cpp
karthikrangasai: Hello @burmako , I first tried making the code to be as simple as `el / abs(el)`.

But the spec for `divide` operation says:
* (C1) `lhs`, `rhs` and `result` have the same type.

The spec of `abs` says:
* (C2)  `operand` and `result` have the same element type, except when the element type of the `operand` is complex type, in which case the element type of the `result` is the element type of the complex type (e.g. the element type of the `result` is `f64` for operand type `complex<f64>`).

So for a given floating-point type `T` and number of elements `N`:

`tensor<Nxcomplex<T>> / abs(tensor<Nxcomplex<T>>)` becomes `tensor<Nxcomplex<T>> / tensor<NxT>`

and the latter is not supported. So that's why I wrote the code like this.

If I have assumed or understood anything incorrectly please do correct me and I shall make the changes accordingly.

stablehlo/reference/Element.cpp
karthikrangasai: Hello @burmako ,

The logic for floating-type has an error here.

`elVal.isNegative()` will be `true` for `-0.0` as well.
Then it will then return `-1.0` instead of `-0.0`.

The order should be the following imo:

```c++
    auto elVal = el.getFloatValue();
    if (elVal.isNaN()) return el;
    if (elVal.isNegZero()) return Element(type, (double)-0.0);
    if (elVal.isPosZero()) return Element(type, (double)+0.0);
    if (elVal.isNegative()) return Element(type, (double)-1.0);
    return Element(type, (double)1.0);
  }
```

stablehlo/tests/interpret_sign.mlir
karthikrangasai: Hello @ghpvnist ,

Shouldn't the output for `sign_op_test_c128()` be the following:

```mlir
dense<[
    (0x7FF0000000000001, 0.0),
    (0x7FF0000000000001, 0.0),
    (0.0, 1.0)
]> : tensor<3xcomplex<f64>>
```

Wouldn't the imaginary part be zero ?

stablehlo/reference/Element.cpp
burmako: Nice catch! I've updated #1395 to reflect this order as well.

stablehlo/reference/Element.cpp
burmako: You're right - this also indicates a gap in the spec, which says `divide(x, abs(x))`. Let me think how to best fix this in the spec. Will follow up in a few hours after my meetings.

stablehlo/reference/Element.cpp
karthikrangasai: Thanks 😄

stablehlo/reference/Element.cpp
karthikrangasai: Sure, will do.

Also, is there a plan to include `complex / float`, `complex / int`, `float / int` in the spec ?

Like wise for all other ops `+`, `-`, and `*`.

burmako: I think that the idiomatic way of going about this in HLO is to use `convert`: https://github.com/openxla/stablehlo/blob/main/docs/spec.md#convert, so we probably should do that as well.

karthikrangasai: #1349 implements it. I should be able to leverage that when it is merged.

stablehlo/tests/interpret_sign.mlir
ghpvnist: Hi @karthikrangasai, as explained in my earlier comment, HLO also returns NaNs for both real and imaginary parts. As StableHLO aims to follow closely with HLO, we should also do the same here.
For complex types, abs(x) returns a float, so the actual computation is like:
```
x / (abs(x), abs(x)) =  (x.real / NaN, x.imag / NaN) = (NaN, NaN)
```

karthikrangasai: Thanks a lot @ghpvnist .

stablehlo/reference/Element.cpp
```diff
@@ -720,6 +720,57 @@ Element rsqrt(const Element &el) {
...
+Element sign(const Element &el) {
+  Type type = el.getType();
+
+  if (isSupportedIntegerType(type)) {
+    auto intEl = el.getIntegerValue();
+    if (intEl.isZero()) return Element(type, (int64_t)0);
+    if (intEl.isNegative())
+      return Element(type, (int64_t)-1);
+    else
+      return Element(type, (int64_t)1);
```
burmako, line 732: Let's reduce nesting by refactoring this into:

```C++
if (intEl.isNegative()) return Element(type, (int64_t)-1);
return Element(type, (int64_t)1);
```

stablehlo/reference/Element.cpp
```diff
@@ -720,6 +720,57 @@ Element rsqrt(const Element &el) {
...
+  }
+
+  if (isSupportedFloatType(type)) {
+    auto elVal = el.getFloatValue();
+
+    if (elVal.isNaN()) return Element(el);
+    if (elVal.isNegZero()) return Element(type, (double)-0.0);
+    if (elVal.isPosZero()) return Element(type, (double)0.0);
+    if (elVal.isNegative()) return Element(type, (double)-1.0);
+    if (!elVal.isNegative()) return Element(type, (double)1.0);
```
burmako, line 742: Drop `if (!elVal.isNegative())` to signify that this is the final case of the chain of conditionals?

stablehlo/tests/interpret_sign.mlir
burmako: This conversation has been resolved, but I don't think the code has been updated?

stablehlo/reference/Element.cpp
burmako: I think we don't even have to wait for convert to be merged, since there's already a compact way of expressing this: `return el / Element(type, abs(el).getFloatValue().convertToDouble())`.

stablehlo/reference/Element.cpp
```diff
@@ -720,6 +720,57 @@ Element rsqrt(const Element &el) {
...
+
+    if (elVal.real().isNaN() || elVal.imag().isNaN())
+      return Element(type,
+                     std::complex<APFloat>(APFloat::getNaN(elSemantics),
+                                           APFloat::getZero(elSemantics)));
+
+    if (elVal.real().isZero() && elVal.imag().isZero())
+      return Element(type,
+                     std::complex<APFloat>(APFloat::getZero(elSemantics),
+                                           APFloat::getZero(elSemantics)));
```
burmako, line 757: Also, how about we drop the check for whether el is equal to zero to simplify the implementation?

The spec suggests that this is implementation-defined behavior, so we are free to do whatever we want here: "StableHLO programs are validated through an extensive set of constraints for individual ops, which rules out many classes of errors prior to run time. However, error conditions are still possible, e.g. through integer overflows, out-of-bounds accesses, etc. Unless explicitly called out, all these errors result in implementation-defined behavior, but this may change in the future".

burmako (changes requested):

stablehlo/reference/Element.cpp
burmako: Huh interesting it looks like this actually might need to be part of the spec: https://github.com/openxla/stablehlo/pull/1395#discussion_r1164397409. Let's discuss there, update the spec if needed and then reproduce this in the interpreter.

stablehlo/tests/interpret_sign.mlir
karthikrangasai: Oh sorry, I thought I pushed the changes. Pushing them now.

burmako (approved):

stablehlo/reference/Element.cpp
karthikrangasai: Hello @burmako ,

`stablehlo/testdata/sign_special_0_dtype_complex64.mlir` defines `stablehlo.sign(0+0i)` outputs `0+0i`. But the suggested implementation returns `NaN + NaN i`.

This is the only test failing in CI. Any suggestion on what I should do ?

burmako: How about we disable the test for now and take care of it in #1395?

karthikrangasai: Sure, that works.

ghpvnist (approved):

ghpvnist (changes requested):
The code lgtm. In general, I'd suggest to keep the constraints verbatim from the spec as they have explicit meaning:

From:
```
(I1) operand is a tensor of integer, floating-point, or complex type.
(C1) operand and result have the same shape and type.
```
To:
```
(I1) operand is a tensor of signed integer, floating-point, or complex type.
(C1) operand and result have the same type.
```

From:
```
I1: a) operand is not a tensor of integer, floating-point, or complex type. (Covered by ODS).
C1: a) shape(operand) != shape(result). (Covered by ODS).
    b) element_type(operand) != element_type(result). (Covered by ODS).
```
To:
```
I1: a) operand is not a tensor of signed integer, floating-point, or complex type. (Covered by ODS).
C1: a) type(operand) != type(result). (Covered by ODS).
```

Feedback:
* SignOp only supports [Signed Integers](https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/StablehloOps.td#L579).
* Listing as "same shape and type" is redundant as shape is covered by type. The [`HLO_CompatibleOperandsAndResultType`](https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/StablehloOps.td#L578) Trait takes care of both the shape and the element type.

PR #1383, opened 2023-04-09
Drop indices_of_shape_operands once they are no longer needed

burmako:
--stablehlo-canonicalize-dynamism can now drop indices_of_shape_operands once they are no longer needed, similarly to how it transforms dynamic_foo ops to foo ops when operands like output_shape are no longer needed.

To make sure that we only do the transformation on valid IR, I also wrote verification logic for indices_of_shape_operands as well as tests for this logic.

In the future, when we upgrade indices_of_shape_operands from an experiment to a full-fledged StableHLO feature, this logic will be moved to a proper verifier.

### DIFF
--- a/stablehlo/dialect/Base.cpp
+++ b/stablehlo/dialect/Base.cpp
@@ -411,8 +411,8 @@ LogicalResult getShapeRefinements(
   if (indicesAttr.getNumElements() != op->getNumResults())
     return emitOptionalError(location, "indices_of_shape_operands: number of ",
                              "elements (", indicesAttr.getNumElements(), ") ",
-                             "must be equal to the number of results (",
-                             op->getNumResults(), ")");
+                             "must be equal to the number of operation results",
+                             " (", op->getNumResults(), ")");
   if (indicesAttr.getType().getRank() != 1)
     return emitOptionalError(location, "indices_of_shape_operands: must have ",
                              "rank = 1");
--- a/stablehlo/dialect/Base.cpp
+++ b/stablehlo/dialect/Base.cpp
@@ -421,8 +421,8 @@ LogicalResult getShapeRefinements(
                              "i64 element type");
 
   auto resultIndex = 0;
-  for (auto [operandIndex, result] :
-       llvm::zip(indicesAttr.getValues<int64_t>(), op->getResults())) {
+  for (auto [operandIndex, resultType] :
+       llvm::zip(indicesAttr.getValues<int64_t>(), op->getResultTypes())) {
     if (operandIndex < 0 || operandIndex >= op->getNumOperands())
       return emitOptionalError(location, "indices_of_shape_operands: index #",
                                resultIndex, " (", operandIndex, ") ",
@@ -432,11 +432,11 @@ LogicalResult getShapeRefinements(
     Value operand = op->getOperand(operandIndex);
     SmallVector<int64_t> refinement;
     if (failed(hlo::matchInts(operand, refinement))) return failure();
-    if (!isCompatibleForHloTypeInference(operand, result.getType()))
+    if (!isCompatibleForHloTypeInference(operand, resultType))
       return emitOptionalError(
           location, "indices_of_shape_operands: refinement #", resultIndex,
           " ([", refinement, "]) must be compatible with operation result (",
-          result.getType(), ")");
+          resultType, ")");
     refinements.emplace_back(refinement);
     ++resultIndex;
   }
--- a/stablehlo/dialect/Base.cpp
+++ b/stablehlo/dialect/Base.cpp
@@ -18,6 +18,7 @@ limitations under the License.
 
 #include <optional>
 
+#include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/TypeSwitch.h"
 #include "mlir/Dialect/Quant/QuantTypes.h"
 #include "mlir/Dialect/Shape/IR/Shape.h"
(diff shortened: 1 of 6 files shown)


stablehlo/transforms/StablehloCanonicalizeDynamism.cpp
```diff
@@ -34,6 +35,54 @@ namespace stablehlo {
...
+      newAttrs.push_back(attr);
+    }
+
+    // Discard the operands that correspond to indices_of_shape_operands.
+    // Similarly, we rely on the verification logic implemented in
+    // getShapeRefinements to make sure that indices_of_shape_operands is
+    // consistent with these operands.
+    SmallVector<Value> newOperands;
+    auto resultIndex = 0;
+    for (auto& operand : op->getOpOperands()) {
```
GleasonK, line 68: `llvm::enumerate`?

stablehlo/tests/stablehlo_canonicalize_dynamism.mlir
```diff
@@ -1,7 +1,97 @@
...
+  // CHECK: stablehlo.custom_call @foo(%arg0, %arg1)
+  %0 = stablehlo.custom_call @foo(%arg0, %arg1) {
+    indices_of_shape_operands = dense<[1]> : tensor<1xi64>
+  } : (tensor<4xf32>, tensor<2xi64>) -> tensor<1x?xf32>
+  return %0 : tensor<1x?xf32>
+}
+
+// -----
+
+func.func @custom_call_failure_dynamic_result_type(%arg0: tensor<4xf32>) -> tensor<1x?xf32> {
```
GleasonK, line 82: Any better name than `_failure_` for these? Or perhaps the other's should be called `_invalid_` and the pass failures are `_failure_`.

stablehlo/tests/stablehlo_canonicalize_dynamism.mlir
```diff
@@ -1,7 +1,97 @@
-// RUN: stablehlo-opt --stablehlo-canonicalize-dynamism %s | FileCheck %s
+// RUN: stablehlo-opt --stablehlo-canonicalize-dynamism --split-input-file --verify-diagnostics %s | FileCheck %s
 
-// CHECK-LABEL: func @canonicalize_dynamic_conv
-func.func @canonicalize_dynamic_conv(%arg0: tensor<100x26x26x32xf32>, %arg1: tensor<3x3x1x32xf32>) -> tensor<100x28x28x1xf32> {
+// CHECK-LABEL: func @custom_call_success
+func.func @custom_call_success(%arg0: tensor<4xf32>) -> (tensor<1x2xf32>, tensor<3x4xf32>) {
```
GleasonK, line 4: Would be good to see a few more positive tests to document what's ok under this API. A few thoughts for tests, if the behavior shouldn't be allowed, consider adding verifiers to `Base.cpp` for them:

- Repeating `indices_of_shape_operands = [1, 1]`, given that we had the recent bug in refinement with repeated return arguments.
- Indices of shape argument not last `custom_call @foo(%0, %1) {
    indices_of_shape_operands = dense<[0]> : tensor<2xi64>
  } : (tensor<2xi64>, tensor<4xf32>) -> (tensor<3x4xf32>)`
- Constant operand used in `indices_of_shape_operands` has negative value `[1, -1]`

GleasonK (approved):

stablehlo/dialect/Base.cpp
```diff
@@ -126,6 +127,7 @@ bool isCompatibleForHloTypeInference(ArrayRef<int64_t> shape1, Type tp2) {
 bool isCompatibleForHloTypeInference(Value shape1, Type tp2) {
   SmallVector<int64_t> shapeVec1;
   if (!succeeded(matchInts(shape1, shapeVec1))) return true;
+  if (llvm::any_of(shapeVec1, [&](int64_t x) { return x < 0; })) return false;
```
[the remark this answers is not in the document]
burmako: We need to check all negative numbers, so that the `RankedTensorType::get` below doesn't crash due to a verification failure (see [verifier code](https://github.com/llvm/llvm-project/blob/51f0b809ca76c3cb037ea1195253af48f9d212e2/mlir/lib/IR/BuiltinTypes.cpp#L322-L323)).

stablehlo/dialect/Base.cpp
```diff
@@ -399,5 +401,47 @@ LogicalResult inferMostSpecificTypeComponents(
   return success();
 }
 
+LogicalResult getShapeRefinements(
+    std::optional<Location> location, Operation* op,
+    SmallVector<ShapedTypeComponents>& refinements) {
+  auto indicesAttr = op->getAttr("indices_of_shape_operands")
+                         .dyn_cast_or_null<DenseIntElementsAttr>();
```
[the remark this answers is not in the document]
burmako: Yeah, will keep DenseElementsAttr, to align with other similar attributes in StableHLO.

stablehlo/transforms/StablehloCanonicalizeDynamism.cpp
burmako: getOpOperands already includes an equivalent of llvm::enumerate, since its results have `getOperandNumber`. If you meant to ask whether we can avoid having the `resultIndex` variable, I don't think we can because this variable is not incremented on every loop iteration.

stablehlo/tests/stablehlo_canonicalize_dynamism.mlir
burmako: Both this comment and the comment above are good ideas. Will address them in a child PR which refactors the tests for this pass.

PR #1384, opened 2023-04-09
Harden --stablehlo-canonicalize-dynamism

burmako:
NOTE: This pull request is based on https://github.com/openxla/stablehlo/pull/1383.

When working on improvements for indices_of_shape_operands, I noticed
that --stablehlo-canonicalize-dynamism canonicalized some operations
that it shouldn't have.

I figured that I should add negative tests for all failure cases in
StablehloCanonicalizeDynamismPass, and this uncovered several issues
in the logic of the pass:
  * DynamicBroadcastInDimOp::output_dimensions,
    DynamicIotaOp::output_shape, DynamicReshapeOp::output_shape were
    discarded even if they weren't constants.
  * DynamicGatherOp/DynamicPadOp/RealDynamicSliceOp canonicalizations
    were using type inference to create static versions of the ops,
    which means that they could change the original result type,
    potentially producing invalid programs if someone downstream was
    relying on this exact result type.

This pull request provides fixes to those issues and comprehensively
covers all notifyMatchFailure cases in
StablehloCanonicalizeDynamismPass to make sure that this doesn't happen
again.


stablehlo/transforms/StablehloCanonicalizeDynamism.cpp
```diff
@@ -181,6 +195,9 @@ struct CanonicalizeDynamicReshapeOpPattern
                                 PatternRewriter& rewriter) const override {
     // This pattern ignores and discards the output_shape operand. We rely on
     // the verifier to make sure that its value is consistent with result type.
+    SmallVector<int64_t> outputShape;
+    if (!succeeded(hlo::matchInts(op.getOutputShape(), outputShape)))
```
GleasonK, line 199: If this pattern of unused vector for `matchInts` grows in usage (3 times in this PR), we could consider making a `matchInts` overload that takes a single argument and returns a LogicalResult, that way this vector doesn't have to stick around for the lifetime of this function call

GleasonK (approved):
LGTM. Hopefully this pass isn't used in production yet so these bugs don't have a large impact? This would likely be something we could consider releasing a patch for after 1.0.0 release

PR #1386, opened 2023-04-09
Audit casts to ShapedType and TensorType

burmako:
When working on adding support for tuple types to
RefineCustomCallOpPattern, I noticed that some logic in
StablehloRefineShapes.cpp assumes shaped types where we aren't
guaranteed to have shaped types and casts to ShapedType at will.

1) This led me to audit our uses of casts to ShapedType, and during that
audit I realized that many of these casts are not needed because in
many cases TableGen generates TypedValue accessors for stuff like
`::mlir::TypedValue<::mlir::TensorType> getLhs()`. This simplified
the code quite a bit.

2) Furthermore, I realized that in some places in our code we use
TensorType, so I audited casts to TensorType as well. To simplify
further refactorings, I figured that we could standardize on just one
kind of type between ShapedType and TensorType, and since ShapedType
is more general, I replaced all usages of TensorType with ShapedType.

3) Overall, after all the refactorings we still have some casts to
ShapedType in ChloOps.cpp, StablehloOps.cpp and TypeInference.cpp.
They are there for the following reasons:
  * Unlike ops, op adaptors don't produce TypedValue, and use
    regular Value instead. This means that there's a tradeoff between:
    a) TypeInference.cpp using Value and doing casts internally,
    b) TypeInference.cpp using TypedValue and forcing all users to
       do casts. I think a) is preferable.
  * Unlike singular fields, variadic fields don't produce TypedValue
    either, and instead use ValueRange, so stuff that comes from there
    also has to be cast.

4) The refactorings exposed the following latent bugs in our code which
could lead to crashes in the future. Even though these bugs cannot be
exercised at the moment, I figured it would be good to safeguard against
them:
  * inferReturnTypeComponentsFromOperands used to cast the result of
    inferReturnTypes to ShapedType. An oversight in user-level
    implementation of inferReturnTypes could lead to a crash.
  * ConstantOp::isCompatibleReturnTypes used to cast both the left-hand
    side and the right-hand side to TensorType. Same as the previous
    bullet.
  * refineReturnTypes used to cast types in op->getResultTypes()
    to ShapedType. Calling this functions on ops which can take
    tuples could lead to a crash.

### DIFF
--- a/stablehlo/dialect/AssemblyFormat.cpp
+++ b/stablehlo/dialect/AssemblyFormat.cpp
@@ -175,9 +175,9 @@ ParseResult parseVariadicOperandWithAttribute(
 // Operation Printers and Parsers
 //===----------------------------------------------------------------------===//
 
-void printComplexOpType(OpAsmPrinter& p, Operation* op, Type lhs, Type rhs,
-                        Type result) {
-  Type realType = createRealType(result.cast<TensorType>());
+void printComplexOpType(OpAsmPrinter& p, Operation* op, ShapedType lhs,
+                        ShapedType rhs, ShapedType result) {
+  Type realType = createRealType(result);
 
   if (lhs != realType || rhs != realType) {
     p.printFunctionalType(op);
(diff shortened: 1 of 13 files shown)


GleasonK (approved):
What is the difference between TensorType and ShapedType? Asking since TensorType sounds more like our spec.

PR #1391, opened 2023-04-11
Update IsFiniteOp test to use check op

ghpvnist:


sdasgup3 (approved):
Thanks for pointing it out.

PR #1392, opened 2023-04-11
Clarifying the semantics of has_side_effect in custom calls

### LINKED ISSUE #1365: Clarifying the semantics of `has_side_effect` in custom calls
### Request description

There are two kinds of side effects in systems:
a) Persistent side effects (e.g. print, fileIO) - reordering of the functions/dispatches/ops that perform print/fileIO is prohibited because the reordering is visible by observing the print/fileIO.
b) Non-persistent side effects (e.g. malloc, pthread_create) - reordering of the functions/dispatches/ops that perform malloc is not visible externally.

When custom calls have an attribute `has_side_effect`, it is not clear what side effects are meant here: has no persistent side effects, or has no persistent or non-persistent side effects. If this attribute is used to decide if reordering of custom calls are allowed or not, then it should mean only persistent side effects are disallowed.

The other alternative is to have a multi-value attribute specifying one of the three: a) no side effect, b) non-persistent side effect, c) persistent side effect. In this case, the compiler/runtime system will have more information to perform the appropriate optimizations.

sdasgup3:
fixes #1365.

Please refer to the issue for more information.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -81,7 +81,7 @@ completely numeric to simplify generation of StableHLO programs.
 
 ```ebnf
 Type         ::= ValueType | NonValueType
-ValueType    ::= TensorType | TokenType | TupleType
+ValueType    ::= TensorType | QuantizedTensorType | TokenType | TupleType
 NonValueType ::= ElementType | FunctionType | StringType
 ```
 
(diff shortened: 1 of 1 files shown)


docs/spec.md
```diff
@@ -2117,9 +2117,11 @@ tensor and produces a `result` tensor.
 #### Semantics
 
 Encapsulates an implementation-defined operation `call_target_name` that takes
-`inputs` and `called_computations` and produces `results`. `has_side_effect`,
-`backend_config` and `api_version` may be used to provide additional
-implementation-defined metadata.
+`inputs` and `called_computations` and produces `results`. The metadata
+`has_side_effect` is used to prevent optimizations like speculation (e.g
+loop-invariant code motion) and dead-code elimination. `backend_config`
```
burmako, line 2122: I'm not sure I understand from this description what program transformations are blocked by `has_side_effects`.

For comparison, [the spec of OptimizationBarrierOp](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#optimization_barrier) is very clear - program transformations cannot move operations across the barrier. So, all transformations that do this are illegal, and all transformations that don't do this are legal, as far as this operation is concerned.

Can we have a similarly clear criterion here?

burmako (changes requested):

sdasgup3:
I am planning to switch to a different branch to host the PR. Hence I closed this one and will publish another PR (considering the review comments) from a different branch.

PR #1395, opened 2023-04-12
Streamline the pseudocode describing SignOp

burmako:
Inspired by #1382, this PR: 1) reduces nesting, 2) drops the final ifs from the is_integer and is_float code blocks, 3) makes it clear that the NaN case for complex numbers returns NaN for both real and imaginary components.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -5028,7 +5028,7 @@ def sign(x):
     return 1.0
   elif is_complex(x):
     if x.real is NaN or x.imag is NaN: return (NaN, NaN)
-    return divide(x, abs(x))
+    return divide(x, convert(abs(x), type(x)))
 ```
 
 #### Inputs
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -5022,9 +5022,9 @@ def sign(x):
     return 1
   elif is_float(x):
     if x is NaN: return NaN
-    if compare(x, 0.0, LT, FLOAT): return -1.0
     if compare(x, -0.0, EQ, FLOAT): return -0.0
     if compare(x, +0.0, EQ, FLOAT): return +0.0
+    if compare(x, 0.0, LT, FLOAT): return -1.0
     return 1.0
   elif is_complex(x):
     if x.real is NaN or x.imag is NaN: return (NaN, NaN)
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -5027,7 +5027,7 @@ def sign(x):
     if compare(x, +0.0, EQ, FLOAT): return +0.0
     return 1.0
   elif is_complex(x):
-    if x.real is NaN or x.imag is NaN: return NaN
+    if x.real is NaN or x.imag is NaN: return (NaN, NaN)
     return divide(x, abs(x))
 ```
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -5028,6 +5028,7 @@ def sign(x):
     return 1.0
   elif is_complex(x):
     if x.real is NaN or x.imag is NaN: return (NaN, NaN)
+    if compare(x, (0.0, 0.0), EQ, FLOAT): return (0.0, 0.0)
     return divide(x, convert(abs(x), type(x)))
 ```
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -5019,20 +5019,16 @@ def sign(x):
   if is_integer(x):
     if compare(x, 0, LT, SIGNED): return -1
     if compare(x, 0, EQ, SIGNED): return 0
-    if compare(x, 0, GT, SIGNED): return 1
+    return 1
   elif is_float(x):
-    if x is NaN:
-      return NaN
-    else:
-      if compare(x, 0.0, LT, FLOAT): return -1.0
-      if compare(x, -0.0, EQ, FLOAT): return -0.0
-      if compare(x, +0.0, EQ, FLOAT): return +0.0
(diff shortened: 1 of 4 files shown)


docs/spec.md
```diff
@@ -4935,20 +4935,16 @@ def sign(x):
...
+    if compare(x, +0.0, EQ, FLOAT): return +0.0
+    if compare(x, 0.0, LT, FLOAT): return -1.0
+    return 1.0
   elif is_complex(x):
-    if x.real is NaN or x.imag is NaN:
-      return NaN
-    else:
-      return divide(x, abs(x))
+    if x.real is NaN or x.imag is NaN: return (NaN, NaN)
+    return divide(x, abs(x))
```
ghpvnist, line 4947: abs(x) returns a float, so how about explicitly writing:
```
return divide(x, (abs(x).real, abs(x).imag))
```

ghpvnist (approved):
One comment and lgtm.

docs/spec.md
burmako: "abs(x) returns a float". Good point, we've also discussed this in #1382. You probably meant `divide(x, (abs(x), 0))`, right? Anyway, I used convert instead to hopefully make this more explicit.

"We should also have a separate case for handling zeros (regardless of sign should return positive zero)". Is this behavior specced somewhere? Here are the data points that seem relevant:
  * [elemental_ir_emitter.py](https://github.com/tensorflow/tensorflow/blob/f6ca332b93cc5584dd09ca991865a6a7091f3eff/tensorflow/compiler/xla/service/elemental_ir_emitter.cc#L1054-L1064) does what's proposed in the comment above.
  * [lax.py](https://github.com/google/jax/blob/d9383fc80dc739443fd10365c85c7e781e370b01/jax/_src/lax/lax.py#L234-L235) only says `x / |x|`.
  * Surprisingly, [numpy.sign](https://numpy.org/doc/stable/reference/generated/numpy.sign.html) does something quite different: `sign(x.real) + 0j if x.real != 0 else sign(x.imag) + 0j`, so we shouldn't probably index too much on that.

ghpvnist: Ah yes, I meant `(abs(x), 0)`. Handling zeros is a specific behavior in `elemental_ir_emitter.py`. And though the jax doc you linked seem to suggest otherwise, as mentioned in https://github.com/openxla/stablehlo/pull/1382#discussion_r1165960432, lowering to hlo in [lax.py](https://github.com/google/jax/blob/d9383fc80dc739443fd10365c85c7e781e370b01/jax/_src/lax/lax.py#L1714-L1724) where the `sign_p` comes from the [line](https://github.com/google/jax/blob/d9383fc80dc739443fd10365c85c7e781e370b01/jax/_src/lax/lax.py#L237) under the doc also follows this behavior.

docs/spec.md
```diff
@@ -4935,20 +4935,16 @@ def sign(x):
...
+    if compare(x, -0.0, EQ, FLOAT): return -0.0
+    if compare(x, +0.0, EQ, FLOAT): return +0.0
+    if compare(x, 0.0, LT, FLOAT): return -1.0
+    return 1.0
   elif is_complex(x):
-    if x.real is NaN or x.imag is NaN:
-      return NaN
-    else:
-      return divide(x, abs(x))
+    if x.real is NaN or x.imag is NaN: return (NaN, NaN)
```
burmako, line 4946: Based on the outcome of the conversation about how `(0.0, 0.0)` should be handled by SignOp, we will want to enable and/or change `stablehlo/testdata/sign_special_0_dtype_complex64.mlir`.

docs/spec.md
burmako: Upon further reflection, I am convinced now that it is a good idea to specialcase `(0.0, 0.0)` in the specification, consistently with how `0.0` is specialcased as well. I've updated the PR accordingly.

docs/spec.md
burmako: Will wait for #1382 to be merged and will then update this PR to include interpreter changes to accommodate `(0.0, 0.0)` and enable this test.

burmako:
Getting spurious CMake build errors across multiple pull requests:

```
Ign:1 http://azure.archive.ubuntu.com/ubuntu jammy/universe amd64 lld amd64 1:14.0-55~exp2
Ign:1 http://azure.archive.ubuntu.com/ubuntu jammy/universe amd64 lld amd64 1:14.0-55~exp2
Ign:1 http://azure.archive.ubuntu.com/ubuntu jammy/universe amd64 lld amd64 1:14.0-55~exp2
Err:1 http://azure.archive.ubuntu.com/ubuntu jammy/universe amd64 lld amd64 1:14.0-55~exp2
  Could not connect to azure.archive.ubuntu.com:80 (52.147.219.192), connection timed out
E: Failed to fetch [http://azure.archive.ubuntu.com/ubuntu/pool/universe/l/llvm-defaults/lld_14.0-55%7eexp2_amd64.deb](http://azure.archive.ubuntu.com/ubuntu/pool/universe/l/llvm-defaults/lld_14.0-55~exp2_amd64.deb)  Could not connect to azure.archive.ubuntu.com:80 (52.147.219.192), connection timed out
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```

I've just run a clean CMake build locally, and it seems to work, so I think that this is safe to merge. I'll rerun the build again on my machine just to be sure and will force merge this PR once the local build succeeds.

PR #1398, opened 2023-04-12
Unify APIs in Element.h

### LINKED ISSUE #1397: Unify APIs in Element.h
Majority of the functions in this file returns back an `Element` type, but recently merged PR ([isFiniteOp](#1385)) and some existing functions (i.e. functions overriding operators) return `bool`. The benefits of unifying the APIs to return `Element` type consistently is bigger than keeping them currently as is:

Pros:
* Simplify `CompareOp` implementation (currently wraps booleans back into `Element` type).
* Unifies the return type across similar meta functions.

Cons:
* Need to unwrap them if needed in the future, but no such cases as of now.
* Avoids explicit creation of `i1` types from the builder (none in the codebase as of now but don't anticipate this to be prominent).

ghpvnist:
closes #1397

### DIFF
--- a/stablehlo/reference/Element.cpp
+++ b/stablehlo/reference/Element.cpp
@@ -264,8 +264,8 @@ std::complex<APFloat> Element::getComplexValue() const {
 }
 
 Element Element::operator!() const {
-  return Element(mlir::Builder(this->getType().getContext()).getI1Type(),
-                 !this->getBooleanValue());
+  return Element(IntegerType::get(getType().getContext(), 1),
+                 !getBooleanValue());
 }
 
 Element Element::operator!=(const Element &other) const {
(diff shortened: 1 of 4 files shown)


stablehlo/reference/Element.cpp
```diff
@@ -263,7 +263,12 @@ std::complex<APFloat> Element::getComplexValue() const {
   return std::complex<APFloat>(floatPair.first, floatPair.second);
 }
 
-bool Element::operator!=(const Element &other) const {
+Element Element::operator!() const {
+  return Element(mlir::Builder(this->getType().getContext()).getI1Type(),
+                 !this->getBooleanValue());
```
burmako, line 268: Let's use `IntegerType::get(getType().getContext(), 1)` directly instead, to avoid creating builders?

burmako, line 268: Can we drop `this->` here and below?

burmako (changes requested):

burmako (approved):

PR #1400, opened 2023-04-12
Add interpreter for Atan2Op

### LINKED ISSUE #1097: Add  interpreter for Atan2Op

ghpvnist:
```
(I1) `lhs` is a tensor of floating-point or complex type.
(I2) `rhs` is a tensor of floating-point or complex type.
(C1) `lhs`, `rhs`, and `result` have the same type.
```
I1, I2, and C1 are covered by the ODS, so no additional tests are added.

closes #1097

### DIFF
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -698,7 +698,8 @@ def StableHLO_AddOp : StableHLO_BinaryElementwiseOp<"add",
 }
 
 def StableHLO_Atan2Op : StableHLO_BinaryElementwiseOp<"atan2",
-      [Pure, HLO_CompatibleOperandsAndResultType], HLO_FpOrComplexTensor> {
+      [Pure, HLO_CompatibleOperandsAndResultType /*atan2_c1*/],
+      HLO_FpOrComplexTensor /*atan2_i1, atan2_i2*/> { /*atan2_c1*/
   let summary = "Atan2 operation";
   let description = [{
     Performs element-wise atan2 operation on `lhs` and `rhs` tensor and produces
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -710,7 +710,7 @@ def StableHLO_Atan2Op : StableHLO_BinaryElementwiseOp<"atan2",
 
     Example:
     ```mlir
-    %result = stablehlo.atan2 %lhs, %rhs : tensor<3xf32>
+    %result = stablehlo.atan2 %lhs, %rhs : tensor<3xf64>
     ```
   }];
 }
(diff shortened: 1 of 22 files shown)


stablehlo/reference/Element.cpp
```diff
@@ -585,6 +585,22 @@ Element areApproximatelyEqual(const Element &e1, const Element &e2) {
                                      debugString(type).c_str()));
 }
 
+Element atan2(const Element &e1, const Element &e2) {
+  auto type = e1.getType();
+  if (isSupportedFloatType(e1.getType()))
+    return Element(type, std::atan2(e1.getFloatValue().convertToDouble(),
+                                    e2.getFloatValue().convertToDouble()));
+
+  if (isSupportedComplexType(type)) {
```
sdasgup3, line 594: In the spec we mentioned that atan2 for complex computes "complex atan2". Is "complex atan2" and the associated formula ([as mentioned in HLO evaluator](https://github.com/tensorflow/tensorflow/blob/487b79112fa1d8d9afd70a6785a06fa179109cd4/tensorflow/compiler/xla/hlo/evaluator/hlo_evaluator_typed_visitor.h#L438)) a standard textbook thing?

stablehlo/dialect/StablehloOps.td
```diff
@@ -698,7 +698,8 @@ def StableHLO_AddOp : StableHLO_BinaryElementwiseOp<"add",
 }
 
 def StableHLO_Atan2Op : StableHLO_BinaryElementwiseOp<"atan2",
-      [Pure, HLO_CompatibleOperandsAndResultType], HLO_FpOrComplexTensor> {
+      [Pure, HLO_CompatibleOperandsAndResultType /*atan2_c1*/],
+      HLO_FpOrComplexTensor /*atan2_i1, atan2_i2*/> { /*atan2_c1*/
```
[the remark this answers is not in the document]
ghpvnist: We should be having this according to the updated guidelines, but AddOp and various other ops are not up to date with the rest. #1002 should track this.

stablehlo/tests/interpret_atan2.mlir
```diff
@@ -0,0 +1,21 @@
...
+  func.return
+}
+
+// -----
+
+func.func @atan2_op_test_c128() {
+  %0 = stablehlo.constant dense<[(0.0, 0.0), (1.0, 0.0), (-1.0, 0.0)]> : tensor<3xcomplex<f64>>
+  %1 = stablehlo.constant dense<[(0.0, 0.0), (0.0, 0.0), (0.0, 0.0)]> : tensor<3xcomplex<f64>>
+  %2 = stablehlo.atan2 %0, %1 : tensor<3xcomplex<f64>>
+  check.expect_almost_eq_const %2, dense<[(0x7FF8000000000000, 0x7FF8000000000000),
```
sdasgup3, line 17: Should we avoid the NaNs here? How about just keeping the last input.

sdasgup3 (approved):
LGTM with one comment.

Issue #1407, opened 2023-04-14
Represent unknown scales and zero-points


sdasgup3:
The goal of the ticket is to track the support of unknown scales and zero-points. This is required to represent the scales and zero-points, in StableHLO graph, calculated on the fly by the training program while quantizing the activations.

Please refer to relevant discussion [here](https://github.com/openxla/stablehlo/pull/1352#discussion_r1164976969).

lgeiger:
Support for unknown scales would be incredibly useful for quantization aware training (QAT). What is the current status on this?

Maybe a bit of context, our use case is focused on QAT targeting a fully int8 quantized TFLite inference model. Currently we're relying on [`tf.quantization.fake_quant_with_min_max_vars`](https://www.tensorflow.org/api_docs/python/tf/quantization/fake_quant_with_min_max_vars) on the training side. As far as I'm aware this is the only supported way at the moment but it would be great to be able to directly output StableHLO from jax or maybe even PyTorch for greater flexibility and better usability.

@abattery mentioned that he's interested in QAT as well and the [odml team seems to have a way to inject `stablehlo.uniform_quantize` ops](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/mlir/lite/stablehlo/tests/compose-uniform-quantized-type.mlir) but I'm not sure what the latest status on these efforts are.

@sdasgup3 do you know whether there is interest in supporting QAT workflows via StableHLO from frontends like jax or PyTorch? I'd be very interested in getting involved and contributing towards any consolidated effort here since QAT is much easier to deal with from an ML training standpoint compared to post training quantization which always has the potential to introduce accuracy degradations if not done carefully.

sdasgup3:
@lgeiger Thanks for bringing this up and providing details about your case. This is on our radar for sometime, but did get a sufficiently motivating use-case (and bandwidth) to initiate work on it. 

>> do you know whether there is interest in supporting QAT workflows via StableHLO from frontends like jax or PyTorch? I'd be very interested in getting involved and contributing towards any consolidated effort here 

Much appreciated on your willingness to contribute! I will get back to you on you the question.

Issue #1406, opened 2023-04-14
Finalize the semantics of `storage_min`, `storage_max`


sdasgup3:
The goal of this ticket is to investigate whether or not the quantized type specification should limit the allowed values of `storage_min` and `storage_max`. Currently, the allowed values are limited to `[min_value(storage_type, max_value(storage_type]`, but this limit may not be followed by some use cases.

Please refer to [ref](https://github.com/openxla/stablehlo/pull/1352#discussion_r1163128362) to follow the discussion.

**update**
The [current spec](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#types) expects default values of `min_value(storage_type)` and `max_value(storage_type)` for `storage_min` and `storage_max` resp., if they are not explicitly provided. Moreover, the specification documents the following general constraint on `storage_min` and `storage_max`

```
`min_value(storage_type) <= storage_min < storage_max <= max_value(storage_type)`.
```

sdasgup3:
From the POV of the semantics of an op producing quantized results, the relevance `storage_min/storage_max` values are  to clamp the result(s) . The [current spec](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#types) expects default values of `min_value(storage_type)` and `max_value(storage_type)` for `storage_min` and `storage_max` resp., if they are not explicitly provided. Moreover, the specification documents the following general constraint on `storage_min` and `storage_max`

```
`min_value(storage_type) <= storage_min < storage_max <= max_value(storage_type)`.
```

My personal opinion: The above constraint seems sufficiently general to accommodate the existing implementations (e.g. TFLite) as well as allow  the adoption of new or experimental quantization schemes.  The op semantics should not add additional constraints unless there is a strong reason to do that. 

Having said that, let me re-propose a draft PR for quantized add. As some of you might remember that, at some point, we [decided](https://github.com/openxla/stablehlo/pull/1352#discussion_r1166196224) to drop the introduction of its specification mainly because we were unsure about the fate of https://github.com/openxla/stablehlo/issues/1406.

The proposal will be almost same as before except that it does not have any additional constraint imposed by the op's semantics on `storage_min` or `storage_max`.

Please let me know your feedback and/or feel free to review the upcoming PR.

jingpu:
Here are some pointers regarding the "narrow_range" INT8 feature in TFLite.

Op specs: https://www.tensorflow.org/lite/performance/quantization_spec
```
CONV_2D
  Input 0:
[...]

Issue #1404, opened 2023-04-14
Quantized type based on integer multiplier/shift


sdasgup3:
Various customers of StableHLO quantization have communicated their interest in investigating integer-based quantized types. Specifically, to represent the floating-point quantization scale using integer multiplier/shift, and in allowing a defined conversion between the two formats. This ticket tracks progress on this issue and its resolution.

Please refer to [ref](https://github.com/openxla/stablehlo/pull/1352#discussion_r1149974605) to follow the discussion.

sdasgup3:
Hello folks, I am writing to inform you that I will be planning for a proposal in next week. Thanks for your time and consideration!

sdasgup3:
Per https://github.com/openxla/stablehlo/pull/1413#discussion_r1186088789, it would be useful in having a reference conversion from integer [multiplier.shift] to integer rescale operation. Let us track that as part of the current PR.

sdasgup3:
Per https://github.com/openxla/stablehlo/pull/1647#discussion_r1266995360, the epsilon parameter of `batch_norm_grad`, `batch_norm_training`, and `batch_norm_inference` is floating point which may prevent integer only computation. Explore if this parameter can be quantized.

sdasgup3:
We are planning to prioritize this issue in H2'24.

Issue #1401, opened 2023-04-13
Tighten the notation that we're using in the interpreter


burmako:
Similarly to #289, I think there are some notational improvements that we can make in the interpreter to make the implementation easier to follow. Let's keep this ticket open for now and then address all that together once we have an initial implementation for all or most of the ops.

burmako:
Something that came to mind when I was reviewing #1371 is can we reach the same level of conciseness of the batch_norm_inference (and friends) implementation as we have in the spec. These are the things that are currently more verbose in the C++ implementation:
  1) `evalFooOp` could be renamed to just `foo`.
  2) `evalFooOp` functions take a result type which could be easily computed for many of these functions like add, sqrt, subtract, etc. Not sure if this means that we should stop passing result types to these functions and keep passing result types to some other functions, or maybe we should have helper overloads without result types where appropriate - something to think about in the long term.

burmako:
Might be interesting to achieve full parity in how metafunctions are represented in code, e.g. define `rank(const Tensor &tensor) -> int64_t`, `dim(const Tensor &tensor, int64_t i) -> int64_t`, etc.

PR #1413, opened 2023-04-15
Specification for uniform quantize dot-general op

sdasgup3:
The PR proposes the spec for quantized dot-general op along with the specifications for a few other ops on which the dot-general depends on, for example, `slice`, `transpose`, and `reshape`.

Given `fp = tensor with floating-point type and q = tensor with uniformed quantized type`, the PR covers the semantics of 
(1) Static range quantized `dot_general` op `dot_general(q, q)`, and 
(2) Hrbrid quantized `dot_general` op `dot_general(fp, q)`: Currently, this version of the op only supports dynamic range quantization, where the on-the-fly quantization of `lhs` is fused in the op-semantics. IMO, once we support https://github.com/openxla/stablehlo/issues/1407, the quantization logic can be un-fused and made explicit in the MLIR graph (cc @sngyhan).

Side note:
The plan forward is to propose a PR for convolution op in very near future. I realized that the spec for convolution  depends on dot-general and a split might help the review process.

Please let me know your review feedback.


docs/spec.md
```diff
@@ -2263,23 +2263,58 @@ Computes dot products between slices of `lhs` and slices of `rhs` and produces a
...
-  body=lambda x, y: add(x, y))`.
+* For `is_non_quantized_tensor(lhs) and is_non_quantized_tensor(rhs)`:
+  * `dot_product = reduce(
+    inputs=[multiply(reshaped_lhs_slice, reshaped_rhs_slice)],
+    init_values=[0],
+    dimensions=[0, ..., size(lhs_contracting_dimensions) - 1],
+    body=lambda x, y: add(x, y))`.
+* For `is_quantized_tensor(lhs) and is_quantized_tensor(rhs)`:
+  * `float_dot_product = reduce(
+    inputs=[multiply((reshaped_lhs_slice - zero_point(reshaped_lhs_slice)) * scale(reshaped_lhs_slice),
```
loganchien, line 2304: Can we lift the `* scale(reshaped_{lhs,rhs}_slice)` to the second step?

```
`rounded_dot_product = round_nearest_even(float_dot_product * (scale(reshaped_lhs_slice) * scale(reshape_rhs_slice) / scale(result)))`.
```

docs/spec.md
```diff
@@ -2263,23 +2263,58 @@ Computes dot products between slices of `lhs` and slices of `rhs` and produces a
...
+                     (reshaped_rhs_slice - zero_point(reshaped_rhs_slice)) * scale(reshaped_rhs_slice)],
+    init_values=[0],
+    dimensions=[0, ..., size(lhs_contracting_dimensions) - 1],
+    body=lambda x, y: add(x, y))`.
+  * `rounded_dot_product = round_nearest_even(float_dot_product / scale(result))`.
+  * `dot_product = clamp(storage_min(result), rounded_dot_product + zero_point(result), storage_max(result))`.
+* For `is_non_quantized_tensor(lhs) and is_quantized_tensor(rhs)`:
+  * `dot_product = reduce(
+    inputs=[multiply((reshaped_lhs_slice - zero_point(reshaped_lhs_slice)) * scale(reshaped_lhs_slice),
```
loganchien, line 2313: If I read the commit message properly, you are specifying dynamic range quantization here, which quantize lhs, perform integer dot, and then scale it back to floating point output.  So, similar to static quantization `dot_general`, lift the scale multiplier to a separate step?

docs/spec.md
```diff
@@ -2263,23 +2263,58 @@ Computes dot products between slices of `lhs` and slices of `rhs` and produces a
 More formally, `result[result_index] = dot_product`, where:
 
 <!-- markdownlint-disable line-length -->
-* `lhs_result_dimensions = [d for d in axes(lhs) and d not in lhs_batching_dimensions and d not in lhs_contracting_dimensions]`.
+* `If lhs uses non-quantized tensor and rhs uses quantized tensor:`
+
+  ```python
+   for each batch_size in [0, dim(lhs, lhs_batching_dimensions[0])]:
```
loganchien, line 2269: Is this `batch_size` or `batch_idx`?

Also, what are you iterating for?  Do you expect `batch_size` to be `0` and then `dim(lhs, lhs_batching_dimensions[0])`?  Or do you expect `batch_size` to be `0`, `1`, `2`, ..., `dim(lhs, lhs_batching_dimensions[0]) - 1`?  If it is later, please just write python code:

```
  for batch_size in range(0, dim(lhs_batching_dimensions[0]):
    ...
```

loganchien, line 2269: Why we only doing quantization on `lhs_batching_dimensions[0]`?

If this is a generalization of `tfl.fully_connected`, we should quantize on all batches (i.e. `lhs_batching_dimensions[0] * lhs_batching_dimensions[1] * ... * lhs_batching_dimensions[n - 1]`).

abattery:
> (2) Hybrid quantized dot_general op dot_general(fp, q): Currently, this version of the op only supports dynamic range quantization, where the on-the-fly quantization of lhs is fused in the op-semantics. IMO, once we support https://github.com/openxla/stablehlo/issues/1407, the quantization logic can be un-fused and made explicit in the MLIR graph (cc @sngyhan).

I have a slightly different view on this part. Basically, I wouldn't see that we are going to force to implement specific kernel implementations for both static range quantization and hybrid quantization to ML hardware makers. It is truly up to ML hardware makers to decide their implementation once the kernel satisfies the operator contracts based on their HW capabilities. For example, even for the float matrix multiplication operator, we usually don't enforce how to treat the input tensors in the computations. Especially, for hybrid kernel cases, they can be implemented with either weight only quantization or dynamic range quantization up to the hardware implementer.

docs/spec.md
sdasgup3: Agreed. PLAT at the revised version.

docs/spec.md
sdasgup3: Sure, that makes sense. IIUC, the rationale here is to highlight the fact that quantized operands enables integer-domain computation.

docs/spec.md
sdasgup3: >> Or do you expect batch_size to be 0, 1, 2, ..., dim(lhs, lhs_batching_dimensions[0]) - 1
Yes, that is what I expect. 

I will do the make the suggested in the next commit.

rewu93:
> > (2) Hybrid quantized dot_general op dot_general(fp, q): Currently, this version of the op only supports dynamic range quantization, where the on-the-fly quantization of lhs is fused in the op-semantics. IMO, once we support #1407, the quantization logic can be un-fused and made explicit in the MLIR graph (cc @sngyhan).
> 
> I have a slightly different view on this part. Basically, I wouldn't see that we are going to force to implement specific kernel implementations for both static range quantization and hybrid quantization to ML hardware makers. It is truly up to ML hardware makers to decide their implementations once those kernels satisfies the operator contracts based on their HW capabilities. For example, even for the float matrix multiplication operator, we usually don't enforce how to treat the input tensors in the computations and enforce using the certain algorithm (e.g, iterative, divide-and-conquer, strassen and so on). Especially, for hybrid kernel cases, they can be implemented with either weight only quantization or dynamic range quantization up to the hardware implementers.

Agree. TFLite is also considering adding weight only quantization (i.e., dequantize weight) for hybrid kernels.

docs/spec.md
```diff
@@ -2288,23 +2288,59 @@ Computes dot products between slices of `lhs` and slices of `rhs` and produces a
...
+      inputs=[multiply((reshaped_lhs_slice - zero_point(reshaped_lhs_slice)),
+                       (reshaped_rhs_slice - zero_point(reshaped_rhs_slice))],
+      init_values=[0],
+      dimensions=[0, ..., size(lhs_contracting_dimensions) - 1],
+      body=lambda x, y: add(x, y))`.
+  * `rounded_dot_product = round_nearest_even(integer_dot_product * (scale(reshaped_lhs_slice) * scale(reshape_rhs_slice) / scale(result)))`.
+  * `dot_product = clamp(storage_min(result), rounded_dot_product + zero_point(result), storage_max(result))`.
+* For `is_non_quantized_tensor(lhs) and is_quantized_tensor(rhs)`:
+  * `integer_dot_product = reduce(
+      inputs=[multiply((reshaped_lhs_slice - zero_point(reshaped_lhs_slice)) * scale(reshaped_lhs_slice),
```
rewu93, line 2338: Not sure if I understand correctly here. Do we need to calculate scale and zero point for lhs? isn't it `is_non_quantized_tensor(lhs)`?

rewu93: Also, `(reshaped_lhs_slice - zero_point(reshaped_lhs_slice)) * scale(reshaped_lhs_slice` dequantize the integer back to float so it may be inappropriate to call it  `integer_dot_product `

rewu93: what about `is_quantized_tensor(lhs) and is_non_quantized_tensor(rhs)`?

sdasgup3: Hi @rewu93 

>> it may be inappropriate to call it integer_dot_product 

Will be fixed in next commit. 


>> what about is_quantized_tensor(lhs) and is_non_quantized_tensor(rhs)?

We have  constraint to disallow that. As per (C14),

```
 If the operation uses quantized tensors:
  * (C14) Either `is_quantized_tensor(lhs) and is_quantized_tensor(rhs)` or  `is_non_quantized_tensor(lhs) and is_quantized_tensor(rhs)`.
```

docs/spec.md
```diff
@@ -2322,49 +2358,70 @@ planning to address this in
...
 #### Constraints
 
-* (C1) `lhs` and `rhs` have the same element type.
-* (C2) size(`lhs_batching_dimensions`) $=$ size(`rhs_batching_dimensions`).
-* (C3) size(`lhs_contracting_dimensions`) $=$
+<!-- markdownlint-disable line-length -->
+* (C1) size(`lhs_batching_dimensions`) $=$ size(`rhs_batching_dimensions`).
+* (C2) size(`lhs_contracting_dimensions`) $=$
   size(`rhs_contracting_dimensions`).
 * (C4) `lhs_batching_dimensions` and `lhs_contracting_dimensions` combined are
```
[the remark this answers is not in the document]
sdasgup3: Sorry for the confusion. We actually meant the following by "combining"

```
Given
  result = concatenate(lhs_batching_dimensions, lhs_contracting_dimensions, dimension = 0),

result will have unique elements. 
```  

I believe all such rough edges will be revisited as part of https://github.com/openxla/stablehlo/issues/1401.

docs/spec.md
```diff
@@ -2322,49 +2358,70 @@ planning to address this in
...
+* (C4) `rhs_batching_dimensions` and `rhs_contracting_dimensions` combined are
   unique.
-* (C6) 0 $\le$ `lhs_batching_dimensions[i]` $\lt$ rank(`lhs`) for all `i`
+* (C5) 0 $\le$ `lhs_batching_dimensions[i]` $\lt$ rank(`lhs`) for all `i`
   $\in$ [0, size(`lhs_batching_dimensions`)).
-* (C7) 0 $\le$ `lhs_contracting_dimensions[i]` $\lt$ rank(`lhs`) for all `i`
+* (C6) 0 $\le$ `lhs_contracting_dimensions[i]` $\lt$ rank(`lhs`) for all `i`
   $\in$ [0, size(`lhs_contracting_dimensions`)).
-* (C8) 0 $\le$ `rhs_batching_dimensions[d]` $\lt$ rank(`rhs`) for all `i`
+* (C7) 0 $\le$ `rhs_batching_dimensions[d]` $\lt$ rank(`rhs`) for all `i`
```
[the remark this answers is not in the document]
sdasgup3: Good catch! It should be `rhs_batching_dimensions[i]`.

I will be fixing in in the next commit.

docs/spec.md
```diff
@@ -2322,49 +2358,70 @@ planning to address this in
...
   $\in$ [0, size(`rhs_batching_dimensions`)).
-* (C9) 0 $\le$ `rhs_contracting_dimensions[d]` $\lt$ rank(`rhs`) for all `i`
+* (C8) 0 $\le$ `rhs_contracting_dimensions[d]` $\lt$ rank(`rhs`) for all `i`
```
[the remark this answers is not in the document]
sdasgup3: same as above. Thanks!

docs/spec.md
```diff
@@ -2322,49 +2358,70 @@ planning to address this in
...
-* (C13) shape(`result`) $=$ dim(`lhs`, `lhs_batching_dimensions`) +
+* (C11) size(`precision_config`) $=$ 2.
+* (C12) shape(`result`) $=$ dim(`lhs`, `lhs_batching_dimensions`) +
   dim(`lhs`, `lhs_result_dimensions`) + dim(`rhs`, `rhs_result_dimensions`).
+* If the operation uses non-quantized tensors:
+  * (C13) element_type(`lhs`) $=$ element_type(`rhs`).
+* If the operation uses quantized tensors:
+  * (C14) Either `is_quantized_tensor(lhs) and is_quantized_tensor(rhs)` or  `is_non_quantized_tensor(lhs) and is_quantized_tensor(rhs)`.
+  * For `is_quantized_tensor(lhs) and is_quantized_tensor(rhs)`:
+    * (C15) `storage_type(lhs) =  storage_type(rhs) = storage_type(result)`.
```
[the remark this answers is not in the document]
sdasgup3: That is a good point! 

In fact we have a [ticket](https://github.com/openxla/stablehlo/issues/369) to address  not only  the dot_general op, but [other ops](https://github.com/openxla/stablehlo/issues/369#issuecomment-1334859933) which allows mixed precision.

docs/spec.md
```diff
@@ -2322,49 +2358,70 @@ planning to address this in
...
-* (C11) dim(`lhs`, `lhs_contracting_dimensions[i]`) $=$
+* (C10) dim(`lhs`, `lhs_contracting_dimensions[i]`) $=$
   dim(`rhs`, `rhs_contracting_dimensions[i]`) for all `i` $\in$ [0,
   size(`lhs_contracting_dimensions`)).
-* (C12) size(`precision_config`) $=$ 2.
-* (C13) shape(`result`) $=$ dim(`lhs`, `lhs_batching_dimensions`) +
+* (C11) size(`precision_config`) $=$ 2.
+* (C12) shape(`result`) $=$ dim(`lhs`, `lhs_batching_dimensions`) +
   dim(`lhs`, `lhs_result_dimensions`) + dim(`rhs`, `rhs_result_dimensions`).
+* If the operation uses non-quantized tensors:
```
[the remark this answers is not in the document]
sdasgup3: >> you mean both lhs and rhs are non-quantized tensor here?

That is right.

docs/spec.md
rewu93: Just try to be as explicit as possible to avoid confusion. What I meant is that since the hybrid kernel is evaluated as DRQ, instead of 
* `is_non_quantized_tensor(lhs) and is_quantized_tensor(rhs)`:
  * `integer_dot_product = reduce(
      inputs=[multiply((reshaped_lhs_slice - zero_point(reshaped_lhs_slice)) * scale(reshaped_lhs_slice),(reshaped_rhs_slice - zero_point(reshaped_rhs_slice)) * scale(reshaped_rhs_slice)],
      init_values=[0],
  * `dot_product = integer_dot_product * scale(reshaped_lhs_slice) * scale(reshape_rhs_slice)`.

this should be 
* `is_non_quantized_tensor(lhs) and is_quantized_tensor(rhs)`:
  * `integer_dot_product = reduce(
      inputs=[multiply((reshaped_lhs_slice - zero_point(reshaped_lhs_slice)),(reshaped_rhs_slice - zero_point(reshaped_rhs_slice))],
      init_values=[0],
  * `dot_product = integer_dot_product * scale(reshaped_lhs_slice) * scale(reshape_rhs_slice)`.

@loganchien I suppose we don't need to round and clamp for the hybrid case here

sdasgup3: Thanks @rewu93 for the fix. This is exactly the fix I have in mind. Please take a look.

rewu93: LGTM

abattery: In my opinions, dynamic range quantization is somewhat unique, which is implemented only in TFLite CPU kernel in the world as far as I know (even all other TFLite variants (XNNPACK and so on) do not support dynamic range quantization kernels). IMO, in terms of adoptions, it is bit early to make dynamic range quantization kernels as a part of the language specification in StableHLO. As dynamic range quantization support in TFLite CPU kernel can be recognized as a legacy support, it seems like a fine choice to use custom ops to express dynamic range quantization for the time being. However, in the long run, there are some ideas too support dynamic range quantization in the graph level later. At the same time, this might be a marathon race to ask the ML hardware implementers to support dynamic range quantization kernels more effectively.

sdasgup3: Thanks @abattery for the elaborate feedback.

>> it seems like a fine choice to use custom ops for expressing dynamic range quantization kernels for the time being.

This seems like a reasonable choice. As you pointed out, `In the long run, there are some ideas to support dynamic range quantization in the graph level later.`, IMO, it might be reasonable to not include any hybrid op for now (which have the potential of getting deprecated in the long run) and instead use custom ops to express dynamic range quantization.

Anyways, I will be happy to understand any other opinion.

docs/spec.md
sdasgup3: Hi @loganchien 
Thanks for explaining in detail (offline) what exactly we need here.

As this change is related to the hybrid op (floating point `lhs` and quantized `rhs`) case, I propose that we hold on this change for a while until we  decide on whether the specification requires a hybrid op.
[ref](https://github.com/openxla/stablehlo/pull/1413#discussion_r1183127043).

docs/spec.md
```diff
@@ -2288,23 +2288,59 @@ Computes dot products between slices of `lhs` and slices of `rhs` and produces a
...
-  dimensions=[0, ..., size(lhs_contracting_dimensions) - 1],
-  body=lambda x, y: add(x, y))`.
+* For `is_non_quantized_tensor(lhs) and is_non_quantized_tensor(rhs)`:
+  * `dot_product = reduce(
+      inputs=[multiply(reshaped_lhs_slice, reshaped_rhs_slice)],
+      init_values=[0],
+      dimensions=[0, ..., size(lhs_contracting_dimensions) - 1],
+      body=lambda x, y: add(x, y))`.
+* For `is_quantized_tensor(lhs) and is_quantized_tensor(rhs)`:
+  * `integer_dot_product = reduce(
```
dominicsymes, line 2328: Does the specification define the type or range for integer_dot_product? For example, for 8-bit quantized data I would expect integer_dot_product to have 32-bit integer range. For 16-bit quantized data I would expect a larger range such as 48-bit integer.

docs/spec.md
```diff
@@ -2288,23 +2288,59 @@ Computes dot products between slices of `lhs` and slices of `rhs` and produces a
...
+      inputs=[multiply((reshaped_lhs_slice - zero_point(reshaped_lhs_slice)),
+                       (reshaped_rhs_slice - zero_point(reshaped_rhs_slice))],
+      init_values=[0],
+      dimensions=[0, ..., size(lhs_contracting_dimensions) - 1],
+      body=lambda x, y: add(x, y))`.
+  * `rounded_dot_product = round_nearest_even(integer_dot_product * (scale(reshaped_lhs_slice) * scale(reshape_rhs_slice) / scale(result)))`.
```
dominicsymes, line 2334: The scaling on this line is defined in terms of floating-point scale values and so can be problematic for integer only processing. Similar to the discussion on Add quantization, can this scaling be converted to an integer multiply and shift rescale for integer processing?”

docs/spec.md
sdasgup3: That a good point and you are right that the specifics of accumulation range is not specified in the specification document for ops like dot_general or convolution. IMO, the problem is not limited to quantized types but non-quantized types as well, which allows us to isolate it from the current PR. We have an open issue to address that https://github.com/openxla/stablehlo/issues/755. (As mention in a [comment](https://github.com/openxla/stablehlo/issues/1439#issuecomment-1527815787),  precision_config has ties with the accumulation range).

Let me know if you have any feedback on the above arrangement.

docs/spec.md
sdasgup3: Hi @dominicsymes 

>> can this scaling be converted to an integer multiply and shift rescale for integer processing?”

The current proposal uses the [textbook](https://arxiv.org/pdf/1712.05877.pdf) [Section 2.2] approach to define the specification. Specifically to ensure  bit-exactness, we can have the following arrangement: 

Following the ongoing discussion in https://groups.google.com/a/openxla.org/g/openxla-discuss/c/iwE9is49SS4, here is the rough sketch of the proposal that we can plan for:

1. Other than the uniform quantized type with floating-point scale, StableHLO to have a type with integer [multiplier, shift]. 
2. There will be defined conversion between the the two types. 
3. For the sake generality of the specification, I would like to propose that the consumers of StableHLO to be given the liberty to generate rescale operation, normalizing the operands of any particular op, in their preferred way. Given that, even for for particular add op, there could be different options of doing the rescale, I think these rescaling choices to be better governed and applied by the consumers only.

Happy to understand your perceptive on this.

docs/spec.md
dominicsymes: Thanks @sdasgup3 for the link to [#755](https://github.com/openxla/stablehlo/issues/755) and related comments. I agree these issues are related. The precision_config described in [#755](https://github.com/openxla/stablehlo/issues/755) is mainly talking about accuracy of floating-point accumulations in terms of DEFAULT, HIGH, HIGHEST. For an integer sum, I think the main issue is likely to be accumulator overflow rather than accuracy. For this the operation should describe the minimum accumulator width supported.

There are some operations where different accumulator widths may be useful to select. For example, for quantized 16-bit by 8-bit dot product operations, both 32-bit and 48-bit accumulation range may be useful depending on the usage. This could be selected by something like DEFAULT/HIGH but it would be range rather than accuracy.

Are there plans to expand [#755](https://github.com/openxla/stablehlo/issues/755) to cover quantized accumulator widths as part of the precision definition?

docs/spec.md
dominicsymes: Hi @sdasgup3

Thanks for describing the current proposal. The general scheme seems fine to me. In terms of bit-exactness, I think it would be useful to have a reference conversion to integer rescale operations. This could then be implemented bit-exactly or used as a comparison point if a consumer varies how the rescale operations are generated as you propose in (3).

docs/spec.md
sdasgup3: Thanks @dominicsymes for explaining. I agree with you. 

>> the operation should describe the minimum accumulator width supported.

As you said,  the `precision_config`  covers part of the picture (the precision at which the computation should happens). In general, for dot_general or convolution, the accumulation should be done in a range determined by at least  the element-type of the output tensor type (which [preferred_element_type](https://github.com/openxla/stablehlo/issues/600) can override).

I should say that the combination of #755 and #600 should help bring clarity into this.

sdasgup3:
I have updated the spec with two changes:
1. For now removed the hybrid format of the dot_general op. The intention is to facilitate the review without getting bothered by the details of how hybrid ops to be handled. I hope this decision still aligned with the [ongoing discussion](https://github.com/openxla/stablehlo/pull/1413#discussion_r1183148586) where we are gathering feedback on the relevance of a hybrid op in StableHLO spec. 
2. Following the proposal in https://github.com/openxla/stablehlo/issues/1406#issuecomment-1528185382, I have removed the constraints on the `storage_min` and `storage_max` values. 

Feel free to let me know your feedback!

docs/spec.md
```diff
@@ -2331,47 +2341,58 @@ planning to address this in
...
   size(`lhs_contracting_dimensions`)).
-* (C12) size(`precision_config`) $=$ 2.
-* (C13) shape(`result`) $=$ dim(`lhs`, `lhs_batching_dimensions`) +
+* (C11) size(`precision_config`) $=$ 2.
+* (C12) shape(`result`) $=$ dim(`lhs`, `lhs_batching_dimensions`) +
   dim(`lhs`, `lhs_result_dimensions`) + dim(`rhs`, `rhs_result_dimensions`).
+* If the operation uses non-quantized tensors:
+  * (C13) element_type(`lhs`) $=$ element_type(`rhs`).
+* If the operation uses quantized tensors:
+  * (C14) `is_quantized_tensor(lhs) and is_quantized_tensor(rhs)`.
```
loganchien, line 2388: and is_quantized_tensor(result)

docs/spec.md
```diff
@@ -2331,47 +2341,58 @@ planning to address this in
...
+* (C10) dim(`lhs`, `lhs_contracting_dimensions[i]`) $=$
   dim(`rhs`, `rhs_contracting_dimensions[i]`) for all `i` $\in$ [0,
   size(`lhs_contracting_dimensions`)).
-* (C12) size(`precision_config`) $=$ 2.
-* (C13) shape(`result`) $=$ dim(`lhs`, `lhs_batching_dimensions`) +
+* (C11) size(`precision_config`) $=$ 2.
+* (C12) shape(`result`) $=$ dim(`lhs`, `lhs_batching_dimensions`) +
   dim(`lhs`, `lhs_result_dimensions`) + dim(`rhs`, `rhs_result_dimensions`).
+* If the operation uses non-quantized tensors:
+  * (C13) element_type(`lhs`) $=$ element_type(`rhs`).
```
loganchien, line 2386: (C?) is_not_quantized_tensor(`lhs`) and is_not_quantized_tensor(`rhs`) and is_not_quantized_tensor(`result`)

docs/spec.md
```diff
@@ -2331,47 +2341,58 @@ planning to address this in
...
+* If the operation uses quantized tensors:
+  * (C14) `is_quantized_tensor(lhs) and is_quantized_tensor(rhs)`.
+  * (C15) `storage_type(lhs) =  storage_type(rhs) = storage_type(result)`.
```
loganchien, line 2389: I think we may want to allow `storage_type(result) != storage_type(rhs)` so that it is possible to dump the summation directly before applying the activation function.

For example:

```
%t = "stablehlo.dot_general"(%input, %param) : (tensor<... x !quant.uniform<ui8:f32, s1:zp1>>, tensor<... x !quant.uniform<ui8:f32, s2:zp2>> -> tensor<... x !quant.uniform<i32:f32, s3:zp3>>
%dq = "stablehlo.dequantize"(%t) : (tensor<... x !quant.uniform<i32:f32, s3:zp3>>) -> tensor<... x f32>
%actfn = "stablehlo.logistic"(%dq) : (tensor<... x f32>) -> tensor<... x f32>
%q = "stablehlo.quantize"(%actfn) : (tensor<... x f32>) -> ...
```

sdasgup3: Thanks for pointing that!I agree with the point. 
Additionally, even for non-quantized type, we do allow `element_type(result)` is be different from `element_type(lhs) (or element_type(rhs))`. 

Also, I note that as per https://github.com/openxla/stablehlo/issues/1439, we will be revisiting the type constraint between `lhs` and `rhs`.

docs/spec.md
sdasgup3: Hi @dominicsymes 

The usefulness of a reference conversion to rescale operations makes perfect sense to me. I believe that could be realized using a legalization pass from StableHLO, with int [miltiplier,shift] parameters, to TOSA rescale operators, similar to the one we have in [legalize_tfl.cc](https://github.com/tensorflow/tensorflow/blob/60f7a770d64cb6cb3a93f84c291272ec51304d31/tensorflow/compiler/mlir/tosa/transforms/legalize_tfl.cc#L1). 

Let me get back you with more information on addressing this.

docs/spec.md
```diff
@@ -2331,47 +2341,61 @@ planning to address this in
...
+* (C12) shape(`result`) $=$ dim(`lhs`, `lhs_batching_dimensions`) +
   dim(`lhs`, `lhs_result_dimensions`) + dim(`rhs`, `rhs_result_dimensions`).
+* If the operation uses non-quantized tensors:
+  * (C13) `is_non_quantized_tensor(lhs) and is_non_quantized_tensor(rhs) and
+    is_non_quantized_tensor(result)`.
+  * (C14) element_type(`lhs`) $=$ element_type(`rhs`).
+* If the operation uses quantized tensors:
+  * (C15) `is_quantized_tensor(lhs) and is_quantized_tensor(rhs) and
+    is_quantized_tensor(result)`.
+  * (C16) `storage_type(lhs) =  storage_type(rhs)`.
```
loganchien, line 2392: Remove one space character after =

loganchien (approved):

loganchien (approved):

docs/spec.md
sdasgup3: Hi @dominicsymes 
I have updated the issue #1404 with  https://github.com/openxla/stablehlo/issues/1404#issuecomment-1539225320 to track progress on the proposed reference converter.

abattery:
> (2) Hybrid quantized dot_general op dot_general(fp, q): Currently, this version of the op only supports dynamic range quantization, where the on-the-fly quantization of lhs is fused in the op-semantics. IMO, once we support https://github.com/openxla/stablehlo/issues/1407, the quantization logic can be un-fused and made explicit in the MLIR graph (cc @sngyhan).

Could you change this part of the description to reflect the recent discussions?

docs/spec.md
```diff
@@ -2302,16 +2302,26 @@ More formally, `result[result_index] = dot_product`, where:
...
+      dimensions=[0, ..., size(lhs_contracting_dimensions) - 1],
+      body=lambda x, y: add(x, y))`.
+* For `is_quantized_tensor(lhs) and is_quantized_tensor(rhs)`:
+  * `integer_dot_product = reduce(
+      inputs=[multiply((reshaped_lhs_slice - zero_point(reshaped_lhs_slice)),
+                       (reshaped_rhs_slice - zero_point(reshaped_rhs_slice))],
+      init_values=[0],
+      dimensions=[0, ..., size(lhs_contracting_dimensions) - 1],
+      body=lambda x, y: add(x, y))`.
+  * `rounded_dot_product = round_nearest_even(integer_dot_product * (scale(reshaped_lhs_slice) * scale(reshape_rhs_slice) / scale(result)))`.
```
abattery, line 2323: Similar to the comment for the round_nearest_even method.

https://github.com/openxla/stablehlo/pull/1446/files#r1188012676

docs/spec.md
dominicsymes: Thanks @sdasgup3  - I agree some combination of #755 and #600 should work (but output type is not sufficient on its own for accumulation range, as it is after scaling). I’ve added a comment in [#755](https://github.com/openxla/stablehlo/issues/755) to note the integer use case as well.

docs/spec.md
sdasgup3: Hi @abattery 
Similar to the add op case, IMO, we have in the current spec one possible variant of the round method. I note that  implementation can chose a rounding method of their choice.

sdasgup3:
Hello reviewers

I would like to propose that we remove 

```
(C17) `is_signed(storage_type(lhs)) = true`.
```

That is,  we do not have to limit to signed storage type.

@loganchien, who already approved the PR, has already on board into this. 
Please let me know if other reviewers have any concerns.

@dominicsymes @abattery @rewu93 @burmako

abattery (approved):

rewu93 (approved):

docs/spec.md
sdasgup3: Thanks @dominicsymes for adding the information. I will explore more on this as  part of the issue.

burmako (approved):

nutsiepully (commented):
LGTM, thanks.

Quick question about the clamp for storage min/max at the end of the operation. If you need that, hasn't an overflow already happened?

nutsiepully (approved):
LGTM, thanks.

Quick question about the clamp for storage min/max at the end of the operation. If you need that, hasn't an overflow already happened?

docs/spec.md
nutsiepully: If we can express dynamic range quantization using not a single StableHLO Op, but a composite set of StableHLO ops it would be fine. We can fallback to the existing TFLite implementation by pattern matching to it, however I think it's important it be expressible via StableHLO Ops, not just as a custom Op. There is a fair bit of production models depending on this behavior.

sdasgup3:
>> Quick question about the clamp for storage min/max at the end of the operation. If you need that, hasn't an overflow already happened?

That is a good point. I would say it depends. For example,  if the `storage_min/storage_max`  is not equal to `min_value(storage_type)` and `max_value(storage_type)` (in fact the the spec is open about the choice of `storage_min/storage_max` and does not put any such constraint on its values), then there is no overflow.  Another possibility for not having an overflow is, as you already pointed you, the clamp is "not needed".

However, if the  `storage_min/storage_max` are equal to the integer limits and the clamp is "really needed", then that would mean we already encountered an overflow.

dominicsymes (approved):
This seems fine to me provided the floating-point post scaling definition for quantized operation given here is not required to be matched and that a full integer implementation is possible. I think the comments on integer accumulator range and integer reference have been captured in other issues [#755](https://github.com/openxla/stablehlo/issues/755 ) and [#1404](https://github.com/openxla/stablehlo/issues/1404).

docs/spec.md
sdasgup3: Hi @nutsiepully 
Thanks for the information. We will explore more into the options.

PR #1419, opened 2023-04-19
Clarify semantics of custom_call::has_side_effect

sdasgup3:
fixes https://github.com/openxla/stablehlo/issues/1365.

Please refer to the issue for more information.

Evolution of this PR: Closed the initial  [attempt](https://github.com/openxla/stablehlo/pull/1392), which has some feedback comments which are now incorporated in the current PR.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -2202,9 +2202,11 @@ Encapsulates an implementation-defined operation `call_target_name` that takes
 `backend_config` and `api_version` may be used to provide additional
 implementation-defined metadata.
 
-At the moment, this operation contains a fairly disorganized collection of
-metadata which reflects organic evolution of its counterpart operation in
-the XLA compiler. In the future, we are planning to unify this metadata
+The metadata `has_side_effect` is used to prevent optimizations that can
+eliminate the op or execute it speculatively. The op has other metadata, like
+`backend_config` and `api_version`, which are fairly disorganized at the moment
+and refects organic evolution of its counterpart operation in the XLA compiler.
+In the future, we are planning to unify this metadata
 ([#741](https://github.com/openxla/stablehlo/issues/741)).
 
 #### Inputs


docs/spec.md
```diff
@@ -2202,9 +2202,11 @@ Encapsulates an implementation-defined operation `call_target_name` that takes
...
 
-At the moment, this operation contains a fairly disorganized collection of
-metadata which reflects organic evolution of its counterpart operation in
-the XLA compiler. In the future, we are planning to unify this metadata
+The metadata `has_side_effect` is used to prevent optimizations that can
+eliminate the op or execute it speculatively. The op has other metadata, like
+`backend_config` and `api_version`, which are fairly disorganized at the moment
+and refects organic evolution of its counterpart operation in the XLA compiler.
+In the future, we are planning to unify this metadata
 ([#741](https://github.com/openxla/stablehlo/issues/741)).
```
burmako, line 2210: It looks like the current wording represents a certain level of duplication between the first and the second paragraphs. How about the following rewording?

```
Encapsulates an implementation-defined operation `call_target_name` that takes
`inputs` and `called_computations` and produces `results`. 

`has_side_effect` can be used to ...

`backend_config` and `api_version` may be used to provide additional
implementation-defined metadata. This represents a fairly disorganized 
collection of metadata which reflects organic evolution of its counterpart 
operation in the XLA compiler. In the future, we are planning to unify this 
metadata ([#741](https://github.com/openxla/stablehlo/issues/741)).
```

docs/spec.md
```diff
@@ -2202,9 +2202,11 @@ Encapsulates an implementation-defined operation `call_target_name` that takes
 `backend_config` and `api_version` may be used to provide additional
 implementation-defined metadata.
 
-At the moment, this operation contains a fairly disorganized collection of
-metadata which reflects organic evolution of its counterpart operation in
-the XLA compiler. In the future, we are planning to unify this metadata
+The metadata `has_side_effect` is used to prevent optimizations that can
+eliminate the op or execute it speculatively. The op has other metadata, like
```
burmako, line 2206: What does it mean to execute the op speculatively? 

Is this referring to reordering it with other ops? If that's the case, then this formulation looks inconsistent with the discussion in #1365: "The semantics of a true value of the proto field prevent optimizations like DCE, CSE etc on the side effecting custom_call. However, it does not impose any ordering restrictions. Two unknown side-effecting custom call nodes in HLO DAG without any explicit data dependency between them can be executed in any order".

burmako (changes requested):

burmako:
This pull request hasn't been active in a while. I'll be closing it for now, to be reopened when we have more cycles.

Issue #1155, opened 2023-02-11
Audit correspondence between the spec and VHLO


burmako:
Before we start providing compatibility guarantees for VHLO, it sounds like a good idea to go through all ops in the spec and see where the correspondence between the spec and VHLO is not 1:1. All these cases should be documented in Frontend Contract tickets.

burmako:
Documenting the current progress of the audit:

1\) Attributes: #1188, #1190, #1191, #1192, #1194 and #1195 - these pull requests have made significant progress with aligning VHLO attributes with the StableHLO spec. Remaining work:
  * #740 (ArgResultAliasAttr/OutputOperandAliasAttr).
  * #1186 (VHLO_COMPARISON_TYPE_NOTYPE, VHLO_CUSTOM_CALL_API_VERSION_UNSPECIFIED and VHLO_TRANSPOSE_INVALID).
  * #425 (DictionaryAttr/TypeAttr).
  * #848 (The entirety of VhloEnums.td).
  * #8 (TypeExtensionsAttr).

2\) Types: #1189 makes some refactorings to align VHLO types with the StableHLO spec. Remaining work:
  * #8 (IndexType, RankedTensorType::encoding, UnrankedTensorType, Witness).
  * #588 (UniformQuantizedType).

3) Operations: #1184 streamlines VhloOps.td, ordering the ops alphabetically to simplify comparison with the StableHLO spec. Based on this work, I should be able to complete the audit shortly.

burmako:
#1184 completes the audit of the correspondence between the spec and VHLO. Remaining work:
  * #3 (BroadcastOp, CreateTokenOp, CrossReplicaSumOp, DotOp, EinsumOp, TorchIndexSelectOp, TraceOp, UnaryEinsumOp).
  * #629 (CustomCallOp, InfeedOp, OutfeedOp).
  * #740 (CustomCallOp).
  * #1187 (CustomCallOp).
  * #425 (CallOp, FuncOp, ReturnOp).
  * #588 (UniformDequantizeOp, UniformQuantizeOp).
  * #8 (ComputeReshapeShapeOp, CstrReshapableOp, DynamicBroadcastInDimOp, DynamicConvOp, DynamicGatherOp, DynamicIotaOp, DynamicPadOp, DynamicReshapeOp, RealDynamicSliceOp).

[...]

Issue #740, opened 2022-12-11
Decide on modelling buffer donation


burmako:
[Buffer donation](https://jax.readthedocs.io/en/latest/faq.html#buffer-donation) is an important technique for reducing memory pressure that we want to model in StableHLO.

When forking from MHLO, we have inherited two attributes to model buffer donation: [ArgResultAlias](https://github.com/openxla/stablehlo/blob/6ee83d2e55ea158f955be479196a284f7b6e9f3a/stablehlo/dialect/StablehloAttrs.td#L111-L142) and [OutputOperandAlias](https://github.com/openxla/stablehlo/blob/6ee83d2e55ea158f955be479196a284f7b6e9f3a/stablehlo/dialect/StablehloAttrs.td#L67-L109). These attributes are: 1) very similar to each other due to unclear reasons, 2) account for tuples which is something that we want to remove from StableHLO in the long term (#598).

Let's clean this up.

PR #1421, opened 2023-04-19
Remove uses of ArgResultAliasAttr as it is not used, specced, or supported in VHLO

GleasonK:
Was looking at VHLO code coverage for serialization and realized the only unused attribute is ArgResultAliasAttr, because it was unsupported in VHLO legalizations. Speaking with @burmako it sounds like this attribute is unused and should not be in StableHLO/VHLO currently without an RFC first.

This is not an incompatibility since serialization would fail for programs that use this attribute currently.

Adding @sdasgup3 as reviewer as well for spec implications (should be none?)


sdasgup3:
@GleasonK Just wanted to make sure that we will be having a separate Frontend contract ticket to remove the same from StableHLO per https://github.com/openxla/stablehlo/issues/1155.  Is that right?

burmako:
Removing my assignment for now to get a review from @sdasgup3 first. I'll chime in afterwards.

sdasgup3 (approved):

burmako (approved):
I agree that this attribute can be removed without an RFC. To the best of my knowledge, it is unused (I added it to StableHLO in August when forking from MHLO, where it is currently unused as well).

Overall, the functionality implemented by this attribute is useful. Buffer donation is an important feature of the XLA compiler, so we should be able to represent it in StableHLO per the overall philosophy that we expose HLO semantics that existed at the time of forking MHLO into StableHLO. It's just that this has to be done in a deliberate way, and we have #740 to follow up on this.

Therefore, the overall idea behind this PR is LGTM. I have left a few minor comments - please feel free to merge after they are addressed.

burmako:
@GleasonK Gentle ping. If I recall correctly, the proposed changes are fairly lightweight, so I think we should be fairly close to merging this after some minor follow-up work.

PR #1423, opened 2023-04-19
Add interpreter for RoundNearestEvenOp

### LINKED ISSUE #1111: Add interpreter for RoundNearestEvenOp

ghpvnist:
Here are the constraints for the RoundNearestEvenOp:
```
(I1) operand is a tensor of floating-point type.
(C1) `operand` and `result` have the same type.
```
I1 and C1 are covered by the ODS, so no additional tests are added.

closes #1111

### DIFF
--- a/stablehlo/reference/Element.h
+++ b/stablehlo/reference/Element.h
@@ -191,6 +191,9 @@ Element real(const Element &e);
 /// Returns the remainder for two Element objects.
 Element rem(const Element &e1, const Element &e2);
 
+/// Returns the value rounded to the nearest even of Element object.
+Element roundNearestEven(const Element &el);
+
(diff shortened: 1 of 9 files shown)


stablehlo/reference/Element.h
```diff
@@ -191,6 +191,9 @@ Element real(const Element &e);
 /// Returns the remainder for two Element objects.
 Element rem(const Element &e1, const Element &e2);
 
+/// Returns the value rounded to the nearest even of Element object.
```
sdasgup3, line 194: value rounded to nearest integer, breaking ties towards the even, of the Element object.

sdasgup3 (approved):
LGTM with one comment.

Issue #1416, opened 2023-04-18
Implement semantic compatibility testing


GleasonK:
### Request description

Currently we have adequate serialization compatibility tests in `stablehlo_legalize_to_vhlo.<version>.mlir` and `stablehlo_legalize_to_vhlo.<version>.mlir.bc`. However we do not have adequate nor scalable semantics tests that use the reference interpreter.

The existing testdata approach does not scale well. A few possibilities include versioning our unit tests, or adding some representative model tests that get versioned with each release.

PR #1425, opened 2023-04-19
Cleanup testdata bytecode tests

GleasonK:
Recently we gathered some code coverage metrics for testdata tests, and compared that to our unit tests.

It turns out that our testdata coverage was a strict subset of unit tests, including in generated files, which makes them not particularly useful as serialization and deserialization tests, and minimally useful as versioned semantic compatibility tests. For semantic compatibility, this approach does not scale well, as the overhead for cloning all these files at each StableHLO version bump is too high for its limited value (25MB of bytecode files per version, difficult to review due to PR size, consistently breaks our tooling). We have created #1416 to rethink semantic compatibility testing, and are planning to only version `stablehlo_legalize_to_vhlo.mlir` for serialization testing, which alone has 95% coverage of compatibility machinery, covering all ops, types, and attributes.

This PR was created using the following commands (meaning it should be uniform):

```
cd stablehlo/testdata
rm *.mlir
sed -i "/0_9_0/d" *.mlir
```


burmako (approved):

GleasonK:
Updated files (and commands used).

In the previous revision we weren't running interpreter tests at all on testdata, in this one we are running interpreter tests on the MLIR files.

burmako (approved):

GleasonK:
Updated commit description with some alternatives considered and rationale for removal.

PR #1436, opened 2023-04-20
Add interpreter for GetDimensionSizeOp

### LINKED ISSUE #1432: Add interpreter for GetDimensionSizeOp

ghpvnist:
Here are the constraints for the GetDimensionSizeOp:
```
(I1) operand is a tensor.
(I2) dimension is a constant of type `si64`.
(C1) 0 <= dimension < rank(operand).
```

```
I2: a) dimension is not a constant of type `si64`. (Covered by ODS).
C1: a) 0 < dimension.
    b) dimension < rank(operand).
```

```
C1a: 0 < dimension.
C1b: dimension < rank(operand).
```

closes #1432

### DIFF
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -2351,22 +2351,22 @@ def StableHLO_GatherOp: StableHLO_Op<"gather", [InferTensorTypeWithReify, Pure]>
 }
 
 def StableHLO_GetDimensionSizeOp: StableHLO_Op<"get_dimension_size",
-      [Pure, InferTensorType]> {
+      [Pure, InferTensorType /*get_dimension_size_c1*/]> {
   let summary = "GetDimensionSize operation";
   let description = [{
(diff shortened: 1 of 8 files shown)


stablehlo/dialect/StablehloOps.td
```diff
@@ -2351,22 +2351,22 @@ def StableHLO_GatherOp: StableHLO_Op<"gather", [InferTensorTypeWithReify, Pure]>
 }
 
 def StableHLO_GetDimensionSizeOp: StableHLO_Op<"get_dimension_size",
-      [Pure, InferTensorType]> {
+      [Pure, InferTensorType /*get_dimension_size_c1*/]> {
```
sdasgup3, line 2354: not sure if `c1` is appropriate in here? As it does not talk about the `result`. The constraint for `result` are all covered in the `Outputs` table entry and we do not have a label for the output like `i*`.

sdasgup3 (approved):

PR #1437, opened 2023-04-20
enable a few interpreter tests

sdasgup3:
A few interpreter tests which might have missed enabling.


ghpvnist (approved):
Thanks for the PR. I haven't gotten a chance to work on #1418, but once this is in place, all disabled tests with supported ops due to floating-point issues would be labeled as such. As for missing out on a few interpreter tests, I need to look into my script again to see why it's missing some obvious files like `stablehlo/testdata/atan_shape_complex64_20_20.mlir` (I plan to address this along #1418).

PR #1442, opened 2023-04-27
Bump patch to 0.10.1 after integrate

GleasonK:
After a StableHLO release is integrated into OpenXLA ([openxla/xla](https://github.com/openxla/xla/tree/main/third_party/stablehlo)), bump the patch version so HEAD remains ahead of the latest release.

### DIFF
--- a/stablehlo/integrations/python/tests/stablehlo.py
+++ b/stablehlo/integrations/python/tests/stablehlo.py
@@ -210,7 +210,9 @@ def test_api_version():
 @run
 def test_serialization_apis():
   curr_version = stablehlo.get_current_version()
-  assert curr_version == "0.10.1"
+  major_minor_patch = curr_version.split('.')
+  assert len(major_minor_patch) == 3
+  assert major_minor_patch[1] == "10"
 
   ASM = """
   func.func @test(%arg0: tensor<2xf32>) -> tensor<2xf32> {
--- a/stablehlo/integrations/python/tests/stablehlo.py
+++ b/stablehlo/integrations/python/tests/stablehlo.py
@@ -210,8 +210,8 @@ def test_api_version():
 @run
 def test_serialization_apis():
   curr_version = stablehlo.get_current_version()
-  assert curr_version == "0.10.0"
-  
+  assert curr_version == "0.10.1"
+
   ASM = """
   func.func @test(%arg0: tensor<2xf32>) -> tensor<2xf32> {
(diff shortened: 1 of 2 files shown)


stablehlo/integrations/python/tests/stablehlo.py
```diff
@@ -210,8 +210,8 @@ def test_api_version():
 @run
 def test_serialization_apis():
   curr_version = stablehlo.get_current_version()
-  assert curr_version == "0.10.0"
-  
+  assert curr_version == "0.10.1"
+
```
burmako, line 214: I wonder if we perhaps should not hardcode the version number here and instead say something along the lines of `assert curr_version is not None` or `assert type(current_version) = str`.

burmako (changes requested):

stablehlo/integrations/python/tests/stablehlo.py
GleasonK:
```
  major_minor_patch = curr_version.split('.')
  assert len(major_minor_patch) == 3
  assert major_minor_patch[1] == "10"
```

Meaning this test would be updated on all minor version bumps. Happy to also leave that out and just test that it is in major, minor, patch format.

burmako: "Happy to also leave that out and just test that it is in major, minor, patch format". Yeah, let's do that, perhaps via a more targeted regex like `r"\d+\.\d+\.\d+"` (mainly because I think it would be great to have a SoT for the version number).

GleasonK: done

sdasgup3 (approved):

stablehlo/integrations/python/tests/stablehlo.py
```diff
@@ -210,8 +211,8 @@ def test_api_version():
 @run
 def test_serialization_apis():
   curr_version = stablehlo.get_current_version()
-  assert curr_version == "0.10.0"
-  
+  assert bool(re.match('\d+\.\d+\.\d+', curr_version))
```
burmako, line 214: How about adding `^` and `$` to the beginning and the end of the regex correspondingly?

burmako, line 214: Also, do we need a `bool` cast here?

burmako (changes requested):

burmako (approved):

PR #1443, opened 2023-04-27
Add type check to matchInts for shape refinement

### LINKED ISSUE #1328: Stop crashing matchInts in --stablehlo-refine-shapes
```C++
template <typename T>
LogicalResult matchInts(Value value, SmallVector<T>& result) {
  DenseIntElementsAttr attr;
  if (!matchPattern(value, m_Constant(&attr))) return failure();
  for (auto element : attr.getValues<APInt>()) {
    if constexpr (std::is_same<T, int64_t>::value) {
      result.push_back(element.getSExtValue());
    } else {
      result.push_back(element);
    }
  }
  return success();
}
```

The `getValues` call will crash if we're asking for int64_t, and the element type is different (e.g. i32 or ui64). If that happens, we should return failure instead of crashing.

GleasonK:
Converting to uint asserted when passing a `SmallVector<int64_t>` to `matchInts`. Add a check that the requested input type is valid for the element type of the tensor being matched.

Fixes #1328

### DIFF
--- a/stablehlo/dialect/Base.h
+++ b/stablehlo/dialect/Base.h
@@ -110,12 +110,17 @@ LogicalResult inferMostSpecificTypeComponents(
     std::optional<Location> location, TypeRange inputTypes,
     SmallVectorImpl<ShapedTypeComponents> &inferredReturnShapes);
 
-inline bool matchNumericType(Type elementType, SmallVector<int64_t> &result) {
+template <typename T>
+bool matchElementType(Type elementType) = delete;
+
+template <>
+inline bool matchElementType<int64_t>(Type elementType) {
   // FIXME: Discuss if this should be signless 64?
   return elementType.isSignlessInteger();
 }
 
-inline bool matchNumericType(Type elementType, SmallVector<APInt> &result) {
+template <>
+inline bool matchElementType<APInt>(Type elementType) {
   return elementType.isa<IntegerType>();
 }
 
@@ -124,7 +129,7 @@ template <typename T>
 LogicalResult matchInts(Value value, SmallVector<T> &result) {
   DenseIntElementsAttr attr;
   if (!matchPattern(value, m_Constant(&attr))) return failure();
-  if (!matchNumericType(attr.getElementType(), result)) return failure();
+  if (!matchElementType<T>(attr.getElementType())) return failure();
   for (auto element : attr.getValues<APInt>()) {
     if constexpr (std::is_same<T, int64_t>::value) {
       result.push_back(element.getSExtValue());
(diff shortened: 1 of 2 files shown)


stablehlo/dialect/Base.h
```diff
@@ -110,11 +110,23 @@ LogicalResult inferMostSpecificTypeComponents(
     std::optional<Location> location, TypeRange inputTypes,
     SmallVectorImpl<ShapedTypeComponents> &inferredReturnShapes);
 
+inline bool matchNumericType(DenseIntElementsAttr attr,
+                             SmallVector<int64_t> &result) {
+  // FIXME: Discuss if this should be signless 64?
```
GleasonK, line 115: What is the intention of passing `matchInts(..., SmallVector<int64_t>)` vs `APInt`? Is this used to indicate signless integer type? Or is the intent that this would mean `i64` type? It is currently used as general signless integer type.

Here is a snippet from `DynamicConv`:

```
SmallVector<int64_t> padding;
if (failed(hlo::matchInts(op.getDPadding(), padding)))
  return rewriter.notifyMatchFailure(op, "expected constant d_padding");
```

DPadding is often i32 as well. Tests fail if `matchNumericType` were to check for `isSignlessInteger(64)` instead of just general signless integer.

burmako:
Expecting to either review this or send an alternative PR tomorrow 5/2.

stablehlo/dialect/Base.h
burmako: I've provided additional documentation around this in #1451. The overall idea was to have an int64_t-based API for when the underlying type doesn't matter (e.g. if we pass the results to int64_t-based APIs like ShapedTypeComponents), and an APInt-based API when it does matter (e.g. when we add the results together or something).

stablehlo/dialect/Base.h
```diff
@@ -110,11 +110,26 @@ LogicalResult inferMostSpecificTypeComponents(
...
+inline bool matchElementType<APInt>(Type elementType) {
+  return elementType.isa<IntegerType>();
+}
+
 // Matches a constant tensor with integer values into a 1-dimensional vector.
 template <typename T>
 LogicalResult matchInts(Value value, SmallVector<T> &result) {
   DenseIntElementsAttr attr;
   if (!matchPattern(value, m_Constant(&attr))) return failure();
+  if (!matchElementType<T>(attr.getElementType())) return failure();
```
burmako, line 132: "Converting to uint asserted when passing a SmallVector<int64_t> to matchInts". The crash was actually caused by something else - namely by `DenseIntElementsAttr::get` encountering a mismatch between the provided type (uint-based) and the provided values (int64_t-based).

I think it would be better to fix the crash at the source - see https://github.com/openxla/stablehlo/pull/1452/commits/ddf7e6895cdb725e1399c71c0f7308d1cda25f8a for details - and keep matchInts unchanged.

burmako (changes requested):

burmako:
Alternative fix has just landed in #1452.

Issue #1439, opened 2023-04-21
dot_general type constraints in spec are too strict


hawkinsp:
### What happened?

https://github.com/openxla/stablehlo/blob/main/docs/spec.md#dot_general
states that the LHS and RHS of a `dot_general` must have equal types. (Constraint C1)

This is stricter than what XLA already supports: XLA allows the LHS and RHS to have different types.

ghpvnist:
Linking related issue as I've also pointed this out in #755

sdasgup3:
Hi @hawkinsp 
Thanks for pointing to the issue. I agree that the constraint is too strict and (rightfully) not honored  by the current verifier. 
From [here](https://github.com/tensorflow/tensorflow/blob/49959d2f7d5e497edf90d6cdebc9afa3247234b7/tensorflow/compiler/xla/hlo/evaluator/hlo_evaluator_typed_visitor.h#L1411), hlo_verifier, and conversation of XLA/HLO experts, I realized that there are no particular constraints on the type of the operands and the result is as precise as if the operands were casted to the output type. 

As pointed by @ghpvnist, there is closely related concerns around the interpretation of the enum values value of precison config which governs at what precision the computation happens on those operations and at what precision the accumulation of results happens.    

So here is my proposal:
1. Remove the constraint from the spec.md
2. Clarify the spec.md with the interpretation of different operand types as part of https://github.com/openxla/stablehlo/issues/1354.

@hawkinsp Please let us know if this sounds right to you?

sdasgup3:
From my offline conversation,  @hawkinsp is onboard with the proposal. Here is the [PR](https://github.com/openxla/stablehlo/pull/1445) removing the redundant constraint. With that let us close the current ticket to be superseded by the [action item](https://github.com/openxla/stablehlo/issues/755#issuecomment-1483590911) discussed in issue https://github.com/openxla/stablehlo/issues/755.

sdasgup3:
Hi @hawkinsp 
As it turned out in the [conversation](https://github.com/openxla/stablehlo/pull/1445#discussion_r1183176358) that the "removing the constraint from the spec" cannot be isolated from addressing #755. Do you mind if we address the current issue once we have clarity on #755?

sdasgup3:
From my offline discussion with @hawkinsp, let us revisit this as proposed above. Unassigning for now.

PR #1445, opened 2023-04-28
Remove a redundant constraint on dot_general

sdasgup3:
fixes https://github.com/openxla/stablehlo/issues/1439

The constraint `C1:lhs and rhs have the same element type.` is redundant as this is not enforced on the `dot_general` op by the StableHLO static verifier. The PR proposes to remove it.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -2322,48 +2322,47 @@ planning to address this in
 
 #### Inputs
 
-| Label | Name                         | Type                                                         | Constraints                           |
-|-------|------------------------------|--------------------------------------------------------------|---------------------------------------|
-| (I1)  | `lhs`                        | tensor                                                       | (C1), (C6), (C7), (C10), (C11), (C13) |
-| (I2)  | `rhs`                        | tensor                                                       | (C1), (C8), (C9), (C10), (C11), (C13) |
-| (I3)  | `lhs_batching_dimensions`    | 1-dimensional tensor constant of type `si64`                 | (C2), (C4), (C6), (C10), (C13)        |
-| (I4)  | `rhs_batching_dimensions`    | 1-dimensional tensor constant of type `si64`                 | (C2), (C5), (C8), (C10)               |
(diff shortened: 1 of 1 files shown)


docs/spec.md
```diff
@@ -2322,48 +2322,47 @@ planning to address this in
...
 #### Outputs
 
 | Name     | Type   | Constraints |
 |----------|--------|-------------|
-| `result` | tensor | (C13)       |
+| `result` | tensor | (C12)       |
 
 #### Constraints
 
-* (C1) `lhs` and `rhs` have the same element type.
```
burmako, line 2343: What other ops need to be updated in a similar way? E.g. probably ConvolutionOp as well at the very least?

burmako, line 2343: Does the Semantics section need to be updated as well? As it's currently written, the `dot_product` not going to work out if the element types of `lhs` and `rhs` happen to be different.

burmako, line 2343: More generally, is this the mixed precision feature that we discussed in #369?

burmako (changes requested):

docs/spec.md
sdasgup3: Thanks for pointing out. 

Other than ConvolutionOp, there are also a set of ops mentioned in #369, where the spec says that the element type has to be equal but, as per HLO semantics,  they can be off by precision. I note that the the problem about the constraints checks of dot_general to be too strict [#1439](https://github.com/openxla/stablehlo/issues/1439) is equally relevant for others ops (mentioned in #369).

I realized that mere removing the constraint will not do the job. As you pointed out, the semantics section, as it right now, does not allow lhs and rhs to be differently typed. As part of this fix,  the semantics section needs to be updated with the interpretation of different typed operands, which is something we cannot do unless we  have a clear story on precision_config #755.   

IMO, we should revisit this as part of solving precision config.  Let me know what do you think? 

PS: Will also check with the author of #1439 for feedback on the decision.

sdasgup3: answered above.

sdasgup3: Please see above for the answer.

PR #1446, opened 2023-04-28
Specification for quantized  AddOp

sdasgup3:
## Summary 

The PR proposes the specification for quantized add op.

## A few details

At some point we [decided](https://github.com/openxla/stablehlo/pull/1352#discussion_r1166196224) to drop the introduction of the specification of this op mainly because we were unsure about the fate of https://github.com/openxla/stablehlo/issues/1406.
 
Please have a look at my revised proposal on https://github.com/openxla/stablehlo/issues/1406 and let me know if I am missing something. Otherwise, let us review this op  and let me know your feedback.   

Side note: For those who are already aware of the context of prior introduction of this op, please note that the current proposal is almost same as before except that it does not have any additional constraint imposed by the op's semantics on `storage_min` or `storage_max`.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -593,23 +593,34 @@ Performs element-wise addition of two tensors `lhs` and `rhs` and produces a
 * For integers: integer addition.
 * For floats: `addition` from IEEE-754.
 * For complex numbers: complex addition.
+* For quantized types:
+  * `float_result = (lhs - zero_point(lhs)) * scale(lhs) +
+    (rhs - zero_point(rhs)) * scale(rhs)`.
+  * `rounded_result = round_nearest_even(float_result / scale(result))`.
+  * `result = clamp(storage_min(result), rounded_result + zero_point(result), storage_max(result))`.
 
 #### Inputs
 
-| Label | Name  | Type   | Constraints |
-|-------|-------|--------|-------------|
-| (I1)  | `lhs` | tensor | (C1)        |
-| (I2)  | `rhs` | tensor | (C1)        |
+| Label | Name  | Type                       | Constraints |
+|-------|-------|----------------------------|-------------|
+| (I1)  | `lhs` | tensor or quantized tensor | (C1-C4)     |
+| (I2)  | `rhs` | tensor or quantized tensor | (C1-C3)     |
 
 #### Outputs
 
-| Name     | Type   | Constraints |
-|----------|--------|-------------|
-| `result` | tensor | (C1)        |
+| Name     | Type                       | Constraints |
+|----------|----------------------------|-------------|
+| `result` | tensor or quantized tensor | (C1-C3)     |
 
 #### Constraints
 
-* (C1) `lhs`, `rhs` and `result` have the same type.
+* (C1) `shape(lhs) = shape(rhs) = shape(result)`.
+* If the operation uses non-quantized tensors:
+  * (C2) `element_type(lhs) = element_type(rhs) = element_type(result)`.
+* If the operation uses quantized tensors:
+  * (C3) `element_type(lhs) = element_type(rhs) = element_type(result)`,
+    except for scales and zero points which may differ.
+  * (C4) `quantization_dimension(lhs)` is empty.
 
 #### Examples
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -619,7 +619,7 @@ Performs element-wise addition of two tensors `lhs` and `rhs` and produces a
   * (C2) `element_type(lhs) = element_type(rhs) = element_type(result)`.
 * If the operation uses quantized tensors:
   * (C3) `element_type(lhs) = element_type(rhs) = element_type(result)`,
-    except for scales and zero points which may differ.
+    except for quantization parameters which may differ.
   * (C4) `quantization_dimension(lhs)` is empty.
 
 #### Examples


loganchien (approved):
LGTM

Note: There can be multiple ways to implement quant add.  This proposal uses the dequantize-add-quantize approach.  Some implementations may rescale both lhs/rhs to a common scale and then do integer add to do (almost) everything integer.  As long as there is no bit-exact requirement, I think this should be fine.

eric-k256 (approved):
I'd agree with @loganchien's comments above. As indicated in the thread on openxla-discuss, with TOSA we would implement this with a full integer pipeline. For implementations that choose to implement via floating-point this seems reasonable.

docs/spec.md
```diff
@@ -593,23 +593,34 @@ Performs element-wise addition of two tensors `lhs` and `rhs` and produces a
 * For integers: integer addition.
 * For floats: `addition` from IEEE-754.
 * For complex numbers: complex addition.
+* For quantized types:
+  * `float_result = (lhs - zero_point(lhs)) * scale(lhs) +
+    (rhs - zero_point(rhs)) * scale(rhs)`.
+  * `rounded_result = round_nearest_even(float_result / scale(result))`.
```
abattery, line 599: Floor implementation can be varied per hardware. For example, some cases prefer floor(x+ 0.5) instead of round_nearest_even.

Can we consider not specifying the round method here?

docs/spec.md
```diff
@@ -593,23 +593,34 @@ Performs element-wise addition of two tensors `lhs` and `rhs` and produces a
...
 #### Constraints
 
-* (C1) `lhs`, `rhs` and `result` have the same type.
+* (C1) `shape(lhs) = shape(rhs) = shape(result)`.
+* If the operation uses non-quantized tensors:
+  * (C2) `element_type(lhs) = element_type(rhs) = element_type(result)`.
+* If the operation uses quantized tensors:
+  * (C3) `element_type(lhs) = element_type(rhs) = element_type(result)`,
+    except for scales and zero points which may differ.
+  * (C4) `quantization_dimension(lhs)` is empty.
```
abattery, line 623: Just for the record, personally, I have not seen that per channel quantization addition operator implementation.  It seems okay for now. However, we might update this part if we are really going to implement such operation in the future.

docs/spec.md
abattery: If this explanation describes just one possible way to calculate the given case, it should be fine.

docs/spec.md
sdasgup3: I have the same observation and hence the specification to keep it simple for now. This is open for update in future in needed.

docs/spec.md
sdasgup3: Hi @abattery 

>> Round implementations can be varied per hardware.
I agree. The current variant of round follows from https://mlir.llvm.org/docs/Quantization/#relation and it is one one possible way.

abattery (approved):

rewu93 (approved):

burmako (approved):

nutsiepully:
LGTM

As mentioned by others, our implementation will be done in int-space rather than via dequantization + quantization and hence the numerics will be different. Likely, the StableHLO reference implementation will also reflect that?

nutsiepully (approved):

sdasgup3:
@nutsiepully 
>> Likely, the StableHLO reference implementation will also reflect that?

I can see the how the integer implementation will help match the numeric closely. Will keep that in mind while implementing that.

docs/spec.md
```diff
@@ -593,23 +593,34 @@ Performs element-wise addition of two tensors `lhs` and `rhs` and produces a
...
 
 #### Constraints
 
-* (C1) `lhs`, `rhs` and `result` have the same type.
+* (C1) `shape(lhs) = shape(rhs) = shape(result)`.
+* If the operation uses non-quantized tensors:
+  * (C2) `element_type(lhs) = element_type(rhs) = element_type(result)`.
+* If the operation uses quantized tensors:
+  * (C3) `element_type(lhs) = element_type(rhs) = element_type(result)`,
+    except for quantization parameters which may differ.
```
doyeonkim0, line 622: I think quantization_dimension may also differ because quantization_dimension(lhs) is enforced to be empty.
Also, when they differ, quantization_dimension(result) = quantization_dimension(rhs) will hold.

loganchien: I think in this spec **quantization parameters** only refer to `scales` and `zero_points`.

Xref.

> Quantized element types represent integer values of a storage type in the range from storage_min to storage_max (inclusive) that correspond to floating-point values of an expressed type. For a given integer value `i`, the corresponding floating-point value f can be computed as `f = (i - zero_point) * scale`, where `scale` and `zero_point` are called quantization parameters.

> In quantized tensors, quantization can be per-tensor, meaning, having one `scale` and `zero_point` for the entire tensor or can be per-axis, meaning, having multiple `scales` and `zero_points`, one pair per slice of a particular dimension `quantized_dimension`. More formally, in a tensor t of with per-axis quantization, there are `dim(t, quantized_dimension)` slices of the `quantized_dimension`: `t[:, ..., 0, ..., :]`, `t[:, ..., 1, ..., :]`, etc. All elements in the ith slice use `scales[i]` and `zero_points[i]` as their quantization parameters.

Also, I think it will be necessary to require `quantized_dimension(rhs) == quantized_dimension(result)` (same as current text).  If we allow different `rhs`/`result` `quantized_dimension`, the implementation will need 2-D scales to rescale the immediate result to the output scale.

sdasgup3: >> I think in this spec quantization parameters only refer to scales and zero_points.
That is right.

>> quantization_dimension(lhs) is enforced to be empty.
This constraint is no longer there in favor of the generality of the semantics. Just an FYI about the updated spec of the add in [here](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#add). The only constraint we have right now is:
`(C1) baseline_type(lhs) = baseline_type(rhs) = baseline_type(result)` which allows only the quantization parameters (scales/zero-points) of `lhs`, `rhs`, and `result` can differ. 

I will get back on the suggestion to differ `quantized_dimension`.

sdasgup3: @loganchien One follow up question:

If an implementation follows the current way defining semantics `dequant_op_quant`, how does it matter if the `quantization_dimenstion(rhs) != quantization_dimension(result)`. The intermediate floating point results can be quantized back to the result type with any quantization dimension appropriate to the result type. Or am I missing something?

I have updated the https://github.com/openxla/stablehlo/issues/1574 with the above information which addressees per-axis quantization. Let us talk about this more at #1574.

loganchien: If `quantization_dimenstion(rhs) != quantization_dimension(result)`, the scale will become 2-D.  For example:

```
a = dense<[[1, 2], [3, 4], [5, 6]]> : tensor<3x2x!quant.uniform<u8:f32, 0.5:0>>
b = dense<[[1, 2], [3, 4], [5, 6]]> : tensor<3x2x!quant.uniform<u8:f32:0, {3.0:0,4.0:0,5.0:0}>>
c = add(a, b) : (tensor<3x2x!quant.uniform<u8:f32, 0.5:0>>, tensor<3x2x!quant.uniform<u8:f32:0, {3.0:0,4.0:0,5:0}>>) -> tensor<3x2x!quant.uniform<u8:f32:1, {6.0:0,7.0:0}>>
```

The computation for `c`:

```
c[0][0] = (a[0][0] * 0.5 + b[0][0] * 3.0) * 6.0
c[0][1] = (a[0][1] * 0.5 + b[0][1] * 3.0) * 7.0
c[1][0] = (a[1][0] * 0.5 + b[1][0] * 4.0) * 6.0
c[1][1] = (a[1][1] * 0.5 + b[1][1] * 4.0) * 7.0
c[2][0] = (a[1][0] * 0.5 + b[1][0] * 5.0) * 6.0
c[2][1] = (a[1][1] * 0.5 + b[1][1] * 5.0) * 7.0
```

There are 6 scales `3 * 6`, `3 * 7`, `4 * 6`, `4 * 7`, `5 * 6`, and `5 * 7`.

PR #1451, opened 2023-05-02
Refactor matchInts to use APSInt instead of APInt

burmako:
As I was cleaning up the uses of matchInts, I figured I'd do the long-standing refactoring to switch to using APSInt.

This had the benefit of simplifying partial evaluation logic in --stablehlo-refine-shapes because with APSInts we no longer need to branch on whether the underlying APInts are signed or unsigned.

### DIFF
--- a/stablehlo/transforms/StablehloRefineShapes.cpp
+++ b/stablehlo/transforms/StablehloRefineShapes.cpp
@@ -55,6 +55,33 @@ namespace stablehlo {
 
 namespace {
 
+// DenseElementsAttr can be constructed from ArrayRef<APInt> but not from
+// ArrayRef<APSInt>. This helper bridges the gap.
+DenseIntElementsAttr getTensorAttr(ShapedType type, ArrayRef<APSInt> values) {
+  SmallVector<APInt> supportedValues;
+  for (auto value : values) {
+    supportedValues.push_back(value);
+  }
+  return DenseIntElementsAttr::get(type, supportedValues);
+}
+
+APSInt getAPSInt(Type type, uint64_t value) {
+  unsigned numBits;
+  bool isUnsigned;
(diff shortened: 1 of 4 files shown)


stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -55,6 +55,33 @@ namespace stablehlo {
 
 namespace {
 
+// DenseElementsAttr can be constructed from ArrayRef<APInt> but not from
+// ArrayRef<APSInt>. This helper bridges the gap.
+DenseIntElementsAttr getTensorAttr(ShapedType type, ArrayRef<APSInt> values) {
+  SmallVector<APInt> supportedValues;
```
GleasonK, line 61: Does the vector constructor allow coercion?

```
SmallVector<APInt> supportedValues(values.begin(), values.end());
```

stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -55,6 +55,33 @@ namespace stablehlo {
...
+  return DenseIntElementsAttr::get(type, supportedValues);
+}
+
+APSInt getAPSInt(Type type, uint64_t value) {
+  unsigned numBits;
+  bool isUnsigned;
+  if (auto integerType = type.dyn_cast<IntegerType>()) {
+    numBits = integerType.getWidth();
+    // Signless types are treated as signed, per StableHLO convention.
+    isUnsigned = integerType.getSignedness() == IntegerType::Unsigned;
```
GleasonK, line 74: `integerType.isUnsigned()` ?

GleasonK (approved):

stablehlo/transforms/StablehloRefineShapes.cpp
burmako: Nice! It indeed does.

PR #1452, opened 2023-05-02
Refactor EvalSliceOpPattern to use APSInt-based matchInts

burmako:
This PR is based on #1451.

Erasing the bitness and the signedness of the underlying values to
int64_t is reasonable when we're going to use these values in the
int64_t context (e.g. take a shape of DynamicReshapeOp and then put it
into ShapedTypeComponents).

However, in the case of data movement ops like BroadcastInDimOp,
ConcatenateOp and now SliceOp, this doesn't make much sense.

This even caused a crash when matchInts was called when trying to
evaluate slices of unsigned tensors. When we were trying to call
DenseIntElementsAttr::get with an unsigned tensor type and values of
type SmallVector<int64_t>, that led to a crash.

Now that we have switched to SmallVector<APSInt>, this crash is fixed.
I've also audited all occurrences of DenseIntElementsAttr::get to make
sure that we no longer have any type mismatches that could lead to
further crashes elsewhere.


stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -359,20 +359,20 @@ struct EvalSliceOpPattern : public OpRewritePattern<SliceOp> {
     if (!resultType.hasRank() || resultType.getRank() != 1)
       return rewriter.notifyMatchFailure(op, "expected 1-dimensional type");
 
-    SmallVector<int64_t> operand;
+    SmallVector<APSInt> operand;
     if (failed(hlo::matchInts(op.getOperand(), operand)))
```
GleasonK, line 363: Do we have a test that exercises matchInts failing?

Perhaps add an eval slice ui64 test? Can pick from 
https://github.com/openxla/stablehlo/pull/1443/files#diff-803b64030ac18cf321610effd8bb8b56a7b8d52629e683f284bddc597d8a2a7b

GleasonK (approved):

stablehlo/transforms/StablehloRefineShapes.cpp
burmako: ui64 isn't going to fail - see tests added in this PR for an example. I'll add an f32 test though, that's a good point.

PR #1461, opened 2023-05-04
Add Python Serialization APIs that operate on strings

### LINKED ISSUE #1361: Add python APIs for serialization that accept modules as string arguments
### Request description

Users that have StableHLO built in a separate shared library cannot use the current API's since passing an MLIRContext across shared library boundaries causes issues.

```
def serialize_portable_artifact_str(module_str, target) -> bytes
def deserialize_portable_artifact_str(artifact_bytes):  -> str
```

GleasonK:
More details on rationale in API comments of `stablehlo/dialect/Serialization.h`.

After learning more about Python bindings, unless build is set up in a specific way where all dialect extensions built together so type IDs are accurate, passing/returning strings is safer. These APIs give the option to do either.

Backport of https://github.com/tensorflow/mlir-hlo/commit/6d62e3157aec86d9a6c023595c1c7f89ecf928da

Closes #1361

### DIFF
--- a/stablehlo/dialect/Serialization.cpp
+++ b/stablehlo/dialect/Serialization.cpp
@@ -15,20 +15,27 @@ limitations under the License.
 
 #include "stablehlo/dialect/Serialization.h"
 
-#include <sstream>
-
 #include "llvm/Support/raw_ostream.h"
 #include "mlir/Bytecode/BytecodeWriter.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/IR/BuiltinOps.h"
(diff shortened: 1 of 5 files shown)


stablehlo/dialect/Serialization.cpp
```diff
@@ -15,20 +15,27 @@ limitations under the License.
```
burmako, line 15: Since this is a backport from MLIR-HLO, let's note that in the commit message. I've also tagged the PR accordingly.

burmako (approved):
LGTM but please update the PR description.

Issue #1460, opened 2023-05-03
Explore adding support for big endian architectures


ghpvnist:
This was brought up from the [discussions](https://github.com/openxla/xla/discussions/5) on the OpenXLA side. Consider adding support for it in StableHLO as supported endianness for `bitcast_convert` is currently implementation defined.

PR #1463, opened 2023-05-05
Add interpreter for BitcastConvertOp

### LINKED ISSUE #1098: Add  interpreter for BitcastConvertOp

ghpvnist:
```
(I1) operand is a tensor.
(C1) Let `E` and `E'` be the `operand` and `result` element type,
respectively and `R = rank(operand)`:
* If `num_bits(E')` = `num_bits(E)`, shape(`result`) = shape(`operand`).
* If `num_bits(E')` < `num_bits(E)`:
  * `rank(result) = R+1`.
  * dim(`result`, `i`) = dim(`operand`, `i`) for all `i` in [0, `R`-1].
  * `dim(result, R) = num_bits(E)/num_bits(E')`.
* If `num_bits(E')` > `num_bits(E)`:
  * `rank(result) = R-1`.
  * dim(`result`, `i`) = dim(`operand`, `i`) for all `i` in [0, `R`-1).
  * `dim(operand, R-1) = num_bits(E')/num_bits(E)`.
(C2) Conversion between complex and non-complex types is not permitted.
```

```
C1: a) If `num_bits(E')` = `num_bits(E)`, shape(`result`) != shape(`operand`).
C1: b) If `num_bits(E')` < `num_bits(E)`: `rank(result) != R+1`.
C1: c) If `num_bits(E')` < `num_bits(E)`: dim(`result`, `i`) != dim(`operand`, `i`) for any `i` in [0, `R`-1].
C1: d) If `num_bits(E')` < `num_bits(E)`: `dim(result, R) != num_bits(E)/num_bits(E')`.
C1: e) If `num_bits(E')` > `num_bits(E)`: `rank(result) != R-1`.
C1: f) If `num_bits(E')` > `num_bits(E)`: dim(`result`, `i`) != dim(`operand`, `i`) for all `i` in [0, `R`-1).
C1: g) If `num_bits(E')` > `num_bits(E)`: `dim(operand, R-1) != num_bits(E')/num_bits(E)`.
(C2) Conversion between complex and non-complex types is permitted.
```

```
C1a: If `num_bits(E')` = `num_bits(E)`:
     shape(`result`) != shape(`operand`).
C1b: If `num_bits(E')` < `num_bits(E)`:
     `rank(result) != R+1`.
C1c: If `num_bits(E')` < `num_bits(E)`:
     dim(`result`, `i`) != dim(`operand`, `i`) for any `i` in [0, `R`-1].
C1d: If `num_bits(E')` < `num_bits(E)`:
     `dim(result, R) != num_bits(E)/num_bits(E')`.
C1e: If `num_bits(E')` > `num_bits(E)`:
     `rank(result) != R-1`.
C1f: If `num_bits(E')` > `num_bits(E)`:
     dim(`result`, `i`) != dim(`operand`, `i`) for all `i` in [0, `R`-1).
C1g: If `num_bits(E')` > `num_bits(E)`:
     `dim(operand, R-1) != num_bits(E')/num_bits(E)`.
C2: Conversion between complex and non-complex types is permitted.
```

Notes:
* The interpreter assumes little endian representation, and we have #1460 to address adding support for big endian architectures.

closes #1098


stablehlo/reference/Ops.cpp
```diff
@@ -402,6 +406,52 @@ Tensor evalAtan2Op(const Tensor &lhs, const Tensor &rhs,
   return result;
 }
 
+Tensor evalBitcastConvertOp(const Tensor &operand, ShapedType resultType) {
+  Tensor result(resultType);
+  auto operandElementType = operand.getElementType();
+  auto resultElementType = result.getElementType();
+  int64_t operandBitWidth = isSupportedComplexType(operandElementType)
```
sdasgup3, line 413: complex type are not supported right?

stablehlo/reference/Element.cpp
```diff
@@ -598,6 +604,210 @@ Element atan2(const Element &e1, const Element &e2) {
                                      debugString(type).c_str()));
 }
 
+Element bitcastConvertManyToOne(Type type, ArrayRef<Element> elements) {
```
sdasgup3, line 607: IMO, I don't think we need to have these three functions. For example, we can just copy the storage of operand to target at _byte granularity_. That is also followed from what we have in the spec.

With that we do not have to deal with interpreting the storage with a particular type and and write code specific to that type.

stablehlo/tests/interpret_bitcast_convert.mlir
```diff
@@ -0,0 +1,224 @@
...
+func.func @bitcast_convert_op_test_c128_to_c128() {
+  %operand = stablehlo.constant dense<(0x0123456789ABCDEF, 0x0000000011111111)> : tensor<complex<f64>>
+  %result = "stablehlo.bitcast_convert"(%operand) : (tensor<complex<f64>>) -> tensor<complex<f64>>
+  check.expect_eq_const %result, dense<(0x0123456789ABCDEF, 0x0000000011111111)> : tensor<complex<f64>>
+  func.return
+}
+
+// -----
+
+func.func @bitcast_convert_op_test_c64_to_c128() {
```
sdasgup3, line 219: complex is not supported for this op right?

stablehlo/tests/interpret_bitcast_convert.mlir
```diff
@@ -0,0 +1,224 @@
+// RUN: stablehlo-translate --interpret -split-input-file %s
+
+func.func @bitcast_convert_op_test_i1_to_i1() {
```
sdasgup3, line 3: IMO the implementation code  may not have to deal with each type, if we just copy the source bytes to destination. With that this test file will be simplified with few cases.

stablehlo/reference/Ops.cpp
ghpvnist: Conversions between complex and non-complex types aren't supported but complex to complex is supported.

stablehlo/tests/interpret_bitcast_convert.mlir
ghpvnist: Conversions between complex and non-complex types aren't supported but complex to complex is supported.

stablehlo/reference/Element.cpp
ghpvnist: There are two problems with this approach:
1. Some element types like int4 or booleans do not take up full 8 bits. The remaining bits are unset (could be anything), so by reinterpreting them as smaller or larger bits, it would result in wrong interpretation of the bits. E.g.
```
8x booleans -> 1x int8
little endian representation -> binary
Expected:
false, true, false, true, false, false, true, true -> -54
1111000[1] 1111000[1] 1111000[0] 1111000[0] 1111000[1] 1111000[0] 1111000[1] 1111000[0] -> 11001010

Actual (the actual representation may differ but the idea is same):
false, true, false, true, false, false, true, true -> -16
11110001 11110001 11110000 11110000 11110001 11110000 11110001 [11110000] -> 11110000
```
and even if we go with this approach, we'd still need to specialize booleans and int4/uint4 types + handle the three types of conversions separately, so there won't be any simplifications made.

2. The implementation does not align with the spec. I've updated the PR to not expose the three functions in `Ops.cpp` and put them in `Elements.cpp` as helper functions, which makes the implementation in `Ops.cpp` closer to the spec.

stablehlo/reference/Element.cpp
```diff
@@ -173,6 +173,221 @@ bool areApproximatelyEqual(APFloat f, APFloat g) {
...
+                       APFloat(baseType.getFloatSemantics(), binaryReal),
+                       APFloat(baseType.getFloatSemantics(), binaryImag)));
+  }
+
+  report_fatal_error(invalidArgument("Unsupported conversion from %s to %s",
+                                     debugString(operandType).c_str(),
+                                     debugString(type).c_str()));
+}
+
+// Returns bitcast converted larger bitwidth Element object to array of smaller
```
sdasgup3, line 261: Element object `el` ...

stablehlo/reference/Element.cpp
```diff
@@ -173,6 +173,221 @@ bool areApproximatelyEqual(APFloat f, APFloat g) {
   return std::fabs(f.convertToDouble() - g.convertToDouble()) <= 0.0001;
 }
 
+APInt leftShiftAndMask(APInt binary, int64_t shift, APInt mask) {
+  binary = binary.shl(shift);
+  binary |= mask;
+  return binary;
+}
+
+// Returns bitcast converted array of smaller bitwidth Element objects to one
```
sdasgup3, line 182: Element objects `elements` to ..

stablehlo/reference/Element.cpp
```diff
@@ -173,6 +173,221 @@ bool areApproximatelyEqual(APFloat f, APFloat g) {
...
+                                     .getIntOrFloatBitWidth()
+                               : type.getIntOrFloatBitWidth();
+
+  // Splits element into 'operandBitWidth' / 'resultBitWidth' elements, converts
+  // to Element object of type 'type', and appends them to 'results'.
+  auto applyIntOrFloatConversion = [&](APInt mask) {
+    for (auto i = 0; i < operandBitWidth; i += resultBitWidth) {
+      auto binary = leftShiftAndMask(
+          APInt(resultBitWidth, 0), resultBitWidth,
+          mask.extractBits(resultBitWidth, i).zext(resultBitWidth));
```
sdasgup3, line 285: extractBits generates a bitstring os size `resultBitWidth`. Zero-extending it by `resultBitWidth` seems like a no-op.

stablehlo/reference/Element.cpp
```diff
@@ -173,6 +173,221 @@ bool areApproximatelyEqual(APFloat f, APFloat g) {
...
+      if (isSupportedBooleanType(operandType))
+        mask = APInt(resultBitWidth, element.getBooleanValue());
+      else if (isSupportedIntegerType(operandType))
+        mask = element.getIntegerValue().zext(resultBitWidth);
+      else if (isSupportedFloatType(operandType))
+        mask = element.getFloatValue().bitcastToAPInt().zext(resultBitWidth);
+      binary = leftShiftAndMask(binary, operandBitWidth, mask);
+    }
+
+    if (isSupportedBooleanType(type)) return elements[0];
```
sdasgup3, line 214: Given that numbits(elements[0].getType()) < type, IMO, type cannot be boolean. Seems like dead check to me. Can  you please double check?

sdasgup3 (commented):
I will add more comments in a while

stablehlo/reference/Element.cpp
```diff
@@ -173,6 +173,221 @@ bool areApproximatelyEqual(APFloat f, APFloat g) {
...
+                               ? type.cast<ComplexType>()
+                                     .getElementType()
+                                     .cast<FloatType>()
+                                     .getIntOrFloatBitWidth()
+                               : type.getIntOrFloatBitWidth();
+  if (isSupportedBooleanType(operandType) && isSupportedBooleanType(type))
+    return el;
+
+  if (isSupportedIntegerType(operandType)) {
+    if (isSupportedIntegerType(type)) return el;
```
sdasgup3, line 362: It seems it may not work if operandType is `in` and type is `uin`. Also, we should add a test case for that as well.

stablehlo/reference/Element.cpp
```diff
@@ -173,6 +173,221 @@ bool areApproximatelyEqual(APFloat f, APFloat g) {
...
+
+    auto binary = APInt(resultBitWidth, 0);
+    APInt mask = el.getIntegerValue().zext(resultBitWidth);
+    binary = leftShiftAndMask(binary, operandBitWidth, mask);
+
+    if (isSupportedFloatType(type))
+      return Element(
+          type, APFloat(type.cast<FloatType>().getFloatSemantics(), binary));
```
sdasgup3, line 370: Add a test case for that, like `i64_to_f64`?

stablehlo/reference/Element.cpp
```diff
@@ -173,6 +173,221 @@ bool areApproximatelyEqual(APFloat f, APFloat g) {
...
+  }
+
+  if (isSupportedFloatType(operandType)) {
+    if (isSupportedFloatType(type)) return el;
```
sdasgup3, line 374: Similarly f16 -> bf16 may not work. Also, add a test case.

stablehlo/reference/Element.cpp
```diff
@@ -173,6 +173,221 @@ bool areApproximatelyEqual(APFloat f, APFloat g) {
...
+
+    APInt mask = el.getFloatValue().bitcastToAPInt().zext(resultBitWidth);
```
sdasgup3, line 376: As above the next two computations seem redundant. Can we remove them?

stablehlo/reference/Element.cpp
```diff
@@ -173,6 +173,221 @@ bool areApproximatelyEqual(APFloat f, APFloat g) {
...
+    auto binary =
+        leftShiftAndMask(APInt(resultBitWidth, 0), operandBitWidth, mask);
+
+    if (isSupportedIntegerType(type)) return Element(type, binary);
+  }
+
+  if (isSupportedComplexType(operandType) && isSupportedComplexType(type))
```
sdasgup3, line 383: a test for this?

stablehlo/reference/Element.cpp
sdasgup3, line 285: I think we can just say
```
auto binary  = mask.extractBits(resultBitWidth, i).zext(resultBitWidth));
```

stablehlo/reference/Element.cpp
```diff
@@ -173,6 +173,221 @@ bool areApproximatelyEqual(APFloat f, APFloat g) {
...
+  if (isSupportedComplexType(operandType) && isSupportedComplexType(type)) {
+    // splits operand into 'resultBitWidth' / 'operandBitWidth' elements,
+    // applies conversion to type 'type'. It converts two elements at a
+    // time, one for the real part and the other for the imaginary part, and
+    // appends them to 'results'.
+    auto applyComplexConversion = [&](APInt mask) {
+      auto baseType = type.cast<ComplexType>().getElementType();
+      for (auto i = 0; i < operandBitWidth; i += resultBitWidth) {
+        auto binaryReal = leftShiftAndMask(
+            APInt(resultBitWidth, 0), resultBitWidth,
```
sdasgup3, line 319: same redundancy as mentioned above

sdasgup3 (commented):
This concludes my review for this PR.

stablehlo/reference/Element.cpp
sdasgup3: agreed. Thanks for explaining.

stablehlo/tests/interpret_bitcast_convert.mlir
sdasgup3: nvm.

docs/spec.md
```diff
@@ -1306,12 +1306,12 @@ type of the `result` tensor.
 Let `E` and `E'` be the `operand` and `result` element type respectively, and
 `R = rank(operand)`:
 
-* If `num_bits(E')` $=$ `num_bits(E)`,
-  `bits(result[i0, ..., iR-1]) = bits(operand[i0, ..., iR-1])`.
 * If `num_bits(E')` $\lt$ `num_bits(E)`,
   `bits(result[i0, ..., iR-1, :]) = bits(operand[i0, ..., iR-1])`.
 * If `num_bits(E')` $\gt$ `num_bits(E)`,
   `bits(result[i0, ..., iR-2]) = bits(operand[i0, ..., iR-2, :])`.
+* If `num_bits(E')` $=$ `num_bits(E)`,
```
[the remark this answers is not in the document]
ghpvnist: The reshuffle matches 1:1 with how the implementation is written.

stablehlo/reference/Element.cpp
ghpvnist: Nice finding! This works because the underlying storage does not change and APInt currently does not distinguish between signed/unsigned. However, we plan to distinguish them separately in #1469 so we can address them together. I'll make a note on this.

stablehlo/reference/Element.cpp
ghpvnist: Similarly in comment above, the underlying bits do not change and the bitwidths are same. The Element's type is wrong but the Tensor's type is correct, thus passing the test. Will fix in the next commit.

stablehlo/reference/Element.h
```diff
@@ -149,6 +149,19 @@ Element atan2(const Element &e1, const Element &e2);
...
+/// the same type:
+// * size(elements) = 1 and num_bits(E') < num_bits(E):
+//   returns n Elements such that n = bitwidth(E)/bitwidth(E') and n *
+//   num_bits(E') == num_bits(E).
+// * size(elements) > 1 and num_bits(E') > num_bits(E):
+//   returns a single Element such that size(elements) * num_bits(E) ==
+//   num_bits(E').
+// * size(elements) = 1 and num_bits(E') == num_bits(E):
+//   returns a single Element of type `type`.
+SmallVector<Element> bitcastConvert(Type type, ArrayRef<Element> elements);
```
burmako, line 163: I left the rest of the feedback in https://github.com/ghpvnist/stablehlo/pull/8.

burmako (changes requested):

stablehlo/reference/Element.cpp
sdasgup3: >> The underlying element is of different type but since the tensor type is correct, the test passes.
That is right. However with the changes proposed in https://github.com/ghpvnist/stablehlo/pull/8 we may need to return the operand value `el`. 

In any case, distinguish them separately in https://github.com/openxla/stablehlo/issues/1469 make sense. 

Marking it resolved.

stablehlo/reference/Element.cpp
sdasgup3: Plus the fact that that we are interpreting  `bf16` and `f16` as `uint16` (as apposed to a float).
IMO it might makes sense to add a check in `Tensor::set(const Index &index, const Element &element)` between `getType().getElementType()`  and `element.getType()` to ensure all the `Element`s of a `Tensor` object  are of same type which is equal to the Tensor's `getType().getElementType()`.

stablehlo/tests/interpret_bitcast_convert.mlir
```diff
@@ -0,0 +1,278 @@
...
```
sdasgup3, line 3: The tests covers the following non complex types like `i1`, `i64`, `i4`, `ui64`, `f64`, `f16` `bf16`.

Given that the new proposal https://github.com/ghpvnist/stablehlo/pull/8 decouples `bitcastConvertManyToOne`, `bitcastConvertOneToMany` and `bitcastConvertOneToOne`  from  handling individual type conversions and rather puts all the type handling logic in   `fromBits` and `toBits`, do we need to use all the tests here? 

From the code coverage point of view, may be we need to cover  function like `fromBits` and `toBits` while writing tests with different types. For example, we may use tests to cover non-complex types

```
// to cover the non-complex types in fromBits

bitcast_convert_op_test_i1_to_i64
bitcast_convert_op_test_i64_to_i64
bitcast_convert_op_test_f64_to_i64

// to cover the non-complex types in toBits

bitcast_convert_op_test_i64_to_i1
bitcast_convert_op_test_i64_to_i64
bitcast_convert_op_test_i64_to_f64
```

what do you think?

sdasgup3 (commented):
Reviewed the PR except for 

1.  `bitcastConvertManyToOne`, `bitcastConvertOneToMany` and `bitcastConvertOneToOne` from this PR as there is a alternative proposal provided [here](https://github.com/ghpvnist/stablehlo/pull/8).

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3115,36 +3106,32 @@ LogicalResult verifyBitcastConvertOp(std::optional<Location> location,
...
   ArrayRef<int64_t> smallerEltPrefix;
   auto smallerEltBitwidth = std::min(targetEltBitwidth, operandEltBitwidth);
   auto biggerEltBitwidth = std::max(targetEltBitwidth, operandEltBitwidth);
+  // bitcast_convert_c1
   if (operandEltBitwidth != targetEltBitwidth) {
-    if (smallerEltShape.empty()) {
-      return emitOptionalError(location,
-                               "does not allow the smaller element type to be "
-                               "part of a 0d tensor, but got: ",
-                               operandType, " and ", targetType, ".");
```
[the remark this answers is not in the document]
ghpvnist: [This](https://github.com/tensorflow/tensorflow/commit/614c6e1ac980afbca3eeaf15a5174603327303e4) is the original commit that introduced this. I'm guessing it was added as part of adding verifiers to MHLO, but I don't have further context behind the change.

stablehlo/tests/interpret_bitcast_convert.mlir
```diff
@@ -0,0 +1,278 @@
...
+  check.expect_eq_const %result, dense<(0x0123456789ABCDEF, 0x0000000011111111)> : tensor<complex<f64>>
+  func.return
+}
+
+// -----
+
+func.func @bitcast_convert_op_test_c64_to_c128() {
+  %operand = stablehlo.constant dense<[(0x01234567, 0x89ABCDEF), (0x00000000, 0x11111111)]> : tensor<2xcomplex<f32>>
+  %result = stablehlo.bitcast_convert %operand : (tensor<2xcomplex<f32>>) -> tensor<complex<f64>>
+  check.expect_eq_const %result, dense<(0x89ABCDEF01234567, 0x1111111100000000)> : tensor<complex<f64>>
```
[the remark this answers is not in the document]
ghpvnist: XLA implementation interprets the bits as a different type without any manual conversions, so I followed the behavior of the experimental results I've got from running the HLO interpreter.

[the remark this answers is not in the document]
ghpvnist: I found the [expander code](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/bitcast_dtypes_expander.cc#L92C3-L92C3) that does the `shift and or` behavior. So yes, it does match the XLA behavior.

stablehlo/reference/Element.cpp
```diff
@@ -173,6 +173,174 @@ bool areApproximatelyEqual(APFloat f, APFloat g) {
...
+                                   el.getIntegerValue()));
+  }
+  if (isSupportedFloatType(operandType)) {
+    if (isSupportedFloatType(type))
+      return Element(type, APFloat(type.cast<FloatType>().getFloatSemantics(),
+                                   el.getFloatValue().bitcastToAPInt()));
+    if (isSupportedIntegerType(type))
+      return Element(type, APSInt(el.getFloatValue().bitcastToAPInt(),
+                                  /*isUnsigned=*/type.isUnsignedInteger()));
+  }
```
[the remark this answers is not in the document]
ghpvnist: The [expander code](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/bitcast_dtypes_expander.cc#L47-L50) uses the input/output bitwidths to get an unsigned `PrimitiveType`. However booleans aren't treated as unsigned here so bitcast conversions to and from booleans do not work at all, unless we are converting booleans to booleans, in which case it is a noop. It's unclear whether this was intentional or a miss.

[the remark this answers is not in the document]
ghpvnist: To follow the same behavior, we should also disallow conversions to/from booleans except bool -> bool.

stablehlo/reference/Element.h
ghpvnist: Thanks for the suggestions!

stablehlo/tests/interpret_bitcast_convert.mlir
ghpvnist: We need to cover test conversions for:
* `any type` to `any type` (not necessarily every combination).
* many to one
* one to many
* one to one

The following three tests has full coverage (I chose `bf16` for shorter tests compared to `i1`):
bf16 -> i64
i64 -> f64
f64 -> bf16

sdasgup3: The suggestion also looks good except that the we need a test for `i1` as source and destination. This is because boolean type is handled explicitly in both `from_bits` and `to_bits`. 

What about we modify the suggestion with replacing `bf16` with `i1`. Which that we still cover all of many to one,
one to many, and one to one. Plus test `from_bits` (and `to_bits`) with `bool`, `int`, `float`, `complex`  as source (and destination)

```
i1 -> i64
i64 -> f64
f64 -> i1
c128 -> c64
```

sdasgup3 (approved):
I approve the changes with a comment about the test case selection.

stablehlo/tests/interpret_bitcast_convert.mlir
ghpvnist: Sure, makes sense.

stablehlo/dialect/TypeInference.cpp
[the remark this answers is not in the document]
burmako: Alright, LGTM.

stablehlo/reference/Element.cpp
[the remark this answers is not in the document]
burmako: "It's unclear whether this was intentional or a miss". Agreed that it's not clear. Let's open a Spec ticket to find out if BitcastConvertOp is supposed to work with i1 element types and move forward with this PR as is.

burmako (approved):

PR #1467, opened 2023-05-07
Simplify verification of ReduceOp and ReduceWindowOp

burmako:
While reviewing PRs that implement interpreters for ReduceOp and ReduceWindowOp, I noticed that the verifiers can be somewhat simplified. This PR does the necessary cleanup.


stablehlo/dialect/TypeInference.cpp
```diff
@@ -454,14 +454,14 @@ LogicalResult verifyReplicaGroups(std::optional<Location> location,
 }
```
burmako, line 454: To be merged only after #1280 and #1336 are merged.

ghpvnist (approved):
LGTM

PR #1475, opened 2023-05-10
Introduce the specification of quantized type with multiplier and shift

sdasgup3:
Addresses https://github.com/openxla/stablehlo/issues/1404

The current PR does the followings:
1. Specifies grammar rules to introduce the integer multipler and shift  in `QuantizedElementType`. 
2. Informally defines the semantics of the multiplier and shift and demonstrates one possible implementation of deriving the values of multiplier and scale. 

There are a couple of open questions which I expect to resolve as part of the review:
1. Do the specification need to demonstrate an implementation (the one we current have using C++ code)? Maybe just the high level definition of multiplier and scale is fine. 
2.  Should the specification also need to explain, maybe using an example, how this new type enable integer-only computation?

Please let me know your feedback! 

**Update**
Based on the feedback received during the community meeting on 06/08/2023,  let me propose to limit this PR with (1) Introduction of multipliers and shifts in the type signature of QuantizedTensorType and (2) Specifying a conversion from floating-point scales to integer multipliers and shifts.

We will address the the specification of individual ops (like dot_general, add) with integers multipliers and shifts in a separate PR.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -175,9 +175,11 @@ following constraints:
 * (C11) If `is_empty(quantization_dimension)`, then `size(scales) = 1`.
 * (C12) `0 <= quantization_dimension`.
 
-In order to allow operation using only integer arithmetic, the floating-point
-`scale`,  can be realized using integer `multipler`, and `shift` values,
-such that `round_nearest_even(scale * 2^shift) == multipler`.
+In a given StableHLO operation, all the quantized tensor types are either
+represented using floating-point scales, or integer scales, with multipliers and
+shifts. The floating-point scale `scale` value and the integer scale, with
+`multipler`, and `shift` values are related as: `round_nearest_even(scale *
+2^shift) == multipler`.
 
 The following demonstrates, using C++ code, a possible implementation of
 deriving the integer parameters `multipler` and `shift` of type `i32` from a
@@ -222,6 +224,17 @@ above.
     std::fesetround(FE_TONEAREST);
     static_cast<int32_t>(std::rint(scale * std::pow(2, shift))) = multipler;
 ```
+
+The above conversion allows for the transformation of a StableHLO program that
+uses quantized types with floating-point scales to quantized types with integer
+scales. The latter allows for integer-only arithmetic. At a high-level, any
+computations specified in the operational semantics using floating-point scales
+can be realized using integer scales with integer-only arithmetic. Refer to
+[dot_general](#dot_general) op for an example. The exact manner in which a
+computation involving floating-point scales is converted to a computation using
+integer scales for each individual operation is implementation-defined. Also,
+the bit width of the multiplier and shift, chosen for each operation, is
+implementation defined.
 
 For simplicity, we will only use floating-point scale when defining the
 operation specification, unless otherwise specified.
@@ -2363,6 +2376,13 @@ More formally, `result[result_index] = dot_product`, where:
       dimensions=range(size(lhs_contracting_dimensions)),
       body=lambda x, y: add(x, y))`.
   * `rounded_dot_product = round_nearest_even(integer_dot_product * (scale(reshaped_lhs_slice) * scale(reshape_rhs_slice) / scale(result)))`.
+    If the `scale(reshaped_lhs_slice)`, `scale(reshape_rhs_slice)`, and
+    `scale(result)` are respectively represented as integer multiplier and
+    shift pairs, like,
+    `(M_reshaped_lhs_slice, n_reshaped_lhs_slice), (M_reshaped_rhs_slice, n_reshaped_rhs_slice), and (M_result, n_result)`,
+    then the computation can be implemented as
+    `rounded_dot_product = (integer_dot_product * M + (1<<(n-1))) >> n`, where
+    `M = ((M_reshaped_lhs_slice * M_reshaped_rhs_slice) / M_result)) and n = (n_reshaped_lhs_slice + n_reshaped_rhs_slice - n_result)`.
   * `dot_product = clamp(storage_min(result), rounded_dot_product + zero_point(result), storage_max(result))`.
 <!-- markdownlint-enable line-length -->
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -174,8 +174,9 @@ following constraints:
 * (C10) `size(scales) = size(zero_points)`.
 * (C11) If `is_empty(quantization_dimension)`, then `size(scales) = 1`.
 * (C12) `0 <= quantization_dimension`.
-* (C13) `0 <= multipliers`.
-* (C14) `0 < shifts`.
+* (C13) `size(multipliers) = size(shifts) = size(zero_points)`.
+* (C14) `0 <= multipliers`.
+* (C15) `0 < shifts`.
 
 All quantized tensor types, in a given StableHLO operation, either
 use scales or multipliers and shifts. The `scales`, `multiplers` and `shifts`,
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -133,9 +133,12 @@ QuantizationExpressedType ::= FloatType
 QuantizationDimension ::= IntegerConstant
 QuantizationParameters ::= QuantizationParameter
                          | '{' QuantizationParameter {',' QuantizationParameter} '}'
-QuantizationParameter ::= QuantizationScale ':' QuantizationZeroPoint
+QuantizationParameter ::= ( QuantizationScale | QuantizationMultiplierShift ) ':' QuantizationZeroPoint
 QuantizationScale ::= FloatConstant
-QuantizationZeroPoint ::= IntegerConstant
+QuantizationMultiplierShift ::= '<' QuantizationMultiplier , QuantizationShift '>'
+QuantizationMultiplier ::= IntegerConstant
+QuantizationShift ::= IntegerConstant
+QuantizationZeroPoint ::=  IntegerConstant
 ```
 
 | Name                     | Type                                        | Constraints                  |
@@ -171,10 +174,56 @@ following constraints:
 * (C11) If `is_empty(quantization_dimension)`, then `size(scales) = 1`.
 * (C12) `0 <= quantization_dimension`.
 
-At the moment, `QuantizationScale` is a floating-point constant, but there is
-strong interest in integer-based scales, represented with multipliers and
-shifts. We are planning to explore this in the near future
-([#1404](https://github.com/openxla/stablehlo/issues/1404)).
+In order to allow operation using only integer arithmetic, the floating-point
+`scale`, `S`, is realized using integer `multipler`, `M` and `shift` value,
+`n >= 0`, such that `round_nearest_integer(S) = (M + round)* 2^-n`, where
+`round` is added to recover the precision loss while deriving the values of
+`M` and `n` from `S`.
+
+The following demonstrates, using C++ code, a possible implementation of
+deriving the integer parameters `M` and `n` of type `i32` from a
+floating-point scale `S` of type `f64`:
+
+```c++
+    int32_t shift;
+    double mantissa = std::frexp(S, &shift);
+    auto shiftedMantissa = static_cast<int64_t>(std::round(mantissa * (int64_t(1) << 31)));
+
+    // Ensure that shiftedMantissa is within limits of 32-bit integer type.
+    if (shiftedMantissa == (int64_t(1) << 31)) {
+        shiftedMantissa /= 2;
+        shift++;
+    }
+
+    int64_t adjustedMantissa = shiftedMantissa;
+
+    // For small `S` value, flush `adjustedMantissa` to zero.
+    if (shift < -31) {
+        shift = 0;
+        adjustedMantissa = 0;
+    }
+
+    // Saturate `adjustedMantissa`, if shift > 30,
+    if (shift > 30) {
+        shift = 30;
+        adjustedMantissa = (1LL << 31) - 1;
+    }
+
+    n = (-shift) + 31;
+    M = static_cast<int32_t>(adjustedMantissa);
+```
+
+Using the above implementation, the following C++ code demonstrates the
+relationship between the floating-point scale `S` and the integer parameters `M`
+and `n`.
+
+```c++
+    int64_t round = static_cast<int64_t>(1) << (n - 1);
+    static_cast<int32_t>(std::floor(S + 0.5)) = clamp(
+                                    (static_cast<int64_t>(M) + round) >> n,
+                                    static_cast<int64_t>(std::numeric_limits<int32_t>::min()),
+                                    static_cast<int64_t>(std::numeric_limits<int32_t>::max()));
+```
 
 There is an ongoing discussion on the semantics of `QuantizationZeroPoint`,
 including the type, the values and whether there can be just one or
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -135,7 +135,7 @@ QuantizationParameters ::= QuantizationParameter
                          | '{' QuantizationParameter {',' QuantizationParameter} '}'
 QuantizationParameter ::= ( QuantizationScale | QuantizationMultiplierShift ) ':' QuantizationZeroPoint
 QuantizationScale ::= FloatConstant
-QuantizationMultiplierShift ::= '<' QuantizationMultiplier , QuantizationShift '>'
+QuantizationMultiplierShift ::= '<' QuantizationMultiplier ',' QuantizationShift '>'
 QuantizationMultiplier ::= IntegerConstant
 QuantizationShift ::= IntegerConstant
 QuantizationZeroPoint ::=  IntegerConstant
@@ -176,7 +176,7 @@ following constraints:
 
 In order to allow operation using only integer arithmetic, the floating-point
 `scale`, `S`, is realized using integer `multipler`, `M` and `shift` values each
-with bit width `W >= 2`, such that `round_nearest_even(S * 2^n) = M`, where `1
+with bit width `W >= 2`, such that `round_nearest_even(S * 2^n) == M`, where `1
 <= n <= 2*W - 2` and `0 <= M < 2^(W-1)`.
 
 The following demonstrates, using C++ code, a possible implementation of
@@ -204,7 +204,7 @@ floating-point scale `S` of type `f64`:
         adjustedMantissa = 0;
     }
 
-    // Saturate `adjustedMantissa`, if shift > 30,
+    // Saturate `adjustedMantissa`, if shift > 30.
     if (shift > 30) {
         shift = 30;
         adjustedMantissa = (1LL << 31) - 1;
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -178,7 +178,7 @@ following constraints:
 In a given StableHLO operation, all the quantized tensor types are either
 represented using floating-point scales, or integer scales, with multipliers and
 shifts. The floating-point scale `scale` value and the integer scale, with
-`multipler`, and `shift` values are related as: `round_nearest_even(scale *
+`multipler`, and `shift` values, are related as: `round_nearest_even(scale *
 2^shift) == multipler`.
 
 The following demonstrates, using C++ code, a possible implementation of
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -133,9 +133,10 @@ QuantizationExpressedType ::= FloatType
 QuantizationDimension ::= IntegerConstant
 QuantizationParameters ::= QuantizationParameter
                          | '{' QuantizationParameter {',' QuantizationParameter} '}'
-QuantizationParameter ::= ( QuantizationScale | QuantizationMultiplierShift ) ':' QuantizationZeroPoint
-QuantizationScale ::= FloatConstant
-QuantizationMultiplierShift ::= '<' QuantizationMultiplier ',' QuantizationShift '>'
+QuantizationParameter ::= QuantizationScale  ':' QuantizationZeroPoint
+QuantizationScale ::= ( QuantizationFloatingPointScale  | QuantizationIntegerScale )
+QuantizationFloatingPointScale ::= FloatConstant
+QuantizationIntegerScale ::= '<' QuantizationMultiplier ',' QuantizationShift '>'
 QuantizationMultiplier ::= IntegerConstant
 QuantizationShift ::= IntegerConstant
 QuantizationZeroPoint ::=  IntegerConstant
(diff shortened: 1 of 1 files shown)


docs/spec.md
```diff
@@ -160,10 +163,56 @@ following constraints:
 * (C12) If `quantization_dimension` is not empty, then
   `0 <= quantization_dimension`.
 
-At the moment, `QuantizationScale` is a floating-point constant, but there is
-strong interest in integer-based scales, represented with multipliers and
-shifts. We are planning to explore this in the near future
-([#1404](https://github.com/openxla/stablehlo/issues/1404)).
+In order to allow operation using only integer arithmetic, the floating-point
+`scale`, `S`, is realized using integer `multipler`, `M` and `shift` value,
+`n >= 0`, such that `round_nearest_integer(S) = (M + round)* 2^-n`, where
```
dominicsymes, line 168: I think it may be better to express the formula as:

M = round_nearest_integer(S*(2^n))

For example, if S=0.25 then we have M=(1<<30) and n=32. Then round_nearest_integer(S)=0 giving both sides as zero. round_nearest_inter(S*(2^n)) = (1<<30) =M gives a non zero comparison.

docs/spec.md
```diff
@@ -160,10 +163,56 @@ following constraints:
...
+    M = static_cast<int32_t>(adjustedMantissa);
+```
+
+Using the above implementation, the following C++ code demonstrates the
+relationship between the floating-point scale `S` and the integer parameters `M`
+and `n`.
+
+```c++
+    int64_t round = static_cast<int64_t>(1) << (n - 1);
+    static_cast<int32_t>(std::floor(S + 0.5)) = clamp(
```
dominicsymes, line 211: Similar to the comment on line 168, I think both sides become zero in the case of small S like 0.25 or lower. The left hand side could be changed to floor(S*(1<<n) + 0.5) and right hand size to M.

docs/spec.md
sdasgup3: That is a good point @dominicsymes. 

I agree with your point and incorporated in the spec. PTAL.
I note that, in order to maintain consistency with the rounding mode used in other quantized ops, I have used round_nearest_even consistently. The specification here demonstrates one way of using the rounding mode, other should be applicable as well.

docs/spec.md
sdasgup3: Also updated the code here. PTAL.

sdasgup3:
>> I think it would be useful to give an example

Thanks for the feedback. I will publish the updated spec soon.

docs/spec.md
```diff
@@ -160,10 +163,50 @@ following constraints:
 * (C12) If `quantization_dimension` is not empty, then
   `0 <= quantization_dimension`.
 
-At the moment, `QuantizationScale` is a floating-point constant, but there is
-strong interest in integer-based scales, represented with multipliers and
-shifts. We are planning to explore this in the near future
-([#1404](https://github.com/openxla/stablehlo/issues/1404)).
+In order to allow operation using only integer arithmetic, the floating-point
+`scale`, `S`, is realized using integer `multipler`, `M` and `shift` value,
+`n >= 0`, such that `round_nearest_even(S * 2^n) = M`.
```
dominicsymes, line 168: The n>=0 puts a lower bound on n but no constraint on the upper bound. The example code below produces n clipped to the range 1<=n<=62 and 0 <= M < (1<<31). I think it would be good if the specification could provide some constraints on the range of n and M such as these.

docs/spec.md
dominicsymes: Thanks @sdasgup3 for the update. I agree that round_nearest_even() is more consistent. The example code could also be updated to use this (I've added a comment below).

docs/spec.md
dominicsymes: Thanks - one new comment below on the updated line.

docs/spec.md
```diff
@@ -160,10 +163,50 @@ following constraints:
...
+
+    int64_t adjustedMantissa = shiftedMantissa;
+
+    // For small `S` value, flush `adjustedMantissa` to zero.
+    if (shift < -31) {
+        shift = 0;
+        adjustedMantissa = 0;
+    }
+
+    // Saturate `adjustedMantissa`, if shift > 30,
```
abattery, line 193: A minor comment: use period instead of the comma at the end of line

docs/spec.md
```diff
@@ -160,10 +163,50 @@ following constraints:
 * (C12) If `quantization_dimension` is not empty, then
   `0 <= quantization_dimension`.
 
-At the moment, `QuantizationScale` is a floating-point constant, but there is
-strong interest in integer-based scales, represented with multipliers and
-shifts. We are planning to explore this in the near future
-([#1404](https://github.com/openxla/stablehlo/issues/1404)).
+In order to allow operation using only integer arithmetic, the floating-point
+`scale`, `S`, is realized using integer `multipler`, `M` and `shift` value,
```
loganchien, line 167: Either (1) need to explicitly specify `int32_t` multiplier/shift or (2) need to update the syntax to allow different multiplier/shift type (and update the C++ code below).

docs/spec.md
```diff
@@ -160,10 +163,50 @@ following constraints:
...
+    n = (-shift) + 31;
+    M = static_cast<int32_t>(adjustedMantissa);
+```
+
+Using the above implementation, the following C++ code demonstrates the
+relationship between the floating-point scale `S` and the integer parameters `M`
+and `n`.
+
+```c++
+    static_cast<int32_t>(roundeven(M * std::pow(2, n))) = M;
```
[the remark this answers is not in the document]
loganchien: +1, I think this should be:

```
round_nearest_tie_to_even(S * std::pow(2, n)) == M
```

docs/spec.md
```diff
@@ -160,10 +163,50 @@ following constraints:
...
+`n >= 0`, such that `round_nearest_even(S * 2^n) = M`.
+
+The following demonstrates, using C++ code, a possible implementation of
+deriving the integer parameters `M` and `n` of type `i32` from a
+floating-point scale `S` of type `f64`:
+
+```c++
+    int32_t shift;
+    double mantissa = std::frexp(S, &shift);
+    auto shiftedMantissa = static_cast<int64_t>(std::round(mantissa * (int64_t(1) << 31)));
```
loganchien, line 177: General design question: is the constant 31 general to all ops?  Or, different ops (e.g. add vs. multiply) may want different fractional bits?

docs/spec.md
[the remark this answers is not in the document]

docs/spec.md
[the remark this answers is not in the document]
sdasgup3: Thanks for pointing that out! Updated the spec

docs/spec.md
sdasgup3: @dominicsymes That makes sense. 
I have added the bounds in this line only. PTAL.

docs/spec.md
sdasgup3: Hi @loganchien 

I have updated the above line to allow multiplier/shift with generic bit widths to provide a generic description of the conversion and the relationship between them. In the C++ code below, I have demonstrated the scale to (multiplier, shift) conversions using a specific bit width.

If we decide on making the code generic, then we might have to decide on a common type to store the values for the multiplier and shift. For example, both [computeMultiplierAndShiftTosaScale32](https://github.com/llvm/llvm-project/blob/94058c41d43a4aae0e0914b151c166f2729affbe/mlir/lib/Dialect/Tosa/Utils/QuantUtils.cpp#L58) and [computeMultiplierAndShiftTosaScale16](https://github.com/llvm/llvm-project/blob/94058c41d43a4aae0e0914b151c166f2729affbe/mlir/lib/Dialect/Tosa/Utils/QuantUtils.cpp#L22) uses `int32_t`  to store those values. For scaling other than 32-bit and 16-bits, it will be more of an implementation details. Please let me know your opinion.

One general question: Is 32-bit and 16-bit scaling (as show in the links above) the only ones supported while legalizing to TOSA? Or it could be completely arbitrary. 
cc @dominicsymes

dominicsymes: Thanks for the updates.

TOSA rescale supports 32-bit and 16-bit multiplier as you say. It seems unlikely to me that multipliers larger than 32-bit would be used since quantization is usually done to reduce data type size. So, 32-bit I would think is an upper bound. Multipliers smaller than 16-bit may be useful to consider. These can be legalized to the 16-bit case with a smaller value of the multiplier.

In terms of legalization to TOSA and comment from @loganchien below on operations, I think many operations will not use these multiplier and shift values directly, but a derived multiplier and shift specific to that operation. For example, for dot product the output scale will be based on (input1_scale * input2_scale / output_scale). This derivation may want to reduce the number of bits used, for example reducing from 32 to 16 bits.

In terms of example usage for an operation, such as dot product discussed around question 2. of the original comment, can this be considered here, or do you think this will be another PR?

sdasgup3: Thanks @dominicsymes for explaining. 

>> In terms of example usage for an operation, such as dot product discussed around question 2. of the original comment, can this be considered here, or do you think this will be another PR?

Yes, I am planning to consider it here in the current PR. I am currently looking into the details and planning to post the update asap.

sdasgup3: @dominicsymes 

Regarding your comment on: `I think it would be useful to give an example, perhaps based around the dot product operation of https://github.com/openxla/stablehlo/pull/1413 since in this operation only a single scale is needed of the final accumulator.`

I was considering a few options here. One is:

Given a dot.general op with floating-point scale as
```
%result = "stablehlo.dot_general"(%lhs, %rhs) {
  dot_dimension_numbers = #stablehlo.dot<
    lhs_batching_dimensions = [0],
    rhs_batching_dimensions = [0],
    lhs_contracting_dimensions = [2],
    rhs_contracting_dimensions = [1]
  >,
  precision_config = [#stablehlo<precision DEFAULT>, #stablehlo<precision DEFAULT>]
} : (tensor<2x2x2x!quant.uniform<i8<-127:127>:f32, 1.25:4>>,
     tensor<2x2x2x!quant.uniform<i8<-127:127>:f32, 1.25:4>>) ->
     tensor<2x2x2x!quant.uniform<i8<-127:127>:f32, 1.25:4>>
```

Applying the scale to 32-bit (multiplier,shift) conversion, as mentioned above, we have

```
%result = "stablehlo.dot_general"(%lhs, %rhs) {
  dot_dimension_numbers = #stablehlo.dot<
    lhs_batching_dimensions = [0],
    rhs_batching_dimensions = [0],
    lhs_contracting_dimensions = [2],
    rhs_contracting_dimensions = [1]
  >,
  precision_config = [#stablehlo<precision DEFAULT>, #stablehlo<precision DEFAULT>]
} : (tensor<2x2x2x!quant.uniform<i8<-127:127>:f32, <1342177280, 30>:4>>, 
     tensor<2x2x2x!quant.uniform<i8<-127:127>:f32, <1342177280, 30>:4>>) ->
     tensor<2x2x2x!quant.uniform<i8<-127:127>:f32, <1342177280, 30>:4>>
```

My intention was to show how the above conversion to this new type format can be used to enable all integer computation. I am not sure if rescale operations help here to capture that.  

Another options is to refer to the  semantic of `dot_general` op (quoted below) 

```
For is_quantized_tensor(lhs) and is_quantized_tensor(rhs):
  - integer_dot_product = reduce( inputs=[multiply((reshaped_lhs_slice - zero_point(reshaped_lhs_slice)), (reshaped_rhs_slice - zero_point(reshaped_rhs_slice))], init_values=[0], dimensions=[0, ..., size(lhs_contracting_dimensions) - 1], body=lambda x, y: add(x, y)).
  - rounded_dot_product = round_nearest_even(integer_dot_product * (scale(reshaped_lhs_slice) * scale(reshape_rhs_slice) / scale(result))).   
   - dot_product = clamp(storage_min(result), rounded_dot_product + zero_point(result), storage_max(result)).
```

and say that the expression
```
scale(reshaped_lhs_slice) * scale(reshape_rhs_slice) / scale(result)
```

can be expressed as the multiplier and shit. But, IMO, this does not demonstrate the benefit of this newly introduced  type. 

I think what I am presenting follow your comment " I think many operations will not use these multiplier and shift values directly, but a derived multiplier and shift specific to that operation. For example, for dot product the output scale will be based on (input1_scale * input2_scale / output_scale)."

Please let me know if I am missing something.

sdasgup3: Thanks @dominicsymes, a lot, for the explanation. This clarifies a lot of my dilemma. 

One thing to double check:
>> (numerator_M/denominator_M) * 2^(-total_n) = M * 2^(-n)

The operation `(numerator_M/denominator_M)`  should use integer arithmetic, right?  IMO, that would enable an all integer computation.

sdasgup3: Hi @dominicsymes 
  
I have attached the revised write-up. PTAL.
I have also added some clarity on a few things:

1. Can we have a mix of quantized types with floating-point and integer scales? I think the answer is No.

2. Given, the example we provided for `dot_general` (in the updated PR), where we have shown how the "derived floating-point scale" (which is `(scale(reshaped_lhs_slice) * scale(reshape_rhs_slice) / scale(result))`) is converted to "derived integer scale", one might be tempted to think that, for [addOp](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#add), the derived floating-point scale is

```
scale(lhs)/scale(result) for re-scaling lhs, and
scale(rhs)/scale(result) for re-scaling lhs.
```

which may not match what is done in [practice](https://groups.google.com/a/openxla.org/g/openxla-discuss/c/iwE9is49SS4/m/VTBpjwDrBAAJ) for TOSA re-scaling.  I remember that we already discussed about this,  [here](https://github.com/openxla/stablehlo/pull/1413#discussion_r1185585993), and I thought about clarifying it in the spec as well.

docs/spec.md
sdasgup3: Thanks for bring this up @loganchien. 

>>  different ops (e.g. add vs. multiply) may want different fractional bits?

I think the scaling bit-width of an operation depends on
 - operation itself (so not same constant 31 for all ops)
 - storage bitwidth of the output quantized type
 - implementation. 
 
 With that I temped to think that the bit-scaling to use for an individual  operation as "implementation defined."

@loganchien  @dominicsymes  What do you think?

docs/spec.md
```diff
@@ -2323,6 +2389,13 @@ More formally, `result[result_index] = dot_product`, where:
       dimensions=[0, ..., size(lhs_contracting_dimensions) - 1],
       body=lambda x, y: add(x, y))`.
   * `rounded_dot_product = round_nearest_even(integer_dot_product * (scale(reshaped_lhs_slice) * scale(reshape_rhs_slice) / scale(result)))`.
+    If the `scale(reshaped_lhs_slice)`, `scale(reshape_rhs_slice)`, and
+    `scale(result)` are respectively represented as integer multiplier and
+    shift pairs, like,
+    `(M_reshaped_lhs_slice, n_reshaped_lhs_slice), (M_reshaped_rhs_slice, n_reshaped_rhs_slice), and (M_result, n_result)`,
+    then the computation can be implemented as
+    `rounded_dot_product = (integer_dot_product * M + (1<<(n-1))) >> n`, where
+    `M = ((M_reshaped_lhs_slice * M_reshaped_rhs_slice) / M_result)) and n = (n_reshaped_lhs_slice + n_reshaped_rhs_slice - n_result)`.
```
dominicsymes, line 2398: I think there are some issues with doing this integer M divide directly as stated. It can overflow if for example M_reshaped_lhs_slice = M_reshaped_rhs_slice = (1<<31)-1 and M_result=(1<<30) then M=(1<<32)-4 which is too large for int32 range. It could also underflow and should ideally be rounded. Since the resulting scale is M * 2^(-n), overflow can be avoided if n>1 by reducing n (which reduces M). Similarly, underflow can be avoided by increasing n (if not at maximum n). So, I would suggest writing this as n = (n_reshaped_lhs_slice + n_reshaped_rhs_slice - n_result), M = scaling_divide(M_reshaped_lhs_slice*M_reshaped_rhs_slice, M_result, &n) where this function divides and adjusts for overflow and underflow by altering n.

docs/spec.md
dominicsymes: Hi @sdasgup3 

> The operation (numerator_M/denominator_M) should use integer arithmetic, right? IMO, that would enable an all integer computation.

Yes, it should be an integer divide but I think there are some issues involving overflow/underflow which I’ve commented on in the new patch.

For (1) do you mean within an operation or within a graph? Within an operation this seems unnecessary and would complicate. Within a graph it may be useful provided there are operations to convert between the two types.

For (2) yes, I think it is worth to clarify this as well for addOp. Scaling directly to result scale for add can lead to precision loss. For example, if output scale is 1.0 and input scales are 0.4 then doing sum of 0.4+0.4 at input scale will give 0.4 + 0.4 = 0.8 round to 1.0 at output scale. Doing sum at output scale would give 0 + 0 = 0. The [TFLite legalization](https://github.com/tensorflow/tensorflow/blob/60f7a770d64cb6cb3a93f84c291272ec51304d31/tensorflow/compiler/mlir/tosa/transforms/legalize_tfl.cc#L695-L700) discussed in the thread you point to chooses an add scale which does not overflow the integer add but also has extra precision. For example, the TFLite legalization does the add at scale 2*max(scale(lhs), scale(rhs))/(1<<n) where n=20 for 8-bit quantization and n=15 for 16-bit quantization. For these scales the add will not overflow 32 bits.

docs/spec.md
dominicsymes: I think there will be different conversions from these quantization parameters to the rescale parameters. For example to 16-bit or 32-bit rescale. However, I also think it would be useful to provide example/reference conversions as we are discussing above for dot product and for add operation.

docs/spec.md
rewu93: +1 on clarifying the scale issue in the add op

sdasgup3: Thanks @dominicsymes! 

>> Yes, it should be an integer divide but I think there are some issues involving overflow/underflow which I’ve commented on in the new patch.

Yes, I agree. Looking into it. Thanks!

>> For (1) do you mean within an operation or within a graph? Within an operation this seems unnecessary and would complicate. Within a graph it may be useful provided there are operations to convert between the two types.

I agree with you. 
Actually I meant at both the levels. At first I thought that mixing float vs int scales is equally bad at any level, but realized that in the current form we are specifying the semantics of an op when all the scales, of operands and outputs,  are the same (example add or dot_general ops). With that we cannot  afford having the mix within an operation, without adding addition complexity in specification (for example, what is the semantics of add with lhs and rhs scales are of different "types"). However, at program (graph) level the mixing seems OK.

>> For (2)

I agree with your point about the precision loss. 

Currently, I am trying to answer a few open questions related to the following (also open for discussion ) and get back asap. 
 
1. To what extend we should do the exemplification. All the ops (involving some numeric computation) ? 
2. Can we specify the  "derived scale" in a generic manner, without getting tied to a specific implementation (like tfl legalization to TOSA). Specifically in cases, like AddOp,when the derived integer scale is not intuitive from the derived floating-point scale.

docs/spec.md
sdasgup3: Thanks for the elaborate write up! That is super helpful.

One follow up on a detail: Given that a floating-point scale is always positive, and the way `M` is calculated in the PR, I assume that M, the multiplier, will be in the range `0 <= M < 2^(W-1)` with `W`-bit  storage. With that we can skip the underflow case, right? Let me know if I missing something.

docs/spec.md
```diff
@@ -153,17 +157,79 @@ following constraints:
...
+        adjustedMantissa = 0;
+    }
+
+    // Saturate `adjustedMantissa`, if exponent > 30.
+    if (exponent > 30) {
+        exponent = 30;
+        adjustedMantissa = (1LL << 31) - 1;
+    }
+
+    shift = (-exponent) + 31;
```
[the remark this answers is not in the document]
sdasgup3: Thanks for bringing this up. 
The file you point out just computes the shift value but did not adjust it as per the scaling bitwidth requirements. 

Let us have look at https://github.com/tensorflow/tensorflow/blob/8d81019452089bfdb0a04994b5d783316cd2bf28/tensorflow/lite/kernels/internal/quantization_util_test.cc#L429 to see how the computed shift value is multiplied with the double multiplier to get the quantized multiplier. 

Specifically at https://github.com/tensorflow/tensorflow/blob/8d81019452089bfdb0a04994b5d783316cd2bf28/tensorflow/lite/kernels/internal/common.h#L153 we will see that the shift is adjusted. 

Please let me know if that makes sense.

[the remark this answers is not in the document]
rewu93: That makes sense. @nutsiepully Do you know why we didn't have the adjusted shift at first?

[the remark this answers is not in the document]
nutsiepully: Is there a reason why we are leaking the implementation details of how the scale computation is done into the spec? Different hardware is bound to do this differently. TFLite just happens to do it via multipliers/shifts, but it's not necessarily guaranteed to be so in downstream hardware.

We discussed in a previous [rfc](https://github.com/openxla/stablehlo/blob/e38a4b6181ddbc42300ecf3feeac8bf07c369876/rfcs/20230210-stablehlo-tosa.md#tosa-quantized-implementation).

docs/spec.md
sdasgup3: Hi @dominicsymes 
Thanks for the above reply.
 
I think I still have a few open questions related to how one the internals of  the `scaling_divide`.
 
For the overflow case, 
Let us take same example, M_reshaped_lhs_slice = M_reshaped_rhs_slice = (1<<31)-1 and M_result=(1<<30)

Inside scaling_divide, if we  do the integer math with (M_reshaped_rhs_slice * M_reshaped_rhs_slice / M_result) we will get M = 4294967292. As the resulting scale is M * 2^(-n), we need to increase n. How is the value of n chosen? Is it by solving some inequality like Mx2^-n <= 2^31-1  or maybe some fixed value of n?


For the underflow case,if M_reshaped_lhs_slice = M_reshaped_rhs_slice = (1<<14) and M_result=(1<<30), M = 0.  
With that, the following integer computation is zero irrespective of the value on n. Maybe I am missing the math involved here. I want to better understand how  reducing n helps here.

 `rounded_dot_product = (integer_dot_product * M + (1<<(n-1))) >> n`

dominicsymes: Hi @sdasgup3 
The scale should be of the form M*2^(-n) where n is chosen so M does not overflow and has good precision. For example, to generate an M in range  0 <= M < (1<<31) the following should adjust n=(n_reshaped_lhs_slice + n_reshaped_rhs_slice - n_result) by decreasing to prevent overflow, increasing to prevent underflow:
- T = M_reshaped_lhs_slice * M_reshaped_rhs_slice
- while  T >= (M_result<<31) and n >= minimum_n then T=T/2; n=n-1;
- while  T < (M_result<<30) and n < maximum_n then T=T*2; n=n+1;
- M = T/M_result
- if n < minimum_n then saturate to maximum scale of M=(1<<31)-1, n=minimum_n

sdasgup3:
FYI: 
[revised](https://github.com/openxla/stablehlo/pull/1475/commits/85b337bf52d1ca9b1d70fcf7ee17b3ead92da787) the spec slightly to reflect that  the specification on how the integer multiplier and shift are going to be applied to every  op is under progress and to be decided soon.

docs/spec.md
sdasgup3: Let me propose to limit this PR with (1) Introduction of multipliers and shifts in the type signature of `QuantizedTensorType` and (2) Specifying a conversion from floating-point scales to integer multipliers and shifts.  

We will address the  the specification of individual ops (like `dot_general`, `add`) with integers multipliers and shifts  in a separate PR.

Let me know.

docs/spec.md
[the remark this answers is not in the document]
sdasgup3: Hi @nutsiepully 
The conversion defined in the spec is one possible implementation and we also clarified that aspect in the spec in the spec as well. This provides some basis to how this newly introduced multipliers and shifts is related to the floating-point scales.

docs/spec.md
dominicsymes: Hi @sdasgup3,

Your proposal makes sense. The discussion above is more specific to the operation (in this case dot_product) and I agree would be more easily handled in a separate PR relating to that operation.

docs/spec.md
```diff
@@ -170,11 +174,64 @@ following constraints:
...
-strong interest in integer-based scales, represented with multipliers and
-shifts. We are planning to explore this in the near future
-([#1404](https://github.com/openxla/stablehlo/issues/1404)).
+* (C13) `0 <= multipliers`.
+* (C14) `0 < shifts`.
+
+All quantized tensor types, in a given StableHLO operation, either
+use scales or multipliers and shifts. The `scales`, `multiplers` and `shifts`,
+are related as:
+`round_nearest_even(scales... * 2^shifts...) = multipler`.
```
dominicsymes, line 183: Should "multiplier" on the right hand side be "multipliers..." to match the list of scales and shifts on the left?

dominicsymes (approved):
This looks good to me as the basis of an integer only definition for quantization. As discussed in the comments, integer computation will also need conversion of the QuantizationMultiplier, QuantizationShift parameters to operation scales such as the dot_poduct output scale. I agree with the proposal from @sdasgup3 that this is probably best handled in separate PR specific to individual ops.

docs/spec.md
sdasgup3: Thanks @dominicsymes!
fixed the typo from `multiplier` -> `multipliers`.
Regarding the `...`, per [link](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#formulas)[B], it is used mainly in the context of `f(xs...)` to apply `f` on every element of `xs`. In this case, IMHO, it might not be needed as the `round_nearest_even(scales... * 2^shifts...)` with result in a list/tensor and it would suffice if the rhs `multipliers` is also of the same type. We have a related ticket https://github.com/openxla/stablehlo/issues/1589 to further tighten the notation to avoid any confusion around the use of `...`.

dominicsymes: Thanks for clarifying. I follow now why multipliers has no ... and that makes sense.

sdasgup3:
Given that introducing the new types would result in putting a lot of implementation details, related to rescale semantics, in the spec, we (in sync with @dominicsymes ) decided to skip that approach. We now believe it is better to declare quantized type numerics as implementation specific https://github.com/openxla/stablehlo/pull/2050.

With that I am closing this ticket here.

PR #1477, opened 2023-05-10
Specification for quantized ConvolutionOp

sdasgup3:
## Summary 
The PR proposes the specification for quantized convolution op along with the specifications for a few other ops on which the convolution specification depends on, for example, `pad`, `reverse`.

## A few details
1. Given `q = tensor with uniformed quantized type`, the PR covers the semantics of  static range quantized  `convolution(q, q)`. The specification is very similar to the earlier proposed [quantized dot_general op](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#dot_general).

2. Given, that many ops support both per-tensor and per-axis quantization schemes, it might be redundant to specify for each op the semantics of the computation involved in both the scheme. To avoid that we decided that the op-level semantics will only specify the per-tensor semantics, whereas, semantics of per-axis computations will be described generally in one place.

Please let me know your review feedback.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -3668,17 +3668,17 @@ More formally, `result[result_index]` is defined as:
 
 | Label | Name                | Type                                         | Constraints   |
 |-------|---------------------|----------------------------------------------|---------------|
-| (I1)  | `operand`           | tensor or quantized tensor                   | (C1), (C3-C6) |
-| (I2)  | `padding_value`     | 0-dimensional tensor or quantized tensor     | (C4)          |
-| (I3)  | `edge_padding_low`  | 1-dimensional tensor constant of type `si64` | (C1), (C3)    |
-| (I4)  | `edge_padding_high` | 1-dimensional tensor constant of type `si64` | (C1), (C3)    |
-| (I5)  | `interior_padding`  | 1-dimensional tensor constant of type `si64` | (C1-C3)       |
+| (I1)  | `operand`           | tensor or quantized tensor                   | (C1-C2), (C4) |
+| (I2)  | `padding_value`     | 0-dimensional tensor or quantized tensor     | (C1)          |
+| (I3)  | `edge_padding_low`  | 1-dimensional tensor constant of type `si64` | (C2), (C4)    |
+| (I4)  | `edge_padding_high` | 1-dimensional tensor constant of type `si64` | (C2), (C4)    |
+| (I5)  | `interior_padding`  | 1-dimensional tensor constant of type `si64` | (C2-C4)       |
 
 #### Outputs
 
-| Name     | Type                       | Constraints   |
-|----------|----------------------------|---------------|
-| `result` | tensor or quantized tensor | (C3-C4), (C6) |
+| Name     | Type                       | Constraints |
+|----------|----------------------------|-------------|
+| `result` | tensor or quantized tensor | (C1)        |
 
 #### Constraints
 
@@ -4330,16 +4330,16 @@ and produces a `result` tensor. More formally,
 
 #### Inputs
 
-| Label | Name         | Type                                         | Constraints      |
-|-------|--------------|----------------------------------------------|------------------|
-| (I1)  | `operand`    | tensor   or quantized tensor                 | (C3-C8)          |
-| (I2)  | `dimensions` | 1-dimensional tensor constant of type `si64` | (C1-C2), (C7-C8) |
+| Label | Name         | Type                                         | Constraints |
+|-------|--------------|----------------------------------------------|-------------|
+| (I1)  | `operand`    | tensor   or quantized tensor                 | (C1), (C3)  |
+| (I2)  | `dimensions` | 1-dimensional tensor constant of type `si64` | (C2), (C3)  |
 
 #### Outputs
 
 | Name     | Type                       | Constraints |
 |----------|----------------------------|-------------|
-| `result` | tensor or quantized tensor | (C2-C8)     |
+| `result` | tensor or quantized tensor | (C1), (C3)  |
 
 #### Constraints
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -3668,17 +3668,17 @@ More formally, `result[result_index]` is defined as:
 
 | Label | Name                | Type                                         | Constraints   |
 |-------|---------------------|----------------------------------------------|---------------|
-| (I1)  | `operand`           | tensor or quantized tensor                   | (C1-C2), (C4) |
-| (I2)  | `padding_value`     | 0-dimensional tensor or quantized tensor     | (C1)          |
-| (I3)  | `edge_padding_low`  | 1-dimensional tensor constant of type `si64` | (C2), (C4)    |
-| (I4)  | `edge_padding_high` | 1-dimensional tensor constant of type `si64` | (C2), (C4)    |
-| (I5)  | `interior_padding`  | 1-dimensional tensor constant of type `si64` | (C2-C4)       |
+| (I1)  | `operand`           | tensor or quantized tensor                   | (C1), (C3-C6) |
+| (I2)  | `padding_value`     | 0-dimensional tensor or quantized tensor     | (C4)          |
+| (I3)  | `edge_padding_low`  | 1-dimensional tensor constant of type `si64` | (C1), (C3)    |
+| (I4)  | `edge_padding_high` | 1-dimensional tensor constant of type `si64` | (C1), (C3)    |
+| (I5)  | `interior_padding`  | 1-dimensional tensor constant of type `si64` | (C1-C3)       |
 
 #### Outputs
 
-| Name     | Type                       | Constraints |
-|----------|----------------------------|-------------|
-| `result` | tensor or quantized tensor | (C1)        |
+| Name     | Type                       | Constraints   |
+|----------|----------------------------|---------------|
+| `result` | tensor or quantized tensor | (C3-C4), (C6) |
 
 #### Constraints
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -210,19 +210,6 @@ quantization parameters. Quantized tensor types have the following constraints:
   * (C12) `quantization_dimension < rank(self)`.
   * (C13) `dim(self, quantization_dimension) = size(scales)`.
 
-For simplicity, we only define individual operation semantics for per-tensor
-quantized type. Semantics of per-axis type, if supported by the op, can be
-derived as follows:
-
-Given an operation specification with per-tensor quantized type, let the
-semantics of the operation includes some application of the quantization
-parameters `zero_point` and `scale` to a tensor element`e` at index = `[i0, ...,
-id, ..., iR-1]`. The specification for the semantics of the same operation on
-tensor element `e`, with per-axis quantized type with quantization dimension `d`
-, would exactly be the same as the one with per-tensor type except that the
-`zero_point` and `scale` will be replaced with their per-axis counterparts,
-`zero_points[id]` and `scales[id]` resp.
-
 ```ebnf
 TokenType ::= 'token'
 ```
@@ -1974,8 +1961,8 @@ If `feature_group_count = 1` and `batch_group_count = 1`, then for all
 `output_spatial_index` in `index_space(dim(result, output_spatial_dimensions...))`,
 `result[result_shape(:, output_spatial_index, :)] = dot_product` where:
 
-* `padded_lhs_value = is_quantized(lhs)? zero_point(lhs) : 0`.
-* `padded_lhs = pad(lhs, padded_lhs_value, lhs_padding[:, 0], lhs_padding[:, 1], lhs_base_dilations - 1)`.
+* `padding_value = is_quantized_tensor(lhs) ? constant(zero_point(lhs), quantized_element_type(lhs)) : constant(0, element_type(lhs))`.
+* `padded_lhs = pad(lhs, padding_value, lhs_padding[:, 0], lhs_padding[:, 1], lhs_base_dilations - 1)`.
 * `lhs_window_start = lhs_shape(0, output_spatial_index, 0) * lhs_window_strides`.
 * `lhs_window = slice(padded_lhs, lhs_window_start, lhs_window_start + lhs_window_dimensions, lhs_window_dilations)`.
 * `reversed_lhs_window = reverse(lhs_window, [input_spatial_dimensions[dim] for dim in range(size(window_reversal)) if window_reversal[dim] = true])`.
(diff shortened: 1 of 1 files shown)


docs/spec.md
```diff
@@ -1988,7 +2001,8 @@ If `feature_group_count = 1` and `batch_group_count = 1`, then for all
 `output_spatial_index` in the index space of `dim(result, output_spatial_dimensions)`,
 `result[result_shape(:, output_spatial_index, :)] = dot_product` where:
 
-* `padded_lhs = pad(lhs, 0, lhs_padding[:, 0], lhs_padding[:, 1], lhs_base_dilations[:] - 1)`.
+* `padded_lhs_value = is_quantized_tensor(lhs)? zero_point(lhs) : 0`.
```
abattery, line 2004: A minor comment: add a space before the question mark:

* `padded_lhs_value = is_quantized_tensor(lhs) ? zero_point(lhs) : 0`.

loganchien (approved):

docs/spec.md
```diff
@@ -2093,8 +2106,21 @@ If `batch_group_count > 1`:
...
+* If the operation uses non-quantized tensors:
+  * (C27 `is_non_quantized_tensor(lhs) and is_non_quantized_tensor(rhs) and
+    is_non_quantized_tensor(result)`.
+  * (C28) element_type(`lhs`) $=$ element_type(`rhs`) $=$ element_type(`result`).
+* If the operation uses quantized tensors:
+  * (C29) `is_quantized_tensor(lhs) and is_quantized_tensor(rhs) and
+    is_quantized_tensor(result)`.
+  * (C30) `storage_type(lhs) =  storage_type(rhs)`.
+  * (C31) `expressed_type(lhs) = expressed_type(rhs) = expressed_type(result)`.
+  * (C32) `zero_points(rhs) = [0, 0, ..., 0]`.
```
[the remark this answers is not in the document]
sdasgup3: Thanks @loganchien!

Do you mind sharing a bit more about the use-case? Actually, given the close parity between the specs for convolution and dot_general, I am trying to evaluate if the use-case is applied to dot_general as well. In that case we can remove [C18](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#dot_general).

docs/spec.md
```diff
@@ -3690,19 +3715,19 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
-|-------|---------------------|----------------------------------------------|------------------|
-| (I1)  | `operand`           | tensor                                       | (C1), (C2), (C4) |
-| (I2)  | `padding_value`     | 0-dimensional tensor                         | (C1)             |
-| (I3)  | `edge_padding_low`  | 1-dimensional tensor constant of type `si64` | (C2), (C4)       |
-| (I4)  | `edge_padding_high` | 1-dimensional tensor constant of type `si64` | (C2), (C4)       |
-| (I5)  | `interior_padding`  | 1-dimensional tensor constant of type `si64` | (C2-C4)          |
+| Label | Name                | Type                                         | Constraints   |
+|-------|---------------------|----------------------------------------------|---------------|
+| (I1)  | `operand`           | tensor or quantized tensor                   | (C1-C2), (C4) |
+| (I2)  | `padding_value`     | 0-dimensional tensor or quantized tensor     | (C1)          |
```
eric-k256, line 3721: Sorry this is a bit pedantic, but Is padding_value a quantized tensor, or the zero_point of the quantized tensor? In Convolution, you are passing zero_point to pad(), but here it seems to imply you would send a tensor with single value 0 and a zero_point. 
I suppose the above comment where the value used is padding_value[] implies that the zero_point of a passed in padding_value is ignored.

docs/spec.md
```diff
@@ -199,6 +199,19 @@ quantization parameters. Quantized tensor types have the following constraints:
...
+derived as follows:
+
+Given an operation specification with per-tensor quantized type, let the
+semantics of the operation includes some application of the quantization
+parameters `zero_point` and `scale` to a tensor element`e` at index = `[i0, ...,
+id, ..., iR-1]`. The specification for the semantics of the same operation on
+tensor element `e`, with per-axis quantized type with quantization dimension `d`
+, would exactly be the same as the one with per-tensor type except that the
+`zero_point` and `scale` will be replaced with their per-axis counterparts,
+`zero_points[id]` and `scales[id]` resp.
```
eric-k256, line 213: Could it be easier to always use `zero_points` and `scales`, and if per-tensor quantization is used, then the size of zero_points/scales must be 1, and only zero_points/scales[0] would be used?

docs/spec.md
loganchien: Here is my understanding of `stablehlo.pad` op:

`stablehlo.pad` can pad whatever value specified in `pad_value`.  For example, I can decide to pad `5.0` at the boundary of the tensor.

The comment in the `stablehlo.conv` op is re-using the semantics of `stablehlo.pad` but specifies the zero point of the `stablehlo.conv` `input` operand as the pad value.

For example, if the input tensor is `tensor<4x4x!quant.uniform<ui8:f32, 1.0:5>>`, then one step of `stablehlo.conv` will call `stablehlo.pad(value = dense<5> : tensor<ui8:f32, 1.0:5>>)`, where `5` is the zero point.

Maybe need some clarification in the comment in `stablehlo.conv`.

docs/spec.md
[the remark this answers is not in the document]
loganchien: Yeah.  I think we should remove C18 from `dot_general` as well.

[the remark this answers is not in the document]
abattery: +1 for removing C32 constraint from here and other places, which is too strict. Asymmetric per-axis quantization is somewhat we would like to keep.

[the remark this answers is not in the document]
sdasgup3: Thanks for the feedback! fixed.

docs/spec.md
sdasgup3: Hi @eric-k256 , Thanks for bringing this up. I remember having the same concern while speccing this and tried to address it.

Thanks @loganchien for  capturing the intention here.

I guess part of the confusion  here is that in the convolution op spec, when 'lhs' is a quantized tensor,  we are using the spec of pad op as

```
`padded_lhs_value = is_quantized_tensor(lhs) ? zero_point(lhs) : 0`.
`padded_lhs = pad(lhs, padded_lhs_value, lhs_padding[:, 0], lhs_padding[:, 1], lhs_base_dilations[:] - 1)`.
```

But at this point it is not clear that the `padded_lhs_value` is  a quantized tensor with same element type as `lhs`.  

This is what I had in mind when I wrote this:

1. The fact that the `padded_lhs_value` is of the same element type as `lhs` can be derived from the constraint `C1` of the pad op

```
 (C1) `operand`, `padding_value`, `result` have the same element type.
```

2. I note here that even for the non quantized scenario, when we have
```
`padded_lhs_value =  0`. // as is_quantized_tensor(lhs) is false
`padded_lhs = pad(lhs, padded_lhs_value, lhs_padding[:, 0], lhs_padding[:, 1], lhs_base_dilations[:] - 1)`.
```

We are just passing the just the (zero) value to pad irrespective of the type of `lhs`. This `0` value is expected to be interpreted as a value of type equal to the `lhs`, which could be integer, floating-point, or complex type. 

>> I suppose the above comment where the value used is padding_value[] implies that the zero_point of a passed in padding_value is ignored.

That is true. From the POV of padding operation, specification should just consider the value ignoring the zero_point of the padding value. The zero_point here will just match with the zero_point of the  `operand` per `C1`.

@eric-k256 Please let me know if it makes sense.

docs/spec.md
sdasgup3: Hi @eric-k256 
Thanks for pointing that point. I remember that at some point I tried to consider the option and here was my findings:

For per-axis quantization with quantization axis  = d, an element at index `[i0, .., id,..., iR-1]` determines the  values for the `zero_points[id]` and `scales[id]` applied to it. Having said that if we use the per-axis as the default case, while specifying the op semantics, then the semantic need to capture the above relation, which might make the specification hard to follow.  

As an example, for AddOp,
```
* For quantized types:
  * `float_result = (lhs - zero_point(lhs)) * scale(lhs) +
    (rhs - zero_point(rhs)) * scale(rhs)`.
  * `rounded_result = round_nearest_even(float_result / scale(result))`.
  * `result = clamp(storage_min(result), rounded_result + zero_point(result), storage_max(result))`.
```
 
because we are  using per-tensor as default case, we can seamlessly use `scale(lhs)`  and `zero_point(lhs)` without referring to which `lhs` slice it is applied to.  

Please let me know if it makes sense.

docs/spec.md
eric-k256: Yes, this makes sense. Both you and Logan confirmed what I had assumed was the case in this instance.

docs/spec.md
eric-k256: Yes, it makes sense. It does describe the operation appropriately. Thanks for confirming.

eric-k256 (approved):
Thanks for the comments. The answers seem reasonable to me.

abattery (approved):

docs/spec.md
```diff
@@ -2084,8 +2097,20 @@ If `batch_group_count > 1`:
...
+* (C26) rank(`result`) $= N$.
+* If the operation uses non-quantized tensors:
+  * (C27 `is_non_quantized_tensor(lhs) and is_non_quantized_tensor(rhs) and
+    is_non_quantized_tensor(result)`.
+  * (C28) element_type(`lhs`) $=$ element_type(`rhs`) $=$ element_type(`result`).
+* If the operation uses quantized tensors:
+  * (C29) `is_quantized_tensor(lhs) and is_quantized_tensor(rhs) and
+    is_quantized_tensor(result)`.
+  * (C30) `storage_type(lhs) =  storage_type(rhs)`.
+  * (C31) `expressed_type(lhs) = expressed_type(rhs) = expressed_type(result)`.
```
rewu93, line 2109: I think hybrid kernel and mixed precision is a different topic, like in the dot general case?

docs/spec.md
```diff
@@ -2084,8 +2097,20 @@ If `batch_group_count > 1`:
...
+  * (C32) `quantization_dimension(lhs)` is empty.
+  * (C33) If `quantization_dimension(rhs)` is not empty, then `quantization_dimension(rhs) = kernel_output_feature_dimension`.
```
rewu93, line 2111: @paulinesho can you check these conditions?

docs/spec.md
sdasgup3: Yes, I also think so. "hybrid kernel" enables different interpretation of the kernel. The essence can  of hybrid kernel is discussed in some of the comment like [here](https://github.com/openxla/stablehlo/pull/1413#discussion_r1183127043). Mixed precision related to allowed types like bf13 and f32 to coexist.

docs/spec.md
```diff
@@ -2084,8 +2097,20 @@ If `batch_group_count > 1`:
...
+* If the operation uses non-quantized tensors:
+  * (C27 `is_non_quantized_tensor(lhs) and is_non_quantized_tensor(rhs) and
+    is_non_quantized_tensor(result)`.
+  * (C28) element_type(`lhs`) $=$ element_type(`rhs`) $=$ element_type(`result`).
+* If the operation uses quantized tensors:
+  * (C29) `is_quantized_tensor(lhs) and is_quantized_tensor(rhs) and
+    is_quantized_tensor(result)`.
+  * (C30) `storage_type(lhs) =  storage_type(rhs)`.
+  * (C31) `expressed_type(lhs) = expressed_type(rhs) = expressed_type(result)`.
+  * (C32) `quantization_dimension(lhs)` is empty.
```
paulinesho, line 2110: Just for clarification, does empty quantization_dimension(tensor) mean per-tensor only quantization?

docs/spec.md
paulinesho: LGTM

docs/spec.md
sdasgup3: That is right!

rewu93 (approved):

docs/spec.md
paulinesho: Relevant thread: https://github.com/openxla/stablehlo/pull/1413#discussion_r1186530590

paulinesho (approved):

docs/spec.md
```diff
@@ -2084,8 +2097,20 @@ If `batch_group_count > 1`:
     * `padded_input_shape[lhs_dim] = padding[spatial_dim, 0] + dilated_input_shape[lhs_dim] + padding[spatial_dim, 1]`.
     * `dilated_window_shape[lhs_dim] = dim(rhs, rhs_dim) == 0 ? 0 : (dim(rhs, rhs_dim) - 1) * rhs_dilation[spatial_dim] + 1`.
     * `num_windows = (padded_input_shape[lhs_dim] == 0 || dilated_window_shape[lhs_dim] > padded_input_shape[lhs_dim]) ? 0 : floor((padded_input_shape[lhs_dim] - dilated_window_shape[lhs_dim]) / window_strides[spatial_dim]) + 1`.
-* (C27) element_type(`result`) $=$ element_type(`lhs`).
-* (C28) rank(`result`) $= N$.
+* (C26) rank(`result`) $= N$.
+* If the operation uses non-quantized tensors:
+  * (C27 `is_non_quantized_tensor(lhs) and is_non_quantized_tensor(rhs) and
+    is_non_quantized_tensor(result)`.
```
burmako, line 2103: Doesn't C27 already follow from "If the operation uses non-quantized tensors"?

docs/spec.md
burmako: +1. We got to get to the bottom of this, although this doesn't have to be in this PR. We already have #369 to decide on mixed precision, so I raised its priority since it has repeatedly come up lately.

docs/spec.md
```diff
@@ -2084,8 +2097,20 @@ If `batch_group_count > 1`:
...
+  * (C28) element_type(`lhs`) $=$ element_type(`rhs`) $=$ element_type(`result`).
+* If the operation uses quantized tensors:
+  * (C29) `is_quantized_tensor(lhs) and is_quantized_tensor(rhs) and
+    is_quantized_tensor(result)`.
+  * (C30) `storage_type(lhs) =  storage_type(rhs)`.
+  * (C31) `expressed_type(lhs) = expressed_type(rhs) = expressed_type(result)`.
+  * (C32) `quantization_dimension(lhs)` is empty.
+  * (C33) If `quantization_dimension(rhs)` is not empty, then `quantization_dimension(rhs) = kernel_output_feature_dimension`.
+  * (C34) If `quantization_dimension(result)` is not empty, then `quantization_dimension(result) = output_feature_dimension`.
+  * (C35) If `quantization_dimension(rhs)` is empty, then `quantization_dimension(result)` is empty.
```
burmako, line 2113: Nit: Let's reorder constraints, so that they go: C35, C33, C34, similarly to how DotGeneralOp's constraints are ordered.

docs/spec.md
```diff
@@ -4359,14 +4383,14 @@ and produces a `result` tensor. More formally,
 
 | Label | Name         | Type                                         | Constraints |
 |-------|--------------|----------------------------------------------|-------------|
-| (I1)  | `operand`    | tensor                                       | (C1)        |
+| (I1)  | `operand`    | tensor   or quantized tensor                 | (C1), (C3)  |
```
burmako, line 4386: Typo: extraneous whitespace.

docs/spec.md
```diff
@@ -1979,7 +1992,8 @@ If `feature_group_count = 1` and `batch_group_count = 1`, then for all
 `output_spatial_index` in the index space of `dim(result, output_spatial_dimensions)`,
 `result[result_shape(:, output_spatial_index, :)] = dot_product` where:
 
-* `padded_lhs = pad(lhs, 0, lhs_padding[:, 0], lhs_padding[:, 1], lhs_base_dilations[:] - 1)`.
+* `padded_lhs_value = is_quantized_tensor(lhs) ? zero_point(lhs) : 0`.
```
burmako, line 1995: How about renaming to `padding_value`, instead of `padded_value`, to match the name of the corresponding input of PadOp?

burmako, line 1995: Also, what is the element type of `padding_value`? Let's express that explicitly by saying `padding_value = constant(is_quantized_tensor(lhs) ? zero_point(lhs) : 0, ???)`.

I see there was a conversation about this in https://github.com/openxla/stablehlo/pull/1477#discussion_r1194496076, and the conclusion was that "This 0 value is expected to be interpreted as a value of type equal to the lhs, which could be integer, floating-point, or complex type", but I think we should make this explicit to 100% avoid confusion.

docs/spec.md
burmako, line 213: As discussed offline, I don't think this definition works well beyond elementwise ops. 

To be more specific: we already have `ReshapeOp`, `SliceOp` and `TransposeOp` which support per-axis quantization at HEAD, and this PR proposes to add support for per-axis quantization to `PadOp` and `ReverseOp`. 

For all of these ops, this definition is insufficient. For example, what if a reshape affects the quantization dimension? Or what if pad pads the quantization dimension? Similarly for other ops called out above. It doesn't appear that we can have a uniform way of deriving per-axis semantics from per-tensor semantics.

Let's delete these paragraphs from this PR and sort this out separately - in #1574.

burmako (changes requested):

docs/spec.md
```diff
@@ -2084,8 +2097,20 @@ If `batch_group_count > 1`:
...
-* (C27) element_type(`result`) $=$ element_type(`lhs`).
-* (C28) rank(`result`) $= N$.
+* (C26) rank(`result`) $= N$.
+* If the operation uses non-quantized tensors:
+  * (C27 `is_non_quantized_tensor(lhs) and is_non_quantized_tensor(rhs) and
+    is_non_quantized_tensor(result)`.
+  * (C28) element_type(`lhs`) $=$ element_type(`rhs`) $=$ element_type(`result`).
+* If the operation uses quantized tensors:
+  * (C29) `is_quantized_tensor(lhs) and is_quantized_tensor(rhs) and
+    is_quantized_tensor(result)`.
```
burmako, line 2107: I also think that we should figure out a story for hybrid quantization, but let's do that separately from this PR which already represents a good amount of forward progress. I opened #1575 to keep track of this.

docs/spec.md
rewu93: +1. Per-axis quantization (and mixed precision) is important for precision demanding tasks. Having a clear spec is important

docs/spec.md
sdasgup3: acknowledged.

docs/spec.md
sdasgup3: That is right. fixed.

docs/spec.md

docs/spec.md
sdasgup3: done. PTAL.

docs/spec.md
sdasgup3: noted.

docs/spec.md
burmako: @paulinesho Thank you for the review and the approval! Resolving this conversation.

PR #1480, opened 2023-05-12
Split portable StableHLO APIs into separate file

GleasonK:
Portable APIs are APIs with signatures that do not depend on MLIR. This provides a way to access some StableHLO APIs without needing an MLIR dependency/visibility at the call site.

These APIs also can be safer in cases where shared objects are used, as passing MLIR Context across shared objects can cause problems.


stablehlo/integrations/python/StablehloModule.cpp
```diff
@@ -479,33 +479,14 @@ PYBIND11_MODULE(_stablehlo, m) {
       });
 
   //
-  // Utility APIs.
+  // Non-MLIR APIs
```
burmako, line 482: Let's call this "Portable APIs", since we already have a name for that?

burmako (approved):

PR #1484, opened 2023-05-13
Add limited support for multiple functions in shape refinement

burmako:
--stablehlo-refine-shapes currently has a limitation of only supporting one function to avoid dealing with complexities of potential loops in the dataflow graphs.

This PR slightly relaxes this limitation by not erroring out on multiple functions and instead refining shapes in just the `main` function among those (in case one doesn't exist, that would be an error).

MLIR-HLO commit: https://github.com/tensorflow/mlir-hlo/commit/36225413eab276eda72f936065d3ab361cf53889.

### DIFF
--- a/stablehlo/transforms/StablehloRefineShapes.cpp
+++ b/stablehlo/transforms/StablehloRefineShapes.cpp
@@ -1089,18 +1089,29 @@ struct StablehloRefineShapesPass
     // Current use cases are served well by inlining multiple functions into
     // a single function, so we leave native support for multiple functions to
     // future work.
+    // To enable modules that contain CustomCallOp::called_computations,
+    // we allow multiple functions, in which case we only refine the main
+    // function called "main", assuming that the called computations will have
+    // static shapes. Lifting this assumption and expanding refinement to
+    // multiple functions is left for future work.
     ModuleOp module = getOperation();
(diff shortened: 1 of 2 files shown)


stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -1089,18 +1089,30 @@ struct StablehloRefineShapesPass
...
+    // function called "main", assuming that the called computations will have
+    // static shapes. Lifting this assumption and expanding refinement to
+    // multiple functions is left for future work.
     ModuleOp module = getOperation();
     auto funcs = llvm::to_vector(module.getOps<func::FuncOp>());
     if (funcs.empty()) return;
-    if (funcs.size() != 1) {
-      module.emitOpError() << "must have no more than one function";
+    func::FuncOp func;
+    if (funcs.size() == 1) {
```
GleasonK, line 1101: Is there no standard name for a single function? Any way to restrict this to only refine `@main`?

GleasonK (approved):

stablehlo/transforms/StablehloRefineShapes.cpp
burmako: Will be addressed in https://github.com/openxla/stablehlo/issues/1494. For now, merging this PR to simplify integrate.

PR #1485, opened 2023-05-13
Fix issues identified during integrate

burmako:
* BUILD.bazel & CMakeLists.txt: add a missing dependency.
  * stablehlo/dialect/TypeInference.cpp, stablehlo/dialect/TypeInference.h, stablehlo/reference/Ops.cpp, stablehlo/reference/Tensor.h: fix clang-tidy warnings.
  * stablehlo/integrations/python/tests/stablehlo.py: fix formatting.

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -376,8 +376,8 @@ verifyWindowAttributesAndInferWindowDimensions(
 //            padding_high,
 //                         dilatedBound(window_shape[d]))
 //      where (padding_low, padding_high) is the padding-pair for d.
-SmallVector<int64_t> inferWindowOutputShape(
-    const ArrayRef<int64_t> baseShape, const ArrayRef<WindowDimension> window) {
+SmallVector<int64_t> inferWindowOutputShape(ArrayRef<int64_t> baseShape,
+                                            ArrayRef<WindowDimension> window) {
   assert(baseShape.size() == window.size() &&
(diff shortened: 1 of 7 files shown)


GleasonK (approved):

stablehlo/dialect/TypeInference.cpp
```diff
@@ -376,8 +376,8 @@ verifyWindowAttributesAndInferWindowDimensions(
 //            padding_high,
 //                         dilatedBound(window_shape[d]))
 //      where (padding_low, padding_high) is the padding-pair for d.
-SmallVector<int64_t> inferWindowOutputShape(
-    const ArrayRef<int64_t> baseShape, const ArrayRef<WindowDimension> window) {
+SmallVector<int64_t> inferWindowOutputShape(ArrayRef<int64_t> baseShape,
```
GleasonK, line 379: Looks like this issue `const ArrayRef` has existed for awhile, any reason we are just catching this now? Any preventative steps we should take?

burmako: I came across this issue after receiving a ClangTidy warning during the integrate. I think that happened because the integrate touched TypeInference.{cpp,h}, so the presubmits decided to check those files with ClangTidy.

One way to avoid these issues in the future is to codify fixing ClangTidy warnings during integrates (I don't think this is a clear-cut part of the integrate process at the moment). Another way is to run ClangTidy as part of GitHub actions, maybe nightly or something if this ends up being too slow.

PR #1278, opened 2023-03-06
Fix for float diff error on Mac platform

sdasgup3:
fixes https://github.com/openxla/stablehlo/issues/1275

Please refer to https://github.com/openxla/stablehlo/issues/1275 for the motivation behind the current change.

PR #1492, opened 2023-05-16
Update spec and interpreter for ReduceWindowOp to properly calculate window_end

ghpvnist:
The current spec does not use `slice` with the same semantics as what `stablehlo.slice` does. `stablehlo.slice` version of `limit_indices` accounts for dilation and also complies with SliceOp's constraint: `(C3) 0 <= start_indices[d] <= limit_indices[d] <= dim(operand, d) for all dimension d.`

ReduceWindowOp's `limit_indices` calculation does not account for the dilation and (C3), so we need to add dilation (minus 1 since limit_indices is exclusive).

This was discussed previously in the interpreter PR https://github.com/openxla/stablehlo/pull/1336#discussion_r1142826672. We kept the formulation as-is to assume that the program provides correct values, but we may not assume this is the case when we are explicitly constructing the input values.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -4131,8 +4131,8 @@ where:
 * `padded_inputs = pad(inputs..., init_values..., padding[:, 0], padding[:, 1],
   base_dilations - 1)`.
 * `window_start = result_index * window_strides`.
-* `window_end = clamp(0, window_start + window_dimensions * window_dilations,
-  shape(inputs[0]))`.
+* `window_end = clamp(0, window_start + window_dimensions + window_dilations -
+  1, shape(inputs[0]))`.
 * `windows = slice(padded_inputs..., window_start, window_end,
   window_dilations)`.
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -4131,7 +4131,8 @@ where:
 * `padded_inputs = pad(inputs..., init_values..., padding[:, 0], padding[:, 1],
   base_dilations - 1)`.
 * `window_start = result_index * window_strides`.
-* `window_end = window_start + window_dimensions`.
+* `window_end = clamp(0, window_start + window_dimensions * window_dilations,
+  shape(inputs[0]))`.
 * `windows = slice(padded_inputs..., window_start, window_end,
   window_dilations)`.
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -4131,8 +4131,8 @@ where:
 * `padded_inputs = pad(inputs..., init_values..., padding[:, 0], padding[:, 1],
   base_dilations - 1)`.
 * `window_start = result_index * window_strides`.
-* `window_end = clamp(0, window_start + window_dimensions + window_dilations -
-  1, shape(inputs[0]))`.
+* `window_end = clamp(window_start, window_start + window_dimensions +
+  window_dilations - 1, shape(padded_inputs[0]))`.
 * `windows = slice(padded_inputs..., window_start, window_end,
   window_dilations)`.
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -4131,8 +4131,7 @@ where:
 * `padded_inputs = pad(inputs..., init_values..., padding[:, 0], padding[:, 1],
   base_dilations - 1)`.
 * `window_start = result_index * window_strides`.
-* `window_end = clamp(window_start, window_start + window_dimensions +
-  window_dilations - 1, shape(padded_inputs[0]))`.
+* `window_end = window_start + window_dimensions + window_dilations - 1`.
 * `windows = slice(padded_inputs..., window_start, window_end,
   window_dilations)`.
 
--- a/stablehlo/reference/Ops.cpp
+++ b/stablehlo/reference/Ops.cpp
@@ -1286,7 +1286,7 @@ SmallVector<Tensor> evalReduceWindowOp(
     SmallVector<Tensor> windows;
     auto windowStart = (*resultIt) * windowStrides;
     auto windowEnd =
-        clamp(0, windowStart + windowDimensions + windowDilations - 1,
+        clamp(windowStart, windowStart + windowDimensions + windowDilations - 1,
               paddedInputs[0].getShape());
     for (const auto &paddedInput : paddedInputs)
       windows.push_back(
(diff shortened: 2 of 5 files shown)


docs/spec.md
```diff
@@ -4173,7 +4173,8 @@ where:
 <!-- markdownlint-disable line-length -->
 * `padded_inputs = pad(inputs[:], init_values[:], padding[:, 0], padding[:, 1], base_dilations[:] - 1)`.
 * `window_start = result_index * window_strides`.
-* `windows = slice(padded_inputs[:], window_start, window_start + window_dimensions, window_dilations)`.
+* `window_end = clamp(0, window_start + window_dimensions * window_dilations, shape(inputs[0]))`.
```
sdasgup3, line 4176: Two comments 
1. Before applying the `window_dimension` to the padded input, it can be grown using `window_dimaltion`, by `window_dilation -1`. That is `window_start =  window_start + window_dimensions  +  window_dilations - 1`. Can you explain why are we multiplying? 
2. Should we consider `padded_input`'s dimensions for clamping? Not sure why using `shape(inputs[0])`  ?

ghpvnist: 1. Yes you're right, should be adding not multiplying.
2. `window_start` contain the index relative to the `paddedInput`. Similarly, `window_end` should contain the index relative to the `padded_input`. The `SliceOp` here would be slicing a window defined by its start index (`window_start`) and limit index (`window_end`) out of the `padded_input`.

sdasgup3: Re: On 2. Should we consider padded_input's dimensions for clamping? Not sure why using shape(inputs[0]) ?

>>Similarly, window_end should contain the index relative to the padded_input. 

In that case,  why not clamp it using the limits of `padded_input`?

Also as per slice semantics, `limit_indices >= start_indices`, so the clamp should be

`window_end = clamp(window_start, window_start + window_dimensions  + window_dilations - 1, shape(inputs[0]))`, right?

ghpvnist: I see what you mean, yes `window_end` should be (corrected typo on the shape):
`window_end = clamp(window_start, window_start + window_dimensions  + window_dilations - 1, shape(padded_inputs[0]))`

The max clamping part was reflected in the implementation but not the spec, which is updated now along with the min clamping on both.

stablehlo/testdata/asinh_shape_float16_20_20.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN-DISABLED: stablehlo-opt -inline %s | stablehlo-translate --interpret
+// RUN-DISABLED(#1278): stablehlo-opt -inline %s | stablehlo-translate --interpret
```
burmako, line 1: Let's factor the change that adds `#1278` out of this PR, so that it focuses on just one thing?

docs/spec.md
burmako: The PR description still mentions multiplication. Let's update it as well?

stablehlo/reference/Ops.cpp
```diff
@@ -1280,10 +1280,12 @@ SmallVector<Tensor> evalReduceWindowOp(
...
+              paddedInputs[0].getShape());
     for (const auto &paddedInput : paddedInputs)
-      windows.push_back(evalSliceOp(paddedInput, windowStart,
-                                    windowStart + windowDimensions,
-                                    windowDilations));
+      windows.push_back(
+          evalSliceOp(paddedInput, windowStart, windowEnd, windowDilations));
 
     auto reducedValues =
         evalReduceOp(windows, initValues, inputs[0].getAxes(), body, scope);
```
burmako, line 1291: What's our process for updating the interpreter? Does it also include updating the corresponding `interpret_foo.mlir` files?

stablehlo/testdata/gather_from_take_indices_name__5_oob__axis_0_enable_xla_True_mode_fill.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN-DISABLED: stablehlo-opt -inline %s | stablehlo-translate --interpret
```
burmako, line 1: I see that most of the enabled tests pass even with `auto windowEnd = windowStart + windowDimensions;`. The only failing tests are:

```
Failed Tests (3):
  STABLEHLO_TESTS_SUITE :: reduce_window_add_window_dilation_shape_float32_4_6__initvalue_0_windowdimensions__2_2__windowstride1490052185320441557.mlir
  STABLEHLO_TESTS_SUITE :: select_and_gather_add_dilations_shape_float32_4_6__selectprim_le_windowdimensions__2__2__windowstrid-1234442230378059557.mlir
  STABLEHLO_TESTS_SUITE :: select_and_gather_add_dilations_shape_float32_4_6__selectprim_le_windowdimensions__2__2__windowstrid4494141771612406334.mlir
```

Let's only touch these three failing tests in this PR, and modify the other testdata/ tests in two separate PRs - one that adds a reference to 1278, and another one that enables the tests which are already passing.

burmako (changes requested):

burmako:
@ghpvnist I reassigned this PR to myself earlier today, but I think the ball's in your court at the moment. Please take a look.

stablehlo/testdata/asinh_shape_float16_20_20.mlir
ghpvnist: Sure, created #1597

stablehlo/testdata/gather_from_take_indices_name__5_oob__axis_0_enable_xla_True_mode_fill.mlir
ghpvnist: Ah I just saw this. Created a single PR to address the comment #1597.

docs/spec.md
ghpvnist: Thanks, updated the description.

stablehlo/reference/Ops.cpp
ghpvnist: The process would be the same (aka also change `interpret_foo.mlir` files if touched), but since this PR does not change the tests, I've excluded them initially. It would be prudent to update the spec example to include clamping and thus also update this test file.

burmako (approved):

PR #1496, opened 2023-05-18
 Specification for UniformQuantizeOp and UniformDequantizeOp

### LINKED ISSUE #530: Add spec for UniformDequantizeOp
### Request description

To spec `stablehlo.uniform_dequantize`

### LINKED ISSUE #531: Add spec for UniformQuantizeOp
### Request description

To spec `stablehlo.uniform_quantize`

sdasgup3:
fixes #531
fixes #530 

## Summary 
The PR proposes the specification for `uniform.quantize` and `uniform.dequantize` ops. 

The specification of `uniform.quantize` also captures the re-quantization conversions from quantized tensor to quantized tensors. 

#### uniform_dequantize

We have the following constraints from the spec  

```
(I1)  `operand` is a  quantized tensor
(C1) `shape(operand) = shape(result)`.
(C2) `element_type(result) = expressed_type(operand)`.
```

```
I1: a) `operand: quantized tensors`. (Covered by ODS).
C1: a) `shape(operand) != shape(result)`. (Covered by ODS)
C2: a) `element_type(result) != expressed_type(operand)`.
```

```
C2: a) `element_type(result) != expressed_type(operand)`.
```

We already has a type inference test to cover the above. 

#### uniform_quantize

We have the following constraints from the spec  

```
(I1)  `operand: tensor of floating-point or quantized type`.
(C1) `shape(operand) = shape(result)`.
(C2) `expressed_type(result) = is_float(operand) ? element_type(operand) :
  expressed_type(operand)`.
```

```
I1: a) `operand: quantized tensors`. (Covered by ODS).
C1: a) `shape(operand) != shape(result)`. (Covered by ODS)
C2: a) if_float(operand): `expressed_type(result) != element_type(operand)`.
      b) if_quantized(operand): `expressed_type(result) !=  expressed_type(operand)`.
```

```
C2: a) if_float(operand): `expressed_type(result) != element_type(operand)`.
      b) if_quantized(operand): `expressed_type(result) !=  expressed_type(operand)`.
```

The above will be covered as part of https://github.com/openxla/stablehlo/issues/1603.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -5555,7 +5555,7 @@ Formally, `result = dequantize(operand)`.
 #### Constraints
 
 * (C1) `shape(operand) = shape(result)`.
-* (C2) `expressed_type(result) = element_type(operand)`.
+* (C2) `element_type(result) = expressed_type(operand)`.
 
 #### Examples
 
@@ -6212,6 +6212,41 @@ def baseline_type(x: Value | Placeholder | Type) -> Type:
     return baseline_element_type(type(x))
 ```
 
+* `dequantize` is defined on quantized tensor types and turns them into
+floating-point tensor types. This happens via converting quantized elements
+which represent integer values of the storage type into corresponding
+floating-point values of the expressed type using the zero point and scale
+associated with the quantized element type. At the moment, this function only
+works for per-tensor quantization. Per-axis quantization is work in progress
+([#1574](https://github.com/openxla/stablehlo/issues/1574)).
+
+```python
+def dequantize(x: Value) -> Value:
+  assert is_quantized(x)
+  x_storage = bitcast_convert(x, storage_type(x))
+  x_storage_sub = x_storage - zero_point(x_storage)
+  x_expressed_sub = convert(x_storage_sub, expressed_type(x))
+  return x_expressed_sub * scale(x)
+```
+
+* `quantize` is defined on floating-point tensor types and turns them into
+quantized tensor types. This happens via converting floating-point values
+of the expressed type into corresponding integer values of the storage type
+using the zero point and scale associated with the quantized element type.
+At the moment, this function only works for per-tensor quantization. Per-axis
+quantization is work in progress
+([#1574](https://github.com/openxla/stablehlo/issues/1574)).
+
+```python
+def quantize(x: Value, type: Type) -> Value:
+  assert is_float(x) and is_quantized(type)
+  x_expressed_rounded = round_nearest_even(x / scale(type))
+  x_storage_rounded = convert(x_expressed_rounded, storage_type(type))
+  x_storage_add = x_storage_rounded + zero_point(type)
+  x_storage = clamp(storage_min(type), x_storage_add, storage_max(type))
+  return bitcast_convert(x_storage, type)
+```
+
 * `dequantize_op_quantize` is used to specify element-wise computations on
 quantized tensors. It dequantizes, i.e. turns quantized elements into their
 expressed types, then performs an operation, and then quantizes, i.e. turns
@@ -6220,19 +6255,11 @@ works for per-tensor quantization. Per-axis quantization is work in progress
 ([#1574](https://github.com/openxla/stablehlo/issues/1574)).
 
 ```python
-
-def dequantize(x):
-  return (x - zero_point(x)) * scale(x)
-
-def quantize(x, type):
-  rounded_result = round_nearest_even(x / scale(type))
-  return clamp(storage_min(type), rounded_result + zero_point(type), storage_max(type))
-
 def dequantize_op_quantize(op, *inputs_and_output_type):
   inputs = inputs_and_output_type[:-1]
   output_type = inputs_and_output_type[-1]
 
-  float_inputs = [dequantize(x) for x in inputs]
+  float_inputs = map(dequantize, inputs)
   float_result = op(*float_inputs)
   return quantize(float_result, output_type)
 ```
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -5538,7 +5538,7 @@ Performs element-wise conversion of quantized tensor `operand` to a
 floating-point tensor `result` according to the quantization parameters defined
 by the `operand` type.
 
-Formally, `result = dequantize(operand)`.
+More formally, `result = dequantize(operand)`.
 
 #### Inputs
 
@@ -5573,7 +5573,7 @@ Performs element-wise conversion of floating-point tensor or quantized tensor
 `operand` to a quantized tensor `result` according to the quantization
 parameters defined by the `result` type.
 
-Formally,
+More formally,
 
 * If `is_float(operand)`:
   * `result = quantize(operand, type(result))`.
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -5535,7 +5535,7 @@ Produces a `result` tuple from values `val`.
 #### Semantics
 
 Performs element-wise conversion of uniform quantized tensor `operand` to a
-floating point tensor `result` according to the quantization parameters defined
+floating-point tensor `result` according to the quantization parameters defined
 by the `operand` type.
 
 Formally, `result = (operand - zero_point(operand)) * scale(operand)`.
@@ -5554,14 +5554,14 @@ Formally, `result = (operand - zero_point(operand)) * scale(operand)`.
 
 #### Constraints
 
-* (C1) `expressed_type(operand) = element_type(result)`.
-* (C2) `shape(operand) = shape(result)`.
+* (C1) `shape(operand) = shape(result)`.
+* (C2) `expressed_type(result) = element_type(operand)`.
 
 #### Examples
 
 ```mlir
 // %operand: [10, 10]
-%result = "stablehlo.uniform_dequantize"(%operand) : (tensor<2x!quant.uniform<i8<-128:127>:f32:0, {0.1:-30, 0.5:-20}>>) -> tensor<2xf32>
+%result = "stablehlo.uniform_dequantize"(%operand) : (tensor<2x!quant.uniform<i8:f32:0, {0.1:-30,0.5:-20}>>) -> tensor<2xf32>
 // %result: [4.0, 15.0]
 ```
 
@@ -5575,45 +5575,40 @@ quantization parameters defined by the `result` type.
 
 Formally,
 
-* For `element_type(operand)` a floating-point type,
+* If `is_float(operand)`:
   * `rounded_result = round_nearest_even(operand / scale(result))`.
   * `result = clamp(storage_min(result), rounded_result + zero_point(result), storage_max(result))`.
-
-* For `element_type(operand)` a quantized type,
-  * `float_result = (operand - zero_point(operand)) * scale(operand)`.
-  * `rounded_result = round_nearest_even(float_result / scale(result))`.
-  * `result = clamp(storage_min(result), rounded_result + zero_point(result), storage_max(result))`.
+* If `is_quantized(operand)`: performs `dequantize_op_quantize(lambda operand:
+        operand, operand, type(result))`
 
 #### Inputs
 
-| Label | Name      | Type                                        | Constraints      |
-|-------|-----------|---------------------------------------------|------------------|
-| (I1)  | `operand` | tensor of floating-point or quantized  type | (C1), (C2), (C3) |
+| Label | Name      | Type                                        | Constraints |
+|-------|-----------|---------------------------------------------|-------------|
+| (I1)  | `operand` | tensor of floating-point or quantized  type | (C1), (C2)  |
 
 #### Outputs
 
-| Name     | Type             | Constraints      |
-|----------|------------------|------------------|
-| `result` | quantized tensor | (C1), (C2), (C3) |
+| Name     | Type             | Constraints |
+|----------|------------------|-------------|
+| `result` | quantized tensor | (C1), (C2)  |
 
 #### Constraints
 
-* (C1) If `element_type(operand)` is a floating-point type,
-  * `element_type(operand) = expressed_type(result)`.
-* (C2) If `element_type(operand)` is a quantized type,
-  * `expressed_type(operand) = expressed_type(result)`.
-* (C3) `shape(operand) = shape(result)`.
+* (C1) `shape(operand) = shape(result)`.
+* (C2) `expressed_type(result) = is_float(operand) ? element_type(operand) :
+  expressed_type(operand)`.
 
 #### Examples
 
 ```mlir
 // %operand: [4.0, 15.0]
-%result = "stablehlo.uniform_quantize"(%operand) : (tensor<2xf32>) -> tensor<2x!quant.uniform<i8<-128:127>:f32:0, {0.1:-30, 0.5:-20}>>
+%result = "stablehlo.uniform_quantize"(%operand) : (tensor<2xf32>) -> tensor<2x!quant.uniform<i8:f32:0, {0.1:-30,0.5:-20}>>
 // %result: [10, 10]
 
 // %operand: [10, 10]
-%result = "stablehlo.uniform_quantize"(%operand) : (tensor<2x!quant.uniform<i8<-128:127>:f32:0, {0.1:-30, 0.5:-20}>>) -> tensor<2x!quant.uniform<i8<-128:127>:f32:0, {0.1:-30, 0.5:-20}>>
-// %result: [10, 10]
+%result = "stablehlo.uniform_quantize"(%operand) : (tensor<2x!quant.uniform<i8:f32:0, {0.1:-30,0.5:-20}>>) -> tensor<2x!quant.uniform<i8:f32:0, {0.1:-20,0.2:-30}>>
+// %result: [20, 45]
 ```
 
 ### while
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -5534,11 +5534,11 @@ Produces a `result` tuple from values `val`.
 
 #### Semantics
 
-Performs element-wise conversion of uniform quantized tensor `operand` to a
+Performs element-wise conversion of quantized tensor `operand` to a
 floating-point tensor `result` according to the quantization parameters defined
 by the `operand` type.
 
-Formally, `result = (operand - zero_point(operand)) * scale(operand)`.
+Formally, `result = dequantize(operand)`.
 
 #### Inputs
 
@@ -5569,23 +5569,23 @@ Formally, `result = (operand - zero_point(operand)) * scale(operand)`.
 
 #### Semantics
 
-Performs element-wise conversion of floating-point tensor or uniform quantized
-tensor `operand` to a uniform quantized tensor `result` according to the
-quantization parameters defined by the `result` type.
+Performs element-wise conversion of floating-point tensor or quantized tensor
+`operand` to a quantized tensor `result` according to the quantization
+parameters defined by the `result` type.
 
 Formally,
 
 * If `is_float(operand)`:
-  * `rounded_result = round_nearest_even(operand / scale(result))`.
-  * `result = clamp(storage_min(result), rounded_result + zero_point(result), storage_max(result))`.
-* If `is_quantized(operand)`: performs `dequantize_op_quantize(lambda operand:
-        operand, operand, type(result))`
+  * `result = quantize(operand, type(result))`.
+* If `is_quantized(operand)`:
+  * `float_result = dequantize(operand)`.
+  * `result = quantize(float_result, type(result))`.
 
 #### Inputs
 
-| Label | Name      | Type                                        | Constraints |
-|-------|-----------|---------------------------------------------|-------------|
-| (I1)  | `operand` | tensor of floating-point or quantized  type | (C1), (C2)  |
+| Label | Name      | Type                                       | Constraints |
+|-------|-----------|--------------------------------------------|-------------|
+| (I1)  | `operand` | tensor of floating-point or quantized type | (C1), (C2)  |
 
 #### Outputs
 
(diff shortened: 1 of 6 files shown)


docs/spec.md
```diff
@@ -5570,6 +5565,89 @@ Produces a `result` tuple from values `val`.
...
+| Name     | Type                     | Constraints      |
+|----------|--------------------------|------------------|
+| `result` | quantized tensor | (C1), (C2), (C3) |
+
+#### Constraints
+
+* (C1) If `element_type(operand)` is a floating-point type,
+  * `element_type(operand) = expressed_type(result)`.
+* (C2) If `element_type(operand)` is a quantized type,
+  * `num_bits(storage_type(operand)) >= num_bits(storage_type(result))`.
```
dominicsymes, line 5639: This constraint seems to allow conversion from 16-bit quantized type to 8-bit quantized type but not allow conversion of 8-bit quantized type to 16-bit quantized type. Both directions seem useful. Is this constraint needed?

loganchien: +1 agreed.  Both directions have use cases.

docs/spec.md
```diff
@@ -5570,6 +5565,89 @@ Produces a `result` tuple from values `val`.
...
 
+### uniform_dequantize
+
+#### Semantics
+
+Performs element-wise conversion of uniform quantized tensor `operand` to a
+floating point tensor `result` according to the quantization parameters defined
+by the `operand` type.
+
+Formally, `result = (operand - zero_point(operand)) * scale(operand)`.
```
loganchien, line 5576: Btw, I don't find the definition of `scale(...)` and `zero_point(...)` in the Types section.  Do these function implicitly handle the `quantization_dimension`?

Also, please add a per-axis example below.

loganchien (approved):
Overall LGTM, which a few minor comments.

docs/spec.md
sdasgup3: Thanks for the feedback! Will be incorporated in the next commit.

docs/spec.md
sdasgup3: >> Btw, I don't find the definition of scale(...) and zero_point(...) in the Types sectiIn some parts of the spec document we are informally using some macros like `scale(...), zero_point(...), quantization_dimension(...)`. This treatment is similar to the usages of other macros like like `element_type(...), rank(...), shape(...)`.    We have a [ticket](https://github.com/openxla/stablehlo/issues/289) to augment the  specification doc with the details of all of these.  I have added an action item to address the quantization related macros  https://github.com/openxla/stablehlo/issues/289#issuecomment-1559823158.


>> Also, please add a per-axis example below.

Done.

I note that, as per the [guideline](https://github.com/openxla/stablehlo/blob/main/docs/spec_checklist.md#stablehlo-specification-checklist), Item 6 , we only use one example per op, irrespective of the type ([AddOp](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#add) as an example). In future we have plans to explore further into this and, if needed, improve the situation. 

I believe the exception in this case makes sense because in the spec we do not have any example for per-axis quantization.

paulinesho (approved):

docs/spec.md
dominicsymes: Thanks for the update - the changes look good.

sdasgup3:
@dominicsymes 
Thanks for the review.

>> I think allowing also QuantizationIntegerScale should be discussed following that PR

That is right. Once we have https://github.com/openxla/stablehlo/pull/1475 landed, then the quantize/dequantize op will allow that type too.

docs/spec.md
```diff
@@ -5570,6 +5565,96 @@ Produces a `result` tuple from values `val`.
...
+
+* (C1) If `element_type(operand)` is a floating-point type,
+  * `element_type(operand) = expressed_type(result)`.
+* (C2) If `element_type(operand)` is a quantized type,
+  * `expressed_type(operand) = expressed_type(result)`.
+* (C3) `shape(operand) = shape(result)`.
+
+#### Examples
+
+```mlir
```
dansuh17, line 5648: For the sake of completeness, can we add an example for the case when the operand is a quantized type?

docs/spec.md
```diff
@@ -5587,6 +5582,92 @@ Produces a `result` tuple from values `val`.
...
+Formally,
+
+* For `element_type(operand)` a floating-point type,
+  * `rounded_result = round_nearest_even(operand / scale(result))`.
+  * `result = clamp(storage_min(result), rounded_result + zero_point(result), storage_max(result))`.
+
+* For `element_type(operand)` a quantized type,
+  * `float_result = (operand - zero_point(operand)) * scale(operand)`.
+  * `rounded_result = round_nearest_even(float_result / scale(result))`.
+  * `result = clamp(storage_min(result), rounded_result + zero_point(result), storage_max(result))`.
```
burmako, line 5637: There's duplication going on between the implementations of AddOp, UniformDequantizeOp and UniformQuantizeOp, but let's address it as part of #1566. It's fine to keep this PR as is.

burmako (changes requested):

docs/spec.md
```diff
@@ -5587,6 +5582,92 @@ Produces a `result` tuple from values `val`.
...
+#### Outputs
+
+| Name     | Type             | Constraints      |
+|----------|------------------|------------------|
+| `result` | quantized tensor | (C1), (C2), (C3) |
+
+#### Constraints
+
+* (C1) If `element_type(operand)` is a floating-point type,
+  * `element_type(operand) = expressed_type(result)`.
```
burmako, line 5654: I think that (C1) should go on this line, same for (C2) below, to be consistent with "If the operation uses non-quantized types" etc.

docs/spec.md
```diff
@@ -5587,6 +5582,92 @@ Produces a `result` tuple from values `val`.
...
+* (C2) If `element_type(operand)` is a quantized type,
+  * `expressed_type(operand) = expressed_type(result)`.
```
burmako, line 5656: How does per-axis quantization work for this operation? E.g. do quantization dimensions have to match if they are non-empty?

burmako (changes requested):

docs/spec.md
```diff
@@ -5587,6 +5582,92 @@ Produces a `result` tuple from values `val`.
...
+#### Examples
+
+```mlir
+// %operand: [4.0, 15.0]
+%result = "stablehlo.uniform_quantize"(%operand) : (tensor<2xf32>) -> tensor<2x!quant.uniform<i8<-128:127>:f32:0, {0.1:-30, 0.5:-20}>>
+// %result: [10, 10]
+
+// %operand: [10, 10]
+%result = "stablehlo.uniform_quantize"(%operand) : (tensor<2x!quant.uniform<i8<-128:127>:f32:0, {0.1:-30, 0.5:-20}>>) -> tensor<2x!quant.uniform<i8<-128:127>:f32:0, {0.1:-30, 0.5:-20}>>
+// %result: [10, 10]
```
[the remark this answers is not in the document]
dansuh17: +1 , it would be nice to have an example where the scale and/or zero point is different.

docs/spec.md
sdasgup3: The op, when `is_quantized(operand)` is true, is doing the following two steps:

1. `temp = stablehlo.uniform_dequantize(operand)`, using the quantization parameters of `operand`.
2. `reesult = stablehlo.uniform_quantize(temp)`, using  using the quantization parameter of `result`.

The quantization parameters of steps (1) and (2)  are independent and hence  quantization_dimension of `operand` and `result`  does not have to match. In fact one can be empty and the other non-empty.  Please let me know if I am missing something.

docs/spec.md
sdasgup3: Tried to get rid of some of the duplication for unifor.quantize case with `is_quantized(operand)`. Please take a look.

docs/spec.md
sdasgup3: I might not have followed the exact suggestion here. But I would like to know about it. 

In the meanwhile, proposed something along the line merging two cases: operand with fp type and operand with quantized type. 

PTAL.

docs/spec.md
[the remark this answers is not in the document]
sdasgup3: proposed a different example. PTAL.

docs/spec.md
```diff
@@ -5587,6 +5582,92 @@ Produces a `result` tuple from values `val`.
...
+tensor `operand` to a uniform quantized tensor `result` according to the
+quantization parameters defined by the `result` type.
+
+Formally,
+
+* For `element_type(operand)` a floating-point type,
+  * `rounded_result = round_nearest_even(operand / scale(result))`.
+  * `result = clamp(storage_min(result), rounded_result + zero_point(result), storage_max(result))`.
+
+* For `element_type(operand)` a quantized type,
```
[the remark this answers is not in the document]

docs/spec.md
```diff
@@ -5587,6 +5582,92 @@ Produces a `result` tuple from values `val`.
...
+
+#### Semantics
+
+Performs element-wise conversion of floating-point tensor or uniform quantized
+tensor `operand` to a uniform quantized tensor `result` according to the
+quantization parameters defined by the `result` type.
+
+Formally,
+
+* For `element_type(operand)` a floating-point type,
```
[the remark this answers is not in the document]

docs/spec.md
```diff
@@ -5587,6 +5582,92 @@ Produces a `result` tuple from values `val`.
 // %result: ([1.0, 2.0], (3))
 ```
 
+### uniform_dequantize
+
+#### Semantics
+
+Performs element-wise conversion of uniform quantized tensor `operand` to a
+floating point tensor `result` according to the quantization parameters defined
```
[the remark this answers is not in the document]

rewu93 (approved):

docs/spec.md
```diff
@@ -319,11 +319,6 @@ in StableHLO programs. In the meanwhile, here is the list of these operations:
   `dynamic_gather`, `dynamic_iota`, `dynamic_pad`, `dynamic_reshape`,
   `real_dynamic_slice`, `set_dimension_size`
   ([#8](https://github.com/openxla/stablehlo/issues/8)).
-* "Quantization" category of StableHLO operations - they were bootstrapped from
-  MHLO, but we haven't specced them yet: `uniform_quantize`
-  ([#531](https://github.com/openxla/stablehlo/issues/531)) and
-  `uniform_dequantize`
-  ([#530](https://github.com/openxla/stablehlo/issues/530)).
```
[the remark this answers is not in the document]
sdasgup3: **update**

Agreed!

I note that, for `stablehlo.quantize`, the verification check for `C2` is missing and hence verification is `no`. For `stablehlo.dequantize`, both `C1` and `C2` are verified (via ODS and https://github.com/openxla/stablehlo/blob/a5bce5ea7ac348a26c685e614dd70160df617776/stablehlo/dialect/TypeInference.cpp#L2984 ), hence the verification and type-inference columns have a `yes`.

sdasgup3:
My proposal is to merge the current PR and consider merging the two ops to convert op at a later point  via https://github.com/openxla/stablehlo/issues/1576.

dansuh17 (approved):

docs/spec.md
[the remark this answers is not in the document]
burmako: I think uniform_quantize falls under "revisit", per guidance from status.md: "no: there is no implementation, but working on that is part of the roadmap. Note that Verifier can never be labeled as "no" because the ODS already implements some verification".

stablehlo/dialect/StablehloOps.td
```diff
@@ -3073,16 +3073,16 @@ def StableHLO_UniformDequantizeOp : StableHLO_UnaryElementwiseOp<"uniform_dequan
       [InferTensorType, Pure], HLO_QuantizedIntTensor, TensorOf<[F32, BF16]>> {
```
burmako, line 3073: `TensorOf<[F32, BF16]>` doesn't look right. Same for `UniformQuantizeOp` above.

docs/spec.md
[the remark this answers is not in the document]
burmako: Also, opened #1603 to track C2 for UniformQuantizeOp.

docs/spec.md
```diff
@@ -5516,6 +5511,87 @@ Produces a `result` tuple from values `val`.
 // %result: ([1.0, 2.0], (3))
 ```
 
+### uniform_dequantize
+
+#### Semantics
+
+Performs element-wise conversion of uniform quantized tensor `operand` to a
```
burmako, line 5518: I don't think we have defined what is a uniform quantized tensor. Replace with "quantized tensor" here, below and in Ops.td?

docs/spec.md
```diff
@@ -5516,6 +5511,87 @@ Produces a `result` tuple from values `val`.
...
+  * `rounded_result = round_nearest_even(operand / scale(result))`.
+  * `result = clamp(storage_min(result), rounded_result + zero_point(result), storage_max(result))`.
+* If `is_quantized(operand)`: performs `dequantize_op_quantize(lambda operand:
+        operand, operand, type(result))`
+
+#### Inputs
+
+| Label | Name      | Type                                        | Constraints |
+|-------|-----------|---------------------------------------------|-------------|
+| (I1)  | `operand` | tensor of floating-point or quantized  type | (C1), (C2)  |
```
burmako, line 5569: Typo: extraneous whitespace.

docs/spec.md
```diff
@@ -5516,6 +5511,87 @@ Produces a `result` tuple from values `val`.
...
+tensor `operand` to a uniform quantized tensor `result` according to the
+quantization parameters defined by the `result` type.
+
+Formally,
+
+* If `is_float(operand)`:
+  * `rounded_result = round_nearest_even(operand / scale(result))`.
+  * `result = clamp(storage_min(result), rounded_result + zero_point(result), storage_max(result))`.
+* If `is_quantized(operand)`: performs `dequantize_op_quantize(lambda operand:
+        operand, operand, type(result))`
```
burmako, line 5563: Looks like it's time to introduce `dequantize(x)` and `quantize(operand, type)` functions? Then we can use them directly in these two ops, and we'll also be able to express `dequantize_op_quantize` in terms of them as well?

burmako (changes requested):

docs/spec.md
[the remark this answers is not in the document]
sdasgup3: Thanks!

docs/spec.md

docs/spec.md

docs/spec.md
sdasgup3: done. Please have a look.

stablehlo/dialect/StablehloOps.td
sdasgup3: That is an oversight. Used `HLO_Float`  instead to comply with the current spec. PTAL.

docs/spec.md
```diff
@@ -6126,13 +6202,21 @@ works for per-tensor quantization. Per-axis quantization is work in progress
 ([#1574](https://github.com/openxla/stablehlo/issues/1574)).
 
 ```python
+
+def dequantize(x):
+  return (x - zero_point(x)) * scale(x)
+
+def quantize(x, type):
+  rounded_result = round_nearest_even(x / scale(type))
+  return clamp(storage_min(type), rounded_result + zero_point(type), storage_max(type))
```
burmako, line 6211: I think this PR is a good opportunity to sharpen the definitions of dequantize and quantize, because current definitions derived from `dequantize_op_quantize`:
  1) Make use of implicit type conversions, e.g. in `rounded_result + zero_point(type)`, which obscure what's going on and that came up in recent offline discussions.
  2) Call elementwise ops like SubtractOp on quantized tensors, which leads to a virtual stack overflow (i.e. circular definition in notation) because elementwise ops on quantized tensors are defined (in #1566) in terms of dequantize and quantize :)

Here's how the sharpened definitions could look like (I think they should live in separate bullet points in this section, so that they can have their own prose in addition to the Python implementation):

* `dequantize` is defined on quantized tensor types and turns them into 
floating-point tensor types. This happens via converting quantized elements 
which represent integer values of the storage type into corresponding 
floating-point values of the expressed type using the zero point and scale 
associated with the quantized element type. At the moment, this function only 
works for per-tensor quantization. Per-axis quantization is work in progress
([#1574](https://github.com/openxla/stablehlo/issues/1574)).

```python
def dequantize(x: Value) -> Value:
  assert is_quantized(x)
  x_storage = bitcast_convert(x, storage_type(x))
  x_storage_sub = x_storage - zero_point(x_storage)
  x_expressed_sub = convert(x_storage_sub, expressed_type(x))
  return x_expressed_sub * scale(x)
```

* `quantize` is defined on floating-point tensor types and turns them into
quantized tensor types. This happens via converting floating-point values
of the expressed type into corresponding integer values of the storage type
using the zero point and scale associated with the quantized element type. 
At the moment, this function only works for per-tensor quantization. Per-axis 
quantization is work in progress
([#1574](https://github.com/openxla/stablehlo/issues/1574)).

```python
def quantize(x: Value, type: Type) -> Value:
  assert is_float(x) and is_quantized(type)
  x_expressed_rounded = round_nearest_even(x / scale(type))
  x_storage_rounded = convert(x_expressed_rounded, storage_type(type))
  x_storage_add = x_storage_rounded + zero_point(type)
  x_storage = clamp(storage_min(type), x_storage_add, storage_max(type))
  return bitcast_convert(x_storage, type)
```

burmako, line 6211: The sharpened definitions of dequantize and quantize from the comment above enable us to reason about a specific discussion item that came up offline:
  1) Are we comfortable computing `x_storage - zero_point(x_storage)` and `x_storage_rounded + zero_point(type)` in `storage_type(x)` or do we need to upcast to e.g. `i32` first and then downcast back to the storage type?
  2) Also, now that I'm thinking about that, does `clamp(storage_min(type), x_storage_add, storage_max(type))` even make sense as it's currently written? If `type(x_storage_add) = storage_type(x)`, then clamping to minimum and maximum values of the storage type is a no-op.

burmako (changes requested):

docs/spec.md
burmako, line 6211: Similar questions apply to the definition of quantized DotGeneralOp. They don't have to be addressed in this PR, but let's at least open a ticket to nail this down.

burmako: You know, perhaps we could use the same `dequantize_op_quantize` approach for DotGeneralOp? This would avoid the need to think specifically about DotGeneralOp - we'll be able to leverage whatever solution we end up with for `dequantize` and `quantize`.

burmako: Also, if we're able to use `dequantize_op_quantize` for these ops, then we wouldn't need to worry about per-axis quantization for the underlying data movement ops, right?

docs/spec.md
```diff
@@ -6126,13 +6202,21 @@ works for per-tensor quantization. Per-axis quantization is work in progress
...
+
 def dequantize_op_quantize(op, *inputs_and_output_type):
   inputs = inputs_and_output_type[:-1]
   output_type = inputs_and_output_type[-1]
-  float_inputs = [(x - zero_point(x)) * scale(x) for x in inputs]
+
+  float_inputs = [dequantize(x) for x in inputs]
```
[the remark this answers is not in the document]

stablehlo/dialect/StablehloOps.td
```diff
@@ -3055,38 +3055,38 @@ def StableHLO_RngBitGeneratorOp : StableHLO_Op<"rng_bit_generator", [Pure]> {
 
 // TODO(b/230662142): Implement unknown scales/zero_point cases.
 def StableHLO_UniformQuantizeOp : StableHLO_UnaryElementwiseOp<"uniform_quantize",
```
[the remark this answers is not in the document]
sdasgup3: https://github.com/openxla/stablehlo/pull/1626
done.

docs/spec.md
sdasgup3: Thanks!
included.

sdasgup3: Thanks for pointing it point. Let me propose to keep the PR as is (using the conversion to  `storage_type`) and revisit the issues separately in https://github.com/openxla/stablehlo/issues/1627.

sdasgup3: Opened https://github.com/openxla/stablehlo/issues/1628 for exploring the topic. Resolving it for now.

stablehlo/dialect/StablehloOps.td
```diff
@@ -3055,38 +3055,39 @@ def StableHLO_RngBitGeneratorOp : StableHLO_Op<"rng_bit_generator", [Pure]> {
...
-    %result = stablehlo.uniform_quantize %operand : (tensor<16x16xf32>) -> tensor<16x16x!quant.uniform<ui8:f32, 34.0:16>>
+    %result = stablehlo.uniform_quantize %operand : (tensor<2xf32>) -> tensor<2x!quant.uniform<i8:f32:0, {0.1:-30,0.5:-20}>>
     ```
   }];
 }
 
 def StableHLO_UniformDequantizeOp : StableHLO_UnaryElementwiseOp<"uniform_dequantize",
-      [InferTensorType, Pure], HLO_QuantizedIntTensor, TensorOf<[F32, BF16]>> {
+      [InferTensorType, Pure], HLO_QuantizedIntTensor /*uniform_dequantize_i1*/,
+      HLO_FpTensor> { /*uniform_dequantize_c1*/
```
burmako, line 3078: uniform_dequantize_c2 as well.

docs/spec.md
```diff
@@ -5535,6 +5530,87 @@ Produces a `result` tuple from values `val`.
...
 
+### uniform_dequantize
+
+#### Semantics
+
+Performs element-wise conversion of quantized tensor `operand` to a
+floating-point tensor `result` according to the quantization parameters defined
+by the `operand` type.
+
+Formally, `result = dequantize(operand)`.
```
burmako, line 5541: Nit: the canonical way of expressing this in the document is "More formally, ...".

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -5106,26 +5106,25 @@ func.func @uniform_requantize(%arg: tensor<16x16x!quant.uniform<i8:f32, 5.0:20>>
...
 
-// CHECK: func @uniform_dequantize_unranked
-func.func @uniform_dequantize_unranked(%arg: tensor<*x!quant.uniform<i8:f32, 34.0:16>>) -> tensor<*xf32> {
-  %0 = stablehlo.uniform_dequantize %arg : (tensor<*x!quant.uniform<i8:f32, 34.0:16>>) -> tensor<*xf32>
-  func.return %0 : tensor<*xf32>
+func.func @uniform_dequantize_i1(%arg: tensor<16x16xf32>) -> tensor<16x16xf32> {
+  // expected-error@+1 {{operand #0 must be tensor of 4/8/16/32-bit uniform quantized signed integer or 4/8/16/32-bit uniform quantized unsigned integer values, but got 'tensor<16x16xf32>'}}
+  %0 = stablehlo.uniform_dequantize %arg : (tensor<16x16xf32>) -> tensor<16x16xf32>
+  func.return %0 : tensor<16x16xf32>
 }
```
burmako, line 5121: Delete this test, given that this is covered by ODS?

docs/spec.md
burmako: Sounds good.

burmako (approved):

stablehlo/dialect/StablehloOps.td
sdasgup3: done.

docs/spec.md

stablehlo/tests/ops_stablehlo.mlir
sdasgup3: done

Issue #1318, opened 2023-03-11
Establish forward compatibility testing


burmako:
Given that https://reviews.llvm.org/D143647 is about to land, I wonder how we can set up tests for forward compatibility, i.e. that portable artifacts serialized by HEAD can be deserialized by 0.9.

PR #1498, opened 2023-05-18
[RFC] Forward Compatibility Testing

GleasonK:
This is a proposal for forward compatibility testing in StableHLO. This ensures that payloads serialized at HEAD can be deserialized at a previous release.

Note, this plan is built on top of our existing backward compatibility testing infrastructure, and as such does not include semantic forward compatibility testing. Once we have addressed semantic compatibility testing in #1416 we can expand this machinery to include semantic forward compatibility testing.

- **Proposed design prototype:** https://github.com/GleasonK/stablehlo/pull/35
- **Alternate design prototype:** https://github.com/GleasonK/stablehlo/pull/33

Notification of this PR will be sent to the [openxla-discuss](https://groups.google.com/a/openxla.org/g/openxla-discuss) group.

Related to #1318.

UPDATE 5/25: While this RFC is in review, some form of forward incompatibility detection is still required by StableHLO. In https://github.com/openxla/stablehlo/pull/1525 a stop-gap solution based on the preferred design was implemented. It will be modified in accordance with feedback on this RFC.

### DIFF
--- a/rfcs/20230517-forward-compatibility-testing.md
+++ b/rfcs/20230517-forward-compatibility-testing.md
@@ -8,7 +8,7 @@ check backward compatibility by versioning and serializing
 compatibility. This document makes the following proposal:
 
 **Proposal:** Add tests to compare bytes of bytecode file serialized at HEAD
-with existing [backward compatibility bytecode test files](https://github.com/search?q=repo%3Aopenxla%2Fstablehlo+path%3A**%2Fstablehlo_legalize_to_vhlo.0_*&type=code).
+with existing [backward compatibility bytecode test files](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/stablehlo_legalize_to_vhlo.0_9_0.mlir.bc).
 
 ## Requirements
 
@@ -56,12 +56,12 @@ Other benefits of this approach include that it is lightweight (_R3_), can be ru
 locally to detect compatibility issues during development, and it fits nicely into
 existing CI build-and-test jobs. If the underlying byte-identical assumption is
 invalidated, we can fall back to the alternate design in a hybrid testing approach,
-since this approach will safely identify all forward incompatibilities, with a
+since byte-comparison will safely identify all forward incompatibilities, with a
 slight risk of false positives that should be tested using the alternate design.
 
 ## Alternate Design: Serialize at HEAD, Deserialize at Target Release
 
-This design is explored more in a [previous commit of this RFC](https://github.com/openxla/stablehlo/pull/1498/commits/0792eb75e85c54f9d106878569b088d03c568b70),
+This design is explored more in a [previous commit of this RFC](https://github.com/openxla/stablehlo/blob/0792eb75e85c54f9d106878569b088d03c568b70/rfcs/20230517-forward-compatibility-testing.md#preferred-design-serialize-at-head-deserialize-at-target-release),
 and is summarized in this version for brevity. If byte-equality cannot be safely
 relied on, then we may need to use this approach, or a hybrid of the two. This
 design can be summarized as:
--- a/rfcs/20230517-forward-compatibility-testing.md
+++ b/rfcs/20230517-forward-compatibility-testing.md
@@ -8,7 +8,7 @@ check backward compatibility by versioning and serializing
 compatibility. This document makes the following proposal:
 
 **Proposal:** Add tests to compare `stablehlo_legalize_to_vhlo.0_X_0.mlir`
-serialized at HEAD with existing [`stablehlo_legalize_to_vhlo.0_X_0.mlir.bc`](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/stablehlo_legalize_to_vhlo.0_9_0.mlir.bc)
+serialized at HEAD using `--target 0.X.0` with existing [`stablehlo_legalize_to_vhlo.0_X_0.mlir.bc`](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/stablehlo_legalize_to_vhlo.0_9_0.mlir.bc)
 test files.
 
 ## Requirements
@@ -24,10 +24,10 @@ of time.
 ## Proposed Design: Statically Test Forward Compatibility
 
 In this design, forward incompatibilities are detected by serializing
-`stablehlo_legalize_to_vhlo.mlir` at head and comparing it to a known-good file
-serialized serialized at previous releases. This approach operates on the
-assumption that at HEAD we are able to produce a byte-identical serialized
-artifact, excluding custom header info and debug locations.
+`stablehlo_legalize_to_vhlo.0_X_0.mlir` at HEAD and comparing it to a
+known-good file serialized at previous releases. This approach
+operates on the assumption that at HEAD we are able to produce a byte-identical
+serialized artifact, excluding custom header info and debug locations.
 
 Detecting forward incompatiblities can be accomplished by adding the following
 `RUN` line to the versioned [`stablehlo_legalize_to_vhlo.0_X_0.mlir`](https://github.com/search?q=repo%3Aopenxla%2Fstablehlo+path%3A**%2Fstablehlo_legalize_to_vhlo.0_*&type=code)
@@ -42,8 +42,11 @@ in StableHLO (_R1_) and LLVM revision bumps (_R2_).
 
 This will not work as is because `bytecode::producer`, `op::location` and
 `block_argument::location` fields may change between when the original bytecode
-file has been produced and when the test runs at HEAD. To account for that,
-we can strip this information before comparing the payloads.
+file has been produced and when the test runs at HEAD. We can set a consistent
+producer string to take care of the first item. For debug locations, one option
+would be to explicitly specify debug locations, however this is very cumbersome
+in a 2.2k line test file. Another option is to strip this debug information
+when generating the bytecode files.
 
 The StableHLO process for [Contributing Incompatible Changes](https://github.com/openxla/stablehlo/blob/main/docs/vhlo.md#add-versioned-serialization-test)
 requires `stablehlo_legalize_to_vhlo.mlir` to be serialized for each
@@ -51,11 +54,12 @@ compatibility breaking change, which means that we can test forward
 compatibility between HEAD and each minor release of StableHLO. Given that
 all changes to the StableHLO opset, as well as changes to serialization
 machinery like targeting a newer version of the MLIR bytecode format, require
-bumping the minor version, there is a low chance of causing forward
-incompatibilities between different patch versions of StableHLO (i.e.
-`0.10.1 → 0.10.2`). Ensuring compatibility between HEAD and all minor releases
-(`0.9.0`, `0.10.0`, etc.) should provide thorough enough coverage of the two
-mentioned sources of forward incompatibilities.
+bumping the minor version, patch versions _must_ produce identical bytecode to
+their minor versions. As such we should prevent the ability to target bytecode
+generation for patch versions other than `0`. With that restriction, ensuring
+compatibility between HEAD and all minor releases (`0.9.0`, `0.10.0`, etc.)
+provides full coverage of the two mentioned sources of forward
+incompatibilities.
 
 Other benefits of this approach include that it is lightweight (_R3_), can be run
 locally to detect compatibility issues during development, and it fits nicely into
--- a/rfcs/20230517-forward-compatibility-testing.md
+++ b/rfcs/20230517-forward-compatibility-testing.md
@@ -7,8 +7,9 @@ check backward compatibility by versioning and serializing
 `stablehlo_legalize_to_vhlo.mlir`, and need something similar for forward
 compatibility. This document makes the following proposal:
 
-**Proposal:** Add tests to compare bytes of bytecode file serialized at HEAD
-with existing [backward compatibility bytecode test files](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/stablehlo_legalize_to_vhlo.0_9_0.mlir.bc).
+**Proposal:** Add tests to compare `stablehlo_legalize_to_vhlo.0_X_0.mlir`
+serialized at HEAD with existing [`stablehlo_legalize_to_vhlo.0_X_0.mlir.bc`](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/stablehlo_legalize_to_vhlo.0_9_0.mlir.bc)
+test files.
 
 ## Requirements
 
@@ -22,42 +23,54 @@ of time.
 
 ## Proposed Design: Statically Test Forward Compatibility
 
-In this design, forward incompatibilities are detected by serializing a file at
-head and comparing it to a known-good file serialized at the previous release.
-This approach operates on the assumption that at HEAD we are able to produce
-a byte-identical serialized artifact, excluding custom header info and debug
-locations.
+In this design, forward incompatibilities are detected by serializing
+`stablehlo_legalize_to_vhlo.mlir` at head and comparing it to a known-good file
+serialized serialized at previous releases. This approach operates on the
+assumption that at HEAD we are able to produce a byte-identical serialized
+artifact, excluding custom header info and debug locations.
 
 Detecting forward incompatiblities can be accomplished by adding the following
 `RUN` line to the versioned [`stablehlo_legalize_to_vhlo.0_X_0.mlir`](https://github.com/search?q=repo%3Aopenxla%2Fstablehlo+path%3A**%2Fstablehlo_legalize_to_vhlo.0_*&type=code)
 test files.
 
 ```bash
-diff %s.bc <(stablehlo-opt %s | stablehlo-translate --serialize --target=0.X.0)
+diff %s.bc <(stablehlo-translate --serialize --target=0.X.0)
 ```
 
-Passing through `stablehlo-opt` in this example removes debug info. An option
-to `stablehlo-translate` can be added to strip debug info and header info. This
-test should be run in CI on all PRs to prevent merging incompatible changes in
-StableHLO (_R1_) and LLVM revision bumps (_R2_).
+This test should be run in CI on all PRs to prevent merging incompatible changes
+in StableHLO (_R1_) and LLVM revision bumps (_R2_).
+
+This will not work as is because `bytecode::producer`, `op::location` and
+`block_argument::location` fields may change between when the original bytecode
+file has been produced and when the test runs at HEAD. To account for that,
+we can strip this information before comparing the payloads.
 
 The StableHLO process for [Contributing Incompatible Changes](https://github.com/openxla/stablehlo/blob/main/docs/vhlo.md#add-versioned-serialization-test)
-requires a versioned test be serialized for each compatibility breaking change,
-which means that we can test forward compatibility between HEAD and each minor
-release of StableHLO. Given that all changes to the StableHLO opset, as well as
-changes to serialization machinery like targeting a newer version of the MLIR
-bytecode format, require bumping the minor version, there is a low chance of
-causing forward incompatibilities between different patch versions of StableHLO
-(i.e. `0.10.1 → 0.10.2`). Ensuring compatibility between HEAD and all minor
-releases (`0.9.0`, `0.10.0`, etc.) should provide thorough enough coverage of the
-two mentioned sources of forward incompatibilities.
+requires `stablehlo_legalize_to_vhlo.mlir` to be serialized for each
+compatibility breaking change, which means that we can test forward
+compatibility between HEAD and each minor release of StableHLO. Given that
+all changes to the StableHLO opset, as well as changes to serialization
+machinery like targeting a newer version of the MLIR bytecode format, require
+bumping the minor version, there is a low chance of causing forward
+incompatibilities between different patch versions of StableHLO (i.e.
+`0.10.1 → 0.10.2`). Ensuring compatibility between HEAD and all minor releases
+(`0.9.0`, `0.10.0`, etc.) should provide thorough enough coverage of the two
+mentioned sources of forward incompatibilities.
 
 Other benefits of this approach include that it is lightweight (_R3_), can be run
 locally to detect compatibility issues during development, and it fits nicely into
-existing CI build-and-test jobs. If the underlying byte-identical assumption is
-invalidated, we can fall back to the alternate design in a hybrid testing approach,
-since byte-comparison will safely identify all forward incompatibilities, with a
-slight risk of false positives that should be tested using the alternate design.
+existing CI build-and-test jobs.
+
+### Assumption: Comparing Bytecode Payloads
+
+This design relies on the assumption that at HEAD we will be able to produce
+a byte-identical serialized artifact, excluding `bytecode::producer`,
+`op::location` and `block_argument::location` fields
+
+If this doesn't work, we can fall back to the alternate design in a hybrid
+testing approach, since byte-comparison will safely identify all forward
+incompatibilities, with some risk of false positives that should be tested
+using the alternate design.
 
 ## Alternate Design: Serialize at HEAD, Deserialize at Target Release
 
@@ -67,9 +80,10 @@ relied on, then we may need to use this approach, or a hybrid of the two. This
 design can be summarized as:
 
 * Maintain bytecode files serialized at HEAD, which are re-serialized before
-  each StableHLO release (~2x/wk): `stablehlo_legalize_to_vhlo.0_X_0.mlirbc.HEAD`
+  each StableHLO release (~2x/wk):
+  `stablehlo_legalize_to_vhlo.0_X_0.mlir.bc.HEAD`
 * Add CI jobs that build StableHLO/MLIR at every tagged minor release
-  (`v0.9.0`, `v0.10.0`, etc.) and check out the `0_X_0.mlirbc.HEAD` file that
+  (`v0.9.0`, `v0.10.0`, etc.) and check out the `0_X_0.mlir.bc.HEAD` file that
   corresponds to the tagged release to deserialize and run FileCheck tests on.
 
 This approach is more resilient to compatible changes in the bytecode format.
@@ -79,7 +93,7 @@ like projects that rely on StableHLO and want to run StableHLO tests. And
 lastly it is more costly and time consuming, as it requires building several
 releases of StableHLO and LLVM/MLIR.
 
-Given the pros/cons of these two designs, this CI should only be used as needed.
+Given the pros/cons of this design, this CI should only be used as needed.
 If there are changes that cause bytecode differences that cannot be tested
 statically, this infrastructure can be used. Otherwise, when possible, the
 static verification should be used.
--- a/rfcs/20230517-forward-compatibility-testing.md
+++ b/rfcs/20230517-forward-compatibility-testing.md
@@ -2,16 +2,13 @@
 
 ## Background
 
-StableHLO needs to verify both forward and backward compatibility via unit
-tests. Currently we check backward compatibility by versioning
-`stablehlo_legalize_to_vhlo.mlir`. This document makes the following proposals:
+StableHLO needs to verify both forward and backward compatibility. Currently we
+check backward compatibility by versioning and serializing
+`stablehlo_legalize_to_vhlo.mlir`, and need something similar for forward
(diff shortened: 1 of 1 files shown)


rfcs/20230517-forward-compatibility-testing.md
```diff
@@ -0,0 +1,102 @@
...
+consuming than needed, and doesn't scale well with multiple releases a week
+(_R3_). Given that all changes to the StableHLO opset (_R1_), as well as
+changes to serialization machinery like targeting a newer version of the MLIR
+bytecode format (_R2_), require bumping the minor version, there is a low
+chance of causing forward incompatibilities between different patch versions of
+StableHLO (i.e. `0.10.1 → 0.10.2`). Ensuring compatibility between HEAD and all
+minor releases (`0.9.0`, `0.10.0`, etc.) should provide thorough enough
+coverage of the two mentioned sources of forward incompatibilities.
+
+Currently, this means that we only serialize a file at HEAD for all supported
```
burmako, line 89: What does "supported" mean? E.g. v0.9.0 is already 2.5 months old, i.e. is well outside the current forward compatibility window. Is it still supported?

burmako (changes requested):

rfcs/20230517-forward-compatibility-testing.md
```diff
@@ -0,0 +1,102 @@
...
+is both useful and practical:
+
+* **R1.** Tests must detect changes in StableHLO that break forward compatibility.
+* **R2.** Tests must detect LLVM revision bumps that break forward compatibility.
+* **R3.** Test job must scale with releases and execute in a reasonable amount
+of time.
+
+## Preferred Design: Serialize at HEAD, Deserialize at Target Release
+
+_Note: Section only discusses how to test a generic release `0.X.0`. Discussion
```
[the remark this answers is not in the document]
GleasonK: Removed section in rework.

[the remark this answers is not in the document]
GleasonK: Removed section in rework.

rfcs/20230517-forward-compatibility-testing.md
```diff
@@ -0,0 +1,102 @@
...
+  stablehlo-translate --serialize --target=0.X.0 |
+    stablehlo-opt --mlir-print-op-generic | FileCheck %s
+```
+
+**Proposal 1:** Maintain bytecode at HEAD to use for forward compatibility
+testing.
+
+### Add Forward Compatibility CI
+
+A CI job which blocks pull request merges to main can be added to ensure
```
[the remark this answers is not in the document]
GleasonK: On every PR. With build caching it isn't too too expensive, but we want to avoid breaking changes being merged from both StableHLO PRs (R1), and LLVM revision bumps (R2), meaning we would want this to block most non-documentation PRs.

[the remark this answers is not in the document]
GleasonK: Clarified in updated version

rfcs/20230517-forward-compatibility-testing.md
```diff
@@ -0,0 +1,102 @@
...
+forward compatibility of PRs. To test version `0.X.0` the CI job does the
+following:
+
+* Checkout `openxla/stablehlo` at tag `v0.X.0`.
+* Checkout file at HEAD `stablehlo_legalize_to_vhlo.0_X_0.mlir.bc.HEAD`
```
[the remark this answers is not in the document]
GleasonK: We could. I think this would require a single CI job which builds at HEAD, serializes, then checks out older revisions of StableHLO/LLVM and rebuilds sequentially. This can't be parallelized as easily as the current approach (see prototype).

Alternatively this would require each CI job to rebuild at HEAD then check out an older tag and rebuild StableHLO/LLVM at those revisions. Slightly higher parallelism, but more things being built. 

Either way, this is being reworked into the alternative design, so lets explore the proposed design before optimizing this approach.

rfcs/20230517-forward-compatibility-testing.md
```diff
@@ -0,0 +1,102 @@
...
+**Proposal 2:** Add CI which builds at tagged release commits to check forward
+compatibility.
+
+## What releases should be tested?
+
+Testing every single release would be sound, but more expensive and time
+consuming than needed, and doesn't scale well with multiple releases a week
+(_R3_). Given that all changes to the StableHLO opset (_R1_), as well as
+changes to serialization machinery like targeting a newer version of the MLIR
+bytecode format (_R2_), require bumping the minor version, there is a low
```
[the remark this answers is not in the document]
GleasonK: I'll propose an amendment to the existing process when I get around to integrating Jacques' recent changes to bytecode format target versions, being able to specify a version is a relatively new feature. 

Will do that in the next few days. But the plan is that we should always correlate a StableHLO release to the most recent version of upstream MLIR bytecode format. We can add a lint check for this as well.

rfcs/20230517-forward-compatibility-testing.md
GleasonK: I suppose supported is up to us to define, since our compatibility guarantees are a minimum.

Given that one use of forward compatibility is to provide users adequate time to upgrade, `v0.9.0` is only allowed to be removed as a supported forward compatible version once `v0.10.0` is a month old. 

Until we explicitly prevent it and clean up the code, the infrastructure allows targeting `v0.9.0`. We should consider a process for this cleanup, i.e. how aggressively do we want to cleanup IR downgrades?

rfcs/20230517-forward-compatibility-testing.md
```diff
@@ -0,0 +1,85 @@
...
+test files.
+
+```bash
+diff %s.bc <(stablehlo-opt %s | stablehlo-translate --serialize --target=0.X.0)
+```
+
+Passing through `stablehlo-opt` in this example removes debug info. An option
+to `stablehlo-translate` can be added to strip debug info and header info. This
+test should be run in CI on all PRs to prevent merging incompatible changes in
+StableHLO (_R1_) and LLVM revision bumps (_R2_).
```
burmako, line 42: How about rewriting this as follows, to first explain how this approach would work and only then state the assumptions?

```
Detecting forward incompatiblities can be accomplished by adding the following
`RUN` line to the versioned [`stablehlo_legalize_to_vhlo.0_X_0.mlir`](https://github.com/search?q=repo%3Aopenxla%2Fstablehlo+path%3A**%2Fstablehlo_legalize_to_vhlo.0_*&type=code)
test files.

diff %s.bc <(stablehlo-translate --serialize --target=0.X.0)

This test should be run in CI on all PRs to prevent merging incompatible changes 
in StableHLO (_R1_) and LLVM revision bumps (_R2_).

This will not work as is because `bytecode::producer`, `op::location` and 
`block_argument::location` fields may change between when the original bytecode
file has been produced and when the test runs at HEAD. To account for that,
we can strip this information before comparing the payloads.
```

rfcs/20230517-forward-compatibility-testing.md
```diff
@@ -0,0 +1,85 @@
...
+Detecting forward incompatiblities can be accomplished by adding the following
+`RUN` line to the versioned [`stablehlo_legalize_to_vhlo.0_X_0.mlir`](https://github.com/search?q=repo%3Aopenxla%2Fstablehlo+path%3A**%2Fstablehlo_legalize_to_vhlo.0_*&type=code)
+test files.
+
+```bash
+diff %s.bc <(stablehlo-opt %s | stablehlo-translate --serialize --target=0.X.0)
+```
+
+Passing through `stablehlo-opt` in this example removes debug info. An option
+to `stablehlo-translate` can be added to strip debug info and header info. This
```
burmako, line 40: We could also write a tool that strips this information from bytecode payloads. I think this might be preferable to complicating the contract of `stablehlo-translate` with adding this functionality specific to testing there.

rfcs/20230517-forward-compatibility-testing.md
```diff
@@ -0,0 +1,85 @@
...
+test should be run in CI on all PRs to prevent merging incompatible changes in
+StableHLO (_R1_) and LLVM revision bumps (_R2_).
+
+The StableHLO process for [Contributing Incompatible Changes](https://github.com/openxla/stablehlo/blob/main/docs/vhlo.md#add-versioned-serialization-test)
+requires a versioned test be serialized for each compatibility breaking change,
```
burmako, line 45: "requires stablehlo_legalize_to_vhlo.mlir to be serialized"?

rfcs/20230517-forward-compatibility-testing.md
```diff
@@ -0,0 +1,85 @@
...
+which means that we can test forward compatibility between HEAD and each minor
+release of StableHLO. Given that all changes to the StableHLO opset, as well as
+changes to serialization machinery like targeting a newer version of the MLIR
+bytecode format, require bumping the minor version, there is a low chance of
```
burmako, line 49: Does this process leave any chance for forward incompatibilities?

rfcs/20230517-forward-compatibility-testing.md
```diff
@@ -0,0 +1,85 @@
...
+(i.e. `0.10.1 → 0.10.2`). Ensuring compatibility between HEAD and all minor
+releases (`0.9.0`, `0.10.0`, etc.) should provide thorough enough coverage of the
+two mentioned sources of forward incompatibilities.
+
+Other benefits of this approach include that it is lightweight (_R3_), can be run
+locally to detect compatibility issues during development, and it fits nicely into
+existing CI build-and-test jobs. If the underlying byte-identical assumption is
+invalidated, we can fall back to the alternate design in a hybrid testing approach,
+since byte-comparison will safely identify all forward incompatibilities, with a
+slight risk of false positives that should be tested using the alternate design.
```
burmako, line 60: Do we have evidence that it's going to be a slight risk? Perhaps drop the "slight" part?

burmako, line 60: How about emphasizing the assumption by moving it to a separate section, for example:

```
### Assumption: Comparing Bytecode Payloads

This design relies on the assumption that at HEAD we will be able to produce 
a byte-identical serialized artifact, excluding `bytecode::producer`, 
`op::location` and `block_argument::location` fields

If this doesn't work, we can fall back to the alternate design in a hybrid 
testing approach, since byte-comparison will safely identify all forward 
incompatibilities, with some risk of false positives that should be tested 
using the alternate design.
```

rfcs/20230517-forward-compatibility-testing.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Alternate Design: Serialize at HEAD, Deserialize at Target Release
+
+This design is explored more in a [previous commit of this RFC](https://github.com/openxla/stablehlo/blob/0792eb75e85c54f9d106878569b088d03c568b70/rfcs/20230517-forward-compatibility-testing.md#preferred-design-serialize-at-head-deserialize-at-target-release),
+and is summarized in this version for brevity. If byte-equality cannot be safely
+relied on, then we may need to use this approach, or a hybrid of the two. This
+design can be summarized as:
+
+* Maintain bytecode files serialized at HEAD, which are re-serialized before
+  each StableHLO release (~2x/wk): `stablehlo_legalize_to_vhlo.0_X_0.mlirbc.HEAD`
```
burmako, line 70: Fwiw the related files in the GitHub repository are currently called `stablehlo_legalize_to_vhlo.0_X_0.mlir.bc`. (With a period between "mlir" and "bc").

rfcs/20230517-forward-compatibility-testing.md
```diff
@@ -0,0 +1,85 @@
...
+  corresponds to the tagged release to deserialize and run FileCheck tests on.
+
+This approach is more resilient to compatible changes in the bytecode format.
+The downsides of this approach include that it cannot easily be tested locally,
+requiring CI jobs. As such it doesn't work out of the box in other environments,
+like projects that rely on StableHLO and want to run StableHLO tests. And
+lastly it is more costly and time consuming, as it requires building several
+releases of StableHLO and LLVM/MLIR.
+
+Given the pros/cons of these two designs, this CI should only be used as needed.
```
burmako, line 82: "this design"?

burmako (changes requested):

rfcs/20230517-forward-compatibility-testing.md
GleasonK: Removed all mention. The important piece is that debug info is stripped.

rfcs/20230517-forward-compatibility-testing.md
GleasonK: FileA: Serialized@0.x.0

Release 0.x.0 - Can serialize a file to the bytes of FileA, can read a file with the bytes of FileA
Release 0.x.1 - Can serialize a file to FileA, can read a file with the bytes of FileA
Release 0.x.2 - Can serialize a file to FileA, can read a file with the bytes of FileA
...

Assuming that a release never broke backward compatibility (can read FileA), and its successor release doesn't break forward compatibility (can generate FileA), then there should be no forward compatibility issues. Given that there are no opset changes or bytecode changes during patch releases, targeting minor versions is preferable and safe (and the ability to target patch versions should possibly be removed).

rfcs/20230517-forward-compatibility-testing.md
GleasonK: Good catch, at the same time I was working on a PR to rename to `.mlirbc` since the MLIR vscode extension recognizes this, but I don't think there is much benefit to being able to look at/edit the textual assembly of VHLO so dropped it

rfcs/20230517-forward-compatibility-testing.md
```diff
@@ -0,0 +1,99 @@
...
+
+* **R1.** Tests must detect changes in StableHLO that break forward compatibility.
+* **R2.** Tests must detect LLVM revision bumps that break forward compatibility.
+* **R3.** Tests must scale with releases and execute in a reasonable amount
+of time.
+
+## Proposed Design: Statically Test Forward Compatibility
+
+In this design, forward incompatibilities are detected by serializing
+`stablehlo_legalize_to_vhlo.mlir` at head and comparing it to a known-good file
```
burmako, line 27: Do you mean "serializing `stablehlo_legalize_to_vhlo.0_X_0.mlir` at HEAD"? (Apologies, I know I asked for the current wording in my previous comment, but I didn't fully understand the proposal at that point).

rfcs/20230517-forward-compatibility-testing.md
```diff
@@ -0,0 +1,99 @@
...
+serialized serialized at previous releases. This approach operates on the
```
burmako, line 28: Typo: "serialized serialized".

rfcs/20230517-forward-compatibility-testing.md
```diff
@@ -0,0 +1,99 @@
...
+assumption that at HEAD we are able to produce a byte-identical serialized
+artifact, excluding custom header info and debug locations.
```
burmako, line 30: "excluding producer information and locations"? 1) I don't think "custom header" is a thing in [the bytecode format spec](https://mlir.llvm.org/docs/BytecodeFormat/). 2) Also, locations are called just "locations", not "debug locations".

rfcs/20230517-forward-compatibility-testing.md
```diff
@@ -0,0 +1,99 @@
...
+requires `stablehlo_legalize_to_vhlo.mlir` to be serialized for each
+compatibility breaking change, which means that we can test forward
+compatibility between HEAD and each minor release of StableHLO. Given that
+all changes to the StableHLO opset, as well as changes to serialization
+machinery like targeting a newer version of the MLIR bytecode format, require
+bumping the minor version, there is a low chance of causing forward
+incompatibilities between different patch versions of StableHLO (i.e.
+`0.10.1 → 0.10.2`). Ensuring compatibility between HEAD and all minor releases
+(`0.9.0`, `0.10.0`, etc.) should provide thorough enough coverage of the two
+mentioned sources of forward incompatibilities.
```
burmako, line 58: How do we define "thorough enough"? The current version of [compatibility.md](https://github.com/openxla/stablehlo/blob/main/docs/compatibility.md) is pretty strict about these things.

rfcs/20230517-forward-compatibility-testing.md
burmako: I'm not sure I understand the difference between "Can serialize a file to the bytes of FileA" and "Can serialize a file to FileA". Also, can you update the text of the RFC to elaborate on what "a low chance" means exactly?

burmako (changes requested):

rfcs/20230517-forward-compatibility-testing.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Background
+
+StableHLO needs to verify both forward and backward compatibility. Currently we
+check backward compatibility by versioning and serializing
+`stablehlo_legalize_to_vhlo.mlir`, and need something similar for forward
+compatibility. This document makes the following proposal:
+
+**Proposal:** Add tests to compare `stablehlo_legalize_to_vhlo.0_X_0.mlir`
+serialized at HEAD with existing [`stablehlo_legalize_to_vhlo.0_X_0.mlir.bc`](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/stablehlo_legalize_to_vhlo.0_9_0.mlir.bc)
```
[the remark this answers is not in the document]
GleasonK: I had a comment on this elsewhere, in some sense we don't want these files to be human editable since VHLO is not intended to be editable. But I prototyped that change last week [here](https://github.com/openxla/stablehlo/compare/main...GleasonK:stablehlo:bytecode-rename-ext) but never sent PR after thinking about this more.

rfcs/20230517-forward-compatibility-testing.md
GleasonK: Will do. The current wording is based on `--mlir-pretty-debuginfo` / `--strip-debuginfo`, which prints / strips locations (and only locations, looked at the strip debuginfo pass).

rfcs/20230517-forward-compatibility-testing.md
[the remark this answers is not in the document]
jpienaar: +1 to making consistent. Even if not allowing editing, its useful to be able to view it and editor plugins make that easy.

rfcs/20230517-forward-compatibility-testing.md
jpienaar: That's just for the textual format, those flags don't affect bytecode. That being said, locations should be retained, for the .mlir file you may want to explicitly specify all locations so that it is roundtrip.

rfcs/20230517-forward-compatibility-testing.md
```diff
@@ -0,0 +1,99 @@
...
+test files.
+
+```bash
+diff %s.bc <(stablehlo-translate --serialize --target=0.X.0)
+```
+
+This test should be run in CI on all PRs to prevent merging incompatible changes
+in StableHLO (_R1_) and LLVM revision bumps (_R2_).
+
+This will not work as is because `bytecode::producer`, `op::location` and
```
jpienaar, line 43: The latter two can be fixed, the former too by setting a define I believe.

rfcs/20230517-forward-compatibility-testing.md
```diff
@@ -0,0 +1,99 @@
...
+compatibility between HEAD and each minor release of StableHLO. Given that
+all changes to the StableHLO opset, as well as changes to serialization
+machinery like targeting a newer version of the MLIR bytecode format, require
+bumping the minor version, there is a low chance of causing forward
+incompatibilities between different patch versions of StableHLO (i.e.
+`0.10.1 → 0.10.2`). Ensuring compatibility between HEAD and all minor releases
+(`0.9.0`, `0.10.0`, etc.) should provide thorough enough coverage of the two
+mentioned sources of forward incompatibilities.
+
+Other benefits of this approach include that it is lightweight (_R3_), can be run
```
jpienaar, line 60: So binary wise, you only build one tool but store multiple versions of the mlirbc file in the repo?

GleasonK: Correct, only build the tool at head, and have `.mlirbc` files from previous releases checked in.

The `.mlirbc` files are not identical, since the file changes as new StableHLO features are introduced, i.e. if new type is added, a test is added to the file we version.

If we did something like this in upstream MLIR it would probably be a file for each bytecode format version number. Ensuring that the test file exercises the interesting pieces of a bytecode file (uses properties, dialect versioning, etc).

rfcs/20230517-forward-compatibility-testing.md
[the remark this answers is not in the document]
GleasonK: Sounds good. Thinking harder about this, the concern is: If a user edits and saves, will this still be forward compatible?

- StableHLO/VHLO: this is not an issue, since we will still be serializing the proper versions of VHLO ops when a user saves.
- Bytecode format: This is an issue, the bytecode format version used currently ([ref](https://github.com/llvm/llvm-project/blob/936dcc89da4f6263ab17e18a85791cd45d86d7cc/mlir/lib/Tools/mlir-lsp-server/MLIRServer.cpp#L889)).

That would be a useful enhancement to add to vscode bytecode editing. I could look into that. Then it would be much safer to allow viewing/editing of StableHLO portable artifacts.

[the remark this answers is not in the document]
jpienaar: Oh yeah I'd not want folks to edit here ... that being said, if we just exposed the "desired bytecode" option to vscode, then that should work.

rfcs/20230517-forward-compatibility-testing.md
GleasonK: Backward compatibility:

```
diff <(stablehlo-translate --deserialize %s.bc | stablehlo-opt) <(stablehlo-opt --strip-debuginfo %s)
#      ^                                                          ^
#      + checked in at 0.10.0 can deserialize at HEAD             |
#                                   HEAD generates the same MLIR  +
```

Forward compatibility:

```
diff %s.bc <(stablehlo-translate --serialize --target=0.X.0)
#    ^       ^
#    + checked in at 0.10.0, was readable at 0.10.0
#            + HEAD can generate exact bytes that worked at 0.10.0
```

Each `.mlirbc` file is checked in prior to a StableHLO release, `0_10_0.mlirbc` is created using `StableHLO@0.10.0`.

- This `.mlirbc` file _must_ be readable at v0.10.0 else tests would have failed in 0.10.0.
- If we are able to generate an exact match of `0_10_0.mlirbc` at `v0.11.0+`, then we know that this file is readable at v0.10.0.

rfcs/20230517-forward-compatibility-testing.md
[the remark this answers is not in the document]
GleasonK: Created https://github.com/openxla/stablehlo/issues/1580

rfcs/20230517-forward-compatibility-testing.md
GleasonK: Yes, changed to that wording.

rfcs/20230517-forward-compatibility-testing.md
GleasonK: I believe `strip-debuginfo` runs a pass on the IR to replace locations with `Unknown`?

We could include them and standardize the directory that tests are run from, that may not work well when StableHLO gets exported to other projects. We could also consider explicitly specifying locations, but that is a bit cumbersome in the test file (2.2k line test file).

rfcs/20230517-forward-compatibility-testing.md
GleasonK: Changed to "full coverage" let's resolve the discussion of full coverage in https://github.com/openxla/stablehlo/pull/1498#discussion_r1206067372, resolving this comment.

rfcs/20230517-forward-compatibility-testing.md
GleasonK: I have a patch that adds the ability to specify producer string, will upstream that as a part of upstream forward compatibility tests, and you're right, we can standardize locations in a test file to take care of the latter two.

rfcs/20230517-forward-compatibility-testing.md
GleasonK: Created https://github.com/openxla/stablehlo/issues/1631

rfcs/20230517-forward-compatibility-testing.md
GleasonK: Created https://github.com/openxla/stablehlo/issues/1633 to experiment with keeping source locations in these test files.

rfcs/20230517-forward-compatibility-testing.md
GleasonK: @burmako let me know if this makes sense now and can be resolved.

burmako: Okay, let's open a ticket to prohibit targeting patch versions and then this comment can be resolved. After that, let's implement this ticket with very high priority.

burmako (approved):

rfcs/20230517-forward-compatibility-testing.md
GleasonK: Created https://github.com/openxla/stablehlo/issues/1653

rfcs/20230517-forward-compatibility-testing.md
GleasonK: Also as a PoC, this plan, which we implemented as a stop-gap solution while this RFC was reviewed, successfully detected an upstream forward incompatibility before we bumped our LLVM revision to that version (see: https://reviews.llvm.org/D151531). The next step (if this testing strategy makes sense) is to add similar tests to upstream MLIR (https://github.com/openxla/stablehlo/issues/1632).

rfcs/20230517-forward-compatibility-testing.md
```diff
@@ -0,0 +1,103 @@
...
+
+This test should be run in CI on all PRs to prevent merging incompatible changes
+in StableHLO (_R1_) and LLVM revision bumps (_R2_).
+
+This will not work as is because `bytecode::producer`, `op::location` and
+`block_argument::location` fields may change between when the original bytecode
+file has been produced and when the test runs at HEAD. We can set a consistent
+producer string to take care of the first item. For debug locations, one option
+would be to explicitly specify debug locations, however this is very cumbersome
+in a 2.2k line test file. Another option is to strip this debug information
```
jpienaar, line 48: There is a simple way: parse and print with debug info. The default behavior is to use file:line:col. That results in every location getting set to the current one.

rfcs/20230517-forward-compatibility-testing.md
jpienaar: I'm assuming the checked in ^ means the bytecode file rather than the tool.

SG. So this is checking bit-equivalence with a version that previously worked.

jpienaar (approved):
OOC where will these files be hosted? e.g., checking thse into github will probably result in a lot of churn and version history issues. Doing it outside github could limit visibility of changes. (well github need not be this repo either :))

rfcs/20230517-forward-compatibility-testing.md
GleasonK: Ah that's a good point, actually I think we can incorporate hard-coded debug locations. Updated https://github.com/openxla/stablehlo/issues/1633 to include this with a proposed plan.

GleasonK:
> OOC where will these files be hosted? e.g., checking thse into github will probably result in a lot of churn and version history issues. Doing it outside github could limit visibility of changes. (well github need not be this repo either :))

Currently we only add tests for minor releases (process described in [vhlo.md > Add versioned serialization test](https://github.com/openxla/stablehlo/blob/main/docs/vhlo.md#add-versioned-serialization-test)).

In [stablehlo/tests/](https://github.com/openxla/stablehlo/tree/main/stablehlo/tests) folder we have a several `stablehlo_legalize_to_vhlo.X_Y_0.mlir` test files, ex: [stablehlo_legalize_to_vhlo.0_12_0.mlir](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/stablehlo_legalize_to_vhlo.0_12_0.mlir). For now this works fine since there aren't that many releases. We can consider moving this to another location as the number of files grows as you mention.

burmako:
Thank you everyone for the great conversation! It looks like all feedback on this RFC has been addressed over an appropriately large review period + there is a prototype solution that works well and has already helped find compatibility breakages. Given that, we can pronounce this RFC approved.

PR #1501, opened 2023-05-19
Incorporate tosa/ tests into our test infrastructure

burmako:
In #1466, we ran into some issues when adding the newly introduced tosa/ suite to relevant test targets, so this PR revamps this system to properly integrate tosa/ as well as simplify similar work in the future.

Instead of the state of the art with one `check-stablehlo` custom target and a bunch of suites becoming its dependencies, we now have three custom targets:
  1) check-stablehlo-ci whose name makes it clear that this is what
     runs in CI.
  2) check-stablehlo-slow for slow-running tests like the testdata/
     suite that we'd like to separate from the other suites, so that
     humans don't have to run them every time.
  3) check-stablehlo-quick for everything else.

Each suite becomes a dependency of either -slow or -quick, making its nature explicit and clearly documented.

Also, it looks like the tosa/ suite got broken by one of the LLVM bumps that happened between when its PR got created and when it got merged. I fixed those breakages too.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,7 +116,21 @@ set(STABLEHLO_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 set(STABLEHLO_BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR})
 set(STABLEHLO_TOOLS_DIR ${CMAKE_BINARY_DIR}/bin)
 
-add_custom_target(check-stablehlo)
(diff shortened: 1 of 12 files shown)


GleasonK (approved):
LGTM. I like the name `check-stablehlo-ci` for clarity.

Could consider swapping `quick` and `slow` with `unit` and `integration` and I believe that is the current distinction between the two?

jpienaar:
Thanks!

PR #1504, opened 2023-05-19
Rename tosa/ passes to follow pass name conventions

burmako:
For the legalization passes, we've been following the foo-legalize-to-bar convention. And for passes in general, we've been adhering to dialectname-something-something. These conventions are applied to the recently introduced tosa/ passes in this PR.

Also, to further the consistency between existing passes and newly introduced passes, this PR removes manual definitions of createFooPass functions. At some point, TableGen got the ability to automatically generate these definitions, so we're leveraging it here.

Thank you, @GleasonK for your feedback that led to the creation of this pull request.


GleasonK (approved):
Ah, after re-reviewing, I like standardizing on the names even more.

`TosaLegalizeStablehlo` is confusing next to `StablehloLegalizeToVhlo`, where the idiom is `SourceLegalizeTarget`. Makes it sound like this is a TOSA->StableHLO conversion.


As discussed, this reduces the copy-pastability of existing users familiar with the mhlo-based passes, but given that this is a new pass and would require using `stablehlo` now, this is a reasonable time to change the naming.

PR #1511, opened 2023-05-21
Use bytecodeVersion = 1 when serializing MLIR bytecode

burmako:
https://reviews.llvm.org/D149515 has just landed, so we must explicitly specify bytecodeVersion when using BytecodeWriter to avoid breaking forward compatibility guarantees.

More specifically, we want to avoid a situation where:
  1) A StableHLO producer using a post-D149515 version of LLVM
     serializes a StableHLO program to bytecode. (This will use
     bytecodeVersion = 2 by default).
  2) A StableHLO consumer using a pre-D149515 version of LLVM within
     the 1 month StableHLO forward compatibility window cannot
     deserialize the StableHLO program.

We don't have forward compatibility tests yet, so this PR doesn't have tests either. See #1498 for an RFC for forward compatibility testing.

### DIFF
--- a/stablehlo/api/PortableApi.cpp
+++ b/stablehlo/api/PortableApi.cpp
@@ -61,6 +61,15 @@ LogicalResult deserializePortableArtifact(StringRef artifactStr,
   loadSerializationDialects(&context);
   auto module = deserializePortableArtifact(artifactStr, &context);
   if (!module) return failure();
+
+  // TODO(gleasonk): I don't think we have formalized our compatibility policy
+  // for bytecode produced by deserializePortableArtifact, so I'm not sure how
+  // this config should look like.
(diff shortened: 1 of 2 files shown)


burmako:
This PR is part of an emergency downstream integrate to accommodate potentially breaking changes in MLIR bytecode implementation: https://reviews.llvm.org/D149515. I will force merge it once CI passes.

@GleasonK Please review post-merge. There are two callsites to `writeBytecodeToFile` which I modified - one is straightforward, but another one was not obvious. I left a todo for you, so that we can figure it out later.

burmako:
This PR introduced BytecodeWriterConfig to our serialization logic, but I forgot to actually use it. #1512 fixes this oversight.

PR #1512, opened 2023-05-21
Actually use the BytecodeWriterConfig in writeBytecodeToFile

burmako:
#1511 introduced BytecodeWriterConfig to our serialization logic,
but I forgot to actually use it. This PR fixes this oversight.

This mistake on my part happened because we don't have forward compatibility tests yet. To make sure that I didn't mess up anything else this time, I have manually verified the following:
  * At HEAD, build/bin/stablehlo-translate --serialize produces different payloads when: 1) local clone of llvm-project is pristine, 2) local clone of llvm-project has kVersion manually changed to 2. The only difference is the bytecodeVersion field in the serialized payloads.
  * With this PR, build/bin/stablehlo-translate --serialize produces the same payloads when: 1) local clone of llvm-project is pristine, 2) local clone of llvm-project has kVersion manually changed to 2. As expected, using BytecodeWriterConfig overrides kVersion.

### DIFF
--- a/stablehlo/api/PortableApi.cpp
+++ b/stablehlo/api/PortableApi.cpp
@@ -70,7 +70,7 @@ LogicalResult deserializePortableArtifact(StringRef artifactStr,
   // ultimately unneeded, but until we've established that, let's play it safe.
   BytecodeWriterConfig writerConfig;
   writerConfig.setDesiredBytecodeVersion(1);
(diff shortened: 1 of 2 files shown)


burmako:
This PR is part of an emergency downstream integrate to accommodate potentially breaking changes in MLIR bytecode implementation: https://reviews.llvm.org/D149515. I will force merge it once CI passes.

PR #1516, opened 2023-05-22
Add producer to StableHLO portable artifacts.

GleasonK:
Producers reads `StableHLO_v<target_version>`, for example:

```
$ stablehlo-translate --serialize --target=0.9.0 file.mlir
ML?RStableHLO_v0.9.0[...]
```

Also moved some logic around and added `minimum` as supported target version now that we have a getMinimumVersion API.

Currently there are no APIs that inspect the producer string, so this is purely debug info and does not impact forward or backward compatibility.

### DIFF
--- a/stablehlo/api/PortableApi.cpp
+++ b/stablehlo/api/PortableApi.cpp
@@ -62,11 +62,8 @@ LogicalResult deserializePortableArtifact(StringRef artifactStr,
   auto module = deserializePortableArtifact(artifactStr, &context);
   if (!module) return failure();
 
-  // This bytecode does not need to specify verison number or header.
-  // This function, like serializePortableArtifact using module string,
-  // provides the reliability of passing / receiving textual assembly
-  // format. Bytecode is used here since it is more compact and faster
-  // to read / write in APIs that require passing strings.
+  // This bytecode does not need to specify verison number or header,
+  // since it is not required to be any more stable than textual assembly.
   return writeBytecodeToFile(*module, os);
 }
 
--- a/stablehlo/api/PortableApi.cpp
+++ b/stablehlo/api/PortableApi.cpp
@@ -62,15 +62,12 @@ LogicalResult deserializePortableArtifact(StringRef artifactStr,
   auto module = deserializePortableArtifact(artifactStr, &context);
   if (!module) return failure();
 
-  // TODO(gleasonk): I don't think we have formalized our compatibility policy
-  // for bytecode produced by deserializePortableArtifact, so I'm not sure how
-  // this config should look like.
-  // With that in mind, bytecodeVersion = 1 looks like the best option at the
-  // moment of writing. The forward compatibility provided by it might be
-  // ultimately unneeded, but until we've established that, let's play it safe.
-  BytecodeWriterConfig writerConfig;
-  writerConfig.setDesiredBytecodeVersion(1);
-  return writeBytecodeToFile(*module, os, writerConfig);
+  // This bytecode does not need to specify verison number or header.
+  // This function, like serializePortableArtifact using module string,
+  // provides the reliability of passing / receiving textual assembly
+  // format. Bytecode is used here since it is more compact and faster
+  // to read / write in APIs that require passing strings.
+  return writeBytecodeToFile(*module, os);
 }
 
 }  // namespace stablehlo
--- a/stablehlo/api/PortableApi.cpp
+++ b/stablehlo/api/PortableApi.cpp
@@ -62,7 +62,7 @@ LogicalResult deserializePortableArtifact(StringRef artifactStr,
   auto module = deserializePortableArtifact(artifactStr, &context);
   if (!module) return failure();
 
-  // This bytecode does not need to specify verison number or header,
+  // This bytecode does not need to specify verison number or producer string,
   // since it is not required to be any more stable than textual assembly.
   return writeBytecodeToFile(*module, os);
 }
--- a/stablehlo/dialect/Serialization.cpp
+++ b/stablehlo/dialect/Serialization.cpp
@@ -28,12 +28,12 @@ namespace mlir {
 namespace stablehlo {
 
 LogicalResult serializePortableArtifact(ModuleOp module,
-                                        StringRef targetVersion,
+                                        StringRef targetVersionStr,
                                         raw_ostream& os) {
   MLIRContext* context = module.getContext();
-  auto version = vhlo::Version::fromString(targetVersion);
-  if (failed(version))
-    return module.emitError("Invalid version string " + targetVersion);
+  auto targetVersion = vhlo::Version::fromAlias(targetVersionStr);
+  if (failed(targetVersion))
+    return module.emitError("Invalid version string " + targetVersionStr);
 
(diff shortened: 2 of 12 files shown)


stablehlo/tests/stablehlo_legalize_to_vhlo.0_9_0.mlir
```diff
@@ -1,6 +1,7 @@
 // RUN: stablehlo-opt --mlir-print-op-generic %s.bc | FileCheck %s
 // RUN: stablehlo-translate --deserialize %s.bc | stablehlo-translate --serialize --target=0.9.0 | stablehlo-opt --mlir-print-op-generic | FileCheck %s
 // RUN: diff <(stablehlo-translate --deserialize %s.bc | stablehlo-opt) <(stablehlo-opt %s)
+// RUN: diff <(stablehlo-translate --serialize --target=minimum %s | stablehlo-translate --deserialize | stablehlo-opt) <(stablehlo-opt %s)
```
burmako, line 4: Am I right in thinking that the purpose of this RUN directive is orthogonal to this specific PR?

stablehlo/transforms/VhloToVersion.cpp
```diff
@@ -67,15 +67,10 @@ class VhloToVersionConverter : public TypeConverter {
   }
 };
 
-FailureOr<Version> parseTargetVersion(llvm::StringRef versionRef) {
-  if (versionRef == "current") return Version::getCurrentVersion();
-  return Version::fromString(versionRef);
-}
```
burmako, line 73: If we're cleaning up version alias parsing, we could also clean up similar logic in getMinVersion / getMaxVersion methods defined in TableGen for VHLO program elements.

stablehlo/tests/stablehlo_legalize_to_vhlo.0_9_0.mlir
GleasonK: Test that "minimum" as a serialization target works? Seems related to this PR?

stablehlo/api/PortableApi.cpp
```diff
@@ -62,15 +62,12 @@ LogicalResult deserializePortableArtifact(StringRef artifactStr,
...
-  // moment of writing. The forward compatibility provided by it might be
-  // ultimately unneeded, but until we've established that, let's play it safe.
-  BytecodeWriterConfig writerConfig;
-  writerConfig.setDesiredBytecodeVersion(1);
-  return writeBytecodeToFile(*module, os, writerConfig);
+  // This bytecode does not need to specify verison number or header.
+  // This function, like serializePortableArtifact using module string,
+  // provides the reliability of passing / receiving textual assembly
+  // format. Bytecode is used here since it is more compact and faster
+  // to read / write in APIs that require passing strings.
```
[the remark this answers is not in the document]
GleasonK: Done.

stablehlo/dialect/Version.cpp
```diff
@@ -49,6 +49,8 @@ static FailureOr<std::array<int64_t, 3>> extractVersionNumbers(
 }  // namespace
 
 FailureOr<Version> Version::fromString(llvm::StringRef versionRef) {
+  if (versionRef == "current") return Version::getCurrentVersion();
+  if (versionRef == "minimum") return Version::getMinimumVersion();
```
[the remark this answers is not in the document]
GleasonK: Agree. But making as a part of this change to avoid the temporary redundancy.

There is some oddity that this aliasing is now needed in two places:
1. Calls to serialize via `stablehlo-translate --serialize --target=current`
2. Explicit pass option to `--vhlo-to-version='target=current'`

Making this change now avoids the redundancy.

stablehlo/dialect/Serialization.cpp
```diff
@@ -31,6 +31,9 @@ LogicalResult serializePortableArtifact(ModuleOp module,
                                         StringRef targetVersion,
                                         raw_ostream& os) {
   MLIRContext* context = module.getContext();
+  auto version = vhlo::Version::fromString(targetVersion);
+  if (failed(version))
```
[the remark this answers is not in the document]
GleasonK: This is used later to write the producer string. Renamed to `targetVersionStr` and `targetVersion`.

stablehlo/dialect/Version.cpp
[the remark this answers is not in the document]
GleasonK: Actually.. we don't need to support aliases in VhloToVersion.. Can make sure call sites use explicit version numbers and bump this logic into `Serialization.cpp`.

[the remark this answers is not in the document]
GleasonK: Moved this logic to `Serialization.cpp` and only accept fully qualified version numbers in `VhloToVersion`.

stablehlo/dialect/Serialization.cpp
```diff
@@ -26,11 +26,23 @@ limitations under the License.
...
+FailureOr<vhlo::Version> parseTargetVersion(StringRef targetVersion) {
+  if (targetVersion == "current") return vhlo::Version::getCurrentVersion();
+  if (targetVersion == "minimum") return vhlo::Version::getMinimumVersion();
+  return vhlo::Version::fromString(targetVersion);
+}
+}  // namespace
 
 LogicalResult serializePortableArtifact(ModuleOp module,
-                                        StringRef targetVersion,
+                                        StringRef targetVersionStr,
```
burmako, line 38: How about renaming the local variable instead, to something like `parsedTargetVersion`? Otherwise, for consistency reasons we'd have to rename the parameter in Serialization.h, as well as in PortableApi.h, as well as the python API (which would also involve bumping the API version).

upd: Given the conversation that we just had, it sounds like we won't need the local variable anymore, so I guess this comment is moot now.

stablehlo/dialect/Version.h
```diff
@@ -35,6 +35,8 @@ class Version {
  public:
   /// Convenience method to extract major, minor, patch and create a Version
   /// from a StringRef of the form `#.#.#`. Returns failure if invalid string.
+  /// Returns current version if input string is "current", and minimum version
+  /// if input string is "minimum".
```
burmako, line 39: This comments seems stale.

stablehlo/tests/stablehlo_legalize_to_vhlo.0_9_0.mlir
burmako: As discussed just now, let's delete this RUN directive?

stablehlo/dialect/Serialization.cpp
```diff
@@ -26,11 +26,23 @@ limitations under the License.
 
 namespace mlir {
 namespace stablehlo {
+namespace {
+FailureOr<vhlo::Version> parseTargetVersion(StringRef targetVersion) {
+  if (targetVersion == "current") return vhlo::Version::getCurrentVersion();
+  if (targetVersion == "minimum") return vhlo::Version::getMinimumVersion();
+  return vhlo::Version::fromString(targetVersion);
+}
+}  // namespace
```
burmako, line 35: Let's move this logic into stablehlo-translate?

burmako (changes requested):

PR #1517, opened 2023-05-22
Bytecode negative tests

GleasonK:
Consolidate VhloToVersion negative tests. Use https://github.com/openxla/stablehlo/pull/1516 as diffbase.


stablehlo/tests/lit.cfg.py
```diff
@@ -33,6 +33,7 @@
   'FileCheck',
   'stablehlo-opt',
   'stablehlo-translate',
+  'not',
```
burmako, line 36: You'll also need to add `not` to https://github.com/openxla/stablehlo/blob/05223052b7e2f387b3534496b0a0109094a6af2b/stablehlo/tests/BUILD.bazel#L142-L148.

burmako (changes requested):

PR #1519, opened 2023-05-22
Add option to stablehlo-translate to generate bytecode without debug info

GleasonK:
Exposes `--strip-debuginfo` as a pass from `stablehlo-translate`. This is currently available in `stablehlo-opt`.

Alternatively, we could use:

```
stablehlo-opt file.mlir --strip-debuginfo --emit-bytecode | stablehlo-translate --serialize --target=X.Y.Z
```

Must use bytecode across the bash pipe, otherwise debug info gets populated with <stdin> values:

```
#loc972 = loc("<stdin>":853:5)
#loc973 = loc("<stdin>":855:3)
#loc975 = loc("<stdin>":856:10)
#loc976 = loc("<stdin>":857:5)
```

### DIFF
--- a/stablehlo/tools/StablehloTranslateMain.cpp
+++ b/stablehlo/tools/StablehloTranslateMain.cpp
@@ -16,10 +16,12 @@ limitations under the License.
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/IR/OpDefinition.h"
 #include "mlir/InitAllDialects.h"
+#include "mlir/Pass/PassManager.h"
 #include "mlir/Support/DebugStringHelper.h"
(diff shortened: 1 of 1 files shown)


GleasonK:
@burmako - Let me know what you think about having this as a part of the translate tool vs using opt+bytecode to generate our versioned test files.

The `stablehlo-opt` tool has access to this pass since it is registered in tablegen with the `strip-debuginfo` name. I'm not sure if theres a good way to use existing pass CLOptions in Translate tooling. I can dig further if we want to try that direction as well.

burmako (approved):
LGTM given that we're in a rush to set up forward compatibility testing. Let's open a ticket to consider a better solution for this.

PR #1520, opened 2023-05-22
Add API to determine bytecode format version of StableHLO release

GleasonK:
Adds a mapping of StableHLO version to bytecode format version. Map only indicates the releases where the format version changes:

```c++
<0.10.0, 1> // bytecode format incremented to v1 in 0.10.0
<0.9.0, 0> // bytecode format started at v0 in 0.9.0
```

Comparison algorithm validates supported version range, and walks the list until it finds a version less or equal to the requested version:

```c++
// <- 0.12.0 is above curr version, failure()
// <- 0.11.0 uses v1 
// <- 0.10.0 uses v1 
<0.10.0, 1>
// <- 0.9.2 uses v0
// <- 0.9.0 uses v0
<0.9.0, 0>
// <- 0.8.0 is below minimum, failure()
```


stablehlo/dialect/Version.h
```diff
@@ -43,6 +43,10 @@ class Version {
   /// Return a Version representing the minimum supported VHLO dialect version.
   static Version getMinimumVersion() { return Version(0, 9, 0); }
 
+  /// Return the MLIR Bytecode Format associated with the input `verison`.
+  /// Returns failure if version is not supported.
+  static FailureOr<int64_t> getBytecodeFormatVersion(const Version& version);
```
burmako, line 48: `getBytecodeVersion` to be consistent with the corresponding MLIR API?

burmako, line 48: Also, can we turn this into an instance method?

stablehlo/dialect/Serialization.cpp
```diff
@@ -53,17 +56,13 @@ LogicalResult serializePortableArtifact(ModuleOp module,
...
-  // serialize portable StableHLO artifacts.
-  // Theoretically speaking, StableHLO v0.9.0 which introduced compatibility
-  // guarantees was released on 3/2/2023 and bytecodeVersion = 1 was released
-  // on 3/10/2023, so there was a time period when we guaranteed compatibility
-  // for StableHLO consumers which only supported bytecodeVersion = 0.
-  // However, this time period (1 month of forward compatibility) has expired,
-  // so it's fine to hardcode bytecodeVersion = 1 here.
-  writerConfig.setDesiredBytecodeVersion(1);
+  auto bytecodeVersion = vhlo::Version::getBytecodeFormatVersion(
+      vhlo::Version::fromString(targetVersion).value());
```
burmako, line 63: What happens if `vhlo::Version::fromString(targetVersion)` fails?

stablehlo/dialect/Version.cpp
```diff
@@ -55,6 +58,35 @@ FailureOr<Version> Version::fromString(llvm::StringRef versionRef) {
...
+  auto it =
+      std::find_if(formatVersionMap.begin(), formatVersionMap.end(),
+                   [&version](auto& pair) { return pair.first <= version; });
+
+  // Should always have value given validation above.
+  if (it == formatVersionMap.end())
+    llvm::report_fatal_error("version value missing from formatVersionMap");
+
+  return it->second;
+}
```
burmako, line 88: I like the declarativeness of `formatVersionMap`, but the current implementation feels like a considerable overkill. How about:

```
FailureOr<int64_t> Version::getBytecodeFormatVersion() {
  if (*this < Version(0, 9, 0)) return failure();
  if (*this < Version(0, 10, 0)) return 0;
  if (*this <= getCurrentVersion()) return 1;
  return failure();
}
```

stablehlo/dialect/Version.cpp
```diff
@@ -55,6 +58,35 @@ FailureOr<Version> Version::fromString(llvm::StringRef versionRef) {
   return Version(versionArr[0], versionArr[1], versionArr[2]);
 }
 
+FailureOr<int64_t> Version::getBytecodeFormatVersion(const Version& version) {
+  // Fail if requested version is not in supported versions.
+  if (vhlo::Version::getCurrentVersion() < version ||
```
burmako, line 63: I don't think we need `vhlo::Version::` here?

stablehlo/dialect/Version.h
```diff
@@ -43,6 +43,10 @@ class Version {
   /// Return a Version representing the minimum supported VHLO dialect version.
   static Version getMinimumVersion() { return Version(0, 9, 0); }
 
+  /// Return the MLIR Bytecode Format associated with the input `verison`.
```
burmako, line 46: Typo: version.

stablehlo/dialect/Version.cpp
```diff
@@ -55,6 +58,35 @@ FailureOr<Version> Version::fromString(llvm::StringRef versionRef) {
...
+    return failure();
+  }
+
+  // This map is updated on all bytecode format increments.
+  // Updating to use a newer bytecode format requires a new minor version
+  // which must be added to this map.
+  const static std::vector<std::pair<Version, int64_t>> formatVersionMap{
+      // All versions greater than first entry use version of first entry.
+      {Version(0, 10, 0), 1},
+      {Version(0, 9, 0), 0},
```
burmako, line 74: I think that in practice most or even all StableHLO portable artifacts with targetVersion = 0.9.0 have bytecodeVersion = 1?

burmako (changes requested):

stablehlo/dialect/Serialization.cpp
GleasonK: Would have errored out a few lines above in:

```
pm.addPass(stablehlo::createVhloToVersionPass({targetVersion.str()}));
```

burmako: Ah, you're right. I think it's fine to not do an extra round of validation here then.

stablehlo/dialect/Version.cpp
GleasonK: If a user requests "bytecode that can run at tagged release v0.9.0", and this API returns `1`, then it will generate bytecode that cannot run at tagged release v0.9.0.

A better way of describing the current situation is "We don't have users running tagged release v0.9.0", which can be deprecated since v0.10.0 was tagged on Apr 20. In the near future we should figure out a deprecation process and deprecate that release as it is no longer in the compatibility window.

burmako (approved):

PR #1521, opened 2023-05-22
Update CustomCallOp status to `yes` and update ODS

ghpvnist:
We already have an implementation to handle custom_call ops by passing a fallback function to handle them. This PR updates the status to reflect that. Other minor change include updating the example in td file to use pretty-print format following the [guide](https://github.com/openxla/stablehlo/blob/main/docs/reference.md#testing-guidelines).

### DIFF
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -2140,13 +2140,10 @@ def StableHLO_CustomCallOp: StableHLO_Op<"custom_call",
 
     Example:
     ```mlir
-    %results = "stablehlo.custom_call"(%input0) {
-      call_target_name = "foo",
-      has_side_effect = false,
+    %results = stablehlo.custom_call @foo(%input0) {
       backend_config = "bar",
-      api_version = 1 : i32,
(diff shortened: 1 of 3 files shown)


stablehlo/dialect/StablehloOps.td
```diff
@@ -2140,11 +2140,8 @@ def StableHLO_CustomCallOp: StableHLO_Op<"custom_call",
 
     Example:
     ```mlir
-    %results = "stablehlo.custom_call"(%input0) {
-      call_target_name = "foo",
-      has_side_effect = false,
+    %results = stablehlo.custom_call @foo(%input0) {
```
sdasgup3, line 2143: Can you check if the pretty printed version is round trippable to preserve the attributes like `api_version` and `has_side_effects`?

ghpvnist: Yes, it will keep all of the relevant info and checked with this command:
```
stablehlo-opt file.mlir | stablehlo-opt
```
Though it will drop values that are already default.

sdasgup3 (approved):

PR #1522, opened 2023-05-22
Update bytecode docs to mention MLIR Bytecode Format and StableHLO releases

GleasonK:
Placed in VHLO markdown file since this is more implementation detail than user-facing documentation.

The plan is to increment the minor release of StableHLO whenever MLIR Bytecode Format updates, so we can keep StableHLO closely tied to the latest MLIR Bytecode Format. This also allows us to provide more strict forward compatibility requirements. I.e. we have no dependency on old bytecode versions a month after the newer bytecode version was adopted.


docs/vhlo.md
```diff
@@ -72,6 +72,24 @@ backends (consumers) only need to support the latest version, which is the
...
+it was built for "the benefits that a binary format brings to the table; namely
+serialization speed and size, mmap capabilities, more easily enabled
+versioning, etc." Performance, serialization size, and memory tests were run
+using large test from various dialects to validate the format.
+
+In order to maintain forward compatibility, StableHLO versions have an
+associated MLIR Bytecode Format version. Additionally, the most recent version
+of StableHLO uses the most up-to-date version of the MLIR Bytecode Format.
+When MLIR bytecode is incremented, the next release of StableHLO will increment
+the minor version and use the latest MLIR Bytecode Format version.
```
burmako, line 91: "the next release of StableHLO will increment the minor version". That only works if the bytecode upgrade is backward compatible, right?

docs/bytecode.md
```diff
@@ -1,18 +0,0 @@
...
-programs. From the [MLIR RFC](https://discourse.llvm.org/t/rfc-a-binary-serialization-format-for-mlir/63518),
-it was built for "the benefits that a binary format brings to the table; namely
-serialization speed and size, mmap capabilities, more easily enabled
-versioning, etc." Performance, serialization size, and memory tests were run
-using large test from various dialects to validate the format.
-
-MLIR bytecode was not specifically built to make MLIR stable, but the MLIR RFC
-notes that it would be possible to provide compatibility guarantees on top of
-this format, which we successfully did for StableHLO
-(see [compatibility.md](compatibility.md)).
```
burmako, line 18: Given the visibility of MLIR bytecode in the upstream ecosystem, I think this file has merit, because it can be quickly used to confirm that StableHLO is indeed using bytecode without going into deep details associated with VHLO. What do you think?

burmako (changes requested):

docs/bytecode.md
GleasonK: That works for me - I'll move these parts back. But also want to note somewhere that StableHLO/VHLO releases are tied to a bytecode release. Do you think this belongs in bytecode.md or should this info go in vhlo.md?

burmako: Agreed that this belongs in vhlo.md.

docs/vhlo.md
GleasonK: Technically all of this has a requirement of forward and backward compatibility. Upstream MLIR has been evolving in a forward/backward compatible fashion so far, would probably be best to consider lack of compatibility as an exception to the guideline.

We could also hold off on updating this document until upstream compatibility discussions take place.

burmako: "would probably be best to consider lack of compatibility as an exception to the guideline". Sounds good. Let's update the document now assuming compatible evolution, and if something changes based on upstream discussions, we can update it again.

burmako (approved):

PR #1524, opened 2023-05-23
Update bytecode artifacts to remove debuginfo and include StableHLO producer string

GleasonK:
Note in the below examples, the suffix `-09`, `-010`, `-011` refers to tooling at the StableHLO@v0.X.0 release:

```bash
# for `-09`:
git checkout v0.9.0
# Checkout LLVM at `./build_tools/llvm_version.txt` version
# Build LLVM and StableHLO
```

Commands used to generate the files below. 

```
stablehlo-opt-v09 stablehlo/tests/stablehlo_legalize_to_vhlo.0_9_0.mlir --strip-debuginfo --stablehlo-legalize-to-vhlo --vhlo-to-version='target=0.9.0' --emit-bytecode | sed 's/MLIR17.0.0git/StableHLO_v0.9.0/' > /tmp/stablehlo_legalize_to_vhlo.0_9_0.mlir.bc
stablehlo-opt-v010 stablehlo/tests/stablehlo_legalize_to_vhlo.0_10_0.mlir --strip-debuginfo --emit-bytecode | stablehlo-translate-v010 --serialize --target=0.10.0 | sed 's/MLIR17.0.0git/StableHLO_v0.10.0/' > /tmp/stablehlo_legalize_to_vhlo.0_10_0.mlir.bc
stablehlo-opt-v011 stablehlo/tests/stablehlo_legalize_to_vhlo.0_11_0.mlir --strip-debuginfo --emit-bytecode | stablehlo-translate-v011 --serialize --target=0.11.0 | sed 's/MLIR17.0.0git/StableHLO_v0.11.0/' > /tmp/stablehlo_legalize_to_vhlo.0_11_0.mlir.bc
```

This is the equivalent way at the time of each release to generate what the following command does today:

```
stablehlo-translate --serialize --target=0.X.0 --strip-debuginfo
```

Lastly, `--strip-debuginfo` is added to the diff check comparison since ReduceOp prettyprinting currently depends on debug info, and the serialized artifacts do not have debuginfo: https://github.com/openxla/stablehlo/blob/4cd6f24257a364857add0e3c1dc11b2364669d50/stablehlo/dialect/StablehloOps.cpp#L1489-L1490

This is an issue that will be addressed separately: https://github.com/openxla/stablehlo/issues/1523

PR #1525, opened 2023-05-23
Add stop-gap static forward compatibility tests

GleasonK:
Use https://github.com/openxla/stablehlo/pull/1524 as diffbase.

This is a stop-gap measure to improve the detection of forward incompatibilities in the StableHLO repo, and repos where StableHLO is exported like openxla/xla, while the Forward Compatibility Testing RFC (https://github.com/openxla/stablehlo/pull/1498) is reviewed. These tests will be reworked based on the outcome of the RFC review.

The forward compatibility test is a byte-wise comparison using:

```bash
# %s = stablehlo/tests/stablehlo_legalize_to_vhlo.0_10_0.mlir
diff %s.bc <(stablehlo-translate --serialize --target=0.10.0 --strip-debuginfo %s)
```


burmako (approved):

burmako:
After merging this PR, let's also: 1) update the RFC explaining that the proposed approach has been tentatively implemented as a stopgap measure, 2) send an email about that to openxla-discuss.

Issue #1405, opened 2023-04-14
Finalize the constraints on QuantizationZeroPoint


sdasgup3:
The goal of the ticket is to resolve the following discussions around the quantization `zero_point`:
1. [Can zero point usage be restricted to 8-bit and lower?](https://github.com/openxla/stablehlo/pull/1352#discussion_r1147883143)
2. [Following the TFLite behavior, should zero_point granularity be restricted to  per-tensor?](https://github.com/openxla/stablehlo/pull/1352#discussion_r1163130568)

**update**
3. Following the TFlite op [behavior](https://www.tensorflow.org/lite/performance/quantization_spec#int8_quantized_operator_specifications), should zero_points be be restricted to certain ranges (e.g., `tanh` op)?

jingpu:
Hi, I didn't get a chance to comment during the discussion meeting today, so I'd add my comment here. I feel 8bit zero point is more of a TFLite x INT8 quant specific choice, and in StableHLO spec it is better to be more general than that. For example, if we think about INT4 quantization, do we want to restrict the zero point to 8bit too? In this case, 8-bit seems a much more magic/arbitrary number in this case (why not 4bit, 5bit, etc.).

igorsafo:
Hi, thanks for the today's discussion!
I think limiting zero point to dst data type range can restrict quantization schemes supported. If we have the following quantization:
```
x_f32 = alpha * (x_u8 - z_x)
min(x_f32) -> u8_min
max(x_f32) -> u8_max

alpha = (max(x_f32) - min(x_f32)) / (u8_max - u8_min)
z_x = u8_max - max(x_f32)/alpha
```

If `0_f32 < min(x_f32)` then `z_x` might not fit into destination range. For example:
```
>>> x = [4, 6]
>>> alpha = (max(x) - min(x)) / (255-0)
>>> z = 255 - max(x)/alpha
>>> alpha
0.00784313725490196
>>> z
-510.0
>>> 4 / alpha + z
0.0
>>> 6 / alpha + z
255.0
```
As you can see the zero point is -510 which doesn't fit into u8 data type.

sdasgup3:
Thanks @igorsafo for bringing this up and thanks for the super useful example.

Let me share a bit of history on why we end up at `type(zero_point) = storage_type`.
[...]

PR #1535, opened 2023-05-24
Add constraints to some quantized ops related to per-axis scheme

sdasgup3:
### Summary 

The PR augments the a few ops, whose specs is already published,  with constraints in the following way:

Constraints added w.r.t the per-axis qunatization scheme,
 - **transpose**:  Transpose the `quantization_dimension` of `output`.
 - **reshape**: Avoid merging or splitting of elements along `quantization dimension` of  `operand`.
 - **slice**: Slice the `zero_points(operand)` and `scales(operand)`.
 - **reverse**: reverse the `scales` and `zero_points` of the  `result` if the `quantization_dimension(operand)` is in `dimensions` to reverse. 
 - **pad**: The op is simplified to use only per-tensor scheme. The rationale for this is as follows:
     - If the `padding_value`  is added in axis other than `quantization_dimension(operand)`, then the `zero_point(padding_value)`  should match the `zero_points(operand)[i]` for all `i`.  That means all the values in `zero_points(operand)` are the same, which is basically the per-tensor case.
     - Only meaningful case is if the `padding_value` is added __only__ along the  `quantization_dimension(operand)`. In this case, the `zero_points(result)` and `scales(result)` need to be updated accordingly. I am not if there is a use case for this particular case and hence kept things simple by allowing on per-tensor scheme for this op. 

Constraints added to comply with other similar specs
 - **add**: For `QuantizedTensorType`, we introduced a new non-terminal `QuantizedElementType` which is different from the non-terminal `ElementType` used for `TensorType`. The proposed change accommodates that. 

PS: The PR involves shuffling the constraint labels (C*). Once the PR reaches a consensus, I will update the verification/type infenrence  code to match the constraint labels.  

**update**:  As per my [proposal](https://groups.google.com/a/openxla.org/g/openxla-discuss/c/iwE9is49SS4/m/mi9AyvKZAQAJ), I have updated the PR to specify the semantics using per-tensor scheme. We can find that it greatly simplifies the specification. With that my proposal is to add per-axis details on a need basis.

**update 06/20/2023**

It turns out that the current PRs becomes "empty" after  a few PRs are merged after the introduction of the current PR. 

I note that:
1.  Some of the  ops addressed in the current PRs (like `transpose, slice, reshape`) have already been introduced with quantized type in #1413 which got merged after the PR was introduced.  
2.  Other ops addressed in the current PRs (like `pad, reverse`) are also been introduced with quantized type in #1477 which got merged after the PR was introduced.  
3. The newly introduced constraints in the current PRs are all simplified while rebasing with https://github.com/openxla/stablehlo/pull/1629.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -4343,8 +4343,6 @@ and produces a `result` tensor. More formally,
 * (C1) `type(operand) = type(result)`.
 * (C2) `is_unique(dimensions)`.
 * (C3) `0 <= dimensions < rank(result)`.
-* If the operation uses quantized tensors:
-  * (C4) `is_quantized_tensor(operand) and is_quantized_tensor(result)`.
 
 #### Examples
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -5382,17 +5382,10 @@ where `result_index[d] = operand_index[permutation[d]]`.
 
 #### Inputs
 
-<<<<<<< HEAD
 | Label | Name          | Type                                         | Constraints |
 |-------|---------------|----------------------------------------------|-------------|
 | (I1)  | `operand`     | tensor or per-tensor quantized tensor        | (C1-C3)     |
 | (I2)  | `permutation` | 1-dimensional tensor constant of type `si64` | (C2), (C3)  |
-=======
-| Label | Name          | Type                                         | Constraints      |
-|-------|---------------|----------------------------------------------|------------------|
-| (I1)  | `operand`     | tensor or quantized tensor                   | (C1-C7)          |
-| (I2)  | `permutation` | 1-dimensional tensor constant of type `si64` | (C1), (C2), (C7) |
->>>>>>> bd17bb47 (Augment the slice op spec with per-axis quantization constraints)
 
 #### Outputs
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -3679,8 +3679,8 @@ More formally, `result[result_index]` is defined as:
   element_type(result)`.
 * (C2) `size(edge_padding_low) = size(edge_padding_high) =
   size(interior_padding) = rank(operand)`.
-* (C3) `0 <= interior_padding`.
-* (C4) `shape(result) = shape(operand) + edge_padding_low +
+* (C2) `0 <= interior_padding`.
+* (C3) `shape(result) = shape(operand) + edge_padding_low +
   max(shape(operand) - 1, 0) * interior_padding + edge_padding_high`.
 
 #### Examples
@@ -4286,18 +4286,22 @@ ordering of `index_space(result)` and `index_space(operand)`.
 
 | Label | Name      | Type                                  | Constraints |
 |-------|-----------|---------------------------------------|-------------|
-| (I1)  | `operand` | tensor or per-tensor quantized tensor | (C1-C2)     |
+| (I1)  | `operand` | tensor or per-tensor quantized tensor | (C1-C4)     |
 
 #### Outputs
 
 | Name     | Type                                  | Constraints |
 |----------|---------------------------------------|-------------|
-| `result` | tensor or per-tensor quantized tensor | (C1-C2)     |
+| `result` | tensor or per-tensor quantized tensor | (C1-C4)     |
 
 #### Constraints
 
-* (C1) `element_type(operand) = element_type(result)`.
-* (C2) `size(operand) = size(result)`.
+* (C1) `size(operand) = size(result)`.
+* If the operation uses non-quantized tensors:
+  * (C2) `element_type(operand) = element_type(result)`.
+* If the operation uses quantized tensors:
+  * (C3) `is_quantized_tensor(operand) and is_quantized_tensor(result)`.
+  * (C4) `quantized_element_type(operand) = quantized_element_type(result)`.
 
 #### Examples
 
@@ -4339,6 +4343,8 @@ and produces a `result` tensor. More formally,
 * (C1) `type(operand) = type(result)`.
 * (C2) `is_unique(dimensions)`.
 * (C3) `0 <= dimensions < rank(result)`.
+* If the operation uses quantized tensors:
+  * (C4) `is_quantized_tensor(operand) and is_quantized_tensor(result)`.
 
 #### Examples
 
@@ -5391,7 +5397,7 @@ where `result_index[d] = operand_index[permutation[d]]`.
 
 | Name     | Type                                  | Constraints |
 |----------|---------------------------------------|-------------|
-| `result` | tensor or per-tensor quantized tensor | (C1), (C3)  |
+| `result` | tensor or per-tensor quantized tensor | (C2-C5)     |
 
 #### Constraints
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -3679,8 +3679,8 @@ More formally, `result[result_index]` is defined as:
   element_type(result)`.
 * (C2) `size(edge_padding_low) = size(edge_padding_high) =
   size(interior_padding) = rank(operand)`.
-* (C2) `0 <= interior_padding`.
-* (C3) `shape(result) = shape(operand) + edge_padding_low +
+* (C3) `0 <= interior_padding`.
+* (C4) `shape(result) = shape(operand) + edge_padding_low +
   max(shape(operand) - 1, 0) * interior_padding + edge_padding_high`.
 
 #### Examples
@@ -4286,22 +4286,18 @@ ordering of `index_space(result)` and `index_space(operand)`.
 
 | Label | Name      | Type                                  | Constraints |
 |-------|-----------|---------------------------------------|-------------|
-| (I1)  | `operand` | tensor or per-tensor quantized tensor | (C1-C4)     |
+| (I1)  | `operand` | tensor or per-tensor quantized tensor | (C1-C2)     |
 
 #### Outputs
 
 | Name     | Type                                  | Constraints |
 |----------|---------------------------------------|-------------|
-| `result` | tensor or per-tensor quantized tensor | (C1-C4)     |
+| `result` | tensor or per-tensor quantized tensor | (C1-C2)     |
 
 #### Constraints
 
-* (C1) `size(operand) = size(result)`.
-* If the operation uses non-quantized tensors:
-  * (C2) `element_type(operand) = element_type(result)`.
-* If the operation uses quantized tensors:
-  * (C3) `is_quantized_tensor(operand) and is_quantized_tensor(result)`.
-  * (C4) `quantized_element_type(operand) = quantized_element_type(result)`.
+* (C1) `element_type(operand) = element_type(result)`.
+* (C2) `size(operand) = size(result)`.
 
 #### Examples
 
@@ -5395,7 +5391,7 @@ where `result_index[d] = operand_index[permutation[d]]`.
 
 | Name     | Type                                  | Constraints |
 |----------|---------------------------------------|-------------|
-| `result` | tensor or per-tensor quantized tensor | (C2-C5)     |
+| `result` | tensor or per-tensor quantized tensor | (C1), (C3)  |
 
 #### Constraints
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -5382,10 +5382,17 @@ where `result_index[d] = operand_index[permutation[d]]`.
 
 #### Inputs
 
+<<<<<<< HEAD
 | Label | Name          | Type                                         | Constraints |
 |-------|---------------|----------------------------------------------|-------------|
 | (I1)  | `operand`     | tensor or per-tensor quantized tensor        | (C1-C3)     |
 | (I2)  | `permutation` | 1-dimensional tensor constant of type `si64` | (C2), (C3)  |
+=======
+| Label | Name          | Type                                         | Constraints      |
+|-------|---------------|----------------------------------------------|------------------|
+| (I1)  | `operand`     | tensor or quantized tensor                   | (C1-C7)          |
+| (I2)  | `permutation` | 1-dimensional tensor constant of type `si64` | (C1), (C2), (C7) |
+>>>>>>> bd17bb47 (Augment the slice op spec with per-axis quantization constraints)
 
 #### Outputs
 


docs/spec.md
```diff
@@ -4323,18 +4330,34 @@ spaces of `result` and `operand`.
...
+  * If `quantization_dimension(operand)` is non-empty:
+    * (C5) `quantized_element_type(operand) = quantized_element_type(result)`,
+      expect for the `quantization_dimension(operand)` and
+      `quantization_dimension(result)` which may differ.
+    * (C6) Given `k = quantization_dimension(operand)`,
+      `k' = quantization_dimension(result)`,
+      `shape(operand) = [d0, d1, ..., dk, ..., dR-1]`, and
+      `shape(result) = [s0, s1, ..., sk', ..., sR'-1]`, then
+       * `dk = sk'`.
+       * `d0 x ... x dk-1 = s0 x ... x sk'-1`.
```
loganchien, line 4359: Use `*` instead of `x` for multiplication.

docs/spec.md
```diff
@@ -4357,23 +4380,38 @@ and produces a `result` tensor. More formally,
 
 #### Inputs
 
-| Label | Name         | Type                                         | Constraints |
-|-------|--------------|----------------------------------------------|-------------|
-| (I1)  | `operand`    | tensor                                       | (C1)        |
-| (I2)  | `dimensions` | 1-dimensional tensor constant of type `si64` | (C2), (C3)  |
+| Label | Name         | Type                                         | Constraints      |
+|-------|--------------|----------------------------------------------|------------------|
+| (I1)  | `operand`    | tensor   or quantized tensor                 | (C3-C8)          |
```
loganchien, line 4385: Remove two extra space characters between `tensor` and `or`.

loganchien (approved):

sdasgup3:
Hello reviewers,

I am seeking [community discussion](https://groups.google.com/a/openxla.org/g/openxla-discuss/c/iwE9is49SS4/m/mi9AyvKZAQAJ) on the operations to consider for per-axis quantization scheme. Please find the link to my rationale in the same post and feel free to add your input. I believe that prior discussion on this topic has the potential to simplify the spec.

sdasgup3:
Hello All,

As per my  [proposal](https://groups.google.com/a/openxla.org/g/openxla-discuss/c/iwE9is49SS4/m/mi9AyvKZAQAJ),  I have updated the PR to specify the semantics using per-tensor scheme. We can find that it greatly simplifies the specification. With that my proposal is to add per-axis details on a need basis. I would love to hear your opinion on ops which should be considered using per-axis scheme.

I note that all the updates leading to this current change are available as separate code commits, which will help us navigate/switch between them, if needed.

qedawkins:
> Hello All,
> 
> As per my [proposal](https://groups.google.com/a/openxla.org/g/openxla-discuss/c/iwE9is49SS4/m/mi9AyvKZAQAJ), I have updated the PR to specify the semantics using per-tensor scheme. We can find that it greatly simplifies the specification. With that my proposal is to add per-axis details on a need basis. I would love to hear your opinion on ops which should be considered using per-axis scheme.
> 
> I note that all the updates leading to this current change are available as separate code commits, which will help us navigate/switch between them, if needed.

Are there any plans to support finer quantization granularities than per-axis? In particular, "grouped" scales/zero points such as that seen in GPTQ (3g128 in https://github.com/IST-DASLab/gptq#new-features). These finer grain quantization schemes are gaining traction in LLMs ([LLaMA + others](https://github.com/ggerganov/llama.cpp), [+ GPTQ](https://github.com/qwopqwop200/GPTQ-for-LLaMa)) and it would be cool to see support for this in StableHLO. I've only seen this for convolution and GEMM so far so I would expect it to be a generalization of per-axis (happening in [brevitas](https://github.com/Xilinx/brevitas/tree/llm) also).

powderluv:
Agree with @qedawkins . GPTQ like quantization is table stakes today for LLMs and would be nice to make sure we can support it.

stellaraccident:
> Agree with @qedawkins . GPTQ like quantization is table stakes today for LLMs and would be nice to make sure we can support it.

Agreed. Per-axis is pretty old at this point and I would like to see us have representational capacity for the more modern schemes.

sdasgup3:
Thanks @qedawkins, @powderluv and @stellaraccident for your inputs!

Given the use-cases cited here and with some of my quick exploration, it seems an very interesting direction to make progress on. I have opened a [ticket](https://github.com/openxla/stablehlo/issues/1569)  to explore and track this item. 

Currently, we are focusing on making progress towards [v1.0](https://github.com/openxla/stablehlo/blob/main/docs/roadmap.md#milestones) which includes adding the quantization specification modelled roughly after [TFLite quantization spec](https://www.tensorflow.org/lite/performance/quantization_spec) and, unfortunately, that might affect addressing this item right away. 

In any case, let me know if there is strong use case which demands prioritizing this item. Also, I would highly encourage contribution in the form of an RFC, with details about the change, which we can start brainstorming on early.

stellaraccident:
> In any case, let me know if there is strong use case which demands prioritizing this item. Also, I would highly encourage contribution in the form of an RFC, with details about the change, which we can start brainstorming on early.

LLMs (like the LLamA derivitives with GPTQ) are a quite strong use case. Stablehlo not being able to model modern workloads represents one of the main reasons we have to lower directly from framework to internal IR for things like IREE.

I think the priorities are inverted.

powderluv:
There are a lot of silicon partners / designs we (nod.ai) team supports.  Our tested path right now is PyTorch->Torch-mlir->LinAlg-on-Tensors->... There were a few TOSA implementations but we are recommending moving customers over to StableHLO. All of these customers _have_ to support (Quantized) LLMs / Generative AI workloads on day 1. It would be unfortunate for StableHLO "1.0" release without GPTQ support (int4/int5/int3/int8 in that order). 

We are happy to participate in the RFC process - but we probably want to do some prep work to try and lower [important](https://huggingface.co/spaces/HuggingFaceH4/open_llm_leaderboard) quantized LLMs (Torch-MLIR-->StableHLO) and then fill in how the ops have to look before reaching RFC stage. 

(BTW no-objections to this RFC by itself)

docs/spec.md
```diff
@@ -603,24 +603,25 @@ Performs element-wise addition of two tensors `lhs` and `rhs` and produces a
...
 
 #### Constraints
 
 * (C1) `shape(lhs) = shape(rhs) = shape(result)`.
 * If the operation uses non-quantized tensors:
   * (C2) `element_type(lhs) = element_type(rhs) = element_type(result)`.
 * If the operation uses quantized tensors:
-  * (C3) `element_type(lhs) = element_type(rhs) = element_type(result)`,
+  * (C3) `is_quantized_tensor(lhs) and is_quantized_tensor(rhs) and is_quantized_tensor(result)`.
+  * (C4) `quantized_element_type(lhs) = quantized_element_type(rhs) = quantized_element_type(result)`,
```
burmako, line 622: Let's keep `element_type` here and leave potential tightening to #1556. Overall, I think it'd be okay to say that element_type can produce both regular and quantized element types - that'll be considerably more concise than the alternative.

docs/spec.md
```diff
@@ -3703,30 +3704,38 @@ More formally, `result[i0, ..., iR-1]` is equal to:
...
-|----------|--------|-------------|
-| `result` | tensor | (C1)        |
+| Name     | Type                       | Constraints |
+|----------|----------------------------|-------------|
+| `result` | tensor or quantized tensor | (C3-C6)     |
 
 #### Constraints
 
 <!-- markdownlint-disable line-length -->
-* (C1) `operand`, `padding_value`, `result` have the same element type.
```
burmako, line 3723: Similarly, we can keep C1 as is because it applies to both cases is we allow element_type to produce both regular and quantized element types. Then below, we'll only have one case to take care of - "if the operation uses quantized tensors". (The same applies for all other ops below).

docs/spec.md
```diff
@@ -4336,18 +4345,23 @@ spaces of `result` and `operand`.
...
 
-* (C1) `operand` and `result` have the same element type.
-* (C2) `operand` and `result` have the same number of elements.
+* (C1) `operand` and `result` have the same number of elements.
+* If the operation uses non-quantized tensors:
+  * (C2) `operand` and `result` have the same element type.
+* If the operation uses quantized tensors:
+  * (C3) `is_quantized_tensor(operand) and is_quantized_tensor(result)`.
+  * (C4) `quantized_element_type(operand) = quantized_element_type(result)`.
+  * (C5) `is_empty(quantization_dimension(operand))`.
```
burmako, line 4364: It would appear that reshaping a per-axis quantized tensor is needed to define semantics of DotGeneralOp per the definition of the op. What do we do about that? (Same question about SliceOp and TransposeOp).

burmako (changes requested):

docs/spec.md
sdasgup3: Let us discuss this [here](https://github.com/openxla/stablehlo/pull/1566#discussion_r1220364289).

burmako: Moved the discussion thread here: https://github.com/openxla/stablehlo/pull/1567#discussion_r1217167342.

docs/spec.md
```diff
@@ -3652,29 +3652,35 @@ More formally, `result[result_index]` is defined as:
...
   size(interior_padding) = rank(operand)`.
-* (C3) `0 <= interior_padding`.
-* (C4) `shape(result) = shape(operand) + edge_padding_low +
+* (C2) `0 <= interior_padding`.
+* (C3) `shape(result) = shape(operand) + edge_padding_low +
   max(shape(operand) - 1, 0) * interior_padding + edge_padding_high`.
+* (C4) If the operation uses non-quantized tensors:
+  * `element_type(operand) = element_type(padding_value) =
+  element_type(result)`.
+* If the operation uses quantized tensors:
```
abattery, line 3679: for pad op, it might be safe to consider this operator required the same quantization parameter.

docs/spec.md
```diff
@@ -4279,18 +4285,22 @@ ordering of `index_space(result)` and `index_space(operand)`.
 
 | Label | Name      | Type                                  | Constraints |
 |-------|-----------|---------------------------------------|-------------|
-| (I1)  | `operand` | tensor or per-tensor quantized tensor | (C1-C2)     |
+| (I1)  | `operand` | tensor or per-tensor quantized tensor | (C1-C4)     |
```
abattery, line 4288: Is there any specific use case for applying to the MHLO.replica_id op with quantization?

Could you explain why we need to add certain restriction (e.g., allowing only per-tensor quantized tensor in the parallelism operator)?

docs/spec.md
```diff
@@ -4318,14 +4328,14 @@ and produces a `result` tensor. More formally,
 
 | Label | Name         | Type                                         | Constraints |
 |-------|--------------|----------------------------------------------|-------------|
-| (I1)  | `operand`    | tensor                                       | (C1)        |
-| (I2)  | `dimensions` | 1-dimensional tensor constant of type `si64` | (C2), (C3)  |
+| (I1)  | `operand`    | tensor or per-tensor quantized tensor        | (C1)        |
```
abattery, line 4331: For reshape op, it might be better to have a constraint where the same quantization parameters over the given operand and result.

docs/spec.md
```diff
@@ -5377,20 +5391,24 @@ where `result_index[d] = operand_index[permutation[d]]`.
...
 
-* (C1) `element_type(operand) = element_type(result)`.
-* (C2) `permutation` is a permutation of `range(rank(operand))`.
-* (C3) `shape(result) = dim(operand, permutation...)`.
+* (C1) `permutation` is a permutation of `range(rank(operand))`.
+* (C2) `shape(result) = dim(operand, permutation...)`.
+* If the operation uses non-quantized tensors:
+  * (C3) `element_type(operand) = element_type(result)`.
+* If the operation uses quantized tensors:
+  * (C4) `is_quantized_tensor(operand) and is_quantized_tensor(result)`.
```
abattery, line 5410: For tanh operator, TFLite quantization prefers having a fixed quantization parameter range, instead of allowing the arbitrary quantization range. I wonder that kind of the restriction makes sense to the StableHLO quantization spec.

docs/spec.md
sdasgup3: That is good point. I would propose to address that as part of https://github.com/openxla/stablehlo/issues/1628. Added to [comment](https://github.com/openxla/stablehlo/issues/1628#issuecomment-1599241549) to make sure we do not miss it.

docs/spec.md
sdasgup3: Per
```
* (C1) `element_type(operand) = element_type(padding_value) =
  element_type(result)`.
```

we make sure that the quantization parameters are the same. I note that `element_type` is an overloaded auxiliary function applicable to both `TensorType` and `QuantizedTensorType`. PTAL.

docs/spec.md
sdasgup3: That is good point. Thanks!
I will  keep this point in mind while proposing the specification for parallelism operators.

docs/spec.md
sdasgup3: The constraint 
```
* (C1) `element_type(operand) = element_type(result)`.
```

ascertains that. PTAL.

docs/spec.md
sdasgup3: Thanks @abattery for pointing the observation.  

I agree that StableHLO spec, in its current form, does not have the restriction ` (scale, zero_point) = (1.0 / 128.0, 0)`. The decision was deliberate for the initial version of the spec to be "open" preferring generality  and symmetry of IR design. On a related note, we have a open ticket https://github.com/openxla/stablehlo/issues/1405 to discuss similar issue. Also, updated the description of the issue to capture your point. Please let me know if it makes sense.

burmako:
Reassigning - the PR appears to be empty.

sdasgup3:
>> Reassigning - the PR appears to be empty.

Thanks for pointing that out. 
The reason for the PR to be empty are the following:
1.  Some of the  ops addressed in the current PRs (like transpose, slice, reshape) have already been introduced with quantized type in #1413 which got merged after the PR was introduced.  
2.  Other ops addressed in the current PRs (like pad, reverse) are also been introduced with quantized type in #1477 which got recently merged. 
3. The newly introduced constraints in the current PRs are all simplified while rebasing with https://github.com/openxla/stablehlo/pull/1629.


All in all the effect is that the PR does not propose any changes. The spec in the repo HEAD should specific the  quantized semantics of the ops referred in the PR description.    

My proposal is to keep the PR to as is till the feed backs are resolved after which we can close it. 

@burmako There is  one follow up [thread](https://github.com/openxla/stablehlo/pull/1535#discussion_r1217039774) for you to look at in this current PR. 

@abattery Please let me know if your concerns are properly addressed.

docs/spec.md
burmako: Sounds like a plan.

burmako (approved):

docs/spec.md
paulinesho: Could we more precisely define what it means for element types to be equivalent elsewhere in the spec before it's widely replicated? I'm afraid it's not immediately clear to me.

sdasgup3: Thanks @paulinesho for bringing this up. 

Recently we have re-defined the semantics of `element_type` to be  inclusive for both `TensorType` and `QuantizedTensorType`. Earlier it applies only to `TensorType`. You can find the semantics [here](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#functions-on-types).

In this context, it means
1. The element type of `operand` and `result` should be same. That is, either they are both non-quantized element types or both quantized element type.
2. In the event when `operand` and `result` are quantized element type, then all the components of the `QuantizedElementType` grammar rule, like `storage_type`, `expressed_type`, `scales`, `zero_points` etc., will be the same for `operand` and `result`.

Please let me know if that clarifies. 
Again, sorry @paulinesho and @abattery for the confusion.

paulinesho (approved):

abattery:
+1

> With that my proposal is to add per-axis details on a need basis.

abattery (approved):

PR #1538, opened 2023-05-25
Specification for quantized ReduceOp, ReduceWindowOp, and SelectAndScatterOp

sdasgup3:
The PR proposes the specification of quantized `ReduceOp`,  `ReduceWindowOp`, and `SelectAndScatterOp`.

For ReduceOp (and the other proposed ops using ReduceOp in their specification ), the specification allows different quantized element types for `inputs` (or `init_values`) and the `results` .  The reason is to allow the accumulation to happen in higher storage_type. However, the quantized element types for `inputs` and `init_values` are the same. This behavior is same as other reduction ops like dot_general.

The following shows an example usage of reduce op:

```
%reduce = "stablehlo.reduce"(%input, %init_value) ({
    ^bb0(%elem1: tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>,
         %elem2: tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>):

        %elem1_rescaled = "stablehlo.uniform_quantize"(%elem1)
            : (tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>)
            -> tensor<!quant.uniform<i32:f32, output_scale:output_zp>>

         %elem1_rescaled = "stablehlo.uniform_quantize"(%elem2)
            : (tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>)
            -> tensor<!quant.uniform<i32:f32, output_scale:output_zp>>


        %add = "stablehlo.add"(%elem1_rescaled, %%elem2_rescaled)
            : (tensor<!quant.uniform<i32:f32, output_scale:output_zp>>,
               tensor<!quant.uniform<i32:f32, output_scale:output_zp>>)
            -> tensor<!quant.uniform<i32:f32, output_scale:output_zp>>

        "stablehlo.return"(%add)
            : (tensor<!quant.uniform<i32:f32, output_scale:output_zp>>) -> ()
    }) {
        dimensions = dense<...> : tensor<1xi64>
    } : (tensor<... x !quant.uniform<ui8:f32, input_scale:input_zp>,
         tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>)
    -> tensor<... x !quant.uniform<i32:f32, output_scale:outout_zp>>
```

Note that, in the example above, like the non-quantized version of reduce op, the type of blocks arguments  are the same. That is because from the POV of the semantic, there are any two elements to be added. The bitwidth of the accumulation is captured via output quantized element type.  

Please let me know your review comments.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -3955,33 +3955,41 @@ More formally, `results[:][j0, ..., jR-1] = reduce(input_slices)` where:
 
 #### Inputs
 
-| Label | Name          | Type                                         | Constraints         |
-|-------|---------------|----------------------------------------------|---------------------|
-| (I1)  | `inputs`      | variadic number of tensors                   | (C1-C4), (C6), (C7) |
-| (I2)  | `init_values` | variadic number of 0-dimensional tensors     | (C2), (C3)          |
-| (I3)  | `dimensions`  | 1-dimensional tensor constant of type `si64` | (C4), (C5), (C7)    |
-| (I4)  | `body`        | function                                     | (C6)                |
+| Label | Name          | Type                                                          | Constraints            |
+|-------|---------------|---------------------------------------------------------------|------------------------|
+| (I1)  | `inputs`      | variadic number of tensors or quantized tensor                | (C1-C3), (C5-C10)      |
+| (I2)  | `init_values` | variadic number of 0-dimensional tensors or quantized tensors | (C2), (C6), (C8), (C9) |
+| (I3)  | `dimensions`  | 1-dimensional tensor constant of type `si64`                  | (C3-C5)                |
+| (I4)  | `body`        | function                                                      | (C7), (C10)            |
 
 #### Outputs
 
-| Name      | Type                       | Constraints      |
-|-----------|----------------------------|------------------|
-| `results` | variadic number of tensors | (C2), (C3), (C7) |
+| Name      | Type                                            | Constraints                |
+|-----------|-------------------------------------------------|----------------------------|
+| `results` | variadic number of tensors or quantized tensors | (C2), (C5-C6), (C8), (C10) |
 
 #### Constraints
 
 * (C1) All `inputs` have the same shape.
-* (C2) element_type(`inputs[k]`) $=$ element_type(`init_values[k]`) $=$
-  element_type(`results[k]`) for all `k` $\in$ [0, N).
-* (C3) size(`inputs`) $=$ size(`init_values`) $=$ size(`results`) $=$ N where
+* (C2) size(`inputs`) $=$ size(`init_values`) $=$ size(`results`) $=$ N where
   N >= 1.
-* (C4) 0 $\le$ `dimensions[d]` $\lt$ rank(`inputs[0][d]`) for all dimension `d`.
-* (C5) All dimensions in `dimensions` are unique.
-* (C6) `body` has type `(tensor<E0>, ..., tensor<EN-1>, tensor<E0>, ...,`
-  `tensor<EN-1>) -> (tensor<E0>, ..., tensor<EN-1>)` where
-  `Ek = element_type(inputs[k])`.
-* (C7) shape(`results[k]`) $=$ shape(`inputs[k]`) except that the dimension
+* (C3) 0 $\le$ `dimensions[d]` $\lt$ rank(`inputs[0][d]`) for all dimension `d`.
+* (C4) All dimensions in `dimensions` are unique.
+* (C5) shape(`results[k]`) $=$ shape(`inputs[k]`) except that the dimension
   sizes of `inputs[k]` corresponding to `dimensions` are not included.
+* If the operation uses non-quantized tensors:
+  * (C6) element_type(`inputs[k]`) $=$ element_type(`init_values[k]`) $=$
+    element_type(`results[k]`) for all `k` $\in$ [0, N).
+  * (C7) `body` has type `(tensor<E0>, ..., tensor<EN-1>, tensor<E0>, ...,`
+    `tensor<EN-1>) -> (tensor<E0>, ..., tensor<EN-1>)` where
+    `Ek = element_type(inputs[k])`.
+* If the operation uses quantized tensors:
+  * (C8) `is_quantized_tensor(inputs) and is_quantized_tensor(init_values) and is_quantized_tensor(results)`.
+  * (C9) quantized_element_type(`inputs[k]`) $=$
+    quantized_element_type(`init_values[k]`)  for all `k` $\in$ [0, N).
+  * (C10) `body` has type `(tensor<E0>, ..., tensor<EN-1>, tensor<E0>, ...,`
+    `tensor<EN-1>) -> (tensor<E'0>, ..., tensor<E'N-1>)` where
+    `Ek = quantized_element_type(inputs[k]) and E'k = quantized_element_type(results[k])`.
 
 #### Examples
 
@@ -4178,22 +4186,22 @@ where:
 
 #### Inputs
 
-| Label | Name                | Type                                         | Constraints                                     |
-|-------|---------------------|----------------------------------------------|-------------------------------------------------|
-| (I1)  | `inputs`            | variadic number of tensors                   | (C1-C4), (C6), (C8), (C10), (C12), (C13), (C15) |
-| (I2)  | `init_values`       | variadic number of 0-dimensional tensors     | (C1), (C13), (C16)                              |
-| (I3)  | `window_dimensions` | 1-dimensional tensor constant of type `si64` | (C4), (C5), (C15)                               |
-| (I4)  | `window_strides`    | 1-dimensional tensor constant of type `si64` | (C6), (C7), (C15)                               |
-| (I5)  | `base_dilations`    | 1-dimensional tensor constant of type `si64` | (C8), (C9), (C15)                               |
-| (I6)  | `window_dilations`  | 1-dimensional tensor constant of type `si64` | (C10), (C11), (C15)                             |
-| (I7)  | `padding`           | 2-dimensional tensor constant of type `si64` | (C12), (C15)                                    |
-| (I8)  | `body`              | function                                     | (C13)                                           |
+| Label | Name                | Type                                                          | Constraints                                 |
+|-------|---------------------|---------------------------------------------------------------|---------------------------------------------|
+| (I1)  | `inputs`            | variadic number of tensors or quantized tensors               | (C1-C3), (C5), (C7), (C9), (C11), (C13-C18) |
+| (I2)  | `init_values`       | variadic number of 0-dimensional tensors or quantized tensors | (C1), (C14), (C16-C17)                      |
+| (I3)  | `window_dimensions` | 1-dimensional tensor constant of type `si64`                  | (C3), (C4), (C13)                           |
+| (I4)  | `window_strides`    | 1-dimensional tensor constant of type `si64`                  | (C5-C6), (C13)                              |
+| (I5)  | `base_dilations`    | 1-dimensional tensor constant of type `si64`                  | (C7-C8), (C13)                              |
+| (I6)  | `window_dilations`  | 1-dimensional tensor constant of type `si64`                  | (C9), (C10), (C13)                          |
+| (I7)  | `padding`           | 2-dimensional tensor constant of type `si64`                  | (C11), (C13)                                |
+| (I8)  | `body`              | function                                                      | (C15), (C18)                                |
 
 #### Outputs
 
-| Name      | Type                       | Constraints     |
-|-----------|----------------------------|-----------------|
-| `results` | variadic number of tensors | (C1), (C14-C16) |
+| Name      | Type                                            | Constraints                   |
+|-----------|-------------------------------------------------|-------------------------------|
+| `results` | variadic number of tensors or quantized tensors | (C1), (C12-C14), (C16), (C18) |
 
 #### Constraints
 
@@ -4201,27 +4209,34 @@ where:
 * (C1) size(`inputs`) $=$ size(`init_values`) $=$ size(`results`) $=$ N and
   N $\ge$ 1.
 * (C2) All `inputs` have the same shape.
-* (C3) `element_type(inputs[k]) = element_type(init_values[k])` for all k
-  $\in$ [0, N).
-* (C4) size(`window_dimensions`) $=$ rank(`inputs[0]`).
-* (C5) `window_dimensions[i]` $\gt 0$ for all i $\in$ [0, size(`window_dimensions`)).
-* (C6) size(`window_strides`) $=$ rank(`inputs[0]`).
-* (C7) `window_strides[i]` $\gt 0$ for all i $\in$ [0, size(`window_strides`)).
-* (C8) size(`base_dilations`) $=$ rank(`inputs[0]`).
-* (C9) `base_dilations[i]` $\gt 0$ for all i $\in$ [0, size(`base_dilations`)).
-* (C10) size(`window_dilations`) $=$ rank(`inputs[0]`).
-* (C11) `window_dilations[i]` $\gt 0$ for all i $\in$ [0, size(`window_dilations`)).
-* (C12) dim(`padding`, 0) $=$ rank(`inputs[0]`) and dim(`padding`, 1) = 2.
-* (C13) `body` has type `(tensor<E0>, ..., tensor<EN-1>, tensor<E0>, ..., tensor<EN-1>) -> (tensor<E0>, ..., tensor<EN-1>)`
-  where `Ek = element_type(inputs[0])`.
-* (C14) All `results` have the same shape.
-* (C15) `shape(results[0]) = num_windows`
+* (C3) size(`window_dimensions`) $=$ rank(`inputs[0]`).
+* (C4) `window_dimensions[i]` $\gt 0$ for all i $\in$ [0, size(`window_dimensions`)).
+* (C5) size(`window_strides`) $=$ rank(`inputs[0]`).
+* (C6) `window_strides[i]` $\gt 0$ for all i $\in$ [0, size(`window_strides`)).
+* (C7) size(`base_dilations`) $=$ rank(`inputs[0]`).
+* (C8) `base_dilations[i]` $\gt 0$ for all i $\in$ [0, size(`base_dilations`)).
+* (C9) size(`window_dilations`) $=$ rank(`inputs[0]`).
+* (C10) `window_dilations[i]` $\gt 0$ for all i $\in$ [0, size(`window_dilations`)).
+* (C11) dim(`padding`, 0) $=$ rank(`inputs[0]`) and dim(`padding`, 1) = 2.
+* (C12) All `results` have the same shape.
+* (C13) `shape(results[0]) = num_windows`
   * `dilated_input_shape = shape(inputs[0]) == 0 ? 0 : (shape(inputs[0]) - 1) * base_dilations + 1`.
   * `padded_input_shape = padding[:, 0] + dilated_input_shape + padding[:, 1]`.
   * `dilated_window_shape = window_dimensions == 0 ? 0 : (window_dimensions - 1) * window_dilations + 1`.
   * `num_windows = (padded_input_shape == 0 || dilated_window_shape > padded_input_shape) ? 0 : floor((padded_input_shape - dilated_window_shape) / window_strides) + 1`.
-* (C16) `element_type(results[k]) = element_type(init_values[k])` for all k
-  $\in$ [0, N).
+* If the operation uses non-quantized tensors:
+  * (C14) `element_type(inputs[k]) = element_type(init_values[k]) = element_type(results[k])` for all k
+    $\in$ [0, N).
+  * (C15) `body` has type `(tensor<E0>, ..., tensor<EN-1>, tensor<E0>, ..., tensor<EN-1>) -> (tensor<E0>, ..., tensor<EN-1>)`
+    where `Ek = element_type(inputs[0])`.
+* If the operation uses quantized tensors:
+  * (C16) `is_quantized_tensor(inputs) and is_quantized_tensor(init_values) and is_quantized_tensor(results)`.
+  * (C17) quantized_element_type(`inputs[k]`) $=$
+    quantized_element_type(`init_values[k]`)  for all `k` $\in$ [0, N).
+  * (C18) `body` has type `(tensor<E0>, ..., tensor<EN-1>, tensor<E0>, ...,`
+    `tensor<EN-1>) -> (tensor<E'0>, ..., tensor<E'N-1>)` where
+    `Ek = quantized_element_type(inputs[k]) and E'k = quantized_element_type(results[k])`.
+
 <!-- markdownlint-enable line-length -->
 
 #### Examples
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -4870,39 +4870,48 @@ More formally:
 
 | Label | Name                | Type                                         | Constraints             |
 |-------|---------------------|----------------------------------------------|-------------------------|
-| (I1)  | `operand`           | tensor                                       | (C1-C5), (C7), (C9-C12) |
-| (I2)  | `source`            | tensor                                       | (C2), (C3)              |
-| (I3)  | `init_value`        | 0-dimensional tensor                         | (C4)                    |
-| (I4)  | `window_dimensions` | 1-dimensional tensor constant of type `si64` | (C1), (C3), (C5), (C6)  |
-| (I5)  | `window_strides`    | 1-dimensional tensor constant of type `si64` | (C3), (C7), (C8)        |
-| (I6)  | `padding`           | 2-dimensional tensor constant of type `si64` | (C3), (C9)              |
-| (I7)  | `select`            | function                                     | (C10)                   |
-| (I8)  | `scatter`           | function                                     | (C11)                   |
+| (I1)  | `operand`           | tensor or quantized tensor                   | (C1-C6), (C8), (C9-C17) |
+| (I2)  | `source`            | tensor or quantized tensor                   | (C2), (C9), (C13-C14)   |
+| (I3)  | `init_value`        | 0-dimensional tensor                         | (C3)                    |
+| (I4)  | `window_dimensions` | 1-dimensional tensor constant of type `si64` | (C1-C2), (C4-C5)        |
+| (I5)  | `window_strides`    | 1-dimensional tensor constant of type `si64` | (C2), (C6), (C7)        |
+| (I6)  | `padding`           | 2-dimensional tensor constant of type `si64` | (C2), (C8)              |
+| (I7)  | `select`            | function                                     | (C10), (C15)            |
+| (I8)  | `scatter`           | function                                     | (C11), (C16)            |
 
 #### Outputs
 
-| Name     | Type   | Constraints |
-|----------|--------|-------------|
-| `result` | tensor | (C12)       |
+| Name     | Type                       | Constraints          |
+|----------|----------------------------|----------------------|
+| `result` | tensor or quantized tensor | (C12-C13), (C16-C17) |
 
 #### Constraints
 
 <!-- markdownlint-disable line-length -->
 * (C1) rank(`operand`) $=$ size(`window_dimensions`).
-* (C2) `operand` and `source` have the same element type.
-* (C3) `shape(source) = (padded_operand_shape == 0 || window_dimensions > padded_operand_shape) ? 0 : floor((padded_operand_shape - window_dimensions) / window_strides) + 1:`
+* (C2) `shape(source) = (padded_operand_shape == 0 || window_dimensions > padded_operand_shape) ? 0 : floor((padded_operand_shape - window_dimensions) / window_strides) + 1:`
   * `padded_operand_shape = padding[:, 0] + shape(operand) + padding[:, 1]`.
-* (C4) element_type(`init_value`) $=$ element_type(`operand`).
-* (C5) size(`window_dimensions`) $=$ rank(`operand`).
-* (C6) `window_dimensions[i]` $\gt 0$ for all i $\in$ [0, size(window_dimensions)).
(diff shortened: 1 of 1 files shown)


loganchien:
I only skimmed through the commit message, but I think only extending the return type of the compute region doesn't work because the input of the compute region can take larger size too.

For example: The schedule can be:

```
[init_val/8] [input/8]  [input/8]
    \        /         /
  [compute/32]        /
        \            /
         [compute/32]
```

sdasgup3:
@loganchien Thanks for the feedback.

Yes, I can see your point. 

Upon further thinking into this: Here is an high level overview of an option in order to accommodate higher (than `input` and `init_value`) bit-width accumulation. I am borrowing the symbol names from the example mentioned above:

1. `input` and `init_value` to have the same storage element type. This is motivated by the fact that the operand types for  `input` and `init_value`  represent the value types and accumulation semantics is expressed in the reduced block.
2. The block arguments of reduce can have same storage type (which is the case when the reducer block is performing non-accumulative operations like min/max) **OR**  different storage types (for reducer block with accumulation).  From the POV of specification of constraints, this means that there will be no constraint between the storage type of block arguments. 
   - In the latter case, either (A) both be equal to `input`'s  storage type or (B) both be equal to accumulation storage type or (C) a mix of storage types of `input`  and accumulator.  For (C), the order is not important, as the op semantics does not require any specific order to fetch the binary tree nodes for computation.  
4. The result storage type of the operation should match the block-return storage type. 
5. If the block_arguments have the same storage_type , then  storage_type of block-result is equal to the storage type of one of the block arguments. Else, storage_type of block-result is equal to the one with large storage bitwidth.
6. The expressed_type, quantization parameters of `input`, `init_value`, `block_arguments`, `block_result` and `result` will be the same. 

Notes:

 - (4) (5) ensures that the op return type can be inferred. 
 - I assume 
 - IMO, we should not impose any restriction of the storage_type of block arguments. Th accumulation bitwidth will be defined by the user.  

Please, let me know your feedback!

wecing:
Thank you @sdasgup3 for the update.

> For (C), the order is not important, as the op semantics does not require any specific order to fetch the binary tree nodes for computation.

Does that imply that for `reduce(+, [1, 2, 3, 4])`, the spec allows `((a + b) + (c + d))` as well? If that's the case then maybe we should require the computation to be performed on the accumulator type.

Also what do you think about explicitly defining the conversion behavior between input/output type and accumulator type, e.g. "converted by zero-extending storage value"?

sdasgup3:
Hi @loganchien 


>> Does that imply that for reduce(+, [1, 2, 3, 4]), the spec allows ((a + b) + (c + d)) as well? If that's the case then maybe we should prohibit the hybrid option C.

Yes, the spec allows that. 

Let say `V` = value type, `A` = accumulation type. There are four possible signatures for a block, which represents computations with specific types.
(1) block(V, V): The computation involves up casting both block arguments to accum type and doing the computation. 
(2) block(A, A): The computation involves just doing the computation (no up casting to accum type needed)
(3) block(A, V) // hybrid: The computation invove up-casting one of the arg with value type to accum type and then doing the computation.
(4) block(V, A) // hybrid: Same as the above


The specific schedule `((a + b) + (c + d)) ` allows computations (1) and (2). 
There could be other schedules  like (((a+b)+c) +d) which involve computation (1) and (3/4)
Also, for schedule ((a+b) + c) + (d +e) involve all the four computations. 

The open question, that I was pondering about, is how a block, with a fixed static type signature, like any one of  1-4, can be used for any of these schedules.  One solution could be use only (2) and update the specification of schedule execution to **always** up cast the binary tree nodes to accum type before  before applying (2).
 
>> Also what do you think about explicitly defining the conversion behavior between input/output type and accumulator type, e.g. "converted by zero-extending storage value"?

May be I am not sure about what you meant by "defining the conversion behavior between input/output type and accumulator type". Can you suggest an example? 

On a related note how about we explicitly zero-extending the `input` and `init_value` tensors to accumulation type before doing the reduction operation. This need minimalist changes to the spec (same type constraints as we currently have). WDYT?

loganchien:
@wecing  wrote:
> defining the conversion behavior between input/output type and accumulator type
>
> Also what do you think about explicitly defining the conversion behavior between input/output type and accumulator type, e.g. "converted by zero-extending storage value"?

@sdasgup3 wrote:
> On a related note how about we explicitly zero-extending the input and init_value tensors to accumulation type before doing the reduction operation

This is another idea that circulates in our team.  Keep the reduce computation with symmetric semantics but allow users to define `input->intermediate` or `intermediate->output` computation:

```
%result = "stablehlo.reduce"(%input, %init_value) ({
    ^input_conversion(
            %input: tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>):
        %input_rescaled = "stablehlo.uniform_quantize"(%input)
            : (tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>)
            -> tensor<!quant.uniform<i32:f32, input_scale:0>>
        "stablehlo.return"(%input_rescaled)
            : (tensor<!quant.uniform<i32:f32, input_scale:0>>) -> ()

    }, {
    ^reduce_computation(
            %lhs: tensor<!quant.uniform<i32:f32, input_scale:0>>,
            %rhs: tensor<!quant.uniform<i32:f32, input_scale:0>>):
        %add = "stablehlo.add"(%lhs, %rhs)
            : (tensor<!quant.uniform<i32:f32, input_scale:0>>,
               tensor<!quant.uniform<i32:f32, input_scale:0>>)
            -> tensor<!quant.uniform<i32:f32, input_scale:0>>
        "stablehlo.return"(%add)
            : (tensor<!quant.uniform<i32:f32, input_scale:0>>) -> ()
    }, {
    ^output_conversion(
            %intermediate_result: tensor<!quant.uniform<i32:f32, input_scale:0>>):
        %output_rescaled = "stablehlo.uniform_quantize"(%intermediate_result)
            : (tensor<!quant.uniform<i32:f32, input_scale:0>>)
            -> tensor<!quant.uniform<ui8:f32, output_scale:output_zp>>
        "stablehlo.return"(%output_rescaled)
            : (tensor<!quant.uniform<ui8:f32, output_scale:output_zp>>) -> ()
    }) {
        dimensions = dense<1> : tensor<1xi64>
    } : (tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>,
         tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>)
    -> tensor<!quant.uniform<ui8:f32, output_scale:output_zp>>
```

sdasgup3:
>> On a related note how about we explicitly zero-extending the input and init_value tensors to accumulation type before doing the reduction operation

Just for the completeness of my thoughts, the following is what i meant: 

```
%input_rescaled = "stablehlo.uniform_quantize"(%input)
            : (tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>)
            -> tensor<!quant.uniform<i32:f32, input_scale:0>>

%init_value_rescaled = "stablehlo.uniform_quantize"(%init_value)
            : (tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>)
            -> tensor<!quant.uniform<i32:f32, input_scale:0>>

%result = "stablehlo.reduce"(%input, %init_value) ({
    ^reduce_computation(
            %lhs: tensor<!quant.uniform<i32:f32, input_scale:0>>,
            %rhs: tensor<!quant.uniform<i32:f32, input_scale:0>>):
        %add = "stablehlo.add"(%lhs, %rhs)
            : (tensor<!quant.uniform<i32:f32, input_scale:0>>,
               tensor<!quant.uniform<i32:f32, input_scale:0>>)
            -> tensor<!quant.uniform<i32:f32, input_scale:0>>
        "stablehlo.return"(%add)
            : (tensor<!quant.uniform<i32:f32, input_scale:0>>) -> ()
    }) {
        dimensions = dense<1> : tensor<1xi64>
    } : (tensor<!quant.uniform<i32:f32, input_scale:0>>,
         tensor<!quant.uniform<i32:f32, input_scale:0>>)
    -> tensor<!quant.uniform<i32:f32, output_scale:0>>

```

With that, spec for reduce can will be pretty much the same (and simple) in terms of type constraints.  Also, I note that  to accommodate multiple input operands in the stablehlo.reduce, we might need to support `stablehlo.uniform_quantize`  with variadic-input support. Do you see any problem with that?


>> This is another idea that circulates in our team. Keep the reduce computation with symmetric semantics but allow users to define input->intermediate or intermediate->output computation:

Thanks @loganchien for the explanatory example. That sounds like a valid option.

A few observations, along with some questions, that I have here:
1. I can understand that `output_conversion` is introduced to achieve symmetry in computation enabling the `reduce_computation` to have a fixed type signature.  IMO, this is at the cost of redundancy in application of   `output_conversion`  after every `reduce_computation`, right? 
2. There could be scenarios when the application `input_conversion` should be skipped. For example, when the `reduce_computation` is not accumulative in nature (like for min/max computations) or for non-quantized scenario. It seems that you meant the `input_conversion` and `output_conversion` blocks to be optional and to be provided only when the the block argument type of `reducer_computation` in different from `%input` (or `%init_value`). From the POV of specification, there should be verification constraints to ensure that.

Let me know if I miss out anything in my observations.

loganchien:
> IMO, this is at the cost of redundancy in application of output_conversion after every reduce_computation, right?

Not every `reduce_computation`.  It is only applied once after all reduce computation are done for a particular output element.

> There could be scenarios when the application input_conversion should be skipped.

Yes.  In that case, it can be optional and the input element is copied as-is.  The verification constraint should be: if the types are different, there must be a conversion block.  If types are the same, it is optional (people can do some preprocessing if they want) but can be skipped for brevity.

sdasgup3:
>> Not every reduce_computation. It is only applied once after all reduce computation are done for a particular output element.

IIUC, for all the input and init_value s contributing to a particular output element, we first use the `input_conversion` to scale all of them up to accum type and then use the `reduce_computation` pairwise on the scaled up elements and at the end  scale down the final result, for that output element. I think you meant this to be one possible implementation in order to avoid redundancy. right?

sdasgup3:
Also looking forward to some feedback on 

>> On a related note how about we explicitly zero-extending the input and init_value tensors to accumulation type before doing the reduction operation

loganchien:
> IIUC, for all the input and init_value s contributing to a particular output element, we first use the input_conversion to scale all of them up to accum type and then use the reduce_computation pairwise on the scaled up elements and at the end scale down the final result, for that output element. I think you meant this to be one possible implementation in order to avoid redundancy. right?

It is not about redundancy.  It is about not rounding/saturating value too early.

````
input_convert(qui8) -> qi32
reduce(qi32, qi32) -> qi32
output_convert(qi32) -> qui8
```

For a 4-element qi8 input, we only want to do:

```
init_value_converted = input_convert(init_value)

input_converted[0] = input_convert(input[0])
input_converted[1] = input_convert(input[1])
input_converted[2] = input_convert(input[2])
input_converted[3] = input_convert(input[3])

accumulated = reduce(reduce(input_converted[0], input_converted[1]), reduce(input_converted[2], input_converted[3]))

output = output_convert(accumulate)
```

We don't want to round off the bits or saturate at qui8 range when we are doing the reduce, so that the inputs get the chance to cancel each other.  **Allowing them to be cancelled** is the motivating reason to do this kind of extension.

For example, if the zero-point adjusted input is `[64, 64, -128, -128]`, we want the final result to be `-128` instead of `saturate_add_i8(saturate_add_i8(64, 64), saturate_add_i8(-128, -128)) = saturate_add_i8(127, -128) = -1`.

> On a related note how about we explicitly zero-extending the input and init_value tensors to accumulation type before doing the reduction operation.

That is the third option being circulated in the internal doc.  But I don't like it because doing tensor-level type conversion means 4x (run-time) memory usage.  All serious compilers/interpreters must do a graph-level rewriting to remove the inefficiency before the they compile or execute the model.

loganchien:
> Also, I note that to accommodate multiple input operands in the stablehlo.reduce, we might need to support stablehlo.uniform_quantize with variadic-input support. Do you see any problem with that?

No.  I don't see problem with multiple-operand reductions.  Multiple-operand reductions only need multiple operand `input_conversion`.  They don't need multiple-operand `stablehlo.uniform_quantize`:

```
%result = "stablehlo.reduce"(%input0, %init_value0, %input1, %init_value1) ({
    ^input_conversion(
            %input0: tensor<!quant.uniform<ui8:f32, input0_scale:input0_zp>>,
            %input1: tensor<!quant.uniform<ui8:f32, input0_scale:input0_zp>>):
        %input0_rescaled = "stablehlo.uniform_quantize"(%input0)
            : (tensor<!quant.uniform<ui8:f32, input0_scale:input0_zp>>)
            -> tensor<!quant.uniform<i32:f32, input0_scale:0>>
        %input1_rescaled = "stablehlo.uniform_quantize"(%input1)
            : (tensor<!quant.uniform<ui8:f32, input1_scale:input1_zp>>)
            -> tensor<!quant.uniform<i32:f32, input1_scale:0>>
        "stablehlo.return"(%input0_rescaled, %input1_rescaled)
            : (tensor<!quant.uniform<i32:f32, input0_scale:0>>,
               tensor<!quant.uniform<i32:f32, input1_scale:0>>) -> ()
    }, {
    ^reduce_computation(
            %lhs: tensor<!quant.uniform<i32:f32, input0_scale:0>>,
            %rhs: tensor<!quant.uniform<i32:f32, input1_scale:0>>):
        // ....
    }, {
    ^output_conversion(
            %intermediate0_result: tensor<!quant.uniform<i32:f32, input0_scale:0>>
            %intermediate1_result: tensor<!quant.uniform<i32:f32, input1_scale:0>>):
       // ....
    }) {
        dimensions = dense<1> : tensor<1xi64>
    } : (tensor<... x !quant.uniform<ui8:f32, input0_scale:input_zp>>,
         tensor<... x !quant.uniform<ui8:f32, input0_scale:input_zp>>,
         tensor<... x !quant.uniform<ui8:f32, input1_scale:input_zp>>,
         tensor<... x !quant.uniform<ui8:f32, input1_scale:input_zp>>)
    -> (tensor<... x !quant.uniform<ui8:f32, output0_scale:output0_zp>>,
        tensor<... x !quant.uniform<ui8:f32, output1_scale:output1_zp>>)
```

wecing:
So, if I understand it correctly, to summarize the discussions so far (please let me know if I missed anything):

(Let `V` = input type, `A` = accumulator type)

1. @loganchien pointed out that the original proposal `reduce_region(V, V) -> A` doesn't work since it implies `A -> V` truncation in the middle of reduction;
2. @wecing suggests since `stablehlo.reduce(+, [1, 2, 3, 4])` allows `(1 + 2) + (3 + 4)`, we may have to use `reduce_region(A, A) -> A`, which brings up the issue of how/where to do the initial `V -> A` conversion;
3. @loganchien brings up the idea of adding `input_conversion` and `output_conversion` regions to `stablehlo.reduce`;
4. @sdasgup3 proposes using separate casting ops outside of `stablehlo.reduce` (before and after it) to do the `qi8 <=> qi32` conversion, so that `stablehlo.reduce` only needs to deal with `qi32`;
5. @sdasgup3 mentioned varargs `stablehlo.uniform_quantize` but @loganchien think it's unnecessary.

There are also some discussions between @loganchien and @sdasgup3 where IIUC @sdasgup3 was asking about `from_accum(reduce(to_accum(input)))` and @loganchien was also saying reduction should happen on accumulator type, i.e. we shouldn't truncate too early.

---

Here are my thoughts:

1. StableHLO needs to support the `reduce_region(A, A) -> A` use case (but does not require it, so people could still write `reduce_region(V, V) -> V` if they really want to). The question is whether to do `V <=> A` conversion inside (via `input_conversion` and `output_conversion` regions) or outside (via casting ops) of `stablehlo.reduce`. I think either way works, but personally I think doing that inside `stablehlo.reduce` is slightly easier for my case.
2. We could add restrictions on the `V` and `A` type, e.g. we could require them to have the same quant params (scale and zp), and only (optionally) differ in the storage type. In this way we could drop the requirement for `input_conversion` and `output_conversion` regions and specify that `stablehlo.reduce` will zero-extend / truncate the values for us. So I am envisioning something like that:

```
%result = "stablehlo.reduce"(%input, %init_value) ({
    ^reduce_computation(
            %lhs: tensor<!quant.uniform<i32:f32, input_scale:0>>,
            %rhs: tensor<!quant.uniform<i32:f32, input_scale:0>>):
        %add = "stablehlo.add"(%lhs, %rhs)
            : (tensor<!quant.uniform<i32:f32, input_scale:0>>,
               tensor<!quant.uniform<i32:f32, input_scale:0>>)
            -> tensor<!quant.uniform<i32:f32, input_scale:0>>
        "stablehlo.return"(%add)
            : (tensor<!quant.uniform<i32:f32, input_scale:0>>) -> ()
    }) {
        dimensions = dense<1> : tensor<1xi64>
    } : (tensor<!quant.uniform<i8:f32, input_scale:0>>,
         tensor<!quant.uniform<i8:f32, input_scale:0>>)
    -> tensor<!quant.uniform<i8:f32, output_scale:0>>
```

What do you think?

sdasgup3:
Hi @wecing 
Thanks for summarizing the proposals. 

Regarding your proposal:
It proposes to use a reducer block which can work only on the accumulator type. One thing that stuck me initailly that  how this reducer block be used to compute something like `(+ 1 2)`, which expects the reducer block to take as argument two value of type   `tensor<!quant.uniform<i8:f32, input_scale:0>>`.  Now it is clarified that what you meant in that the leaf node of the schedule binary tree has to be implicitly converted.  I believe @loganchien proposal makes that conversion step explicit by using `input_conversion` and `output_conversion` blocks.  Also, I note that, per @loganchien proposal, for reduce op with quantized types 
  - The `input_conversion` will be applied to convert just the leaf nodes of the binary tree.
  - The  `output_conversion` is applied once after all reduce computation are done for a particular output element.



@jingpu

Re:
Logan>> That is the third option being circulated in the internal doc. But I don't like it because doing tensor-level type conversion means 4x (run-time) memory usage. All serious compilers/interpreters must do a graph-level rewriting to remove the inefficiency before the they compile or execute the model.

Just for context, @loganchien is referring to [A](https://github.com/openxla/stablehlo/pull/1538#issuecomment-1565121567)  where the proposal is about using separate  quantize and dequantize ops before and after  stablehlo.reduce to do the `qi8 <=> qi32` conversion, so that stablehlo.reduce only needs to deal with `qi32`. 

Can you please elaborate on the problem @loganchien is referring to? 

If we have one single op, as proposed in [B](https://github.com/openxla/stablehlo/pull/1538#issuecomment-1565242735),  then I believe it can be readily mapped an efficient reduce kernel. For (A), pattern matching could be an additional concern and I am trying to understand how much is that an overhead.

rewu93 (commented):
One dumb question. How does the user define the accumulator type `A` if it is inside the `reduce` block?

sdasgup3:
>> One dumb question. How does the user define the accumulator type A if it is inside the reduce block?

Per @loganchien proposal, the input conversion block with specify the accumulation type.
 Per my proposal, the `stablehlo.uniform_quantize` will specify the accumulation type.

wecing:
Hey @sdasgup3 -- I synced with @jingpu offline, and unfortunately neither of us know what @loganchien's concern on your proposal A was. But, as I mentioned above in the summary comment,

> There were also some discussions between @loganchien and @sdasgup3 where IIUC @sdasgup3 was asking about `from_accum(reduce(to_accum(input)))` and @loganchien was also saying reduction should happen on accumulator type, i.e. we shouldn't truncate too early.

My guess is that:

1. It was partially a mis-communication issue and @loganchien was worried about potential early truncation, but your proposal doesn't really have this issue (FYI: I think @loganchien will be back next week; hopefully we could get more clarification by then);
2. As you quoted, `All serious compilers/interpreters must do a graph-level rewriting to remove the inefficiency before the they compile or execute the model`, the concern was that this would make certain things inefficient, e.g. interpreters, and some debug tools, would have to do the `q8->q32` quantization, which may be impractical for very large models, e.g. LLMs.

loganchien:
Let me elaborate my thinking on two different ways to express "accumulator" type in the reduction:

The first one (my favorite one for now) is to introduce `input_conversion` and `output_conversion` as described in the [earlier comment](https://github.com/openxla/stablehlo/pull/1538#issuecomment-1565072256):

```
%result = "stablehlo.reduce"(%input, %init_value) ({
    ^input_conversion(
            %input: tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>):
        %input_rescaled = "stablehlo.uniform_quantize"(%input)
            : (tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>)
            -> tensor<!quant.uniform<i32:f32, input_scale:0>>
        "stablehlo.return"(%input_rescaled)
            : (tensor<!quant.uniform<i32:f32, input_scale:0>>) -> ()

    }, {
    ^reduce_computation(
            %lhs: tensor<!quant.uniform<i32:f32, input_scale:0>>,
            %rhs: tensor<!quant.uniform<i32:f32, input_scale:0>>):
        %add = "stablehlo.add"(%lhs, %rhs)
            : (tensor<!quant.uniform<i32:f32, input_scale:0>>,
               tensor<!quant.uniform<i32:f32, input_scale:0>>)
            -> tensor<!quant.uniform<i32:f32, input_scale:0>>
        "stablehlo.return"(%add)
            : (tensor<!quant.uniform<i32:f32, input_scale:0>>) -> ()
    }, {
    ^output_conversion(
            %intermediate_result: tensor<!quant.uniform<i32:f32, input_scale:0>>):
        %output_rescaled = "stablehlo.uniform_quantize"(%intermediate_result)
            : (tensor<!quant.uniform<i32:f32, input_scale:0>>)
            -> tensor<!quant.uniform<ui8:f32, output_scale:output_zp>>
        "stablehlo.return"(%output_rescaled)
            : (tensor<!quant.uniform<ui8:f32, output_scale:output_zp>>) -> ()
    }) {
        dimensions = dense<1> : tensor<1xi64>
    } : (tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>,
         tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>)
    -> tensor<!quant.uniform<ui8:f32, output_scale:output_zp>>
```

Its execution looks like:

```
[In0:i8]           [In1:i8]           [In2:i8]           [In3:i8]
|                  |                  |                  |
v                  v                  v                  v
[InputConversion]  [InputConversion]  [InputConversion]  [InputConversion]
(i32)              (i32)              (i32)              (i32)
|                  |                  |                  |
v                  v                  v                  v                  \
[ReduceComputation_]                  [ReduceComputation_]                  |
(i32)                                 (i32)                                 |
|                                     |                                     | Reduce schedule can be arbitrary
v                                     v                                     |
[ReduceComputation____________________]                                     /
(i32)
|
v
[OutputConversion]
(i8)
```

The second one is proposed by @sdasgup3, which simply suggest adding `stablehlo.convert`/`stablehlo.uniform_quantize` ops before and after the reduce op, which operates on the "accumulator" type:

```
%widen = "stablehlo.convert"(%input)
    : (tensor<... x !quant.uniform<ui8:f32, ...>>) -> tensor<... x !quant.uniform<i32:f32, ...>>
%reduce = "stablehlo.reduce"(%widen) {
    ^reduce_computation(%lhs: !quant.uniform<i32:f32, ...>, %rhs: !qunat.uniform<i32:f32, ...>):
        // reduce_computation_block
    }
    : (tensor<... x !quant.uniform<i32:f32, ...>>) -> tensor<... x !quant.uniform<i32:f32, ...>>
%narrowed = "stablehlo.convert"(%reduce)
    : (tensor<... x !quant.uniform<i32:f32, ...>>) -> tensor<... x !quant.uniform<ui8:f32, ...>>
```

Both options have pros and cons:

The first option enables programmers to program at (almost) baremetal (*argument1).  If the HW can support reduction computation in narrow type (e.g. in the SIMD instruction set, we typically do widening/compute/narrowing within the kernel to save the memory bandwidth), the programmer can explicitly request for that.  The second option can't express the intent.  It will rely on the compiler to do pattern matching and optimization to avoid the 4x memory bandwidth increase (*argument2).  The compiler (or even interpreter) must pattern match 3 operations and map them into some internal representation (that is not expressible in StableHLO) before their compilation or execution (*argument3).

In the second option, there is also a risk that the consecutive `stablehlo.convert` operations might be folded or converted to something else.  And then, the best-effort approach of the backend may fail.  For example, in the `bf16` reduction:

```
%widen0 = "stablehlo.convert"(%input0)
    : (tensor<... x bf16>) -> tensor<... x f32>
%reduce0 = "stablehlo.reduce"(%widen0) {
    ^reduce_computation(%lhs: f32, %rhs: f32):
        // reduce_computation_block
    }
    : (tensor<... x f32>) -> tensor<... x f32>
%narrowed0 = "stablehlo.convert"(%reduce0)        //
    : (tensor<... x f32>) -> tensor<... x bf16>   //
                                                  // potential target of so-called target-independent optimization
%widen1 = "stablehlo.convert"(%narrowed0)         //
    : (tensor<... x bf16>) -> tensor<... x f32>   //
%reduce1 = "stablehlo.reduce"(%widen1) {
    ^reduce_computation(%lhs: f32, %rhs: f32):
        // reduce_computation_block
    }
    : (tensor<... x f32>) -> tensor<... x f32>
%narrowed1 = "stablehlo.convert"(%reduce1)
    : (tensor<... x f32>) -> tensor<... x bf16>
```

Some optimizations or folders may *transform* the code above into:

```
%widen0 = "stablehlo.convert"(%input0)
    : (tensor<... x bf16>) -> tensor<... x f32>
%reduce0 = "stablehlo.reduce"(%widen0) {
    ^reduce_computation(%lhs: f32, %rhs: f32):
        // reduce_computation_block
    }
    : (tensor<... x f32>) -> tensor<... x f32>

%reduce1 = "stablehlo.reduce"(%reduce0) {
    ^reduce_computation(%lhs: f32, %rhs: f32):
        // reduce_computation_block
    }
    : (tensor<... x f32>) -> tensor<... x f32>
%narrowed1 = "stablehlo.convert"(%reduce1)
    : (tensor<... x f32>) -> tensor<... x bf16>
```

and the best-effort approach in the backend can't kick in and the compiler must allocate an `f32` buffer for `%reduce0` (*argument4).

The second option still have some pros though.  It doesn't require significant changes to existing spec (ReduceOp in particular).  However, it puts the burden to the backend writer.

----

From my own perspective, since my backend must do semantic graph rewrite anyway, the differences between two options are only about (1) the ease of the pattern matching implementation and (2) keep early transformations away from breaking the best-effort patterns.  However, I believe the extra expressiveness might find some use cases in other more general architectures (e.g. CPU/GPU).

----

Regarding the point on extending `stablehlo.uniform_quantize` for variadic operands, I think it is not necessary even if we choose the first option.  For example, if a `stablohlo.reduce` takes multiple input operands, we can expand the input/output conversion into multiple `stablehlo.uniform_quantize` operations in the conversion block if needed:

```
%result = "stablehlo.reduce"(%input0, %init_value0, %input1, %init_value1) ({
    ^input_conversion(
            %input0: tensor<!quant.uniform<ui8:f32, input0_scale:input0_zp>>,
            %input1: tensor<!quant.uniform<ui8:f32, input1_scale:input1_zp>>):
        %input0_rescaled = "stablehlo.uniform_quantize"(%input0)
            : (tensor<!quant.uniform<ui8:f32, input0_scale:input0_zp>>)
            -> tensor<!quant.uniform<i32:f32, input0_scale:0>>
        %input1_rescaled = "stablehlo.uniform_quantize"(%input1)
            : (tensor<!quant.uniform<ui8:f32, input1_scale:input1_zp>>)
            -> tensor<!quant.uniform<i32:f32, input1_scale:0>>
        "stablehlo.return"(%input0_rescaled, %input1_rescaled)
            : (tensor<!quant.uniform<i32:f32, input0_scale:0>>,
               tensor<!quant.uniform<i32:f32, input1_scale:0>>) -> ()
    }, {
    ^reduce_computation(
            %lhs0: tensor<!quant.uniform<i32:f32, input0_scale:0>>,
            %rhs0: tensor<!quant.uniform<i32:f32, input0_scale:0>>,
            %lhs1: tensor<!quant.uniform<i32:f32, input1_scale:0>>,
            %rhs1: tensor<!quant.uniform<i32:f32, input1_scale:0>>):
        // ... skipped ...
        "stablehlo.return"(..., ...)
            : (tensor<!quant.uniform<i32:f32, input0_scale:0>>,
               tensor<!quant.uniform<i32:f32, input1_scale:0>>) -> ()
    }, {
    ^output_conversion(
            %intermediate0_result: tensor<!quant.uniform<i32:f32, input0_scale:0>>
            %intermediate1_result: tensor<!quant.uniform<i32:f32, input1_scale:0>>):
        %output0_rescaled = "stablehlo.uniform_quantize"(%input0)
            : (tensor<!quant.uniform<i32:f32, input0_scale:0>>)
            -> tensor<!quant.uniform<ui8:f32, output0_scale:output0_zp>>
        %output1_rescaled = "stablehlo.uniform_quantize"(%input1)
            : (tensor<!quant.uniform<i32:f32, input1_scale:0>>)
            -> tensor<!quant.uniform<ui8:f32, output1_scale:output1_zp>>
        "stablehlo.return"(%output0_rescaled, %output1_rescaled)
            : (tensor<!quant.uniform<ui8:f32, output0_scale:output0_zp>>,
               tensor<!quant.uniform<ui8:f32, output1_scale:output1_zp>>) -> ()
    }) {
        dimensions = dense<1> : tensor<1xi64>
    } : (tensor<!quant.uniform<ui8:f32, input0_scale:input0_zp>>,
         tensor<!quant.uniform<ui8:f32, input0_scale:input0_zp>>,
         tensor<!quant.uniform<ui8:f32, input0_scale:input1_zp>>,
         tensor<!quant.uniform<ui8:f32, input0_scale:input1_zp>>)
    -> (tensor<!quant.uniform<ui8:f32, output0_scale:output0_zp>>,
        tensor<!quant.uniform<ui8:f32, output1_scale:output1_zp>>)
```

sdasgup3:
@loganchien Thanks for the elaborate explanation and I agree with your point. 
As the proposal needs some structural changes in the op (adding additional blocks), it should go an RFC process. 
Let me document that which we all can further iterate over.

burmako:
Superseded by https://github.com/openxla/stablehlo/pull/1664 - which has summarized the discussion here into an RFC, which is under discussion and will result in spec updates once approved.

PR #1541, opened 2023-05-26
Avoid out of bound reads in the spec of GatherOp

ghpvnist:
In a recent [discussion](https://github.com/openxla/stablehlo/pull/1058#discussion_r1203386357) on the GatherOp implementation, we pointed out that the semantics did not fully address what to do with out of bounds indexes. To make sure the bounds are always correct, clamping on the `start_index` is needed. This reflects the change needed in the spec and the reference implementation.


docs/spec.md
```diff
@@ -2823,6 +2825,7 @@ More formally, `result[result_index] = operand[operand_index]` where:
 * `operand_index` = `add(full_start_index, full_offset_index)`.
   If `operand_index` is out of bounds for `operand`, then the behavior is
   implementation-defined.
```
burmako, line 2827: "If operand_index is out of bounds for operand, then the behavior is implementation-defined". Does this still apply?

docs/spec.md
```diff
@@ -2813,8 +2814,9 @@ More formally, `result[result_index] = operand[operand_index]` where:
     `index_vector_dim` < `rank(start_indices)`.
   * `[start_indices[batch_index]]` otherwise.
 * For `d_operand` in `axes(operand)`,
-  * `full_start_index[d_operand]` = `start_index[d_start]` if
-    `d_operand = start_index_map[d_start]`.
+  * `full_start_index[d_operand]` =
+    `clamp(start_index[d_start], 0, dim(operand, d_operand) - slice_sizes[d_operand])`
+    if `d_operand = start_index_map[d_start]`.
```
burmako, line 2819: Why do we need to clamp specifically here instead of e.g. at the very last step?

burmako, line 2819: Also how about we move clamping to a separate bullet, i.e. 

```
* `start_index` =
  * `start_indices[bi0, ..., :, ..., biN]` where `bi` are individual elements in
    `batch_index` and `:` is inserted at the `index_vector_dim` index, if
    `index_vector_dim` < `rank(start_indices)`.
  * `[start_indices[batch_index]]` otherwise.
* `clamped_start_index = ...`.
* For `d_operand` in `axes(operand)`,
  * `full_start_index[d_operand]` = `clamped_start_index[d_start]` if
    `d_operand = start_index_map[d_start]`.
```

This will make the intent more explicit and will also avoid the need for markdownlint-disable line-length.

docs/spec.md
burmako, line 2827: Also, what should happen to ScatterOp?

burmako (changes requested):

docs/spec.md
burmako: Oops, I see now that clamping cannot happen before applying start_index_map - it should happen afterwards. I still think that modularization could be useful though, along the lines of `clamped_full_start_index = ...`.

docs/spec.md
```diff
@@ -2805,6 +2805,7 @@ indices and explains in detail which `operand` indices they correspond to.
```
burmako, line 2805: Also, what do we do with the image? Do we need to update it as well?

burmako (changes requested):

docs/spec.md
burmako: Okay nevermind - I now think that moving clamping into a separate bullet isn't worth it overall, because it would make updating the associated image considerably harder and the benefits are pretty minor all things considered. Resolving this thread.

ghpvnist: We still want to keep the information about how much to offset for `fullOffsetIndex`. By clamping after the addition, we lose this information. For example, if `start_index` had -100, and offset 2, we would get `clamp(min=0, -100+2, max) = min` whereas we want (min+2) as the index.

burmako: "whereas we want (min+2) as the index". To dig a bit deeper, can you share a reference to documentation or code that depends on / implements this behavior?

docs/spec.md
ghpvnist: HLO skips computing out of bounds indices: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/hlo/evaluator/hlo_evaluator.cc#L3020-L3025

burmako: Alright, it looks like we may want to change the spec for ScatterOp as well, following the similar process as for GatherOp - first submit an interpreter PR as is, and then follow up with a delta.

docs/spec.md
ghpvnist: Perhaps this code snippet is what you're looking for?https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/hlo/evaluator/hlo_evaluator.cc#L2635-L2644

Operation Semantics do not explicitly state clamping, but the comments in the code state the clamping before addition.

docs/spec.md
ghpvnist: Thanks, this should be removed.

docs/spec.md
ghpvnist: The svg export did its magic, so I couldn't quite make the changes as minimal as hoped.

docs/spec.md
burmako: A link to hlo_evaluator is useful, but since this semantics is not part of https://www.tensorflow.org/xla/operation_semantics#gather, we'd need a bit more. Otherwise, it's hard to say whether the clamping behavior is perhaps implementation-defined behavior. 

Can we have some additional paper trail, e.g. links to some frameworks making use of this semantics (e.g. JAX) or testimonials from representatives of these frameworks?

ghpvnist: Here is the link to JAX docs on gather: https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.gather.html
Note for JAX, the mode can be one of three values `{CLIP, FILL_OR_DROP, PROMISE_IN_BOUNDS}`. 
https://jax.readthedocs.io/en/latest/jax.lax.html#jax.lax.GatherScatterMode

PROMISE_IN_BOUNDS:
The user promises that indices are in bounds. No additional checking will be performed. In practice, with the current XLA implementation this means that, out-of-bounds gathers will be clamped but out-of-bounds scatters will be discarded. Gradients will not be correct if indices are out-of-bounds.

docs/images/spec/gather.svg
burmako, line 1: Can we also change one block of values in start_indices (e.g. `|1 0|`) to be out of bounds to demonstrate clamping?

burmako, line 1: How about "then clamp to shape(operand) - slice_sizes = [2, 2, 0]"?

docs/spec.md
burmako: I don't think that "indices are in bounds" and "out-of-bounds gathers will be clamped" indicate specifically that clamping should happen at the `full_start_index` level and not at the `operand_index` level?

burmako (changes requested):

docs/images/spec/gather.svg
burmako: By the way, this is the same example that is used in the spec, so if we update the image, let's also update the code example in the spec (and in the interpreter).

burmako: @ghpvnist I have moved the StableHLO Spec Images deck from Google's internal drive to OpenXLA's public drive: https://docs.google.com/presentation/d/1vr6x3VtV77NPIwR26uiHEJQb407QtRXMUfIi5CSs8aA/edit#slide=id.g225b8f7527f_0_0. Please apply the updates there.

docs/spec.md
ghpvnist: Here is another [reference](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/elemental_ir_emitter.cc#L2587-L2602) where this behavior is similarly replicated.

burmako (approved):
LGTM but please also get Peter's approval

ghpvnist:
Hi @hawkinsp, the PR is now ready for review, and I've assigned you as a reviewer. Feel free to skip unrelated changes and just review [docs/spec.md](https://github.com/openxla/stablehlo/pull/1541/files#diff-d7d183da3129291ef1210410b2898ac6a65a4e4de8ede9775601173545570f57) and [stablehlo/reference/Ops.cpp](https://github.com/openxla/stablehlo/pull/1541/files#diff-55cd0d2c93314baca4e69a6ee8c83cb016040a45b684d995bb58ef262cea8018).

docs/images/spec/gather.svg
burmako, line 1: I've proposed a minor update to the diagram to explicitly call out that one of the full start indices is clamped: https://docs.google.com/presentation/d/1vr6x3VtV77NPIwR26uiHEJQb407QtRXMUfIi5CSs8aA/edit#slide=id.g24e7ed57a8e_0_28.

burmako (changes requested):

hawkinsp (approved):
I think this is correct. But I'm hoping you've validated the spec against what XLA does.

burmako (approved):

PR #1550, opened 2023-05-28
Remove an implementation-defined test for GatherOp

burmako:
In the removed test, one of the start_indices is 10 which is out of bounds for the operand, and that's currently specced as implementation-defined. Per interpreter checklist, we should remove this test.

### DIFF
--- a/stablehlo/tests/interpret_gather.mlir
+++ b/stablehlo/tests/interpret_gather.mlir
@@ -23,22 +23,3 @@ func.func @gather_op_test() {
                                          [[17, 18], [19, 20]]]]> : tensor<2x3x2x2xi64>
   func.return
 }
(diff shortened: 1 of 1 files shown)


burmako:
It looks like this test is causing spurious CI failures: https://github.com/openxla/stablehlo/actions/runs/5113709444. Let's prioritize reviewing & merging.

ghpvnist (approved):
The change lgtm, but we still want to have this test for `index_vector_dim == rank(start_indices)` case. How about changing `start_indices` from `[0, 10]` to `[0, 1]`?

burmako:
As discussed, merging this now to unblock CI, and I’ll address feedback once I’m back in office.

Issue #425, opened 2022-11-01
Decide on Func/Call/Return ops


burmako:
Open questions:
  1. Should we continue using `func.func`, `func.call` and `func.return`? (Probably not).
  2. Should we use `ml_program`? (Probably yes, but this will need an RFC + `ml_program.call` will need to be introduced).
  3. Or maybe we should add `stablehlo.func` and `stablehlo.call`? (Probably not, but this will be one of the alternatives in the RFC).
  4. Also, what does this mean for `stablehlo.return`? (Probably should merge it into whatever `return` op we decide to use).

burmako:
ModuleOp/FuncOp should provide a landing pad for a bunch of metadata from [HloModuleProto](https://github.com/tensorflow/tensorflow/blob/7f273c2516ccbbd3544145b440432e5a30cb9038/tensorflow/compiler/xla/service/hlo.proto#L456) that has been in our blind spot until recently (MHLO uses `builtin.module` and `func.func` which unlike MHLO ops don't carry HLO-specific metadata except via unregistered attributes, so I didn't realize how much we're missing until recently).

```Proto
  // Describes alias information between inputs and outputs.
  HloInputOutputAliasProto input_output_alias = 8;

  DynamicParameterBindingProto dynamic_parameter_binding = 9;

  repeated CrossProgramPrefetch cross_program_prefetches = 10;

  // True if the module contains dynamic computation.
  bool is_dynamic = 11;

  xla.OpSharding spmd_output_sharding = 12;

  repeated xla.OpSharding spmd_parameters_shardings = 14;

  // Uses AutoSharding pass or not.
  bool use_auto_spmd_partitioning = 16;

  // The type of optimization profile in use for module-level optimizations.
  enum ProfileType {
    INVALID = 0;
    FLAG = 1;
    FUSION = 2;
    LAYOUT = 3;
    DOT = 4;
  }

  // Information about the optimization profile that this module contains.
  message ProfileInfo {
    // The optimization profiles that this module contains.
    ProfileType profile_type = 1;
    // Speedup of tuned config compared to default config.
    double relative_speedup = 2;
    // The source of the optimization profile that this module contains.
    xla.ProfileSource profile_source = 3;
    // The compilation event that triggered the use of the profile.
    xla.CompilationEvent compilation_event = 4;
  }

  // Profile information for the HLO module.
  repeated ProfileInfo profile_info = 13;

  // DeviceAssignment object information.
[...]

PR #1555, opened 2023-05-29
Use func.func and func.return in the spec

burmako:
The stablehlo.func and stablehlo.return syntax in the opening example of the spec was aspirational, and at the time we expected that we'll soon adopt it.

A lot of the time has passed, and we still haven't gotten around to doing this, while we've been getting regular questions about this. Let's reflect reality in the example for now - we can always change it back once we address #425.


burmako:
Overriding CI since it's run into a spurious failure which is addressed elsewhere (#1550).

PR #1488, opened 2023-05-16
Add interpreter for ScatterOp

ghpvnist:
We have the following constraints in the spec:

```
(I1) `inputs` variadic number of tensors.
(I2) `scatter_indices` tensor of integer type.
(I3) `updates` variadic number of tensors.
(I4) `update_window_dims` 1-dimensional tensor constant of type `si64`.
(I5) `inserted_window_dims` 1-dimensional tensor constant of type `si64`.
(I6) `scatter_dims_to_operand_dims` 1-dimensional tensor constant of type `si64`.
(I7) `index_vector_dim` constant of type `si64`.
(I8) `indices_are_sorted` constant of type `i1`.
(I9) `unique_indices` constant of type `i1`.
(I10) `update_computation` function.
(C1) All `inputs` have the same shape.
(C2) rank(`inputs`[0]) = size(`update_window_dims`) +
     size(`inserted_window_dims`).
(C3) All `updates` have the same shape.
(C4) `shape(updates[0])` =
      `combine(update_scatter_dim_sizes, update_window_dim_sizes)` where:
* `update_scatter_dim_sizes` = `shape(scatter_indices)` except that
  the dimension size of `scatter_indices` corresponding to
  `index_vector_dim` is not included.
* `update_window_dim_sizes` <= `shape(inputs[0])` except that
  the dimension sizes in `inputs[0]` corresponding to `inserted_window_dims`
  are not included.
* `combine` puts `update_scatter_dim_sizes` at axes corresponding to
 `update_scatter_dims` and `update_window_dim_sizes` at axes corresponding
 to `update_window_dims`.
(C5) N = size(`inputs`) = size(`updates`) and N >= 1.
(C6) `element_type(updates[k]) = element_type(inputs[k])` for all k $\in$
     [0, N).
(C7) All dimensions in `update_window_dims` are unique and sorted.
(C8) For all i in [0, size(`update_window_dims`)), 0 <=
`update_window_dims`[i] < rank(`updates`[0]).
(C9) All dimensions in `inserted_window_dims` are unique and sorted.
(C10) For all i in [0, size(`inserted_window_dims`)), 0 <=
`inserted_window_dims`[i] < rank(`inputs`[0]).
(C11) size(`scatter_dims_to_operand_dims`) =
     `index_vector_dim` < rank(`scatter_indices`) ?
     dim(`scatter_indices`, `index_vector_dim`) : 1.
(C12) All dimensions in `scatter_dims_to_operand_dims` are unique.
[...]

PR #1561, opened 2023-05-31
Avoid out of bound writes in the spec of ScatterOp

ghpvnist:
As a follow up to https://github.com/openxla/stablehlo/pull/1488#discussion_r1207464016, this PR updates the spec to clarify the behavior of OOB result indices and to match HLO semantics and the current reference implementation.


docs/spec.md
```diff
@@ -4667,8 +4667,8 @@ Given that, `results = exec(schedule, inputs)`, where:
     update_computation(results[:][result_index], updates[:][update_index])`.
   * `updated_results` is a copy of `results` with `results[:][result_index]`
     set to `updated_values[:]`.
-  * If `result_index` is out of bounds for `shape(results[:])`, the behavior
-    is implementation-defined.
+  * If `result_index` is out of bounds for `shape(results[:])`, the update is
+    discarded.
```
burmako, line 4671: How about the following formulation which defines more precisely what it means to discard an update:

```
* `exec([update_index, ...], results) = exec([...], updated_results)` where:
  * If `result_index` is in bounds for `shape(results[:])`
    * `updated_values =
      update_computation(results[:][result_index], updates[:][update_index])`.
    * `updated_results` is a copy of `results` with `results[:][result_index]`
      set to `updated_values[:]`.
  * Otherwise
    * `updated_results = results`.
```

burmako, line 4671: Same question as in #1541: can we leave a paper trail about why we're making this change?

burmako (changes requested):

docs/spec.md
ghpvnist: Here is the [reference](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/scatter_expander.cc#L364-L368) for this semantics.

docs/spec.md
```diff
@@ -4663,12 +4663,13 @@ Given that, `results = exec(schedule, inputs)`, where:
...
+  * If `result_index` is in bounds for `shape(results[:])`
+    * `updated_values =
+      update_computation(results[:][result_index], updates[:][update_index])`.
+    * `updated_results` is a copy of `results` with `results[:][result_index]`
+      set to `updated_values[:]`.
+  * Otherwise
+    * `updated_results = results`.
 * `exec([], results) = results`.
 
 If `indices_are_sorted` is `true` then the implementation can assume that
```
burmako, line 4675: Let's also update the image similarly to #1541? We could similarly change `0 | 2` to `0 | 9`, but unlike with GatherOp, we'd need to recolor one of the green rectangles into white, and I would also drop the green highlighting from the corresponding index: https://docs.google.com/presentation/d/1vr6x3VtV77NPIwR26uiHEJQb407QtRXMUfIi5CSs8aA/edit#slide=id.g24e7ed57a8e_0_5.

burmako: I'd also propose that we update `interpret_scatter.mlir` accordingly and drop the second test, similarly to what's done for GatherOp in #1541.

docs/spec.md
burmako: 1) Why is it that unlike for GatherOp, the write is not clamped but skipped?

2) Also, it's not evident to me that the code in scatter_expander.cc is equivalent to what we have in this PR. E.g. in that code, I see DynamicSlice and DynamicUpdateSlice ops and comments reference the adjusted_start_indices behavior, but that is not at all reflected in the updated spec or the interpreter. Could you elaborate?

burmako (changes requested):

docs/spec.md
ghpvnist: Gather is clamped because some backends that doesn't implement Gather use DynamicSlice, and OOB indices in DynamicSlice is clamped to provide a uniform behavior across different hardwares.

OOB indices in Scatter on the other hand is skipped because nobody wanted the clamping behavior on Scatter.

burmako (approved):

PR #1566, opened 2023-06-01
Specification for quantized element-wise operations

sdasgup3:
### Summary

The PR proposes the specification for quantized element wise operations (total 47).

### Details

Overall, we propose treating quantized elements as floating-point 
elements, and therefore ops on quantized tensors as ops on floating-point 
tensors, along the lines of dequantize -> float computation -> quantize. This 
principle works for most elementwise ops, although there are some exceptions 
discussed below.

Furthermore, the proposal is to only support per-tensor quantization for
elementwise ops. We haven't yet come across use cases for per-axis quantization
for these ops, so let's start small. The story for per-axis quantization will
be worked out in #1574.

Finally, we propose to not support hybrid quantization (i.e. situations when
some inputs/outputs are quantized and some are not) for now. The story for this
will be worked out in #1575.

- **Ops that support tensors of floating-point types (33)**: These ops support
quantized tensors, with semantics following dequantize -> float computation -> 
quantize. We are using the `dequantize_op_quantize` function to express it:
  - Binary (10): `add, atan2, compare, divide, maximum, minimum, multiply, 
    power, remainder, subtract`.
  - Unary (20): `abs, cbrt, ceil, cosine, exponential, exponential_minus_one, 
    floor, is_finite, log, logistic, log_plus_one, negate, reduce_precision, 
    round_nearest_afz, round_nearest_even, rsqrt, sign, sine, sqrt, tanh`.
  - Ternary (2): `clamp, select`.
  - Other (1): `map`.
  
- **Ops that don't support tensors of floating-point types (9)**: These ops
(`and, count_leading_zero, not, or, popcnt, shift_left, shift_right_arithmetic, 
shift_right_logical, xor`) don't support quantized tensors. If there is a need 
to perform computations on the underlying integer representation of these 
tensors, they can be bitcast_convert'ed to integers.

- **Ops that involve complex types (3)**: These ops (`complex`, `imag`, `real`) don't support quantized
tensors because quantization doesn't compose with complex types at the moment.

- **Conversion ops (2)**: 
  - `convert`: A convert from a quantized type to any type can be realized using 
    `stablehlo.uniform_dequantize` followed by `stabhle.convert` to convert the 
    dequantized floating-point type to type of choice. Similarly, a convert from 
    any type to quantized type can be realized using `stablehlo.convert` to 
    floating-point type followed by `stablehlo.uniform_quantize`. It's not
    necessarily great that we have 3 ops to represent something that could
    theoretically be represented by 1 op, and we're planning to explore 
    a potential simplification in #1576.
  - `bitcast_convert`: Works with low-level representations, so it treats
    quantized elements as integer elements.


docs/spec.md
```diff
@@ -541,6 +541,55 @@ bytes is implementation-defined. String literals have type `string`.
...
+
+```python
+##  operand: tensor with quantized type.
+##  unary_computaion: unary function of type: (floating-point) ->
+##      floating-point.
+##  out_storage_min, out_storage_max, out_scale, and out_zero_point: integer
+##      constants
+def unary_computation_with_quantized_types(operand, unary_computaion,
+    out_storage_min, out_storage_max, out_scale, out_zero_point):
+  float_result = unary_computaion((operand - zero_point(operand)) *
```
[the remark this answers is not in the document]
sdasgup3: done

docs/spec.md
```diff
@@ -942,24 +1000,26 @@ Performs element-wise AND of two tensors `lhs` and `rhs` and produces a `result`
...
-| `result` | tensor of boolean or integer type | (C1)        |
+| Name     | Type                                         | Constraints |
+|----------|----------------------------------------------|-------------|
+| `result` | tensor of boolean, integer or quantized type | (C1)        |
 
 #### Constraints
 
 * (C1) `lhs`, `rhs` and `result` have the same type.
+* If the operation uses quantized tensors:
+  * (C2) `quantization_dimension(lhs)` is empty.
```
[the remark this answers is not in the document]
sdasgup3: In this case, the constraint (C1) captures that the entire type of `lhs`, `rhs` and `result` are the same, which includes the quantization parameter.  Please let me know if it makes sense.

docs/spec.md
```diff
@@ -3166,23 +3337,32 @@ along the `iota_dimension` dimension. More formally,
 
 Performs element-wise check whether the value in `x` is finite (i.e. is neither
 +Inf, -Inf, nor NaN) and produces a `y` tensor. Implements the `isFinite`
-operation from the IEEE-754 specification.
+operation from the IEEE-754 specification. For quantized types, performs
+`unary_computation_with_quantized_types(x, lambda x: isFinite(x),
```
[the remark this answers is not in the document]
sdasgup3: Thanks @dominicsymes  for pointing this out!
I agree with you. IMO, is_finite should always be `true`.  PTAL at the revised version and let me know what you think.

docs/spec.md
```diff
@@ -4027,26 +4306,36 @@ More formally:
   range provided by `exponent_bits`, the intermediate result overflows to
   infinity using the original sign or underflows to zero using the
   original sign.
+* For quantized types, the above three steps are performed using the original
+  floating-point value as `operand - zero_point(operand)) * scale(operand)` to
+  produce an intermediate result `float_result`. The final `result` is obtained
+  as:
+  * `rounded_result = round_nearest_even(float_result / scale(result))`
+  * `result = clamp(storage_min(result), rounded_result + zero_point(result),
+            storage_max(result)`
```
burmako, line 4315: Can this be simplified to `dequantize_op_quantize(reduce_precision, operand, type(result))`?

docs/spec.md
```diff
@@ -5092,32 +5433,45 @@ def sign(x):
...
-    if compare(x, +0.0, EQ, FLOAT): return +0.0
-    if compare(x, 0.0, LT, FLOAT): return -1.0
-    return 1.0
+    def sign_computaion_float_type(x):
+        if x is NaN: return NaN
+        if compare(x, -0.0, EQ, FLOAT): return -0.0
+        if compare(x, +0.0, EQ, FLOAT): return +0.0
+        if compare(x, 0.0, LT, FLOAT): return -1.0
+        return 1.0
+    sign_computaion_float_type(x)
```
burmako, line 5442: I don't think we need this change. Below, we can just say `dequantize_op_quantize(sign, x, type(result))`.

docs/spec.md
```diff
@@ -5295,7 +5659,7 @@ More formally, for all `result_index` in the index space of `results[0]`:
 * (C3) All tensors in `inputs` and `results` have the same shape.
 * (C4) `-R` $\le$ `dimension` $\lt$ `R`, where `R` is rank of `inputs[0]`.
 * (C5) `comparator` has type
-  `(tensor<E1>, tensor<E1>, ..., tensor<EN-1>, tensor<EN-1>) -> tensor<i1>`,
+  `(tensor<E1>, tensor<E1>, ..., tensor<EN-1>, tensor<EN-1>) -> tensor<i1>`.
```
burmako, line 5662: I don't think this punctuation computes, given that there's a "where" on the next line?

docs/spec.md
```diff
@@ -3317,25 +3527,31 @@ and will likely be removed in the future
...
-* (C4) `computation` has type `(tensor<E0>, ..., tensor<EN-1>) -> tensor<E'>`
-  where `Ek` $=$ element_type(`inputs[k]`) and `E'` $=$ element_type(`result`).
+* If the operation uses non-quantized tensors:
+  * (C4) `computation` has type `(tensor<E0>, ..., tensor<EN-1>) -> tensor<E'>`
+     where `Ek` $=$ element_type(`inputs[k]`) and `E'` $=$ element_type(`result`).
+* If the operation uses quantized tensors:
+  * (C5) is_quantized_tensor(inputs[i]) for all `i in range(size(inputs))`.
+  * (C6) `computation` has type `(tensor<E0>, ..., tensor<EN-1>) -> tensor<E'>`
+     where `Ek` $=$ quantized_element_type(`inputs[k]`) and `E'` $=$
+     quantized_element_type(`result`).
```
burmako, line 3554: Can we keep C4 as is? The semantics of `map` would then be: `computation` is a function that takes whatever and returns whatever, and map simply applies it elementwise to inputs and write the result to the output.

docs/spec.md
```diff
@@ -5658,24 +6055,26 @@ Performs element-wise XOR of two tensors `lhs` and `rhs` and produces a `result`
...
+|----------|----------------------------------------------|-------------|
+| `result` | tensor of boolean, integer or quantized type | (C1)        |
 
 #### Constraints
 
 * (C1) `lhs`, `rhs` and `result` have the same type.
+* If the operation uses quantized tensors:
+  * (C2) `is_empty(quantization_dimension(lhs))`.
 
 #### Examples
```
burmako, line 6080: For the integer-only ops, how about we just decline to support them for quantized tensors?

This way, we can describe the semantics of elementwise ops with respect to quantization as "we treat quantized tensors as floating-point tensors and wrap floating-point computations in dequantize/quantize".

If someone wants to do integer-only computations on quantized tensors, they can always bitcast_convert to the storage type.

docs/spec.md
```diff
@@ -4790,22 +5122,25 @@ where `pred_val = rank(pred) == 0 ? pred : pred[i0, ..., iR-1]`.
...
+|----------|----------------------------|-------------|
+| `result` | tensor or quantized tensor | (C2)        |
 
 #### Constraints
 
 * (C1) Either `rank(pred)` $=$ `0` or `shape(pred)` $=$ `shape(on_true)`.
 * (C2) `on_true`, `on_false` and `result` have same type.
+* If the operation uses quantized tensors:
+  * (C3) `is_quantized_tensor(on_true) and is_quantized_tensor(on_false)`.
+  * (C4) `is_empty(quantization_dimension(on_true))`.
```
burmako, line 5143: Any particular reason to treat SelectOp differently from ClampOp?

docs/spec.md
```diff
@@ -1709,7 +1819,7 @@ Performs element-wise comparison of `lhs` and `rhs` tensors according to
 The values of `comparison_direction` and `compare_type` have the following
 semantics:
 
-For boolean and integer element types:
+For boolean, integer or quantized element types:
```
burmako, line 1822: I would propose that we don't introduce a special case for `compare` (similarly to the proposal about integer-only ops) and use the same `dequantize_op_quantize` mechanism for it as for other elementwise ops. 

This way, we'll be able to consistently reason about all elementwise ops, and if someone wants to compare storage types, they can bitcast convert to the storage type and perform the comparison on the result.

burmako (changes requested):
I like that this PR is working towards establishing a general principle for supporting quantized tensors, as compared with e.g. having ad-hoc rules on per-op basis. I proposed further sharpening of the principle, as well as several simplification on multiple fronts.

burmako:
#1556 has been updated with the proposed feedback for quantization-related notation, including `dequantize_op_quantize`, `baseline_type` and more.

docs/spec.md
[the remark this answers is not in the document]
dominicsymes: Thanks for clarifying - that makes sense.

docs/spec.md
```diff
@@ -2268,23 +2396,35 @@ produces a `result` tensor. Depending on the element type, does the following:
 * For integers: integer division.
```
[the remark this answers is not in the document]
dominicsymes: Thanks for clarifying. If a certain behavior is assumed (as it seems to be in this case) then I think that should be specified.

docs/spec.md
```diff
@@ -541,6 +541,55 @@ bytes is implementation-defined. String literals have type `string`.
...
+        out_zero_point):
+  float_result = ternary_computaion(
+          (operand1 - zero_point(operand1)) * scale(operand1),
+          (operand2 - zero_point(operand2)) * scale(operand2),
+          (operand3 - zero_point(operand3)) * scale(operand3))
+  rounded_result = round_nearest_even(float_result / out_scale)
+  result = clamp(out_storage_min, rounded_result + out_zero_point,
+                 out_storage_max)
+  return result
+```
```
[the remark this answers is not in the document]
rewu93: Might be a dumb question. Why not use `quantize` and `dequantize` function directly inside `dequantize_op_quantize`?

docs/spec.md
rewu93: I do find it is confusing about integer-only computation and computation on quantized tensors. My understanding is that there are two ways to do computation on quantized tensors: 
1. dequant->float op->quant
2. integer-only computation with multiplier and shift (#1475)
Here `integer-only computation` is about implementation. 

From the spec perspective,  an `integer-only computation` op seems to mean that mathematically it can only act on integer operands. In that sense `we just decline to support them for quantized tensors` is rigorous. 

Let me know if my understanding is correct. If so, is there a way to make the distinction more clear?

burmako: Apologies for the confusion! What I meant to discuss in this comment thread is the ops which don't support floating-point tensors, i.e.: `and`, `count_leading_zero`, `not`, `or`, `popcnt`, `shift_left`, `shift_right_arithmetic`, `shift_right_logical`, `xor`.

The PR description proposes to support quantized tensors in these ops and directly use integer representations of their elements, as if these elements were actually of integer type and not of quantized element type. I would propose something different - to not allow quantized tensors in these ops at all, and if someone wants to operate on integer representations directly, then they would need to bitcast_convert.

What would happen to quantized element types with multiplier and shift - that seemed to be out of scope of this PR, so I haven't thought about that yet :)

docs/spec.md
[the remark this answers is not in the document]
burmako: Good point! Since we don't yet have specs for UniformQuantizeOp and UniformDequantizeOp, we don't yet have the corresponding `uniform_quantize` and `uniform_dequantize` functions available to use in our formulas. But once we do, I agree we should switch to them.

docs/spec.md
rewu93: sgtm. `bitcast_convert` is rigorous.

docs/spec.md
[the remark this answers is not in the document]
sdasgup3: Sure. Will update in the next commit.

docs/spec.md
```diff
@@ -551,26 +600,35 @@ tensor. Depending on the element type, does the following:
 * For signed integers: integer modulus.
 * For floats: `abs` from IEEE-754.
 * For complex numbers: complex modulus.
+* For quantized types: `unary_computation_with_quantized_types(operand, lambda
+        x: abs(x), storage_min(result), storage_max(result), scale(result),
+        zero_point(result))`.
```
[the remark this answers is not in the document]
sdasgup3: Agreed!

docs/spec.md
```diff
@@ -551,26 +600,35 @@ tensor. Depending on the element type, does the following:
...
+* If the operation uses non-quantized tensors:
+  * (C2)  `operand` and `result` have the same element type, except when the
+    element type of the `operand` is complex type, in which case the element
+    type of the `result` is the element type of the complex type (e.g. the
+    element type of the `result` is `f64` for operand type `complex<f64>`).
+* If the operation uses quantized tensors:
+  * (C3) `is_quantized_tensor(operand) and is_quantized_tensor(result)`.
+  * (C4) `quantized_element_type(operand) = quantized_element_type(result)`,
+    except for quantization parameters which may differ.
+  * (C5) `is_empty(quantization_dimension(operand))`.
```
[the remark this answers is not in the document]
sdasgup3: Following how add op is specced in https://github.com/openxla/stablehlo/pull/1556, added per-tensor type in the Inputs/Outputs table.

docs/spec.md
```diff
@@ -942,24 +1000,26 @@ Performs element-wise AND of two tensors `lhs` and `rhs` and produces a `result`
...
 #### Inputs
 
-| Label | Name  | Type                              | Constraints |
-|-------|-------|-----------------------------------|-------------|
-| (I1)  | `lhs` | tensor of boolean or integer type | (C1)        |
-| (I2)  | `rhs` | tensor of boolean or integer type | (C1)        |
+| Label | Name  | Type                                         | Constraints |
+|-------|-------|----------------------------------------------|-------------|
+| (I1)  | `lhs` | tensor of boolean, integer or quantized type | (C1-C2)     |
+| (I2)  | `rhs` | tensor of boolean, integer or quantized type | (C1)        |
```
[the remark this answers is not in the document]
sdasgup3: done

docs/spec.md
```diff
@@ -942,24 +1000,26 @@ Performs element-wise AND of two tensors `lhs` and `rhs` and produces a `result`
...
-| `result` | tensor of boolean or integer type | (C1)        |
+| Name     | Type                                         | Constraints |
+|----------|----------------------------------------------|-------------|
+| `result` | tensor of boolean, integer or quantized type | (C1)        |
 
 #### Constraints
 
 * (C1) `lhs`, `rhs` and `result` have the same type.
+* If the operation uses quantized tensors:
+  * (C2) `is_empty(quantization_dimension(lhs))`.
```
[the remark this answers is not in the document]
sdasgup3: Followed the quantized add spec to simplify it further.

docs/spec.md
```diff
@@ -979,23 +1039,35 @@ Performs element-wise atan2 operation on `lhs` and `rhs` tensor and produces a
...
+* (C1) `shape(lhs) = shape(rhs) = shape(result)`.
+* If the operation uses non-quantized tensors:
+  * (C2) `element_type(lhs) = element_type(rhs) = element_type(result)`.
+* If the operation uses quantized tensors:
+  * (C3) `is_quantized_tensor(lhs) and is_quantized_tensor(rhs) and
+    is_quantized_tensor(result)`.
+  * (C4) `quantized_element_type(lhs) = quantized_element_type(rhs) =
+    quantized_element_type(result)`.
+    except for quantization parameters which may differ.
+  * (C5) `is_empty(quantization_dimension(lhs))`.
```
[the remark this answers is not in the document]
sdasgup3: Uses `baseline_type` from https://github.com/openxla/stablehlo/pull/1556 which simplifies this further.

docs/spec.md
sdasgup3: Makes sense. While doing so we might need to consider a few things: 

1. After the dequantize, the floating-point  operands need to be compared, but  which compare_type is use: FLOAT or TOTAL_ORDER? We can use `FLOAT` unless we have a special usecase to use TOTAL_ORDER. 

2. After applying op to the floating-point operands we will get tensor with boolean type. For quantization, one option is to convert the boolean type tensor to floating-point tensor and then do the quantization. But `dequantize_op_quantize` may not be used out of the box. 

3. Should we constraint  the  result quantized type to have the default quantization parameter (scale=1 and zero_point=0 ) and storage_type = `i1`.

docs/spec.md
sdasgup3: That would be great!
The only concern I had here is that  the member name function  `element_type` returns the `ElementType` part of a `TensorType`.  For `QuantizedTensorType`, I am using the member name function `quantized_element_type`. Also, I note that `baseline_element_type` may not be suitable here if we just want to use `C4` as is. 

Please let me know what do you think.

docs/spec.md
```diff
@@ -1348,6 +1421,12 @@ representation of element types is implementation-defined as well.
     * dim(`result`, `i`) $=$ dim(`operand`, `i`) for all `i` $\in$ [0, `R`-1).
     * `dim(operand, R-1) = num_bits(E')/num_bits(E)`.
 * (C2) Conversion between complex and non-complex types is not permitted.
+* If the operation uses quantized tensors:
+  * (C3) `is_quantized_tensor(operand) and is_quantized_tensor(result)`.
+  * (C4) `quantized_element_type(operand) = quantized_element_type(result)`,
+    expect for `storage_type(operand)` and `storage_type(result)` which may
+    differ.
+  * (C5) `is_empty(quantization_dimension(operand))`.
```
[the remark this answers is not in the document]
sdasgup3: I agree. It should only re-interpreting the bits of storage_type to some other type or the other way round. Will fix it.

docs/spec.md
sdasgup3: Agree with the proposal. Also updated the PR description accordingly.

docs/spec.md
sdasgup3: As the op selects one or the other operand based on `pred` value, I thought a simplifiled semantics can be proposed. 

However, based on 
>> This way, we can describe the semantics of elementwise ops with respect to quantization as "we treat quantized tensors as floating-point tensors and wrap floating-point computations in dequantize/quantize".

I agree that it would be better to have a common theme around elementwise ops.
Towards that I have updated the `select` op sepc. PTAL. 
I have added a new quantization computation `dequantize_op_quantize_select` to accommodate the boolean `pred` operand. I am not very excited about introducing a new function for an op. Even if represent `pred` as a quantized tensor, so as to use the original `dequantize_op_quantize`, I have to convert the dequantized floating-point `pred` to boolean type before applying the operation. 

Let me know what do you think.

docs/spec.md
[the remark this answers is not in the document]
sdasgup3: PTAL.

docs/spec.md
```diff
@@ -551,26 +600,35 @@ tensor. Depending on the element type, does the following:
...
+* (C1) `shape(operand) = shape(result)`.
+* If the operation uses non-quantized tensors:
+  * (C2)  `operand` and `result` have the same element type, except when the
+    element type of the `operand` is complex type, in which case the element
+    type of the `result` is the element type of the complex type (e.g. the
+    element type of the `result` is `f64` for operand type `complex<f64>`).
+* If the operation uses quantized tensors:
+  * (C3) `is_quantized_tensor(operand) and is_quantized_tensor(result)`.
+  * (C4) `quantized_element_type(operand) = quantized_element_type(result)`,
+    except for quantization parameters which may differ.
```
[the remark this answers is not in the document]
sdasgup3: I have update the spec based on the notations introduced in #1556. PTAL.

docs/spec.md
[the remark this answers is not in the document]
sdasgup3: makes sense to me.

docs/spec.md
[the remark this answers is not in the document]
sdasgup3: Agreed that we do not need them. Updated the spec. PTAL.

docs/spec.md
sdasgup3: done. Please take a look.

docs/spec.md
```diff
@@ -572,25 +572,26 @@ tensor. Depending on the element type, does the following:
 * For signed integers: integer modulus.
```
burmako, line 572: Consider refactoring the PR description to incorporate the recent round of reviews, e.g. along the lines of:

```
### Summary

The PR proposes the specification for quantized element wise operations (total 47).

### Details

Overall, we propose treating quantized elements as floating-point 
elements, and therefore ops on quantized tensors as ops on floating-point 
tensors, along the lines of dequantize -> float computation -> quantize. This 
principle works for most elementwise ops, although there are some exceptions 
discussed below.

Furthermore, the proposal is to only support per-tensor quantization for
elementwise ops. We haven't yet come across use cases for per-axis quantization
for these ops, so let's start small. The story for per-axis quantization will
be worked out in #1574.

Finally, we propose to not support hybrid quantization (i.e. situations when
some inputs/outputs are quantized and some are not) for now. The story for this
will be worked out in #1575.

- **Ops that support tensors of floating-point types (33)**: These ops support
quantized tensors, with semantics following dequantize -> float computation -> 
quantize. We are using the `dequantize_op_quantize` function to express it:
  - Binary (10): `add, atan2, compare, divide, maximum, minimum, multiply, 
    power, remainder, subtract`.
  - Unary (20): `abs, cbrt, ceil, cosine, exponential, exponential_minus_one, 
    floor, is_finite, log, logistic, log_plus_one, negate, reduce_precision, 
    round_nearest_afz, round_nearest_even, rsqrt, sign, sine, sqrt, tanh`.
  - Ternary (2): `clamp, select`.
  - Other (1): `map`.
  
- **Ops that don't support tensors of floating-point types (9)**: These ops
(`and, count_leading_zero, not, or, popcnt, shift_left, shift_right_arithmetic, 
shift_right_logical, xor`) don't support quantized tensors. If there is a need 
to perform computations on the underlying integer representation of these 
tensors, they can be bitcast_convert'ed to integers.

- **Ops that involve complex types (3)**: These ops don't support quantized
tensors because quantization doesn't compose with complex types at the moment.

- **Conversion ops (2)**: 
  - `convert`: A convert from a quantized type to any type can be realized using 
    `stablehlo.uniform_dequantize` followed by `stabhle.convert` to convert the 
    dequantized floating-point type to type of choice. Similarly, a convert from 
    any type to quantized type can be realized using `stablehlo.convert` to 
    floating-point type followed by `stablehlo.uniform_quantize`. It's not
    necessarily great that we have 3 ops to represent something that could
    theoretically be represented by 1 op, and we're planning to explore 
    a potential simplification in #1576.
  - `bitcast_convert`: Works with low-level representations, so it treats
    quantized elements as integer elements.
```

docs/spec.md
```diff
@@ -572,25 +572,26 @@ tensor. Depending on the element type, does the following:
...
-| `result` | tensor of signed integer or floating-point type | (C1), (C2)  |
+| Name     | Type                                                                           | Constraints |
+|----------|--------------------------------------------------------------------------------|-------------|
+| `result` | tensor of signed integer or floating-point type or per-tensor quantized tensor | (C1-C2)     |
 
 #### Constraints
 
-* (C1) `shape(result) = shape(operand)`.
-* (C2) `element_type(result)` is defined as:
+* (C1) `shape(operand) = shape(result)`.
```
burmako, line 591: We probably don't need the change to C1, to keep the PR focused?

docs/spec.md
```diff
@@ -1303,20 +1305,21 @@ implementation-defined as well.
...
-| `result` | tensor | (C1), (C2)  |
+| Name     | Type                                  | Constraints |
+|----------|---------------------------------------|-------------|
+| `result` | tensor or per-tensor quantized tensor | (C1-C2)     |
 
 #### Constraints
 
-* (C1) Given `E = element_type(operand)`, `E' = element_type(result)`, and
-  `R = rank(operand)`:
+* (C1) Given `E = is_quantized_tensor(operand) ? storage_type(operand) :
```
burmako, line 1320: This can be compressed to `is_quantized`.

docs/spec.md
```diff
@@ -1303,20 +1305,21 @@ implementation-defined as well.
 
 #### Inputs
 
-| Label | Name      | Type   | Constraints |
-|-------|-----------|--------|-------------|
-| (I1)  | `operand` | tensor | (C1), (C2)  |
+| Label | Name      | Type                                  | Constraints |
+|-------|-----------|---------------------------------------|-------------|
+| (I1)  | `operand` | tensor or per-tensor quantized tensor | (C1-C2)     |
```
burmako, line 1310: Since this is a reinterpret cast, let's support all quantized tensors?

docs/spec.md
```diff
@@ -1489,23 +1493,24 @@ Performs element-wise cubic root operation on `operand` tensor and produces a
 
 Performs element-wise ceil of `operand` tensor and produces a `result` tensor.
 Implements the `roundToIntegralTowardPositive` operation from the IEEE-754
-specification.
+specification. For quantized types, performs
+`dequantize_op_quantize(ceil, operand, type(result)`.
```
burmako, line 1497: Typo: missing closing parenthesis.

docs/spec.md
```diff
@@ -1578,32 +1583,34 @@ Clamps every element of the `operand` tensor between a minimum and maximum
 value and produces a `result` tensor. More formally, `result[result_index] =
 minimum(maximum(operand[result_index], min_element), max_element)`,
 where `min_element = rank(min) = 0 ? min : min[result_index]`,
-`max_element = rank(max) = 0 ? max : max[result_index]`.
+`max_element = rank(max) = 0 ? max : max[result_index]`. For quantized types,
+performs `dequantize_op_quantize(clamp, min_val, operand, max_val,
```
burmako, line 1587: Typo: min and max.

docs/spec.md
```diff
@@ -1724,27 +1731,30 @@ when `comparison_direction` is `GE`, `GT`, `LE` or `LT`
...
+| Name     | Type                                                | Constraints |
+|----------|-----------------------------------------------------|-------------|
+| `result` | tensor of boolean type or per-tensor quantized type | (C2)        |
 
 #### Constraints
 
-* (C1) `element_type(lhs) = element_type(rhs)`.
-* (C2) `shape(lhs) = shape(rhs) = shape(result)`.
+* (C1) `type(lhs) = type(rhs)`.
+* (C2) `shape(lhs) = shape(result)`.
```
burmako, line 1748: How about we keep the existing C1/C2 (replacing element_type with baseline_element_type) to keep this PR focused and defer this change to a follow-up PR?

docs/spec.md
```diff
@@ -1724,27 +1731,30 @@ when `comparison_direction` is `GE`, `GT`, `LE` or `LT`
...
 | (I4)  | `compare_type`         | enum of `FLOAT`, `TOTALORDER`, `SIGNED`, and `UNSIGNED` | (C3)        |
 
 #### Outputs
 
-| Name     | Type                   | Constraints |
-|----------|------------------------|-------------|
-| `result` | tensor of boolean type | (C2)        |
+| Name     | Type                                                | Constraints |
+|----------|-----------------------------------------------------|-------------|
+| `result` | tensor of boolean type or per-tensor quantized type | (C2)        |
```
burmako, line 1743: I think we should keep just "tensor of boolean type" here. 

This follows the spirit of "if a thing doesn't support a tensor of floating-point type, then it shouldn't support a quantized tensor", which led us to decline support for quantization in integer-only operations. Similarly, in IsFiniteOp, we don't say that its output can be a quantized tensor, we just keep "tensor of boolean type".

docs/spec.md
```diff
@@ -2244,26 +2263,29 @@ the XLA compiler. In the future, we are planning to unify this metadata
...
 #### Inputs
 
-| Label | Name  | Type                                              | Constraints |
-|-------|-------|---------------------------------------------------|-------------|
-| (I1)  | `lhs` | tensor of integer, floating-point or complex type | (C1)        |
-| (I2)  | `rhs` | tensor of integer, floating-point or complex type | (C1)        |
+| Label | Name  | Type                                                                              | Constraints |
+|-------|-------|-----------------------------------------------------------------------------------|-------------|
+| (I1)  | `lhs` | tensor of integer, floating-point, or complex type or per-tensor quantized tensor | (C1)        |
+| (I2)  | `rhs` | tensor of integer, floating-point, or complex type or per-tensor quantized tensor | (C1)        |
```
burmako, line 2278: I don't think we're using commas for similar situations anywhere else. Defer to a follow-up PR?

docs/spec.md
```diff
@@ -3273,25 +3306,31 @@ the future ([#487](https://github.com/openxla/stablehlo/issues/487)).
...
 * (C1) `shape(inputs...) = shape(result)`.
 * (C2) `0 < size(inputs) = N`.
 * (C3) `dimensions = range(rank(inputs[0]))`.
-* (C4) `computation` has type `(tensor<E0>, ..., tensor<EN-1>) -> tensor<E'>`
-  where `Ei = element_type(inputs[i])` and `E' = element_type(result)`.
+* If the operation uses non-quantized tensors:
+  * (C4) `computation` has type `(tensor<E0>, ..., tensor<EN-1>) -> tensor<E'>`
+    where `Ei = element_type(inputs[i])` and `E' = element_type(result)`.
+* If the operation uses quantized tensors:
+  * (C5) is_quantized_tensor(inputs[i]) for all `i in range(size(inputs))`.
```
burmako, line 3330: Do we really need C5 here? Given that MapOp is just a passthrough op, I think the individual element types shouldn't matter to it? This way, we'll also be able to merge C4 and C6.

docs/spec.md
```diff
@@ -3324,23 +3363,25 @@ Performs element-wise max operation on tensors `lhs` and `rhs` and produces a
...
-|----------|--------|-------------|
-| `result` | tensor | (C1)        |
+| Name     | Type                                  | Constraints |
+|----------|---------------------------------------|-------------|
+| `result` | tensor or per-tensor quantized tensor | (C1)        |
 
 #### Constraints
 
-* (C1) `type(lhs) = type(rhs) = type(result)`.
+* (C1) `baseline_type(operand) = baseline_type(result)`.
```
burmako, line 3384: It looks like for a number of ops, "type(lhs) = type(rhs) = type(result)" got accidentally replaced with "baseline_type(operand) = baseline_type(result)".

docs/spec.md
```diff
@@ -5044,23 +5101,25 @@ def sign(x):
     if is_nan(real(x)) or is_nan(imag(x)): return (NaN, NaN)
     if compare(x, (0.0, 0.0), EQ, FLOAT): return (0.0, 0.0)
     return divide(x, convert(abs(x), type(x)))
+  elif is_quantized_type(x):
+    return dequantize_op_quantize(sign, operand, type(result))`.
```
burmako, line 5105: Typo: extraneous backtick.

burmako, line 5105: I don't think `dequantize_op_quantize` works well in this function, since it's a scalar function. Let's drop these lines from the diff, and say something like of "For quantized types, performs
`dequantize_op_quantize(sign, operand, type(result))`" a few lines below.

docs/spec.md
```diff
@@ -6138,6 +6203,14 @@ def dequantize_op_quantize(op, *inputs_and_output_type):
   float_result = op(*float_inputs)
   rounded_result = round_nearest_even(float_result / scale(output_type))
   return clamp(storage_min(output_type), rounded_result, storage_max(output_type))
+
+def dequantize_op_quantize_select(op, pred, *inputs_and_output_type):
```
burmako, line 6207: If we're specializing dequantize_op_quantize to a specific op, let's go all the way:

```python
def dequantize_select_quantize(pred, on_true, on_false):
  float_on_true = (on_true - zero_point(on_true)) * scale(on_true)
  float_on_false = (on_false - zero_point(on_false)) * scale(on_false)
  float_result = select(pred, float_on_true, float_on_false)
  rounded_result = round_nearest_even(float_result / scale(output_type))
  return clamp(storage_min(output_type), rounded_result, storage_max(output_type))
```

burmako (changes requested):

docs/spec.md
sdasgup3: done. Thanks!

docs/spec.md
sdasgup3: done.

docs/spec.md

docs/spec.md
sdasgup3: agree. done.

docs/spec.md

docs/spec.md

docs/spec.md

docs/spec.md
```diff
@@ -3767,23 +3814,24 @@ produces a `result` tensor. Depending on the element type, does the following:
...
 #### Inputs
 
-| Label | Name  | Type                                               | Constraints |
-|-------|-------|----------------------------------------------------|-------------|
-| (I1)  | `lhs` | tensor of integer, floating-point, or complex type | (C1)        |
-| (I2)  | `rhs` | tensor of integer, floating-point, or complex type | (C1)        |
+| Label | Name  | Type                                                                              | Constraints |
+|-------|-------|-----------------------------------------------------------------------------------|-------------|
+| (I1)  | `lhs` | tensor of integer, floating-point, or complex type or per-tensor quantized tensor | (C1)        |
+| (I2)  | `rhs` | tensor of integer, floating-point, complex, or quantized  type                    | (C1)        |
```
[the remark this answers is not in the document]

docs/spec.md
```diff
@@ -3979,24 +4027,26 @@ More formally:
   range provided by `exponent_bits`, the intermediate result overflows to
   infinity using the original sign or underflows to zero using the
   original sign.
+* For quantized types, performs `dequantize_op_quantize(reduce_precision,
+        operand, type(result), exponent_bits, mantissa_bits)`.
```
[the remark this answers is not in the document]

docs/spec.md
```diff
@@ -4728,26 +4784,27 @@ Produces a `result` tensor where each element is selected from `on_true` or
 `on_false` tensor based on the value of the corresponding element of `pred`.
 More formally, `result[result_index] = pred_element ? on_true[result_index] :
 on_false[result_index]`, where `pred_element = rank(pred) = 0 ? pred :
-pred[result_index]`.
+pred[result_index]`. For quantized types, performs
+`dequantize_op_quantize_select(select, pred, on_true, on_false. type(result))`.
```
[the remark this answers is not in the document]

docs/spec.md

docs/spec.md

docs/spec.md
sdasgup3: >> Do we really need C5 here? Given that MapOp is just a passthrough op, I think the individual element types shouldn't matter to it? 
agreed.

>> This way, we'll also be able to merge C4 and C6.
We still need different member_name function like `element_type` and `quantized_element_type` to cover for non-quantized and quantized cases. right? `baseline_element_type` may not work here as the we want the block arguments to match exactly the input types and block return to match the output type.

docs/spec.md
sdasgup3: makes sense.

docs/spec.md
burmako: 1) "We still need different member_name function like element_type and quantized_element_type to cover for TensorType and QuantizedTensorType cases. right?". Sure. We can do something like `is_quantized(inputs[i]) ? quantized_element_type(inputs[i]) : element_type(inputs[i])`.

2) "With that we can define define element_type as a function on types such that...". I think it's useful to keep the two separate, to ensure that handling of a quantized type is an opt in (which is in line with the design principles that we've been using so far, e.g. by splitting tensor type and quantized tensor type).

3) "NonValueType ::= ElementType | FunctionType | StringType". Agreed, this should include QuantizedElementType as well.

burmako (approved):
One minor comment about MapOp, but the rest LGTM.

docs/spec.md
sdasgup3: Done.
Will address (3) in a separate PR.

sdasgup3:
@rewu93 @dominicsymes Please let me know if you have additional feedback on  the PR in its current form.

docs/spec.md
```diff
@@ -2244,26 +2262,29 @@ the XLA compiler. In the future, we are planning to unify this metadata
...
-|----------|---------------------------------------------------|-------------|
-| `result` | tensor of integer, floating-point or complex type | (C1)        |
+| Name     | Type                                                                              | Constraints |
+|----------|-----------------------------------------------------------------------------------|-------------|
+| `result` | tensor of integer, floating-point, or complex type or per-tensor quantized tensor | (C1)        |
 
 #### Constraints
 
-* (C1) `type(lhs) = type(rhs) = type(result)`.
+* (C1) `baseline_type(lhs) = baseline_type(rhs) = baseline_type(result)`.
```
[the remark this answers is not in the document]
sdasgup3: Thanks @wecing for the proposal. 

>> I think option 1 would be slightly better, since a more flexible IR is easier to manipulate.

I agree with that , but , IMHO, this comes at the cost of making the op complex by fusing a re-quantize in the semantics of divide. Can you please elaborate a bit of the benefit on such a fused version of `divide` op? Maybe it will lead to less pattern matching by the downstream consumers of StableHLO. IMO, such overhead is not too much, preserving the  generality of op plus the associated separation of concerns  is something worth considering. 
   
On a related note: Is divide is the only op which needs this change or there are other candidates (maybe `rsqrt`) which might need this similar  treatment?
 
cc @burmako

dominicsymes (approved):
Looks fine to me with exception of C6 typo as commented.

docs/spec.md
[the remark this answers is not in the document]
wecing: > this comes at the cost of making the op complex

> such overhead is not too much

Agreed.

> Is divide is the only op which needs this change

`stablehlo.add` also has this problem. For convolution followed by bias add, we may want convolution to return `qi32` and bias add be `(qi32, qi32) -> qi8`.

But since this usage is only applicable to a few ops, I think preferring simplicity in IR design and keeping the current proposal is fine.

docs/spec.md
```diff
@@ -3273,25 +3305,27 @@ the future ([#487](https://github.com/openxla/stablehlo/issues/487)).
...
+| (I3)  | `computation` | function                                                   | (C4)        |
 
 #### Outputs
 
-| Name     | Type   | Constraints |
-|----------|--------|-------------|
-| `result` | tensor | (C1), (C4)  |
+| Name     | Type                                  | Constraints      |
+|----------|---------------------------------------|------------------|
+| `result` | tensor or per-tensor quantized tensor | (C1), (C4), (C6) |
```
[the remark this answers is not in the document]
sdasgup3: fixed.

docs/spec.md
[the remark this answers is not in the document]
sdasgup3: >> stablehlo.add also has this problem. For convolution followed by bias add, we may want convolution to return qi32 and bias add be (qi32, qi32) -> qi8.

Thanks for the information. 

>> But since this usage is only applicable to a few ops, I think preferring simplicity in IR design and keeping the current proposal is fine.
Thanks for the feedback! Agreed.

wecing (approved):

sdasgup3:
@rewu93 Please let me know if the PR in its current form looks good to you.

rewu93 (approved):

docs/spec.md
```diff
@@ -6138,6 +6200,18 @@ def dequantize_op_quantize(op, *inputs_and_output_type):
...
+  float_on_false = (on_false - zero_point(on_false)) * scale(on_false)
+  float_result = select(pred, float_on_true, float_on_false)
+  rounded_result = round_nearest_even(float_result / scale(output_type))
+  return clamp(storage_min(output_type), rounded_result, storage_max(output_type))
+
+def dequantize_compare(lhs, rhs, comparison_direction):
+  float_lhs = (lhs - zero_point(lhs)) * scale(lhs)
+  float_rhs = (rhs - zero_point(rhs)) * scale(rhs)
+  return compare(float_lhs, float_rhs, comparison_direction, FLOAT)
 ```
```
burmako, line 6215: Just a reminder so that we don't forget. We'd want to use `quantize` and `dequantize` from #1496 in these new definitions as well.

sdasgup3: done.

docs/spec.md
```diff
@@ -952,23 +953,24 @@ Performs element-wise atan2 operation on `lhs` and `rhs` tensor and produces a
...
 
 #### Inputs
 
-| Label | Name  | Type                                     | Constraints |
-|-------|-------|------------------------------------------|-------------|
-| (I1)  | `lhs` | tensor of floating-point or complex type | (C1)        |
-| (I2)  | `rhs` | tensor of floating-point or complex type | (C1)        |
+| Label | Name  | Type                                                                    | Constraints |
+|-------|-------|-------------------------------------------------------------------------|-------------|
+| (I1)  | `lhs` | tensor of floating-point or complex type or per-tensor quantized tensor | (C1)        |
```
abattery, line 962: The atan2 quantization support isn't ready at the TFLite ecosystem. As far as I know, there will low possibility to use quantization on the atan2 quantization.

In this change, what is our strategy in StableHLO to define the quantization support of each operator?

sdasgup3: Hi @abattery 

>> what is our strategy in StableHLO to define the quantization support of each operator?

IMO the strategy is best described in the description and hence quoting that

>> Overall, we propose treating quantized elements as floating-point
elements, and therefore ops on quantized tensors as ops on floating-point
tensors, along the lines of dequantize -> float computation -> quantize. This
principle works for most elementwise ops, although there are some exceptions ...

I agree that there might be some ops which might not be use for quantization. In this particular case, I  acknowledge that the PR prefers ease of specification + generality over narrowing down the list of candidates.

>> there will low possibility to use quantization on the `atan2` quantization.

We found that, like for most of the element wise ops, the above applies to `tanh` as well and hence thus we went ahead to proposing its specification. 

Nevertheless, I would like to understand if there is a fundamental problem in quantizing `tanh` in the way we proposed. Please let me know. 

If not, then we might keep it for generality purpose.

docs/spec.md
```diff
@@ -1292,20 +1294,21 @@ implementation-defined as well.
...
 
 #### Constraints
 
-* (C1) Given `E = element_type(operand)`, `E' = element_type(result)`, and
-  `R = rank(operand)`:
+* (C1) Given `E = is_quantized(operand) ? storage_type(operand) :
+  element_type(operand)`, `E' = is_quantized(result) ?
+  storage_type(result) : element_type(result)`, and `R = rank(operand)`:
   * If `num_bits(E') = num_bits(E)`, `shape(result) = shape(operand)`.
   * If `num_bits(E') < num_bits(E)`:
```
loganchien, line 1313: Missed earlier, but there should be a constraint on remainder:

    `num_bits(E) % num_bits(E') == 0`

loganchien: Line 1314 as well, but in reverse:

    `num_bits(E') % num_ebits(E) == 0`

docs/spec.md
```diff
@@ -4729,26 +4779,27 @@ Produces a `result` tensor where each element is selected from `on_true` or
...
-| `result` | tensor | (C2)        |
+| Name     | Type                                  | Constraints |
+|----------|---------------------------------------|-------------|
+| `result` | tensor or per-tensor quantized tensor | (C2)        |
 
 #### Constraints
 
 * (C1) `rank(pred) = 0 or shape(pred) = shape(on_true)`.
-* (C2) `type(on_true) = type(on_false) = type(result)`.
+* (C2) `baseline_type(on_true) = baseline_type(on_false) = baseline_type(result)`.
```
loganchien, line 4802: I think we want to limit rescaling in select as well, thus instead of using `baseline_type(...)`, we should keep this as `element_type(...)`

docs/spec.md
sdasgup3: Thanks @loganchien! That is a good point. I will resolve this in a separate [PR](https://github.com/openxla/stablehlo/pull/1637).

sdasgup3: Thanks @loganchien! That is a good point. I will resolve this in a separate [PR](https://github.com/openxla/stablehlo/pull/1637).

docs/spec.md
sdasgup3: Hi @loganchien
This "extra" general behavior of select is to draw parity with other element wise operations. IMHO, this allows treating all op in the category uniformly. Let me know if is makes sense.

loganchien: If we allow different LHS/RHS/Output Scale/Zp, then we will need a transformation to rescale input scales (LHS/RHS) to the output scale (just like what we handled `tfl.maximum` before the TFLite team added the constraint to `tfl.maximum`):

```
%lhs_rescaled = "stablehlo.quantize_uniform"(%lhs)
    : (tensor<... x !quant.uniform<ui8:f32, lhs_scale:lhs_zp>>)
    -> tensor<... x !qunat.uniform<ui8:f32, out_scale:out_zp>>
%rhs_rescaled = "stablehlo.quantize_uniform"(%rhs)
    : (tensor<... x !quant.uniform<ui8:f32, rhs_scale:rhs_zp>>)
    -> tensor<... x !qunat.uniform<ui8:f32, out_scale:out_zp>>
%out = "stablehlo.select"(%cond, %lhs_rescaled, %rhs_rescaled) : ...
```

So it is the question on whether we want to prefer same scale (and explictly rescale when why are different) or to prefer different scale (and internally rescale to output scale if the backend doesn't have other means).

sdasgup3: Thanks @loganchien for clearly specificying the trade-offs.

The PR in the current form prefers the second of the two you mentioned. Actually, this is the same approach we used for almost all the element wise operations in this PR, where in, the explicit re-scaling  of operands are avoided; the op semantics will internally take care of it and, as you said, the backend will rescale if no other means are available.

loganchien: Most common implementations I am aware of: (1) use trivial element copy while selection, (2) use `cond * a + (!cond) * b` or `(mask(cond) & a) | (mask(~cond) & b)`, or  (3) SIMD instructions (AArch64 `SEL`, RISCV `VMERGE`, etc) prefers the scales to be the same.

But probably it is fine.  Seeing different scales is not the end of day.  The backend can always apply the transformation I mentioned earlier to ensure all scales are same.

This can be marked as resolved.

sdasgup3: Thanks for the additional insights.

loganchien (approved):

abattery (approved):

PR #1567, opened 2023-06-01
Specification of the remaining quantized data movement ops

sdasgup3:
The PR proposes the specification of remaining data movement ops. Specifically, the PR covers `concatenate`, `sort`, `broadcast_in_dim`, `gather`, `scatter`, `dynamic_slice`, and `dynamic_update_slice`.  The rest are covered in a separate [PR](https://github.com/openxla/stablehlo/pull/1535). 

I note that 
-  The specifications are very similar to each other and hence i thought about putting it in a single PR.
 - Following the discussion item in [link](https://groups.google.com/a/openxla.org/g/openxla-discuss/c/iwE9is49SS4/m/mi9AyvKZAQAJ), I proposed the specification to use using pex-tensor quantization granularity.  Feel free to provide your input if any of the op has a use-case to support per-axis scheme.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -5217,18 +5217,18 @@ More formally, for all `result_index` in `index_space(results[0])`:
 
 #### Inputs
 
-| Label | Name         | Type                       | Constraints |
-|-------|--------------|----------------------------|-------------|
-| (I1)  | `inputs`     | variadic number of tensors | (C1)        |
-| (I2)  | `dimension`  | constant of type `si64`    | (C4)        |
-| (I3)  | `is_stable`  | constant of type `i1`      |             |
-| (I4)  | `comparator` | function                   | (C5)        |
+| Label | Name         | Type                                            | Constraints |
+|-------|--------------|-------------------------------------------------|-------------|
+| (I1)  | `inputs`     | variadic number of tensors or quantized tensors | (C1-C7)     |
+| (I2)  | `dimension`  | constant of type `si64`                         | (C4)        |
+| (I3)  | `is_stable`  | constant of type `i1`                           |             |
+| (I4)  | `comparator` | function                                        | (C5)        |
 
 #### Outputs
 
-| Name      | Type                       | Constraints |
-|-----------|----------------------------|-------------|
-| `results` | variadic number of tensors | (C2), (C3)  |
+| Name      | Type                                            | Constraints |
+|-----------|-------------------------------------------------|-------------|
+| `results` | variadic number of tensors or quantized tensors | (C2), (C3)  |
 
 #### Constraints
 
@@ -5239,6 +5239,9 @@ More formally, for all `result_index` in `index_space(results[0])`:
 * (C5) `comparator` has type
   `(tensor<E1>, tensor<E1>, ..., tensor<EN-1>, tensor<EN-1>) -> tensor<i1>`,
   where `Ei = element_type(inputs[i])`.
+* If the operation uses quantized tensors:
+  * (C6) `is_quantized_tensor(inputs...) and is_quantized_tensor(result)`.
+  * (C7) `is_empty(quantization_dimension(inputs[0]))`.
 
 #### Examples
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -2804,7 +2804,7 @@ behavior is undefined. More formally, for all `i1 < i2` from `indices(result)`,
 
 | Label | Name                   | Type                                         | Constraints                   |
 |-------|------------------------|----------------------------------------------|-------------------------------|
-| (I1)  | `operand`              | tensor                                       | (C1), (C10-C12), (C14)        |
+| (I1)  | `operand`              | tensor or quantized tensor                   | (C1), (C10-C12), (C14-C17)    |
 | (I2)  | `start_indices`        | tensor of integer type                       | (C2), (C3), (C13)             |
 | (I3)  | `offset_dims`          | 1-dimensional tensor constant of type `si64` | (C1), (C4), (C5), (C13)       |
 | (I4)  | `collapsed_slice_dims` | 1-dimensional tensor constant of type `si64` | (C1), (C6), (C7), (C8), (C13) |
@@ -2815,9 +2815,9 @@ behavior is undefined. More formally, for all `i1 < i2` from `indices(result)`,
 
 #### Outputs
 
-| Name     | Type   | Constraints        |
-|----------|--------|--------------------|
-| `result` | tensor | (C5), (C13), (C15) |
+| Name     | Type                       | Constraints     |
+|----------|----------------------------|-----------------|
+| `result` | tensor or quantized tensor | (C5), (C13-C16) |
 
 #### Constraints
 
@@ -2842,7 +2842,12 @@ behavior is undefined. More formally, for all `i1 < i2` from `indices(result)`,
     in `slice_sizes` corresponding to `collapsed_slice_dims` are not included.
   * `combine` puts `batch_dim_sizes` at axes corresponding to `batch_dims` and
    `offset_dim_sizes` at axes corresponding to `offset_dims`.
-* (C14) `element_type(operand) = element_type(result)`.
+* If the operation uses non-quantized tensors:
+  * (C14) `operand` and `result` have the same element type.
+* If the operation uses quantized tensors:
+  * (C15) `is_quantized_tensor(operand) and is_quantized_tensor(result)`.
+  * (C16) `quantized_element_type(operand) = quantized_element_type(result)`.
+  * (C17) `is_empty(quantization_dimension(operand)`.
 
 #### Examples
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -1344,14 +1344,14 @@ in the `operand` tensor and produces a `result` tensor. More formally,
 
 | Label | Name                   | Type                                         | Constraints      |
 |-------|------------------------|----------------------------------------------|------------------|
-| (I1)  | `operand`              | tensor or quantized tensor                   | (C1-C2), (C4-C8) |
-| (I2)  | `broadcast_dimensions` | 1-dimensional tensor constant of type `si64` | (C1-C4)          |
+| (I1)  | `operand`              | tensor or per-tensor quantized tensor        | (C1-C2), (C5-C6) |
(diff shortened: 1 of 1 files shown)


docs/spec.md
```diff
@@ -1372,27 +1372,32 @@ dimensions `k` in `operand`.
...
+* (C3) All dimensions in `broadcast_dimensions` are unique.
+* (C4) For all dimensions `j` in `operand`:
   * `dim(operand, j) = 1` or
   * `dim(operand, j) = dim(result, i)` where `i = broadcast_dimensions[j]`.
+* If the operation uses non-quantized tensors:
+  * (C5) `operand` and `result` have the same element type.
+* If the operation uses quantized tensors:
+  * (C6) `is_quantized_tensor(operand) and is_quantized_tensor(result)`.
+  * (C7) `quantized_element_type(operand) = quantized_element_type(result)`.
+  * (C8) `is_empty(quantization_dimension(operand)`.
```
burmako, line 1400: Same feedback as in #1566: let's minimize the changes to the spec by: 1) using `baseline_type` and/or `baseline_element_type` where appropriate instead of branching on whether the operation uses or doesn't use quantized tensors, 2) pushing is_empty into types, as in "tensor or per-tensor quantized tensor".

docs/spec.md
```diff
@@ -1372,27 +1372,32 @@ dimensions `k` in `operand`.
```
burmako, line 1372: From PR description: "Following the discussion item in link, I proposed the specification using pex-axis quantization granularity". Did you mean "per-tensor quantization granularity"?

burmako (changes requested):

docs/spec.md
sdasgup3: Yup. Fixed.

docs/spec.md
sdasgup3: 2) done for all the ops. 

Re:1

I agree that it would be great if we could avoid branching on qunatized type. As we discussed in https://github.com/openxla/stablehlo/pull/1566#discussion_r1220364289, we need to somehow merge  `element_type` and `quantized_element_type` as they are related to differnet non-terminals  `TensorType` and `QuantizedTensorType` resp. 

A few options that I can think of: 
1. `element_type(operand) = element_type(result)` if `is_quantized(operand)`, `quantized_element_type(operand) = quantized_element_type(result)` otherwise. (Not a big fan for this as this seems too verbose).

2. Similar to how `type(x)` refers to the type of both tensor or quantized tensor types, can we use `element_type` for the element types of both. 

Specifically,

We can have a dedicated non-terminal to represent the element-type of Tensor type. Like
```
TensorType ::= 'tensor' '<' Shape TensorElementType '>'
``` 

With that we can reserve the function name `element_type` to be defined as
```
def element_type(x: Value | Placeholder | Type):
 if type(x) == TensorType:
    return tensor_element_type(x)
  if type(x) == QuantizedTensorType:
    return quantized_element_type(x)
  if type(x) is not Type:
    return element_type(type(x))
```

An additional benefit with the above approach would be that we can skip constraints like `(C6) is_quantized_tensor(operand) and is_quantized_tensor(result)`.

Please let me know your opinion.

sdasgup3:
@jingpu, @rewu93, @dominicsymes, @abattery Feel free to add your review comments.

dominicsymes (approved):
This looks fine to me in that quantized tensors can be handled in similar way to non-quantized for data movement operations. A couple of minor comments added.

docs/spec.md
burmako: Originally, I thought that we shouldn't introduce `element_type` at all, so that each potential callsite is explicit about the distinction between element types and quantized element types. However, upon further thinking, I am now convinced that this will be extra verbose without real benefits.

Overall, I agree with the proposal, except that if we rename ElementType to TensorElementType, then we should also rename QuantizedElementType to QuantizedTensorElementType.

How about doing this refactoring in a separate PR, which would also replace usages of `quantized_element_type` with `element_type` for ConvolutionOp and PadOp? I'll quickly review, we'll merge and then you'll be able to rebase #1566 (elementwise), #1535 (data movement 1) and this PR (data movement 2)?

sdasgup3: makes sense. [Here](https://github.com/openxla/stablehlo/pull/1629) is the proposed PR.

docs/spec.md
```diff
@@ -5215,18 +5229,18 @@ More formally, for all `result_index` in `index_space(results[0])`:
...
-|-------|--------------|----------------------------|-------------|
-| (I1)  | `inputs`     | variadic number of tensors | (C1)        |
-| (I2)  | `dimension`  | constant of type `si64`    | (C4)        |
-| (I3)  | `is_stable`  | constant of type `i1`      |             |
-| (I4)  | `comparator` | function                   | (C5)        |
+| Label | Name         | Type                                                       | Constraints |
+|-------|--------------|------------------------------------------------------------|-------------|
+| (I1)  | `inputs`     | variadic number of tensors or per-tensor quantized tensors | (C1-C7)     |
+| (I2)  | `dimension`  | constant of type `si64`                                    | (C4)        |
+| (I3)  | `is_stable`  | constant of type `i1`                                      |             |
```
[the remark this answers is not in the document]
sdasgup3: That is a very good point. Thanks for pointing out. 

I skimmed through all the uses of if `i1` as indices and found that we use some context to disambiguate them from type usages. For example, `for all i1 < i2 from indices(result)`.  

Nevertheless, I agree that we should be careful to (1) always use such notation with a proper context or (2) use a different letter to represent indices (as `i` is reserved for integer), to avoid future ambiguity. 

cc @burmako

docs/spec.md
```diff
@@ -5215,18 +5229,18 @@ More formally, for all `result_index` in `index_space(results[0])`:
...
 
-| Label | Name         | Type                       | Constraints |
-|-------|--------------|----------------------------|-------------|
-| (I1)  | `inputs`     | variadic number of tensors | (C1)        |
-| (I2)  | `dimension`  | constant of type `si64`    | (C4)        |
-| (I3)  | `is_stable`  | constant of type `i1`      |             |
-| (I4)  | `comparator` | function                   | (C5)        |
+| Label | Name         | Type                                                       | Constraints |
+|-------|--------------|------------------------------------------------------------|-------------|
+| (I1)  | `inputs`     | variadic number of tensors or per-tensor quantized tensors | (C1-C7)     |
```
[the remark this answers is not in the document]
sdasgup3: done

docs/spec.md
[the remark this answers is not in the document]
dominicsymes: I agree for the specification it is possible to tell from the context in the places I’ve seen. Just mentioning in case this may cause any ambiguity in other places these terms are likely to be used (eg reference code/other documents). Maybe, like you say, it is useful to have a different standard letter (such as j or k) for indices where there could be any confusion.

docs/spec.md
```diff
@@ -1344,14 +1344,14 @@ in the `operand` tensor and produces a `result` tensor. More formally,
 
 | Label | Name                   | Type                                         | Constraints   |
 |-------|------------------------|----------------------------------------------|---------------|
-| (I1)  | `operand`              | tensor                                       | (C1-C3), (C5) |
+| (I1)  | `operand`              | tensor or per-tensor quantized tensor        | (C1-C2), (C5) |
```
abattery, line 1347: why do we need to support only per tensor quantization here? When adding bias value and the result of quantized convolution, they might need to add broadcast_in_dim op if the shapes of two tensors are not the exactly same.

burmako (approved):

docs/spec.md
sdasgup3: Hi @abattery Thanks for the comment and the information.
We have a separate issue https://github.com/openxla/stablehlo/issues/1574 just to to figure out the story for per-axis quantization: where we are seeking for some clear criterion on supporting per-axis scheme for StableHLO ops. 

>> When adding bias value and the result of quantized convolution, they might need to add broadcast_in_dim op if the shapes of two tensors are not the exactly same.

From the above it seems that both  `broadcast_in_dim` and `add` could be candidates for supporting per-axis scheme. 
@abattery Do you mind help further the discussion in #1574 with your opinion? For example, could there be other candidates?

loganchien: While I think there might be a use case when we have a scalar bias (scalar in terms of storage type) in the convolution:

```
%conv = "stablehlo.convolution"(%input, %weight)
%bias_scalar_cst = "stablehlo.constant"() {value = dense<1> : tensor<i32>}
    : tensor<!quant.uniform<ui8:f32, 1.0:0>>
%bias = "stablehlo.broadcast_in_dim"(%bias_scalar_cst)
    {broadcast_dimensions = [0]}
    : (tensor<!quant.uniform<ui8:f32, 1.0:0>>)
    -> tensor<4x!quant.uniform<ui8:f32:0, 1.0:0,2.0:0,3.0:0,4.0:0>> 
%add = "stabehlo.add"(%conv, %bias) : ...
```

However, we will have to make it explicit that the broadcast operates on storage type.  In this example, it means we broadcast the storage type value `1` into 4 elements.  And their corresponding dequant values will be `[1.0, 2.0, 3.0, 4.0]`.  Also, we might have some problem in output type inference (since per-axis quantization parameters are unknowns) though this problem is nothing new to `broadcast_in_dim` (since the output shape is unknown too).

OTOH, for this particular case, I think it will make more sense to utilize the `isScalar` in the `DenseElementsAttr`, i.e. the constant can be specified as:

```
%conv = "stablehlo.convolution"(%input, %weight)
%bias = "stablehlo.constant"() {value = dense<1> : tensor<4xi32>}
    :tensor<4x!quant.uniform<ui8:f32:0, 1.0:0,2.0:0,3.0:0,4.0:0>>
%add = "stabehlo.add"(%conv, %bias) : ...
```

loganchien (approved):
Approve (as this change focus on per-tensor quantization)

docs/spec.md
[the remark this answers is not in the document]
sdasgup3: Point noted.

paulinesho (approved):

docs/spec.md
abattery: Thanks Logan for the detailed example. Also I'm fine with the discussion separately.

abattery (approved):

Issue #1572, opened 2023-06-03
Revisit verification and type inference tests as part of conformance suite


burmako:
When writing the spec and then when implementing the interpreter, we put a lot of effort into adding structure to verification and type inference tests, but I still find them somewhat hard to wrap my head around. For example, some things that immediately come to mind are:
  1) Verifiers for some ops are in ops_stablehlo.mlir and for other ops are in separate verify_foo.mlir files. I think we should have separate verify_foo.mlir tests for all ops, similarly to how we have separate interpret_foo.mlir tests.
  2) In these files, should we also explicitly call out the constraints which are not covered by ODS as deliberately out of scope? Right now, we just skip them and don't mention them at all.
  3) We don't have crisp guidelines for type inference tests, so things are hectic there overall.
  4) Positive tests don't have crisp guidelines either. Do we even need to have them?

burmako:
Related ticket: #1004.

burmako:
On a related note, because of #1571, we may end up skipping another round of audits / introducing additional structure for most of the Not in HLO and Miscellaneous ops. These ops, if they end up staying in the StableHLO opset, could use extra attention.

PR #1573, opened 2023-06-04
Fail interpretation on Not in HLO and Miscellaneous ops

burmako:
As we're getting closer and closer to completion of the intepreter thanks to great work by Gunhyun, our corner cases are becoming more and more prominent.

One of these corner cases is support for Not in HLO ops, which we are planning to remove from the StableHLO opset but which are still part of the StableHLO dialect.

Another corner case is support for Miscellaneous ops, all of which except for FftOp have decomposition passes implemented either on MHLO or HLO. Some of these decompositions are really tricky, and it would be a shame to duplicate their code and to maintain this duplicated code in the interpreter.

As a result, I propose that we fail when encountering these ops and leave decomposition of these ops to be the responsibility of the users of the interpreter for now.

In the future, we should provide for these ops in this repository (or move these ops out of the StableHLO opset). We can do this via a a combination of moving the already existing MLIR-based passes to this repository (and running them before the interpreter runs) and/or porting the already existing HLO-based passes to MLIR in this repository.

This proposal also involves deleting the already implemented eval functions and tests for BatchNormGradOp, BatchNormInferenceOp, BatchNormTrainingOp and CholeskyOp. A lot of work went into these implementations (which will not be forgotten!), and I'm hesitating to remove them. However, this is a non-trivial amount of code and it may need non-trivial amount of maintenance down the line, so my recommendation would be to remove it.


ghpvnist (changes requested):
> In the future, we should provide for these ops in this repository (or move these ops out of the StableHLO opset).

Sounds good to me as we can always rollback the changes if we end up needing it again.

Some comments:
1. Can you also update `spec.md` to remove the "More Example" links?
2. How should we go about skipping the checklist for Miscellaneous ops? They won't be tagging the constraints, renaming/deduplicating tests, etc.

burmako:
1) Done.
2) Great question! This is why I opened [this ticket](https://github.com/openxla/stablehlo/issues/1572#issuecomment-1575301772). Overall, I'm thinking that the Conformance Suite workstream could be the proper home for this work, so when we get to it, we would address this item.

ghpvnist (approved):

Issue #1514, opened 2023-05-21
Consider adding TopK


burmako:
Merged PR that introduced kTopK in HLO: https://github.com/openxla/xla/pull/2539. Pending PR that introduces the same functionality in MHLO: https://github.com/openxla/xla/pull/3127.

burmako:
@kuhar This is the ticket that we're using on the StableHLO side to track the work for potentially adding TopK to the StableHLO opset, following its introduction to MHLO.

burmako:
One of the questions that has come is whether the output of TopK is sorted or not. chlo::TopK [currently says that it is](https://github.com/openxla/stablehlo/blob/4344550b2a239705a0c5b236b1e6251822b5ead8/stablehlo/dialect/ChloOps.td#L817-L818).

Issue #1396, opened 2023-04-12
TopKOp with dynamic shapes 


gnecula:
### Request description

JAX users are asking for TopKOp with dynamic value for the `k` parameter.

burmako:
A prototype has landed in https://github.com/tensorflow/mlir-hlo/commit/9137ea725a4a817154e58daf3d2e4135a51abb52, a long-term solution will be presented in the upcoming dynamism RFC.

PR #1593, opened 2023-06-06
[RFC] Add TopK op to StableHLO

kuhar:
TopK is a common operation in ML frameworks and dialects, but is currently missing from StableHLO. This RFC proposed to add a new op to the StableHLO dialect: `stablehlo.topk`, and argues why `sort` is insufficient.

Note that both CHLO and MHLO already support TopK:
-  PR that introduced TopK in HLO: openxla/xla#2539.
-  PR that introduces the same functionality in MHLO: openxla/xla#3127.

Issue: openxla/stablehlo#1514
openxla-discuss: https://groups.google.com/a/openxla.org/g/openxla-discuss/c/hhieJAN4Eis

### DIFF
--- a/rfcs/20230606-topk.md
+++ b/rfcs/20230606-topk.md
@@ -14,7 +14,7 @@ the IR lowering chain.
 
 ## Background and Motivation
 
-The TopK operation is ubiquitous in both ML frameworks (e.g.,
+The TopK operation is ubiquitous in many ML frameworks (e.g.,
 [JAX](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.top_k.html),
 [tensorflow](https://www.tensorflow.org/api_docs/python/tf/math/top_k),
 [PyTorch](https://pytorch.org/docs/stable/generated/torch.topk.html)) and
@@ -77,6 +77,10 @@ Sample IR:
   attribute, we can more faithfully model input models and differentiate from
   sorting, especially when `k` is dynamic. This also makes it much more explicit
   whether sorting is performed, reducing the need to consult the StableHLO spec.
+* Make ordering a function argument to dynamic selection between sorted and
+  unordered results. It is unclear that there is need for this kind of dynamism.
+  The alternative -- emitting a `stablehlo.select` of the  two variants of
+  `stablehlo.top` provides an easy to employ workaround.
 
 ## Acknowledgements
 
--- a/rfcs/20230606-topk.md
+++ b/rfcs/20230606-topk.md
@@ -1,0 +1,88 @@
+# RFC: Add TopK Op to StableHLO
+
+Status: Draft<br/>
+Initial version: 2023-06-06<br/>
+Last updated: 2023-06-06
+
+## Summary
+
+We propose to add a new op: `stablehlo.topk`, which returns the first `k`
+elements of a tensor and their indices, based on a user-provided comparator. The
+goal of the new op is to bridge a gap between existing and emerging uses of TopK
+in models (especially transformers and LLMs, e.g.: [^1], [^2]), as well as in
+the IR lowering chain.
+
+## Background and Motivation
+
+The TopK operation is ubiquitous in both ML frameworks (e.g.,
+[JAX](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.top_k.html),
+[tensorflow](https://www.tensorflow.org/api_docs/python/tf/math/top_k),
+[PyTorch](https://pytorch.org/docs/stable/generated/torch.topk.html)) and
+compiler IRs ([XLA](https://www.tensorflow.org/xla/operation_semantics#top-k),
+[CHLO](https://github.com/openxla/stablehlo/blob/6f0d25ef2e05f4549c5e4ffc506436a4a13af51f/stablehlo/dialect/ChloOps.td#L811-L842),
+[MHLO](https://github.com/openxla/xla/commit/11a4424760f8814faa6da29a0e6dd9d5c168d22a#diff-9573e4e3a2bf686edee3437726f9dbccaf1b296e634d44f4df504d94bf2576d1),
+[LinalgExt](https://github.com/openxla/iree/blob/b5e45cd08eecbd97b0568f9c5fdac28b0f5e8f11/llvm-external-projects/iree-dialects/include/iree-dialects/Dialect/LinalgExt/IR/LinalgExtOps.td#L436-L480)),
+which makes its absence from StableHLO a point of friction. For example, in IREE
+we have to pattern match stablehlo `iota` + `sort` + `slice` and raise it back
+to `chlo.top_k`, so that we can efficiently lower it to `linalg_ext.topk`. See
+the IREE issue for more details: <https://github.com/openxla/iree/issues/13905>.
+
+The definitions of TopK in frontend frameworks and IRs vary in two places:
+
+* Whether `k` is static or dynamic.
+* Whether the results are sorted, unordered, or whether ordering is decided based
+  on a function argument.
+
+Note that when the input TopK gets lowered to `sort` followed by a dynamic slice,
+we generally lose the information about both. If `k` is a runtime value, we no
+longer know whether we can expect it to be much smaller than the number of
+elements, which makes it difficult to tell whether to rewrite sort as TopK or
+whether we truly have to sort all elements. Second, we may be introducing
+unnecessary ordering when the input semantics do not require for the results to
+be sorted.
+
+Note that we want to avoid degrading to a general `sort`, as this can lead to
+significant bottlenecks in real-world models, e.g.:
+<https://github.com/openxla/iree/issues/13865>.
+
+This loss of high level semantic information and high friction along the
+lowering chain makes us argue for a dedicated TopK op in StableHLO.
+
+## Implementation
+
+We propose to add a `stablehlo.topk` op with dynamic argument `k` and an
+attribute `order` (either `SORTED` or `UNORDERED`). This version of TopK should
+be a faithful representation of TopK in input frameworks that does not introduce
+needless sorting.
+
+Sample IR:
+
+```mlir
+%values, %indices = stablehlo.topk %values, %k, UNORDERED {
+  ^bb0(%arg0: tensor<f32>, %arg1: tensor<f32>):
+    %predicate = stablehlo.compare GT, %arg0, %arg1 : (tensor<f32>, tensor<f32>) -> tensor<i1>
+    stablehlo.return %predicate : tensor<i1>
+} : (tensor<100xf32>, tensor<i64>) -> (tensor<5xf32>, tensor<5xi32>)
+```
+
+### Additional Considerations
+
+* When `k` is static, we should be able to trivially recognize it is a constant
+  and use that constant in further lowering to a `lower` dialect. When it is a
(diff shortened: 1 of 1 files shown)


kuhar:
This the first draft of this RFC and also my very first contribution to StableHLO, so I'd really appreciate your feedback.

kuhar:
> @kuhar, thank you for your contribution! I'll review this RFC shortly. Let's also solicit community feedback by posting a link to this RFC on [openxla-discuss](https://groups.google.com/a/openxla.org/g/openxla-discuss/).

OK, I'll cross-post it there in a few mins

rfcs/20230606-topk.md
```diff
@@ -0,0 +1,88 @@
...
+whether we truly have to sort all elements. Second, we may be introducing
+unnecessary ordering when the input semantics do not require for the results to
+be sorted.
+
+Note that we want to avoid degrading to a general `sort`, as this can lead to
+significant bottlenecks in real-world models, e.g.:
+<https://github.com/openxla/iree/issues/13865>.
+
+This loss of high level semantic information and high friction along the
+lowering chain makes us argue for a dedicated TopK op in StableHLO.
```
joker-eph, line 49: In the recent discussions around MHA, we were floating the idea of having a different mechanism for ops that can be decomposed (like special function call) to preserve them as subgraph but providing any StableHLO compiler a straightforward way to match it to a first class op internally.
Wouldn't that apply here as well?

kuhar: Do you have some link with a reference to these discussions?

rfcs/20230606-topk.md
```diff
@@ -0,0 +1,88 @@
...
+
+We propose to add a new op: `stablehlo.topk`, which returns the first `k`
+elements of a tensor and their indices, based on a user-provided comparator. The
+goal of the new op is to bridge a gap between existing and emerging uses of TopK
+in models (especially transformers and LLMs, e.g.: [^1], [^2]), as well as in
+the IR lowering chain.
+
+## Background and Motivation
+
+The TopK operation is ubiquitous in both ML frameworks (e.g.,
```
impjdi, line 17: s/both/many ?

rfcs/20230606-topk.md
impjdi: This is a pain point StableHLO representation should solve.  And I don't think this should lead to `pollution` of arbitrary ops into the StableHLO op set.  Let me give you an example that I recently ran into:

Gaussian Error Linear Unit (GELU) https://arxiv.org/pdf/1606.08415.pdf is an activation layer we needed to support.  It's used in some other text models as well, so it's not a special case.  In the current status, we would have used the first approximation listed in the paper which would get decomposed into 7 ops.  It was performance critical, yes, it was important to us, yes, but do I think it should be added to the StableHLO op set?  My answer would be a no.  The user should do the graph matching and do a custom op for this case if they really care about performance.  Without this, there will be a GELU (original), GELU_V2 (first approximation), and GELU_V3 (second approximation), because one version didn't work for everybody (there are some numerical differences, making the network produce worse output).

GELU was one example.   If we start allowing one performance op into the builtin op set, we can't say no to any other RFCs introducing custom ops into the StableHLO op set.  Someone will come with a hash table look up op and another would want a k-means op.

What StableHLO must resolve is the preserving the high level semantic information mentioned here, so that it's easy for the users to write custom acceleration.

silvasean: Some links:
https://github.com/pytorch/pytorch/blob/main/torch/_decomp/decompositions.py
https://github.com/llvm/torch-mlir/blob/2480cb7a512d249a0cc41b2079313b5ada3c399d/python/torch_mlir/dynamo.py#L39
https://github.com/llvm/torch-mlir/blob/2480cb7a512d249a0cc41b2079313b5ada3c399d/include/torch-mlir/Dialect/Torch/Transforms/Passes.h#L45

GleasonK: > a winning strategy here is to declare part of StableHLO the "primitive"/"core" subset and have "decompositions" that can turn any non-primitive op into a combination of core ops.

Fully agree with this! @burmako and I iterated on a design that allows dialects to define StableHLO decompositions, and would appreciate any feedback: [[RFC] StableHLO Extensibility](https://docs.google.com/document/d/1bSyyLA-p1F7KjZgjo563F1WFsPwcZc4eaH5WyQfbsi0/edit#heading=h.kfv34azf3j5k)

This doc proposes two StableHLO interfaces that arbitrary dialects / ops can implement that would allow for backends to specify supported non-primitive ops _at the time of deserialization_. If a non-primitive op is unsupported by a backend, a StableHLO decomposition is used instead.

> We would then have a standardized library of such decompositions that users of StableHLO can select.

This is a separate design point worth discussing, if there should be one canonical extension dialect, or several extension dialects. The proposed design is for arbitrary dialects, but with slight modification would work for a single dialect.

rfcs/20230606-topk.md
```diff
@@ -0,0 +1,88 @@
...
+
+We propose to add a `stablehlo.topk` op with dynamic argument `k` and an
+attribute `order` (either `SORTED` or `UNORDERED`). This version of TopK should
+be a faithful representation of TopK in input frameworks that does not introduce
+needless sorting.
+
+Sample IR:
+
+```mlir
+%values, %indices = stablehlo.topk %values, %k, UNORDERED {
```
changm, line 61: small nit: Why an attribute versus a boolean arg?

rfcs/20230606-topk.md
```diff
@@ -0,0 +1,88 @@
+# RFC: Add TopK Op to StableHLO
+
+Status: Draft<br/>
+Initial version: 2023-06-06<br/>
+Last updated: 2023-06-06
+
+## Summary
+
+We propose to add a new op: `stablehlo.topk`, which returns the first `k`
```
changm, line 9: nit: Why topk versus DynamicTopK if it supports dynamic operands?

rfcs/20230606-topk.md
silvasean: I would try to create an ecosystem around this. What I'm seeing from Torch-MLIR is that a lot of people see a lot of value in StableHLO outside the stable artifact itself. So in addition to the core value of the stable artifact, an orthogonal direction is serving as a "lingua franca" for ML compilation, which often will not have all the same constraints as the stable artifact.

Reading Kevin's proposal, it seems like a great idea with the serialized decomposition inline in the module which allows implementations that are unaware of the custom op at least correctly execute :+1:

One design point I'm not sure about is the idea that the proxy is a proxy for a specific op in another dialect, or whether the "custom operation" is abstractly defined independent of any other dialect. I'm not sure that "MLIR dialects" are a strong enough backplane for defining an ecosystem. Probably some org.mycompany.Foo type scheme would be better. Or a Khronos style central registry with vendor proposals that get mixed and matched by vendors and eventually rolled into the next version of the standard if successful. That seems like it would be more durable and build more consensus and glide path to adoption.

I think the balance here is defining
1) a way for vendors to tunnel through arbitrary operations in a well-defined way that is lightweight (is asking for an HLO reference too much to ask for? What about stuff that is not expressible as HLO, such as some linked list thing or other weirdness). Idea: java-style org.mycompany.Foo improved, more-disciplined version of custom_call
2) a way for the community to gradually coalesce around more stable and winning designs. Idea: Khronos style extensions
3) The "official StableHLO"
4) The primitive subset of "official StableHLO"

Ideally we provide a glide path for each operation to graduate core-wise to the maximal level that makes sense.

Note that one of the core value-adds of StableHLO is that it is a small set so that it is plausible to write "a X rule for each operator" (e.g. partitioning), but we have seen that e.g. for TopK it is actually valuable to partition the non-decomposed version, which is more work (have to write a new rule), but better performance, so worth it.

The name of the game here is gradients, allowing things to be stood up and more work invested for greater interop and performance as needed.

rfcs/20230606-topk.md
kuhar: I thought about it some more and I think that you are right and static `k` is the better choice

rfcs/20230606-topk.md
kuhar: I had similar questions and doubts to @silvasean but decided to continue on openxa-discuss, at this is not directly related to the TopK RFC itself: https://groups.google.com/a/openxla.org/g/openxla-discuss/c/Ao5K8fvXoEk/m/JG0DGzhqAgAJ

rfcs/20230606-topk.md
burmako: Interestingly, we have [a feature request](https://github.com/openxla/stablehlo/issues/1396) from @gnecula about supporting dynamic `k` :)

rfcs/20230606-topk.md
GleasonK: This sounds very promising. The initial design I proposed for composites last September was basically "custom_call with default implementation", and the primary feedback was that this may cause ecosystem fragmentation and we should look for an approach that leads to convergence. I think backing this with a public spec may be one of the missing pieces: It would allow backends to safely optimize, community members to reuse existing specs, and frameworks don't need additional dependencies as long as StableHLO provides a means of implementing the spec (maybe functions+metadata, maybe custom_call++).

One possible option: Allow any non-stable abstractions to be made for experimentation, but only permit specced features to appear in portable artifacts.

rfcs/20230606-topk.md
changm: Hmm thanks for that spec! If this op should support dynamic k, what's the StableHLO convention for dynamic ops? For example, Slice has SliceOp and DynamicOp for the static / dynamic pair. If this op should support dynamic k, is this really a DynamicTopK and we'll have a corresponding TopK for static?

burmako: I'm hoping to discuss this in the upcoming dynamism RFC in more detail (ETA: end of week), but I would propose that we support both in the same TopKOp. Overall, we have received feedback that the Foo/DynamicFoo split makes things harder than they should be for both producers and consumers.

rfcs/20230606-topk.md
kuhar: In all of the uses of topk in frameworks I saw, the ordering was decided statically for each callsite. If dynamism is required, it should be simple to emit a select with two variants.

rfcs/20230606-topk.md
kuhar: The 'both' refers to (1) ML frameworks and (2) compiler IRs. This is a bit clumsy, so I changed it like you suggested. Thanks.

rfcs/20230606-topk.md
kuhar: I added this to 'Additional Considerations'

GleasonK:
Hello! I spoke with @kuhar back in the fall and they mentioned that this RFC is no longer a priority and they will not be submitting it, so I will be closing it as such.

I do want to note that there are two important topics covered by this PR:
1. A mechanism to preserve higher level ops later in the compilation pipeline
2. A place for winning abstractions to reside -- either as StableHLO primitives with expanders, or in another opset with decompositions.

In some sense, I think another design for this could be to have CHLO be an ATen-like abstraction opset and improve its ability to be optionally preserved / decomposed later in the compilation pipeline (would require some sort of serialization support).

These are both topics that I hope to address in the coming months, and will use TopK as a case study, but will hope to propose something more generic. In the meantime I will be closing this PR as it still unclear what is the best path forward and there is no clear owner of figuring out that task yet.

Issue #1549, opened 2023-05-28
Add a CMake configuration for external projects


burmako:
Right now, StableHLO is oftentimes depended on through MLIR-HLO, but given [the plans to sunset the MLIR-HLO repository](https://discourse.llvm.org/t/sunsetting-the-mlir-hlo-repository/70536), we need to tighten our story in this area. 

Direct dependencies on the StableHLO repository already work through Bazel (e.g. this is how [XLA depends on StableHLO](https://github.com/openxla/xla/tree/main/third_party/stablehlo)), but CMake support is currently missing. We would need to implement something along the lines of [what MLIR-HLO does](https://github.com/tensorflow/mlir-hlo/tree/master/cmake/modules) - this doesn't appear to be particularly hard.

PR #1612, opened 2023-06-13
[CMake] Make StableHLO build consistent with MLIR-HLO

kuhar:
This applies when stablehlo is built by other projects enabled as an
external llvm project.

This change makes it match the mlir-hlo configuration more closely:
https://github.com/tensorflow/mlir-hlo/blob/51ce2fa22bc9acff37332ad02ef39412f21ffc98/CMakeLists.txt#L113-L130,
and ease the transition from mlir-hlo to stablehlo:                                                                      
https://discourse.llvm.org/t/sunsetting-the-mlir-hlo-repository/70536.

Tested by embedding stablehlo into IREE.

Issue: https://github.com/openxla/stablehlo/issues/1549

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,17 +37,25 @@ endif()
 #-------------------------------------------------------------------------------
 # Options and settings
 #-------------------------------------------------------------------------------
+option(STABLEHLO_BUILD_EMBEDDED "Build StableHLO as part of another project" OFF)
 option(STABLEHLO_ENABLE_BINDINGS_PYTHON "Enables StableHLO Python bindings" OFF)
 option(STABLEHLO_ENABLE_STRICT_BUILD "Build StableHLO with strict warnings and warnings as errors" OFF)
 
 #-------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -37,19 +37,27 @@ endif()
 #-------------------------------------------------------------------------------
 # Options and settings
 #-------------------------------------------------------------------------------
+option(STABLEHLO_BUILD_EMBEDDED "Build StableHLO as part of another project" OFF)
```
burmako, line 40: There are quite a few related boolean variables - STABLEHLO_BUILD_EMBEDDED, STABLEHLO_STANDALONE_BUILD and STABLEHLO_EXTERNAL_PROJECT_BUILD - and I think that not all combinations of them are valid, so it's a bit confusing. 

However, given that this PR aims to exactly reproduce the MLIR-HLO setup, I think we should keep these variables as is to first establish the baseline and then think about refining it. I'll maybe look into this in a follow-up PR.

burmako (approved):

PR #1614, opened 2023-06-14
Further align StableHLO and MLIR-HLO builds

burmako:
To follow up on #1612, this PR further aligns StableHLO and MLIR-HLO builds to facilitate the transition from mlir-hlo to stablehlo.

Changes in this PR:
  1) Adds Ccache support.
  2) Adds support for LLVM_ENABLE_ZLIB.
  3) Removes STABLEHLO_STANDALONE_BUILD, to better align with MLIR-HLO
     terminology.

Deltas between CMakeLists.txt in this PR and CMakeLists.txt in MLIR-HLO:
  a) Says "StableHLO" instead of "MHLO".
  b) Supports STABLEHLO_ENABLE_STRICT_BUILD (there is no equivalent
     in the MLIR-HLO build).
  c) Doesn't support ${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules yet,
     as well as some MLIR_HLO_FOO_DIR variables needed for that.
     (We have #1549 to follow up on this).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,18 +44,39 @@ option(STABLEHLO_ENABLE_STRICT_BUILD "Build StableHLO with strict warnings and w
 #-------------------------------------------------------------------------------
 # Project setup and globals
 #-------------------------------------------------------------------------------
-set(STABLEHLO_STANDALONE_BUILD OFF)
+set(STABLEHLO_EXTERNAL_PROJECT_BUILD OFF)
 
-if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
-  set(STABLEHLO_STANDALONE_BUILD ON)
+if(NOT (CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR) AND NOT MLIR_BINARY_DIR)
+  # Building as part of LLVM via the external project mechanism.
+  set(STABLEHLO_EXTERNAL_PROJECT_BUILD ON)
+else()
+  # Building standalone.
   project(stablehlo LANGUAGES CXX C)
   set(CMAKE_C_STANDARD 11)
   set(CMAKE_CXX_STANDARD 17)
 endif()
 
-if(NOT STABLEHLO_STANDALONE_BUILD AND NOT MLIR_BINARY_DIR)
-  # Building as part of LLVM via the external project mechanism.
-  set(STABLEHLO_EXTERNAL_PROJECT_BUILD ON)
+# Build with ccache if the package is present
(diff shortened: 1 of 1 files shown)


kuhar:
> 1. Adds Ccache support.

Do we really need this? AFAIK IREE removed one a few months ago because CMake supports this well without any project-specific configuration. (So arguably LLVM shouldn't need it either.) I think all you need to do is set:
```shell
-DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_C_COMPILER_LAUNCHER=ccache
```

CMakeLists.txt
```diff
@@ -44,18 +44,39 @@ option(STABLEHLO_ENABLE_STRICT_BUILD "Build StableHLO with strict warnings and w
...
   project(stablehlo LANGUAGES CXX C)
   set(CMAKE_C_STANDARD 11)
   set(CMAKE_CXX_STANDARD 17)
 endif()
 
-if(NOT STABLEHLO_STANDALONE_BUILD AND NOT MLIR_BINARY_DIR)
-  # Building as part of LLVM via the external project mechanism.
-  set(STABLEHLO_EXTERNAL_PROJECT_BUILD ON)
+# Build with ccache if the package is present
+set(LLVM_CCACHE_BUILD OFF CACHE BOOL "Set to ON for a ccache enabled build")
```
GleasonK, line 60: I wonder if we can rely on this flag in CI? Currently we use a ccache step to set this up I believe. Doesn't need to be addressed now.

CMakeLists.txt
```diff
@@ -94,6 +115,10 @@ else()
   message(STATUS "Building StableHLO embedded in another project")
 endif()
 
+if(LLVM_ENABLE_ZLIB)
+  find_package(ZLIB)
```
GleasonK, line 119: Curious on error checking. We have special handling for `find_program` but not `find_package`.

Looks like we can add `find_package(... REQUIRED)` and it will error for us if it cant be found. Same for `find_program` in CMake 3.18+

GleasonK (approved):

CMakeLists.txt
burmako: @kuhar @GleasonK Good point about Ccache! I think we should keep `LLVM_CCACHE_BUILD` in our build for 1:1 compatibility with MLIR-HLO and not do anything further for now. Down the line, when we sunset MLIR-HLO and folks successfully migrate, we can refactor.

CMakeLists.txt
burmako: For now, let's keep this as is to replicate the MLIR-HLO build behavior (even minor divergence may complicate migrations). Can revisit in the future.

PR #1620, opened 2023-06-15
Add stablehlo-lsp-server

joker-eph:
This allows VSCode plugin integration as documented here:

https://mlir.llvm.org/docs/Tools/MLIRLSP/#mlir-lsp-language-server--mlir-lsp-server

### DIFF
--- a/stablehlo/tools/CMakeLists.txt
+++ b/stablehlo/tools/CMakeLists.txt
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 set(LLVM_OPTIONAL_SOURCES
+  StablehloLspServerMain.cpp
   StablehloOptMain.cpp
   StablehloTranslateMain.cpp
   Interpreter.cpp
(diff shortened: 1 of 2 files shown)


stablehlo/tools/StableHloLspServerMain.cpp
```diff
@@ -0,0 +1,22 @@
+#include "mlir/Tools/mlir-lsp-server/MlirLspServerMain.h"
```
burmako, line 1: Nit: Let's rename to `StablehloLspServerMain.cpp` to follow the conventional spelling of StableHLO in code: "Stablehlo" (without capitalization except for the first letter as appropriate).

burmako (approved):
Thank you! Looks great to me, except for one stylistic comment.

burmako:
Oops right I clicked "Merge", I realized that we'll also need to update the Bazel build. Will do that in a moment.

joker-eph:
Hopefully that didn't break anything? (there is no test relying on this, it just means the lsp-server can't be built with Bazel ATM)

burmako:
Yeah, everything's alright - although I've added stablehlo-lsp-server to both Bazel and CMake CI to make sure that it stays this way :)

PR #1629, opened 2023-06-20
Refactor the grammar for element types

sdasgup3:
As brought up in https://github.com/openxla/stablehlo/pull/1567#discussion_r1217167342, there is a need to introduce a computation function, with name `element_type`, which can be applied for both `TensorElementType` and `QuantizedTensorElementType` in order to simplify the operation specification.

The PR does the following:
1. Rename `ElementType` to `TensorElementType`. 
2. Rename `QuantizedElementType` to `QuantizedTensorElementType`.
3. Introduction a quantization computation function `element_type`.
4. Use the function at (3) for `convolution` and `pad`.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -532,11 +532,11 @@ represents a tensor value with the following mapping from indices to elements:
 `{1, 2} => 6`. The order in which these elements are then stored in memory is
 implementation-defined. Tensor constants have the following constraints:
 
-* (C1) `has_syntax(tensor_literal, tensor_element_type(tensor_type))`, where:
-  * `has_syntax(element_literal: Syntax, tensor_element_type: Type) =
+* (C1) `has_syntax(tensor_literal, element_type(tensor_type))`, where:
+  * `has_syntax(element_literal: Syntax, element_type: Type) =
     is_wellformed(element_literal, type)`.
-  * `has_syntax(tensor_literal: List, tensor_element_type: Type) =
-    has_syntax(tensor_literal..., tensor_element_type)`.
+  * `has_syntax(tensor_literal: List, element_type: Type) =
+    has_syntax(tensor_literal..., element_type)`.
 * (C2) `has_shape(tensor_literal, shape(tensor_type))`, where:
   * `has_shape(element_literal: Syntax, []) = true`.
   * `has_shape(tensor_literal: List, shape: List) =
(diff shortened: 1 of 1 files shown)


docs/spec.md
```diff
@@ -6184,6 +6178,20 @@ on types" section via `member_name`.
 
 #### Quantization computations
 
+* `element_type` is defined on tensor types and quantized tensor types and
```
burmako, line 6181: Let's put this function in the list of functions in this section in alphabetical order.

docs/spec.md
```diff
@@ -6198,8 +6206,8 @@ def baseline_type(x: Value | Placeholder | Type) -> Type:
   if type(x) == TensorType:
     return x
   if type(x) == QuantizedTensorType:
-    element_type = quantized_element_type(x)
-    baseline_element_type = QuantizedElementType(
+    element_type = quantized_tensor_element_type(x)
```
burmako, line 6209: I think that we should replace most of the occurrences of `tensor_element_type` and `quantized_tensor_element_type` with just `element_type` to make things more concise. The more specific functions have additional functionality that ensures that the argument is of a specific type, but that's typically checked elsewhere, e.g. by input constraints that say "tensor" or "tensor or quantized tensor".

There are a few uses of more specific functions in this PR which I think are warranted:
  * `tensor_element_type` in the definition of `member_name`.
  * `tensor_element_type` and `quantized_element_type` in the definition of `element_type`.

docs/spec.md
```diff
@@ -1152,7 +1152,7 @@ def batch_norm_inference(operand, scale, offset, mean, variance, epsilon, featur
 
 * (C1) `0 <= feature_index < rank(operand)`.
 * (C2) `operand`, `scale`, `offset`, `mean`, `variance` and `result` have the
-  same element type.
+  same tensor element type.
```
burmako, line 1155: Similarly to the comment about `tensor_element_type` and `quantized_tensor_element_type`, I don't think we should be this specific here and say "tensor element type" - "element type" is just fine. (Here and below).

burmako (changes requested):

docs/spec.md
sdasgup3: done

docs/spec.md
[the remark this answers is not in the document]

docs/spec.md

burmako (approved):

PR #1608, opened 2023-06-09
Adding the missing quantized-type support in the ODS

sdasgup3:
Some of the stablehlo ops does not have support for quantized types in their  tablegen specification, prohibits writing StableHLO quantized programs using those ops. The PR is about adding the missing support for the following ops. Also, I believe the ongoing specification [work](https://github.com/openxla/stablehlo/issues/588), should not deviate much from the  proposed changes here.  

```
stablehlo.atan2
stablehlo.divide
stablehlo.power
stablehlo.remainder
stablehlo.subtract

stablehlo.abs
stablehlo.cbrt
stablehlo.cosine
stablehlo.exponential
stablehlo.exponential_minus_one
stablehlo.log
stablehlo.log_plus_one
stablehlo.logistic
stablehlo.negate
stablehlo.rsqrt
stablehlo.sign
stablehlo.sine
stablehlo.sqrt
stablehlo.tanh

stablehlo.cholesky
stablehlo.triangular_solve
```

Other than these ops, we  have `fft`, `rng`, and `rng_bit_generator` (or something else which I might be missing) which could be  potential candidates for the support. I propose that we add the support after adding the specification of those op as adding the support might need some non-trivial discussion.

PR #1641, opened 2023-06-23
Add interface to allow VHLO ops to verify operand and result types

GleasonK:
https://github.com/openxla/stablehlo/pull/1496 made a subtle change to the StableHLO opset which caused a forward incompatibility. It was possible at HEAD to serialize a file for a target version that would then fail to deserialize on that target version. See `vhlo_to_version_downgrade_invalid.0_13_0.mlir` for repro file.

This is not an issue yet since this change is not a part of any tagged release, nor has it been exported to other openxla repositories yet. This PR fixes the forward incompat by introducing an op interface `VerifyOperandAndResultTypesInterface` which is called on ops during VhloToVersion legalization. This is useful in cases where an Ops structure has not changed (no new attrs, operands, or regions), but the supported operand and result types has changed.

This can also be used to maintain compatibility in https://github.com/openxla/stablehlo/pull/1608.

Using an interface for this came first to mind, to keep this validation tied to individual ops. If you have any better ideas please let me know, as I'm not tied to this design. In its current state it may get a little verbose as we apply it to more ops.


burmako:
If you're reading this and wondering how to keep track of all the compatibility RFCs (we have 3 accepted ones and 1 more is under review), [compatibility.md](https://github.com/openxla/stablehlo/blob/main/docs/compatibility.md) provides the source of truth on the current state of compatibility in StableHLO.

Specifically regarding this PR, in the "Out of scope" section, it says "Unspecced features: We may make incompatible changes to features which are not yet part of the StableHLO specification, e.g. we do not provide compatibility guarantees for unregistered attributes", i.e. pretty much what Kevin wrote above.

PR #1647, opened 2023-06-25
Specification of quantized ops from `Control-flow`, `Distribution`, `Extensibility`, and `Miscellaneous` categories. 

sdasgup3:
### Summary 

This PRs provides the specification of the ops from the following categories:

* **Control-flow ops**: `after_all, case, if, optimization_barrier, while`
  - `afer_all`: Not applicable for quantization as the ops deals with setting data-flow ordering using tokens. 
  - `case, if`: These ops  implicitly captures operands from the parent scope. The result of these operations and the corresponding blocks  can be quantized types. 
  - `optimization_barrier`: The `operand` and `result` can both be quantized tensors.
  - `while`: The `operand` and `result` can both be quantized tensors and of same type. Per the semantics of `while` op, the output of the associated block is used for the next iteration with the `operand` used for the very first iteration. With that,  I do not think we can use the `operand` and `result` with different quantization types. 
 
* **Distribution**: `all_gather, all_reduce, all_to_all, collective_permute, infeed, outfeed, partition_id, recv, reduce_scatter, replica_id, send`
All the ops in this category except for the following ones will support quantized types:
  - `all_reduce, reduce_scatter`:  These involve quantization of reduction operation https://github.com/openxla/stablehlo/pull/1538. I would propose to address these two after the quantization of reduction ops are addressed. 
  - `partition_id, replica_id`: Not applicable 

* **Extensibility**: `custom_call, get_tuple_element, tuple`
All the ops in this category supports all the `ValueType`s including `QuantizedTensorType`. No change is needed in the existing spec. 

* **Miscellaneous**: `batch_norm_grad, batch_norm_inference, batch_norm_training, cholesky, constant, fft, iota, rng, rng_bit_generator, triangular_solve`
  - The specification of `batch_norm_inference`, `cholesky`, `triangular_solve` ops follow the same `dequantize-op-quantize` strategy as we followed for the elementwise ops. 
  - `batch_norm_grad, batch_norm_training`:  These involve quantization of reduction operation https://github.com/openxla/stablehlo/pull/1538. I would propose to address these two after the quantization of reduction ops are addressed. 
  - `constant, iota`: The output type of these ops can be a quantized type of user-defined quantization parameters.
  -  `fft`: This op supports `element_type(operand) = element_type(result) = complex type`, when the `fft_type` is `FFT`  or `IFFT`. For `fft_type = RFFT`, the op supports  floating-point `operand` and complex type `result` and for `fft_type = IRFFT`, it supports complex type `operand` and floating-point `result`. IMO, only for `fft_type = RFFT`, the operand can support quantized type and the semantics could be to quantize the operand to floating-point before doing the computation. 
  - `rng`: The `operand` and `result` can be of quantized type only when the `rng_distribution` is `NORMAL` in which case  the semantics follows the `dequantize_op_quantize` strategy. The motivation for this choice comes from the fact the  op supports floating point type only when `rng_distribution` is `NORMAL` and we propose to treat this op, on floating-point tensors, as an op on quantized tensor, along the lines of dequantize -> float computation -> quantize. 
  - `rng_bit_generator`: To be discussed in "open questions" section  below. 

### Open questions
 -  `rng_bit_generator`: This op generates a randomized output tensor based on a `rng_algorithm` algorithm (`DEFAULT`, `THREE_FRY`, and `PHILOX`). It is not clear how there algorithm should handle generating quantized output. My proposal here is to skip quantization of this op unless we have a use case for it. Please let me know your opinion on this. 

Please review and let me know your feedback. 

**update 06/26/2023**
Scoping `rng` out from quantization based on https://github.com/openxla/stablehlo/pull/1647#issuecomment-1607857449

**update 07/13/2023**
Per https://github.com/openxla/stablehlo/pull/1647#discussion_r1251233481, I proposed not to fuse de-quantization of operand and quantization of result for `fft_type = RFFT` and `fft_type = IRFFT` respectively for fft op. With that a user can explicit de-quantize of operand or quantize of result as needed.

In the current proposal, the `batch_norm_inference` op, do not consider the constant operand `epsilon` to be of quantized type. IMO, considering that operand with quantized type unnecessary complicate the expressiveness of the op.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -517,7 +517,7 @@ have the following constraints:
 * (C2) `is_wellformed(imaginary_part, complex_element_type(complex_type))`.
 
 ```ebnf
-TensorConstant ::= TensorLiteral ':' ( TensorType | QuantizedTensorType )
+TensorConstant ::= TensorLiteral ':' TensorType
 TensorLiteral  ::= 'dense' '<' (DenseLiteral | ElementLiteral) '>'
 DenseLiteral   ::= DenseDimension | DenseElements
 DenseDimension ::= '[' [DenseLiteral {',' DenseLiteral}] ']'
@@ -533,10 +533,8 @@ represents a tensor value with the following mapping from indices to elements:
 implementation-defined. Tensor constants have the following constraints:
 
 * (C1) `has_syntax(tensor_literal, element_type(tensor_type))`, where:
-  * `has_syntax(element_literal: Syntax, tensor_element_type: Type) =
-    is_wellformed(element_literal, tensor_element_type)`.
-  * `has_syntax(element_literal: Syntax, quantized_tensor_element_type: Type) =
-    is_wellformed(element_literal, storage_type(quantized_tensor_element_type))`.
+  * `has_syntax(element_literal: Syntax, element_type: Type) =
+    is_wellformed(element_literal, type)`.
   * `has_syntax(tensor_literal: List, element_type: Type) =
     has_syntax(tensor_literal..., element_type)`.
 * (C2) `has_shape(tensor_literal, shape(tensor_type))`, where:
@@ -545,6 +543,18 @@ implementation-defined. Tensor constants have the following constraints:
     size(tensor_literal) = shape[0] and
     has_shape(tensor_literal..., shape[1:])`.
   * otherwise, `false`.
+
+```ebnf
+QuantizedTensorConstant ::= QuantizedTensorLiteral ':' QuantizedTensorType
+QuantizedTensorLiteral  ::= 'dense' '<' (DenseLiteral | ElementLiteral) '>'
+```
+
+**Quantized tensor constants** represent quantized tensor values using the same
+notation as tensor constants, with elements specified as constants of their
+storage type. Quantized tensor constants have the following constraints:
+
+* (C1) `has_syntax(quantized_tensor_literal, storage_type(quantized_tensor_type))`.
+* (C2) `has_shape(quantized_tensor_literal, shape(quantized_tensor_type))`.
 
 ```ebnf
 StringConstant  ::= StringLiteral
@@ -1420,9 +1430,9 @@ where:
 
 #### Outputs
 
-| Name      | Type                                                               | Constraints |
-|-----------|--------------------------------------------------------------------|-------------|
-| `results` | variadic number of tensors, per-tensor quantized tensors or tokens | (C4)        |
+| Name      | Type                                                    | Constraints |
+|-----------|---------------------------------------------------------|-------------|
+| `results` | variadic number of tensors, quantized tensors or tokens | (C4)        |
 
 #### Constraints
 
@@ -1871,14 +1881,13 @@ Produces an `output` tensor from a constant `value`.
 
 #### Outputs
 
-| Name     | Type                                  | Constraints |
-|----------|---------------------------------------|-------------|
-| `output` | tensor or per-tensor quantized tensor | (C1)        |
+| Name     | Type                       | Constraints |
+|----------|----------------------------|-------------|
+| `output` | tensor or quantized tensor | (C1)        |
 
 #### Constraints
 
-* (C1) `type(value) = is_quantized(output) ? storage_type(output) :
-       type(output)`.
+* (C1) `type(value) = type(output)`.
 
 #### Examples
 
@@ -2712,25 +2721,19 @@ for `fft_type = RFFT`. For example, for `L = 3`:
 * `result2[i0, ..., :, iR-1] = ifft(result1[i0, ..., :, iR-1])`.
 * `result[i0, ..., :] = irfft(result2[i0, ..., :])`.
 
-For quantized types, performs:
-
-* `fft(dequantize(operand), fft_type, fft_length)` if `fft_type = RFFT`.
-* `quantize(fft((operand, fft_type, fft_length), type(result)))` if `fft_type
-   = IRFFT`.
-
 #### Inputs
 
-| Label | Name         | Type                                                                    | Constraints            |
-|-------|--------------|-------------------------------------------------------------------------|------------------------|
-| (I1)  | `operand`    | tensor of floating-point or complex type or per-tensor quantized tensor | (C1), (C2), (C4), (C5) |
-| (I2)  | `fft_type`   | enum of `FFT`, `IFFT`, `RFFT`, and `IRFFT`                              | (C2), (C5)             |
-| (I3)  | `fft_length` | 1-dimensional tensor constant of type `si64`                            | (C1), (C3), (C4)       |
+| Label | Name         | Type                                         | Constraints            |
+|-------|--------------|----------------------------------------------|------------------------|
+| (I1)  | `operand`    | tensor of floating-point or complex type     | (C1), (C2), (C4), (C5) |
+| (I2)  | `fft_type`   | enum of `FFT`, `IFFT`, `RFFT`, and `IRFFT`   | (C2), (C5)             |
+| (I3)  | `fft_length` | 1-dimensional tensor constant of type `si64` | (C1), (C3), (C4)       |
 
 #### Outputs
 
-| Name     | Type                                                                    | Constraints      |
-|----------|-------------------------------------------------------------------------|------------------|
-| `result` | tensor of floating-point or complex type or per-tensor quantized tensor | (C2), (C4), (C5) |
+| Name     | Type                                     | Constraints      |
+|----------|------------------------------------------|------------------|
+| `result` | tensor of floating-point or complex type | (C2), (C4), (C5) |
 
 #### Constraints
 
@@ -2740,12 +2743,12 @@ For quantized types, performs:
     have the same complex type.
   * If `fft_type = IFFT`, `element_type(operand)` and `element_type(result)`
     have the same complex type.
-  * If `fft_type = RFFT`, (`is_float(operand)` or `is_quantized(operand)`) and
-    `is_complex(result)` such that `complex_element_type(element_type(result))
-    = is_quantized(operand) ? expressed_type(operand) : element_type(operand)`.
-  * If `fft_type = IRFFT`, `is_complex(operand)`and (`is_float(result)` or
-    `is_quantized(result)`) such that `complex_element_type(element_type(operand))
-    = is_quantized(result) ? expressed_type(result) : element_type(result)`.
+  * If `fft_type = RFFT`, `element_type(operand)` is a floating-point type and
+    `element_type(result)` is a complex type of the same floating-point
+    semantics.
+  * If `fft_type = IRFFT`, `element_type(operand)` is a complex type and
+    `element_type(result)` is a floating-point type of the same floating-point
+    semantics.
 * (C3) `1 <= size(fft_length) <= 3`.
 * (C4) If among `operand` and `result`, there is a tensor `real` of a
 floating-point type, then `shape(real)[-size(fft_length):] = fft_length`.
@@ -3010,9 +3013,9 @@ pred ? true_branch() : false_branch()`.
 
 #### Outputs
 
-| Name      | Type                                                               | Constraints |
-|-----------|--------------------------------------------------------------------|-------------|
-| `results` | variadic number of tensors, per-tensor quantized tensors or tokens | (C3)        |
+| Name      | Type                                                    | Constraints |
+|-----------|---------------------------------------------------------|-------------|
+| `results` | variadic number of tensors, quantized tensors or tokens | (C3)        |
 
 #### Constraints
 
@@ -3096,9 +3099,9 @@ separate outputs to improve clarity
 
 #### Outputs
 
-| Name      | Type                                                               | Constraints |
-|-----------|--------------------------------------------------------------------|-------------|
-| `results` | variadic number of tensors, per-tensor quantized tensors or tokens | (C1), (C2)  |
+| Name      | Type                                                    | Constraints |
+|-----------|---------------------------------------------------------|-------------|
+| `results` | variadic number of tensors, quantized tensors or tokens | (C1), (C2)  |
 
 #### Constraints
 
@@ -3671,11 +3674,11 @@ Semantics of `outfeed_config` is implementation-defined.
 
 #### Inputs
 
-| Label | Name             | Type                                                       |
-|-------|------------------|------------------------------------------------------------|
-| (I1)  | `inputs`         | variadic number of tensors or per-tensor quantized tensors |
-| (I2)  | `token`          | `token`                                                    |
-| (I3)  | `outfeed_config` | constant of type `string`                                  |
+| Label | Name             | Type                                            |
+|-------|------------------|-------------------------------------------------|
+| (I1)  | `inputs`         | variadic number of tensors or quantized tensors |
+| (I2)  | `token`          | `token`                                         |
+| (I3)  | `outfeed_config` | constant of type `string`                       |
 
 #### Outputs
 
@@ -3921,9 +3924,9 @@ separate outputs to improve clarity
 
 #### Outputs
 
-| Name      | Type                                                               | Constraints |
-|-----------|--------------------------------------------------------------------|-------------|
-| `results` | variadic number of tensors, per-tensor quantized tensors or tokens | (C2), (C3)  |
+| Name      | Type                                                    | Constraints |
+|-----------|---------------------------------------------------------|-------------|
+| `results` | variadic number of tensors, quantized tensors or tokens | (C2), (C3)  |
 
 #### Constraints
 
@@ -4954,13 +4957,13 @@ implementation-defined. This flag duplicates the information provided in
 
 #### Inputs
 
-| Label | Name               | Type                                                       | Constraints |
-|-------|--------------------|------------------------------------------------------------|-------------|
-| (I1)  | `inputs`           | variadic number of tensors or per-tensor quantized tensors |             |
-| (I2)  | `token`            | `token`                                                    |             |
-| (I3)  | `channel_id`       | constant of type `si64`                                    |             |
-| (I4)  | `channel_type`     | enum of `DEVICE_TO_DEVICE` and `DEVICE_TO_HOST`            | (C1)        |
-| (I5)  | `is_host_transfer` | constant of type `i1`                                      | (C1)        |
+| Label | Name               | Type                                            | Constraints |
+|-------|--------------------|-------------------------------------------------|-------------|
+| (I1)  | `inputs`           | variadic number of tensors or quantized tensors |             |
+| (I2)  | `token`            | `token`                                         |             |
+| (I3)  | `channel_id`       | constant of type `si64`                         |             |
+| (I4)  | `channel_type`     | enum of `DEVICE_TO_DEVICE` and `DEVICE_TO_HOST` | (C1)        |
+| (I5)  | `is_host_transfer` | constant of type `i1`                           | (C1)        |
 
 #### Outputs
 
@@ -5706,17 +5709,17 @@ The behavior of an infinite loop is TBD
 
 #### Inputs
 
-| Label | Name      | Type                                                               | Constraints |
-|-------|-----------|--------------------------------------------------------------------|-------------|
-| (I1)  | `operand` | variadic number of tensors, per-tensor quantized tensors or tokens | (C1-C3)     |
-| (I2)  | `cond`    | function                                                           | (C1)        |
-| (I3)  | `body`    | function                                                           | (C2)        |
+| Label | Name      | Type                                                    | Constraints |
+|-------|-----------|---------------------------------------------------------|-------------|
+| (I1)  | `operand` | variadic number of tensors, quantized tensors or tokens | (C1-C3)     |
+| (I2)  | `cond`    | function                                                | (C1)        |
+| (I3)  | `body`    | function                                                | (C2)        |
 
 #### Outputs
 
-| Name      | Type                                                               | Constraints |
-|-----------|--------------------------------------------------------------------|-------------|
-| `results` | variadic number of tensors, per-tensor quantized tensors or tokens | (C3)        |
+| Name      | Type                                                    | Constraints |
+|-----------|---------------------------------------------------------|-------------|
+| `results` | variadic number of tensors, quantized tensors or tokens | (C3)        |
 
 #### Constraints
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -3126,11 +3126,9 @@ separate outputs to improve clarity
 Fills an `output` tensor with values in increasing order starting from zero
 along the `iota_dimension` dimension. More formally,
 
-* For non-quantized case: `output[result_index] = constant(
-        result_index[iota_dimension], element_type(output))`.
-* For quantized case: `output[result_index] = constant(
-        quantize(result_index[iota_dimension], element_type(output)),
-        element_type(output))`.
+`output[result_index] = constant(is_quantized(output) ?
+quantize(result_index[iota_dimension], element_type(output)) :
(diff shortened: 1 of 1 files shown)


loganchien:
Actually, I would suggest scoping out `rng` for now.  Both `RoundToNearestEven` or `RoundAwayFromZero` will change the distribution of the random variables.  Typically, the correct usage of float-to-int conversion on random floats is to truncate them with `floor` (e.g. `floor(rng(...) * N) + M`).  Thus, it doesn't make sense to provide a quantized version.  If people really need them, they can still use `uniform_quantize(rng(...))`, albeit I think it will be an error.

For `rng_bit_generator`, I agree that we don't need quantized version since it is generating raw bits.

docs/spec.md
```diff
@@ -1853,9 +1861,9 @@ Produces an `output` tensor from a constant `value`.
...
 #### Outputs
 
-| Name     | Type   | Constraints |
-|----------|--------|-------------|
-| `output` | tensor | (C1)        |
+| Name     | Type                                  | Constraints |
+|----------|---------------------------------------|-------------|
+| `output` | tensor or per-tensor quantized tensor | (C1)        |
 
 #### Constraints
```
loganchien, line 1869: `stablehlo.constant` needs special definition for quantized constants.  The value actually holds a `DenseIntElementsAttr` with the storage type, thus the element type will be different.  And the semantic can't be defined as dequant_compute_quant, thus you must specify them here.

docs/spec.md
```diff
@@ -4383,25 +4394,29 @@ In conversations with many stakeholders, this op has come up as effectively
 deprecated, so in the future we are planning to explore removing it
 ([#597](https://github.com/openxla/stablehlo/issues/597)).
 
+For quantized types, performs
+`quantize(rng(dequantize(a), dequantize(b), shape, rng_distribution), type(result))`.
```
loganchien, line 4398: As discussed in the comment, I think it will be better to scope out `rng`, as the rounding for float-to-quant conversion may change the random variable distribution, which will make this unusable.

docs/spec.md
sdasgup3: Thanks @loganchien for the details.
IMO, those details will be pretty helpful in incorporating/implementing  quantized type in StableHLO `constant` op. 

 >> And the semantic can't be defined as `dequant_compute_quant`
I agree. 


>> you must specify them here

In the current PR, the semantics of this op is the input `value` of `constant` type to be interpreter with type as `quantized tensor`. I have updated the constraints of this op to reflect the type constraint. Also updated the syntax of Tensor constants to include constants with quantized type. PTAL.

Also, let me know if I am missing anything.

docs/spec.md
sdasgup3: Thanks this makes sense!
Will be fixed in the next commit.

docs/spec.md
loganchien: LGTM now.

loganchien (approved):

docs/spec.md
```diff
@@ -2677,13 +2688,16 @@ for `fft_type = RFFT`. For example, for `L = 3`:
 * `result2[i0, ..., :, iR-1] = ifft(result1[i0, ..., :, iR-1])`.
 * `result[i0, ..., :] = irfft(result2[i0, ..., :])`.
 
+For quantized types, performs
+`quantize(fft(dequantize(operand), fft_type, fft_length), type(result))`.
```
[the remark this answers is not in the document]
sdasgup3: Thanks @paulinesho for the observation. You are right that the `quantize` here does not makes sense given we do not support `result` to be of quantized type.

Upon thinking more on this topic, I would propose supporting  both quantized `operand` for RFFT and quantized `result` for IRFFT. I updated the PR with the necessary changes to achieve that.  PTAL.

paulinesho (approved):

docs/spec.md
```diff
@@ -1531,18 +1537,22 @@ strict lower triangle correspondingly, are implementation-defined.
 If there exists `i` where the input matrix is not an Hermitian positive-definite
 matrix, then the behavior is undefined.
 
+For quantized types, performs
+`quantize(cholesky(dequantize(a), lower), type(result))`.
```
burmako, line 1541: `quantize_op_dequantize(lambda operand: cholesky(operand, lower), operand, type(result))`. Let's centralize the quantize/dequantize logic in a single function. Same everywhere below.

docs/spec.md
```diff
@@ -1641,15 +1651,15 @@ Afterwards, `result@process` is given by:
 
 | Label | Name                  | Type                                         | Constraints |
 |-------|-----------------------|----------------------------------------------|-------------|
-| (I1)  | `operand`             | tensor                                       | (C5)        |
+| (I1)  | `operand`             | tensor or per-tensor quantized tensor        | (C5)        |
```
burmako, line 1654: Does `constant(0, element_type(result))` work correctly for the quantized case?

docs/spec.md
```diff
@@ -1862,13 +1872,14 @@ Produces an `output` tensor from a constant `value`.
...
-| `output` | tensor | (C1)        |
+| Name     | Type                                  | Constraints |
+|----------|---------------------------------------|-------------|
+| `output` | tensor or per-tensor quantized tensor | (C1)        |
 
 #### Constraints
 
-* (C1) `type(value) = type(output)`.
+* (C1) `type(value) = is_quantized(output) ? storage_type(output) :
+       type(output)`.
```
burmako, line 1882: I'm not sure I follow why this change is needed. Why wouldn't `type(value)` be equal to `type(output)` for the quantized case?

docs/spec.md
```diff
@@ -1415,9 +1421,9 @@ where:
 
 #### Outputs
 
-| Name      | Type                                 | Constraints |
-|-----------|--------------------------------------|-------------|
-| `results` | variadic number of tensors or tokens | (C4)        |
+| Name      | Type                                                               | Constraints |
+|-----------|--------------------------------------------------------------------|-------------|
+| `results` | variadic number of tensors, per-tensor quantized tensors or tokens | (C4)        |
```
burmako, line 1426: Any particular reason to not allow per-axis quantized tensors for control flow ops? 

Similarly to e.g. tokens, if there are StableHLO ops which take per-axis quantized tensors, then I think control flow ops should support them. Otherwise, we'd be able to express `op(foo)` but not `bar = cond ? foo1 : foo2; op(bar)` which I think would be unfortunate.

I think that the same logic applies to ConstantOp/InfeedOp/RecvOp which get values from the outside world and OutfeedOp/SendOp which put values into the outside world. If we allow per-axis quantized tensors in parameters and returns, I think we should allow them in the ops mentioned above.

docs/spec.md
```diff
@@ -3114,9 +3131,9 @@ constant(result_index[iota_dimension], element_type(output))`.
 
 #### Outputs
 
-| Name     | Type                                              | Constraints |
-|----------|---------------------------------------------------|-------------|
-| `output` | tensor of integer, floating-point or complex type | (C1)        |
+| Name     | Type                                                                             | Constraints |
+|----------|----------------------------------------------------------------------------------|-------------|
+| `output` | tensor of integer, floating-point or complex type or per-tensor quantized tensor | (C1)        |
```
burmako, line 3136: What does `constant(result_index[iota_dimension], element_type(output))` mean in the cast of quantized tensors? Does result_index[iota_dimension] stand for a value of storage type here or for a value of expressed type? Same for other uses of `constant(..., type)` in the spec where type can be a quantized tensor type.

docs/spec.md
```diff
@@ -1136,23 +1138,27 @@ def batch_norm_inference(operand, scale, offset, mean, variance, epsilon, featur
   return add(multiply(scale_bcast, normalized_operand), offset_bcast)
 ```
 
+For quantized types, performs
+`quantize(batch_norm_inference(dequantize(operand), scale, offset, mean,
+            variance, epsilon, feature_index), type(result))`.
```
burmako, line 1143: Hmm can't say that this definitions looks super satisfying. Why are we quantizing just the operand? Also, C2 doesn't seem to work under the current proposal.

docs/spec.md
```diff
@@ -2702,19 +2713,25 @@ for `fft_type = RFFT`. For example, for `L = 3`:
 * `result2[i0, ..., :, iR-1] = ifft(result1[i0, ..., :, iR-1])`.
 * `result[i0, ..., :] = irfft(result2[i0, ..., :])`.
 
+For quantized types, performs:
+
+* `fft(dequantize(operand), fft_type, fft_length)` if `fft_type = RFFT`.
+* `quantize(fft((operand, fft_type, fft_length), type(result)))` if `fft_type
+   = IRFFT`.
```
burmako, line 2720: Hmm I don't know. This looks sufficiently weird that I think we should just decline supporting quantized types in FftOp, similarly to the overall guideline that quantized types don't compose with complex types.

burmako (changes requested):

docs/spec.md
sdasgup3: When the `type(output)` is a quantized type,   `element_type(output)` provides the `QuantizedTensorElementType`.  The intention with `constant(result_index[iota_dimension], element_type(output))`  is to create the constant with  `QuantizedTensorElementType`.

sdasgup3: Added an [issue](https://github.com/openxla/stablehlo/issues/1687) to  address the usage of such  `constant` meta function on `QuantizedTensorElementType`.

docs/spec.md
```diff
@@ -517,7 +517,7 @@ have the following constraints:
 * (C2) `is_wellformed(imaginary_part, complex_element_type(complex_type))`.
 
 ```ebnf
-TensorConstant ::= TensorLiteral ':' TensorType
+TensorConstant ::= TensorLiteral ':' ( TensorType | QuantizedTensorType )
```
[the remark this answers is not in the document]

docs/spec.md
sdasgup3: You are right. I confused `value` with individual tensor elements. updated.

docs/spec.md
sdasgup3: I see you point. Let us not fuse de-quantization or quantization in this op. Updated the PR description with the revised proposal.

docs/spec.md

docs/spec.md
sdasgup3: Not considering the other operands for quantized type was an oversight. Fixed it. PTAL.

I note that:

Following how the `scale, offset, mean, variance` are used in the spec as part of `broadcast_in_dim` op and from the constraint `C`1` of [broadcast_in_dim](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#broadcast_in_dim), these other operands must be of quantized type if `operand` is of quantized type.  This also follows from the general strategy tahat have been employed so far while defining the quantized spec - all the operands with floating point types can potentially be considered with quantized type. 



Please note that the program constant operand `epsilon` is not consider to be of quantized type. IMO, considering that operand with quantized type unnecessary complicate the expressiveness of the op.

docs/spec.md
```diff
@@ -544,6 +544,18 @@ implementation-defined. Tensor constants have the following constraints:
     has_shape(tensor_literal..., shape[1:])`.
   * otherwise, `false`.
 
+```ebnf
+QuantizedTensorConstant ::= QuantizedTensorLiteral ':' QuantizedTensorType
```
burmako, line 548: This non-terminal also needed to be added to the `Constant ::= ...` rule.

docs/spec.md
burmako: What I meant is whether we should perhaps use `zero_point(result)` instead of `0` for the quantized case?

docs/spec.md
```diff
@@ -1530,22 +1548,26 @@ strict lower triangle correspondingly, are implementation-defined.
...
+For quantized types, performs
+`dequantize_op_quantize(lambda operand: cholesky(operand, lower), a, type(result))`.
+
 #### Inputs
 
 | Label | Name    | Type                                       | Constraints |
 |-------|---------|--------------------------------------------|-------------|
-| (I1)  | `a`     | tensor of floating-point or complex type   | (C1-C3)     |
+| (I1)  | `a`     | tensor of floating-point or complex type or per-tensor
+quantized tensor   | (C1-C3)     |
```
burmako, line 1559: I think we'll also want to align other rows of this table with the updated length of this row.

docs/spec.md
burmako: "The intention [...] is to create the constant with QuantizedTensorElementType". Sure, I understood that. What I didn't understand is whether the first argument of the `constant` function takes values of storage type or expressed type.

burmako (changes requested):

docs/spec.md
sdasgup3: I see your point. Thanks! 
I think that is the right choice as value 0 is not guaranteed to be represented in the the quantized value domain.

sdasgup3: Made similar changes for `imag` op.

docs/spec.md
sdasgup3: The first argument of `constant(result_index[iota_dimension], element_type(output))` takes value of storage type.  IMO, there was a confusion here on whether the second argument of `constant`  function  is the type of the first argument or the casted type of first argument. I was assuming the second one and was expecting the type of `constant` to match the element type of `QuantizedTensorType`. 

Upon further thinking, it seems  the second argument of `constant` is just the type of the first argument and it makes sense to use  `output[result_index] = constant(result_index[iota_dimension], storage_type(output))`. Also updated other places (`collective_permute`, `dot_general`, and `convolution`).

docs/spec.md
sdasgup3: done.

docs/spec.md

docs/spec.md
burmako: Let's revert to `constant(0, element_type(result))` and resolve as part of #1687. I think that there are multiple design options that need to be evaluated carefully, and we don't have to block this PR on that.

docs/spec.md
burmako: "bitcast_convert(constant(result_index[iota_dimension], storage_type(output)), element_type(output))". Let's move forward with whatever's currently written in the PR and discuss this separately in #1687.

burmako (approved):

docs/spec.md
sdasgup3: Done.
Sure.

docs/spec.md
```diff
@@ -1136,34 +1149,40 @@ def batch_norm_inference(operand, scale, offset, mean, variance, epsilon, featur
...
-| (I6)  | `epsilon`       | constant of type `f32`                      |               |
-| (I7)  | `feature_index` | constant of type `si64`                     | (C1), (C3-C6) |
+| Label | Name            | Type                                                                | Constraints   |
+|-------|-----------------|---------------------------------------------------------------------|---------------|
+| (I1)  | `operand`       | tensor of floating-point type or per-tensor quantized tensor        | (C1-C7)       |
+| (I2)  | `scale`         | 1-dimensional tensor of floating-point or per-tensor quantized type | (C2), (C3)    |
+| (I3)  | `offset`        | 1-dimensional tensor of floating-point or per-tensor quantized type | (C2), (C4)    |
+| (I4)  | `mean`          | 1-dimensional tensor of floating-point or per-tensor quantized type | (C5)          |
+| (I5)  | `variance`      | 1-dimensional tensor of floating-point or per-tensor quantized type | (C2), (C6)    |
+| (I6)  | `epsilon`       | constant of type `f32`                                              |               |
```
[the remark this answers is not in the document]
sdasgup3: That is a good point!
The reason epsilon is not provided in  quantized form is because it is a program constant and it would be non trivial to determine the quantization parameters (scale & zero point) for its quantized form.  For example, during the compilation time we do not know  the quantization parameters of the variance. So, the the quantization parameter to be selected for the epsilon would be arbitrary and the op semantics should involve re-scaling it to the quantization parameter of the variance. 

In any case, let me revisit point as part of https://github.com/openxla/stablehlo/issues/1404., I have added an item in that ticket.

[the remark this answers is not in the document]
loganchien: While I think it is fine to allow quantized type for `epsilon` (in other words, no objection from my side), it will be difficult to use.  The purpose of `epsilon` is to avoid division by zero when we calculate `1 / sqrt(variance + epsilon)`.  The model author must knowingly choose a constant value that won't be rounded to zero when `epsilon` is quantized.  Furthermore, the difference between the `scale`/`zero_point` of `variance` and `epsilon` should be small enough to prevent `epsilon` from being ignored.  And, the quantizer must ensure the quantized type of the `sqrt(variance + epsilon)` has enough resolution and won't round down the result value to zero.

In general, I question the feasibility of quantizing normalization-like operations by replacing the types of the floating point normalization implementations.  But that topic will be far beyond of the scope of this CL.  I am fine with either option: (1) submit as-is, (2) allow all operands to get quantized type (and leave the problem to the quantizer), or (3) scope out normalization-like op for further discussion.  All of these options are fine with me.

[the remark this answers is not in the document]
dominicsymes: Thanks for the comments. I think this is fine to be tracked as part of #1404 as it is integer only computation that will be affected by this. So fine with me to submit this and take further discussion in #1404.

docs/spec.md
```diff
@@ -3102,7 +3125,8 @@ separate outputs to improve clarity
 
 Fills an `output` tensor with values in increasing order starting from zero
 along the `iota_dimension` dimension. More formally, `output[result_index] =
-constant(result_index[iota_dimension], element_type(output))`.
+constant(result_index[iota_dimension], is_quantized(output) ?
+        storage_type(output) : element_type(output))`.
```
loganchien, line 3129: I think the specification here is different from what people (or quantizer) expect when we quantize a float tensor into a quantized tensor.

For example, a floating point iota:

```
    %a = "stablehlo.iota"() {iota_dimension = 1} : () -> tensor<2x8xf32>
```

Returns:

```
[[0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0],
 [0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0]]
```

If we quantize this float tensor with scale = 2, zp = 0 (with round to nearest even rounding), the storage-typed tensor should be:

```
[[0, 0, 1, 2, 2, 2, 3, 4],
 [0, 0, 1, 2, 2, 2, 3, 4]]
```

But if I interpret this line correctly, this is specifying:

```
[[0, 1, 2, 3, 4, 5, 6, 7],
 [0, 1, 2, 3, 4, 5, 6, 7]]
```

If this storage-typed tensor is dequantized, the value will be very different from the float tensor.

docs/spec.md
```diff
@@ -544,6 +545,18 @@ implementation-defined. Tensor constants have the following constraints:
     has_shape(tensor_literal..., shape[1:])`.
   * otherwise, `false`.
 
+```ebnf
+QuantizedTensorConstant ::= QuantizedTensorLiteral ':' QuantizedTensorType
```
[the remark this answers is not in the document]
sdasgup3: Thanks for the feedback! I agree that there is a  mismatch and I have added a [ticket](https://github.com/openxla/stablehlo/issues/1691) to address that.

docs/spec.md
sdasgup3: I completely agree with you that the quantized iota is just referring to the storage type and how that is going to be interpreted in the floating-point world depends on the choice (user-driven) of the quantization parameters. For example, with zp=0, scale=1, one can expect to get the desired correspondence in floating-point world.  

Having said that, IMO, we can keep the current specification as in the specification mainly because we made it clear in the spec that the stored value are of type `storage_type` . What do you think?

burmako (changes requested):

loganchien (approved):

docs/spec.md
sdasgup3: Update the ops `collective_permute`, `convolution`, `dot_general`, and `iota`. Now we have,

* `collective_permute`
    * Updated text: `broadcast_in_dim(constant(zero_point(result), element_type(result)), [], type(result))`
* `convolution`: 
    * Same as before: `padding_value = constant(is_quantized_tensor(lhs) ? zero_point(lhs) : 0, element_type(lhs))`.
* `dot_general`:
   * Updated text: `init_values=[constant(0, storage_type(result)]`
   * Rationale: The parent reduce op, which the `init_values` corresponds to, is doing integer computation. 
* `iota`:
  * Update test as suggested above.   

PTAL.

burmako: dot_general & iota look fine. Now that I looked at this again, I'm realizing that `reshaped_lhs_slice - zero_point(reshaped_lhs_slice)` doesn't typecheck, but we can leave resolving this to some other venue - this seems out of scope for this PR.

collective_permute - How would this work for non-quantized types?  `zero_point(result)` is going to return `None` in that case, which is probably not what we want.

convolution - This notation no longer works now that we have specified that QuantizedTensorConstant "has_syntax(quantized_tensor_literal, storage_type(quantized_tensor_type))". This is why I proposed writing things up explicitly for iota and why I chose to use `quantize` in that proposal.

burmako (changes requested):
See comment above

docs/spec.md
sdasgup3: >> collective_permute - How would this work for non-quantized types? zero_point(result) is going to return None in that case, which is probably not what we want.

That's my bad. fixed it. 

>> convolution - This notation no longer works

Just want to make sure I get it right. `constant(zero_point(lhs), element_type(lhs))` does not work because the type of `zero_point(lhs)` is not `element_type(lhs)`. To ensure the type parity your proposal is to explicitly express the quantize operation. 

With the above understanding, I have changed  
-* `padding_value = constant(is_quantized(lhs) ? zero_point(lhs) : 0, element_type(lhs))`.

to

+* `padding_value = constant(is_quantized(lhs) ? quantize(0, element_type(lhs)) : 0, element_type(lhs))`.

Please let me know if I am missing something.

docs/spec.md
```diff
@@ -1136,34 +1149,40 @@ def batch_norm_inference(operand, scale, offset, mean, variance, epsilon, featur
...
+| `result` | tensor of floating-point type or per-tensor quantized tensor | (C2), (C7)  |
 
 #### Constraints
 
 * (C1) `0 <= feature_index < rank(operand)`.
-* (C2) `operand`, `scale`, `offset`, `mean`, `variance` and `result` have the
-  same element type.
+* (C2) `baseline_element_type(operand) = baseline_element_type(scale) =
+       baseline_element_type(offset) = baseline_element_type(mean) =
+       baseline_element_type(variance) = baseline_type(result)`.
```
doyeonkim0, line 1180: nit: shouldn't this be baseline_element_type(result) instead of baseline_type(result)?

abattery (approved):
LGTM except one minor comment from Doyeon.

docs/spec.md
```diff
@@ -3101,8 +3124,13 @@ separate outputs to improve clarity
...
 Fills an `output` tensor with values in increasing order starting from zero
-along the `iota_dimension` dimension. More formally, `output[result_index] =
-constant(result_index[iota_dimension], element_type(output))`.
+along the `iota_dimension` dimension. More formally,
+
+* For non-quantized case: `output[result_index] = constant(
+        result_index[iota_dimension], element_type(output))`.
+* For quantized case: `output[result_index] = constant(
+        quantize(result_index[iota_dimension], element_type(output)),
+        element_type(output))`.
```
burmako, line 3133: Let's shorten this using the `is_quantized(output) ? quantize(foo, element_type(output)) : foo` notation from elsewhere in the spec. I like it better than my original proposal.

docs/spec.md
burmako: LGTM.

burmako (approved):
LGTM with the final minor comment

docs/spec.md

Issue #1530, opened 2023-05-23
Consider offering serialization/deserialization APIs as Pass Pipelines


GleasonK:
### Request description

Based on a discussion in the IREE discord server.

This would help StableHLO interoperate more seamlessly in the MLIR ecosystem. Need to investigate how to express targetVersion in a pass pipeline. Perhaps as a ModuleOp attribute, then the pipeline can take care of the rest.

GleasonK:
Sent PR for deserialization pipeline.

Thinking more about this - I'm wondering if we want to offer serialization as a pass pipeline, or only deserialization? Serialization requires writing bytecode, which requires an additional step beyond the pass pipeline, and we want portable artifacts to be easy to identify with specific header information, which is all provided by the serialization API.

PR #1651, opened 2023-06-28
Offer StableHLO portable artifact deserialization as a pass pipeline

GleasonK:
Some users want to write APIs that accept portable artifacts _and_ other programs. Currently adding support for portable artifacts requires additional logic to test if input is a portable artifact, then deserialize it before running any other passes. This is not ergonomic, as the more MLIR way of doing this is to offer a pass that will deserialize _if_ the input is a portable artifact - for example, support for StableHLO was added to some TF APIs using a `createHloLegalizeToStablehlo` pass.

To get to a point to permit this, we need two changes:

1. Don't error on non-VHLO dialects when deserializing, currently only happens in `--vhlo-to-version`.
This isn't a big issue, as `--stablehlo-legalize-to-vhlo` already validates that programs are stable. The current way VHLO to version is written is already more lenient than StableHLO->VHLO, as it only errors on Func and StableHLO ops, meaning, for example, that arith ops are permitted even without this change (would be caught by StableHLO->VHLO though).

2. Add deserialization pipeline. This allows users of StableHLO to add these passes to their input pipeline to quickly support StableHLO portable artifacts.

Part of https://github.com/openxla/stablehlo/issues/1530.

### DIFF
--- a/stablehlo/transforms/PassPipelines.cpp
+++ b/stablehlo/transforms/PassPipelines.cpp
@@ -19,7 +19,7 @@ limitations under the License.
 namespace mlir {
 namespace stablehlo {
 
-void createDeserializePortableArtifactPipeline(OpPassManager &pm) {
+void createStablehloDeserializePipeline(OpPassManager &pm) {
   // Convert VHLO(version x.y.z) --> VHLO(current).
   pm.addPass(stablehlo::createVhloToVersionPass(
       {vhlo::Version::getCurrentVersion().toString()}));
@@ -31,7 +31,7 @@ void createDeserializePortableArtifactPipeline(OpPassManager &pm) {
 void registerPassPipelines() {
   PassPipelineRegistration<>("stablehlo-deserialize",
                              "Run an example pipeline.",
-                             createDeserializePortableArtifactPipeline);
+                             createStablehloDeserializePipeline);
 }
 
 }  // namespace stablehlo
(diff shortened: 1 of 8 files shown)


stablehlo/transforms/PassPipelines.cpp
```diff
@@ -0,0 +1,38 @@
...
+      {vhlo::Version::getCurrentVersion().toString()}));
+
+  // Convert VHLO --> StableHLO. Will not fail within compatibility window.
+  pm.addPass(stablehlo::createVhloLegalizeToStablehloPass());
+}
+
+void registerPassPipelines() {
+  PassPipelineRegistration<>("stablehlo-deserialize",
+                             "Run an example pipeline.",
+                             createDeserializePortableArtifactPipeline);
```
burmako, line 34: Should this pipeline also take target version as an option, since it uses `--vhlo-to-version` which does?

burmako (changes requested):

stablehlo/transforms/PassPipelines.cpp
GleasonK: No, because deserialize always targets `getCurrentVersion` to convert back to StableHLO. If we decide to offer serialization as a pipeline then yes there should be an option.

burmako: Okay let's mention that in the comment for `createStablehloDeserializePipeline` and LGTM.

burmako (approved):

Issue #1578, opened 2023-06-05
StableHLO dialect is missing on some uses of DenseArray


joker-eph:
### What happened?

Seems like there is still quite a few ops using DenseElementsAttr instead of DenseArray, for example:

```
   %1 = "stablehlo.slice"(%arg0) {limit_indices = dense<[2, 1]> : tensor<2xi64>, start_indices = dense<0> : tensor<2xi64>, strides = dense<1> : tensor<2xi64>} : (tensor<2x2xui32>) -> tensor<2x1xui32>
```
   
 DenseArray construct is more efficient and offers better APIs in general.

(seems like this op is also missing on a custom synax)

joker-eph:
No I don't have any specific, I just saw I while trying to read some dumps from some models. I thought this was tracked already but I couldn't find a public issues, so maybe I was thinking about an internal bug.

PR #1657, opened 2023-06-29
Add custom print for stablehlo.slice

joker-eph:
Uses a syntax close to XLA HLO

Subset of #1578


stablehlo/dialect/AssemblyFormat.cpp
```diff
@@ -276,6 +278,73 @@ ParseResult parseDenseI64Array(OpAsmParser& parser,
...
+    llvm::interleaveComma(strides, p);
+    p << "]";
+    return;
+  }
+
+  llvm::interleaveComma(
+      llvm::zip(start_indices, limit_indices, strides), p,
+      [&](std::tuple<llvm::APInt, llvm::APInt, llvm::APInt> pack) {
+        auto [start, limit, stride] = pack;
+        p << std::get<0>(pack) << ":" << std::get<1>(pack);
```
burmako, line 304: You probably meant to use `start` and `limit` here, instead of `std::get<0>` and `std::get<1>`?

stablehlo/tests/ops_stablehlo_roundtrip.mlir
```diff
@@ -1,7 +1,4 @@
 // RUN: stablehlo-opt %s | stablehlo-opt
-// RUN: diff <(stablehlo-opt %s) <(stablehlo-opt -emit-bytecode %s | stablehlo-opt)
-// RUN: stablehlo-opt -emit-bytecode -debug-only=stablehlo-bytecode %s 2>&1 | (! grep 'Not Implemented')
-// RUN: stablehlo-opt -emit-bytecode %s | stablehlo-opt -debug-only=stablehlo-bytecode 2>&1 | (! grep 'Not Implemented')
```
burmako, line 4: Can you tell more about the motivating for disabling these RUN directives? cc @GleasonK for context since he added them here.

stablehlo/tests/stablehlo_canonicalize_dynamism.mlir
```diff
@@ -534,11 +522,7 @@ func.func @real_dynamic_slice_to_slice_success_static_result_type(%arg0: tensor<
 // CHECK-LABEL: func @real_dynamic_slice_to_slice_success_dynamic_result_type
 func.func @real_dynamic_slice_to_slice_success_dynamic_result_type(%arg0: tensor<4xf32>) -> tensor<?xf32> {
   //  CHECK-NOT: stablehlo.real_dynamic_slice
-  //      CHECK: "stablehlo.slice"(%arg0) {
-  // CHECK-SAME:   limit_indices = dense<1> : tensor<1xi64>,
-  // CHECK-SAME:   start_indices = dense<0> : tensor<1xi64>,
-  // CHECK-SAME:   strides = dense<2> : tensor<1xi64>
-  // CHECK-SAME: } : (tensor<4xf32>) -> tensor<?xf32>
+  //      CHECK: stablehlo.slice %arg0 [0:1:2]
```
burmako, line 525: I like the syntax - it matches Python and is close to HLO. Also cc @GleasonK since he may have had some thoughts about prettyprinting for slicing as well.

burmako (changes requested):
Thank you for the contribution! I have a few minor comments, and I wanted to also loop in Kevin since he looked into related areas in the past.

stablehlo/dialect/StablehloOps.td
```diff
@@ -1604,11 +1604,15 @@ def StableHLO_SliceOp: StableHLO_Op<
...
+    // Same in generic form
     %result = "stablehlo.slice" (%operand) {
-      start_indices = dense<[1, 2]> : tensor<2xi64>,
-      limit_indices = dense<[3, 4]> : tensor<2xi64>,
-      strides = dense<1> : tensor<2xi64>
-    } : (tensor<3x4xi64>) -> tensor<2x2xi64>
+      start_indices = dense<[1, 4]> : tensor<2xi64>,
+      limit_indices = dense<[3, 8]> : tensor<2xi64>,
+      strides = dense<[1, 2]> : tensor<2xi64>
+    } : (tensor<3x8xi64>) -> tensor<2x2xi64>
```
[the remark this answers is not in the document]
joker-eph: I don't really care about the generic syntax: to me it was a shortcut to explain the in-memory representation and avoid having to write a paragraph about it :)

stablehlo/dialect/StablehloOps.td
```diff
@@ -1604,11 +1604,15 @@ def StableHLO_SliceOp: StableHLO_Op<
 
     Example:
     ```mlir
+    %result = stablehlo.slice %operand [1:3, 4:8:1]
```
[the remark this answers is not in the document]
joker-eph: Indeed

stablehlo/dialect/AssemblyFormat.cpp
joker-eph: Indeed

stablehlo/dialect/AssemblyFormat.h
```diff
@@ -186,6 +186,17 @@ void printDenseI64Array(OpAsmPrinter& p, Operation* op,
...
+// SliceRanges - Used to print multi-dimensional ranges for slice.
+void printSliceRanges(OpAsmPrinter& p, Operation* op,
+                      DenseIntElementsAttr start_indices,
+                      DenseIntElementsAttr limit_indices,
+                      DenseIntElementsAttr strides);
+
+ParseResult parseSliceRanges(OpAsmParser& parser,
+                             DenseIntElementsAttr& start_indices,
+                             DenseIntElementsAttr& limit_indices,
+                             DenseIntElementsAttr& strides);
```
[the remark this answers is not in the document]
joker-eph: Of course!

stablehlo/tests/ops_stablehlo_roundtrip.mlir
joker-eph: My mistake: I was debugging something, restored now!

stablehlo/dialect/AssemblyFormat.cpp
```diff
@@ -276,6 +278,73 @@ ParseResult parseDenseI64Array(OpAsmParser& parser,
   return success();
 }
 
+void printSliceRanges(OpAsmPrinter& p, Operation* op,
+                      DenseIntElementsAttr start_indices,
+                      DenseIntElementsAttr limit_indices,
+                      DenseIntElementsAttr strides) {
```
burmako, line 284: This function still appears to be using the Google naming style for its parameters. (Yes, we should really set up an automatic check for this kind of stuff - we tried clang-tidy at some point and it turned out to be too slow, but that was before we got more powerful runners).

stablehlo/dialect/StablehloOps.td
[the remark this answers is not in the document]
burmako: Sounds good - I'm okay with breaking the convention here, since this is done for documentation reasons. Later on, we'll find a dedicated place to document the pretty syntax.

burmako (approved):
LGTM, but let's also ask Kevin to take a look.

GleasonK (approved):
LGTM. Could we also add a test to `tests/print_stablehlo.mlir` which is used to explicitly test custom printers, looks like our coverage of this printer comes only from canonicalize dynamism currently.

joker-eph:
> LGTM. Could we also add a test to `tests/print_stablehlo.mlir` which is used to explicitly test custom printers, looks like our coverage of this printer comes only from canonicalize dynamism currently.

Done, PTAL :)

GleasonK (approved):

GleasonK:
```
/home/runner/work/stablehlo/stablehlo/stablehlo/conversions/tosa/tests/unary.mlir:94:12: error: CHECK: expected string not found in input
 // CHECK: "stablehlo.slice"
```

Looks like the quotes on this check need to be removed

PR #1664, opened 2023-06-30
[RFC] Specification of quantized reduction operation

sdasgup3:
From the discussion emerged in https://github.com/openxla/stablehlo/pull/1538 about the specification of quantized `reduce` op, we zeroed in to a proposal which need some structural changes in the op (adding additional blocks). This proposed RFC is to further explore the proposal in detail.


rfcs/20230622-quantized-reduction.md
```diff
@@ -0,0 +1,521 @@
...
+index is computed as shown in the above diagram.
+
+Please refer to the [formal spec](#specification-of-reduce-op) of the proposed
+reduce op.
+
+### Implementation details
+
+From the implementation POV of the proposed spec, we note that
+`input_conversion` and `output_conversion` can very well be optional with
+default values as identity functions. For example, the following code snippet
```
loganchien, line 138: The example below is not identity function.  The identity function should be:

(Some `0` should be replaced with `input_zp`)

```
%result = "stablehlo.reduce"(%input, %init_value) ({
    ^input_conversion(
            %input: tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>):
        "stablehlo.return"(%input)
            : (tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>) -> ()
    }, {
    ^reduce_computation(
            %lhs: tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>,
            %rhs: tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>):
        %add = "stablehlo.add"(%lhs, %rhs)
            : (tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>,
               tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>)
            -> tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>
        "stablehlo.return"(%add)
            : (tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>) -> ()
    }, {
    ^output_conversion(
            %intermediate_result: tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>):
        "stablehlo.return"(%intermediate_result)
            : (tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>) -> ()
    }) {
        dimensions = dense<...> : tensor<1xi64>
    } : (tensor<... x !quant.uniform<ui8:f32, input_scale:input_zp>>,
         tensor<... x !quant.uniform<ui8:f32, input_scale:input_zp>>)
    -> tensor<... x !quant.uniform<ui8:f32, input_scale:input_zp>>
```

rfcs/20230622-quantized-reduction.md
```diff
@@ -0,0 +1,521 @@
...
+```
+
+Note that with default values, the  input/result type of `reduce` op matches
+with the argument or the result type of the `reduce_computation`, including the
+quantization parameters.
+
+It is important to enforce the following aspect in the implementations: both
+conversion functions must be provided or both must be missing. If this is not
+followed, and there are two functions provided in the IR, it would be difficult
+to determine which function is intended for which purpose.
```
loganchien, line 196: I don't understand this paragraph.  From the use case POV, it is reasonable to have on-the-fly input conversion but leave output in accumulated type.

If you are talking about MLIR ODS, we can use `VariadicRegion` and distinguish them with the order and signature.  The regions can be either `[input: (S)->T, reduce: (T, T)->T, output: (T)->V]`, `[reduce: (T, T)->T, output: (T)->V]`, `[input: (S->T), reduce: (T, T)->T]`, or `[reduce: (T, T)->T]`

rfcs/20230622-quantized-reduction.md
```diff
@@ -0,0 +1,521 @@
...
+The existing pretty printing is currently producing the following output
+`stablehlo.reduce(%input init: %init_value) applies stablehlo.add across
+dimensions = [1] : (tensor<1x6xi64>, tensor<i64>) -> tensor<1xi64>`. IMO,
+modifying the above format, with the default conversion function, will create
+clutter. My proposal here is to follow the existing pretty printing when the
+conversion functions are "not provided". In the event, the conversion functions
+are explicitly provided, then the pretty printers will fall back to default
+generic printing,
+**even if the explicitly provided conversion functions are identity function**:
+To avoid identification of identity functions which could be tricky in general.
```
loganchien, line 207: I think we can limit identity to:

```
^block(%arg: TYPE):
    return %arg : TYPE
```

and ignore other semantic identity function (e.g. `%arg + 0`).

rfcs/20230622-quantized-reduction.md
```diff
@@ -0,0 +1,521 @@
...
+%result = "stablehlo.reduce"(%input, %init_value) ({
+  ^input_conversion(%arg0: tensor<bf16>):
+    %0 = "stablehlo.convert"(%arg0): (tensor<bf16>) -> (tensor<f32>)
+    "stablehlo.return"(%0) : (tensor<f32>) -> (tensor<f32>)
+  }, {
+  ^bb0(%arg0: tensor<f32>, %arg1: tensor<f32>):
+    %0 = "stablehlo.add"(%arg0, %arg1) : (tensor<f32>, tensor<f32>) ->
+    tensor<f32>
+    "stablehlo.return"(%0) : (tensor<f32>) -> ()
+  },
```
loganchien, line 336: Missing a `{` at the end of line to start a region

rfcs/20230622-quantized-reduction.md
```diff
@@ -0,0 +1,521 @@
...
+
+### Should we restrict the proposal #1 to quantized types only?
+
+The above proposal #1 of introducing the additional functions is theoretically
+not limited to quantized `reduce` op, but also can be applied to `reduce` op with
+non-quantized types. For example,
+
+```mlir
+%result = "stablehlo.reduce"(%input, %init_value) ({
+  ^input_conversion(%arg0: tensor<bf16>):
```
loganchien, line 328:
```
// This is an implementation of f16 average reduction which is less likely to overflow.

%sum = "stablehlo.reduce"(%input, %init_value) ({
  ^input_conversion(%arg0: tensor<f16>):
    %0 = "stablehlo.convert"(%arg0): (tensor<f16>) -> (tensor<f32>)
    "stablehlo.return"(%0) : (tensor<f32>) -> (tensor<f32>)
  }, {
  ^bb0(%arg0: tensor<f32>, %arg1: tensor<f32>):
    %0 = "stablehlo.add"(%arg0, %arg1) : (tensor<f32>, tensor<f32>) ->
    tensor<f32>
    "stablehlo.return"(%0) : (tensor<f32>) -> ()
  }, {
  ^output_conversion(%arg0: tensor<f32>):
    %0 = "stablehlo.convert"(%arg0): (tensor<f32>) -> (tensor<f16>)
    "stablehlo.return"(%0) : (tensor<f16>) -> (tensor<f16>)
  }) {
  dimensions = dense<1> : tensor<1xf16>
} : (tensor<1x6xf16>, tensor<f16>) -> tensor<1xf32>

%avg_f32 = "stablehlo.divide"(%sum, %cst_6) : (tensor<f32>, tensor<f32>) -> tensor<f32>

%avg = "stablehlo.convert"(%avg_f32) : (tensor<f32>) -> tensor<f16> 
```

rfcs/20230622-quantized-reduction.md
sdasgup3: In all three are present then yes the order is the way to go. 

```
[reduce: (T, T)->T, output: (T)->V], 
[input: (S->T), reduce: (T, T)->T]
```
How to distinguish between the above two just based on types if `T = V  = S`? Also we cannot confuse one with the other as these  (input/output/reduce) functions can have different semantics.

My proposal here is   size(VariadicRegion) = 1 or 3. Let me know if I am missing anything.

rfcs/20230622-quantized-reduction.md
```diff
@@ -0,0 +1,521 @@
...
+  ^output_conversion(%arg0: tensor<f32>):
+    %0 = "stablehlo.convert"(%arg0): (tensor<f32>) -> (tensor<bf16>)
+    "stablehlo.return"(%0) : (tensor<bf16>) -> (tensor<bf16>)
+  }) {
+  dimensions = dense<1> : tensor<1xbf16>
+} : (tensor<1x6xbf16>, tensor<bf16>) -> tensor<1xbf16>
+```
+
+However, it is not clear how such operations will be lowered to other IR
+representations, like HLO, which does not support such additional computation
```
loganchien, line 346: That is the problem of HLO.  Either they decide not to support it and reject the op or extend the functionality of HLO.

The existing backend do 16-bit floating point work by either pattern matching, semantic/algorithmetic rewrite, use `bf16` to wish its range is large enough, or just guessing the intention of the programmer because we don't have a good representation.  I really hope we can include this case.

This design makes sense to CPU or GPU like architecture when they have bf16 to f32 extension internally.

rfcs/20230622-quantized-reduction.md
```diff
@@ -0,0 +1,521 @@
...
+
+### select_and_scatter
+
+This op originally takes two function arguments `select` and `scatter`. As the
+`select` function is supposed to perform a non-accumulative operation, we may
+not need additional conversion functions associated with `select`. But the
+`scatter` function needs be accompanied with `input_conversion` and
+`output_conversion` functions.
+
+```python
```
loganchien, line 484: Wrong line?

rfcs/20230622-quantized-reduction.md
sdasgup3: Maybe I did not get it. 
I was proposing that when the input_conversion or `output_conversion` is missing then their default value will behave as identity function `f(x) = x`. 

>> The example below is not identity function. The identity function should be:

I meant just the `input_conversion` and `output_conversion` are identity function in the above example.  Please let e know what I am missing.

rfcs/20230622-quantized-reduction.md
loganchien: reduce is binary, other are unary.

sdasgup3: That is right. Actually my confusion was how to deduce if a unary function is a input or output conversion function if only two functions are provided. 
I understand that relative order of a those function w.r.t the reduce block is the key here, given that reduce block is mandatory. Having said that I agree that we do not need this restriction. I will update.

rfcs/20230622-quantized-reduction.md
sdasgup3: done.

rfcs/20230622-quantized-reduction.md
sdasgup3: SG!

rfcs/20230622-quantized-reduction.md
sdasgup3: done.

rfcs/20230622-quantized-reduction.md
loganchien: I believe there were **typos** on Line 165, 169, 170, 172, 173, 174, 176, and 179.  All of the occurrences of `input_scale:0` should be replaced with `input_scale:input_zp`.

Otherwise, for example, Line 165 doesn't type check because `%input` is of type `tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>` but the specified type is `: (tensor<!quant.uniform<ui8:f32, input_scale:0>>`.

sdasgup3: fixed. Thanks!

rfcs/20230622-quantized-reduction.md
loganchien: I don't think this comment is addressed.  Why are you wrapping the following sections with python code block?

rfcs/20230622-quantized-reduction.md
```diff
@@ -0,0 +1,522 @@
...
+  sizes of `inputs...` corresponding to `dimensions` are not included.
+```
+
+The above specification of `reduce` op can be used to define the specification
+of other ops as shown below. For brevity, we are only presenting the relevant
+portions of the spec which needs modification.
+
+### reduce_window
+
+```python
```
loganchien, line 437: Why are you wrapping the following sections with python code block?

rfcs/20230622-quantized-reduction.md
```diff
@@ -0,0 +1,522 @@
...
+types.
+
+## Appendix
+
+To provide an estimate of specification changes needed to implement option #1
+I have attempted to provide the blueprint here.
+
+### Specification of reduce op
+
+```python
```
loganchien, line 360: Why are you wrapping the following sections with python code block?

rfcs/20230622-quantized-reduction.md
sdasgup3: Initially I thought about isolating the proposed spec of the ops in a code block for better visibility. Now it seems unnecessary and hence removed.

rfcs/20230622-quantized-reduction.md
sdasgup3: fixed.

rfcs/20230622-quantized-reduction.md
sdasgup3: noted.

rfcs/20230622-quantized-reduction.md
```diff
@@ -0,0 +1,605 @@
...
+
+The reduce syntax to be augmented with a optional [type
+attribute](https://github.com/llvm/llvm-project/blob/51a57074bc63842970c4c160b05c1a7e42db7523/mlir/include/mlir/IR/OpBase.td#L1466)
+as follows:
+
+```mlir
+%0 = stablehlo.reduce(%arg0 init: %arg1) across dimensions = [0] {
+    accumulation_type = tensor<!quant.uniform<i32:f32, 3.400000e+01:16>>
+} : (tensor<16x!quant.uniform<i8:f32, 3.400000e+01:16>>, tensor<!quant.uniform<i8:f32, 3.400000e+01:16>>) -> tensor<!quant.uniform<i8:f32, 3.400000e+01:16>>
+    reducer(%arg2: tensor<!quant.uniform<i32:f32, 3.400000e+01:16>>, %arg3: tensor<!quant.uniform<i32:f32, 3.400000e+01:16>>)  {
```
loganchien, line 552: IMO, the revised plan doesn't work.  Unlike `ConvolutionOp` or `DotGeneralOp` which only uses multiply-accumulation in their computation, `ReduceOp` and `ReduceWindowOp` allow users to specify the `computation`.  For example, the computation can be `ADD`, `ARGMAX`, etc.  In the ARGMAX example, we may mix `QUI8` (values) and `I32` (indices) in the `computation` region.  What does it mean when we specify `accumulation_type` attribute as proposed?

Second, what does it mean when user specify `accumulation_type` in this particular sum-reduction example?  Apparently, if we compile the code as-is, we will unnecessary round/truncate values when we compute 8-bit `stablehlo.add`.  It will be unfortunate if we require the backend to guess which part must be replaced by `accumulation_type` and which part must not (or even worse, making this representation unusable).

If the concern is about that `input_conversion` and `output_conversion` regions are too flexible, I would rather specify that:

> If (1) the input operand type is different from the `computation` block argument type or (2) the op result type is different from the `computation` block return type, there will be implicit type conversion defined by either `stablehlo.convert`, `stablehlo.uniform_quantize`, or `stablehlo.uniform_dequantize`.

sdasgup3: Hi Logan
Thanks for the feedback!

>> It will be unfortunate if we require the backend to guess which part must be replaced by accumulation_type and which part must not (or even worse, making this representation unusable).

The reducer block  
`reducer(%arg2: tensor<!quant.uniform<i32:f32, 3.400000e+01:16>>, %arg3: tensor<!quant.uniform<i32:f32, 3.400000e+01:16>>)` is using the accumulation-typed block arguments to do computations and will be different from the input value type. The type of the attribute `accumulation_type`  must match the block arguments type and block return type.

IMO, the attribute is used to uniformly treat all the reduction based operations (including convolution and dot_general) to explicitly specify a accumulation type. For convolution and dot_general as there is no explicut reduction block, the attribute will inform the backend about the type to accumulate on. For reduction operation, the backend can just use the attribute rather than doing an equality type comparison between block argument type and input  value type and using the block argument type if the comparison is false. 

Let me know what do you think.

loganchien: How does the `accumulation_type` attribute work if there are more than one operations in the `reduce_computation` block?  

How does the backend rewrite the `reduce_computation` block if they want to use `accumulation_type`?

For example, if someone writes a fancy sum-max-argmax reduction and wants to accumulate in `QI32` but keep other stuff unchanged, how do they write it?  How does the backend interpret or compile it?

```
%init_sum = "stablehlo.constant"() {
    value = dense<input_zp> : tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>}
    : () -> tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>
%init_max_value = "stablehlo.constant"() {
    value = dense<MIN_VALUE> : tensor<...>}
    : () -> tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>
%init_max_idx = "stablehlo.constant"() {
    value = dense<0> : tensor<i32>} : () -> tensor<i32>
%indices = "stablehlo.iota"() {dim = 0} : () -> tensor<... x i32>

%result:3 = "stablehlo.reduce"(%input, %input, %indices, %init_sum, %init_max_value, %init_max_idx) ({
    ^reduce(%sum_lhs, %max_lhs, %max_idx_lhs, %sum_rhs, %max_rhs, %max_idx_rhs):
        // sum-reduction
        %sum_value = "stablehlo.add"(%sum_lhs, %sum_rhs)
            : (tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>,
               tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>)
            -> tensor<!quant.uniform<ui8:f32, output_scale:output_zp>>

        // max-reduction and argmax-reduction
        %cmp = "stablehlo.compare"(%max_lhs, %max_rhs) {GE}
            : (tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>,
               tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>)
            -> tensor<i1>
        %max_value = "stablehlo.select"(%cmp, %max_lhs, %max_rhs)
            : (...)
            -> tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>
        %eq = "stablehlo.compare"(%max_lhs, %max_rhs) {EQ}
            : (tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>,
               tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>)
            -> tensor<i1>
        %first_idx = "stablehlo.minimum"(%max_idx_lhs, %max_idx_rhs)
            : (tensor<i32>, tensor<i32>) -> tensor<i32>>
        %max_idx = "stablehlo.select"(%cmp, %max_idx_lhs, %max_idx_rhs)
            : (tensor<i1>, tensor<i32>, tensor<i32>) -> tensor<i32>
        %new_idx = "stablehlo.select"(%eq, %first_idx, %max_idx)
            : (tensor<i1>, tensor<i32>, tensor<i32>) -> tensor<i32>
        "stablehlo.return"(%sum_value, %max_value, %new_idx)
            : (tensor<!quant.uniform<ui8:f32, output_scale:output_zp>>,
               tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>,
               tensor<i32>) -> ()
    }) {
      dimensions = dense<...> : tensor<1xi64>,
      accumulation_type = !quant.uniform<i32:f32, input_scale:0>
    } : (tensor<... x !quant.uniform<ui8:f32, input_scale:input_zp>>,
         tensor<... x !quant.uniform<ui8:f32, input_scale:input_zp>>,
         tensor<... x i32>,
         tensor<i32>)
    -> (tensor<... x !quant.uniform<ui8:f32, output_scale:output_zp>>,
        tensor<... x !quant.uniform<ui8:f32, input_scale:input_zp>>,
        tensor<... x i32>)
```

sdasgup3: Thanks for the elaborate example. That is helpful!

In the above code (A) the type of block arguments `sum_lhs` and `sum_rhs` has to be explicitly expressed as QI32 and (B) `sum_value` needs to be computed at QI32, and (C) the block return include the `sum_value` with QI32 type. I can see that not all block arguments need to be considered with higher accumulation type (like `max_lhs, max_idx_lhs, max_rhs, max_idx_rhs`) and agree that the attribute is not sufficient to convey such granular information.

Let me think a bit more on this this and get back.

rfcs/20230622-quantized-reduction.md
```diff
@@ -0,0 +1,372 @@
...
+ | (A) `stablehlo.uniorm_quantize`   | quantized tensor                       | quantized tensor                                    |
+ | (B) `stablehlo.uniorm_quantize`   | floating point                         | quantized tensor                                    |
+ | (C) `stablehlo.uniorm_dequantize` | quantized tensor                       | floating point                                      |
+ | (D) `stablehlo.convert`           | floating-point                         | integer                                             |
+ | (E) `stablehlo.convert`           | integer                                | floating-point                                      |
+ | (F) `stablehlo.convert`           | floating-point                         | floating-point                                      |
+ | (G) `stablehlo.convert`           | integer                                | integer                                             |
+ | (G) `stablehlo.convert`           | complex                                | complex                                             |
```
loganchien, line 89: (H)

rfcs/20230622-quantized-reduction.md
```diff
@@ -0,0 +1,372 @@
...
+
+`reduce_implicit_convert` is defined as
+
+```python
+def reduce_implicit_convert(x: Value, source_type: Type, destination_type:
+        Type):
+    if source_type == destination_type:
+        return x
+    if is_quantized(source_type) and is_quantized(destination_type):
+        return quantize(x, destination_type)
```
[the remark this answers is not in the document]
sdasgup3: Actually, `quantize` is the meta function we use in the spec (defined at https://github.com/openxla/stablehlo/blob/main/docs/spec.md#quantization-computations). Similarly we might need a meta function for `convert`.

rfcs/20230622-quantized-reduction.md
```diff
@@ -0,0 +1,372 @@
...
+
+The RFC introduces the following proposal, emerged out of discussion in the
+[thread](https://github.com/openxla/stablehlo/pull/1538#issuecomment-1599476906)
+, along with their tradeoffs.
+
+The proposal allows the reducer block to express the computation in a different
+element type (preferably higher accumulation type) than the one used in reduce
+op's ops arguments and return type. For illustrative purposes, in the following
+example, the operand element type `tensor<!quant.uniform<ui8:f32,
+    input_scale:input_zp>>`  is different from the element type for
```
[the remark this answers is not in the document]
sdasgup3: done

rfcs/20230622-quantized-reduction.md
```diff
@@ -0,0 +1,372 @@
...
+data types. However, with quantized data types, such loss in precision is not
+acceptable and hence the motivation is to perform the accumulation in some
+higher data type.
+
+The RFC introduces the following proposal, emerged out of discussion in the
+[thread](https://github.com/openxla/stablehlo/pull/1538#issuecomment-1599476906)
+, along with their tradeoffs.
+
+The proposal allows the reducer block to express the computation in a different
+element type (preferably higher accumulation type) than the one used in reduce
```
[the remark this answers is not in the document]
sdasgup3: done.

rfcs/20230622-quantized-reduction.md
```diff
@@ -0,0 +1,372 @@
...
+ | (C) `stablehlo.uniorm_dequantize` | quantized tensor                       | floating point                                      |
+ | (D) `stablehlo.convert`           | floating-point                         | integer                                             |
+ | (E) `stablehlo.convert`           | integer                                | floating-point                                      |
+ | (F) `stablehlo.convert`           | floating-point                         | floating-point                                      |
+ | (G) `stablehlo.convert`           | integer                                | integer                                             |
+ | (G) `stablehlo.convert`           | complex                                | complex                                             |
+
+At this point there is no use for cases other than (A), (F), and (G).  My
+proposal here would be to address (A), (F), and (G) only.  Note that the (F)
+    partially addresses [Decide on mixed
```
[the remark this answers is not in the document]
sdasgup3: fixed.

rfcs/20230622-quantized-reduction.md
sdasgup3: done

rfcs/20230622-quantized-reduction.md
```diff
@@ -0,0 +1,372 @@
...
+    }) {
+        dimensions = dense<1> : tensor<i64>
+    } : (tensor<5 x 1 x !quant.uniform<ui8:f32, input_scale:input_zp>>,
+         tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>)
+    -> tensor<5 x !quant.uniform<ui8:f32, output_scale:output_zp>>
+```
+
+### Semantics
+
+Depending on (1) the input operand type is different from the reduction block
```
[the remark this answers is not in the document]
sdasgup3: done.

rfcs/20230622-quantized-reduction.md
```diff
@@ -0,0 +1,372 @@
...
+`output_conversion` functions.
+
+#### Constraints
+
+<!-- markdownlint-disable line-length -->
+* (C1) `element_type(operand) = element_type(source)`.
+* (C3) `element_type(init_value) = element_type(operand)`.
+* (C?) `baseline_element_type(inputs...) = baseline_element_type(results...)`.
+* (C10) `scatter` has type `(tensor<E>, tensor<E>) -> tensor<E>` where
+       `is_integer(element_type(operand)) = is_integer(element_type(E]` or
```
loganchien, line 193: s/element_type(E/element_type(E)/

(note: missing `)`)

and below

rfcs/20230622-quantized-reduction.md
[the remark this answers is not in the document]
loganchien: still look the same (w/ weird indentation)

rfcs/20230622-quantized-reduction.md
sdasgup3: done.

rfcs/20230622-quantized-reduction.md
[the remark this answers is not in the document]
sdasgup3: I missed this one. PTAL.

rfcs/20230622-quantized-reduction.md
```diff
@@ -0,0 +1,372 @@
...
+
+If (1) the input operand type is different from the reduction block
+argument type or (2) the op result type is different from the reduction block
+return type, there will be implicit type conversion defined by either
+`stablehlo.convert`, `stablehlo.uniform_quantize`, or
+`stablehlo.uniform_dequantize`. For example,
+
+ | Implicit type conversion op       | element type of operand or result type | element type of block argument or block return type |
+ |-----------------------------------|----------------------------------------|-----------------------------------------------------|
+ | (A) `stablehlo.uniorm_quantize`   | quantized tensor                       | quantized tensor                                    |
```
sherhut, line 82: Nit: typo `uniform_quantize'. Also, of the result type is quantized but the block is float, the conversion would be `uniform_quantize'. The table seems to specify argument conversions.

rfcs/20230622-quantized-reduction.md
```diff
@@ -0,0 +1,372 @@
...
+
+### Semantics
+
+If (1) the input operand type is different from the reduction block
+argument type or (2) the op result type is different from the reduction block
+return type, there will be implicit type conversion defined by either
+`stablehlo.convert`, `stablehlo.uniform_quantize`, or
+`stablehlo.uniform_dequantize`. For example,
+
+ | Implicit type conversion op       | element type of operand or result type | element type of block argument or block return type |
```
sherhut, line 80: Nit: of

rfcs/20230622-quantized-reduction.md
```diff
@@ -0,0 +1,372 @@
...
+
+At this point there is no use for cases other than (A), (F), and (G).  My
+proposal here would be to address (A), (F), and (G) only.  Note that the (F)
+partially addresses
+[Decide on mixed precision](https://github.com/openxla/stablehlo/issues/369)
+for reduce op in that it allows the the input or init value to differ from the
+corresponding block arguments w.r.t the precision of floating-point types.
+However, the mixed precision implementation in HLO seems more detailed in the
+sense that even allows `inputs` and `init_values` to differ in floating-point
+precision. My proposal would be to treat the above ticket separately.
```
sherhut, line 99: Can you explain why this should be its own ticket? Is there a risk that we have to fundamentally revisit the encoding? It seems there is a natural extension to this case by using different type conversions. Is the question then whether this should be supported?
It would be useful to at least discuss whether the mixed precision usecase can be covered.

rfcs/20230622-quantized-reduction.md
```diff
@@ -0,0 +1,372 @@
...
+
+```python
+(C2) element_type(inputs...) = element_type(init_values...) = element_type(results...).
+(C6) body has type tensor<E0>, ..., tensor<EN-1>, tensor<E0>, ..., tensor<EN-1>) -> (tensor<E0>, ..., tensor<EN-1>) where Ei = element_type(inputs[i]).
+```
+
+which constrained the signature of reduce op and its associated reducer function
+`body` to have the same element types for `inputs`, `results` and arguments and
+return for `body`. For reducer function performing an accumulative operation
+like add, this means that the the result of accumulation can overflow in which
```
sherhut, line 31: Nit: the the

sdasgup3: done.

rfcs/20230622-quantized-reduction.md
sdasgup3: seems correct after my change. PTAL.

rfcs/20230622-quantized-reduction.md
sdasgup3: The current proposal does allows mixed precision for reduce op, but it solved the issue partially because there are other ops (than reduce) which need support for mixed precision [list](https://github.com/tensorflow/tensorflow/blob/1d69ba72834b963b72075a82c10959f6bb74e473/tensorflow/compiler/xla/service/hlo_verifier.cc#L1681-L1714). Having said that  my proposal is to address all those ops as part of resolving #369.

>> It would be useful to at least discuss whether the mixed precision usecase can be covered.
I agree with your point. I will update the doc with how the current proposal partially solves #369 for reduce op and the plan ahead to address #369.

cormacmb:
Detail:  today Reduce has the following constraint:

> (C3) 0 < size(inputs) = size(init_values) = size(results) = N.

Useful for the RFC to document out how this will change...

sdasgup3:
> Detail: today Reduce has the following constraint:
> 
> > (C3) 0 < size(inputs) = size(init_values) = size(results) = N.
> 
> Useful for the RFC to document out how this will change...

Hi Cormac
IMO, the above constraint should be unaffected with the proposal as the the number of input operand, number of init operands and number of results should remain the same. 

The current proposal allows operand or init value types to differ from that of reduce block arguments. In that light, the following constraint should get affected. 

```
(C6) body has type tensor<E0>, ..., tensor<EN-1>, tensor<E0>, ..., tensor<EN-1>) -> (tensor<E0>, ..., tensor<EN-1>) where Ei = element_type(inputs[i]).
```

I have proposed the changes to this constraint under `### Revised specification of reduce op`.

Please let me know if I am misinterpreting your comment.

cormacmb:
> > Detail: today Reduce has the following constraint:
> > > (C3) 0 < size(inputs) = size(init_values) = size(results) = N.
> > 
> > 
> > Useful for the RFC to document out how this will change...
> 
> Hi Cormac IMO, the above constraint should be unaffected with the proposal as the the number of input operand, number of init operands and number of results should remain the same.
> 
> The current proposal allows operand or init value types to differ from that of reduce block arguments. In that light, the following constraint should get affected.
> 
> ```
> (C6) body has type tensor<E0>, ..., tensor<EN-1>, tensor<E0>, ..., tensor<EN-1>) -> (tensor<E0>, ..., tensor<EN-1>) where Ei = element_type(inputs[i]).
> ```
> 
> I have proposed the changes to this constraint under `### Revised specification of reduce op`.
> 
> Please let me know if I am misinterpreting your comment.

sgtm

loganchien (approved):

rfcs/20230622-quantized-reduction.md
```diff
@@ -0,0 +1,384 @@
...
+    } : (tensor<5 x 1 x !quant.uniform<ui8:f32, input_scale:input_zp>>,
+         tensor<!quant.uniform<ui8:f32, input_scale:input_zp>>)
+    -> tensor<5 x !quant.uniform<ui8:f32, output_scale:output_zp>>
+```
+
+### Semantics
+
+If (1) the input operand type is different from the reduction block argument
+type or (2) the op result type is different from the reduction block return
+type, there will be implicit type conversion defined by either
```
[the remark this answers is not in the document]
sdasgup3: done

rfcs/20230622-quantized-reduction.md
```diff
@@ -0,0 +1,384 @@
...
+of other ops as shown below. As before, we are only presenting the relevant
+portions of the spec which needs modification.
+
+### Revised specification of  reduce_window op
+
+#### Constraints
+
+* (C?) `element_type(inputs...) = element_type(init_values...)`.
+* (C?) `baseline_element_type(inputs...) = baseline_element_type(results...)`.
+* (C?) `body` has type `tensor<E0>, ..., tensor<EN-1>, tensor<E0>, ...,`
```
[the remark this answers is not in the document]
sdasgup3: done

loganchien (approved):

rfcs/20230622-quantized-reduction.md
```diff
@@ -0,0 +1,384 @@
+# RFC: StableHLO quantization for reduction ops
+
+Status: Review<br/>
```
GleasonK, line 3: Per the discussion on OpenXLA Discuss, please change the status to approved and link to the OpenXLA discuss post at: https://groups.google.com/a/openxla.org/g/openxla-discuss/c/iwE9is49SS4

GleasonK (approved):
Per the discussion on [OpenXLA Discuss](https://groups.google.com/a/openxla.org/g/openxla-discuss/c/iwE9is49SS4) this RFC is now approved! Thanks to all who provided feedback on this design!

rfcs/20230622-quantized-reduction.md
GleasonK: I'll do this in a follow up PR actually

Issue #690, opened 2022-12-04
Allow specifying the return type for DynamicGatherOp


gnecula:
### Request description

When using the Python bindings for `stablehlo.DynamicGatherOp` the result has all dimensions unknown. This is likely because the `DynamicGatherOp` is [defined to support type inference](https://github.com/openxla/stablehlo/blob/444f6f269136a5c493713e77138a47b271839351/stablehlo/dialect/StablehloOps.td#L3057). Instead, there should be a way to pass in the result type, as for examples, for `stablehlo.DynamicReshapeOp` so that a front-end can set more precise return shapes.

For now my workaround is to generate a `stablehlo.BitcastConvertOp` following the `DynamicGatherOp` to set the more precise shape.

burmako:
Long-term fix will likely be done in #621, but for the short-term fix we might be able to come up with something different.

burmako:
"but for the short-term fix we might be able to come up with something different". I've taken the liberty to reformulate the ticket to improve type inference for DynamicGatherOp, because I think that a fix to this line will address the underlying issue: https://github.com/openxla/stablehlo/blob/e2aa7fe97cd09f44d864079c4e8be98064e5b425/stablehlo/dialect/TypeInference.cpp#L1677.

burmako:
I'm realizing that a more precise shape function for DynamicGatherOp won't satisfy the original request, because sometimes StableHLO producers have more shape information than shape functions can possibly infer. 

As mentioned above, because there is a shape function defined for DynamicGatherOp, its autogenerated Python bindings no longer include the result type in the signature of the `DynamicGatherOp.__init__` function. This is not a problem for precise shape functions, but not so much for imprecise shape functions like this one. If the producer knows a more precise type than what the shape function can infer, then `DynamicGatherOp.__init__` won't work for them, and they will need to use the much less convenient `DynamicGatherOp.build_generic`. Here's the workaround that was introduced in JAX ([click here for full code](https://github.com/google/jax/blob/ef4c2a38d65f4111b54f08f0c2e065174dde9b40/jax/_src/lax/slicing.py#L1403-L1415)):

```Python
  # TODO(burmako): Fix overly conservative type inference of DynamicGatherOp.
  # For now use the build_generic so that we can specify the result type.
  # return hlo.DynamicGatherOp(
  #     operand, indices, mlir.shape_tensor(slice_sizes),
  #     dnums, indices_are_sorted=ir.BoolAttr.get(indices_are_sorted)).results
[...]

PR #1668, opened 2023-07-05
Dynamism RFC

burmako:
The current design of dynamism in MHLO and StableHLO has been practically useful. There are success stories of it connecting JAX, PyTorch and TensorFlow to a number of compilers, in a mix of research and production environments. This RFC aims to leverage existing support for dynamism in the StableHLO dialect, discuss improvements to the existing design and then formalize the improved design in the StableHLO opset.

The main challenge with writing this RFC was that it affects the entire opset. The current design involves a lot of corner cases, so it took about a year of practical evaluation by the author - primarily within JAX native serialization, but also correlating with other ongoing and prior projects - to distill the design into just a few general design principles. I think I'm happy with the outcome, but please take a look at the "Summary" section for what that entails.

This RFC addresses a considerable chunk of community feedback on the existing design, but some feedback is deliberately left out of scope for this RFC to enable incremental progress while areas which require additional alignment are developed in parallel. See sections "Community feedback" and "Out of scope" for details.

The only open question at the moment is interoperability with HLO with respect to bounded dynamism. The proposed representation for bounded dynamic types in StableHLO is in parity with the representation for bounded dynamic types in HLO. However, I'm not sure whether this parity covers 100% of bounded dynamism functionality. For example: 1) there appears to be a mismatch in how broadcasts are represented, 2) there is misalignment in representations of dynamic windows (HLO has a high-level representation: VALID and SAME, whereas StableHLO expects the producers to dynamically compute window sizes). Nonetheless, I think that this shouldn't block the initial review of the RFC, since there's a lot of stuff to discuss - and in the meanwhile, I'll be working on confirming interoperability with HLO.

Finally, I'd like to acknowledge Smit Hinsu's work on the [Bounded Dynamism RFC](https://github.com/openxla/stablehlo/pull/194) from Q4 2022, which was superseded by this work. The representation for bounded dynamic types in the StableHLO dialect was designed and implemented by Smit, and Smit's proposal to allow bounded dynamic types everywhere is compatible with the more general proposal from this RFC to enable dynamism for all size-related program elements. Furthermore, Smit contributed the formal spec for get_dimension_size as well as the informal spec for set_dimension_size.


rfcs/20230704-dynamism-101.md
```diff
@@ -0,0 +1,825 @@
...
+
+**Baseline:** The MHLO and CHLO dialects have been initially co-designed with
+the TensorFlow dialect, the MLIR-based TF/XLA bridge and TensorFlow's
+[KernelGen toolchain](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/compiler/mlir/tools/kernel_gen).
+This has been inherited by the StableHLO repository when the StableHLO dialect
+was bootstrapped from MHLO and the CHLO dialect was moved to the StableHLO
+repository.
+
+As a result, there are 2 StableHLO ops (`compute_reshape_shape` and
+`cstr_reshapable`) as well as 4 CHLO ops (`dynamic_reshape`,
```
gnecula, line 261: It seems that JAX also uses `dynamic_reshape`.

burmako: JAX is using the `stablehlo.dynamic_reshape`, but this is talking about `chlo.dynamic_reshape`. The naming of this op is really confusing, all things considered.

rfcs/20230704-dynamism-101.md
```diff
@@ -0,0 +1,825 @@
...
+* [(P4) Enable dynamism for all size-related program elements but keep all axis-related program elements static](#p4).
+* [(P5) Unify static and dynamic versions of StableHLO operations](#p5).
+* [(P6) Represent shape computations as StableHLO operations working on 0-dimensional tensors](#p6).
+<!-- markdownlint-enable line-length -->
+
+These proposals address a considerable chunk of community feedback on the
+existing design, but some feedback is deliberately left out of scope for this
+RFC to enable incremental progress while areas which require additional
+alignment are developed in parallel. See sections "Community feedback" and
+"Out of scope" for details.
```
GleasonK, line 28: Could you link "community feedback" / "out of scope" similar to how you linked proposals?

rfcs/20230704-dynamism-101.md
```diff
@@ -0,0 +1,825 @@
...
+as StableHLO operations working on 0-dimensional tensors. More specifically,
+the proposal is to:
+
+1) Affirm that only shape computations that use StableHLO operations are
+   supported in StableHLO portable artifacts (this can be changed in future
+   RFCs, but is out of scope for this one).
+2) Use `Variadic<0DTensorOf<HLO_DimensionValue>>` instead of
+   `HLO_DimensionTensor` in the StableHLO dialect (this is consistent with
+   "variadic number of 0-dimensional tensors of integer type" from [P5](#p5)).
+3) Drop support for unranked dynamism in StableHLO (otherwise the operations
```
GleasonK, line 756: Is this a minor change in terms of impact? I'm thinking this should maybe be mentioned in the P6 header or elsewhere that is more visible to ensure it doesn't go unnoticed.

rfcs/20230704-dynamism-101.md
```diff
@@ -0,0 +1,825 @@
...
+// "vanilla" PadOp:
+// low, high and interior paddings are implemented via MLIR attributes,
+// i.e. they are static.
+%0 = stablehlo.pad %arg0, %arg1, low = [1, 1], high = [1, 1], interior = [0, 0]
+  : (tensor<16x16xf32>, tensor<f32>) -> tensor<20x20xf32>
+
+// DynamicPadOp:
+// low, high and interior paddings are implemented via MLIR operands,
+// i.e. they are dynamic.
+%1 = stablehlo.dynamic_pad %arg0, %arg1, %arg2, %arg3, %arg4 :
```
changm, line 101: nit: Would be clearer if you gave these args names like "low", "high", "interior" etc

rfcs/20230704-dynamism-101.md
```diff
@@ -0,0 +1,825 @@
...
+// See P5 for the "before" and "after" of unifying static and dynamic ops.
+// This snippet only shows the "after" for this section.
+def StableHLO_PadOp ... {
+  let arguments = (ins
+    HLO_Tensor:$operand,
+    HLO_Tensor:$padding_value,
+    I64ElementsAttr:$static_edge_padding_low,
+    I64ElementsAttr:$static_edge_padding_high,
+    I64ElementsAttr:$static_interior_padding,
+    Variadic<0DTensorOf<HLO_DimensionValue>>:$dynamic_edge_padding_low,
```
loganchien, line 769: I am concerned about the usage of variadic arguments.  It can make writing [Declarative Rewrite Rule (DRR)](https://mlir.llvm.org/docs/DeclarativeRewrites/) more difficult (if not impossible).  I am not sure if this will have some impact on the [PDL dialect](https://mlir.llvm.org/docs/Dialects/PDLOps/).

If we want to go through this approach, we should think about how we can assist declarative rewrite patterns.

----

Another way I can think of to unify static/dynamic shape, is to replace both of them with `1DTensorOf<HLO_DimensionValue>` and provide a utility function that converts variadic arguments into a Tensor:

```
// Static-shape
%static_edge_padding_low = "stablehlo.constant"() {value = dense<[...]> : tensor<4xi64>} : () -> tensor<4xi64>
%static_edge_padding_high = "stablehlo.constant"() {value = dense<[...]> : tensor<4xi64>} : () -> tensor<4xi64>
%static_interior_padding = "stablehlo.constant"() {value = dense<[...]> : tensor<4xi64>} : () -> tensor<4xi64>
%pad = "stablehlo.pad"(%operand, %pad_value, %static_edge_padding_low, %static_eduge_padding_high, %static_interior_padding)

// Dynamic-shape
%edge_padding_low = "stablehlo.make_tensor"(some, value, computed, dynamically) : (i64, i64, i64, i64) -> tensor<4xi64>
%edge_padding_high = "stablehlo.make_tensor"(some, value, computed, dynamically) : (i64, i64, i64, i64) -> tensor<4xi64>
%interior_padding = "stablehlo.make_tensor"(some, value, computed, dynamically) : (i64, i64, i64, i64) -> tensor<4xi64>
%pad = "stablehlo.pad"(%operand, %pad_value, %edge_padding_low, %eduge_padding_high, %interior_padding)
```

The benefit of this approach is that the static-shape version looks similar to the prior arts (e.g. TFLite).  Though, I am not sure if this approach has been discussed or not.

rfcs/20230704-dynamism-101.md
```diff
@@ -0,0 +1,825 @@
...
+like `add` are only defined when inputs and outputs have the same shape.
+For the static shape case, this can be checked **at compile time** (i.e. before
+the execution of the StableHLO program). However, when dynamism is involved,
+this can only be checked **at run time**.
+
+If some operation doesn't make sense at run time because some unknown ranks
+and/or unknown dimension sizes turned out to be incompatible with the operation
+and/or with each other, an error condition called **"shape mismatch"** occurs.
+In order to guard against shape mismatches, StableHLO programs may employ
+**shape checks**.
```
changm, line 116: I may be overindexing from my JavaScript days where they basically do the same thing.

This would require runtime support to bail out when a shape mismatch occurs. Are these shape checks inserted by the runtime as its own internal IR or does StableHLO have an op to represent shape checks? Or should front ends insert:

if (!Shape(op, expected)) return error() type logic?

rfcs/20230704-dynamism-101.md
```diff
@@ -0,0 +1,825 @@
...
+
+<a id="o3" /> **(O3)** Working out a common solution for shape checks would be
+nice. Per [F7](#f7), there are multiple incompatible approaches, so a common
+solution would result in a simplification of the overall stack.
+
+However, based on author's experience, alignment in this area requires
+non-trivial effort (different approaches have different benefits, plus they have
+limited interoperability), so this is left for future work.
+
+<a id="o4" /> **(O4)** Shape inference is an essential part of the StableHLO
```
changm, line 212: Does that mean that Shape Inference is provided by StableHLO? As in front ends shouldn't do their own?

rfcs/20230704-dynamism-101.md
changm: Nevermind, resolved via reading P2.

rfcs/20230704-dynamism-101.md
```diff
@@ -0,0 +1,825 @@
...
+changing the StableHLO dialect to implement the `ConditionallySpeculatable`
+interface, prohibiting speculation for StableHLO ops that involve dynamism
+([documentation](https://mlir.llvm.org/docs/Rationale/SideEffectsAndSpeculation/)).
+
+<!-- markdownlint-disable line-length -->
+## <a id="p3" /> (P3) Ratify the existing convention for relaxed constraints already implemented in the StableHLO dialect
+<!-- markdownlint-enable line-length -->
+
+**Baseline:** At the moment, the StableHLO dialect supports
+[relaxed constraints](https://github.com/openxla/stablehlo/blob/8993ad54839add6648b88801f1d223b7f9bc2e58/stablehlo/dialect/Base.cpp#L102-L120)
```
changm, line 318: Is this true even for unranked if unranked is probably going out of scope?

rfcs/20230704-dynamism-101.md
```diff
@@ -0,0 +1,825 @@
...
+). Note that this section is still using tensors to represent dynamic sizes,
+whereas MLIR upstream is using scalars. As mentioned in [F4](#f4), there's been
+community feedback that this design can be improved, and the next section
+proposes how an improved design could look like.
+
+```tablegen
+// Before: static version.
+def StableHLO_PadOp ... {
+  let arguments = (ins
+    HLO_Tensor:$operand,
```
changm, line 549: nit: I'm curious why these are named HLO_Tensor types and not StableHLO_Tensor?

rfcs/20230704-dynamism-101.md
burmako: Agreed - we have a series of tickets to improve prettyprinting for StableHLO, but they have a relatively lower priority than the work that's on the critical path for v1.0. https://github.com/openxla/stablehlo/blob/main/docs/roadmap.md goes into details about the overall roadmap.

rfcs/20230704-dynamism-101.md
burmako: Shape inference is implemented by the StableHLO dialect for the majority of the StableHLO ops (there are some ops where shape inference is infeasible, and some ops which have known issues). However, even with shape inference provided by libStablehlo, sometimes producers may want to specify their own result types. #690 is one example pertaining to dynamism, but there are other dynamism-, quantization- and sparsity-related examples as well. Overall, this is complicated topic, which is why I didn't include it in this RFC :)

rfcs/20230704-dynamism-101.md
burmako: Because they are shared with multiple *HLO dialects, including CHLO and MHLO.

rfcs/20230704-dynamism-101.md
burmako: Yes, unranked types are also part of these relaxed constraints in the current StableHLO implementation. I didn't call them out in the example below though to avoid overloading it, but they are in scope of the proposed design principle "If a constraint for a specific operation cannot be evaluated at compile time because it involves an unknown rank or an unknown dimension size, it gets deferred until the run time of the operation".

rfcs/20230704-dynamism-101.md
burmako: Done.

rfcs/20230704-dynamism-101.md
burmako: "This mentions the only use of this as TF, are we confident it isn't used externally?". The RFC mentions that we don't have 100% confidence: "the only user of unranked dynamism in StableHLO/MHLO appears to be TensorFlow", so we definitely need to validate this.

"Or is this a call to ask if any StableHLO/OpenXLA users support unranked dynamism?". Yes, exactly. The hope is that the uses that we haven't accounted for (if any) will come up during the review of the RFC. I've also updated the title of P6 to explicitly call out that the proposal is to drop support for unranked dynamism - that's a good idea.

rfcs/20230704-dynamism-101.md
changm: Ack that's OK, I just meant in this proposal in this actual text on line 101. Instead of arg0, arg1, etc, change line 101 to be:

%1 = stablehlo.dynamic_pad %tensor, %low, %high, etc

rfcs/20230704-dynamism-101.md
changm: Ahh ack, then super nit: Maybe rename to AnyHLO? Otherwise I would presume this links to basic HLO or whatever we call regular HLO now :)

rfcs/20230704-dynamism-101.md
burmako: Shape inference in StableHLO is part of the builder interface. When a programmer writes `builder.create<stablehlo::FooOp>(...)`, for most of the ops they can skip providing the result type and it will be inferred. For static shapes, shape inference works very well, so it is idiomatic to use it. For dynamic shapes, sometimes it's overly conservative, so the programmer may decide to provide the result type explicitly.

This is just a summary of what's going on, but there's more. Let's follow up on that offline since shape inference is out of scope for this RFC.

rfcs/20230704-dynamism-101.md
```diff
@@ -0,0 +1,826 @@
...
+Furthermore, different producers have different requirements and different
+preferences for expressing shape checks (e.g. JAX's type system enables it to
+need fewer checks than TensorFlow's type system), so a specific way of
+performing shape checks don't look like something that should be standardized
+at the StableHLO level either.
+
+B\) From the implementation standpoint, this proposal would involve updating
+[the "Errors" section](../docs/spec.md#errors) of the specification, and
+changing the StableHLO dialect to implement the `ConditionallySpeculatable`
+interface, prohibiting speculation for StableHLO ops that involve dynamism
```
yishuangP, line 311: Does it mean all StableHLO ops that could have dynamic input will implement `getSpeculatability`?

burmako: Yes, that's the plan. Fwiw if inputs and outputs of a specific operation are static, we can keep the existing behavior (always allow speculation), but if we have an operation with dynamic inputs, then this creates opportunities for UB, so we'll need to prohibit speculation per this proposal.

rfcs/20230704-dynamism-101.md
```diff
@@ -0,0 +1,826 @@
...
+  %3 = stablehlo.add %arg0, %arg1 : (tensor<?xf32>, tensor<1xf32>) -> tensor<1xf32>
+  %4 = stablehlo.add %arg1, %arg0 : (tensor<1xf32>, tensor<?xf32>) -> tensor<?xf32>
+  %5 = stablehlo.add %arg1, %arg0 : (tensor<1xf32>, tensor<?xf32>) -> tensor<1xf32>
+  %6 = stablehlo.add %arg1, %arg1 : (tensor<1xf32>, tensor<1xf32>) -> tensor<?xf32>
+  %7 = stablehlo.add %arg1, %arg1 : (tensor<1xf32>, tensor<1xf32>) -> tensor<1xf32>
+  func.return
+}
+```
+
+More formally, these relaxed constraints generalize the constraints that are
```
yishuangP, line 337: This means that for dynamic cases we should always check for shape mismatches in kernels at runtime?

burmako: Not really. P2 proposes that shape mismatches constitute undefined behavior at the StableHLO level, i.e. that StableHLO consumers are not obligated to perform shape checks in their kernel implementations (so these shape checks must happen elsewhere).

On a related note, per O3, defining a standard mechanism for shape checks is out of scope for this RFC. TF/XLA uses `shape.assuming` for these purposes, JAX is using `stablehlo.custom_call @shape_assertion` ([example](https://github.com/tensorflow/tensorflow/blob/40f01670c823de04fe8302fa9bc6987a376bbf4f/tensorflow/compiler/tests/xla_call_module_test.py#L464)), and we haven't discussed this from the PyTorch perspective yet.

So, overall, the shape checking story for StableHLO is in a bit of a chaos right now, and it will be desirable to figure it out in a follow-up RFC.

yishuangP: Got it. So far we don't see such shape check ops in the generated stablehlo module, probably because right now JAX only support shape polymorphism? If we are going to have these shape checking ops in the stablehlo module, we need to handle them in our converter.

burmako: @gnecula Can you tell more about about the current status and the future plans for shape checks in JAX lowerings?

gnecula: In the first phase, JAX will use dynamic shapes only when the intermediate shapes depend only on the **shapes** of the inputs, not on the value of the inputs (we call this shape polymorphism). JAX will try to do all the shape checking symbolically at lowering time, but when it is not possible it will insert `custom call shape_assertions`, which are expected to be discharged and removed once the input shapes are known.

In a latter phase JAX plans to support more dynamism, but we have not yet decided whether we'll need runtime shape assertions. Probably we will start by rejecting programs that we cannot verify statically to be well-shaped.

rfcs/20230704-dynamism-101.md
```diff
@@ -0,0 +1,826 @@
...
+([example](https://github.com/tensorflow/mlir-hlo/blob/1fd13ef28d3f423363e59b1a80eb2c26e0c0979d/mhlo/transforms/chlo_legalize_to_hlo/chlo_legalize_to_hlo.cc#L1577-L1587)).
+
+**Proposal:** The RFC proposes to standardize on representing shape computations
+as StableHLO operations working 0-dimensional tensors. More specifically,
+the proposal is to:
+
+1) Affirm that only shape computations that use StableHLO operations are
+   supported in StableHLO portable artifacts (this can be changed in future
+   RFCs, but is out of scope for this one).
+2) Use `Variadic<0DTensorOf<HLO_DimensionValue>>` instead of
```
ghpvnist, line 754: Could you elaborate on using `HLO_DimensionValue` over e.g. `HLO_Int`?  Is the rationale for using `HLO_DimensionValue` that other dialects convert to `Index` instead of integer element type?

Following the proposal in [P5](#p5) ("variadic number of 0-dimensional tensors of integer type") verbatim, I'd think the proposal would use `Variadic<0DTensorOf<HLO_Int>>` instead. I'm asking since the StableHLO spec currently does not define `index` element type even though the StableHLO dialect uses it.

sdasgup3: >> Could you elaborate on using HLO_DimensionValue over e.g. HLO_Int? Is the rationale for using HLO_DimensionValue that other dialects convert to Index instead of integer element type?

That is true. For example, Shape [dialect](https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Dialect/Shape/IR/ShapeOps.td) replies heavily on index.

rfcs/20230704-dynamism-101.md
```diff
@@ -0,0 +1,826 @@
...
+  %6 = stablehlo.add %arg1, %arg1 : (tensor<1xf32>, tensor<1xf32>) -> tensor<?xf32>
+  %7 = stablehlo.add %arg1, %arg1 : (tensor<1xf32>, tensor<1xf32>) -> tensor<1xf32>
+  func.return
+}
+```
+
+More formally, these relaxed constraints generalize the constraints that are
+already documented in the StableHLO specification. If a constraint for a
+specific operation cannot be evaluated at compile time because it involves an
+unknown rank or an unknown dimension size, it gets deferred until the run time
```
GleasonK, line 340: Would be good to clarify what "cannot be evaluated" means.

If there exists some value of `?` where a constraint could succeed, I understand deferring to runtime. I'm unsure about how we should handle cases where no value of `?` would permit verification to succeed, or if this is even feasible.

GleasonK: For an simple example of the above comment, consider the following. StableHLO ConcatenateOp requires that all non-concatenated dimensions be equivalent, making `[2, 1] . [2, 5]` with concat `dim = 1` valid since non-concatenated dims are both `2`.

Enter dynamism:

```
%0 = stablehlo.concatenate %arg0, %arg1, dim = 1 : (tensor<?x1xf32>, tensor<2x1xf32>) -> tensor<?x2xf32>
%1 = stablehlo.concatenate %arg0, %arg1, %arg2, dim = 1 : (tensor<?x1xf32>, tensor<2x1xf32>, tensor<3x1xf32>) -> tensor<?x3xf32>
```

- For `%0`, this is valid if `? == 2`
- For `%1`, no value of `?` satisfies `? == 2 == 3`

Thinking more about my earlier comment, while erroring on `? == 2 == 3` is certainly a more precise representation, I'm not sure _how useful_ this error is in practice..

I see two use cases for dynamically shaped programs:
1. Backend can evaluate dynamic shapes (mostly interpreters).
2. Backend refines shapes to run static program (compiler).

Users of (1), will need to implement identical verification regardless (in this case concat dimension size verification) to ensure during runtime that the provided value is valid.

Users in (2) will get static verification after refinement.

The use case I can imagine being interested in stricter verification would be serializing/shipping dynamic unrefined models, as (some of) these models wouldn't error until they attempt to be executed by a consumer instead of as they are serialized by the producer. Note that even with strict verification it is still possible to refine a program using invalid values and still fail verification, the only difference in strict verification is that "some value of `?` exists which can satisfy a constraint".

rfcs/20230704-dynamism-101.md
burmako: "I'd think the proposal would use Variadic<0DTensorOf<HLO_Int>> instead". You're right. Using `HLO_DimensionValue` was an oversight on my part, this should be `HLO_Int`.

"Is the rationale for using HLO_DimensionValue that other dialects convert to Index instead of integer element type?". Yes, interoperability with other dialects was a factor in the design process, but in the end I decided to keep it out of scope of this RFC, and O2 goes into details about that. tl;dr is that switching to `index` would be a significant amount of work, which I think is best left for a follow-up RFC (this one is already huge).

rfcs/20230704-dynamism-101.md
burmako: "Would be good to clarify what "cannot be evaluated" means". This RFC is using the meaning of evaluation that was informally established in https://github.com/openxla/stablehlo/blob/main/docs/spec.md#formulas.

In a nutshell, evaluation of a constraint takes a constraint written in the StableHLO specification formula language, resolves names in the constraint using the rules from the "Names" section, applies some peculiar semantics described in the "Formulas" section (e.g. ellipses) and produces one of the supported values described in the "Values" section.

I mentioned above that this notion of evaluation hasn't been formalized. When writing up the relevant parts of the StableHLO spec, I was thinking that we can assume that the readers would be familiar with mainstream programming languages like Python, which would allow them to extend their intuition about evaluation in these languages to formulas used in the StableHLO specification.

To make things very precise, we could formalize the notion of evaluation - both for StableHLO formulas and StableHLO programs - using one of the frameworks popular in theoretical works on programming languages. E.g. defining small-step operational semantics could be one way to approach this. On a related note, we actually have a ticket for formalizing evaluation of StableHLO programs (#484), but I haven't yet gottten around to doing that.

"I'm unsure about how we should handle cases where no value of ? would permit verification to succeed, or if this is even feasible". I think this is an interesting idea. I like that it would allow us to reject more invalid StableHLO programs at compile time. 

However, what's unclear to me is how to frame the formula evaluation rules to accommodate this proposal. That would probably mean that Python/C++/etc-style evaluation semantics is out of the question, and we would need some form of solvers?

rfcs/20230704-dynamism-101.md
```diff
@@ -0,0 +1,826 @@
...
+improved design in the StableHLO opset. To that end, this document proposes
+the following:
+
+<!-- markdownlint-disable line-length -->
+* [(P1) Move TensorFlow-specific operations out of StableHLO](#p1).
+* [(P2) Ratify the existing convention for shape mismatches constituting undefined behavior](#p2).
+* [(P3) Ratify the existing convention for relaxed constraints already implemented in the StableHLO dialect](#p3).
+* [(P4) Enable dynamism for all size-related program elements but keep all axis-related program elements static](#p4).
+* [(P5) Unify static and dynamic versions of StableHLO operations](#p5).
+* [(P6) Represent shape computations as StableHLO operations on 0-dimensional tensors and drop support for unranked dynamism](#p6).
```
GleasonK, line 21: After further discussion, P6 has been removed from this RFC, it may still be proposed in a future proposal, but some confidence should be built in the usability of this approach prior to making it the de-facto means for dynamic ops. I.e. this structure of op may be difficult to write declarative rewriter patterns for.

GleasonK: cc @loganchien

loganchien: Is the RFC doc updated?

PR #1669, opened 2023-07-05
Elaborate on the semantics of tokens

### LINKED ISSUE #1577: Elaborate on the semantics of tokens
The current specification of token types says "Tokens are used for imposing execution order on operations as described in the Execution section", but the Execution section doesn't actually describe this.

Also, token-producing and token-consuming operations don't go into much detail either, e.g. this is what AfterAllOp says: "Ensures that the operations producing the `inputs` are executed before any operations that depend on `result`. Execution of this operation does nothing, it only exists to establish data dependencies from `result` to `inputs`".

ghpvnist:
closes #1577

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -1298,14 +1298,14 @@ implementation-defined as well.
 
 #### Inputs
 
-| Label | Name      | Type                        | Constraints |
-|-------|-----------|-----------------------------|-------------|
+| Label | Name      | Type                       | Constraints |
+|-------|-----------|----------------------------|-------------|
 | (I1)  | `operand` | tensor or quantized tensor | (C1-C2)     |
 
 #### Outputs
 
-| Name     | Type                        | Constraints |
-|----------|-----------------------------|-------------|
+| Name     | Type                       | Constraints |
+|----------|----------------------------|-------------|
 | `result` | tensor or quantized tensor | (C1-C2)     |
 
 #### Constraints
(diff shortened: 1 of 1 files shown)


docs/spec.md
```diff
@@ -5789,10 +5781,12 @@ A StableHLO program is executed by providing input values to the `main` function
...
 executing the graph of ops rooted in the corresponding `return` op.
 
-The execution order is implementation-defined, as long as ops are executed
-before their uses. Possible execution orders of the example program above are
-`%0` → `%1` → `%2` → `%3` → `%4` → `return` or `%3` → `%0` → `%1` → `%2` → `%4`
-→ `return`.
+The execution order is implementation-defined as long as it is aligned with
+dataflow, i.e. if ops are executed before their uses. In StableHLO, all
+side-effecting ops consume and/or produce token(s) to reify the side effects as
+a value that other operations can take a data dependency on. Possible execution
```
burmako, line 5787: I'd like to propose a reformulation: "In StableHLO, all side-effecting ops consume a token and produce a token (multiple tokens can be multiplexed via `after_all`), so the execution order of side effects is also aligned with dataflow":
  1) These ops don't just consume and/or produce an unspecified number of tokens. Each such op consumes exactly one and produces exactly one token.
  2) The second part of rewording explicitly focuses on the execution order since this is what this paragraph is talking about.

burmako (changes requested):

docs/spec.md
ghpvnist: 1. Thanks, I originally misinterpreted "variadic number of tensors or tokens" in Case, If, and While op.
2. Sounds good to me.

burmako (approved):

Issue #1563, opened 2023-05-31
Add CI for compatibility version bump


ghpvnist:
As discussed in https://github.com/openxla/stablehlo/pull/1559#issuecomment-1570479355, it would be nice to have a CI to ensure that we include a bytecode file for each minor release. Either adding a `lint_compatibility.sh` to get the current version `X.Y.Z` from `Version.h` and ensure that test/bytecode files like `stablehlo_legalize_to_vhlo.X_Y_0.mlir` and `stablehlo_legalize_to_vhlo.X_Y_0.mlir.bc` exist.

PR #1670, opened 2023-07-05
Add lint to ensure that compatibility tests are implemented for new versions of StableHLO

GleasonK:
This will fail if we bump the version to `Version(0, 15, 0)` without adding new compatibility tests.

Fixes https://github.com/openxla/stablehlo/issues/1563

### DIFF
--- a/build_tools/github_actions/lint_version.sh
+++ b/build_tools/github_actions/lint_version.sh
@@ -1,4 +1,4 @@
-# Copyright 2022 The StableHLO Authors.
+# Copyright 2023 The StableHLO Authors.
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
@@ -12,8 +12,9 @@
 # limitations under the License.
 
 # This lint validates that there are compatibility tests for the current
-# version by extracting the current version from Version.h and checking if
-# files named stablehlo_legalize_to_vhlo.VERSION.mlir and .mlir.bc exist.
+# version by extracting the current version vX.Y.Z from Version.h and
+# checking if files named stablehlo_legalize_to_vhlo.X_Y_0.mlir and .mlir.bc
+# exist.
 
 ## Setup VERSION variable as global:
 VERSION_H="stablehlo/dialect/Version.h"
--- a/build_tools/github_actions/lint_version.sh
+++ b/build_tools/github_actions/lint_version.sh
@@ -1,0 +1,57 @@
+# Copyright 2022 The StableHLO Authors.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


build_tools/github_actions/lint_version.sh
```diff
@@ -0,0 +1,57 @@
+# Copyright 2022 The StableHLO Authors.
```
ghpvnist, line 1:
```
# Copyright 2023 The StableHLO Authors.
```

build_tools/github_actions/lint_version.sh
```diff
@@ -0,0 +1,57 @@
...
+COMPAT_TEST_BASE="stablehlo/tests/stablehlo_legalize_to_vhlo"
+COMPAT_TEST="$COMPAT_TEST_BASE.mlir"
+TEST_VERSION="${VERSION[0]}_${VERSION[1]}_0"
+VERSIONED_COMPAT_TEST="$COMPAT_TEST_BASE.$TEST_VERSION.mlir"
+VERSIONED_COMPAT_TEST_BC="$COMPAT_TEST_BASE.$TEST_VERSION.mlir.bc"
+
+show_help() {
+  HELP_URL="https://github.com/openxla/stablehlo/blob/main/docs/vhlo.md#add-versioned-serialization-test"
+  echo "For details on creating versioned tests for a new minor version of"
+  echo "StableHLO, see the instructions on:"
```
ghpvnist, line 43: Another nit: space after `:`
```
echo "StableHLO, see the instructions on: "
```

build_tools/github_actions/lint_version.sh
```diff
@@ -0,0 +1,57 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This lint validates that there are compatibility tests for the current
+# version by extracting the current version from Version.h and checking if
+# files named stablehlo_legalize_to_vhlo.VERSION.mlir and .mlir.bc exist.
```
ghpvnist, line 16: Nit: Since we are not replacing the variable `VERSION` verbatim, what about something like:
```
# files named stablehlo_legalize_to_vhlo.<X_Y_Z>.mlir and .mlir.bc exist.
```

ghpvnist (approved):
Minor comments, but lgtm overall.

build_tools/github_actions/lint_version.sh
GleasonK: good catch

build_tools/github_actions/lint_version.sh
GleasonK: echo adds a newline, so should be ok without

build_tools/github_actions/lint_version.sh
GleasonK: Should add that copyright lint one of these days..

PR #1671, opened 2023-07-06
Fix spelling errors in .cpp and .h files

GleasonK:
Used `cspell` with a user dictionary. Will upload a lint script for this shortly.

Fixes:
- bitwidth -> bit width. With a space seems far more common. Even LLVM methods use `BitWidth`.
- dnums -> dimNums. Could go either way on this one, but seems like a more descriptive name that passes spell check is fine.
- The rest are typos or improvements to camelcase


ghpvnist (approved):
Nice findings!

PR #1674, opened 2023-07-06
Fix cspell spelling issues in MD files

GleasonK:
Now cspell passes for `**/*.md`.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -229,7 +229,7 @@ func.func @check_tolerance() {
 ```
 
 This is just the first step in testing the numerical accuracy of StableHLO ops.
-At the moment, this is an underspecified area of the StableHLO spec, and there is
+At the moment, this is an under-specified area of the StableHLO spec, and there is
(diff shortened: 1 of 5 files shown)


docs/reference.md
```diff
@@ -229,7 +229,7 @@ func.func @check_tolerance() {
 ```
 
 This is just the first step in testing the numerical accuracy of StableHLO ops.
-At the moment, this is an underspecified area of the StableHLO spec, and there is
+At the moment, this is an under-specified area of the StableHLO spec, and there is
```
burmako, line 232: I don't think this is the right spelling? https://en.wiktionary.org/wiki/underspecify (Same for underdocumented).

burmako (changes requested):

docs/reference.md
GleasonK: fixed in all files

ghpvnist (approved):

burmako (approved):

PR #1675, opened 2023-07-06
Add lint script to spell check using cspell

GleasonK:
This can be added to CI later, for now it is a useful script to have in the repo to test against CPP/H/MD files.

Currently it reports no spelling issues for `'**/*.cpp', '**/*.h', '**/*.md'` files. I haven't taken a look at the rest of the filetypes yet. Script is based on the markdown lint script.


.cspell.json
```diff
@@ -0,0 +1,147 @@
...
+        "cmake",
+        "tablegen"
+    ],
+    "ignoreRegExpList": [
+        "TODO\\([a-zA-Z0-9]+\\)",
+        "#ifndef [A-Z_]+",
+        "#endif .*",
+        "#define [A-Z_]+"
+    ]
+}
```
burmako, line 147: Hmm this sounds like a considerable amount of work (and future maintenance) to avoid typos. I'm not sure the tradeoff is worth it. Can you talk more about the motivation for this change? Maybe I'm missing something.

burmako (changes requested):

.cspell.json
GleasonK: Happy to keep this as a local script!

Motivation is code quality, to catch spelling issues like the ones in:
- https://github.com/openxla/stablehlo/pull/1671
- https://github.com/openxla/stablehlo/pull/1674

I agree that this overhead should not be imposed on the community - that's why I don't have this set up as a CI job, just a useful script if you want to check your own files/code for spelling issues.

GleasonK:
Planning to keep this script locally instead of storing in the repo. Can revisit this decision if other users become interested in it.

PR #1677, opened 2023-07-06
Fix a build warning in conversion to TOSA

burmako:
Apparently, tosa.logical_not doesn't implement type inference, so the build warning was asking to explicitly provide a return type in the PDLL pattern, which this PR does.


GleasonK (approved):
Glad this is fixed! Back to a clean build!

PR #1698, opened 2023-07-25
Add `HLO_Commutative` trait to avoid inheriting folders

GleasonK:
After https://reviews.llvm.org/D155687 the builtin Commutative trait has folders by default. StableHLO does not want to have any default folding or canonicalization. Adding a `hlo::OpTrait::IsCommutative` and `HLO_Commutative` trait to avoid these folders.


burmako (approved):

burmako:
MLIR-HLO PR: https://github.com/openxla/xla/pull/4498.

Issue #1615, opened 2023-06-14
Should dot_general support per-axis quantization scheme? 


sdasgup3:
From the 
- TFLite [spec](https://www.tensorflow.org/lite/performance/quantization_spec#int8_quantized_operator_specifications), and
- https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/kernels/internal/reference/fully_connected.h

It seems like TFLite fully_connected op does not support per-axis. 

The per-axis scheme is [added](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#dot_general) to StableHLO `dot_general` op mainly because of the fact that `dot_general` is used in the [specification](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#convolution) of convolution op. IMO, that should not be the only reason to include per-axis scheme to dot_general. 

The ticket is about re-visiting the spec of `dot_general` and explore use-case for it to support per-axis quantization scheme.

rewu93:
Currently TFLite does not support per-axis quantized FC. However, we do see the rising demands for it due to the dramatic size increase of recent models. We may add it in the future.

jingpu:
+1 per-axis quantization FC seems useful for LLM

sdasgup3:
This issue is tracked as part of https://github.com/openxla/stablehlo/issues/1574. Closing this for now. 

For record: The current state of dot-general is that it support per-tensor quantized type which is implemented via https://github.com/openxla/stablehlo/pull/1700.

PR #1700, opened 2023-07-26
Using dequant-op-quant for convolution and dot_general

### LINKED ISSUE #1628: Consider using `dequant_op_quant` in the `dot_general`/`convolution` spec
As brought up in https://github.com/openxla/stablehlo/pull/1496#discussion_r1231063146, the ticket explores specifying `dot_general` and `convolution` using the `dequant_op_quant` strategy that we have applied for the element wise operations.  An immediate benefit with that would be that we do not have to treat the specification of the above ops any differently than the element wise ops. Please refer to the link above for additional benefits.

sdasgup3:
fixes #1628 

The idea behind the PR is to treat the quantization of `dot_general` and `convolution` similar to how various other element wise ops are specified. Other than the obvious parity of the specification with other ops, there are other benefits too.

1. `dot_general` and `convolution` are potential candidates for per-axis quantization. With this solution we can define per-axis scheme uniformaly in the meta-functions `dequantize`  and `quantize` and thereby do not have to  explicitly handle the per-axis scheme for these ops. 
2. `dot_general` and `convolution` uses some data_movement ops (like `reshape`, `transpose`, `pad`, `slice`) for the underlying specification. There is a concern where  the fact that `dot_general` and `convolution` can potentially use per-axis quantization necessitates handling of the per-axis scheme for the data movement ops just because the later are used in the specification of the former.  With the current solution,  we wouldn't need to worry about per-axis quantization for the underlying data movement ops.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -2391,7 +2391,7 @@ planning to address this in
 | Label | Name                         | Type                                                         | Constraints                   |
 |-------|------------------------------|--------------------------------------------------------------|-------------------------------|
 | (I1)  | `lhs`                        | tensor or per-tensor quantized tensor                        | (C5-C6), (C9-C10), (C12-C16)  |
-| (I2)  | `rhs`                        | tensor or quantized tensor                                   | (C7-C10), (C12-C18)           |
+| (I2)  | `rhs`                        | tensor or per-tensor quantized tensor                        | (C7-C10), (C12)               |
 | (I3)  | `lhs_batching_dimensions`    | 1-dimensional tensor constant of type `si64`                 | (C1), (C3), (C5), (C9), (C12) |
 | (I4)  | `rhs_batching_dimensions`    | 1-dimensional tensor constant of type `si64`                 | (C1), (C4), (C7), (C9)        |
 | (I5)  | `lhs_contracting_dimensions` | 1-dimensional tensor constant of type `si64`                 | (C2), (C3), (C6), (C10)       |
@@ -2400,9 +2400,9 @@ planning to address this in
 
 #### Outputs
 
-| Name     | Type                       | Constraints                |
-|----------|----------------------------|----------------------------|
-| `result` | tensor or quantized tensor | (C12), (C14), (C16), (C18) |
+| Name     | Type                                  | Constraints         |
+|----------|---------------------------------------|---------------------|
+| `result` | tensor or per-tensor quantized tensor | (C12), (C14), (C16) |
 
 #### Constraints
 
@@ -2429,8 +2429,6 @@ planning to address this in
   * (C15) `storage_type(lhs) = storage_type(rhs)`.
   * (C16) `expressed_type(lhs) = expressed_type(rhs) = expressed_type(result)`.
   * (C17) `zero_points(rhs) = 0`.
-  * (C18) If `is_per_tensor_quantized(rhs)`,
-    then `is_per_tensor_quantized(result)`.
 
 #### Examples
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -1999,7 +1999,7 @@ If `feature_group_count = 1` and `batch_group_count = 1`, then for all
 `output_spatial_index` in `index_space(dim(result, output_spatial_dimensions...))`,
 `result[result_shape(:, output_spatial_index, :)] = dot_product` where:
 
-* `padding_value = constant(is_quantized(lhs) ? quantize(0, element_type(lhs)) : 0, element_type(lhs))`.
+* `padding_value = constant(0, element_type(lhs))`.
 * `padded_lhs = pad(lhs, padding_value, lhs_padding[:, 0], lhs_padding[:, 1], lhs_base_dilations - 1)`.
 * `lhs_window_start = lhs_shape(0, output_spatial_index, 0) * lhs_window_strides`.
 * `lhs_window = slice(padded_lhs, lhs_window_start, lhs_window_start + lhs_window_dimensions, lhs_window_dilations)`.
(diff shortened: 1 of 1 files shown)


loganchien (approved):

burmako (approved):

docs/spec.md
```diff
@@ -2346,23 +2356,18 @@ More formally, `result[result_index] = dot_product`, where:
...
-  * `rounded_dot_product = round_nearest_even(integer_dot_product * (scale(reshaped_lhs_slice) * scale(reshape_rhs_slice) / scale(result)))`.
-  * `dot_product = clamp(storage_min(result), rounded_dot_product + zero_point(result), storage_max(result))`.
+* `dot_product = reduce(
+    inputs=[multiply(reshaped_lhs_slice, reshaped_rhs_slice)],
+    init_values=[constant(0, element_type(result))],
+    dimensions=range(size(lhs_contracting_dimensions)),
+    body=lambda x, y: add(x, y))`.
 <!-- markdownlint-enable line-length -->
 
+For quantized types, performs `dequantize_op_quantize(
```
dominicsymes, line 2366: I’d like to check which axes are permitted to be per-axis quantized for dot_general. Convolution states in (C32) and (C33) that per-axis quantization_dimension is an output feature dimension (which makes sense as these are not summed over by the convolution operation). For dot_general, there does not seem to be corresponding constraints. For example, I’d expect a constraint after (C18) that if is_per_axis_quantized(rhs) then quantization_dimension is one of the rhs_result_dimensions. In particular, that it is not one of rhs_contracting_dimensions. Similar constraints also for the result tensor.

sdasgup3: Thanks for pointing out this important aspect!

Just a bit of background: One of the main reasons per-axis quantization is added in dot_general is because of how dot_general is used in the specification of convolution. I acknowledge that is this not an ideal reason for supporting  per-axis quantization for an op and we identified it as a problem in #1615. As mentioned in the ticket it seems that currently TFLite  does not support per-axis scheme for this op. IMO, that is one reason we skip detailing the quantization behavior for the op is isolation. In any case, I agree with you that the relevant details like `quantized dimension` are missing from the `dot_general` op spec and that is something we should to address if we are allowing the `rhs` and `result` to be per-axis quantized.

The current PR  allows both `convolution` and  `dot_general` op to be defined in terms of floating-point semantics. With that the `convolution` op specification will use the floating-point behavior of dot_general (as opposed to using dot_general op with quantized types). The immediate benefit is that  dot_general does not need to be per-axis quantized just because it is used in the specification of `convolution`. There must be a better reason for the support.

Also, your proposal about quantization dimension to not include the contracting dimensions makes sense. 

I propose that we address it as part of https://github.com/openxla/stablehlo/issues/1615. 

Please let me know your opinion?

rewu93 (approved):

docs/spec.md
dominicsymes: Thanks @sdasgup3 for explaining the background and the link to #1615.

Since it is still an open question if dot_general needs per-axis quantization, and it is not required for the convolution specification any more, I agree with your comment that the dot_general specification should be per-tensor quantized for now to avoid confusion. So, label (I2) below (line 2394) defining rhs type and result type (line 2405) should be updated to “tensor or per-tensor quantized tensor”?

As you suggest there could be a further update, if needed, when #1615 is resolved.

sdasgup3: Thanks @dominicsymes 
Makes sense for the `dot_general` to have per-tensor quantization scheme for now. Please take a look at the updated spec.

loganchien (approved):

docs/spec.md
dominicsymes: Thanks for updating. Looks good to me.

dominicsymes (approved):

docs/spec.md
```diff
@@ -2027,6 +2027,16 @@ If `batch_group_count > 1`:
...
+    lambda lhs, rhs: convolution(lhs, rhs, window_strides, padding,
+        lhs_dilation, rhs_dilation, window_reversal, input_batch_dimension,
+        input_feature_dimension, input_spatial_dimensions,
+        kernel_input_feature_dimension, kernel_output_feature_dimension,
+        kernel_spatial_dimensions, output_batch_dimension,
+        output_feature_dimension, output_spatial_dimensions,
+        feature_group_count, batch_group_count, precision_config), lhs, rhs,
+    type(result))`.
+
```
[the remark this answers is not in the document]
sdasgup3: >> If it is dequantize_op_quantize, are C29 and C31 of convolution still relevant?
Yes, there are relevant as the constraints are on the syntax of the quantized convolution. The semantics is modified to first dequantize and then do the operations in floating-point domain. Please let me know if this makes sense. 

>> Also, I think allowing hybrid quantization(relaxing C28?) now makes more sense with this change

Switching the spec to hybrid op does not change the semantics much: Instead of dequantizing both lhs and rhs, we need to to dequantize just the rhs.

docs/spec.md
```diff
@@ -2395,9 +2400,9 @@ planning to address this in
...
 #### Outputs
 
-| Name     | Type                       | Constraints                |
-|----------|----------------------------|----------------------------|
-| `result` | tensor or quantized tensor | (C12), (C14), (C16), (C18) |
+| Name     | Type                                  | Constraints         |
+|----------|---------------------------------------|---------------------|
+| `result` | tensor or per-tensor quantized tensor | (C12), (C14), (C16) |
 
 #### Constraints
```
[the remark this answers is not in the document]
sdasgup3: `C18` is already removed.

As mentioned before there are relevant constraints supporting the syntax of quantized operands for the op.

doyeonkim0 (approved):

Issue #650, opened 2022-11-29
Expose `num_replicas` and `num_partitions` in StableHLO.


sdasgup3:
### Request description

The statically know values  `num_replicas` and `num_partitions`, being provided by [HLOModuleConfig](https://github.com/tensorflow/tensorflow/blob/5631d0d1d017eff1d1a54e32f76ef56f8ca04a8a/tensorflow/compiler/xla/service/hlo_module_config.h#L361), helps to find the [size of each process group ](https://github.com/tensorflow/tensorflow/blob/20c6943d3cd7e07da162f7778a0af5d3776274b4/tensorflow/compiler/xla/service/hlo_verifier.cc#L95)to be employed in parallel execution. Currently these values are not exposed in StableHLO. 

Having these values in StableHLO will enable type inference of CollectiveOps. 

Originated in https://github.com/openxla/stablehlo/pull/503#discussion_r1032732950 and
https://github.com/openxla/stablehlo/pull/503#discussion_r1032839388

burmako:
JAX has just added `mhlo.num_replicas` and `mhlo.num_partitions` to their lowering: https://github.com/google/jax/pull/15586. It's great to know that this information is available during lowering :tada:

PR #1715, opened 2023-08-04
Add interpreter for CollectivePermuteOp

### LINKED ISSUE #1124: Add interpreter for CollectivePermuteOp

ghpvnist:

```
(I1) `operand`: tensor.
(I2) `source_target_pairs`: 2-dimensional tensor constant of type `si64`.
(I3) `channel_id`: constant of type `si64`.
(C1) `dim(source_target_pairs, 1) = 2`.
(C2) `is_unique(source_target_pairs[:, 0])`.
(C3) `is_unique(source_target_pairs[:, 1])`.
(C4) `0 <= source_target_pairs < N`, where N is defined as:
* `num_partitions` if `cross_partition` is used.
(C5) `type(result) = type(operand)`.
```

```
I1: a) `operand` is not a tensor. (Covered by ODS).
I2: a) `source_target_pairs` is not a 2-dimensional tensor constant of type `si64`.
C1: a) `dim(source_target_pairs, 1) != 2`.
C2: a) `is_unique(source_target_pairs[:, 0]) = false`.
C3: a) `is_unique(source_target_pairs[:, 1]) = false`.
C4: a) `source_target_pairs < 0`.
C4: b) `source_target_pairs >= N`, where `N` is defined as:
* `num_partitions` if `cross_partition` is used.
C5: a) `type(result) != type(operand)`.
```

```
I2a: `source_target_pairs` is not a 2-dimensional tensor constant of type `si64`.
C1a: `dim(source_target_pairs, 1) != 2`.
C2a: `is_unique(source_target_pairs[:, 0]) = false`.
C3a: `is_unique(source_target_pairs[:, 1]) = false`.
C4a: `source_target_pairs < 0`.
C4b: `source_target_pairs >= N`, where `N` is defined as:
* `num_partitions` if `cross_partition` is used.
C5a: `type(result) != type(operand)`.
```

Notes:
  * C4b verification is infeasible since `num_replicas` and `num_partitions` are not known statically at the moment (see #650).

closes #1124


stablehlo/reference/ProcessGrid.h
```diff
@@ -0,0 +1,125 @@
...
+
+  bool operator<(const ProcessId &other) const {
+    return std::pair<uint32_t, uint32_t>{replicaId, partitionId} <
+           std::pair<uint32_t, uint32_t>{other.replicaId, other.partitionId};
+  }
+
+  bool operator==(const ProcessId &other) const {
+    return std::pair<uint32_t, uint32_t>{replicaId, partitionId} ==
+           std::pair<uint32_t, uint32_t>{other.replicaId, other.partitionId};
+  }
```
burmako, line 48: Let's move the implementations of the operations into the .cpp file, to make it easier to understand the APIs in this header.

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -0,0 +1,128 @@
...
+  }
+  return processGroups;
+}
+
+SmallVector<std::pair<ProcessId, Tensor>> ProcessGrid::rendezvous(
+    ProcessGroup processGroup, int64_t channelId, ProcessId &processId,
+    const Tensor &operand) {
+  {
+    std::unique_lock<std::mutex> lk(
+        resourceLockMap_[{processGroup, channelId}]);
```
burmako, line 97: From https://llvm.org/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly: "Avoid abbreviations unless they are well known". I don't think we're gaining much by saving two characters at the cost of reducing readability.

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -0,0 +1,128 @@
...
+    channels_[{processGroup, channelId}].push_back({processId, operand});
+  }
+
+  resourceConditionMap_[{processGroup, channelId}].notify_all();
```
burmako, line 101: Why is this needed?

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -0,0 +1,128 @@
...
+                     processGroup.size();
+            }))
+      llvm::report_fatal_error("rendezvous timed out");
+
+    std::sort(channels_[{processGroup, channelId}].begin(),
+              channels_[{processGroup, channelId}].end(),
+              [](std::pair<ProcessId, Tensor> &lhs,
+                 std::pair<ProcessId, Tensor> &rhs) {
+                return lhs.first < rhs.first;
+              });
```
burmako, line 121: I don't think the sort piece is needed at the moment. And in the future it could be hidden as an implementation detail of `RendezvousResult`.

stablehlo/reference/ProcessGrid.cpp
burmako, line 97: Should we use `std::lock_guard` here, given that we don't really use the additional functionality provided by `std::unique_lock`?

stablehlo/reference/ProcessGrid.h
```diff
@@ -0,0 +1,125 @@
...
+  ProcessGroups crossReplica(SmallVector<SmallVector<uint32_t>> replicaGroups);
+
+  /// Cross-partition communication strategy.
+  ProcessGroups crossPartition(
+      SmallVector<SmallVector<uint32_t>> partitionGroups);
+
+  /// Synchronizes running processes in the `ProcessGroup` and returns the
+  /// collected resources from all partitipating processes in the process group.
+  SmallVector<std::pair<ProcessId, Tensor>> rendezvous(
+      ProcessGroup processGroup, int64_t channelId, ProcessId &processId,
```
burmako, line 79: What do you think about defining `using ChannelId = int64_t` in this file, and then using that definition both here and in Ops.h/cpp.

stablehlo/reference/ProcessGrid.cpp
burmako, line 97: Let's refactor the key into something like `std::pair<ProcessGroup, ChannelId> channelKey(processGroup, channelId);`? It's used quite a few times in this method.

stablehlo/reference/ProcessGrid.h
```diff
@@ -0,0 +1,125 @@
...
+#include <utility>
+
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/Support/LLVM.h"
+#include "stablehlo/reference/Tensor.h"
+
+namespace mlir {
+namespace stablehlo {
+
+// Object to represent StableHLO `process_id`.
```
burmako, line 32: Simplify to "StableHLO `process_id`", similarly to "StableHLO `replica_id`" below? (Same for the comment on `ProcessGroup`).

stablehlo/reference/ProcessGrid.h
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  bool operator==(const ProcessId &other) const {
+    return std::pair<uint32_t, uint32_t>{replicaId, partitionId} ==
+           std::pair<uint32_t, uint32_t>{other.replicaId, other.partitionId};
+  }
+};
+
+// Object to represent StableHLO process group.
+struct ProcessGroup : SmallVector<ProcessId> {};
```
burmako, line 52: https://llvm.org/docs/CodingStandards.html#use-of-class-and-struct-keywords: "struct should be used when all members are declared public". Given that not all members of this program element are public, `class` is a better fit.

stablehlo/reference/ProcessGrid.h
```diff
@@ -0,0 +1,125 @@
...
+  std::map<std::pair<ProcessGroup, int64_t>, std::mutex> resourceLockMap_;
+
+  /// Synchronization primitive used in `rendezvous` function.
+  std::map<std::pair<ProcessGroup, int64_t>, std::condition_variable>
+      resourceConditionMap_;
+
+  /// Internal mapping of StableHLO `channel_id` to its value.
+  std::map<std::pair<ProcessGroup, int64_t>,
+           SmallVector<std::pair<ProcessId, Tensor>>>
+      channels_;
```
burmako, line 99:
```C++
  /// Internal storage used to implement `rendezvous`.
  /// Each call to `rendezvous`, i.e. each combination `processGroup` and
  /// `channelId`, has its own key in the map.
  /// Within the implementation of `rendezvous`, the value corresponding to
  /// this key is gradually populated with tensors arriving from different
  /// processes in the process group.
  std::map<std::pair<ProcessGroup, ChannelId>, RendezvousResult> channels_;

  /// Synchronization primitive used to manage concurrent access to `channels_`.
  std::map<std::pair<ProcessGroup, ChannelId>, std::mutex> channelLocks_;

  /// Synchronization primitive used to manage concurrent access to `channels_`.
  std::map<std::pair<ProcessGroup, ChannelId>, std::condition_variable>
      channelConditions_;
```

Feedback:
  1) It's unclear what's a value of a channel_id, so I elaborated on that.
  2) Dropped the `Map` prefixes, because I don't think much is gained by signaling in the name that this is a map - we can kind of get this already from the signature on the definition site and from the called methods on use sites.
  3) Renamed the `resource` prefix to clearly associate these maps with `channels_`.

stablehlo/reference/ProcessGrid.h
```diff
@@ -0,0 +1,125 @@
...
+
+  /// Cross-partition communication strategy.
+  ProcessGroups crossPartition(
+      SmallVector<SmallVector<uint32_t>> partitionGroups);
+
+  /// Synchronizes running processes in the `ProcessGroup` and returns the
+  /// collected resources from all partitipating processes in the process group.
+  SmallVector<std::pair<ProcessId, Tensor>> rendezvous(
+      ProcessGroup processGroup, int64_t channelId, const Tensor &operand);
+};
```
burmako, line 120: Let's split `class Process` into a separate file (Process.h/cpp)? This file is getting pretty intense, so modularizing it would be beneficial. (`ProcessId` would need to stay here, and that's fine imo).

burmako (changes requested):

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -0,0 +1,128 @@
...
+  return processGroups;
+}
+
+SmallVector<std::pair<ProcessId, Tensor>> ProcessGrid::rendezvous(
+    ProcessGroup processGroup, int64_t channelId, ProcessId &processId,
+    const Tensor &operand) {
+  {
+    std::unique_lock<std::mutex> lk(
+        resourceLockMap_[{processGroup, channelId}]);
+    channels_[{processGroup, channelId}].push_back({processId, operand});
```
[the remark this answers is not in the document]
ghpvnist: Updated.

stablehlo/reference/ProcessGrid.h
```diff
@@ -0,0 +1,125 @@
...
+  bool operator==(const ProcessId &other) const {
+    return std::pair<uint32_t, uint32_t>{replicaId, partitionId} ==
+           std::pair<uint32_t, uint32_t>{other.replicaId, other.partitionId};
+  }
+};
+
+// Object to represent StableHLO process group.
+struct ProcessGroup : SmallVector<ProcessId> {};
+
+// Object to represent StableHLO process groups.
```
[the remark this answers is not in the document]
ghpvnist: [StableHLO process groups](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#collective-ops) is defined here.

stablehlo/reference/ProcessGrid.h
ghpvnist: Actually, why not just make all members public and let both `ProcessGroup` and `ProcessGroups` be a `struct`. Both are essentially an alias of `SmallVector`.

stablehlo/reference/Ops.cpp
```diff
@@ -800,6 +820,37 @@ Tensor evalClzOp(const Tensor &operand, ShapedType resultType) {
...
+  if (!process)
+    llvm::report_fatal_error(
+        "collective_permute is only supported when run via "
+        "interpreter.run_parallel");
+
+  ProcessGroups processGroups;
+  if (channelId <= 0) processGroups = process->crossReplica(sourceTargetPairs);
+  if (channelId > 0) processGroups = process->crossPartition(sourceTargetPairs);
+
+  std::optional<Tensor> result;
```
burmako, line 835: Given that Tensor supports the notion of empty tensors, we don't need to represent optional tensors via `std::optional`. Just `Tensor` would suffice. The only thing missing is this:

```C++
class Tensor {
  ...
  explicit operator bool() const { return (bool)impl_; }
  bool operator!() const { return !impl_; }
  ...
}
```

stablehlo/reference/ProcessGrid.h
```diff
@@ -0,0 +1,127 @@
...
+  Tensor lookup(ProcessId processId);
+
+  /// Inserts `tensor` into the map `result_` using the key `processId`.
+  void insert(ProcessId processId, Tensor tensor);
+
+  /// Erases all elements in the map `result_`.
+  void clear();
+
+  /// Returns the size of `result_`.
+  size_t size();
```
burmako, line 69: Let's write comments in terms of the public API, not private fields.

```C++
/// Represents a result of a `ProcessGrid::rendezvous` where multiple processes
/// synchronize at a barrier and contribute a Tensor each.
/// This class is pretty much a map from ProcessId to Tensor, with the 
/// map-like API.
class RendezvousResult {
  /// Iterates through the map `result_` and returns the tensor associated with
  /// the key `processId`.
  /// If key is not found, return an empty `Tensor`.
  Tensor lookup(ProcessId processId);

  /// Inserts `tensor` into the map using the key `processId`.
  void insert(ProcessId processId, Tensor tensor);

  /// Erases all elements in the map.
  void clear();

  /// Returns the number of elements in the map.
  size_t size();
}
```

stablehlo/reference/ProcessGrid.h
```diff
@@ -0,0 +1,127 @@
...
+
+ private:
+  ///
+  std::map<ProcessId, Tensor> result_;
+};
+
+/// Class to model a process grid.
```
burmako, line 76: "/// StableHLO process grid" similarly to above comments?

stablehlo/reference/ProcessGrid.h
```diff
@@ -0,0 +1,127 @@
...
+class ProcessGrid {
+ public:
+  /// \name Constructors
+  /// @{
+  ProcessGrid(uint32_t numReplicas, uint32_t numPartitions);
+  /// @}
+
+  /// StableHLO `cross_replcia`.
```
burmako, line 84: "/// StableHLO `cross_replica` communication strategy"? Same below.

stablehlo/reference/ProcessGrid.h
```diff
@@ -0,0 +1,127 @@
...
+  ProcessGroups crossPartition(
+      SmallVector<SmallVector<uint32_t>> partitionGroups);
+
+  /// Each participating process in the `processGroup` appends its data
+  /// `operand` to the `channels_` map using the pair (processGroup, channelId)
+  /// as a key. `channels_` is cleared before any process adds its data.
+  /// `rendezvous` then returns `RendezvousResult` containing a mapping of
+  /// `processId` to its data `operand` once all participating processes have
+  /// successfully added their data. Throws an error after a timeout when
+  /// synchronization deadlocks.
```
burmako, line 97: Same as above - let's write this comment in terms of a public API.

burmako (changes requested):

stablehlo/reference/ProcessGrid.h
```diff
@@ -0,0 +1,127 @@
...
+
+// StableHLO `process_id`.
+struct ProcessId {
+  /// StableHLO `replica_id`.
+  uint32_t replicaId;
+
+  /// StableHLO `partition_id`.
+  uint32_t partitionId;
+
+  bool operator<(const ProcessId &other) const;
```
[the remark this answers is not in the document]
ghpvnist: It is still needed to sort the keys of `std::map<std::pair<ProcessGroup, ChannelId>, RendezvousResult> channels_`. We could use `std::unordered_map`, but this requires a hash function instead. We will soon need to use this operator to sort `RendezvousResult` anyways hence keeping this change.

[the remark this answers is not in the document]
burmako: "We will soon need to use this operator to sort RendezvousResult anyways hence keeping this change". I think this operator should come with an update to the spec, because this piece is underspecced at the moment. Let's defer to a later PR.

"It is still needed to sort the keys". Hmm good point. Let's switch to a different container then. Let's use llvm::DenseMap to follow LLVM conventions.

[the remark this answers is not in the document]
burmako: "because this piece is underspecced at the moment". Actually, I was wrong - the spec does talk about how `process_groups` are computed exactly. But I still think we shouldn't have `operator<`, since this can be an implementation detail of `RendezvousResult`.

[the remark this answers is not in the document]
ghpvnist: >  Let's switch to a different container then. Let's use llvm::DenseMap to follow LLVM conventions.

I looked into `std::map`, `std::unordered_map`, and `llvm::DenseMap`, and of the three, `std::map` seemed to have the most straightforward, minimal implementation. Since we define how `process_groups` are computed for each communication strategy, we could define `operator<` on the `ProcessGroup`, but this seems like an anti-pattern. I've added a comment to justify the need.

stablehlo/reference/ProcessGrid.h
burmako: The current wording is still formulated in terms of the implementation details which involve an internal map with an internal key - something that users of this API shouldn't need to know about.

```C++
  /// Synchronize a StableHLO process with the `processId` with other StableHLO
  /// processes in the `processGroup` using a `channelId`.
  ///
  /// A call to this method represents a barrier, i.e. it blocks the calling
  /// OS thread until all StableHLO processes from the `processGroup` call this 
  /// method with the same `channelId`. If the calling OS thread doesn't 
  /// correspond to the StableHLO process with `processId`, the behavior is 
  /// undefined.
  ///
  /// If any of the StableHLO processes from `processGroup` fails to arrive
  /// at the barrier within 3 seconds, the `rendezvous` fails with a fatal 
  /// error. This is to make sure that errors in underlying StableHLO programs 
  /// or bugs in the StableHLO interpreter don't deadlock the entire 
  /// interpreter.
  ///
  /// At the barrier, each StableHLO process contributes a tensor, and these
  /// tensors are accumulated in `RendezvousResult` which is returned to all
  /// callers once the barrier has been reached by all StableHLO processes.
```

burmako (changes requested):

stablehlo/reference/ProcessGrid.h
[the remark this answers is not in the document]
ghpvnist: I opened a [ticket](https://github.com/openxla/stablehlo/issues/1720) to address this in the future.

burmako (approved):

PR #1717, opened 2023-08-07
Typo fixes in the example provided for the spec of CollectivePermute

sdasgup3:


ghpvnist (changes requested):
I'd actually be overwriting most of the fixes here in #1715, but it's probably best to address these issues separately. Two more fixes are needed:

1.
```
* `cross_replica(source_target_pairs)` if `channel_id <= 0`.
* `cross_partition(source_target_pairs)` if `channel_id > 0`.
```
2.
```
`0 <= source_target_pairs < N`, where `N` is defined as:
```

ghpvnist (approved):

PR #1721, opened 2023-08-08
Add interpreter for AllReduceOp

### LINKED ISSUE #1119: Add interpreter for AllReduceOp

ghpvnist:

```
(I1) `operand` tensor.
(I2) `replica_groups` variadic number of 1-dimensional tensor constant of type `si64`.
(I3) `channel_id` constant of type `si64`.
(I4) `use_global_device_ids` constant of type `i1`.
(I5) `computation` function.
(C1) `is_unique(replica_groups)`.
(C2) `size(replica_groups)` is defined as:
* `num_replicas` if `cross_replica_and_partition` is used.
* `num_processes` if `flattened_ids` is used.
(C3) `0 <= replica_groups < size(replica_groups)`.
(C4) If `use_global_device_ids = true`, then `channel_id > 0`.
(C5) `computation` has type `(tensor<E>, tensor<E>) -> (tensor<E>)` where
     `E = element_type(operand)`.
(C6) `type(result) = type(operand)`.
```

```
I1: a) `operand` is not a tensor. (Covered by ODS).
I2: a) `replica_groups` is not a variadic number of 1-dimensional tensor.
I2: b) `element_type(replica_groups) != si64`. (Covered by ODS).
I4: a) `use_global_device_ids` is not a constant of type `i1`. (Covered by ODS).
I5: a) `computation` is not a function. (Covered by ODS).
C1: a) `is_unique(replica_groups)` = false.
C2: a) `size(replica_groups)` is not defined as:
* `num_replicas` if `cross_replica_and_partition` is used.
* `num_processes` if `flattened_ids` is used.
C3: a) `replica_groups < 0`.
C3: b) `replica_groups >= size(replica_groups)`.
C4: a) If `use_global_device_ids = true` and `channel_id <= 0`.
C5: a) `computation` does not have type `(tensor<E>, tensor<E>) -> (tensor<E>)`
       where `E = element_type(operand)`.
C6: a) `type(result) != type(operand)`. (Covered by ODS).
```

```
I2: a) `replica_groups` is not a variadic number of 1-dimensional tensor.
C1: a) `is_unique(replica_groups)` = false.
C2: a) `size(replica_groups)` is not defined as:
* `num_replicas` if `cross_replica_and_partition` is used.
* `num_processes` if `flattened_ids` is used.
C3: a) `replica_groups < 0`.
C3: b) `replica_groups >= size(replica_groups)`.
C4: a) If `use_global_device_ids = true` and `channel_id <= 0`.
C5: a) `computation` does not have type `(tensor<E>, tensor<E>) -> (tensor<E>)`
       where `E = element_type(operand)`.
```

Notes:
  * I2 is not tested due to #498.
  * C2a verification is infeasible since `num_replicas` and `num_partitions` are not knwon statically.
  * C3b verification is infeasible since `size(replica_groups)` is not known statically.

closes #1119

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -784,13 +784,13 @@ Afterwards, within each `process_group`:
 
 #### Inputs
 
-| Label | Name                    | Type                                         | Constraints |
-|-------|-------------------------|----------------------------------------------|-------------|
-| (I1)  | `operand`               | tensor                                       | (C5), (C6)  |
-| (I2)  | `replica_groups`        | 2-dimensional tensor constant of type `si64` | (C1-C3)     |
-| (I3)  | `channel_id`            | constant of type `si64`                      | (C4)        |
-| (I4)  | `use_global_device_ids` | constant of type `i1`                        | (C4)        |
-| (I5)  | `computation`           | function                                     | (C5)        |
+| Label | Name                    | Type                                                             | Constraints |
+|-------|-------------------------|------------------------------------------------------------------|-------------|
+| (I1)  | `operand`               | tensor                                                           | (C5), (C6)  |
+| (I2)  | `replica_groups`        | variadic number of 1-dimensional tensor constants of type `si64` | (C1-C3)     |
+| (I3)  | `channel_id`            | constant of type `si64`                                          | (C4)        |
+| (I4)  | `use_global_device_ids` | constant of type `i1`                                            | (C4)        |
+| (I5)  | `computation`           | function                                                         | (C5)        |
 
 #### Outputs
 
@@ -802,7 +802,7 @@ Afterwards, within each `process_group`:
 
 * (C1) `is_unique(replica_groups)`.
 * (C2) `rank(replica_groups) = 2`.
-* (C3) `1 < size(replica_groups)`.
+* (C3) `1 <= size(replica_groups)`.
 * (C4) `0 <= replica_groups < N` where `N` is defined as:
   * `num_replicas` if `cross_replica` is used.
   * `num_replicas` if `cross_replica_and_partition` is used.
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -784,13 +784,13 @@ Afterwards, within each `process_group`:
 
 #### Inputs
 
-| Label | Name                    | Type                                                             | Constraints |
-|-------|-------------------------|------------------------------------------------------------------|-------------|
-| (I1)  | `operand`               | tensor                                                           | (C5), (C6)  |
-| (I2)  | `replica_groups`        | variadic number of 1-dimensional tensor constants of type `si64` | (C1-C3)     |
-| (I3)  | `channel_id`            | constant of type `si64`                                          | (C4)        |
-| (I4)  | `use_global_device_ids` | constant of type `i1`                                            | (C4)        |
-| (I5)  | `computation`           | function                                                         | (C5)        |
+| Label | Name                    | Type                                         | Constraints |
+|-------|-------------------------|----------------------------------------------|-------------|
+| (I1)  | `operand`               | tensor                                       | (C5), (C6)  |
+| (I2)  | `replica_groups`        | 2-dimensional tensor constant of type `si64` | (C1-C3)     |
+| (I3)  | `channel_id`            | constant of type `si64`                      | (C4)        |
+| (I4)  | `use_global_device_ids` | constant of type `i1`                        | (C4)        |
+| (I5)  | `computation`           | function                                     | (C5)        |
 
 #### Outputs
 
@@ -801,36 +801,37 @@ Afterwards, within each `process_group`:
 #### Constraints
 
 * (C1) `is_unique(replica_groups)`.
-* (C2) `size(replica_groups)` is defined as:
+* (C2) `rank(replica_groups) = 2`.
+* (C3) `1 < size(replica_groups)`.
+* (C4) `0 <= replica_groups < N` where `N` is defined as:
   * `num_replicas` if `cross_replica` is used.
   * `num_replicas` if `cross_replica_and_partition` is used.
   * `num_processes` if `flattened_ids` is used.
-* (C3) `0 <= replica_groups < size(replica_groups)`.
-* (C4) If `use_global_device_ids = true`, then `channel_id > 0`.
-* (C5) `computation` has type `(tensor<E>, tensor<E>) -> (tensor<E>)` where
+* (C5) If `use_global_device_ids = true`, then `channel_id > 0`.
+* (C6) `computation` has type `(tensor<E>, tensor<E>) -> (tensor<E>)` where
        `E = element_type(operand)`.
-* (C6) `type(result) = type(operand)`.
+* (C7) `type(result) = type(operand)`.
 
 #### Examples
 
 ```mlir
 // num_replicas: 2
 // num_partitions: 1
-// %operand@(0, 0): [1.0, 2.0, 3.0, 4.0]
-// %operand@(1, 0): [5.0, 6.0, 7.0, 8.0]
+// %operand@(0, 0): [1, 2, 3, 4]
+// %operand@(1, 0): [5, 6, 7, 8]
 %result = "stablehlo.all_reduce"(%operand) ({
-  ^bb0(%arg0: tensor<f32>, %arg1: tensor<f32>):
-    %0 = "stablehlo.add"(%arg0, %arg1) : (tensor<f32>, tensor<f32>) -> tensor<f32>
-    "stablehlo.return"(%0) : (tensor<f32>) -> ()
+  ^bb0(%arg0: tensor<i64>, %arg1: tensor<i64>):
+    %0 = "stablehlo.add"(%arg0, %arg1) : (tensor<i64>, tensor<i64>) -> tensor<i64>
+    "stablehlo.return"(%0) : (tensor<i64>) -> ()
 }) {
   replica_groups = dense<[[0, 1]]> : tensor<1x2xi64>,
-  // channel_id = 0
   channel_handle = #stablehlo.channel_handle<handle = 0, type = 0>
-  // use_global_device_ids = false
-} : (tensor<4xf32>) -> tensor<4xf32>
-// %result@(0, 0): [6.0, 8.0, 10.0, 12.0]
-// %result@(1, 0): [6.0, 8.0, 10.0, 12.0]
+} : (tensor<i64>) -> tensor<i64>
+// %result@(0, 0): [6, 8, 10, 12]
+// %result@(1, 0): [6, 8, 10, 12]
 ```
+
+&nbsp;[More Examples](../stablehlo/tests/interpret_all_reduce.mlir)
 
 ### all_to_all
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -801,16 +801,15 @@ Afterwards, within each `process_group`:
 #### Constraints
 
 * (C1) `is_unique(replica_groups)`.
-* (C2) `rank(replica_groups) = 2`.
-* (C3) `1 <= size(replica_groups)`.
-* (C4) `0 <= replica_groups < N` where `N` is defined as:
+* (C2) `size(replica_groups)` is defined as:
   * `num_replicas` if `cross_replica` is used.
   * `num_replicas` if `cross_replica_and_partition` is used.
   * `num_processes` if `flattened_ids` is used.
-* (C5) If `use_global_device_ids = true`, then `channel_id > 0`.
-* (C6) `computation` has type `(tensor<E>, tensor<E>) -> (tensor<E>)` where
+* (C3) `0 <= replica_groups < size(replica_groups)`.
+* (C4) If `use_global_device_ids = true`, then `channel_id > 0`.
+* (C5) `computation` has type `(tensor<E>, tensor<E>) -> (tensor<E>)` where
        `E = element_type(operand)`.
-* (C7) `type(result) = type(operand)`.
+* (C6) `type(result) = type(operand)`.
 
 #### Examples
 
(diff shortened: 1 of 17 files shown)


stablehlo/dialect/TypeInference.cpp
```diff
@@ -438,6 +439,11 @@ LogicalResult verifyReplicaGroups(std::optional<Location> location,
                              "groups cannot be empty");
 
   auto replicaIds = replicaGroups.getValues<int64_t>();
+
+  // all_reduce_c3
+  if (replicaGroups.getNumElements() == 0)
+    return emitOptionalError(location, "replica group cannot be empty");
```
burmako, line 445: Where is this constraint coming from?

stablehlo/reference/Ops.cpp
```diff
@@ -171,6 +171,27 @@ SmallVector<InterpreterValue> eval(
...
+      auto operand = scope.findTensor(allReduceOp.getOperand());
+
+      auto replicaGroupsAttr = allReduceOp.getReplicaGroups();
+      auto replicaGroupsShape = replicaGroupsAttr.getShapedType().getShape();
+      SmallVector<SmallVector<uint32_t>> replicaGroups(
+          replicaGroupsAttr.getNumElements() / replicaGroupsShape[1]);
+      for (auto &group : replicaGroups) {
+        auto it = replicaGroupsAttr.getValues<int64_t>().begin();
+        for (auto i = 0; i < replicaGroupsShape[1]; ++i) group.push_back(*it++);
+      }
```
burmako, line 184: We'll need to change this code because AllReduceOp allows replica groups of different sizes. Since DenseElementsAttr doesn't support ragged tensors, we're currently using a convention where elements with value -1 represents raggedness (see http://google3/third_party/tensorflow/compiler/xla/translate/mhlo_to_hlo/attribute_exporter.cc;l=102-107;rcl=554330020).

stablehlo/tests/infer_stablehlo.mlir
```diff
@@ -102,6 +102,24 @@ func.func @cholesky(%arg0: tensor<1x2x2xf32>) -> tensor<1x2x2xindex> {
 
 // -----
 
+// CHECK-LABEL: func @all_reduce_c7
```
burmako, line 105: Just "all_reduce".

stablehlo/tests/infer_stablehlo.mlir
```diff
@@ -102,6 +102,24 @@ func.func @cholesky(%arg0: tensor<1x2x2xf32>) -> tensor<1x2x2xindex> {
...
+    %max = stablehlo.maximum %lhs, %rhs : tensor<f32>
+    "stablehlo.return"(%max) : (tensor<f32>) -> ()
+  })
+  {
+    replica_groups = dense<[[0, 2, 4, 6], [1, 3, 5, 7]]> : tensor<2x4xi64>,
+    channel_handle = #stablehlo.channel_handle<handle = 5,type = 2>,
+    use_global_device_ids
+  } : (tensor<10xf32>) -> tensor<10xf32>
+  func.return %0 : tensor<10xf32>
+}
```
burmako, line 119: Also, we'd want to use `hlo_test_infer.get_return_types` here?

stablehlo/reference/ProcessGrid.h
```diff
@@ -66,6 +68,9 @@ class RendezvousResult {
   /// `processId`. If key is not found, return an empty `Tensor`.
   Tensor lookup(ProcessId processId);
 
+  /// Returns the tensor values of the map `result_`.
```
burmako, line 71: Let's write this comment in terms of the public API. I don't think that users of RendezvousResult should know about `result_`.

burmako, line 71: Also, this comment needs more details, specifically around the order in which tensor values are returned.

stablehlo/reference/Tensor.h
```diff
@@ -71,6 +71,8 @@ class Tensor {
   /// \name Constructors
   /// @{
   Tensor();
+  /// This constructor initializes the tensor with a scalar value.
+  explicit Tensor(const Element &initValue);
```
burmako, line 75: I don't think this additional API brings much value at the moment. It's used only once, so I think we should just inline this constructor into the callsite, until we have more use cases for it.

stablehlo/reference/ProcessGrid.h
burmako, line 71: Maybe rename to `sorted` to emphasize that the result is sorted?

stablehlo/reference/ProcessGrid.h
```diff
@@ -51,10 +51,12 @@ struct ProcessId {
 };
 
 // StableHLO `process_group`.
-class ProcessGroup : public SmallVector<ProcessId> {};
+struct ProcessGroup : public SmallVector<ProcessId> {};
 
 // StableHLO `process_groups`.
-class ProcessGroups : public SmallVector<ProcessGroup> {};
+struct ProcessGroups : public SmallVector<ProcessGroup> {
+  ProcessGroups findAll(ProcessId processId);
```
burmako, line 58: Since everything else in this file has comments, this method needs a comment (a detailed comment!) as well.

burmako, line 58: I don't think that `findAll` is the right API to express what we're looking for here, especially since the current caller simply does `[0]` on the result. Something like `std::optional<ProcessGroup> find(...)` looks better.

burmako, line 58: I think we should also rename this to `findProcess` to avoid ambiguity with `std::find`.

stablehlo/reference/Ops.h
```diff
@@ -33,6 +33,11 @@ namespace stablehlo {
 Tensor evalAbsOp(const Tensor &operand, ShapedType resultType);
 Tensor evalAddOp(const Tensor &lhs, const Tensor &rhs, ShapedType resultType);
 Token evalAfterAllOp(ArrayRef<Token> inputs, MLIRContext *context);
+Tensor evalAllReduceOp(const Tensor &operand,
+                       SmallVector<SmallVector<uint32_t>> replicaGroups,
+                       int64_t channelId, bool useGlobalDeviceIds,
+                       Region &region, Process *process, Scope &scope,
```
burmako, line 39: `Region &computation`

stablehlo/reference/Ops.cpp
```diff
@@ -702,6 +723,51 @@ Token evalAfterAllOp(ArrayRef<Token> inputs, MLIRContext *context) {
...
+    auto concatenateResult = evalConcatenateOp(
+        inputs, 0,
+        RankedTensorType::get(concatResultShape, operand.getElementType()));
+
+    result.set(*resultIt,
+               evalReduceOp(concatenateResult,
+                            Tensor(convert(operand.getElementType(), 0.0)), {0},
+                            region, process, scope)[0]
+                   .get({}));
+  }
```
burmako, line 767: I think this implementation complicates things quite a bit. The following implementation that directly follows the spec seems pretty straightforward:

```C++
Tensor result(resultType);
     ++resultIt) {
  Tensor resultElement;
  for (auto groupOperand : groupOperands) {
    Tensor groupOperandElement(
        RankedTensorType::get({}, groupOperand.getElementType()),
        groupOperand.get(*resultIt));
    if (resultElement) {
      resultElement = eval(computation, {resultElement, groupOperandElement},
                           process, &scope)[0]
                          .getTensor();
    } else {
      resultElement = groupOperandElement;
    }
  }
  result.set(*resultIt, resultElement.get({}));
}
```

burmako (changes requested):

stablehlo/dialect/TypeInference.cpp
ghpvnist: This comes from the updated spec constraints. Updated the PR description to reflect it.

stablehlo/tests/infer_stablehlo.mlir
ghpvnist: This is the [guideline](https://github.com/openxla/stablehlo/blob/main/docs/reference_checklist.md) we've been following:

6. In [infer_stablehlo.mlir](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/infer_stablehlo.mlir):
i. Make sure all constraints related to shape inference tests are present in this file, following the same naming guidelines noted above.

stablehlo/tests/infer_stablehlo.mlir
ghpvnist: Actually, since this is already covered by ODS, we don't need a test for this (there's no type inference code either). Removed.

stablehlo/reference/Tensor.h
ghpvnist: Oops the other changes weren't included. This is actually used by several other functions and improves the readability quite a bit. PTAL at the updated code.

docs/spec.md
```diff
@@ -801,37 +801,38 @@ Afterwards, within each `process_group`:
...
 * (C1) `is_unique(replica_groups)`.
-* (C2) `size(replica_groups)` is defined as:
+* (C2) `rank(replica_groups) = 2`.
+* (C3) `1 <= size(replica_groups)`.
+* (C4) `0 <= replica_groups < N` where `N` is defined as:
   * `num_replicas` if `cross_replica` is used.
   * `num_replicas` if `cross_replica_and_partition` is used.
   * `num_processes` if `flattened_ids` is used.
-* (C3) `0 <= replica_groups < size(replica_groups)`.
-* (C4) If `use_global_device_ids = true`, then `channel_id > 0`.
```
burmako, line 809: Why are we changing the spec? The current formulation works for the potentially ragged shape of `replica_groups`, while the new formulation does not (because rank is only defined on regular tensors).

stablehlo/dialect/TypeInference.cpp
burmako: I'm not sure I follow why we need to update the spec constraints in the first place.

stablehlo/reference/Ops.cpp
```diff
@@ -171,6 +171,30 @@ SmallVector<InterpreterValue> eval(
       auto inputs = scope.findTokens(afterAllOp.getInputs());
       auto result = evalAfterAllOp(inputs, afterAllOp->getContext());
       scope.add(afterAllOp.getResult(), result);
+    } else if (auto allReduceOp = dyn_cast<AllReduceOp>(op)) {
+      auto operand = scope.findTensor(allReduceOp.getOperand());
+
+      auto replicaGroupsAttr = allReduceOp.getReplicaGroups();
+      auto replicaGroupsShape = replicaGroupsAttr.getShapedType().getShape();
+      SmallVector<SmallVector<uint32_t>> replicaGroups(replicaGroupsShape[0]);
+      auto it = replicaGroupsAttr.getValues<int64_t>().begin();
```
burmako, line 180: `replicaGroupsIt` and `replicaGroup` here and below?

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -60,6 +76,8 @@ ProcessGroups ProcessGrid::crossReplica(
     SmallVector<SmallVector<uint32_t>> replicaGroups) {
   ProcessGroups processGroups;
   for (auto replicaGroup : replicaGroups) {
+    if (replicaGroup.empty())
+      llvm::report_fatal_error("empty replicaGroup found during crossReplica");
```
burmako, line 80: Why is this a problem? This is not covered in the spec. (Same below for other methods implementing communication strategies).

stablehlo/reference/ProcessGrid.h
```diff
@@ -63,9 +70,13 @@ class ProcessGroups : public SmallVector<ProcessGroup> {};
 class RendezvousResult {
  public:
   /// Iterates through the map and returns the value associated with the key
-  /// `processId`. If key is not found, return an empty `Tensor`.
+  /// `processId`. If key is not found, return an empty Tensor.
   Tensor lookup(ProcessId processId);
 
+  /// Iterates through the (ProcessId, Tensor) map entires and returns a vector
+  /// of Tensors sorted by ProcessId.
```
burmako, line 77: Sorted how?

stablehlo/reference/Tensor.h
burmako: Agreed that this improves readability, but perhaps let move this to a helper function `makeScalar` next to `makeTensor`? While we're at it, perhaps let's change the `ShapedType type, const Element &initValue` constructor into a helper function as well, called `makeSplat` or something?

burmako (changes requested):

docs/spec.md
ghpvnist: I thought rank is always 2 and raggedness only changes  `dim(replica_groups, 1)`?

burmako: The way how TensorConstant is defined doesn't support raggedness in the first place.

ghpvnist: Ah makes sense.

stablehlo/reference/ProcessGrid.cpp
ghpvnist: The original constraint is
* (C3) `0 <= replica_groups < size(replica_groups)`.

This implies the updated constraint:
* (C3 new) `1 <= size(replica_groups)`.
and
* (C2 old) `size(replica_groups)` is defined as:
  * `num_replicas` if `cross_replica_and_partition` is used.
  * `num_processes` if `flattened_ids` is used.

And as per https://github.com/openxla/stablehlo/issues/498#issuecomment-1331396793, replica_groups cannot be empty. If this is checked in the verifier the checks here aren't needed. I guess this circles back to #498. I'll leave the spec alone for now and let the ticket address this as a whole.

docs/spec.md
```diff
@@ -816,22 +816,22 @@ Afterwards, within each `process_group`:
...
-} : (tensor<4xf32>) -> tensor<4xf32>
-// %result@(0, 0): [6.0, 8.0, 10.0, 12.0]
-// %result@(1, 0): [6.0, 8.0, 10.0, 12.0]
+} : (tensor<i64>) -> tensor<i64>
+// %result@(0, 0): [6, 8, 10, 12]
+// %result@(1, 0): [6, 8, 10, 12]
 ```
 
+&nbsp;[More Examples](../stablehlo/tests/interpret_all_reduce.mlir)
+
```
burmako, line 834: It looks like the constraint changes got rolled back, so we'll also want to update the PR description?

burmako (approved):

PR #1722, opened 2023-08-09
Add interpreter for InfeedOp

### LINKED ISSUE #1128: Add interpreter for InfeedOp

ghpvnist:

```
(I1) `token` is a `token`.
(I2) `infeed_config` is a constant of type `string`.
(C1) `0 < size(results)`.
(C2) `is_empty(result[:-1])` or `is_tensor(type(results[:-1]))`.
(C3) `is_token(type(results[-1]))`.
```

```
I1: a) `token` is not a `token`. (Covered by ODS).
I2: a) `infeed_config` is not a constant of type `string`. (Covered by ODS).
C1: a) `size(results) <= 0`.
C2: a) `is_empty(result[:-1]) = false` and `is_tensor(type(results[:-1])) = false`.
C3: a) `type(results[-1]) != token`.
```

```
C1: a) `size(results) <= 0`.
C2: a) `is_empty(result[:-1]) = false` and `is_tensor(type(results[:-1])) = false`.
C3: a) `type(results[-1]) != token`.
```

closes #1128

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -3575,20 +3575,20 @@ LogicalResult verifyInfeedOp(HloDialectInterface* dialect,
         resultTypes.size());
 
   // infeed_c2
-  for (size_t i = 0; i < resultTypes.size() - 1; ++i)
-    if (!resultTypes[i].isa<TensorType>())
+  for (auto resultType : results.drop_back().getTypes())
+    if (!resultType.isa<TensorType>())
       return emitOptionalError(
           location,
           "all elements of result types, except the last element, are expected "
           "to be of tensor type, but got ",
-          resultTypes[i]);
+          resultType);
 
   // infeed_c3
-  if (!dialect->isTokenType(resultTypes[resultTypes.size() - 1]))
+  if (!dialect->isTokenType(results.back().getType()))
     return emitOptionalError(location,
                              "last element of result types is expected to "
                              "be of token type, but got ",
-                             resultTypes[resultTypes.size() - 1]);
+                             results.back().getType());
 
   if (!layout.has_value()) return success();
   if (!layout.value())
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -3563,18 +3563,27 @@ LogicalResult verifyDynamicReshapeOp(std::optional<Location> location,
   return success();
 }
 
-// Checks that the result type is of the form `zero_or_more_type(s),
-// stablehlo::token`
 LogicalResult verifyInfeedOp(HloDialectInterface* dialect,
                              std::optional<Location> location,
                              std::optional<ArrayAttr> layout,
                              ValueRange results) {
   auto resultTypes = results.getType();
+  // infeed_c1
   if (resultTypes.empty())
     return emitOptionalError(
         location, "result is expected to be at least of size 1, but got ",
         resultTypes.size());
 
+  // infeed_c2
+  for (size_t i = 0; i < resultTypes.size() - 1; ++i)
+    if (!resultTypes[i].isa<TensorType>())
+      return emitOptionalError(
+          location,
+          "everything but the last element of result types is expected to "
+          "be of tensor type, but got ",
+          resultTypes[i]);
+
+  // infeed_c3
   if (!dialect->isTokenType(resultTypes[resultTypes.size() - 1]))
     return emitOptionalError(location,
                              "last element of result types is expected to "
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -3579,7 +3579,7 @@ LogicalResult verifyInfeedOp(HloDialectInterface* dialect,
     if (!resultTypes[i].isa<TensorType>())
       return emitOptionalError(
           location,
-          "everything but the last element of result types is expected to "
+          "all elements of result types except the last element is expected to "
           "be of tensor type, but got ",
           resultTypes[i]);
 
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -3579,8 +3579,8 @@ LogicalResult verifyInfeedOp(HloDialectInterface* dialect,
     if (!resultTypes[i].isa<TensorType>())
       return emitOptionalError(
           location,
-          "all elements of result types except the last element is expected to "
-          "be of tensor type, but got ",
+          "all elements of result types, except the last element, are expected "
+          "to be of tensor type, but got ",
           resultTypes[i]);
 
   // infeed_c3
--- a/stablehlo/reference/InterpreterOps.cpp
+++ b/stablehlo/reference/InterpreterOps.cpp
@@ -15,6 +15,8 @@ limitations under the License.
 
 #include "stablehlo/reference/InterpreterOps.h"
 
+#include <queue>
+
 #include "llvm/Support/ThreadPool.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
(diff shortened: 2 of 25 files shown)


burmako:
Let's iterate some more on the design for infeed/outfeed once I'm back from vacation.

GleasonK (changes requested):

stablehlo/reference/Ops.cpp
```diff
@@ -1215,6 +1219,21 @@ Tensor evalImagOp(const Tensor &operand, ShapedType resultType) {
...
+                                           Region &region, Scope &scope) {
+  if (!process)
+    llvm::report_fatal_error(
+        "infeed is only supported when run via interpreter.run_parallel");
+
+  auto mnemonic = process->infeed();
+  auto results = eval(region.getParentOfType<ModuleOp>()
+                          .lookupSymbol<func::FuncOp>(mnemonic)
+                          .getBody(),
+                      {}, process, &scope);
```
GleasonK, line 1232: Is there anything we should verify about these functions? Does a function with multiple return values work? Does each infeed function require that the function returns the same number of results? Will this error gracefully?

GleasonK (changes requested):

stablehlo/reference/Ops.cpp
ghpvnist: Yes, we should add some checks. Here are the verifications I've added:
1. function should return 1 value
2. function return type should be a tensor
3. infeed attribute should be null or non empty.

stablehlo/reference/InterpreterOps.cpp
```diff
@@ -90,6 +92,32 @@ LogicalResult RunParallelOp::verify() {
...
+      if (!func)
+        return emitOptionalError(getLoc(), "Function ", funcName, " not found");
+
+      if (func.getNumResults() != 1)
+        return emitOptionalError(getLoc(), "Function ", funcName,
+                                 " should return 1 tensor but returns ",
+                                 func.getNumResults());
+
+      for (auto type : func.getResultTypes())
+        if (!type.isa<ShapedType>())
```
GleasonK, line 114: Probably don't need a loop if we've verified that there is only one result?

Can use something like:

```
auto resultType = func.getResultTypes()[0];
if (!resultType.isa<ShapedType>()) error(..., resultType);
```

stablehlo/reference/InterpreterOps.cpp
```diff
@@ -90,6 +92,32 @@ LogicalResult RunParallelOp::verify() {
...
 
+  if (getInfeed()) {
+    if (getInfeed()->empty())
+      return emitOptionalError(
+          getLoc(), "infeed attribute is optional or should not be empty");
+
+    for (auto it = getInfeed()->begin(); it != getInfeed()->end(); ++it) {
+      auto funcName = (*it).dyn_cast<FlatSymbolRefAttr>().getAttr();
+      auto func =
+          (*this)->getParentOfType<ModuleOp>().lookupSymbol<func::FuncOp>(
```
GleasonK, line 103: `(*this)` --> `getOperation()` ?

GleasonK (changes requested):

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -114,11 +114,11 @@ void ProcessGrid::ThreadSafeQueue<T>::push(T input) {
 //===----------------------------------------------------------------------===//
 
 ProcessGrid::ProcessGrid(uint32_t numReplicas, uint32_t numPartitions,
-                         std::queue<StringAttr> *infeed)
+                         std::queue<StringAttr> &infeed)
     : numReplicas_(numReplicas), numPartitions_(numPartitions) {
-  while (!infeed->empty()) {
-    auto head = infeed->front();
-    infeed->pop();
+  while (!infeed.empty()) {
```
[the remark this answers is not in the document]
ghpvnist: After adding the constructor, things seemed a bit cramped, so I moved the definitions of ThreadSafeMap and ThreadSafeQueue to a separate namespace.

stablehlo/reference/InterpreterOps.cpp
```diff
@@ -92,16 +92,16 @@ LogicalResult RunParallelOp::verify() {
...
+    if (infeed->empty())
       return emitOptionalError(
           getLoc(), "infeed attribute is optional or should not be empty");
 
-    for (auto it = getInfeed()->begin(); it != getInfeed()->end(); ++it) {
-      auto funcName = (*it).dyn_cast<FlatSymbolRefAttr>().getAttr();
-      auto func =
-          (*this)->getParentOfType<ModuleOp>().lookupSymbol<func::FuncOp>(
-              funcName);
+    for (auto flatSymbolRefAttr : infeed->getAsRange<FlatSymbolRefAttr>()) {
```
GleasonK, line 100: Didn't know this method, awesome!

stablehlo/reference/ProcessGrid.h
```diff
@@ -31,31 +31,103 @@ limitations under the License.
 namespace mlir {
 namespace stablehlo {
 
+struct ProcessId;
+class RendezvousResult;
+
+namespace {
```
GleasonK, line 37: Anonymous namespace in a header file is rarely a good idea. Creates a new namespace per translation unit.

https://stackoverflow.com/questions/357564/uses-for-anonymous-namespaces-in-header-files

I think keeping in ProcessGrid is OK, the alternative would be to add a `detail` namespace, which tells users that these things are implementation details of process grid and should not be relied upon, this is also an OK option. We use `detail` namespace in `Tensor.h`.

GleasonK (changes requested):

stablehlo/reference/ProcessGrid.h
ghpvnist: I also thought about taking this out of ProcessGrid to something like ThreadSafeUtils.h. These are generic enough to be separated, but this is also exclusively used by ProcessGrid, and I don't anticipate any other uses of these classes. Would be good to add `detail` to be consistent for now though.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3539,18 +3539,27 @@ LogicalResult verifyDynamicReshapeOp(std::optional<Location> location,
...
                              ValueRange results) {
   auto resultTypes = results.getType();
+  // infeed_c1
   if (resultTypes.empty())
     return emitOptionalError(
         location, "result is expected to be at least of size 1, but got ",
         resultTypes.size());
 
+  // infeed_c2
+  for (size_t i = 0; i < resultTypes.size() - 1; ++i)
```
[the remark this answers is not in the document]
GleasonK: I believe you can use:

`for (auto resultType : results.drop_back().getTypes())`

and 

`results.back().getType()`

Consistent with the send/receive PR

stablehlo/reference/InterpreterOps.td
```diff
@@ -64,13 +66,15 @@ def Interpreter_RunParallelOp : Op<Interpreter_Dialect, "run_parallel", []> {
     Example:
     ```mlir
     %results:2 = "interpreter.run_parallel"() {
-      programs=[["foo"], ["bar"]]
+      infeed=[%infeed_queue0, @infeed_queue1]
```
GleasonK, line 69: typo `@infeed_queue0`

GleasonK (approved):
Last bit of feedback then LGTM

stablehlo/reference/ProcessGrid.h
GleasonK: Detail is fine. Prefer to not make things overly generic/reusable until there is a need for it

stablehlo/reference/InterpreterOps.td
ghpvnist: Thanks!

PR #1723, opened 2023-08-09
Add interpreter for OutfeedOp

### LINKED ISSUE #1130: Add interpreter for OutfeedOp

ghpvnist:

```
(I1) `inputs`: is a variadic number of tensors.
(I2) `token`: is a `token`.
(I3) `outfeed_config`: is a constant of type `string`.
(C1) `type(result) = token`.
```

```
I1: a) `inputs`: is not a variadic number of tensors. (Covered by ODS).
I2: a) `token`: is not a `token`. (Covered by ODS).
I3: a) `outfeed_config` is not a constant of type `string`. (Covered by ODS).
C1: a) `type(result) != token`.
```

```
C1: a) `type(result) != token`.
```

closes #1130

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -3698,6 +3698,10 @@ Semantics of `outfeed_config` is implementation-defined.
 |----------|---------|
 | `result` | `token` |
 
+#### Constraints
+
+* (C1) `type(result) = token`.
+
 #### Examples
 
 ```mlir
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -3701,10 +3701,12 @@ Semantics of `outfeed_config` is implementation-defined.
 #### Examples
 
 ```mlir
-%result = "stablehlo.outfeed"(%input0, %token) {
+%result = "stablehlo.outfeed"(%inputs0, %token) {
   outfeed_config = ""
-} : (tensor<3x3x3xi32>, !stablehlo.token) -> !stablehlo.token
+} : (tensor<2x2x2xi64>, !stablehlo.token) -> !stablehlo.token
 ```
+
+&nbsp;[More Examples](../stablehlo/tests/interpret_outfeed.mlir)
 
 ### pad
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -3698,10 +3698,6 @@ Semantics of `outfeed_config` is implementation-defined.
 |----------|---------|
 | `result` | `token` |
 
-#### Constraints
-
-* (C1) `type(result) = token`.
-
 #### Examples
 
 ```mlir
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -1028,16 +1028,16 @@ def StableHLO_OutfeedOp : StableHLO_Op<"outfeed",
 
     Example:
     ```mlir
-    %result = "stablehlo.outfeed"(%input0, %token) {
+    %result = "stablehlo.outfeed"(%inputs0, %token) {
       outfeed_config = ""
-    } : (tensor<3x3x3xi32>, !stablehlo.token) -> !stablehlo.token
+    } : (tensor<2x2x2xi64>, !stablehlo.token) -> !stablehlo.token
     ```
   }];
 
   let arguments = (ins
-    Variadic<HLO_Tensor>:$inputs,
-    HLO_Token:$token,
-    DefaultValuedStrAttr<StrAttr, "">:$outfeed_config
+    Variadic<HLO_Tensor>:$inputs, /*outfeed_i1*/
+    HLO_Token:$token, /*outfeed_i2*/
+    DefaultValuedStrAttr<StrAttr, "">:$outfeed_config /*outfeed_i3*/
   );
   let results = (outs HLO_Token);
 }
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -1018,7 +1018,7 @@ def StableHLO_InfeedOp : StableHLO_Op<"infeed", []> {
 }
 
 def StableHLO_OutfeedOp : StableHLO_Op<"outfeed",
-    [DeclareOpInterfaceMethods<InferTypeOpInterface> /*outfeed_c1*/]> {
+    [DeclareOpInterfaceMethods<InferTypeOpInterface>]> {
   let summary = "Outfeed operation";
   let description = [{
     Writes `inputs` to the outfeed and produces a `result` token.
@@ -1028,9 +1028,8 @@ def StableHLO_OutfeedOp : StableHLO_Op<"outfeed",
 
     Example:
     ```mlir
-    %result = "stablehlo.outfeed"(%inputs0, %token) {
-      outfeed_config = ""
-    } : (tensor<2x2x2xi64>, !stablehlo.token) -> !stablehlo.token
+    %result = "stablehlo.outfeed"(%inputs0, %token) :
+        (tensor<2x2x2xi64>, !stablehlo.token) -> !stablehlo.token
     ```
   }];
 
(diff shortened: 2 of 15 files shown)


stablehlo/dialect/StablehloOps.td
```diff
@@ -1028,16 +1028,16 @@ def StableHLO_OutfeedOp : StableHLO_Op<"outfeed",
...
-    } : (tensor<3x3x3xi32>, !stablehlo.token) -> !stablehlo.token
+    } : (tensor<2x2x2xi64>, !stablehlo.token) -> !stablehlo.token
     ```
   }];
 
   let arguments = (ins
-    Variadic<HLO_Tensor>:$inputs,
-    HLO_Token:$token,
-    DefaultValuedStrAttr<StrAttr, "">:$outfeed_config
+    Variadic<HLO_Tensor>:$inputs, /*outfeed_i1*/
```
burmako, line 1038: I think we'd want to update constraints to say that all the inputs except last are tensors and the last input is a token? Something like:

```
(C1) `0 < size(results)`.
(C2) `is_tensor(results[:-1])`.
(C3) `is_token(results[-1])`.
```

stablehlo/reference/Ops.cpp
```diff
@@ -1228,6 +1233,11 @@ Tensor evalOrOp(const Tensor &lhs, const Tensor &rhs, ShapedType resultType) {
   return result;
 }
 
+Token evalOutfeedOp(ArrayRef<Tensor> inputs, Token token, Process *process) {
+  process->outfeed(inputs);
```
burmako, line 1237: Need a check that process is not nullptr.

stablehlo/reference/Process.h
```diff
@@ -39,6 +39,9 @@ class Process {
   ProcessGroups crossPartition(
       SmallVector<SmallVector<uint32_t>> partitionGroups);
 
+  /// See `ProcessGrid::outfeed`.
+  void outfeed(ArrayRef<Tensor> inputs);
```
burmako, line 43: Let's order methods alphabetically? This will make it easier to decide whether to put new methods.

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -85,6 +85,10 @@ ProcessGroups ProcessGrid::crossPartition(
   return processGroups;
 }
 
+void ProcessGrid::outfeed(ArrayRef<Tensor> inputs) {
+  outfeed_.push(llvm::to_vector(inputs));
```
burmako, line 89: `outfeed_.emplace(inputs)`?

stablehlo/tests/interpret_outfeed.mlir
```diff
@@ -0,0 +1,19 @@
+// RUN: stablehlo-translate --interpret -split-input-file %s
+
+module @distribution_ops {
+  func.func public @outfeed(%inputs0 : tensor<2x2x2xi64>, %token : !stablehlo.token) -> !stablehlo.token {
+    %result = "stablehlo.outfeed"(%inputs0, %token) {
+      outfeed_config = ""
```
burmako, line 6: If outfeed_config is empty, we probably shouldn't be specifying it in the first place?

stablehlo/tests/interpret_outfeed.mlir
```diff
@@ -0,0 +1,19 @@
...
+    } : (tensor<2x2x2xi64>, !stablehlo.token) -> !stablehlo.token
+    func.return %result : !stablehlo.token
+  }
+  func.func public @main() {
+    %inputs0 = stablehlo.constant dense<[[[1, 2], [3, 4]],
+                                         [[5, 6], [7, 8]]]> : tensor<2x2x2xi64>
+    %token = stablehlo.after_all : !stablehlo.token
+    %result = "interpreter.run_parallel"(%inputs0, %token) {
+      programs=[["outfeed"]]
+    } : (tensor<2x2x2xi64>, !stablehlo.token) -> !stablehlo.token
```
burmako, line 16: I think we'd also want a way to test the contents of the outfeed.

burmako (changes requested):

stablehlo/dialect/StablehloOps.td
ghpvnist: Since `outfeed` only returns a token, I've added (C3) in your proposal.

docs/spec.md
```diff
@@ -3698,14 +3698,20 @@ Semantics of `outfeed_config` is implementation-defined.
 |----------|---------|
 | `result` | `token` |
 
+#### Constraints
+
+* (C1) `type(result) = token`.
```
burmako, line 3703: Let's drop this constraint since it's already covered by Outputs above?

stablehlo/dialect/StablehloOps.td
burmako: Apologies, I was too hasty when writing the previous review and mixed up InfeedOp and OutfeedOp. Please disregard this comment.

stablehlo/reference/Process.h
burmako: I see that methods are now ordered in ProcessGrid.h, but I don't think they are in this file.

stablehlo/reference/InterpreterOps.cpp
```diff
@@ -123,6 +123,10 @@ SmallVector<InterpreterValue> evalRunParallelOp(
 
   SmallVector<InterpreterValue> results;
   for (auto &future : futures) results.append(future.get());
+
+  while (!processGrid.outfeedEmpty())
+    for (auto outfeed : processGrid.popOutfeed()) results.push_back(outfeed);
```
burmako, line 128: I don't think this design is going to work because it is not going to always be possible to statically determine the number of entries expected in the outfeed (that looks equivalent to solving the halting problem).

burmako, line 128: This made me realize that RunParallelOp::verify should also verify the number of results, not just the number of inputs.

stablehlo/reference/ProcessGrid.h
```diff
@@ -88,13 +89,23 @@ class ProcessGrid {
...
+
+  /// Inserts `inputs` to StableHLO `outfeed`.
+  void outfeed(ArrayRef<Tensor> inputs);
+
+  /// Returns whether the outfeed queue is empty.
+  bool outfeedEmpty();
+
+  /// Pops the first entry in the outfeed queue and returns its contents. The
+  /// method fails with fatal error if the queue is empty.
+  SmallVector<Tensor> popOutfeed();
```
burmako, line 107: Can we perhaps expose `getOutfeed` directly, so that reading it requires only 1 API call and doesn't involve mutating the underlying queue?

stablehlo/reference/InterpreterOps.cpp
burmako: Let's open a ticket for this, leave a TODO in code and move on I think. The changes in this PR are already valuable on their own, and we can iterate on testing outfeeds in follow-up PRs.

burmako (changes requested):

stablehlo/reference/InterpreterOps.cpp
ghpvnist: Opened #1725

stablehlo/reference/ProcessGrid.h
ghpvnist: Actually I realized that this operation is not thread safe so we should probably keep it the way it is and introduce a separate lock for it.

burmako: That's a good catch! As far `getOutfeed` goes, it could take a lock, copy the queue into a separate SmallVector and return the copy. But this is moot at this point because testing will be happening in a follow-up, so this can be figured out then.

burmako (approved):

Issue #666, opened 2022-12-01
Consider unifying `channel_type` and `is_host_transfer` for send/Recv Op


sdasgup3:
### Request description

From the explanations given at
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/hlo/ir/hlo_instruction.h#L859, we have

>> If is_host_transfer is true, then this Send operation transfers data to the host.

and from https://github.com/tensorflow/tensorflow/blob/2bb03ca841bf31c2fc57ae1aca6d14127ca89385/tensorflow/compiler/xla/hlo/ir/hlo_instruction.h#L873, we have

>>  If is_host_transfer is true, then this Recv operation transfers data from the host.

Together we have  channel_handle::channelType (https://github.com/tensorflow/tensorflow/blob/2bb03ca841bf31c2fc57ae1aca6d14127ca89385/tensorflow/compiler/xla/xla_data.proto#L434) which could have the following relevant (to this discussion) values among others. 

  * `DEVICE_TO_HOST`: A channel for sending data from the device to the host. Can only be used with a `stablehlo.send` operation.
  * `HOST_TO_DEVICE`: A channel for sending data from the host to the device. Can only be used with a `stablehlo.recv` operation.

It seems `is_host_transfer` is serving the same purpose as  `channelTypes`. So how about unifying the two in the [definition of send/recv ops](https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/StablehloOps.td#L1042
)?

A thought: It seems that none of the collective and point-to-point communication ops really need this channel_type. Can we remove it.

PR #1732, opened 2023-08-12
Add interpreter for SendOp and RecvOp

### LINKED ISSUE #1132: Add interpreter for RecvOp

### LINKED ISSUE #1137: Add interpreter for SendOp

ghpvnist:
SendOp:

```
(I1) `inputs` is a variadic number of tensors.
(I2) `token` is a `token`.
(I3) `channel_id` is a constant of type `si64`.
(I4) `channel_type` is a enum of `DEVICE_TO_DEVICE` and `DEVICE_TO_HOST`.
(I5) `is_host_transfer` is a constant of type `i1`.
(C1) `channel_type` is defined as:
* `DEVICE_TO_HOST` if `is_host_transfer = true`,
* `DEVICE_TO_DEVICE` otherwise.
```

```
I1: a) `inputs` is not a variadic number of tensors. (Covered by ODS).
I2: a) `token` is not a `token`. (Covered by ODS).
I4: a) `channel_type` is not a enum of `DEVICE_TO_DEVICE` and `DEVICE_TO_HOST`.
I5: a) `is_host_transfer` is not a constant of type `i1`. (Covered by ODS).
C1: a) `channel_type` is not defined as:
* `DEVICE_TO_HOST` if `is_host_transfer = true`,
* `DEVICE_TO_DEVICE` otherwise.
```

```
I4: a) `channel_type` is not a enum of `DEVICE_TO_DEVICE` and `DEVICE_TO_HOST`.
C1: a) `channel_type` is not defined as:
* `DEVICE_TO_HOST` if `is_host_transfer = true`,
* `DEVICE_TO_DEVICE` otherwise.
```

closes #1137

RecvOp:

```
(I1) `token` is a `token`.
(I2) `channel_id` is a constant of type `si64`.
(I3) `channel_type` is a enum of `DEVICE_TO_DEVICE` and `HOST_TO_DEVICE`.
(I4) `is_host_transfer` is a constant of type `i1`.
(C1) `channel_type` is defined as:
* `HOST_TO_DEVICE` if `is_host_transfer = true`,
* `DEVICE_TO_DEVICE` otherwise.
(C2) `0 < size(results)`.
(C3) `is_empty(result[:-1])` or `is_tensor(type(results[:-1]))`.
(C4) `is_token(type(results[-1]))`.
```

```
I1: a) `token` is not a `token`. (Covered by ODS).
I2: a) `channel_id` is not a constant of type `si64`. (Covered by ODS).
I3: a) `channel_type` is not a enum of `DEVICE_TO_DEVICE` and `HOST_TO_DEVICE`.
I4: a) `is_host_transfer` is not a constant of type `i1`. (Covered by ODS).
C1: a) `channel_type` is not defined as:
* `HOST_TO_DEVICE` if `is_host_transfer = true`,
* `DEVICE_TO_DEVICE` otherwise.
C2: a) `size(results) <= 0`.
C3: a) `is_empty(result[:-1]) = false` and `is_tensor(type(results[:-1])) = false`.
C4: a) `is_token(type(results[-1])) = false`.
```

```
I3: a) `channel_type` is not a enum of `DEVICE_TO_DEVICE` and `HOST_TO_DEVICE`.
C1: a) `channel_type` is not defined as:
* `HOST_TO_DEVICE` if `is_host_transfer = true`,
* `DEVICE_TO_DEVICE` otherwise.
C2: a) `size(results) <= 0`.
C3: a) `is_empty(result[:-1]) = false` and `is_tensor(type(results[:-1])) = false`.
C4: a) `is_token(type(results[-1])) = false`.
```

Notes:
  * Also fixes bugs in the compatibility tests. The changes made here is an exception to the regular process (See [compatibility.md](https://github.com/openxla/stablehlo/blob/main/docs/compatibility.md#out-of-scope)) since the verification was not in sync with the spec. Other than the bug fixes, the remaining changes were autogenerated using the script in [compatibility.md](https://github.com/openxla/stablehlo/blob/main/docs/compatibility.md).

closes #1132


docs/spec.md
```diff
@@ -3942,27 +3942,28 @@ separate outputs to improve clarity
 
 | Name      | Type                                                    | Constraints |
 |-----------|---------------------------------------------------------|-------------|
-| `results` | variadic number of tensors, quantized tensors or tokens | (C2), (C3)  |
+| `results` | variadic number of tensors, quantized tensors or tokens | (C2-C4)     |
 
 #### Constraints
 
 * (C1) `channel_type` is defined as:
```
GleasonK, line 3949: Is there any plan to back this value with an enum? It looks like we don't document what value should be used if you wanted `HOST_TO_DEVICE` for example. An int64_t enum in C++ should help resolve that issue for StableHLO C++ users.

Additionally, we could make the pretty printer/parser for channel handle attr accept the string values for valid specced enums.

burmako: There is no such plan at the moment, but there a plan to remove this value altogether: https://github.com/openxla/stablehlo/issues/666 :)

stablehlo/tests/stablehlo_legalize_to_vhlo.0_10_0.mlir
```diff
@@ -2159,6 +2160,13 @@ func.func @type_f8E4M3FNUZ(%arg0: tensor<f8E4M3FNUZ>, %arg1: tensor<f8E4M3FNUZ>)
 }
 // CHECK-LABEL: "type_f8E4M3FNUZ"
 
+func.func @type_f8E4M3B11FNUZ(%arg0: tensor<f8E4M3B11FNUZ>, %arg1: tensor<f8E4M3B11FNUZ>) -> tensor<f8E4M3B11FNUZ> {
```
GleasonK, line 2163: Why is this added?

GleasonK (changes requested):

stablehlo/tests/stablehlo_legalize_to_vhlo.0_10_0.mlir
GleasonK: Confirmed that this test passes once this block is removed.


Please regenrate with:

```
stablehlo-translate --serialize --target=0.10.0 --strip-debuginfo stablehlo/tests/stablehlo_legalize_to_vhlo.0_10_0.mlir > stablehlo/tests/stablehlo_legalize_to_vhlo.0_10_0.mlir.bc
```

And update the run commands of all `0_X_0.mlir` files to what they used to be.

GleasonK (changes requested):

stablehlo/tests/stablehlo_legalize_to_vhlo.0_10_0.mlir
ghpvnist: I'm not sure how this was added either.. Updated.

stablehlo/tests/stablehlo_legalize_to_vhlo.0_10_0.mlir
```diff
@@ -1,7 +1,8 @@
-// RUN: stablehlo-opt --mlir-print-op-generic %s.bc | FileCheck %s
-// RUN: stablehlo-translate --deserialize %s.bc | stablehlo-translate --serialize --target=0.10.0 | stablehlo-opt --mlir-print-op-generic | FileCheck %s
-// RUN: diff <(stablehlo-translate --deserialize %s.bc | stablehlo-opt) <(stablehlo-opt --strip-debuginfo %s)
-// RUN: diff %s.bc <(stablehlo-translate --serialize --target=0.10.0 --strip-debuginfo %s)
+// RUN: stablehlo-opt --stablehlo-legalize-to-vhlo --mlir-print-op-generic --split-input-file %s | FileCheck %s
```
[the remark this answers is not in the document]
ghpvnist: The command seemed to have failed due to a typo in the instruction. I sent #1739 to fix it.

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -154,6 +154,25 @@ void ProcessGrid::outfeed(ArrayRef<Tensor> inputs) {
   outfeed_.emplace(inputs);
 }
 
+SmallVector<Tensor> ProcessGrid::recv(ChannelId channelId,
+                                      ProcessId processId) {
+  {
+    std::lock_guard<std::mutex> lock(sendRecvLocks_[channelId]);
```
GleasonK, line 160: This will probably have the same issue we hit in https://github.com/openxla/stablehlo/pull/1737 right?

We could make a new datatype `MutexMap` which wraps this map and has a synchronized getter method.

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -183,5 +202,24 @@ RendezvousResult ProcessGrid::rendezvous(ProcessGroup processGroup,
   }
 }
 
+void ProcessGrid::send(ArrayRef<Tensor> inputs, ChannelId channelId,
+                       ProcessId processId) {
+  {
+    std::unique_lock<std::mutex> lock(sendRecvLocks_[channelId]);
```
GleasonK, line 208: Same here

GleasonK (changes requested):

stablehlo/reference/ProcessGrid.cpp
GleasonK: Actually wait.. do we know the number of mutexes we will need at creation time? We can instead use a vector of mutexes and index by channel id

ghpvnist: Thanks for following up in this PR! I haven't made the changes following up from #1737 yet.

> do we know the number of mutexes we will need at creation time?

Technically possible (I think), but not feasible. We can have multiple send/recv (or any other distribution ops) with different `channelId`s within a single process. This would mean going through op by op to obtain the set of used `channelId`s.

> We can instead use a vector of mutexes and index by channel id

This won't work since `channelId` can be arbitrary (i.e. does not have to increment 0, 1, 2, etc. unlike `channelType` which is enumerated), and map would be a better fit for this reason.

ghpvnist: Updated.

burmako:
Will review once I'm back - this will need some careful thinking given that this PR involves a non-trivial amount of multithreaded APIs.

stablehlo/reference/ProcessGrid.h
```diff
@@ -154,12 +172,38 @@ class ProcessGrid {
...
+
+  /// Internal storage used to implement `send` and `recv`.
+  /// `send` can write its data to the channel with ChannelId once the ops are
+  /// ready to communicate. `recv` receives data from the same channel once the
+  /// data is ready to read.
+  std::map<ChannelId, RendezvousResult> sendRecvChannels_;
+
+  /// Synchronization primitive used to manage concurrent access to
+  /// `sendRecvChannels_`.
+  std::map<ChannelId, std::mutex> sendRecvChannelLocks_;
```
GleasonK, line 194: Should any of these be `ThreadSafeMap`?

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -149,11 +149,35 @@ std::mutex &ProcessGrid::getRendezvousLock(ProcessGroup processGroup,
...
+  }
+  sendRecvConditions_[channelId].notify_all();
+  {
+    std::unique_lock<std::mutex> lock(getSendRecvOmniLock(channelId));
+    if (!sendRecvConditions_[channelId].wait_for(
+            lock, std::chrono::seconds(3),
+            [&] { return sendRecvChannels_[channelId].size() != 0; }))
+      llvm::report_fatal_error("recv timed out");
+  }
+  return sendRecvChannels_[channelId].getSortedTensors();
```
GleasonK, line 178: Should this access be lock-guarded? Same for `sendRecvConditions_[channelId].notify_all();` Will need to think more about these unguarded accesses of non-thread-safe datatypes.

GleasonK (changes requested):
Providing initial feedback, can you make sure this has a clean TSAN before assigning back to me. I still need to think a bit harder about the communication between send/recv, but overall structure looks good.

stablehlo/reference/Ops.cpp
```diff
@@ -483,6 +483,13 @@ SmallVector<InterpreterValue> eval(
       auto operand = scope.findTensor(realOp.getOperand());
       auto result = evalRealOp(operand, realOp.getType());
       scope.add(realOp.getResult(), result);
+    } else if (auto recvOp = dyn_cast<RecvOp>(op)) {
+      auto token = scope.findToken(recvOp.getToken());
+      ChannelId channelId = 0;
+      if (auto channelHandle = recvOp.getChannelHandleAttr())
```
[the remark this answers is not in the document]
ghpvnist: I updated the code to use the version without `-Attr` suffix. I think shorter names are more readable for the same function.

stablehlo/reference/Ops.cpp
```diff
@@ -636,6 +643,14 @@ SmallVector<InterpreterValue> eval(
       auto onFalse = scope.findTensor(selectOp.getOnFalse());
       auto result = evalSelectOp(pred, onTrue, onFalse, selectOp.getType());
       scope.add(selectOp.getResult(), result);
+    } else if (auto sendOp = dyn_cast<SendOp>(op)) {
+      auto inputs = scope.findTensors(sendOp.getInputs());
+      auto token = scope.findToken(sendOp.getToken());
+      ChannelId channelId = 0;
+      if (auto channelHandle = sendOp.getChannelHandleAttr())
```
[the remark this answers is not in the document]
ghpvnist: This makes no difference in the interpreter.

stablehlo/reference/ProcessGrid.h
ghpvnist: Rebased to HEAD and updated the commit with recent addition of `ThreadSafeMap`. PTAL.

stablehlo/reference/ProcessGrid.cpp
ghpvnist: Rebased to new commit with a few changes.

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -162,6 +197,21 @@ ProcessGroups ProcessGrid::flattenedIds(
 
 void ProcessGrid::outfeed(ArrayRef<Tensor> inputs) { outfeed_.push(inputs); }
 
+SmallVector<Tensor> ProcessGrid::recv(ChannelId channelId,
+                                      ProcessId processId) {
+  sendRecvReady_.insert(channelId);
+  sendRecvConditions_[channelId].notify_one();
+
+  auto &state = sendRecvChannels_[channelId];
+  std::unique_lock<std::mutex> lock(state.mutex);
```
GleasonK, line 206: send uses:

```
std::unique_lock<std::mutex> lock(sendRecvChannels_[channelId].mutex);
```

Maybe pick one representation and use it both places for consistency

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -162,6 +197,21 @@ ProcessGroups ProcessGrid::flattenedIds(
 
 void ProcessGrid::outfeed(ArrayRef<Tensor> inputs) { outfeed_.push(inputs); }
 
+SmallVector<Tensor> ProcessGrid::recv(ChannelId channelId,
+                                      ProcessId processId) {
+  sendRecvReady_.insert(channelId);
+  sendRecvConditions_[channelId].notify_one();
```
GleasonK, line 203: Is there an assumption that `send` will be waiting by the time this code is reached? Wondering if this is a safe assumption to make. I.e. is it possible that this notify_one either notifies nothing, or notifies the recv thread itself, both of which would cause deadlock/timeout

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -228,5 +278,19 @@ std::shared_ptr<RendezvousResult const> ProcessGrid::rendezvous(
...
+                       ProcessId processId) {
+  std::unique_lock<std::mutex> lock(sendRecvChannels_[channelId].mutex);
+  if (!sendRecvConditions_[channelId].wait_for(
+          lock, std::chrono::seconds(3), [&] {
+            return sendRecvReady_.find(channelId) != sendRecvReady_.end();
+          }))
+    llvm::report_fatal_error("send timed out");
+
+  sendRecvChannels_[channelId].result = llvm::to_vector(inputs);
+  sendRecvReady_.erase(channelId);
```
GleasonK, line 291: Can multiple processes be receiving on the same channel? This setup currently only handles 1:1 communication. If this is how XLA handles these then this is OK

stablehlo/reference/ProcessGrid.cpp
GleasonK: Can test this by adding a thread sleep for a second to the send impl.

One possible way to avoid this could be a separate `sendCondition` and `recvCondition`. This doesn't resolve the "notifies nothing" potential issue, but does avoid a self-wakeup if that turns out to be a problem

GleasonK (changes requested):

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -228,5 +278,19 @@ std::shared_ptr<RendezvousResult const> ProcessGrid::rendezvous(
   return result;
 }
 
+void ProcessGrid::send(ArrayRef<Tensor> inputs, ChannelId channelId,
+                       ProcessId processId) {
+  std::unique_lock<std::mutex> lock(sendRecvChannels_[channelId].mutex);
+  if (!sendRecvConditions_[channelId].wait_for(
+          lock, std::chrono::seconds(3), [&] {
+            return sendRecvReady_.find(channelId) != sendRecvReady_.end();
```
GleasonK, line 286: Can we add a contains method? This code currently:
- Gets mutex, calls begin, releases mutex
- Gets mutex, calls end, releases mutex

There is no guarantee (very unlikely though) that the set is unmodified in the meantime? Having a contains function should make this a single mutex event

GleasonK: Also curious if we should check if receive is ready prior to waiting? In case the `recv` thread reached this point first?

ghpvnist: Note that the `wait` function will release the lock when the condition is not satisfied. If send happens first, it will lock, and wait (release the lock). Then recv will then insert to `sendRecvReady_` and notify, at which point send's wait condition is satisfied. send will append to result and notify, at which point recv similarly will reacquire the lock and see that the condition is satisfied and return.

stablehlo/reference/ProcessGrid.cpp
ghpvnist: There should only be 1:1 communication happening within a channel, otherwise it is undefined.

stablehlo/reference/ProcessGrid.cpp
ghpvnist: There are two cases:

1. recv arrives first:
* Insert `channelId` to `sendRecvReady_`.
* `notify_one` notifies nothing and locks `sendRecvChannels_[channelId].mutex` or notifies send to continue.
  * If `notify_one` notifies nothing and recv will wait. Send will then gain control of lock and see that `sendRecvReady_.find(channelId) != sendRecvReady_.end()` condition is satisfied and move on (releasing the lock), and `sendRecvChannels_[channelId].result` is populated and notify, at which point recv condition is also satisfied.
  * If `notify_one` notifies send, `sendRecvChannels_[channelId].result` is populated and notify, at which point recv condition is also satisfied.
2. send arrives first:
* Locks `sendRecvChannels_[channelId].mutex` and wait (release the lock). Recv inserts to `sendRecvReady_` and `notify_one` which wakes send and the condition is satisfied.

I think the "notify nothing" part is okay to have, and it won't cause deadlocks.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -1330,19 +1330,26 @@ LogicalResult OutfeedOp::inferReturnTypes(
 
 LogicalResult SendOp::inferReturnTypes(
     MLIRContext* context, std::optional<Location> location, ValueRange operands,
-    DictionaryAttr, OpaqueProperties, RegionRange,
+    DictionaryAttr attributes, OpaqueProperties properties, RegionRange regions,
     SmallVectorImpl<Type>& inferredReturnTypes) {
+  SendOp::Adaptor adaptor(operands, attributes, properties, regions);
+  bool isDeviceToDevice = adaptor.getChannelHandle().getType() == 1;
```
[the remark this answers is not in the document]
ghpvnist: I think the PR description is okay. The enum type part is covered by ODS, and the value is checked in the verifier and noted as such in the description.

stablehlo/reference/ProcessGrid.cpp
GleasonK: As long as `wait_for` doesn't block if the precondition is true when the statement is reached, this probably isn't an issue, which seems to be the case

GleasonK: Let's still move the lock in recv to the top of the method. That way it is guarantees that either send gets the lock first, and joins the waiting queue, or recv gets the lock first and send's wair_for confition is guaranteed to be true by the time execution reaches there

stablehlo/reference/ProcessGrid.h
```diff
@@ -212,21 +255,44 @@ class ProcessGrid {
                                                      ProcessId processId,
                                                      const Tensor &operand);
 
+  /// Sends `inputs` to a channel with `channelId`.
+  /// The channel with `channel_id` is emptied before the receiving process can
```
GleasonK, line 259: `channelId` also?

stablehlo/dialect/StablehloOps.cpp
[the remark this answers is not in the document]
GleasonK: > The enum type part is covered by ODS

Only that the value is a signed integer is covered by ODS, i.e. `type=-1` is valid under ODS, but not a valid enum value (since there is no enum).

`let parameters = (ins "int64_t":$handle, "int64_t":$type);`

Saying covered by C1 is also how the implementation works, in that invalid integer values ("enums") are only caught by C1 and throw the same error as C1, this is different from other "covered by ODS" constraints which error in generated op verification code

stablehlo/reference/ProcessGrid.h
```diff
@@ -188,6 +224,13 @@ class ProcessGrid {
   /// Inserts `inputs` to StableHLO `outfeed`.
   void outfeed(ArrayRef<Tensor> inputs);
 
+  /// Receives data from a channel with `channelId` and returns the data.
+  /// `recv` has to be called first before `send` to indicate to the sending
+  /// process that the receiver is ready to receive data. The process then waits
+  /// until there is data in the channel. The data in the channel with
+  /// `channel_id` is returned.
```
GleasonK, line 231: `channelId`?

stablehlo/reference/ProcessGrid.h
```diff
@@ -212,21 +255,44 @@ class ProcessGrid {
...
   const uint32_t numReplicas_;
 
   /// StableHLO `num_partitions`.
   const uint32_t numPartitions_;
 
-  /// Interal queue of strings which represents `func::FuncOp` mnemonic that
+  /// Internal queue of strings which represents `func::FuncOp` mnemonic that
   /// returns a vector of Tensor. The function name is stored instead of the
   /// vector of tensors to save memory. See `ThreadSafeQueue`.
   detail::ThreadSafeQueue<StringAttr> infeed_;
```
GleasonK, line 274: I thought we flipped this to `FlatSymbolRefAttrs`? Any reason to use string here?

stablehlo/reference/ProcessGrid.h
```diff
@@ -212,21 +255,44 @@ class ProcessGrid {
...
 
   /// See `ThreadSafeQueue`.
   detail::ThreadSafeQueue<SmallVector<Tensor>> outfeed_;
```
GleasonK, line 277: Could you also add a description for outfeed here? All the other member variables have great descriptions of how they're used

stablehlo/tests/infer_stablehlo.mlir
```diff
@@ -664,6 +664,19 @@ func.func @scatter_bounds(%input_tensor: tensor<200x?x?xf32, #stablehlo.bounds<?
 
 // -----
 
+// CHECK-LABEL: func @send
```
GleasonK, line 667: Do we need a recv test?

GleasonK (changes requested):
Mostly nits at this point

stablehlo/reference/ProcessGrid.cpp
ghpvnist: According to the [cpp reference](https://en.cppreference.com/w/cpp/thread/condition_variable/notify_one) for `notify_one`,

> The notifying thread does not need to hold the lock on the same mutex as the one held by the waiting thread(s); in fact doing so is a pessimization, since the notified thread would immediately block again, waiting for the notifying thread to release the lock.

Some implementations will recognize this and transfer the ownership, but otherwise notifying the other thread while having control of the lock would immediately leave the other thread waiting again.

Here is a [diagram](https://www.tensorflow.org/xla/operation_semantics#send) that may complement my explanation, but essentially, since send is immediately trying to acquire the lock, if it has the lock, it will proceed to wait (and release lock). If it doesn't have the lock, it will wait until the lock is free to acquire. This allows recv to freely insert `channelId` to `sendRecvReady_` at any time, and since send is already already waiting on the condition_variable.

Before I go deeper into the explanation (at this point this little code snippet seems to cause more confusion!), I'm questioning whether the `notify_one` is actually needed. If we remove `notify_one` from both functions, it might be easier to reason about the parallel execution and I think it's unnecessary. Before I make this comment even longer, can you look at the updated code and let's discuss :)

stablehlo/tests/infer_stablehlo.mlir
ghpvnist: Type inference for recv is infeasible since we don't know what value we are "receiving"

stablehlo/reference/ProcessGrid.h
ghpvnist: Mostly because `StringType` is a supported StableHLO type and `StringAttr` seems to be the best representation for this whereas `FlatSymbolRefAttrs` is more closely tied to MLIR related things. Since we now use `FlatSymbolRefsAttr` in the td file, function names are also verified.

stablehlo/reference/ProcessGrid.h
ghpvnist: Thanks!

stablehlo/reference/ProcessGrid.h
ghpvnist: Thanks!

stablehlo/reference/ProcessGrid.cpp
ghpvnist: But I agree that it doesn't hurt to be extra cautious and makes reading the two implementation easier to reason about:

There are now two cases:
1. Recv acquires lock first.
2. Send acquires lock first.

GleasonK: LGTM! I had seen that clause before about "does not need to hold the lock on the same mutex as the one held by the waiting thread(s)", but I think this is a fair means of doing it. Our other ops notify the same way so this is consistent. If we end up finding this to be an issue on some hardware/compiler (hasn't happened yet), then we can re-think a strategy and fix them all together

stablehlo/reference/ProcessGrid.h
GleasonK: Valid reasoning! The benefit of a symbol ref vs string isn't too big while interpreting, so using StringType for the sake of it being spec'ed makes sense.

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -77,6 +77,8 @@ SmallVector<Tensor> RendezvousResult::getSortedTensors() const {
       llvm::map_range(result_, [](const auto &pair) { return pair.second; }));
 }
 
+size_t RendezvousResult::size() const { return result_.size(); }
```
GleasonK: Is this used anywhere? If not lets remove and also remove the size method from thread safe map.

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -182,6 +204,20 @@ void ProcessGrid::outfeed(ArrayRef<Tensor> inputs) {
...
+  std::unique_lock<std::mutex> lock(sendRecvChannels_[channelId].mutex);
+  sendRecvReady_.insert(channelId);
+  sendRecvConditions_[channelId].notify_one();
+
+  if (!sendRecvConditions_[channelId].wait_for(
+          lock, std::chrono::seconds(3),
+          [&] { return !sendRecvChannels_[channelId].result.empty(); }))
+    llvm::report_fatal_error("recv timed out");
+
+  return sendRecvChannels_[channelId].result;
```
GleasonK, line 218: Do we need to re-empty the result once we're done for subsequent recv's? Perhaps use move, or move this value to a temp var and clear it?

stablehlo/reference/Ops.cpp
```diff
@@ -1547,6 +1560,14 @@ Tensor evalRealOp(const Tensor &operand, ShapedType resultType) {
   return result;
 }
 
+SmallVector<InterpreterValue> evalRecvOp(Token token, ChannelId channelId,
+                                         Process *process) {
+  SmallVector<InterpreterValue> results;
+  for (const auto &tensor : process->recv(channelId)) results.push_back(tensor);
+  results.push_back(token);
```
GleasonK, line 1567: What happens if send/recv have different types, i.e. send puts 3 tensors, but receive expects 4? Does this type checking happen elsewhere in the interpreter?

GleasonK (commented):
Let's resolve open comments then good to approve

stablehlo/reference/ProcessGrid.cpp
GleasonK: (outdated, removed)

stablehlo/reference/Ops.cpp
```diff
@@ -560,6 +560,13 @@ SmallVector<InterpreterValue> eval(
       auto operand = scope.findTensor(realOp.getOperand());
       auto result = evalRealOp(operand, realOp.getType());
       scope.add(realOp.getResult(), result);
+    } else if (auto recvOp = dyn_cast<RecvOp>(op)) {
+      auto token = scope.findToken(recvOp.getToken());
+      ChannelId channelId = 0;
```
GleasonK, line 565: How does the spec go about handling default values? Looks like ChannelId has a default value of 0 in the interpreter.. should this be documented elsewhere, or do we tend to leave default values out of the spec?

stablehlo/tests/interpret_send_recv.mlir
```diff
@@ -0,0 +1,28 @@
...
+
+module @distribution_ops {
+  func.func public @send(%operand : tensor<2x2xi64>, %token : !stablehlo.token) -> !stablehlo.token {
+    %result = "stablehlo.send"(%operand, %token) {
+      channel_handle = #stablehlo.channel_handle<handle = 1, type = 2>,
+      is_host_transfer = true
+    } : (tensor<2x2xi64>, !stablehlo.token) -> !stablehlo.token
+    return %result : !stablehlo.token
+  }
+  func.func public @recv(%token : !stablehlo.token) -> (tensor<2x2xi64>, !stablehlo.token) {
```
GleasonK, line 11: Other test ideas:

- single program that does **sequential** `send/receive`, `send/receive` on the **same** channel
- single program that does **parallel** `send/receive`, `send/receive` on the **different** channels
- Send 3 tensors, but receive expects 4 (unless this is handled by generic machinery)

stablehlo/reference/ProcessGrid.cpp
ghpvnist: Good call, updated.

stablehlo/reference/Ops.cpp
ghpvnist: The spec does not go into detail about default values as this is implementation specific.

stablehlo/tests/interpret_send_recv.mlir
ghpvnist: Updated the test to include the first two tests. The third test is verified by the `run_parallel` op verifier.

stablehlo/reference/Ops.cpp
ghpvnist: This should be caught when trying to add incompatible types to the Scope object through an assertion (e.g. Scope expects `Value` with 4 but the Tensor value only has 3)l

stablehlo/dialect/StablehloOps.cpp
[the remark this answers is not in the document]
ghpvnist: I see what you mean, updated the PR description and the comment in the code to reflect the change.

PR #1737, opened 2023-08-15
Fix race condition in `ProcessGrid::rendezvous`

ghpvnist:
There is a race condition when two processes try to access `channelLocks_[channelKey]` at the same time when that key doesn't exist. The two processes will both create a mutex at the same time, acquiring different mutex instead of sharing the same one. It is also unsafe to read from `channels_[channelKey]` without a lock since this value can also be modified by another caller to `rendezvous` with a duplicate `processGroup` and `channelId` by another distribution op.

### DIFF
--- a/stablehlo/reference/ProcessGrid.h
+++ b/stablehlo/reference/ProcessGrid.h
@@ -115,6 +115,9 @@ class ProcessGrid {
   ProcessGroups flattenedIds(
       SmallVector<SmallVector<uint32_t>> flattenedIdGroups);
 
+  /// Synchronizes access to StableHLO `channels`.
+  std::mutex &getChannelLock(std::pair<ProcessGroup, ChannelId> channelKey);
+
   /// Inserts `inputs` to StableHLO `outfeed`.
   void outfeed(ArrayRef<Tensor> inputs);
 
@@ -158,6 +161,11 @@ class ProcessGrid {
   /// this key is gradually populated with tensors arriving from different
   /// processes in the process group.
   std::map<std::pair<ProcessGroup, ChannelId>, RendezvousResult> channels_;
+
+  /// Synchronization primitive used to manage concurrent access to
+  /// `channelLocks_` to prevent multiple processes from creating multiple
+  /// mutexes when the map entry is empty.
+  std::mutex channelLock_;
 
   /// Synchronization primitive used to manage concurrent access to `channels_`.
   std::map<std::pair<ProcessGroup, ChannelId>, std::mutex> channelLocks_;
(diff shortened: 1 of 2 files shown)


ghpvnist:
Thank you @burmako and @GleasonK for helping pinpoint the root cause!

stablehlo/reference/ProcessGrid.h
```diff
@@ -115,6 +115,9 @@ class ProcessGrid {
   ProcessGroups flattenedIds(
       SmallVector<SmallVector<uint32_t>> flattenedIdGroups);
 
+  /// Synchronizes access to StableHLO `channels`.
+  std::mutex &getChannelLock(std::pair<ProcessGroup, ChannelId> channelKey);
```
burmako, line 119: Doesn't look like this should be part of the public API.

burmako, line 119: Pass the key by reference? Or alternatively, split into two parameters?

ghpvnist:
`getChannelLock` function is only used in the implementation but is currently marked public (to access its private member variables). We currently don't have [pimpl idiom](https://en.cppreference.com/w/cpp/language/pimpl) in our codebase. Shall I create a ticket for now?

stablehlo/reference/ProcessGrid.h
```diff
@@ -159,6 +162,11 @@ class ProcessGrid {
   /// processes in the process group.
   std::map<std::pair<ProcessGroup, ChannelId>, RendezvousResult> channels_;
 
+  /// Synchronization primitive used to manage concurrent access to
+  /// `channelLocks_` to prevent multiple processes from creating multiple
+  /// mutexes when the map entry is empty.
+  std::mutex channelLock_;
```
burmako, line 168: Nit: the name is fairly ambiguous with `channelLocks_` defined below. How about `channelsMasterLock` or something?

burmako (changes requested):

stablehlo/reference/ProcessGrid.h
```diff
@@ -140,6 +140,11 @@ class ProcessGrid {
                               ProcessId processId, const Tensor &operand);
 
  private:
+  /// Synchronizes access to `channelLocks_` to prevent multiple processes from
+  /// creating multiple mutexes when the entry in `channelLocks_` is empty.
+  std::mutex &getChannelsMasterLock(ProcessGroup processGroup,
+                                    ChannelId channelId);
+
   /// StableHLO `num_replicas`.
   uint32_t numReplicas_;
```
GleasonK, line 149: While we're making changes - any thoughts on making `numReplicas_` and `numPartitions_` const? These are both read in ProcessGrid methods without any locking, so would be safer to enforce that these values are read-only.

ghpvnist: Agreed, updated.

ghpvnist:
I've kept the second enclosing scope to indicate everything in the scope needs to be handled with a lock. E.g. it may not be obvious that `return channels_[channelKey];` also needs to be thread-safe without the scope.

burmako (approved):

Issue #619, opened 2022-11-25
Add ShardingAttr


burmako:
Sharding on StableHLO/MHLO ops is represented as serialized proto string, e.g. `mhlo.sharding = "\08\03\1A\02\02\01\22\02\00\01"`. We should spec it and model it using a dedicate data structure.

atondwal:
I've been working on a prototype in MHLO, which I thought would be nice to have
to try this out before I send a PR to the spec, but it's trickier than
expected. I'd like to get some feedback on the design before then, since I've
already committed a lot of engineering time to the prototype, and want to get
some early feedback as I'm commit even more time to it. cc @mikedelorimier who was asking about this yesterday.

My proposal is that I implement this initially in MHLO so that op `sharding`  can take
either this sharding attr, or a string (as before). And then I'll add it in stablehlo in a form that only takes the attr.

Right now we don't have sections in the spec for attributes. Here's what
I've been working off of:

## Attributes

### sharding

#### Semantics

The `stablehlo.sharding` attribute is used to specify the sharding strategy for
an operation. The sharding strategy determines how the computation for the
operation is distributed across multiple devices. This sharding strategy is a
ShardingAttr attribute.

## Attribute Types

### ShardingAttr

-   `REPLICATED`: The operation is replicated across all devices, The
    entire operation is executed independently on each device. No other
    fields are used in this case, except `metadata`.
-   `MAXIMAL`: The operation is executed entirely on a single device,
    the first device in the `tile_assignment_devices` list.
-   `TUPLE`: The operation is sharded as a tuple, with different
    shardings applied to each leaf of the tuple. The `tuple_shardings`
    field contains a list of `ShardingAttr` instances, one per leaf node
    in the tuple shape. No other fields are used in this case, except
    `metadata`.
-   `OTHER`: The operation is sharded using the specified tile shape and
    device assignment. The `tile_shape` field defines the shape of the
    sharded tile, and `tile_assignment_dimensions` specifies the shape
    of the tile assignment tensor. The `tile_assignment_devices` field
    contains a flattened list of device IDs for the tile assignment.
    -   If `replicate_on_last_tile_dim` is true, the data is sharded
        according to other dimensions of the `tile_assignment`, but
        replicated across devices along the last dimension.
    -   The `last_tile_dims` field is a list of `ShardingType` values
        representing the sharding type of each subgroup in the
        `tile_assignment_dimensions`. If this is shorter than the rank
[...]

PR #1741, opened 2023-08-16
Specification of sharding attributes

sdasgup3:
This PR adds the specification for sharding attributes in StableHLO clarifying its syntax and semantics.

fixes https://github.com/openxla/stablehlo/issues/619


burmako:
Thank you for your meticulous work! I'll be out of office till the beginning of September, but I'm looking forward to diving into this review once I'm back - this is a very important topic, especially given the importance of LLMs, which has been pretty underspecced so far.

Issue #1574, opened 2023-06-04
Figure out the story for per-axis quantization


burmako:
What ops should support per-axis quantization? All of them? Only some of them? (E.g. dot_general). Need some clear criterion for this, rather than ad-hoc (per-op) rules. #1535 represents some ongoing work in that direction.

Also, how do we spec the semantics of per-axis quantization? Will we need to duplicate per-tensor and per-axis semantics for each supporting op? Or maybe there's some easy way out? In #1477, there's a sketch of the latter.

```
For simplicity, we only define individual operation semantics for per-tensor
quantized type. Semantics of per-axis type, if supported by the op, can be
derived as follows:

Given an operation specification with per-tensor quantized type, let the
semantics of the operation includes some application of the quantization
parameters `zero_point` and `scale` to a tensor element`e` at index = `[i0, ...,
id, ..., iR-1]`. The specification for the semantics of the same operation on
tensor element `e`, with per-axis quantized type with quantization dimension `d`
, would exactly be the same as the one with per-tensor type except that the
`zero_point` and `scale` will be replaced with their per-axis counterparts,
`zero_points[id]` and `scales[id]` resp.
```

burmako:
Here's an example of how a more formal definition of quantized semantics for AddOp could look like:

```
For quantized types, `result[result_index] = result_element` where:
  * `float_lhs = (lhs[result_index] - zero_point(lhs, result_index)) * scale(lhs, result_index)`.
  * `float_rhs = (rhs[result_index] - zero_point(rhs, result_index)) * scale(rhs, result_index)`.
  * `float_result = float_lhs + float_rhs`.
  * `rounded_result = round_nearest_even(float_result / scale(result))`.
  * `result_element = clamp(storage_min(result), rounded_result + zero_point(result), storage_max(result))`.
```

It's more verbose, but it captures the semantics exactly, similarly to how we wrote the rest of the specification. I imagine that we may also find ways to compress it further by introducing some functions in #1556, for example:

```
For quantized types, `result[result_index] = result_element` where:
  * `float_lhs = uniform_dequantize(lhs[result_index], type(lhs), result_index)`
  * `float_rhs = uniform_dequantize(rhs[result_index], type(rhs), result_index)`
  * `float_result = float_lhs + float_rhs`.
  * `result_element = uniform_quantize(float_result, type(result), result_index)`.
```

sdasgup3:
[...]

PR #1747, opened 2023-08-22
Per-axis quantization specification for reshape, transpose, and broadcast_in_dim.

sdasgup3:
Based on the discussion in https://github.com/openxla/stablehlo/issues/1574, I propose the quantization specification for the following data movement ops:
 -  **transpose**:  The `quantization_dimension` of `result` is also transposed based on the input `permutation` order.
 - **reshape**:  The `quantization_dimension` of the operand should not be merged or split.
 - **broadcast_in_dim**: For [degenerate dimension broadcasting](https://www.tensorflow.org/xla/broadcasting#broadcasting_similar-rank_arrays_with_degenerate_dimensions) of the operand's quantized dimension, the scales of the result are expanded.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -1387,8 +1387,9 @@ in the `operand` tensor and produces a `result` tensor. More formally,
 * (C6) `is_per_tensor_quantized(operand) = is_per_tensor_quantized(result)`.
 * (C7) If `is_per_axis_quantized(result)`:
   * `quantized_dimension(result) = broadcast_dimensions[quantized_dimension(operand)]`.
-  * `If dim(operand, quantized_dimension(operand)) = 1, then
-    scales(result)[i] = scales(operand)[0] for i in
+  * If `dim(operand, quantized_dimension(operand)) = 1`, then
+    `scales(result)[i] = scales(operand)[0] and zero_points(result)[i] =
+    zero_points(operand)[0] for i in
     range(dim(result, quantized_dimension(result)))`.
 
 #### Examples
@@ -4385,15 +4386,15 @@ ordering of `index_space(result)` and `index_space(operand)`.
 * (C4) If `is_per_axis_quantized(operand)`:
   * `reduce(dims(operand, [0, 1, ..., quantization_dimension(operand) - 1]),
     init_values=1, dimensions=[0], body=lambda x, y: x * y) =
-    reduce(dims(operand, [0, 1, ..., quantization_dimension(result) - 1]),
+    reduce(dims(result, [0, 1, ..., quantization_dimension(result) - 1]),
     init_values=1, dimensions=[0], body=lambda x, y: x * y)`.
   * `dim(operand, quantization_dimension(operand)) =
     dim(result, quantization_dimension(result))`.
   * `reduce(dims(operand,
     [quantization_dimension(operand) + 1, ..., rank(operand) - 1]),
     init_values=1, dimensions=[0], body=lambda x, y: x * y) =
-    reduce(dims(operand,
-    [quantization_dimension(operand) + 1, ..., rank(operand) - 1]),
+    reduce(dims(result,
+    [quantization_dimension(result) + 1, ..., rank(result) - 1]),
     init_values=1, dimensions=[0], body=lambda x, y: x * y)`.
 
 #### Examples
@@ -5508,8 +5509,8 @@ where `result_index[d] = operand_index[permutation[d]]`.
 * (C2) `permutation` is a permutation of `range(rank(operand))`.
 * (C3) `shape(result) = dim(operand, permutation...)`.
 * (C4 `is_per_tensor_quantized(operand) = is_per_tensor_quantized(result)`.
-* (C5) If `is_per_axis_quantized(result)`, then `quantized_dimension(result) =
-       permutation(quantized_dimension(operand))`.
+* (C5) If `is_per_axis_quantized(result)`, then `quantized_dimension(operand) =
+       permutation(quantized_dimension(result))`.
 
 #### Examples
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -198,11 +198,12 @@ have quantized element types, instead of regular element types.
 In quantized tensors, quantization can be **per-tensor**, meaning, having
 one `scale` and `zero_point` for the entire tensor or can be **per-axis**,
 meaning, having multiple `scales` and `zero_points`, one pair per slice of
-a particular dimension `quantized_dimension`. More formally, in a tensor `t` of
-with per-axis quantization, there are `dim(t, quantized_dimension)` slices
-of the `quantized_dimension`: `t[:, ..., 0, ..., :], t[:, ..., 1, ..., :]`, etc.
-All elements in the `i`th slice use `scales[i]` and `zero_points[i]` as their
-quantization parameters. Quantized tensor types have the following constraints:
+a particular dimension `quantization_dimension`. More formally, in a tensor `t`
+with per-axis quantization, there are `dim(t, quantization_dimension)` slices
+of the `quantization_dimension`: `t[:, ..., 0, ..., :], t[:, ..., 1, ..., :]`,
+etc. All elements in the `i`th slice use `scales[i]` and `zero_points[i]` as
+their quantization parameters. Quantized tensor types have the following
+constraints:
 
 * For per-tensor quantization:
   * No additional constraints.
@@ -1378,11 +1379,11 @@ in the `operand` tensor and produces a `result` tensor. More formally,
 #### Constraints
 
 * (C1) `element_type(result)` is given by:
-  * `element_type(operand)`, if `!is_quantized(operand)`.
+  * `element_type(operand)`, if `!is_per_axis_quantized(operand)`.
   * `element_type(operand)` except that `quantization_dimension(operand)`,
   `scales(operand)`, and `zero_points(operand)` may differ from
   `quantization_dimension(result)`, `scales(result)`, and `zero_points(result)`
-  resp., if `is_quantized(operand)`.
+  resp., otherwise.
 * (C2) `size(broadcast_dimensions) = rank(operand)`.
 * (C3) `0 <= broadcast_dimensions < rank(result)`.
 * (C4) `is_unique(broadcast_dimensions)`.
@@ -1390,11 +1391,11 @@ in the `operand` tensor and produces a `result` tensor. More formally,
   * `dim(operand, d) = 1` or
   * `dim(operand, d) = dim(result, broadcast_dimensions[d])`.
 * (C6) If `is_per_axis_quantized(result)`:
-  * `quantized_dimension(result) = broadcast_dimensions[quantized_dimension(operand)]`.
-  * If `dim(operand, quantized_dimension(operand)) = 1`, then
+  * `quantization_dimension(result) = broadcast_dimensions[quantization_dimension(operand)]`.
+  * If `dim(operand, quantization_dimension(operand)) = 1`, then
     `scales(result)[i] = scales(operand)[0] and zero_points(result)[i] =
     zero_points(operand)[0] for i in
-    range(dim(result, quantized_dimension(result)))`.
+    range(dim(result, quantization_dimension(result)))`.
 
 #### Examples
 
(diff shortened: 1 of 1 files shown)


docs/spec.md
```diff
@@ -1384,6 +1384,12 @@ in the `operand` tensor and produces a `result` tensor. More formally,
 * (C5) For all `d` in `axes(operand)`:
   * `dim(operand, d) = 1` or
   * `dim(operand, d) = dim(result, broadcast_dimensions[d])`.
+* (C6) `is_per_tensor_quantized(operand) = is_per_tensor_quantized(result)`.
+* (C7) If `is_per_axis_quantized(result)`:
+  * `quantized_dimension(result) = broadcast_dimensions[quantized_dimension(operand)]`.
+  * `If dim(operand, quantized_dimension(operand)) = 1, then
+    scales(result)[i] = scales(operand)[0] for i in
```
loganchien, line 1391: `and zero_points(result)[i]] = zero_points(operand)[i]`

docs/spec.md
```diff
@@ -4359,20 +4365,34 @@ ordering of `index_space(result)` and `index_space(operand)`.
...
+    init_values=1, dimensions=[0], body=lambda x, y: x * y) =
+    reduce(dims(operand, [0, 1, ..., quantization_dimension(result) - 1]),
+    init_values=1, dimensions=[0], body=lambda x, y: x * y)`.
+  * `dim(operand, quantization_dimension(operand)) =
+    dim(result, quantization_dimension(result))`.
+  * `reduce(dims(operand,
+    [quantization_dimension(operand) + 1, ..., rank(operand) - 1]),
+    init_values=1, dimensions=[0], body=lambda x, y: x * y) =
+    reduce(dims(operand,
+    [quantization_dimension(operand) + 1, ..., rank(operand) - 1]),
```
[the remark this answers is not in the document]

docs/spec.md
```diff
@@ -1384,6 +1384,12 @@ in the `operand` tensor and produces a `result` tensor. More formally,
 * (C5) For all `d` in `axes(operand)`:
   * `dim(operand, d) = 1` or
   * `dim(operand, d) = dim(result, broadcast_dimensions[d])`.
+* (C6) `is_per_tensor_quantized(operand) = is_per_tensor_quantized(result)`.
+* (C7) If `is_per_axis_quantized(result)`:
+  * `quantized_dimension(result) = broadcast_dimensions[quantized_dimension(operand)]`.
+  * `If dim(operand, quantized_dimension(operand)) = 1, then
```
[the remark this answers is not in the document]

docs/spec.md
```diff
@@ -5469,22 +5490,25 @@ where `result_index[d] = operand_index[permutation[d]]`.
...
+| `result` | tensor or quantized tensor | (C1), (C3)  |
 
 #### Constraints
 
 * (C1) `element_type(operand) = element_type(result)`.
 * (C2) `permutation` is a permutation of `range(rank(operand))`.
 * (C3) `shape(result) = dim(operand, permutation...)`.
+* (C4 `is_per_tensor_quantized(operand) = is_per_tensor_quantized(result)`.
+* (C5) If `is_per_axis_quantized(result)`, then `quantized_dimension(result) =
+       permutation(quantized_dimension(operand))`.
```
loganchien, line 5511: Seems to be the reverse:

`quantized_dimension(operand) = permutation(quantized_dimension(result))`

because according to the code above:

```
result_index[d] = operand_index[permutation[d]]
```

`permutation` is actually a result-axis to operand-axis function (or the array index is the result-axis and the array value is the operand-axis).  so we should put `quantized_dimension(result)` in `permutation[...]`.

sdasgup3: That is good catch! Thanks. 
Updated the spec accordingly.

loganchien (approved):

docs/spec.md
```diff
@@ -1384,6 +1384,13 @@ in the `operand` tensor and produces a `result` tensor. More formally,
 * (C5) For all `d` in `axes(operand)`:
   * `dim(operand, d) = 1` or
   * `dim(operand, d) = dim(result, broadcast_dimensions[d])`.
+* (C6) `is_per_tensor_quantized(operand) = is_per_tensor_quantized(result)`.
+* (C7) If `is_per_axis_quantized(result)`:
+  * `quantized_dimension(result) = broadcast_dimensions[quantized_dimension(operand)]`.
+  * If `dim(operand, quantized_dimension(operand)) = 1`, then
+    `scales(result)[i] = scales(operand)[0] and zero_points(result)[i] =
+    zero_points(operand)[0] for i in
+    range(dim(result, quantized_dimension(result)))`.
```
dansuh17, line 1393: totally optional - can we have an example in the spec that demonstrates the case for per-axis quantized tensor? The constraints might not be very intuitive for first-time users.

sdasgup3: Hi @dansuh17 thanks for the suggestion. 
We do have examples for per-axis quantization for `uniform_quantize` and `uniform_dequantize` ([here](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#uniform_quantize) and [here](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#uniform_dequantize)). For most of the other ops which supports quantized types in addition to other regular element types, we follow the norm to provide example for only one regular element type (ref point 6 of [this](https://github.com/openxla/stablehlo/blob/main/docs/spec_checklist.md#stablehlo-specification-checklist)). While we add constraint checks to the verifier for each op, we will be adding example for each  constraint check. Does that sound like a plan to you?

docs/spec.md
```diff
@@ -1384,6 +1384,13 @@ in the `operand` tensor and produces a `result` tensor. More formally,
 * (C5) For all `d` in `axes(operand)`:
   * `dim(operand, d) = 1` or
   * `dim(operand, d) = dim(result, broadcast_dimensions[d])`.
+* (C6) `is_per_tensor_quantized(operand) = is_per_tensor_quantized(result)`.
```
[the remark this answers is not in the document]
sdasgup3: Thanks for the catch! On a second thought let me remove the constraint altogether in favor of (C1) `element_type(operand) = element_type(result)` which could capture both the cases. PTAL.

doyeonkim0 (approved):

docs/spec.md
dansuh17: SGTM! Will the verifier implementation follow this commit?

dansuh17 (approved):

docs/spec.md
```diff
@@ -1364,16 +1364,16 @@ in the `operand` tensor and produces a `result` tensor. More formally,
...
 #### Outputs
 
-| Name     | Type                                  | Constraints      |
-|----------|---------------------------------------|------------------|
-| `result` | tensor or per-tensor quantized tensor | (C1), (C3), (C5) |
+| Name     | Type                       | Constraints         |
+|----------|----------------------------|---------------------|
+| `result` | tensor or quantized tensor | (C1), (C3), (C5-C6) |
 
 #### Constraints
```
dansuh17, line 1378: Actually, following-up on Doyeon's comment at https://github.com/openxla/stablehlo/pull/1747#discussion_r1302338221, I think (C1) should be broken down into two separate conditions, one for quantized type and non-quantized type.

The reason why we need a separate issue for the quantized type is that the `element_type` is considered different when at least one of the parameter is different. If two quantized types have different quantization axes then they are considered different, e.g. `!quant.uniform<i8:f32:0, {0.1:-30,0.5:-20}> != !quant.uniform<i8:f32:1, {0.1:-30,0.5:-20}>`. If an op modifies the quantization params then C1 is violated.

I think this should be applied to all ops that **may** modify the quantization parameters.

dansuh17 (changes requested):

docs/spec.md
sdasgup3: Yes, that is right.

paulinesho (approved):

docs/spec.md
dansuh17: Thanks! LGTM.

dansuh17 (approved):

docs/spec.md
```diff
@@ -5471,20 +5498,25 @@ where `result_index[d] = operand_index[permutation[d]]`.
 
 | Label | Name          | Type                                         | Constraints |
 |-------|---------------|----------------------------------------------|-------------|
-| (I1)  | `operand`     | tensor or per-tensor quantized tensor        | (C1-C3)     |
-| (I2)  | `permutation` | 1-dimensional tensor constant of type `si64` | (C2), (C3)  |
+| (I1)  | `operand`     | tensor or quantized tensor                   | (C1-C4)     |
+| (I2)  | `permutation` | 1-dimensional tensor constant of type `si74` | (C2-C4)     |
```
dominicsymes, line 5502: "si74" is a typo for si64?

dominicsymes (approved):
Think there is one typo. Otherwise looks good to me.

docs/spec.md
sdasgup3: Thanks for catching this! fixed.

docs/spec.md
```diff
@@ -1364,26 +1364,37 @@ in the `operand` tensor and produces a `result` tensor. More formally,
...
+| `result` | tensor or quantized tensor | (C1), (C3), (C5-C6) |
 
 #### Constraints
 
-* (C1) `element_type(operand) = element_type(result)`.
+* (C1) `element_type(result)` is given by:
+  * `element_type(operand)`, if `!is_quantized(operand)`.
+  * `element_type(operand)` except that `quantization_dimension(operand)`,
+  `scales(operand)`, and `zero_points(operand)` may differ from
+  `quantization_dimension(result)`, `scales(result)`, and `zero_points(result)`
```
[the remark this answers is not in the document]
sdasgup3: Yes, you are right. This should be fixed. 

Based on the Grammar rule
```
QuantizationDimension ::= IntegerConstant
```

and the fact from https://github.com/openxla/stablehlo/blob/main/docs/spec.md#names

```
For everything else, member definitions include structural parts of the program element, named after the corresponding EBNF non-terminals.
```

We can settle on `quantization_dimension`. Fixed all the occurances.

docs/spec.md
```diff
@@ -1364,26 +1364,37 @@ in the `operand` tensor and produces a `result` tensor. More formally,
...
-| `result` | tensor or per-tensor quantized tensor | (C1), (C3), (C5) |
+| Name     | Type                       | Constraints         |
+|----------|----------------------------|---------------------|
+| `result` | tensor or quantized tensor | (C1), (C3), (C5-C6) |
 
 #### Constraints
 
-* (C1) `element_type(operand) = element_type(result)`.
+* (C1) `element_type(result)` is given by:
+  * `element_type(operand)`, if `!is_quantized(operand)`.
```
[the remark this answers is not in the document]
sdasgup3: Thanks for the catch. That is a typo to be fixed. 

This should be `!is_per_axis_quantized`. fixed.

docs/spec.md
```diff
@@ -4361,20 +4373,36 @@ ordering of `index_space(result)` and `index_space(operand)`.
...
 #### Constraints
 
-* (C1) `element_type(operand) = element_type(result)`.
+* (C1) `element_type(result)` is given by:
+  * `element_type(operand)`, if `!is_per_axis_quantized(operand)`.
+  * `element_type(operand)` except that `quantization_dimension(operand)` and
+    `quantization_dimension(result)` may differ, otherwise.
 * (C2) `size(operand) = size(result)`.
+* (C3) If `is_per_axis_quantized(operand)`:
+  * `reduce(dims(operand, [0, 1, ..., quantization_dimension(operand) - 1]),
```
GleasonK, line 4394: Do we use this `reduce` notation, or something similar, elsewhere in the spec? I like that this will match the implementation fairly closely, but was a bit more difficult to understand at first glance.

I think this concern will be resolved once we have a verifier and an error message. I anticipate this error message will be a little difficult to write.. "expected dimensions of rank < quantizedDim of input and output to have the same number of elements" ?

GleasonK (approved):
Spoke with @sdasgup3 offline. LGTM

docs/spec.md
sdasgup3: Yes this notation is used in the following ops. 
[batch_norm_grad](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#batch_norm_grad)
[batch_norm_training](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#batch_norm_training)
[dot_general](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#dot_general)
[reduce_window](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#reduce_window)
[select_and_scatter](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#select_and_scatter)

Or maybe something like `quantization_dimension(operand) cannot be splitted or merged during the reshape operation`

PR #1750, opened 2023-08-24
[NFC] Explicitly set dialects to usePropertiesForAttributes=0

GleasonK:
Port of https://github.com/tensorflow/mlir-hlo/commit/e8a6a7c2b51f7f7e27d050afbc115e805b8a9fc6


GleasonK:
In a followup we can probably change InterpreterOps.td to use properties, I don't think that would break anything. CHLO, StableHLO, and VHLO will likely require some integ work.

ghpvnist (approved):

Issue #1755, opened 2023-08-25
Fix TSAN violations for Distribution ops


ghpvnist:
Currently tests for Distribution ops have TSAN violations, and should enable these tests agian once the issue is fixed.

PR #1756, opened 2023-08-25
Disable multithreaded tests

ghpvnist:
There are TSAN violations for multithreaded tests. I've added #1755 to address it, and we should disable it in the meantime.

### DIFF
--- a/stablehlo/tests/interpret_all_gather.mlir
+++ b/stablehlo/tests/interpret_all_gather.mlir
@@ -1,4 +1,5 @@
-// RUN: stablehlo-translate --interpret -split-input-file %s
+// RUN: echo 'DISABLED'
+// RUN-DISABLED: stablehlo-translate --interpret -split-input-file %s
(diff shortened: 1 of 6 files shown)


stablehlo/tests/interpret_all_gather.mlir
```diff
@@ -1,74 +1,74 @@
 // RUN: stablehlo-translate --interpret -split-input-file %s
```
GleasonK, line 1: Can we confine this to just this line instead of commenting all lines?

If there are multiple run lines we can usually do `// RUN-DISABLED:`, but since there is only one, I believe we'll get an error for there not being any RUN lines.

We could try:

```
// RUN: echo 'DISABLED' # stablehlo-translate --interpret -split-input-file %s
```

GleasonK (changes requested):

GleasonK (approved):

PR #1761, opened 2023-08-30
tests: drop the use of "grep" from lit tests

ashay:
The "grep" utility doesn't exist on Windows (except on MSYS2 systems),
making it difficult for Windows users to run a handful of lit tests in
the stablehlo/tests directory.  As a workaround, this patch replaces the
use of grep in these tests with FileCheck.


GleasonK (approved):
Deleted a bytecode impl to test that the error case still works - worked great!

PR #1764, opened 2023-08-30
tests: avoid sub-shells in lit commands for correct use on Windows

ashay:
On Windows, command substitution fails when the `RUN` command uses
sub-shells, which are often used when the lit test needs to use the
output of commands, for example, `diff <(cmd1 arg1 arg2) <(cmd2 arg1)`.
To avoid such incorrect substitution, it seems the canonical way to use
the output of one command as an input for the subsequent command is to
save the output of each sub-command into a temporary file before running
the original (`diff`) command on the temporary files (see clang lit
tests for examples).

This patch applies this pattern to all tests in the "tests" and
"testdata" directories.  Although this patch is massive, it was generated
using the following sed command:

```bash
sed 's/\/\/ RUN: diff <(\(.*\)) <(\(.*\))/\/\/ RUN: \1 > %t.0\n\/\/ RUN: \2 > %t.1\n\/\/ RUN: diff %t.0 %t.1/' -i stablehlo/tests/*.mlir

sed 's/\/\/ RUN: diff <(\(.*\)) <(\(.*\))/\/\/ RUN: \1 > %t.0\n\/\/ RUN: \2 > %t.1\n\/\/ RUN: diff %t.0 %t.1/' -i stablehlo/testdata/*.mlir
```

As a result of this and previous two patches (#1760 and #1761), the
StableHlo code builds and runs all lit tests successfully on Windows.

### DIFF
--- a/stablehlo/testdata/abs_shape_bfloat16_20_20.mlir
+++ b/stablehlo/testdata/abs_shape_bfloat16_20_20.mlir
@@ -1,5 +1,7 @@
 // RUN: stablehlo-opt -inline %s | stablehlo-translate --interpret
-// RUN: diff <(stablehlo-translate --serialize --target=current %s | stablehlo-translate --deserialize | stablehlo-opt) <(stablehlo-opt %s)
+// RUN: stablehlo-translate --serialize --target=current %s | stablehlo-translate --deserialize | stablehlo-opt > %t.0
+// RUN: stablehlo-opt %s > %t.1
+// RUN: diff %t.0 %t.1
 
 module @jit_testcase {
   func.func public @main() -> tensor<i1> {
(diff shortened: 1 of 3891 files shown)


ashay:
I acknowledge that this patch is huge, but it is the result of executing a couple of `sed` commands.  If there's a better way to accommodate running these tests on Windows, please let me know.  Thanks much!

GleasonK:
Thanks for the contributions!

No major opposition to using temporary files. I think we've had issues in the past with running lit tests like this in parallel since `RUN` command order isn't guaranteed. Let me see if I can repro and suggest a modification better than a long line with a bunch of `&&`s. Don't remember if adding `\` to the end of related lines works.

Will try to review the previous two patches today, and get to testing this tomorrow.

GleasonK (approved):
Ah I take it back, looks like writing to files in separate run commands is a supported workflow: https://llvm.org/docs/TestingGuide.html#writing-new-regression-tests

The bug I was referring to earlier: A first RUN command set environment variables, and the second RUN attempted to use them, but the lit test runner used executed each cmd in its own shell, unless a trailing `\` was used, in which case they were concatenated. We don't need that here.

This LGTM. There is a merge conflict on a few files now that I've merged your other submissions, mind rebasing?

ashay:
Thanks for explaining!  I've rebased to the main branch.

PR #1768, opened 2023-09-02
Fix TSAN violations for Distribution ops

### LINKED ISSUE #1755: Fix TSAN violations for Distribution ops

ghpvnist:
The TSAN violation is due to an unprotected insert shown above and replacing `llvm::RefCountedBase` with its thread-safe alternative `llvm::ThreadSafeRefCountedBase`.

In the original implementation of `rendezvous`, there were some nuances that were undesirable.
For example, it's preferred to clear out the entry once all processes have read the data and clear out the map instead of letting the next distribution op to clear it out. The current implementation is also incorrect since there is no guarantee that processGroup size is the same as previous op calling `rendezvous`.
```
if (channels_[channelKey].size() == processGroup.size())
  channels_[channelKey].clear();

channels_[channelKey].insert(processId, operand);
```

The proposed update enhances the `rendezvous` logic such that each process has a shared pointer to the result object once out of scope, it is automatically deleted.

The PR also reenables tests for distribution ops.

closes #1755

### DIFF
--- a/stablehlo/reference/ProcessGrid.cpp
+++ b/stablehlo/reference/ProcessGrid.cpp
@@ -67,11 +67,8 @@ void RendezvousResult::insert(ProcessId processId, Tensor tensor) {
 }
 
 Tensor RendezvousResult::lookup(ProcessId processId) {
-  auto it = result_.begin();
-  while (it != result_.end()) {
-    if (it->first == processId) return it->second;
-    it++;
-  }
+  auto it = result_.find(processId);
+  if (it != result_.end()) return it->second;
   return {};
 }
 
@@ -198,7 +195,7 @@ std::shared_ptr<RendezvousResult> ProcessGrid::rendezvous(
 
     // The shared result from the state owns one, the last process to contribute
     // owns one, and the remaining processes (except the last) owns one here.
-    if (state.result.use_count() < (int64_t)processGroup.size()) {
+    if (state.result.use_count() < static_cast<int64_t>(processGroup.size())) {
       result = state.result;
       channelConditions_[channelKey].notify_one();
     } else {
--- a/stablehlo/reference/ProcessGrid.cpp
+++ b/stablehlo/reference/ProcessGrid.cpp
@@ -162,12 +162,11 @@ ProcessGroups ProcessGrid::flattenedIds(
 
 void ProcessGrid::outfeed(ArrayRef<Tensor> inputs) { outfeed_.push(inputs); }
 
-std::shared_ptr<RendezvousResult> ProcessGrid::rendezvous(
+const std::shared_ptr<RendezvousResult> ProcessGrid::rendezvous(
     ProcessGroup processGroup, ChannelId channelId, ProcessId processId,
     const Tensor &operand) {
   std::pair<ProcessGroup, ChannelId> channelKey(processGroup, channelId);
-  // Immediately return the result. The logic below doesn't work for a single
-  // process.
+  // Process wait/notify logic below doesn't work for single process.
   if (processGroup.size() == 1)
     return std::make_shared<RendezvousResult>(
         RendezvousResult({std::pair{processId, operand}}));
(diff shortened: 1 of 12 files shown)


stablehlo/reference/Tensor.h
```diff
@@ -36,7 +36,7 @@ namespace stablehlo {
 namespace detail {
 
 /// Underlying storage class for Tensor objects.
-class Buffer : public llvm::RefCountedBase<Buffer> {
+class Buffer : public llvm::ThreadSafeRefCountedBase<Buffer> {
```
GleasonK, line 39: Why does the buffer need to be thread safe? Are two threads ever looking at the same buffer?

Wondering if this was a TSAN issue that we saw only because the map wasn't thread safe, or if it is a separate issue that needs to be fixed

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -59,15 +59,19 @@ std::optional<ProcessGroup> ProcessGroups::findGroup(ProcessId processId) {
...
 
 Tensor RendezvousResult::lookup(ProcessId processId) {
-  auto it = result_.find(processId);
-  if (it != result_.end()) return it->second;
+  auto it = result_.begin();
+  while (it != result_.end()) {
+    if (it->first == processId) return it->second;
+    it++;
+  }
   return {};
```
GleasonK, line 75: Why is this change needed? Looks like the datatype of `result_` is unchanged?

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -142,45 +163,61 @@ ProcessGroups ProcessGrid::flattenedIds(
...
+    // the values into the shared result.
+    if (!channelConditions_[channelKey].wait_for(
+            lock, std::chrono::seconds(3),
+            [&] { return state.result != nullptr; }))
+      llvm::report_fatal_error(
+          "rendezvous timed out: not all processes have contributed yet");
+
+    // The shared result from the state owns one, the last process to contribute
+    // owns one, and the remaining processes (except the last) owns one here.
+    if (state.result.use_count() < (int64_t)processGroup.size()) {
```
GleasonK, line 201: `static_cast<int64_t>`

GleasonK (changes requested):
I'll need to spend more time thinking about this, brain not working this late on a Friday. Shared some initial thoughts.

stablehlo/reference/Tensor.h
ghpvnist: This particular [line](https://github.com/openxla/stablehlo/pull/1768/files#diff-b68de14f89a7656326ce089b57147f2dd5a24d7a1dc3dc427076210306f419e4R181) does a copy of the tensor, and it's not protected by the underlying class using `llvm::RefCountedBase`.

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -142,45 +160,61 @@ ProcessGroups ProcessGrid::flattenedIds(
...
-void ProcessGrid::outfeed(ArrayRef<Tensor> inputs) {
-  std::lock_guard<std::mutex> lock(outfeedLock_);
-  outfeed_.emplace(inputs);
-}
-
-RendezvousResult ProcessGrid::rendezvous(ProcessGroup processGroup,
-                                         ChannelId channelId,
-                                         ProcessId processId,
-                                         const Tensor &operand) {
+std::shared_ptr<RendezvousResult> ProcessGrid::rendezvous(
```
GleasonK, line 165: Here was my attempt to rewrite this in a more understandable way. There are two assumptions:
1. wait/notify respects the order of threads arriving, using a queue - I think this is the case.
2. state.values is not written to again until all processes have returned from this thread (can be tweaked to account for this).

```c++
std::shared_ptr<RendezvousResult> ProcessGrid::rendezvous(
    ProcessGroup processGroup, ChannelId channelId, ProcessId processId,
    const Tensor &operand) {
  std::pair<ProcessGroup, ChannelId> channelKey(processGroup, channelId);
  // Process wait/notify logic below doesn't work for single process.
  if (processGroup.size() == 1)
    return std::make_shared<RendezvousResult>(
        RendezvousResult({std::pair{processId, operand}}));

  auto &state = channels_[channelKey];

  std::unique_lock<std::mutex> lock(state.mutex);
  state.values[processId] = operand;

  if (state.values.size() == processGroup.size()) {
    // If values are full, that means all other processes are currently waiting.
    // The last process to contribute moves the values into the result
    // then waits for each process to return a copy of the result before
    // cleaning up the state variable for future computations in this process
    // grid.
    state.result = std::make_shared<RendezvousResult>(state.values);
    channelConditions_[channelKey].notify_one();
    if (channelConditions_[channelKey].wait_for(
            lock, std::chrono::seconds(3)) == std::cv_status::timeout)
      llvm::report_fatal_error(
          "rendezvous timed out: not all process has received the results yet");

    // Only woken up once all other processes in group have returned.
    // Clean up the shared state.
    state.values.clear();
    return std::move(state.result);
  }

  // The remaining processes wait for the last process to contribute to move
  // the values into the shared result.
  if (channelConditions_[channelKey].wait_for(lock, std::chrono::seconds(3)) ==
      std::cv_status::timeout)
    llvm::report_fatal_error(
        "rendezvous timed out: not all processes have contributed yet");

  if (state.result == nullptr)
    llvm::report_fatal_error("rendezvous expected result to be set");

  // Copy result from state before notifying.
  auto result = state.result;
  channelConditions_[channelKey].notify_one();
  return result;
}
```

GleasonK (changes requested):

stablehlo/reference/ProcessGrid.cpp
ghpvnist: Updated the commit with some minor changes. PTAL

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -142,45 +160,72 @@ ProcessGroups ProcessGrid::flattenedIds(
...
+    llvm::report_fatal_error(
+        "rendezvous timed out: not all process has received the results yet");
+
+  // Copy result from the state before notifying.
+  auto result = state.result;
+  channelConditions_[channelKey].notify_one();
+
+  // Wait for the remaining processes to have retrieved the result. In other
+  // words, wait until the last process to contribute exit the function.
+  if (!channelConditions_[channelKey].wait_for(
```
GleasonK, line 222: Is this wait needed? Once they have all returned their temporary value, do they need to wait for the "leader" to reset state to null before continuing? Is this to protect more than one thing using state at the same time?

GleasonK (changes requested):

stablehlo/reference/ProcessGrid.cpp
ghpvnist: The user code immediately releases the shared ptr which results in a decrease in `use_count`. This will result in a deadlock unless the last process is returned first. And the use of `use_count` as the condition is the workaround I found to ensure all processes have access to the value and address assumption 1) in your proposed change.
```
auto groupOperands = process->rendezvous(*processGroup, channelId, operand)
                           ->getSortedTensors();
```

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -142,45 +160,72 @@ ProcessGroups ProcessGrid::flattenedIds(
...
-              return channels_[channelKey].size() == processGroup.size();
+              return state.result.use_count() >=
+                     static_cast<int64_t>(processGroup.size());
             }))
-      llvm::report_fatal_error("rendezvous timed out");
+      llvm::report_fatal_error(
+          "rendezvous timed out: not all processes have contributed yet");
 
-    return channels_[channelKey];
+    if (state.result.use_count() > static_cast<int64_t>(processGroup.size()))
```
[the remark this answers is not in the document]
ghpvnist: No it should not be possible. This check acts more like an assertion that one process indeed has only one access. It can be removed, but I've added as a guard against potential future changes. Lmk if you think this is unnecessary, I can remove it.

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -142,45 +160,72 @@ ProcessGroups ProcessGrid::flattenedIds(
...
+      llvm::report_fatal_error(
+          "Each process should have only one shared access to the result.");
+
+    // The last process to contribute takes the result from the state to allow
+    // the process that contributed last to exit the function.
+    channelConditions_[channelKey].notify_one();
+    return std::move(state.result);
```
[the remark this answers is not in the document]
ghpvnist: IIUC, this is safe and `state.result` should be marked `nullptr`.

stablehlo/reference/ProcessGrid.cpp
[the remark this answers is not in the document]
GleasonK: Should be ok, was just curious

stablehlo/reference/ProcessGrid.cpp
GleasonK: makes sense, lgtm

GleasonK (approved):
Looks good, could you confirm that TSAN is clean at this final revision before merge?

Issue #1291, opened 2023-03-07
Define next steps for using stablehlo/testdata in interpreter tests


burmako:
To summarize the offline discussion:
  1) What tests can we enable already? (Partially answered by #1290).
  2) When will we be able to enable the rest of the tests?
  3) How should we update [the checklist](https://github.com/openxla/stablehlo/blob/main/docs/reference_checklist.md) now that we have all these tests in stablehlo/testdata?

ghpvnist:
1. There are some floating-point precision issues we need to sort out before enabling the rest. However, we can enable all tests for Data Movement ops, for which if we run into issues, will look into them:
* broadcast_in_dim
* concatenate
* dynamic_slice
* dynamic_update_slice
* gather #1058
* pad
* reshape
* slice
* sort #1283 
* transpose

2. We are hoping to enable all tests soon. However, some issues -- 1) some tests include attributes we plan to remove, 2) currently unsupported ops, 3) tests mixed with supported and unsupported ops, 4) floating-precision issues, 5) currently supported attributes like `precision_config` or having preferred element type for DotGeneral and Convolution which are not yet implemented -- have to be solved before all tests are enabled. The exact ETA to enable all tests is TBD, but the next steps for doing so would be to update the testing infrastructure to better capture difference in precision, which is expected to finish by the end of Q1.

3. I plan to update the checklist to enable any tests that are passing, including floating-point tests as this issue has to do with current testing infrastructure and not with, say wrong implementation (which we may find by enabling these tests progressively).

sdasgup3:
Some stats as of 03/08:
 - Total tests: 3329 
 - Enabled: 308 (All integer tests which are all passing)
 - Disabled: 3021
     - **Tests we could enable by 03/31: 1203 (ideally)** 
         - Out of 1203, 343 are int tests and the rest (860) are float/complex tests. 
         - Out of 1203, 332 are fully covered on 03/13, of which all except 15 are passing and hence they could be enabled right away.
             - The 15 tests failing because of precision mismatch (11) or other problems to be explored (4)

     - Not planned to be enable for Q1'23: 1818
         -  Out of 1818, 1116 are int tests and the rest (702) are float/complex tests.
[...]

PR #1775, opened 2023-09-16
Remove sign-based floating-point comparison

ghpvnist:
If there are two finite values such that the absolute difference is less than zero but their sign is different, I think it should compare approximately true. I.e. `areApproximatelyEqual(0.0, -0.00001)` should evaluate to true.


ghpvnist:
Let's combine this with #1291 to address them altogether.

Issue #1765, opened 2023-08-31
Allow Transpose Reshape and BroadcastInDim ops to have different quantized element types for operand and result. 


sdasgup3:
As a follow up on https://github.com/openxla/stablehlo/pull/1747, the verifiers of the above ops should be updated to allow different quantized element types for operand and result.  In the current [form](https://github.com/openxla/stablehlo/blob/b2090e9774f68c81b65615d18872b99b718eedad/stablehlo/dialect/StablehloOps.td#L2763) , these ops require the element types to be the same. 

cc @doyeonkim0

PR #1782, opened 2023-09-25
Fixing verifier for reshape, transpose, and broadcast_in_dim ops

sdasgup3:
fixes https://github.com/openxla/stablehlo/issues/1765

The existing verifier for these ops  was using the trait [SameOperandsAndResultElementType](https://github.com/openxla/stablehlo/blob/b2090e9774f68c81b65615d18872b99b718eedad/stablehlo/dialect/StablehloOps.td#L2763) to check the equality of the element types which will not work for the per-axis quantized variants of these ops as the quantization dimension and quantization parameters (scales and zero points)  may differ between operand and result.

The PR proposed to replace the trait with [HLO_CompatibleOperandsAndResultElementType ](https://source.corp.google.com/piper///depot/google3/third_party/stablehlo/stablehlo/dialect/Base.td;rcl=567468454;l=216)(which just checks the equality for storage type and expressed type).

Note that the PR does not provide a comprehensive solution here. It just allows the representational capability to these ops thereby allowing writing meaningful quantized versions of the ops w/o verification error. A comprehensive verification solution should include verifying all the constraints related to quantization for these ops and is true for other quantized ops as well. That would need additional work and will be addressed collectively for the full set of quantized ops.

### DIFF
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -1860,7 +1860,7 @@ def StableHLO_BroadcastOp : StableHLO_ShapedInterfaceOp<"broadcast",
 }
 
 def StableHLO_BroadcastInDimOp : StableHLO_Op<"broadcast_in_dim",
-      [Pure, SameOperandsAndResultElementType /*broadcast_in_dim_c1*/]> {
+      [Pure, HLO_CompatibleOperandsAndResultElementType /*broadcast_in_dim_c1*/]> {
   let summary = "BroadcastInDim operation";
   let description = [{
     Expands the dimensions and/or rank of an input tensor by duplicating the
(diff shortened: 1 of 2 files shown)


GleasonK (approved):
Can you update PR description to state that "compatible element type" means identical element type or quantized type with matching expressed/storage/min/max/ types. This is _only_ relaxing the constraint for quantized type's scale/zp values I believe.

Can we also make compatible only apply to PerAxis quantized types? I believe that is a separate subclass?

sdasgup3:
> Can you update PR description to state that "compatible element type" means identical element type or quantized type with matching expressed/storage/min/max/ types. This is _only_ relaxing the constraint for quantized type's scale/zp values I believe.

updated. 

> Can we also make compatible only apply to PerAxis quantized types? I believe that is a separate subclass?

The proposed compatibility is applied to the case when the above operations are per-axis quantized. Note that the same strategy is currently shared by the verification logic of most per-tensor quantized StableHLO operations. The PR include the above operations in the same category.  If we make compatible apply only to per-axis scheme, then it will affect how other per-tensor quantized operation (e.g. [add op](https://github.com/openxla/stablehlo/blob/680d44d1be19a74d3479c2906b3394389ccb1dec/stablehlo/dialect/StablehloOps.td#L691) which eventually uses the [same](https://github.com/openxla/stablehlo/blob/680d44d1be19a74d3479c2906b3394389ccb1dec/stablehlo/dialect/Base.cpp#L70) compatibility definition we are proposing in this PR)

PR #1703, opened 2023-07-27
Add Interpreter dialect

ghpvnist:
The StableHLO reference implementation is nearing its final phase to implement most of the specced operations in the StableHLO opset, and one major remaining piece is the Distribution ops. This PR creates the `Interpreter` dialect specifically targeted for interpreter-related ops that enable running parallel execution of Distribution ops. In the future, this dialect may also be used to add additional functionality to the interpreter.

Issue #1780, opened 2023-09-22
Provide standardized interface for calling reference interpreter


penagos:
### Request description

Today, it isn't practical to call the StableHLO reference interpreter. Invoking the interpreter requires going through the MLIR translate infrastructure (i.e. via `interpretRegistration` in `StablehloTranslateMain.cpp`), and does not currently support specifying user inputs into a model or doing any meaningful post processing on the model outputs short of printing them to screen. This makes it difficult to run StableHLO MLIR models.

It would be great to provide a standardized API to call into the StableHLO reference interpreter, in which clients could specify the following:

1. The StableHLO MLIR (or `ModuleOp`) to be executed
2. The model inputs, as `InterpreterValue` instances
3. An optional interpretation fallback callback (which defaults to the existing `interpreterFallback` in `StablehloTranslateMain.cpp`)

Akin to `eval`, the interface could return a `SmallVector<InterpreterValue>` of result tensors. While I think the `eval` method might be able to be repurposed for this feature, I think lower level abstractions (such as `Region`s, `Scope`s and `Process`es should not be something which are required & exposed to downstream users).

GleasonK:
This has been completed in https://github.com/openxla/stablehlo/pull/1784 and a few followup PRs

PR #1784, opened 2023-09-28
Add reference interpreter probe instrumentation

penagos:
Reopening #1763 due to a branch corruption issue. Original text below:

===

As the number of ML frameworks continues to rapidly grow, it’s becoming increasingly desirable to be able to compare numerical results across different runtimes. This ensures that ML models run equivalently in different environments, and overall helps to instill confidence in ML frameworks. For example, given the recency of the StableHLO representation, it may be desirable to have a way to compare StableHLO interpreter results to other frameworks, such as Tensorflow, or other implementations

This PR proposes the addition of a probe instruction to the new interpreter dialect in StableHLO. The probe instruction will enable us to extract intermediate tensor state from the StableHLO interpreter. Since this new op is intended to initially support debuggability, it will not be added to the StableHLO opset itself, but instead to the new [interpreter dialect](https://github.com/openxla/stablehlo/pull/1703).

The probe instruction will take precisely 1 shaped tensor input, and produce the same shaped tensor output, without modifying it:

```
%1 = "interpreter.probe"(%0) {probe_id="probe0"} : (tensor<*xf32) -> tensor<*xf32>
```

The op will be modeled as having side effects, to prevent it from being moved or removed by the compiler. It will additionally contain a singular, unique string attribute, `probe_id`, which will allow us to later pair this tensor’s data with other runtimes. At compile time, the probe instruction may be inserted after every StableHLO operation (except constants), allowing runtime data to be serialized to disk.

This PR also adds a new StableHLO interpreter configuration flag, `--probe_readout_dir`, which will be used to specify the directory in which to store all probed interpreter values. When executed, the probe instruction will associate the contents of its input tensor with the given probe_id, and serialize the tensor data to disk using the portable [NumPy](https://numpy.org/) format. The filename used will be in the form of `probe_readout_dir/probe_id.npy`. As the interpreter executes `interpreter.probe` operations, in addition to serializing tensor data to NumPy files, it will also append serialized filepaths to a metadata `index.csv` file in the form of:

```
probe_id0,/probe_readout_dir/probe_id0.npy
probe_id1,/probe_readout_dir/probe_id1.npy
…
probe_id0,/probe_readout_dir/probe_id0_1.npy
```

The metadata file can then be used by post processing tools to analyze the instrumented tensor data. Note that a `probe_id` can appear more than once in the metadata file (i.e. if a probe is executed within a loop).

### Validation

To validate the new instruction, we propose extending the existing `stablehlo::check` MLIR dialect with a `expect_serialized_eq` operation, which can be utilized in the following way:

```
// Where `T` is a statically shaped tensor
%0 = …
%1 = interpreter.probe %0 { probe_id = "probe0" } : (tensor<T>) -> tensor<T>
check.expect_serialized_eq %3 { probe_id = "probe0", iteration = 0 : i32 } : tensor<T>
```

The `check.expect_serialized_eq` is expected to appear after the producing `interpreter.probe` instruction with the same `probe_id`. When executed, the instruction will search for an existing `index.csv` metadata file in the same directory as specified by the `--probe_readout_dir` flag, and then attempt to find the serialized filepath for the given `probe_id` and iteration. After which, the interpreter will load the serialized tensor from the NumPy file, and then proceed to compare it with its operand tensor. The operation will assert on an answer mismatch or failure to find/parse the associated NumPy file. The iteration attribute is 0-based and is optional and will default to 0.

### DIFF
--- a/stablehlo/reference/InterpreterOps.cpp
+++ b/stablehlo/reference/InterpreterOps.cpp
@@ -17,6 +17,9 @@ limitations under the License.
 
 #include <queue>
 
+#include "llvm/Support/Errc.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/Path.h"
 #include "llvm/Support/ThreadPool.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/IR/BuiltinAttributes.h"
@@ -27,6 +30,7 @@ limitations under the License.
 #include "mlir/IR/SymbolTable.h"
 #include "mlir/Support/LLVM.h"
 #include "stablehlo/reference/InterpreterValue.h"
+#include "stablehlo/reference/NumPy.h"
 #include "stablehlo/reference/Ops.h"
 #include "stablehlo/reference/ProcessGrid.h"
 
@@ -36,6 +40,32 @@ limitations under the License.
 namespace mlir {
 namespace stablehlo {
 namespace interpreter {
+namespace {
+
+// Appends a new line item to an instrumentation metadata file, `index.json` in
+// the form: `probeId,probeOutputDir/filename`.
+llvm::Error writeProbeMetadata(StringRef probeId, StringRef filename,
+                               StringRef probeOutputDir) {
+  if (probeOutputDir.empty())
+    return createStringError(llvm::errc::invalid_argument,
+                             "Probe serialization directory cannot be empty.");
+
+  llvm::SmallString<128> filepath(probeOutputDir);
+  llvm::sys::path::append(filepath, numpy::kInstrumentationMetadataFilename);
+
+  int fd;
+  if (llvm::sys::fs::openFileForWrite(filepath, fd,
+                                      llvm::sys::fs::CD_CreateAlways,
+                                      llvm::sys::fs::OF_Append))
+    return createStringError(llvm::errc::io_error,
+                             "Failed to open instrumentation metadata file.");
+
+  llvm::raw_fd_ostream out(fd, /*shouldClose=*/true);
+  out << probeId.str() << ',' << filename.str() << '\n';
+
+  return llvm::Error::success();
+}
+}  // namespace
 
 //===----------------------------------------------------------------------===//
 // Interpreter Dialect Constructor
@@ -161,6 +191,28 @@ SmallVector<InterpreterValue> evalRunParallelOp(
   for (auto &future : futures) results.append(future.get());
   // TODO(#1725): Figure out how to test the outfeed queue.
   return results;
+}
+
+llvm::Error evalProbeOp(InterpreterValue input, StringRef probeId,
+                        StringRef probeOutputDir,
+                        llvm::StringMap<int32_t> &probeIterations) {
+  llvm::SmallString<128> filepath(probeOutputDir);
+
+  // To properly support loops, append a suffix denoting how many times this
+  // specific probe_id has executed.
+  const int32_t numTimesExecuted = ++probeIterations[probeId];
+
+  llvm::sys::path::append(
+      filepath, probeId + "_" + std::to_string(numTimesExecuted) + ".npy");
+  auto tensor = input.getTensor();
+  if (auto serializationResultError =
+          numpy::serializeTensor(filepath, tensor.getType(), tensor.getData()))
+    return serializationResultError;
+
+  // After the tensor has been serialized to disk, append it to a metadata file
+  // to associate the serialized probe_id with the filepath. By default, this
+  // will live in an `index.csv` file generated in specified `probeOutputDir`.
+  return writeProbeMetadata(probeId, filepath, probeOutputDir);
 }
 
 }  // namespace interpreter
--- a/stablehlo/reference/InterpreterOps.td
+++ b/stablehlo/reference/InterpreterOps.td
@@ -80,4 +80,38 @@ def Interpreter_RunParallelOp : Op<Interpreter_Dialect, "run_parallel", []> {
   let hasVerifier = 1;
 }
 
+def Interpreter_ProbeOp : Op<Interpreter_Dialect, "probe",
+    [SameOperandsAndResultType]> {
+  let arguments = (ins
+    HLO_Tensor:$operand,
+    StrAttr:$probe_id
+  );
+  let results = (outs HLO_Tensor:$result);
+
+  let description = [{
+    Probe and store the values of the input tensor at runtime, using the NumPy
+    file format. Writes tensor input value to
+    `<output-dir>/<probe_id>_<iteration>.npy` (where output-dir is specified by
+    the `--probe_output_dir` flag). Additionally, adds an entry to
+    <output-dir>/index.csv metadata file which maps probe IDs and iterations to
+    filenames with their tensor values.
+
+    The `probe` operation will not modify its input in any way. Probe
+    instrumentation may however slow down the interpretation of a module as
+    there will be increased file I/O.
+
+    Note that `probe_id` should be unique for each `probe` instruction in a
+    StableHLO module. A `probe` may run more than once, in which case it will
+    produce separate serialized data for each iteration in the form
+    `probe_id_#` where # is a 1-based counter.
+
+    Example:
+    ```mlir
+    %result = interpreter.probe %operand, probe_id = "probe0" : tensor<3xi32>
+    ```
+  }];
+
+  let assemblyFormat = "$operand `,` `probe_id` `=` $probe_id attr-dict `:` type($result)";
+}
+
 #endif  // STABLEHLO_REFERENCE_INTERPRETER_OPS
--- a/stablehlo/reference/NumPy.cpp
+++ b/stablehlo/reference/NumPy.cpp
@@ -1,0 +1,346 @@
+/* Copyright 2023 The StableHLO Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#include "stablehlo/reference/NumPy.h"
+
+#include <cctype>
+#include <fstream>
+#include <regex>
+
+#include "llvm/Support/Endian.h"
+#include "llvm/Support/Errc.h"
+#include "llvm/Support/Error.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/raw_ostream.h"
+#include "mlir/IR/BuiltinTypeInterfaces.h"
+#include "mlir/IR/Types.h"
+
+namespace mlir {
+namespace stablehlo {
+namespace numpy {
+namespace {
+
+constexpr char kMagicString[] = "\x93NUMPY";
+constexpr int kMagicStringSize = 6;
+constexpr char kMajorVersion = 0x01;
+constexpr char kMinorVersion = 0x00;
+
+template <typename T>
+struct IsComplexT : public std::false_type {};
+template <typename T>
+struct IsComplexT<std::complex<T>> : public std::true_type {};
+
+}  // namespace
+
+// Determine the NumPy type abbreviation for the given underlying data type T.
+template <typename T>
+static constexpr char getNumPyType() {
+  if constexpr (std::is_same_v<T, bool>) return 'b';
+  if constexpr (IsComplexT<T>::value) return 'c';
+  if constexpr (std::is_floating_point_v<T>) return 'f';
+  if constexpr (std::is_integral_v<T>) {
+    if constexpr (std::is_signed_v<T>)
+      return 'i';
+    else
+      return 'u';
+  }
+
+  llvm::report_fatal_error("Unknown type");
+}
+
+// Creates the NumPy dict header. This will serialize the `descr`,
+// `fortran_order` and `shape` keys after the NumPy magic string. See
+// `parseDescrHeader` for additional information on how the `descr` key is
+// structured.
+template <typename T>
+static void buildNumpyHeaderDict(llvm::raw_fd_ostream& out,
+                                 ArrayRef<int64_t> shape) {
+  // For now, only little endian machines are supported.
+  const auto descr = std::string(1, /*endianness=*/'<') +
+                     std::string(1, getNumPyType<T>()) +
+                     std::to_string(sizeof(T));
+  const auto shapeString =
+      std::accumulate(shape.begin(), shape.end(), std::string(""),
+                      [](const std::string& dest, size_t dim) {
+                        return dest + std::to_string(dim) + ",";
+                      });
+  std::string dict;
+  std::stringstream outDict(dict);
+
+  outDict << "{'descr': '" << descr << "', ";
+  outDict << "'fortran_order': False, ";
+  outDict << "'shape' : (" << shapeString << "), }";
+
+  // The NumPy header is padded with spaces for proper alignment.
+  // Account for newline and the magic string length.
+  const auto headerSize = static_cast<int>(out.tell());
+  const auto padding = 16 - (headerSize + 1) % 16;
+  outDict << std::string(padding - 1, ' ') << '\n';
+
+  const auto dictLength = static_cast<uint16_t>(outDict.str().size());
+  out << (char)(dictLength & 0xff) << (char)((dictLength >> 8) & 0xff)
+      << outDict.str();
+}
+
+template <typename T>
+static void buildNumpyHeader(llvm::raw_fd_ostream& out,
+                             ArrayRef<int64_t> shape) {
+  out.write(kMagicString, kMagicStringSize);
+  out.write(kMajorVersion);
+  out.write(kMinorVersion);
+  buildNumpyHeaderDict<T>(out, shape);
+}
+
+static llvm::Error parseFortranOrderKey(const std::string& header) {
+  const std::size_t fortranOrderOffset = header.find("'fortran_order':");
+  if (fortranOrderOffset == std::string::npos)
+    return llvm::createStringError(llvm::errc::invalid_argument,
+                                   "Failed to find fortran_order header.");
+
+  if (header.find("False", fortranOrderOffset) == std::string::npos)
+    return llvm::createStringError(llvm::errc::invalid_argument,
+                                   "Only fortran_order: False is supported.");
+
+  return llvm::Error::success();
+}
+
+// Parses the NumPy `descr` header, which is in the following format:
+// 'descr': '<i4'
+// Where the first character determines the endianness of the data (< for little
+// endian, > for big endian), the next character determines the data type (i.e.
+// unsigned int, float, bool, etc) and the last number(s) determine the data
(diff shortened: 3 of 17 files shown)


GleasonK:
Adding link to OpenXLA Discuss RFC:
https://groups.google.com/a/openxla.org/g/openxla-discuss/c/Y1875kCT3Hc

stablehlo/tools/StablehloTranslateMain.cpp
```diff
@@ -83,7 +99,8 @@ llvm::Error evalCustomCallCheckEq(stablehlo::CustomCallOp op,
 
 llvm::Error interpreterFallback(Operation &op, stablehlo::Process *process,
                                 stablehlo::Scope &scope,
-                                llvm::StringRef funcName) {
+                                llvm::StringRef funcName,
+                                llvm::StringMap<int32_t> &instrumentedTensors) {
```
GleasonK, line 103: Per earlier feedback, please update the name you decide to use for `instrumentedTensors` here as well.

I'll think more on if this is how we should expose this. I don't love adding new input arguments here, but this state needs to be stored somewhere. I think this is good for now, but if there were one more thing added that required state, perhaps we should add an `InterpreterConfig` / `InterpreterState` class to manage stateful `InterpreterDialect` ops.

stablehlo/tools/StablehloTranslateMain.cpp
```diff
@@ -160,6 +197,8 @@ TranslateFromMLIRRegistration interpretRegistration(
     [](ModuleOp module, raw_ostream &os) {
       auto numFuncs = 0;
       bool hasMain = false;
+      llvm::StringMap<int32_t> instrumentedTensors;
```
GleasonK, line 200: update name (see earlier comment)

GleasonK (changes requested):
Left a few more thoughts. I'll need to take a deeper look at the numpy serialization code, but overall looks good

stablehlo/reference/InterpreterOps.cpp
```diff
@@ -163,6 +193,28 @@ SmallVector<InterpreterValue> evalRunParallelOp(
   return results;
 }
 
+llvm::Error evalProbeOp(InterpreterValue input, StringRef probeId,
+                        StringRef probeOutputDir,
+                        llvm::StringMap<int32_t> &instrumentedTensors) {
+  auto tensor = input.getTensor();
```
ghpvnist, line 199: Let's move this line close to where its used.

stablehlo/reference/InterpreterOps.cpp
```diff
@@ -163,6 +193,28 @@ SmallVector<InterpreterValue> evalRunParallelOp(
   return results;
 }
 
+llvm::Error evalProbeOp(InterpreterValue input, StringRef probeId,
+                        StringRef probeOutputDir,
+                        llvm::StringMap<int32_t> &instrumentedTensors) {
```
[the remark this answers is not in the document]
ghpvnist: +1, and make sure the variable name is also updated in the header.

ghpvnist (changes requested):

stablehlo/reference/Ops.cpp
```diff
@@ -2003,16 +2003,17 @@ Tuple evalTupleOp(ArrayRef<InterpreterValue> val, TupleType resultType) {
...
+    llvm::function_ref<llvm::Error(Operation &, Process *, Scope &)> fallback) {
   SmallVector<InterpreterValue> results(operand);
 
   auto condResults = eval(cond, operand, process, &scope);
 
   while (condResults[0].getTensor().get({}).getBooleanValue()) {
-    results = eval(body, results, process, &scope);
-    condResults = eval(cond, results, process, &scope);
+    results = eval(body, results, process, &scope, fallback);
+    condResults = eval(cond, results, process, &scope, /*fallback=*/nullptr);
```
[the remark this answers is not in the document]
penagos: Initially I hadn't expected any `probe` instructions to appear in the while condition regions. However I suppose there could be other custom ops that have kernels defined in the fallback function, so added it in.

stablehlo/reference/NumPy.cpp
```diff
@@ -0,0 +1,344 @@
...
+    return Functor<int64_t>()(std::forward<Args>(args)...);
+  if (type.isF16()) return Functor<uint16_t>()(std::forward<Args>(args)...);
+  if (type.isF32()) return Functor<float>()(std::forward<Args>(args)...);
+  if (type.isF64()) return Functor<double>()(std::forward<Args>(args)...);
+  if (auto complexTy = type.dyn_cast<ComplexType>()) {
+    auto complexElemTy = complexTy.getElementType();
+    if (complexElemTy.isF32())
+      return Functor<double>()(std::forward<Args>(args)...);
+
+#ifdef __SIZEOF_INT128__
```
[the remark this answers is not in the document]
penagos: For proper `f64` complex type serialization, we'd need a 128 bit native type to properly go to/from disk (64 bits for each component). Since it's not standardized, I think we need a preprocessor check here. We can however use a standard `double` for `f32` complex types. Added a comment.

stablehlo/tools/StablehloTranslateMain.cpp
penagos: I think maybe we can tackle a more elegant solution as part of #1780?

stablehlo/reference/NumPy.cpp
```diff
@@ -0,0 +1,349 @@
...
+  // This utility is presently only used from the context of the check dialect
+  // for validation. It suffices to only support NumPy v1.0 for now.
+  char majorVersion, minorVersion;
+  in.read(&majorVersion, 1);
+  in.read(&minorVersion, 1);
+
+  if (majorVersion != kMajorVersion || minorVersion != kMinorVersion)
+    return llvm::createStringError(
+        llvm::errc::invalid_argument,
+        "Invalid NumPy version detected. Only v1.0 supported.");
```
ghpvnist, line 196: Can you provide dynamic error message, something like:
`Invalid NumPy version: %s. Expected version to be %s or %s`

stablehlo/tests/interpret_probe.mlir
```diff
@@ -0,0 +1,84 @@
+// RUN: stablehlo-translate --interpret --probe-output-dir=%T -split-input-file %s
+
+func.func @probe_i1() {
+  %0 = stablehlo.constant dense<[[0], [0], [0]]> : tensor<3x1xi1>
+  %1 = stablehlo.constant dense<[[1], [1], [1]]> : tensor<3x1xi1>
+  %2 = stablehlo.add %0, %1 : tensor<3x1xi1>
+  %3 = interpreter.probe %2 { probe_id = "probe_i1" } : (tensor<3x1xi1>) -> tensor<3x1xi1>
```
ghpvnist, line 7: Optional QoL improvement:
Would be nice if we can write:
`%3 = interpreter.probe %2 { probe_id = "probe_i1" } : (tensor<3x1xi1>) -> tensor<3x1xi1>`
to something shorter:
`%3 = interpreter.probe %2, id = "probe_i1" : tensor<3x1xi1>`
or even:
`%3 = interpreter.probe %2, "probe_i1" : tensor<3x1xi1>`

I think this is just a 1 line change in the InterpreterOps.td file:
```
let assemblyFormat = "$operand `,` $probe_id attr-dict `:` type($operand)";
```

stablehlo/tests/CheckOps.cpp
```diff
@@ -75,6 +81,53 @@ llvm::Error evalExpectEqOp(const Tensor &lhs, const Tensor &rhs) {
...
+  llvm::SmallString<128> instrumentationMetadataFile(probeDir);
+  llvm::sys::path::append(instrumentationMetadataFile,
+                          numpy::kInstrumentationMetadataFilename);
+  std::ifstream metadataFile(instrumentationMetadataFile.str().str());
+
+  if (!metadataFile.is_open()) return llvm::errc::io_error;
+
+  std::string line;
+
+  for (uint32_t match = 0; metadataFile >> line && match <= iteration;
```
ghpvnist, line 101: Fail early when match not found?
```
  for (uint32_t match = 0; metadataFile >> line && match <= iteration;
       ++match) {
    auto pos = line.find(probeId);
    if (pos == std::string::npos) return llvm::errc::bad_address;
    if (match == iteration) return line.substr(pos + probeId.size() + 1);
```

stablehlo/tests/interpret_probe.mlir
```diff
@@ -0,0 +1,84 @@
...
+    stablehlo.return %cond : tensor<i1>
+  } do {
+    %new_sum = stablehlo.add %arg1, %one : tensor<i64>
+    %new_sum_instrumented = interpreter.probe %new_sum { probe_id = "probe_iterations" } : (tensor<i64>) -> tensor<i64>
+    %new_i = stablehlo.add %arg0, %one : tensor<i64>
+    stablehlo.return %new_i, %new_sum_instrumented : tensor<i64>, tensor<i64>
+  }
+
+  check.expect_eq_const %results0, dense<2> : tensor<i64>
+  check.expect_serialized_eq %one { probe_id = "probe_iterations", iteration = 0 : ui32 } : tensor<i64>
```
ghpvnist, line 81: Similarly (but also optional):
`check.expect_serialized_eq %one { probe_id = "probe_iterations", iteration = 0 : ui32 } : tensor<i64>`
into:
`check.expect_serialized_eq %one, "probe_iterations", i = 0 : tensor<i64>`

Something like:
```
let assemblyFormat = [{
  $operand `,` $expected `,` $probe_id
  (`,` `i` `=` $iteration^)?
  attr-dict `:` type($expected)
}];
```

ghpvnist (changes requested):
Minor feedbacks and some QoL improvements

stablehlo/tests/CheckOps.cpp
ghpvnist: Unresolving since the updated commit doesn't seem to address this.

stablehlo/tests/interpret_probe.mlir
GleasonK: Two (undocumented) design principles for pretty printing:
1. Prefer N=V format as a default, omitting `Name=` is an exception for cases where it is overly redundant, like equality comparison type (`EQ`) for the sake of looking like `arith` dialect printing. Would suggest `id = $probe_id`
2. Abbreviations are fine, but prefer something that indicates the field name (i=index? i=iteration?) so would prefer `iter`, similar to how we use `dims` instead of `d`.

```
let assemblyFormat = [{
  $operand `,` $expected `,` `id` `=` $probe_id
  (`,` `iter` `=` $iteration^)?
  attr-dict `:` type($expected)
}];
```

stablehlo/tests/CheckOps.cpp
penagos: Sorry meant to reply to this one. I don't think those proposed changes are equivalent to what the loop is intending to achieve. That is, if we don't find the requested `probeId` in the current line, we break from the loop and return `llvm:errc::bad_address` instead of continuing to check other lines.

ghpvnist: Got it, thanks!

stablehlo/tests/interpret_probe.mlir
penagos: I  slightly tweaked the proposed `td` changes to have the attribute on `check.expect_serialized_eq` be `probe_id` instead to imply it's to be used with probe instrumentation.

stablehlo/reference/InterpreterOps.td
```diff
@@ -80,4 +80,38 @@ def Interpreter_RunParallelOp : Op<Interpreter_Dialect, "run_parallel", []> {
   let hasVerifier = 1;
 }
 
+def Interpreter_ProbeOp : Op<Interpreter_Dialect, "probe",
+    [SameOperandsAndResultType, SameOperandsAndResultShape]> {
```
ghpvnist, line 84: `SameOperandsAndResultShape` is redundant with `SameOperandsAndResultType`.

stablehlo/tests/CheckOps.td
```diff
@@ -118,4 +118,31 @@ def CHECK_ExpectEqOp : Op<CHECK_Dialect, "expect_eq", [SameTypeOperands]> {
...
+    Verifies that the value of the serialized tensor `probe_id` matches the
+    optionally specified input tensor at iteration `iteration`, using previously
+    serialized filepaths in `index.csv`.
+
+    ```mlir
+    check.expect_serialized_eq %arg0, 
+      probe_id = "probe0",
+      iter = 0
+    } : tensor<2xi32>
+    ```
```
ghpvnist, line 133: Typo on the `}`:
```mlir
check.expect_serialized_eq %arg0, probe_id = "probe0", iter = 0
    : tensor<2xi32>
```

ghpvnist (changes requested):
LGTM with remaining feedback

ghpvnist (approved):

stablehlo/reference/NumPy.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+  header.erase(std::remove_if(header.begin(), header.end(), isspace),
+               header.end());
+
+  auto wordSizeOrError = parseDescrHeader<T>(header);
+  if (!wordSizeOrError)
+    return llvm::createStringError(llvm::errc::invalid_argument,
+                                   "Failed to parse descr header.");
+
+  if (parseFortranOrderKey(header))
```
GleasonK, line 218: Lets capture this result in `status`, that seems to be the convention for `llvm::Error` handling in our codebase. Otherwise it's confusing why the other two error checks use `!value` and this one doesn't negate

```
auto status = parseFortranOrderKey(header);
if (status) ...
```

stablehlo/reference/InterpreterOps.td
```diff
@@ -80,4 +80,38 @@ def Interpreter_RunParallelOp : Op<Interpreter_Dialect, "run_parallel", []> {
...
+    HLO_Tensor:$operand,
+    StrAttr:$probe_id
+  );
+  let results = (outs HLO_Tensor:$result);
+
+  let description = [{
+    Probe and store the values of the input tensor at runtime, using the NumPy
+    file format and using the `probe_id` string as the serialization filename,
+    storing results in the directory specified by the `--probe-output-dir`
+    flag.
```
GleasonK, line 95: Add a little more detail on the I/O outputs of this op. Given that they are encoded in `evalProbeOp` they are a part of the semantics of the op. I'm realizing this description serves as the only documentation for this interpreter feature currently:

Something like the following (please correct any wording if I have it wrong):

"Writes tensor input value to `<output-dir>/<probe_id>_<iteration>.npy`. Additionally, adds an entry to `<output-dir>/index.csv` metadata file which maps probe IDs and iterations to filenames with their tensor values"

GleasonK (changes requested):
No major code changes, just nits and build file updates at this point. Should be good to merge soon

stablehlo/reference/NumPy.cpp
```diff
@@ -0,0 +1,351 @@
...
+  if (auto quantized = type.dyn_cast<quant::QuantizedType>()) {
+    switch (quantized.getStorageTypeIntegralWidth()) {
+      case 8:
+        return Functor<int8_t>()(std::forward<Args>(args)...);
+      case 16:
+        return Functor<int16_t>()(std::forward<Args>(args)...);
+      default:
+        llvm::report_fatal_error("Unhandled quantized type");
+    }
+  }
```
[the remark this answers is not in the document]
penagos: Hm, is it worthwhile to add this back in if/when it is supported? Or should we leave it in so it works OOB once support is added? (deferring to @ghpvnist )

stablehlo/reference/NumPy.cpp
```diff
@@ -0,0 +1,351 @@
...
+  if (descrOffset == std::string::npos) return llvm::errc::invalid_argument;
+
+  const std::size_t needleSize = header.find(',', descrOffset + kDescrSize + 1);
+  std::string typeString = header.substr(
+      descrOffset + kDescrSize, needleSize - (descrOffset + kDescrSize));
+
+  if (typeString.front() != '\'' || typeString.back() != '\'')
+    return llvm::errc::invalid_argument;
+
+  typeString = typeString.substr(1, typeString.size() - 2);
```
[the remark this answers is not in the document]
penagos: Yes, that's the intended semantics. Added some more comments to make it clearer.

stablehlo/reference/NumPy.cpp
```diff
@@ -0,0 +1,351 @@
...
+  if (majorVersion != kMajorVersion || minorVersion != kMinorVersion)
+    return llvm::createStringError(
+        llvm::errc::invalid_argument,
+        "Invalid NumPy version: %c.%c. Expected version to be %c.%c.",
+        majorVersion, minorVersion, kMajorVersion, kMinorVersion);
+
+  char headerSizeBuffer[2];
+  if (!in.read(headerSizeBuffer, 2))
+    return llvm::createStringError(llvm::errc::io_error,
+                                   "Failed to read NumPy header size.");
```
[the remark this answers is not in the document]
penagos: A well formed NumPy file should have its header size contained within 2 bytes per the spec (see https://numpy.org/doc/1.13/neps/npy-format.html#format-specification-version-1-0)

1 byte for the major version, and 1 for the minor version

stablehlo/reference/NumPy.cpp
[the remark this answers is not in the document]
GleasonK: I'd prefer removal. It's best to consciously add support for these ops instead of predict how the undesigned interpreter feature will work in the future. I.e. maybe we need `QuantizedTensor` type and very specific ops support this new type.. now this code will never be hit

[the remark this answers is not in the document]
penagos: Sounds good, I've removed this from this PR.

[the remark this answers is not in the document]
ghpvnist: I also prefer removing unused features until it is fully supported. Things may change which could make this change stale or break.

stablehlo/reference/InterpreterOps.td
```diff
@@ -80,4 +80,40 @@ def Interpreter_RunParallelOp : Op<Interpreter_Dialect, "run_parallel", []> {
...
+
+    Example:
+    ```mlir
+    %result = "interpreter.probe"() {
+      probe_id="probe0"
+    } : () -> (tensor<ui32>, tensor<ui32>)
+    ```
+  }];
+
+  let assemblyFormat = "$operand `,` $probe_id attr-dict `:` type($result)";
```
GleasonK, line 116:
```suggestion
    Example:
    ```mlir
    %result = interpreter.probe %operand, probe_id = "probe0" : tensor<3xi32>
    ```
  }];

  let assemblyFormat = "$operand `,` `probe_id` `=` $probe_id attr-dict `:` type($result)";
```

stablehlo/tests/interpret_probe.mlir
```diff
@@ -0,0 +1,84 @@
+// RUN: stablehlo-translate --interpret --probe-output-dir=%T -split-input-file %s
+
+func.func @probe_i1() {
+  %0 = stablehlo.constant dense<[[0], [0], [0]]> : tensor<3x1xi1>
+  %1 = stablehlo.constant dense<[[1], [1], [1]]> : tensor<3x1xi1>
+  %2 = stablehlo.add %0, %1 : tensor<3x1xi1>
+  %3 = interpreter.probe %2, "probe_i1" : tensor<3x1xi1>
```
GleasonK, line 7: Per prettyprint changes above:

`probe_id = "probe_i1"`

GleasonK: same with rest of file

GleasonK (approved):

GleasonK:
Changes merged! Thanks for the contribution!

Issue #1576, opened 2023-06-04
Consider merging UniformDequantizeOp and UniformQuantizeOp into ConvertOp


burmako:
Why do we need three ops to express conversion between different kinds of types if one op (ConvertOp) appear to be sufficient? Sure, for unknown scales and zero points we'll want three ops, but we don't support them yet.

sdasgup3:
I agree with this proposal. In the current scenario, where we do not have support for unknown scales and zero-points, we do not need any special operation. We can re-evaluate the need for such additional ops on need basis. 
 
With that I would propose to do the merge sooner than later. Otherwise, it may impact the producers/consumers of StableHLO who might have already adopted the `uniformQuantize/UniformDequantize` ops.

sdasgup3:
@abattery 
Thanks for the feedback! 

>> This simplification might block such efforts.

Can you please elaborate a bit on this?  I am trying to understand if the `stablehlo.convert` op can be used for those efforts. Is this because you plan to add additional arguments to `stablehlo.quantize/stablehlo.dequantize` ops?

abattery:
Yes, in quantized training, scales and zps should be calculated on the fly, which means that ops for quantization/dequantization should be enough flexible to get such parameters in the graph level.

rewu93:
`unform_dequantize` is also used to requantize a quantized tensor. According to the spec, `convert` is `Performs an element-wise conversion from one element type to another on operand tensor and produces a result tensor.` 
Not sure if using `convert` here is clean (operands and results have the same type).

sdasgup3:
>> unform_dequantize is also used to requantize a quantized tensor
I assume you meant `stablehlo.uniform_quantize` can also be used to re-quantize a quantized tensor. 

From the [convert spec](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#convert), only the shape should be the same, element_type could be different.

rewu93:
> > > unform_dequantize is also used to requantize a quantized tensor
> > > I assume you meant `stablehlo.uniform_quantize` can also be used to re-quantize a quantized tensor.
> 
> From the [convert spec](https://github.com/openxla/stablehlo/blob/main/docs/spec.md?rgh-link-date=2023-06-12T18%3A43%3A53Z#convert), only the shape should be the same, element_type could be different.

Yes. But if we requantize a qi8 to another qi8 with different scale/zero point (or different axis), element_type` will be the same.  Maybe I misunderstood `convert` to `cast`.

[...]

PR #1796, opened 2023-10-09
Specification of quantized `reduce` and other related ops

sdasgup3:
The PR implements the approved [RFC for reduce op](https://github.com/openxla/stablehlo/pull/1664) by proposing the specification related changes for `reduce`, `reduce_window` and `select_and_scatter` ops.

In https://github.com/openxla/stablehlo/pull/1647, we talked about some  of the other ops which will depend on the quantized specification of reduce op. Initially, I thought about creating separate PRs for them, but for the interest of time and the fact that their handling is going to be very similar to how `reduce` op is handled, I propose to include their PR in the current PR.

Here are the  additional ops (other than `reduce`, `reduce_window`, `select_and_scatter`) whose specification is added
 - Ops with explicit computation regions: `all_reduce`, `reduce_scatter`, `scatter`: They are handled similar to how `reduce` op is handled. 

 - Ops w/o explicit computation region:  `batch_norm_grad`, `batch_norm_training`: For these ops, the semantics of the operation implicitly does reduction with a custom computation function. As there is no explicit computation function in the IR, the proposal in the RFC cannot be applied. I propose (A) to handle these ops similar to how `batch_norm_inference` is handled with `dequant-op-quant` strategy, (B) we can revisit this op later if there is use case to do that implicit reduction using higher accumulation type.  
 
One implementation detail: The fact that  `batch_norm_grad`, `batch_norm_training` ops returns multiple outputs and current `dequantize_op_quantize` returns a single output, is handled using a dedicated meta function for these two ops. 

Next steps: Once the PR is approved the plan is to propose the corresponding changes in the verifier/shape functions for these ops.

**Only support promotion of reduction operands to reduction-block arguments**
The specific question bought up in the discussion is: should we promote demotion as well? It seems that there aren't any use-cases for that. For example, consider the following cases of implicit conversion for reduction operations. The examples are shown using integer type but can be generalized to other types as well.
 
| input element type | accumulator element type  | result element type | use-cases |
|:--:|:--:|:--:|:--|
| i8| i8 |i8| Reduction using min/max|
| i8| i32 |i32| Reduction using average/sum|
| i8| i32| i8 | Demotion from accumulation type to result type. (No use-case)|
| i8| i32| i64 | Extra promotion from accumulation type to result type. (No use-case) |
| i32 | i8 | i8 | Demotion from input value type to accumulation type. (No use case) |

With that in mind, we propose to develop the specification based on the scenario we have the use-case for (just for promotion: _none of the cases of demotion or extra promotion will be supported in the specification_.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -778,8 +778,8 @@ defined as follows:
 
 Afterwards, within each `process_group`:
 
-* `result@process[result_index] = to_destination_type(exec(schedule),
-  type(result))` for some binary tree `schedule` where:
+* `result@process[result_index] = exec(schedule)` for some binary tree
+  `schedule` where:
   * `exec(node)` = `computation(exec(node.left), exec(node.right))`.
   * `exec(leaf)` = `leaf.value`.
 * `schedule` is an implementation-defined binary tree whose in-order
@@ -800,7 +800,7 @@ Afterwards, within each `process_group`:
 
 | Name     | Type                                  | Constraints |
 |----------|---------------------------------------|-------------|
-| `result` | tensor or per-tensor quantized tensor | (C6)        |
+| `result` | tensor or per-tensor quantized tensor | (C6-C7)     |
 
 #### Constraints
 
@@ -812,8 +812,9 @@ Afterwards, within each `process_group`:
 * (C3) `0 <= replica_groups < size(replica_groups)`.
 * (C4) If `use_global_device_ids = true`, then `channel_id > 0`.
 * (C5) `computation` has type `(tensor<E>, tensor<E>) -> (tensor<E>)` where
-       `is_promotable(E, element_type(operand))`.
-* (C6) `is_promotable(type(result), type(operand))`.
+       `is_promotable(element_type(operand), E)`.
+* (C6) `shape(result) = shape(operand)`.
+* (C7) `element_type(result) = E`.
 
 #### Examples
 
@@ -4074,8 +4075,7 @@ doesn't hold for many popular reductions. E.g. floating-point addition for
 `body` and zero for `init_values` don't actually form a monoid because
 floating-point addition is not associative.
 
-More formally, `results...[j0, ..., jR-1] = to_destination_type(
-reduce(input_slices_converted), type(results...)))` where:
+More formally, `results...[j0, ..., jR-1] = reduce(input_slices_converted)` where:
 
 * `input_slices = inputs...[j0, ..., :, ..., jR-1]`, where `:` are inserted
   at `dimensions`.
@@ -4119,10 +4119,10 @@ reduce(input_slices_converted), type(results...)))` where:
 * (C5) `is_unique(dimensions)`.
 * (C6) `body` has type `(tensor<E0>, ..., tensor<EN-1>, tensor<E0>, ...,`
   `tensor<EN-1>) -> (tensor<E0>, ..., tensor<EN-1>)` where
-  `is_promotable(element_type(inputs[i]), element_type(E[i]))`.
+  `is_promotable(element_type(inputs[i]), Ei)`.
 * (C7) `shape(results...) = shape(inputs...)` except that the dimension
   sizes of `inputs...` corresponding to `dimensions` are not included.
-* (C8) `is_promotable(element_type(inputs...), element_type(results...))`.
+* (C8) `element_type(results[i]) = Ei` for all `i` in `[0,N)`.
 
 #### Examples
 
@@ -4244,7 +4244,7 @@ Afterwards, within each `process_group`:
 
 | Name     | Type                                  | Constraints |
 |----------|---------------------------------------|-------------|
-| `result` | tensor or per-tensor quantized tensor | (C8)        |
+| `result` | tensor or per-tensor quantized tensor | (C8-C9)     |
 
 #### Constraints
 
@@ -4258,10 +4258,11 @@ Afterwards, within each `process_group`:
 * (C5) `0 <= replica_groups < size(replica_groups)`.
 * (C6) If `use_global_device_ids = true`, then `channel_id > 0`.
 * (C7) `computation` has type `(tensor<E>, tensor<E>) -> (tensor<E>)` where
-       `is_promotable(E, element_type(operand))`.
-* (C8) `is_promotable(type(result), type(operand))` except:
+       `is_promotable(element_type(operand), E)`.
+* (C8) `shape(result) = shape(operand)` except:
   * `dim(result, scatter_dimension) = dim(operand, scatter_dimension) /
     dim(process_groups, 1)`.
+* (C9) `element_type(result) = E`.
 
 #### Examples
 
@@ -4318,7 +4319,7 @@ where:
 | Label | Name                | Type                                                                     | Constraints                                     |
 |-------|---------------------|--------------------------------------------------------------------------|-------------------------------------------------|
 | (I1)  | `inputs`            | variadic number of tensors or per-tensor quantized tensors               | (C1-C4), (C6), (C8), (C10), (C12), (C13), (C15) |
-| (I2)  | `init_values`       | variadic number of 0-dimensional tensors or per-tensor quantized tensors | (C1), (C13), (C16)                              |
+| (I2)  | `init_values`       | variadic number of 0-dimensional tensors or per-tensor quantized tensors | (C1), (C13)                                     |
 | (I3)  | `window_dimensions` | 1-dimensional tensor constant of type `si64`                             | (C4), (C5), (C15)                               |
 | (I4)  | `window_strides`    | 1-dimensional tensor constant of type `si64`                             | (C6), (C7), (C15)                               |
 | (I5)  | `base_dilations`    | 1-dimensional tensor constant of type `si64`                             | (C8), (C9), (C15)                               |
@@ -4349,7 +4350,7 @@ where:
 * (C12) `shape(padding) = [rank(inputs[0]), 2]`.
 * (C13) `body` has type `(tensor<E0>, ..., tensor<EN-1>, tensor<E0>, ...,`
   `tensor<EN-1>) -> (tensor<E0>, ..., tensor<EN-1>)` where
-  `is_promotable(element_type(inputs[i]), element_type(E[i]))`.
+  `is_promotable(element_type(inputs[i]), Ei)`.
 * (C14) `same(shape(results...))`.
 * (C15) `shape(results[0]) = num_windows` where:
   * `dilated_input_shape = shape(inputs[0]) = 0 ? 0 : (shape(inputs[0]) - 1) * base_dilations + 1`.
@@ -4357,7 +4358,7 @@ where:
   * `dilated_window_shape = (window_dimensions - 1) * window_dilations + 1`.
   * `is_empty_window = padded_input_shape = 0 || dilated_window_shape > padded_input_shape`.
   * `num_windows = is_empty_window ? 0 : floor((padded_input_shape - dilated_window_shape) / window_strides) + 1`.
-* (C16) `is_promotable(element_type(results...), element_type(init_values...))`.
+* (C16) `element_type(results[i]) = Ei` for all `i` in `[0,N)`.
 <!-- markdownlint-enable line-length -->
 
 #### Examples
@@ -4816,17 +4817,12 @@ Given that, `results = exec(schedule, inputs)`, where:
   `index_space(updates[0])`.
 * `exec([update_index, ...], results) = exec([...], updated_results)` where:
   * If `result_index` is in bounds for `shape(results...)`
-    * `results_converted = to_destination_type(
-      results...[result_index], type(func_inputs(update_computation)
-      [:len(func_inputs(update_computation))//2])... )`
     * `updates_converted = to_destination_type(
       updates...[update_index], type(func_inputs(update_computation)
       [len(func_inputs(update_computation))//2:])... )`
-    * `updated_values = update_computation(result_converted, updates_converted)`
-    * `updated_values_converted = to_destination_type(
-           updated_values, type(results...))`
+    * `updated_values = update_computation(results...[result_index], updates_converted)`
     * `updated_results` is a copy of `results` with `results...[result_index]`
-      set to `updated_values_converted...`.
+      set to `updated_values...`.
   * Otherwise
     * `updated_results = results`.
 * `exec([], results) = results`.
@@ -4860,7 +4856,7 @@ undefined.
 
 | Name      | Type                                                       | Constraints |
 |-----------|------------------------------------------------------------|-------------|
-| `results` | variadic number of tensors or per-tensor quantized tensors | (C15)       |
+| `results` | variadic number of tensors or per-tensor quantized tensors | (C15-C17)   |
 
 #### Constraints
 
@@ -4893,8 +4889,9 @@ undefined.
 * (C14) `0 <= index_vector_dim <= rank(scatter_indices)`.
 * (C15) `update_computation` has type `(tensor<E0>, ..., tensor<EN-1>,
   tensor<E0>, ..., tensor<EN-1>) -> (tensor<E0>, ..., tensor<EN-1>)`,
-  where `is_promotable(Ei, element_type(inputs[i]))`.
-* (C16) `is_promotable(type(inputs...), type(results...))`.
+  where `is_promotable(element_type(inputs[i]), Ei)`.
+* (C16) `shape(inputs...) = shape(results...)`.
+* (C17) `element_type(results[i]) = Ei` for all `i` in `[0,N)`.
 
 #### Examples
 
@@ -5031,7 +5028,7 @@ More formally:
 
 | Name     | Type                                  | Constraints |
 |----------|---------------------------------------|-------------|
-| `result` | tensor or per-tensor quantized tensor | (C11)       |
+| `result` | tensor or per-tensor quantized tensor | (C11-C12)   |
 
 #### Constraints
 
@@ -5050,8 +5047,9 @@ More formally:
 * (C9) `select` has type `(tensor<E>, tensor<E>) -> tensor<i1>` where
        `E = element_type(operand)`.
 * (C10) `scatter` has type `(tensor<E>, tensor<E>) -> tensor<E>` where
-  `is_promotable(element_type(operand), element_type(E))`.
-* (C11) `is_promotable(type(operand), type(result))`.
+  `is_promotable(element_type(operand), E)`.
+* (C11) `shape(operand) = shape(result)`.
+* (C12) `element_type(result) = E`.
 <!-- markdownlint-enable line-length -->
 
 #### Examples
@@ -6342,15 +6340,22 @@ currently used in context of reduction computation (refer to
 
 ```python
 def is_promotable(x: Type, y: Type) -> Value:
-  if x == Type and y == Type:
-    return shape(x) == shape(y) and is_promotable(element_type(x), element_type(y))
+  is_same_type = (is_bool(x) and is_bool(y)) or
+    (is_integer(x) and is_integer(y)) or (is_float(x) and is_float(y)) or
+    (is_complex(x) and is_complex(y)) or
+    (is_quantized(x) and is_quantized(y) and expressed_type(x) = expressed_type(y))
 
-  if x != Type and y != Type:
-    return (is_bool(x) and is_bool(y)) or
-      (is_integer(x) and is_integer(y)) or
-      (is_float(x) and is_float(y)) or
-      (is_complex(x) and is_complex(y)) or
-      (is_quantized(x) and is_quantized(y) and expressed_type(x) = expressed_type(y))
+  if is_same_type == False:
+    return False
+
+  if is_integer(x) or is_float(x):
+    return bitwidth(x) <= bitwidth(y)
+
+  if is_complex(x):
+    return bitwidth(element_type(x)) <= bitwidth(element_type(y))
+
+  if is_quantized(x):
+    return bitwidth(storage_type(x)) <= bitwidth(storage_type(y))
 
   return false
 ```
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -778,13 +778,13 @@ defined as follows:
 
 Afterwards, within each `process_group`:
 
-* `result@process[result_index] = convert_or_quantize_or_dequantize(exec(
-  schedule), type(result))` for some binary tree `schedule` where:
+* `result@process[result_index] = to_destination_type(exec(schedule),
+  type(result))` for some binary tree `schedule` where:
   * `exec(node)` = `computation(exec(node.left), exec(node.right))`.
   * `exec(leaf)` = `leaf.value`.
 * `schedule` is an implementation-defined binary tree whose in-order
-  traversal is `convert_or_quantize_or_dequantize(
-  operands@process_group...[result_index], type(func_inputs(computation)[0]))`.
+  traversal is `to_destination_type(operands@process_group...[result_index],
+  type(func_inputs(computation)[0]))`.
 
 #### Inputs
 
@@ -1092,11 +1092,8 @@ grad_output, type(grad_operand), type(grad_scale), type(feature_index))`.
 #### Constraints
 
 * (C1) `0 <= feature_index < rank(operand)`.
-* (C2) `baseline_element_type(operand) = baseline_element_type(scale) =
-       baseline_element_type(mean) = baseline_element_type(variance) =
-       baseline_element_type(grad_output) = baseline_element_type(grad_operand)
-       = baseline_element_type(grad_scale) =
-       baseline_element_type(grad_offset)`.
+* (C2) `operand`, `scale`, `mean`, `variance`, `grad_output`, `grad_operand`,
+       `grad_scale` and `grad_offset` have the same `baseline_element_type`.
 * (C3) `operand`, `grad_output` and `grad_operand` have the same shape.
 * (C4) `scale`, `mean`, `variance`, `grad_scale` and `grad_offset` have the
        same shape.
@@ -1183,9 +1180,8 @@ feature_index), operand, scale, offset, mean, variance, type(result))`.
 #### Constraints
 
 * (C1) `0 <= feature_index < rank(operand)`.
-* (C2) `baseline_element_type(operand) = baseline_element_type(scale) =
-       baseline_element_type(offset) = baseline_element_type(mean) =
-       baseline_element_type(variance) = baseline_element_type(result)`.
+* (C2) `operand`, `scale`, `offset`, `mean`, `variance` and `result` have the
+       same `baseline_element_type`.
 * (C3) `size(scale) = dim(operand, feature_index)`.
 * (C4) `size(offset) = dim(operand, feature_index)`.
 * (C5) `size(mean) = dim(operand, feature_index)`.
@@ -1275,9 +1271,8 @@ scale, offset, type(output), type(batch_mean), type(batch_var))`.
 #### Constraints
 
 * (C1) `0 <= feature_index < rank(operand)`.
-* (C2) `baseline_element_type(operand) = baseline_element_type(scale) =
-       baseline_element_type(offset) = baseline_element_type(batch_mean) =
-       baseline_element_type(batch_var) = baseline_element_type(output)`.
+* (C2) `operand`, `scale`, `offset`, `batch_mean`, `batch_var` and `output` have
+       the same `baseline_element_type`.
 * (C3) `size(scale) = dim(operand, feature_index)`.
 * (C4) `size(offset) = dim(operand, feature_index)`.
 * (C5) `size(batch_mean) = dim(operand, feature_index)`.
@@ -4079,13 +4074,15 @@ doesn't hold for many popular reductions. E.g. floating-point addition for
 `body` and zero for `init_values` don't actually form a monoid because
 floating-point addition is not associative.
 
-More formally, `results...[j0, ..., jR-1] = convert_or_quantize_or_dequantize(
+More formally, `results...[j0, ..., jR-1] = to_destination_type(
 reduce(input_slices_converted), type(results...)))` where:
 
 * `input_slices = inputs...[j0, ..., :, ..., jR-1]`, where `:` are inserted
   at `dimensions`.
-* `input_slices_converted = convert_or_quantize_or_dequantize(input_slices...,
+* `input_slices_converted = to_destination_type(input_slices...,
   type(func_inputs(body)[:len(func_inputs(body))//2])...)`.
+* `init_values_converted = to_destination_type(init_values...,
+  type(func_inputs(body)[len(func_inputs(body))//2:])...)`.
 * `reduce(input_slices_converted) = exec(schedule)` for some binary tree
   `schedule` where:
   * `exec(node) = body(exec(node.left), exec(node.right))`.
@@ -4096,9 +4093,7 @@ reduce(input_slices_converted), type(results...)))` where:
     `index_space(input_slices_converted)` in the ascending lexicographic order
     of `index`.
   * Interspersed with an implementation-defined amount of
-    `convert_or_quantize_or_dequantize(init_values...,
-    type(func_inputs(body)[len(func_inputs(body))//2:])...)` at
-    implementation-defined positions.
+    `init_values_converted` at implementation-defined positions.
 
 #### Inputs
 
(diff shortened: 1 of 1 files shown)


docs/spec.md
```diff
@@ -4963,8 +4977,12 @@ More formally:
 * (C9) `select` has type `(tensor<E>, tensor<E>) -> tensor<i1>` where
        `E = element_type(operand)`.
 * (C10) `scatter` has type `(tensor<E>, tensor<E>) -> tensor<E>` where
-        `E = element_type(operand)`.
-* (C11) `type(operand) = type(result)`.
+  `is_integer(element_type(operand)) = is_integer(element_type(E))` or
+  `is_float(element_type(operand)) = is_float(element_type(E))` or
+  `is_complex(element_type(operand)) = is_complex(element_type(E))` or
+  `(is_quantized(element_type(operand)) = is_quantized(element_type(E)) and
```
loganchien, line 4983: Same as above.  Replace `=` with `and`.

docs/spec.md
```diff
@@ -4963,8 +4977,12 @@ More formally:
 * (C9) `select` has type `(tensor<E>, tensor<E>) -> tensor<i1>` where
        `E = element_type(operand)`.
 * (C10) `scatter` has type `(tensor<E>, tensor<E>) -> tensor<E>` where
-        `E = element_type(operand)`.
-* (C11) `type(operand) = type(result)`.
+  `is_integer(element_type(operand)) = is_integer(element_type(E))` or
```
loganchien, line 4980: Not sure if you want to define a name for this constraint and share it between 3 occurrences.

docs/spec.md
```diff
@@ -6312,6 +6330,19 @@ function returns `true`. If `x` is not a tensor, returns `None`.
 tensors and returns `num_results` slices of `x` along the axis `axis`.
 If `x` is not a tensor or `dim(x, axis) % num_results != 0`, returns `None`.
 
+* `reduce_implicit_convert(x: Value, source_type: Type, destination_type: Type)
+-> Value` is defined on tensors and returns the converted value of `x` based on
+the `source_type` and `destination_type` as follows:
+
+```python
+def reduce_implicit_convert(x: Value, source_type: Type, destination_type: Type) -> Value:
```
loganchien, line 6338: Maybe rename to `implicit_convert_for_reduce`?

It looks a little weird to see two "reduce" function nested: `reduce_implicit_covert(reduce(...))`.

docs/spec.md
```diff
@@ -4001,49 +4001,58 @@ doesn't hold for many popular reductions. E.g. floating-point addition for
...
-More formally, `results...[j0, ..., jR-1] = reduce(input_slices)` where:
+More formally, `results...[j0, ..., jR-1] = reduce_implicit_convert(reduce(
+input_slices_converted), type(func_outputs(body)...), type(results...)))` where:
 
 * `input_slices = inputs...[j0, ..., :, ..., jR-1]`, where `:` are inserted
   at `dimensions`.
-* `reduce(input_slices) = exec(schedule)` for some binary tree `schedule`
-  where:
+* `input_slices_converted = reduce_implicit_convert(input_slices...,
+  type(inputs...), type(func_inputs(body)...)`.
```
loganchien, line 4010: Only the first half of the `func_inputs(body)`?

docs/spec.md
```diff
@@ -4001,49 +4001,58 @@ doesn't hold for many popular reductions. E.g. floating-point addition for
...
 * `schedule` is an implementation-defined full binary tree whose in-order
   traversal consists of:
-  * `input_slices...[index]` values, for all `index` in
-    `index_space(input_slices)` in the ascending lexicographic order of `index`.
-  * Interspersed with an implementation-defined amount of `init_values`
+  * `input_slices_converted...[index]` values, for all `index` in
+    `index_space(input_slices_converted)` in the ascending lexicographic order
+    of `index`.
+  * Interspersed with an implementation-defined amount of
+    `reduce_implicit_convert(init_values..., type(init_values...), type(func_inputs(body)[len(func_inputs(body)//2)]:)...)`
```
loganchien, line 4021: Should be:

`func_inputs(body)[len(func_inputs(body)//2):])`

(semicolon `:` in the slice operator `[...]`)

loganchien: Actually:

```
func_inputs(body)[len(func_inputs(body))//2:])
```
`len(...)` must not include `//2` either

docs/spec.md
burmako, line 6338: Am I right in thinking that `type(x)` is equal to `source_type` for all calls to `reduce_implicit_convert`? If yes, we probably don't need to pass `source_type` explicitly?

burmako: +1 on further iterating on the name, although my feedback is coming from a bit of a different perspective. 

In my opinion, "reduce_implicit_convert" is a bit too specific to be included in mostly general purpose functions in Notation. E.g. does the name have to include the "reduce" part? There's nothing reduce-specific in this logic, right? What about "implicit"? Is the way how this function is called material for what it does?

burmako (changes requested):

docs/spec.md
```diff
@@ -4001,49 +4001,58 @@ doesn't hold for many popular reductions. E.g. floating-point addition for
...
 * (C3) `0 < size(inputs) = size(init_values) = size(results) = N`.
 * (C4) `0 <= dimensions < rank(inputs[0])`.
 * (C5) `is_unique(dimensions)`.
 * (C6) `body` has type `tensor<E0>, ..., tensor<EN-1>, tensor<E0>, ...,`
   `tensor<EN-1>) -> (tensor<E0>, ..., tensor<EN-1>)` where
-  `Ei = element_type(inputs[i])`.
+  `is_integer(element_type(inputs[i])) = is_integer(element_type(E[i]))` or
+  `is_float(element_type(inputs[i])) = is_float(element_type(E[i]))` or
+  `is_complex(element_type(inputs[i])) = is_complex(element_type(E[i]))` or
+  `(is_quantized(element_type(inputs[i])) = is_quantized(element_type(E[i])) and
```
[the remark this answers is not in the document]
sdasgup3: Agreed. Fixed.

docs/spec.md
sdasgup3: Thanks for pointing point. fixed.

docs/spec.md
sdasgup3: Thanks for pointing this. Yes, we do not need the extra argument.

docs/spec.md
sdasgup3: agreed. fixed.

docs/spec.md
```diff
@@ -4268,16 +4273,17 @@ where:
 * (C10) `size(window_dilations) = rank(inputs[0])`.
 * (C11) `0 < window_dilations`.
 * (C12) `shape(padding) = [rank(inputs[0]), 2]`.
-* (C13) `body` has type `(tensor<E0>, ..., tensor<EN-1>, tensor<E0>, ..., tensor<EN-1>) -> (tensor<E0>, ..., tensor<EN-1>)`
-  where `Ei = element_type(inputs[i])`.
+* (C13) `body` has type `tensor<E0>, ..., tensor<EN-1>, tensor<E0>, ...,`
```
loganchien, line 4276: `(` is missing again

loganchien (approved):

docs/spec.md
sdasgup3: I agree that some components of the name like `reduce` and `implicit` are not relevant here. The function is similar to StableHLO convert op with additional support for quantization. I have revised the name as `convert_with_quantized_type`. Please let me know. 

Note that once https://github.com/openxla/stablehlo/issues/1576 is resolved we can replace the name with convert. Added a note about that in the spec itself.

docs/spec.md
sdasgup3: I have a function on Type with name `same_elementtype_ignoring_bitwidth` for reusing the common. The name, however, is not representative of  how we compare the quantized element types, where we just checks for the equality of expressed_type and allowing the bitwidth, quantized dimension and quantization parameter to differ.

Other alternatives like   `same_elementtype_ignoring_bitwidth_quantized_dimension_and_parameter` is quite mouthful. I am looking for a better alternative and get back on that.

docs/spec.md
```diff
@@ -4001,49 +4001,58 @@ doesn't hold for many popular reductions. E.g. floating-point addition for
...
 * (C4) `0 <= dimensions < rank(inputs[0])`.
 * (C5) `is_unique(dimensions)`.
 * (C6) `body` has type `tensor<E0>, ..., tensor<EN-1>, tensor<E0>, ...,`
   `tensor<EN-1>) -> (tensor<E0>, ..., tensor<EN-1>)` where
-  `Ei = element_type(inputs[i])`.
+  `is_integer(element_type(inputs[i])) = is_integer(element_type(E[i]))` or
+  `is_float(element_type(inputs[i])) = is_float(element_type(E[i]))` or
+  `is_complex(element_type(inputs[i])) = is_complex(element_type(E[i]))` or
+  `(is_quantized(element_type(inputs[i])) = is_quantized(element_type(E[i])) and
+  expressed_type(element_type(inputs[i])) = expressed_type(element_type(E[i])))`.
```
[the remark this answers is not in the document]
sdasgup3: Per https://github.com/openxla/stablehlo/pull/1796#discussion_r1353223442 I am planning to use a function on `Type` for that and looking for a better name for that.

docs/spec.md
sdasgup3: fixed. Thanks!

docs/spec.md
```diff
@@ -4268,16 +4273,17 @@ where:
...
+  `same_elementtype_ignoring_bitwidth(element_type(inputs[i]), element_type(E[i]))`.
 * (C14) `same(shape(results...))`.
 * (C15) `shape(results[0]) = num_windows` where:
   * `dilated_input_shape = shape(inputs[0]) = 0 ? 0 : (shape(inputs[0]) - 1) * base_dilations + 1`.
   * `padded_input_shape = padding[:, 0] + dilated_input_shape + padding[:, 1]`.
   * `dilated_window_shape = (window_dimensions - 1) * window_dilations + 1`.
   * `is_empty_window = padded_input_shape = 0 || dilated_window_shape > padded_input_shape`.
   * `num_windows = is_empty_window ? 0 : floor((padded_input_shape - dilated_window_shape) / window_strides) + 1`.
-* (C16) `element_type(results...) = element_type(init_values...)`.
+* (C16) `baseline_element_type(results...) = baseline_element_type(init_values...)`.
```
loganchien, line 4286: Actually should be:

```
same_elementtype_ignoring_bitwidth(element_type(E[i]), element_type(results[i]))
```

docs/spec.md
```diff
@@ -4001,49 +4001,54 @@ doesn't hold for many popular reductions. E.g. floating-point addition for
...
 * (C3) `0 < size(inputs) = size(init_values) = size(results) = N`.
 * (C4) `0 <= dimensions < rank(inputs[0])`.
 * (C5) `is_unique(dimensions)`.
 * (C6) `body` has type `tensor<E0>, ..., tensor<EN-1>, tensor<E0>, ...,`
   `tensor<EN-1>) -> (tensor<E0>, ..., tensor<EN-1>)` where
-  `Ei = element_type(inputs[i])`.
+  `same_elementtype_ignoring_bitwidth(element_type(inputs[i]), element_type(E[i]))`.
 * (C7) `shape(results...) = shape(inputs...)` except that the dimension
   sizes of `inputs...` corresponding to `dimensions` are not included.
+* (C8) `baseline_element_type(inputs...) = baseline_element_type(results...)`.
```
loganchien, line 4051: Should be:

```
same_elementtype_ignoring_bitwidth(element_type(E[i]), element_type(results[i]))
```

docs/spec.md
```diff
@@ -4963,8 +4969,8 @@ More formally:
 * (C9) `select` has type `(tensor<E>, tensor<E>) -> tensor<i1>` where
        `E = element_type(operand)`.
 * (C10) `scatter` has type `(tensor<E>, tensor<E>) -> tensor<E>` where
-        `E = element_type(operand)`.
-* (C11) `type(operand) = type(result)`.
+  `same_elementtype_ignoring_bitwidth(element_type(operand), element_type(E))`.
+* (C11) `baseline_type(operand) = baseline_type(result)`.
```
loganchien, line 4973:
```
same_elementtype_ignoring_bitwidth(element_type(E), element_type(result))
```

loganchien (approved):

docs/spec.md
```diff
@@ -6312,6 +6331,24 @@ function returns `true`. If `x` is not a tensor, returns `None`.
 tensors and returns `num_results` slices of `x` along the axis `axis`.
 If `x` is not a tensor or `dim(x, axis) % num_results != 0`, returns `None`.
 
+* `convert_with_quantized_type(x: Value, destination_type: Type)
```
burmako, line 6334: Since `destination_type` doesn't have to be quantized, let's think of a more precise name, e.g. `convert_or_quantize`.

docs/spec.md
```diff
@@ -6312,6 +6331,24 @@ function returns `true`. If `x` is not a tensor, returns `None`.
...
+-> Value` is defined on tensors and returns the converted value of `x` based on
+the `type(x)` and `destination_type` as follows:
+
+```python
+def convert_with_quantized_type(x: Value, destination_type: Type) -> Value:
+  if type(x) == destination_type:
+    return x
+  if is_quantized(type(x)) and is_quantized(destination_type):
+    return quantize(x, destination_type)
```
burmako, line 6343: What happens if only one of the types is quantized?

docs/spec.md
burmako, line 6334: Also, how about moving this definition accordingly, so that all the definitions in this section are written in alphabetical order?

docs/spec.md
```diff
@@ -6312,6 +6331,24 @@ function returns `true`. If `x` is not a tensor, returns `None`.
 tensors and returns `num_results` slices of `x` along the axis `axis`.
 If `x` is not a tensor or `dim(x, axis) % num_results != 0`, returns `None`.
 
+* `convert_with_quantized_type(x: Value, destination_type: Type)
+-> Value` is defined on tensors and returns the converted value of `x` based on
```
burmako, line 6335: I don't think the linebreak in the signature is needed anymore.

docs/spec.md
```diff
@@ -6268,6 +6274,19 @@ If `x` is a value or placeholder, this function is a shortcut for
 `member_name(type(x))`.  If `x` is not a type that has an appropriate member, or
 a value or a placeholder of such a type, returns `None`.
 
+* `same_elementtype_ignoring_bitwidth(x: Type, y: Type) -> bool` checks for the
```
burmako, line 6277: Let's align the name of this function with the name of the function that does the conversion (and define semantics in terms of that). E.g. if the latter ends up being called `convert_or_quantize`, then this function could be called `is_convertible_or_quantizable`.

burmako (changes requested):

docs/spec.md
sdasgup3: make sense.

docs/spec.md
sdasgup3: Hi @loganchien 
When `operand`, `E` and `result` are all per-tensor quantized types,
`same_elementtype_ignoring_bitwidth` ensures that only the expressed types are the same, other components (storage_type, quantization params) may vary.
`baseline_element_type` ensures that both storage type and expressed type will be the same,  other components  (quantization params) may vary.

In this PR I am proposing `same_elementtype_ignoring_bitwidth(element_type(operand), element_type(E))` and `baseline_type(operand) = baseline_type(result)` which transitively means the expressed type of `E` and `result` to be the same and (storage type and quantization parameters) may vary, which will be the same constraint   with `same_elementtype_ignoring_bitwidth(element_type(E), element_type(result))`.

However, by using `baseline_type(operand) = baseline_type(result)`  we are adding an extra constraint that the storage type of `operand` and `result` will be the same, which is inline with the RFC proposal. 

Please let me know what you think about this

docs/spec.md
sdasgup3: please refer to my comment https://github.com/openxla/stablehlo/pull/1796#discussion_r1355365362

docs/spec.md
sdasgup3: Thanks @burmako  for the name suggestion!
Also let me capture your comnent @https://github.com/openxla/stablehlo/pull/1796#discussion_r1353673707 here. 

>> What happens if only one of the types is quantized?

I agree that generalizing `convert_or_quantize` to capture other scenarios will help make this function sufficient  on its own and can be reused in other places, if needed.  

>> Let's align the name of this function with the name of the function that does the conversion (and define semantics in terms of that). E.g. if the latter ends up being called convert_or_quantize, then this function could be called is_convertible_or_quantizable.

Per the [RFC](https://github.com/openxla/stablehlo/pull/1664), the criteria for conversion  is limited to `integer -> integer`, `float -> float`, `complex -> complex` and `quantized_type -> quantized_type`. With that `is_convertible_or_quantizable` might not fit as the pre-requisite for the genetic version of `convert_or_quantize`.
 
One option, which I prefer,  is to use generic `convert_or_quantize` and keep the suggested name "is_convertable_or_quantizable" with limited checking criteria and add rationale to state its limited functionality. 

A few other options considered: 
1. Sync the routines corresponding to "check for conversion" and "conversion"
    - (a) Keep both of them limited to specific conversions only. In that case, change the routine names to something like `convert_for_reduction_computations` and `is_convertible_for_reduction_computation`. However, introducing operation specific functions defeats the generality of such meta functions to be re-usable. 
    - (b) Keep both generalized. Per the RFC, this will include additional scenarios of type conversions which we do not have the use-case for.

2. Keep  "check for conversion" routine as special purpose and the "conversion" routine as generic. 
   - With that we can use `convert_or_quantize` as a generic conversion routine and use some special purpose name like  `is_convertible_for_reduction_computation` for "check for conversion" routine. The difference with 1(a): is that it using one (vs two) operations specific function. 

Please let me know your thoughts?

docs/spec.md
```diff
@@ -4001,49 +4016,54 @@ doesn't hold for many popular reductions. E.g. floating-point addition for
...
 #### Constraints
 
 * (C1) `same(shape(inputs...))`.
-* (C2) `element_type(inputs...) = element_type(init_values...) =
-  element_type(results...)`.
+* (C2) `element_type(inputs...) = element_type(init_values...)`.
 * (C3) `0 < size(inputs) = size(init_values) = size(results) = N`.
 * (C4) `0 <= dimensions < rank(inputs[0])`.
 * (C5) `is_unique(dimensions)`.
 * (C6) `body` has type `tensor<E0>, ..., tensor<EN-1>, tensor<E0>, ...,`
```
dominicsymes, line 4061: missing "(" before tensor\<E0\>?

docs/spec.md
```diff
@@ -4154,7 +4174,7 @@ Afterwards, within each `process_group`:
 
 | Label | Name                    | Type                                         | Constraints            |
 |-------|-------------------------|----------------------------------------------|------------------------|
-| (I1)  | `operand`               | tensor                                       | (C1), (C2), (C7), (C8) |
+| (I1)  | `operand`               | tensor or quantized tensor                   | (C1), (C2), (C7), (C8) |
```
dominicsymes, line 4177: Similar to line 792, should "quantized tensor" be per-tensor quantized tensor or per-axis constraints be added?

docs/spec.md
```diff
@@ -6268,6 +6294,19 @@ If `x` is a value or placeholder, this function is a shortcut for
...
+* `is_convertible_or_quantizable(x: Type, y: Type) -> bool` checks for the
+equality of `x` and `y`, ignoring the bitwidth, when they are of type
+`TensorElementType`. When `x` and `y` are `QuantizedTensorElementType`s,
+the function checks for the equality of `QuantizationExpressedType` component.
+
+```python
+def is_convertible_or_quantizable(x: Type, y: Type) -> Value:
+  return is_integer(x) = is_integer(y) or
+    is_float(x) = is_float(y) or
+    is_complex(x) = is_complex(y) or
```
doyeonkim0, line 6306: Shouldn't we have 'and' instead of = here?

docs/spec.md
```diff
@@ -6285,6 +6324,33 @@ and [slicing](https://docs.python.org/3/reference/expressions.html#slicings)
 notations from Python are available to index into tensors, quantized tensors
 and tuples.
 
+* `convert_or_quantize(x: Value, destination_type: Type) -> Value` is defined on
+tensors and returns the converted value of `x` based on the `type(x)` and
+`destination_type` as follows:
+
+```python
+def convert_or_quantize(x: Value, destination_type: Type) -> Value:
```
doyeonkim0, line 6332: For me, 'convert_or_quantize' sounds like it will convert or quantize, and never dequantize, but it actually does dequantize. I'm not sure if 'convert_or_quantize_or_dequantize' is too verbose, but can we somoehow reflect the dequantizing behavior on its name?

docs/spec.md
```diff
@@ -6268,6 +6294,19 @@ If `x` is a value or placeholder, this function is a shortcut for
...
+equality of `x` and `y`, ignoring the bitwidth, when they are of type
+`TensorElementType`. When `x` and `y` are `QuantizedTensorElementType`s,
+the function checks for the equality of `QuantizationExpressedType` component.
+
+```python
+def is_convertible_or_quantizable(x: Type, y: Type) -> Value:
+  return is_integer(x) = is_integer(y) or
+    is_float(x) = is_float(y) or
+    is_complex(x) = is_complex(y) or
+    (is_quantized(x) and is_quantized(y) and expressed_type(x) = expressed_type(y))
```
doyeonkim0, line 6307: Shouldn't we also have `(is_quantized(x) and expressed_type(type(x))=type(y)) or (is_quantized(y) and expressed_type(type(y))=type(x))`?

docs/spec.md
```diff
@@ -6268,6 +6294,19 @@ If `x` is a value or placeholder, this function is a shortcut for
 `member_name(type(x))`.  If `x` is not a type that has an appropriate member, or
 a value or a placeholder of such a type, returns `None`.
 
+* `is_convertible_or_quantizable(x: Type, y: Type) -> bool` checks for the
+equality of `x` and `y`, ignoring the bitwidth, when they are of type
```
doyeonkim0, line 6298: I think we should have something more than equality. eg. compatibility for quantization/dequantization

docs/spec.md
```diff
@@ -778,28 +778,28 @@ defined as follows:
...
 * `schedule` is an implementation-defined binary tree whose in-order
-  traversal is `operands@process_group...[result_index]`.
+  traversal is `convert_or_quantize(operands@process_group...[result_index], type(func_inputs(computation)[0]))`.
 
 #### Inputs
 
 | Label | Name                    | Type                                                             | Constraints |
 |-------|-------------------------|------------------------------------------------------------------|-------------|
-| (I1)  | `operand`               | tensor                                                           | (C5), (C6)  |
+| (I1)  | `operand`               | tensor or quantized tensor                                       | (C5), (C6)  |
```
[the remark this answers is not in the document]
sdasgup3: Thanks for the catch! It was unintended and now fixed.

docs/spec.md
sdasgup3: Thanks for the catch! It was unintended and now fixed.

docs/spec.md
sdasgup3: That is right. Fixed.

docs/spec.md

docs/spec.md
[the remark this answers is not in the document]
sdasgup3: >> The reduction will be more complex if different scale is used for the left and right elements being applied to the computation function

I agree that this will not be straight-forward. The individual operations within the computation blocks are supposed to handle per-axis computation. However, we do not have per-axis support for the entire opset.

docs/spec.md
sdasgup3: Per the [RFC](https://github.com/openxla/stablehlo/pull/1664), the criteria for conversion  is limited to `integer -> integer`, `float -> float`, `complex -> complex` and `quantized_type -> quantized_type`. Hence the limited behavior.

docs/spec.md
sdasgup3: I agree. Let me get back on that.

docs/spec.md
sdasgup3: Hi @doyeonkim0 
I agree that this is a restrictive version of equality and also the name `is_convertible_quantizable` might cause confusion as it as not considering all possible scenarios.  How about we use `is_promotable` and add justification in the function definition about the rationale of using it.

docs/spec.md
sdasgup3: I just revised the name of the function and add justification to its specific behavior. PTAL.

docs/spec.md
sdasgup3: Update:
The current PR is structured around (2),
Suggested name for "check for conversion" routine: `is_promotable` and while defining this function the spec text provided rationale for its specific behavior.  
Suggested name for "conversion" routine: `convert__or_quantize_or_dequantize`.

docs/spec.md
doyeonkim0: Sounds good to me!

doyeonkim0 (approved):

dominicsymes (approved):

docs/spec.md
loganchien: >  by using baseline_type(operand) = baseline_type(result) we are adding an extra constraint that the storage type of operand and result will be the same

I don't think this extra constraint is desired nor inline with the RFC proposal.

One important use case for RFC is to read quantized type in ui8 storage type and then return a larger summation in i32.  This "extra" constraint defeat the purpose of the RFC.

paulinesho (approved):

docs/spec.md
sdasgup3: Thanks for the feedback! Such a case is now handled.

paulinesho (approved):

docs/spec.md
sdasgup3: @loganchien 
updated the PR with the change.  

>> same_elementtype_ignoring_bitwidth(element_type(E), element_type(result))

Currently using
```
is_promotable(element_type(operand), element_type(result))
```

As `E` is not defined in the current constraint. However, transitively it would mean the same. Please let me know your opinion.

docs/spec.md
sdasgup3: updated according. Please refer to my comment below.

loganchien (approved):

GleasonK (changes requested):

GleasonK:
The semantics LGTM. Have a few suggestions to make the spec more readable.

docs/spec.md
```diff
@@ -6285,6 +6336,33 @@ and [slicing](https://docs.python.org/3/reference/expressions.html#slicings)
...
+    return quantize(x, destination_type)
+
+  if is_quantized(type(x)):
+    assert destination_type = expressed_type(type(x))
+    return dequantize(type(x))
+
+  return convert(x, destination_type)
+```
+
+There is plan to merge `convert`, `uniform_quantize` and `uniform_dequantize`
```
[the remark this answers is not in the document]
sdasgup3: updated the text.

docs/spec.md
```diff
@@ -1062,31 +1063,40 @@ def batch_norm_grad(operand, scale, mean, variance, grad_output, epsilon, featur
...
+| `grad_operand` | tensor of floating-point type or per-tensor quantized tensor        | (C2), (C3)  |
+| `grad_scale`   | 1-dimensional tensor of floating-point or per-tensor quantized type | (C2), (C4)  |
+| `grad_offset`  | 1-dimensional tensor of floating-point or per-tensor quantized type | (C2), (C4)  |
 
 #### Constraints
 
 * (C1) `0 <= feature_index < rank(operand)`.
-* (C2) `operand`, `scale`, `mean`, `variance`, `grad_output`, `grad_operand`
-       `grad_scale` and `grad_offset` have the same element type.
+* (C2) `baseline_element_type(operand) = baseline_element_type(scale) =
```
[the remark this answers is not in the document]
sdasgup3: That is a good point. I applied the same to the constraints of batch_norm_* operations which are quite verbose. In other places the readability is fine.

docs/spec.md
```diff
@@ -4001,49 +4017,55 @@ doesn't hold for many popular reductions. E.g. floating-point addition for
...
-  * `input_slices...[index]` values, for all `index` in
-    `index_space(input_slices)` in the ascending lexicographic order of `index`.
-  * Interspersed with an implementation-defined amount of `init_values`
-    at implementation-defined positions.
+  * `input_slices_converted...[index]` values, for all `index` in
+    `index_space(input_slices_converted)` in the ascending lexicographic order
+    of `index`.
+  * Interspersed with an implementation-defined amount of
+    `convert_or_quantize_or_dequantize(init_values...,
+    type(func_inputs(body)[len(func_inputs(body))//2:])...)` at
```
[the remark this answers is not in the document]
sdasgup3: done.

docs/spec.md
```diff
@@ -4736,10 +4759,17 @@ Given that, `results = exec(schedule, inputs)`, where:
   `index_space(updates[0])`.
 * `exec([update_index, ...], results) = exec([...], updated_results)` where:
   * If `result_index` is in bounds for `shape(results...)`
-    * `updated_values =
-      update_computation(results...[result_index], updates...[update_index])`.
+    * `updated_values = update_computation(
+      convert_or_quantize_or_dequantize(results...[result_index], type(
+      func_inputs(update_computation)[:len(func_inputs(update_computation))//2])
```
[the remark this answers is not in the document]
sdasgup3: done.

docs/spec.md
```diff
@@ -6285,6 +6336,33 @@ and [slicing](https://docs.python.org/3/reference/expressions.html#slicings)
 notations from Python are available to index into tensors, quantized tensors
 and tuples.
 
+* `convert_or_quantize_or_dequantize(x: Value, destination_type: Type) -> Value`
+is defined on tensors and returns the converted value of `x` based on the
+`type(x)` and `destination_type` as follows:
+
+```python
+def convert_or_quantize_or_dequantize(x: Value, destination_type: Type) -> Value:
```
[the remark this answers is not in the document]
sdasgup3: Agreed with your point. Thanks for the suggestions!
Let me propose picking  `to_destination_type`. To me it seems a good generic and concise choice without geting tied much into its implementation. `convert_any` looks fine too but I am biased towards the former. 

PTAL.

docs/spec.md
```diff
@@ -6247,6 +6277,27 @@ for `is_quantized(x) and quantization_dimension(x) is not None`.
...
 shortcut for `is_quantized(x) and quantization_dimension(x) is None`.
 
+* `is_promotable(x: Type, y: Type) -> bool` checks if type `x` can be promoted
+to type `y`.  When `x` and `y` are `QuantizedTensorElementType`s, the promotion
+is applied only to the `storage_type`. This specific version of promotion is
+currently used in context of reduction computation (refer to
+[RFC](https://github.com/openxla/stablehlo/pull/1664) for more details).
+
+```python
+def is_promotable(x: Type, y: Type) -> Value:
```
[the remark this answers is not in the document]
sdasgup3: >> Does "promotable" have any implications on bitwidth?

In the current form there is no implication and it does allow conversion in both ways (i32 -> i16 and i16 -> i32). 

Per my offline chat with @loganchien, the output type can be anything to which the accumulation result needs to be converted. So all of `f32->f32` (no conversion), `f32 -> f16`, `f16 -> f32` are allowed.


>> I believe the existing rules permit this, not sure if we want this.
There does not seem like a current use case with the demotion without overflowing the accumulation result (confirmed with @loganchien). IMO,  In its current form it provides generality to the user get the output in any precision they want.

With that I can think of two options:
1. Keep the generality as is. With that, I agree that `is_promotable` might be ambiguous. How about we use something like `is_implicitly_convertable`. There I would like to quality `_convertable` somehow to highlight that this is not a general conversion, but for a specific use-case. What do you think?

2. Add additional check to allow `bitwidth(source) <= bitwidth(destination)` 

I am biased towards the general approach with less constraints, but I would like to know your opinion.

GleasonK (commented):
LGTM. Let's resolve the last open comment and then I'll approve. Will reach out offline to discuss further.

docs/spec.md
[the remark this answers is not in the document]
GleasonK: I like generic _if there are use cases for being generic_. Rationale: Everything we support in the IR has implications for what backends must support.

With that said - Do we think it is trivial overhead for backends to support demotion if we are already supporting promotion? If so then probably OK to be generic. But if it would require more specialization for each of these ops, then perhaps we reconsider and restrict to use cases.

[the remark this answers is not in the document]
GleasonK: Whoops, I misplaced this comment. Moving it to the other open thread and resolving this one.

docs/spec.md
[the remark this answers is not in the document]
GleasonK: I like general if there are use cases for being general. Rationale: Everything we support in the IR has implications for what backends must support.

With that said - Do we think it is trivial overhead for backends to support demotion if we are already supporting promotion? If so then probably OK to be generic. But if it would require more specialization for each of these ops, then perhaps we reconsider and restrict to use cases.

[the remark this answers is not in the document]
sdasgup3: Hi Kevin, thanks for the feedback!

The specific question bought up in the above discussion is: should we promote demotion as well? It seems that there aren't any use-cases for that. For example, consider the following cases of implicit conversion for reduction operations. The examples are shown using integer type but can be generalized to other types as well.
 
| input element type | accumulator element type  | result element type | use-cases |
|:--:|:--:|:--:|:--|
| i8| i8 |i8| Reduction using min/max|
| i8| i32 |i32| Reduction using average/sum|
| i8| i32| i8 | Demotion from accumulation type to result type. (No use-case)|
| i8| i32| i64 | Extra promotion from accumulation type to result type. (No use-case) |
| i32 | i8 | i8 | Demotion from input value type to accumulation type. (No use case) |

With that in mind, we propose to develop the specification based on the scenario we have the use-case for (just for promotion). 
Thanks @loganchien for you input on this.  

@loganchien @GleasonK @paulinesho If the above proposal looks good to you, can you please review the PR. The only changes from the last time are
1. The definition of `is_promotable` now only support the promotion cases with `bit-width(element_type(input)) <= bit-width(element_type(accumulator))`.
2. For all the ops supported in the PR: `element_type(result) == element_type(accumulator)`.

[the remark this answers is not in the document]
loganchien: SGTM.

loganchien (approved):

paulinesho (approved):

docs/spec.md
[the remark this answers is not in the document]
doyeonkim0: LGTM

GleasonK (approved):

PR #1797, opened 2023-10-09
Introduce StableHLO interpreter API

penagos:
Today, it isn't practical to call the StableHLO reference interpreter. Invoking the interpreter requires going through the MLIR translate infrastructure (i.e. via `interpretRegistration` in `StablehloTranslateMain.cpp`), and does not currently support specifying user inputs or doing any meaningful post processing on the model outputs short of printing them to screen. This makes it difficult to build tooling which uses the reference interpreter.

This PR proposes a new reference interpreter API which facilitates the running the StableHLO reference interpreter through a simple, unified interface. At a high level, the proposed interface allows clients to specify a set of inputs (in `InterpreterValue` format), and an `InterpreterConfiguration` object (see below). The proposed interface invokes the reference interpreter and returns a set of outputs (in `InterpreterValue` format). The new interface would live in `stablehlo/reference/InterpreterApi.h` and would expose the following interface:

```
/// Invoke the StableHLO reference interpreter with the given parsed MLIR
/// module input and provided inputs. Returns a list of interpreter outputs.
/// Can optionally pass a fallback interpreter callback which executes when no
/// builtin kernels are matched.
llvm::ErrorOr<SmallVector<InterpreterValue>> evalModule(
    ModuleOp module, ArrayRef<InterpreterValue> inputs,
    const InterpreterConfiguration &config);
```

### Interpreter configuration
As part of this PR, there is an `InterpreterConfiguration` struct which can be used to configure the StableHLO interpreter. Encapsulating configuration state into a singular object allows the interface to remain stable as new configuration options are added. Initially, the configuration object will contain a handful of options:

```
struct InterpreterConfiguration {
  InterpreterConfiguration()
      : fallback(std::make_unique<InterpreterFallback>()) {}

  /// If specified, the directory to which StableHLO interpreter tensors will
  /// be serialized to disk.
  std::string probeInstrumentationDir = "";

  /// Use the specified named function as the main entrypoint into a module.
  /// Defaults to `main` for modules with multiple functions. If a module only
  /// contains 1 function and the default `main` value is used, the singular
  /// function will be used as the entrypoint (irrespective of a function name
  /// match).
  std::string mainFunction = "main";

  /// If specified, use the callback to run on ops which do not have a
  /// registered kernel.
  std::unique_ptr<InterpreterFallback> fallback;
};
```

Note that all of these configuration values are optional, and it is possible to run the interpreter with a default initialized `InterpreterConfiguration` struct (which will by default, not provide a fallback callback).

### Interpreter fallback callback
When defining an interpreter fallback callback, the `InterpreterFallback` struct can be extended:

```
class InterpreterFallback {
 public:
  /// Custom op kernels for any user specified ops not found in the StableHLO
  /// op dialect or StableHLO interpreter dialect.
  virtual llvm::Error operator()(Operation &op, Scope &scope, Process *process);

  virtual ~InterpreterFallback() = default;
};
```
Note that the user specified fallback need not contain any `InterpreterOp` kernels, these will always run before the user specified callback.

Currently, the proposed change exposes `InterpreterConfiguration` and the current `mlir::func::FuncOp` being evaluated to the fallback interface to support the existing flow. Consequently, there’s a bit of gymnastics in `stablehlo::runInterpreter` which I think could be improved on. I'm open to ideas on how we could better handle this.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -393,6 +393,25 @@ cc_library(
 )
 
 cc_library(
+    name = "reference_interpreter_api",
+    srcs = [
+        "stablehlo/reference/InterpreterApi.cpp",
+    ],
+    hdrs = [
+        "stablehlo/reference/InterpreterApi.h",
+    ],
+    strip_include_prefix = ".",
+    deps = [
+        ":reference_errors",
+        ":reference_process",
+        ":reference_scope",
+        ":reference_tensor",
+        "@llvm-project//llvm:Support",
+        "@llvm-project//mlir:IR",
+    ],
+)
+
+cc_library(
     name = "reference_interpretervalue",
     srcs = [
         "stablehlo/reference/InterpreterValue.cpp",
@@ -921,6 +940,7 @@ cc_binary(
     deps = [
         ":interpreter_ops",
         ":reference_errors",
+        ":reference_interpreter_api",
         ":reference_interpretervalue",
         ":reference_ops",
         ":reference_process_grid",
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -367,6 +367,9 @@ cc_library(
 
 cc_library(
     name = "reference_errors",
+    srcs = [
+        "stablehlo/reference/Errors.cpp",
+    ],
     hdrs = [
         "stablehlo/reference/Errors.h",
     ],
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -402,6 +402,7 @@ cc_library(
     ],
     strip_include_prefix = ".",
     deps = [
+        ":interpreter_ops",
         ":reference_errors",
         ":reference_ops",
         ":reference_process",
@@ -422,7 +423,6 @@ cc_library(
     ],
     strip_include_prefix = ".",
     deps = [
-        ":interpreter_ops",
         ":reference_errors",
         ":reference_tensor",
         ":reference_token",
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -407,13 +407,17 @@ cc_library(
     deps = [
         ":interpreter_ops",
         ":reference_errors",
+        ":reference_interpretervalue",
+        ":reference_numpy",
         ":reference_ops",
         ":reference_process",
         ":reference_scope",
         ":reference_tensor",
         ":register",
         "@llvm-project//llvm:Support",
+        "@llvm-project//mlir:FuncDialect",
         "@llvm-project//mlir:IR",
+        "@llvm-project//mlir:Support",
     ],
 )
 
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -411,6 +411,7 @@ cc_library(
         ":reference_process",
         ":reference_scope",
         ":reference_tensor",
+        ":register",
         "@llvm-project//llvm:Support",
         "@llvm-project//mlir:IR",
     ],
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -422,6 +422,7 @@ cc_library(
     ],
     strip_include_prefix = ".",
     deps = [
+        ":interpreter_ops",
         ":reference_errors",
         ":reference_tensor",
         ":reference_token",
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -412,8 +412,6 @@ cc_library(
         ":reference_ops",
         ":reference_process",
         ":reference_scope",
-        ":reference_tensor",
-        ":register",
         "@llvm-project//llvm:Support",
         "@llvm-project//mlir:FuncDialect",
         "@llvm-project//mlir:IR",
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -403,6 +403,7 @@ cc_library(
     strip_include_prefix = ".",
     deps = [
         ":reference_errors",
+        ":reference_ops",
         ":reference_process",
         ":reference_scope",
         ":reference_tensor",
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -407,6 +407,7 @@ cc_library(
     deps = [
         ":interpreter_ops",
         ":reference_errors",
+        ":reference_interpreterconfiguration",
         ":reference_interpretervalue",
         ":reference_numpy",
         ":reference_ops",
@@ -416,6 +417,23 @@ cc_library(
         "@llvm-project//mlir:FuncDialect",
         "@llvm-project//mlir:IR",
         "@llvm-project//mlir:Support",
+    ],
+)
+
+cc_library(
+    name = "reference_interpreterconfiguration",
+    srcs = [
+        "stablehlo/reference/InterpreterConfiguration.cpp",
+    ],
+    hdrs = [
+        "stablehlo/reference/InterpreterConfiguration.h",
+    ],
+    strip_include_prefix = ".",
+    deps = [
+        ":reference_errors",
+        ":reference_process",
+        ":reference_scope",
+        "@llvm-project//llvm:Support",
     ],
 )
 
@@ -469,6 +487,7 @@ cc_library(
         ":reference_element",
         ":reference_errors",
         ":reference_index",
+        ":reference_interpreterconfiguration",
         ":reference_interpretervalue",
         ":reference_process",
         ":reference_process_grid",
--- a/stablehlo/reference/InterpreterApi.cpp
+++ b/stablehlo/reference/InterpreterApi.cpp
@@ -20,7 +20,6 @@ limitations under the License.
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/Path.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
-#include "mlir/Parser/Parser.h"
 #include "mlir/Support/DebugStringHelper.h"
 #include "stablehlo/reference/Errors.h"
 #include "stablehlo/reference/InterpreterOps.h"
--- a/stablehlo/reference/InterpreterApi.cpp
+++ b/stablehlo/reference/InterpreterApi.cpp
@@ -1,0 +1,215 @@
+/* Copyright 2023 The StableHLO Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#include "stablehlo/reference/InterpreterApi.h"
+
+#include "llvm/ADT/StringRef.h"
+#include "llvm/Support/Errc.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/MemoryBuffer.h"
+#include "llvm/Support/Path.h"
+#include "llvm/Support/SourceMgr.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/IR/MLIRContext.h"
+#include "mlir/Parser/Parser.h"
+#include "mlir/Support/DebugStringHelper.h"
+#include "stablehlo/reference/Errors.h"
+#include "stablehlo/reference/InterpreterOps.h"
+#include "stablehlo/reference/NumPy.h"
+#include "stablehlo/reference/Ops.h"
+#include "stablehlo/tests/CheckOps.h"
+
+namespace mlir {
+namespace stablehlo {
+namespace {
+
+llvm::Error wrapStatus(llvm::Error status, llvm::StringRef funcName,
+                       llvm::StringRef fallbackName) {
+  if (status)
+    return stablehlo::invalidArgument(
+        "Error evaluating function: %s. \n\tFallback for %s failed: %s",
+        funcName.data(), fallbackName.data(),
+        toString(std::move(status)).c_str());
+  return llvm::Error::success();
+}
+
+}  // namespace
+
+llvm::Error DefaultInterpreterFallback::operator()(Operation &op,
+                                                   Process *process,
+                                                   Scope &scope) {
+  llvm::StringRef funcName = currentFcn.getSymName();
+  if (auto customCall = dyn_cast<stablehlo::CustomCallOp>(op)) {
+    if (customCall.getCallTargetName() == "check.eq") {
+      auto status = check::evalCustomCallCheckEq(customCall, scope);
+      return wrapStatus(std::move(status), funcName,
+                        "stablehlo.custom_call(@check.eq)");
+    }
+
+    return stablehlo::invalidArgument("Unsupported custom call: %s",
+                                      debugString(op).c_str());
+  }
+
+  if (auto expectAlmostEqOp =
+          dyn_cast<stablehlo::check::ExpectAlmostEqOp>(op)) {
+    auto runtimeLhs = scope.findTensor(expectAlmostEqOp.getLhs());
+    auto runtimeRhs = scope.findTensor(expectAlmostEqOp.getRhs());
+    auto status =
+        stablehlo::check::evalExpectAlmostEqOp(runtimeLhs, runtimeRhs);
+    return wrapStatus(std::move(status), funcName, "check.expect_almost_eq");
+  }
+
+  if (auto expectAlmostEqConstOp =
+          dyn_cast<stablehlo::check::ExpectAlmostEqConstOp>(op)) {
+    auto runtimeOperand = scope.findTensor(expectAlmostEqConstOp.getLhs());
+    auto status = stablehlo::check::evalExpectAlmostEqConstOp(
+        runtimeOperand, expectAlmostEqConstOp.getValue());
+    return wrapStatus(std::move(status), funcName,
+                      "check.expect_almost_eq_const");
+  }
+
+  if (auto expectEqOp = dyn_cast<stablehlo::check::ExpectEqOp>(op)) {
+    auto runtimeLhs = scope.findTensor(expectEqOp.getLhs());
+    auto runtimeRhs = scope.findTensor(expectEqOp.getRhs());
+    auto status = stablehlo::check::evalExpectEqOp(runtimeLhs, runtimeRhs);
+    return wrapStatus(std::move(status), funcName, "check.expect_eq");
+  }
+
+  if (auto expectEqConstOp = dyn_cast<stablehlo::check::ExpectEqConstOp>(op)) {
+    auto runtimeOperand = scope.findTensor(expectEqConstOp.getLhs());
+    auto status = stablehlo::check::evalExpectEqConstOp(
+        runtimeOperand, expectEqConstOp.getValue());
+    return wrapStatus(std::move(status), funcName, "check.expect_eq_const");
+  }
+
+  if (auto expectSerializedEqOp =
+          dyn_cast<stablehlo::check::ExpectSerializedEqOp>(op)) {
+    auto runtimeOperand = scope.findTensor(expectSerializedEqOp.getExpected());
+    auto status = stablehlo::check::evalExpectSerializedEqOp(
+        runtimeOperand, expectSerializedEqOp.getProbeId(),
+        config->probeInstrumentationDir, expectSerializedEqOp.getIteration());
+    return wrapStatus(std::move(status), funcName,
+                      "check.expect_serialized_eq");
+  }
+
+  if (auto probeOp = dyn_cast<stablehlo::interpreter::ProbeOp>(op)) {
+    auto input =
+        stablehlo::InterpreterValue(scope.findTensor(probeOp.getOperand()));
+    auto status = stablehlo::interpreter::evalProbeOp(
+        input, probeOp.getProbeId(), config->probeInstrumentationDir,
+        instrumentedTensors);
+    scope.add(probeOp.getResult(), input);
+    return wrapStatus(std::move(status), funcName, "interpreter.probe");
+  }
+
+  if (auto runParallelOp =
+          dyn_cast<stablehlo::interpreter::RunParallelOp>(op)) {
+    auto runtimeOperands = scope.find(runParallelOp.getInputs());
+    std::queue<StringAttr> infeed;
+    if (auto infeedAttr = runParallelOp.getInfeed())
+      for (auto &value : infeedAttr->getValue())
+        infeed.push(value.cast<FlatSymbolRefAttr>().getAttr());
+
+    SmallVector<SmallVector<StringAttr>> programs(
+        runParallelOp.getPrograms().size());
+    for (auto [i, replica] : llvm::enumerate(runParallelOp.getPrograms()))
+      for (auto &program : replica.cast<ArrayAttr>())
+        programs[i].push_back(program.cast<FlatSymbolRefAttr>().getAttr());
+
+    SymbolTable symbolTable{op.getParentOfType<ModuleOp>()};
+    auto results = stablehlo::interpreter::evalRunParallelOp(
+        runtimeOperands, infeed, programs, symbolTable);
+    scope.add(runParallelOp.getResults(), results);
+    return wrapStatus(llvm::Error::success(), funcName,
+                      "interpreter.run_parallel");
+  }
+
+  return stablehlo::invalidArgument("Unsupported op: %s",
+                                    debugString(op).c_str());
+}
+
+llvm::ErrorOr<SmallVector<InterpreterValue>> runInterpreter(
+    const std::string &mlir, ArrayRef<InterpreterValue> inputs,
+    const InterpreterConfiguration &config) {
+  std::unique_ptr<llvm::MemoryBuffer> program_buffer =
+      llvm::MemoryBuffer::getMemBuffer(mlir);
+
+  llvm::SourceMgr source_mgr;
+  MLIRContext context;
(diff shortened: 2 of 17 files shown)


stablehlo/reference/InterpreterApi.cpp
```diff
@@ -0,0 +1,215 @@
...
+#include "llvm/Support/SourceMgr.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/IR/MLIRContext.h"
+#include "mlir/Parser/Parser.h"
+#include "mlir/Support/DebugStringHelper.h"
+#include "stablehlo/reference/Errors.h"
+#include "stablehlo/reference/InterpreterOps.h"
+#include "stablehlo/reference/NumPy.h"
+#include "stablehlo/reference/Ops.h"
+#include "stablehlo/tests/CheckOps.h"
```
GleasonK, line 32: I don't think CheckOps should ship with the default implementation.

I believe the default fallback should just be "InterpreterOps", and should allow subclassing to support InterpreterOps + their own dialect ops / custom call ops.

Could have methods for:
- `handleInterpreterOp` (this should be marked final, not overridable)
- `handleUnregisteredOp` (check dialect could be added here)
- `handleCustomCallOp` (any additional handling)

Or could be a map of `op_name -> fallbackHandler`, then users could add their own `addCheckDialectFallbacks(handler)` methods to populate the fallback handler class.

But for now I think `CheckOps` should remain a feature of `StablehloTranslateMain`, as exposing them to the wild has broader maintenance implications.

stablehlo/tests/CheckOps.cpp
```diff
@@ -128,6 +136,29 @@ llvm::Error evalExpectSerializedEqOp(const Tensor &expected, StringRef probeId,
   return evalExpectEqOp(expected, *tensorOrError);
 }
 
+llvm::Error evalCustomCallCheckEq(stablehlo::CustomCallOp op,
```
GleasonK, line 139: Similarly, let's keep this custom call handling in `StablehloTranslateMain` as it was. The custom call is a bit of a hack for `testdata` testing, it shouldn't be used more broadly than our tests. A better solution would be adding a pass that converts custom_call -> check dialect, but no need to do it now

GleasonK (changes requested):
Proposed refactoring:

```
InterpreterApi.h/cpp (_only_ handles StableHLO, no check or custom call)
StablehloTranslateMain.cpp (implements InterpreterFallback with Check and custom call)
```

If we need to expose CheckDialect in the future, we can, but for now lets consider that an implementation detail of StableHLO test infrastructure. Non-test code depending on test code is a bit of an antipattern

stablehlo/reference/InterpreterApi.cpp
penagos: Hm, how about just implementing the `InterpreterOp` in the default fallback functor in lieu of 3 separate callbacks? And then leaving the `CheckOp` kernels within the impl of `StablehloTranslateMain`?

stablehlo/reference/InterpreterApi.h
```diff
@@ -0,0 +1,96 @@
...
+struct InterpreterFallback {
+  llvm::Error operator()(Operation &op, Process *process, Scope &scope);
+
+  virtual ~InterpreterFallback() = default;
+
+  /// Set the user provided interpreter configuration.
+  void setConfig(const InterpreterConfiguration &cfg) { config = &cfg; }
+
+  /// Set the topmost currently executing function in the module.
+  void setFcn(func::FuncOp fcn) { currentFcn = fcn; }
```
ghpvnist, line 42: Proposal:
```
void setFunction(func::FuncOp funcOp) { currentFunction = funcOp; }
```

Abbreviating this does not give us much mileage, plus the code in the implementation
```
config.fallback->setFcn(funcOp);
```
would read nicely:
```
config.fallback->setFunction(funcOp);
```

stablehlo/reference/InterpreterApi.h
```diff
@@ -0,0 +1,96 @@
...
+/// Base interpreter fallback callback functor to run when no registered kernels
+/// are found for a given StableHLO operation.
+struct InterpreterConfiguration;
+struct InterpreterFallback {
+  llvm::Error operator()(Operation &op, Process *process, Scope &scope);
+
+  virtual ~InterpreterFallback() = default;
+
+  /// Set the user provided interpreter configuration.
+  void setConfig(const InterpreterConfiguration &cfg) { config = &cfg; }
```
ghpvnist, line 39: `cfg` -> `config` to be consistent with name defined in struct and arg name used in `runInterpreter`

stablehlo/reference/InterpreterApi.h
```diff
@@ -0,0 +1,96 @@
...
+
+  /// The topmost function currently being executed in the module.
+  func::FuncOp currentFcn;
+
+ private:
+  /// Counts how many times a given probe_id has been used while profiling.
+  llvm::StringMap<int32_t> instrumentedTensors;
+};
+
+struct InterpreterConfiguration {
```
ghpvnist, line 60: Let's alphabetically order them for ease of maintenance.

stablehlo/reference/InterpreterApi.cpp
```diff
@@ -0,0 +1,166 @@
...
+  return handleOp(op, process, scope);
+}
+
+llvm::Error InterpreterFallback::handleOp(Operation &op, Process *process,
+                                          Scope &scope) {
+  return stablehlo::invalidArgument("Unsupported op: %s",
+                                    debugString(op).c_str());
+}
+
+llvm::ErrorOr<SmallVector<InterpreterValue>> runInterpreter(
```
ghpvnist, line 94:
```
llvm::ErrorOr<SmallVector<InterpreterValue>> runInterpreter(
    const std::string &mlir, ArrayRef<InterpreterValue> inputs,
    const InterpreterConfiguration &config) {
  llvm::SourceMgr source_mgr;
  source_mgr.AddNewSourceBuffer(llvm::MemoryBuffer::getMemBuffer(mlir),
                                llvm::SMLoc());

  MLIRContext context;
  OwningOpRef<ModuleOp> module(parseSourceFile<ModuleOp>(source_mgr, &context));

  return runInterpreter(module.get(), inputs, config);
}
```

stablehlo/reference/InterpreterApi.cpp
```diff
@@ -0,0 +1,166 @@
...
+        runtimeOperands, infeed, programs, symbolTable);
+    scope.add(runParallelOp.getResults(), results);
+    return wrapStatus(llvm::Error::success(), funcName,
+                      "interpreter.run_parallel");
+  }
+
+  return handleOp(op, process, scope);
+}
+
+llvm::Error InterpreterFallback::handleOp(Operation &op, Process *process,
```
ghpvnist, line 88: I don't think we need this abstracted out yet. Since we don't handle them in multiple ways, let's keep the error message inlined?

stablehlo/reference/InterpreterApi.h
```diff
@@ -0,0 +1,96 @@
...
+#include "stablehlo/reference/Process.h"
+#include "stablehlo/reference/Scope.h"
+
+namespace mlir {
+namespace stablehlo {
+
+/// Base interpreter fallback callback functor to run when no registered kernels
+/// are found for a given StableHLO operation.
+struct InterpreterConfiguration;
+struct InterpreterFallback {
```
ghpvnist, line 33: Let's convert this to a class per https://stackoverflow.com/questions/54585/when-should-you-use-a-class-vs-a-struct-in-c

ghpvnist (changes requested):
Went through first pass. I need a bit more time to think about the `InterpreterFallback` API.

stablehlo/tools/StablehloTranslateMain.cpp
```diff
@@ -15,29 +15,21 @@ limitations under the License.
...
 #include "mlir/Support/LogicalResult.h"
 #include "mlir/Tools/mlir-translate/MlirTranslateMain.h"
 #include "mlir/Tools/mlir-translate/Translation.h"
 #include "mlir/Transforms/Passes.h"
 #include "stablehlo/dialect/Register.h"
 #include "stablehlo/dialect/Serialization.h"
 #include "stablehlo/dialect/StablehloOps.h"
-#include "stablehlo/dialect/VhloOps.h"
-#include "stablehlo/reference/Errors.h"
+#include "stablehlo/dialect/Version.h"
```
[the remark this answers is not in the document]
penagos: This include is needed by:
```
 targetVersion = vhlo::Version::getCurrentVersion().toString();
```

stablehlo/reference/InterpreterApi.cpp
penagos: The initial idea behind splitting this was that if/when user specify a custom fallback, they need not worry about explicitly inheriting or calling the interpreter ops fallback. That is, interpreter ops would always have a kernel defined.

Were you thinking of having the interpreter op kernels be optional in a singular overridable callback?

stablehlo/tools/StablehloTranslateMain.cpp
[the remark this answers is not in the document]
ghpvnist: Ah okay, I didn't see this being used in the diffs initially.

stablehlo/reference/InterpreterApi.h
```diff
@@ -0,0 +1,99 @@
...
+  virtual llvm::Error handleOp(Operation &op, Process *process, Scope &scope);
+
+  /// The user provided interpreter configuration.
+  const InterpreterConfiguration *config;
+
+  /// The topmost function currently being executed in the module.
+  func::FuncOp currentFunction;
+
+ private:
+  /// Counts how many times a given probe_id has been used while profiling.
```
ghpvnist, line 76: Could you elaborate on the comment from the viewpoint of the API; from referencing this doc only, I don't know what `probe_id` is, which is an implementation detail of the ProbeOp.

ghpvnist (changes requested):
I reviewed the rest and left a few comments. PTAL.

ghpvnist (changes requested):

BUILD.bazel
```diff
@@ -392,6 +395,32 @@ cc_library(
...
+        ":reference_interpretervalue",
+        ":reference_numpy",
+        ":reference_ops",
+        ":reference_process",
+        ":reference_scope",
+        ":reference_tensor",
+        ":register",
+        "@llvm-project//llvm:Support",
+        "@llvm-project//mlir:FuncDialect",
+        "@llvm-project//mlir:IR",
```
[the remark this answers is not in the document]
penagos: Removed `:register` dependency, but I do think we need the others due to existing `#include` in the impl file.

BUILD.bazel
```diff
@@ -392,6 +395,32 @@ cc_library(
...
+    deps = [
+        ":interpreter_ops",
+        ":reference_errors",
+        ":reference_interpretervalue",
+        ":reference_numpy",
+        ":reference_ops",
+        ":reference_process",
+        ":reference_scope",
+        ":reference_tensor",
+        ":register",
```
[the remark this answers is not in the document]
penagos: I think we still need some of the `reference_*` dependencies for the header includes.

stablehlo/reference/InterpreterApi.h
```diff
@@ -0,0 +1,95 @@
...
+  /// Set the topmost currently executing function in the module.
+  void setFunction(func::FuncOp function) { currentFunction = function; }
+
+ protected:
+  /// Custom op kernels for any user specified ops not found in the StableHLO
+  /// op dialect or StableHLO interpreter dialect.
+  virtual llvm::Error handleOp(Operation &op, Process *process, Scope &scope);
+
+  /// The user provided interpreter configuration.
+  const InterpreterConfiguration *config;
```
GleasonK, line 73: I don't like this setup where we have an `InterpreterConfig` and an `InterpreterFallback` which both must point at eachother if a fallback exists.

Can we make this a unidirectional dependency where a Config has a Fallback and that's it. If there's ever a reason for the back-pointer, I think we should modify other APIs to accept a config instead of a fallback.

If we do this refactoring we can:
- Remove the forward declaration. 
- Update InterpreterConfig to take an InterpreterFallback by reference (open to thoughts on this, I'm ok with pointer, just worth noting that under the current setup we throw away a fallback every time we use `StablehloTranslate` since we override it)

stablehlo/reference/InterpreterApi.h
```diff
@@ -0,0 +1,95 @@
...
+
+  virtual ~InterpreterFallback() = default;
+
+  /// Set the user provided interpreter configuration.
+  void setConfig(const InterpreterConfiguration &config) {
+    this->config = &config;
+  }
+
+  /// Set the topmost currently executing function in the module.
+  void setFunction(func::FuncOp function) { currentFunction = function; }
```
GleasonK, line 65: Let's remove the `currentFunction` field and this setter. I believe this value will currently be wrong any time we call a function..since it is only updated at the top level?

Instead we can use `op.getParentOfType<func::FuncOp>()` to get the nearest enclosing FuncOp from the IR when it is needed.

stablehlo/reference/InterpreterApi.cpp
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  SmallVector<InterpreterValue> results;
+  config.fallback->setConfig(config);
+
+  auto walkResult = module.walk([&](func::FuncOp funcOp) {
+    if (numFuncs > 1 && funcOp.getSymName() != config.mainFunction)
+      return WalkResult::advance();
+
+    if (config.fallback) config.fallback->setFunction(funcOp);
```
GleasonK, line 109: Delete per other feedback.

stablehlo/reference/InterpreterApi.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+    results = stablehlo::eval(funcOp.getBody(), inputs, /*process=*/nullptr,
+                              /*parent=*/nullptr, *config.fallback);
+
+    if (config.stream)
```
GleasonK, line 114: Let's make this even more flexible, and let `config` take a `resultHandler` function which takes results and returns a `llvm::Error`. The `StablehloTranslate` impl of this function can call print and return success.

This method will just throw the error if there's a failure in the callback.

stablehlo/reference/InterpreterApi.cpp
```diff
@@ -0,0 +1,126 @@
...
+                            stablehlo::numpy::kInstrumentationMetadataFilename);
+    if (llvm::sys::fs::remove(instrumentationMetadataFile))
+      llvm::report_fatal_error(
+          "Failed to remove existing instrumentation metadata file.");
+  }
+
+  SmallVector<InterpreterValue> results;
+  config.fallback->setConfig(config);
+
+  auto walkResult = module.walk([&](func::FuncOp funcOp) {
```
GleasonK, line 105: Per above comment, I don't think we need this walk, if the goal is to just find the main function for executing, lets get it above and just call eval here.

stablehlo/reference/InterpreterApi.h
```diff
@@ -0,0 +1,95 @@
...
+  /// If the input StableHLO program has been instrumented, keep track of how
+  /// many times a given operation has been executed.
+  llvm::StringMap<int32_t> instrumentedTensors;
+};
+
+/// Invoke the StableHLO reference interpreter with the given parsed MLIR
+/// module input and provided inputs. Returns a list of interpreter outputs.
+/// Can optionally pass a fallback interpreter callback which executes when no
+/// builtin kernels are matched.
+llvm::ErrorOr<SmallVector<InterpreterValue>> runInterpreter(
```
GleasonK, line 88: I've been trying to think of a better name, and so far I've come up with:
- referenceEval
- evalModule
- runModule
- interpret
- interpretModule

I think `evalModule` is my preference since it is more clear what the inputs are and what it is doing.

GleasonK (changes requested):

stablehlo/reference/InterpreterApi.cpp
```diff
@@ -0,0 +1,126 @@
...
+}
+
+llvm::ErrorOr<SmallVector<InterpreterValue>> runInterpreter(
+    ModuleOp module, ArrayRef<InterpreterValue> inputs,
+    const InterpreterConfiguration &config) {
+  auto numFuncs = 0;
+  bool hasMain = false;
+
+  assert(!config.mainFunction.empty() && "Main function name cannot be empty");
+  module.walk([&](func::FuncOp funcOp) {
```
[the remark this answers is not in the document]
penagos: I think to preserve the current behavior (i.e. for existing LIT tests), the semantics of the function need to be if there is only 1 function in the module, use that as the main function. Otherwise, use the provided `mainFunction` string to break ties when there are more than 1 function. Slightly tweaked proposed change with this.

stablehlo/reference/InterpreterApi.cpp
penagos: Thinking about this more, we probably don't need this complexity at all. I.e. we can move this specific use case into the `interpretRegistration` callback in `StablehloTranslateMain.cpp`.

stablehlo/reference/InterpreterApi.h
penagos: I agree, it's a bit confusing of a design. I've refactored it to pass through a config instead to `eval` as the default fallback does need some configuration state (so the fallback functor will still take a config, but it doesn't have a chance of being "outdated" by the toplevel caller not correctly calling `setConfig` before the fallback runs).

However, I think there's an opportunity to further clean up some of the current interpreter design here. When changing the interface, I noticed there's a multiple op kernels which take a `Process` and `Scope`. For example, `evalAllReduceOp`, `evalCaseOp`, `evalIfOp` to name a few, which later go on to recursively call back into teh top level `eval`. However, they don't preserve the current top level fallback function.

For example, `evalIfOp` calls back into `eval` for both true/false branches:

```
  return pred.get({}).getBooleanValue()
             ? eval(trueBranch, {}, /*config=*/nullptr, process, &scope)
             : eval(falseBranch, {}, /*config=*/nullptr, process, &scope);
```

But presumably, if either branch had an op which was handled in the top level specified fallback, that wouldn't work. I'm not sure if this is by design, or an omission. If the latter, since `eval` would also need to take an `InterpreterConfiguration` for properly handling interpreter instrumentation, would it be worthwhile to instead encapsulate `Process`, `Scope` and `InterpreterConfiguration` into something like an `InterpreterState` struct?

stablehlo/reference/InterpreterApi.cpp
```diff
@@ -0,0 +1,114 @@
...
+namespace mlir {
+namespace stablehlo {
+namespace {
+func::FuncOp getMainFunction(ModuleOp module, StringRef mainName) {
+  auto functions = module.getOps<func::FuncOp>();
+
+  // If the module has 1 function only, use it as the main function.
+  if (std::distance(functions.begin(), functions.end()) == 1) {
+    return *functions.begin();
+  }
```
GleasonK, line 37: Nit - there is something odd that this function takes a `mainName` that is ignored if there is only 1 function.

```
func.func @not_main() { ... }
```

If I call `evalModule` for `main` on the above module.. it still runs.

stablehlo/reference/Ops.cpp
```diff
@@ -817,18 +818,18 @@ SmallVector<InterpreterValue> eval(
       auto operand = scope.find(whileOp.getOperand());
       auto &cond = whileOp.getCond();
       auto &body = whileOp.getBody();
-      auto results = evalWhileOp(operand, cond, body, process, scope, fallback);
+      auto results = evalWhileOp(operand, cond, body, process, scope, config);
```
GleasonK, line 821: Nit, lets be uniform with where the config arg appears. In `eval` it is before process/scope, in `evalWhileOp` it is a trailing argument. I don't have a strong preference on which position it should be, just that it is consistent.

stablehlo/reference/Ops.cpp
```diff
@@ -199,10 +200,10 @@ SmallVector<Tensor> split(const Tensor &x, int64_t numResults, Axis axis,
 
 }  // namespace
 
-SmallVector<InterpreterValue> eval(
-    Region &region, ArrayRef<InterpreterValue> args, Process *process,
-    Scope *parent,
-    llvm::function_ref<llvm::Error(Operation &, Process *, Scope &)> fallback) {
+SmallVector<InterpreterValue> eval(Region &region,
+                                   ArrayRef<InterpreterValue> args,
+                                   const InterpreterConfiguration *config,
```
GleasonK, line 205: Why not make this a reference? It looks like the important part is that Fallback is optional. I think making config optional and fallback optional makes the code more complicated.

```cpp
if (config && !config->fallback)
==>
if (!config->fallback)

foo(...,/*config=*/nullptr)
==>
foo(..., /*config=*/{})
```

GleasonK: Also for the places where config is nullptr.. is this the proper behavior? I feel like this should be passed through regions? I.e. probe in the body of a `stablehlo.map` operation looks like it wouldn't work currently

stablehlo/reference/Ops.h
```diff
@@ -206,10 +206,9 @@ Tensor evalXorOp(const Tensor &lhs, const Tensor &rhs, ShapedType resultType);
 /// used for evaluating ops which are not supported by the interpreter.
 /// Assumes that `region` has only one block.
 SmallVector<InterpreterValue> eval(
-    Region &region, ArrayRef<InterpreterValue> args, Process *process = nullptr,
-    Scope *parent = nullptr,
-    function_ref<llvm::Error(Operation &, Process *, Scope &)> fallback =
-        nullptr);
+    Region &region, ArrayRef<InterpreterValue> args,
+    const InterpreterConfiguration *config = nullptr,
```
GleasonK, line 210: Per earlier comment, I don't think this should be optional here. Let's flip to reference and pass the context through.

GleasonK (changes requested):

stablehlo/reference/InterpreterApi.h
GleasonK: Ah ignore my other comment, I didn't realize these bugs existed prior to your change. This is on us to clean up. I'll make a ticket with this refactoring / bug fixing

GleasonK: Created https://github.com/openxla/stablehlo/issues/1820

Thanks for calling this out. Feel free to resolve comment, I'm currently impartial to the refactoring you've mentioned. If we end up needing to refactor all callsites with a new signature for this context we may reconsider this

stablehlo/reference/InterpreterApi.cpp
GleasonK: Maybe we add a check that the user didn't specify a custom main function. I.e. only use the 1 function if the function we're looking for is `main`.

Let's move this check below the iteration over all ops, and add a check for:

```cpp
  for (auto funcOp : functions)
    if (funcOp.getSymName().equals(mainName)) return funcOp;
 
bool isSingleFunction = std::distance(functions.begin(), functions.end()) == 1;
bool isDefaultLookup = mainName == "main";
if (isSingleFunction && isDefaultLookup)  return *functions.begin();
```

stablehlo/reference/Ops.cpp
penagos: Yup, https://github.com/openxla/stablehlo/pull/1797#discussion_r1378996442 is the reason this is an optional pointer for the time being, since otherwise the other interfaces that call into `eval` today would need to be updated to take a `config` as described in #1820. I think maybe this could be changed to required pass by reference once those ops are updated to take the fallback interface correctly.

stablehlo/reference/Ops.cpp
penagos: Decided to keep it before `Process` as I think once #1820 is addressed, we can make it a `const InterpreterConfiguration&` and keep it before any other optional pointer parameters.

GleasonK (approved):
LGTM. Thanks for the contribution!

ghpvnist (approved):
Looks like we need to add InterpreterApi dependency. All looks good to me as well otherwise. Thanks again for the contribution!

penagos:
Refactored the interface to avoid the build issue and to try and make it a bit cleaner. Instead of having 2 functions in the fallback callback interface functor, there's just the singular `operator()` now, and I've reverted the `evalXX` changes to pass around a config, and instead pass around a fallback callback pointer as before.

Now, the interpreter interface will create a lightweight wrapper that sits ontop of the optionally overridable fallback callback, to handle kernels for probe/parallel ops. This should avoid the circular include dependency, and somewhat simplify the interface.

stablehlo/reference/InterpreterApi.cpp
```diff
@@ -44,47 +43,63 @@ func::FuncOp getMainFunction(ModuleOp module, StringRef mainName) {
...
+
+      SymbolTable symbolTable{op.getParentOfType<ModuleOp>()};
+      auto results = stablehlo::interpreter::evalRunParallelOp(
+          runtimeOperands, infeed, programs, symbolTable);
+      scope.add(runParallelOp.getResults(), results);
+      return wrapFallbackStatus(llvm::Error::success(), funcName,
+                                "interpreter.run_parallel");
+    }
+
+    return (*config.fallback)(op, scope, process);
```
GleasonK, line 90: If config points back to this class, which I think is a typical usage(?) wouldn't this recurse infinitely?

GleasonK: Ah. This is an internal implementation detail. Please add a comment to the class saying "DefaultInterpreterFallback is an implementation detail of run module. It takes in an InterpreterConfiguration which can have user-implemented fallbacks"

As of now, this setup means that a custom op cannot have a region that uses a default callback (i.e. my.op with a region containing a probe op). This is probably an OK limitation for now, but would require refactoring into something more composable if this ever becomes a requirement.

GleasonK (approved):

PR #1806, opened 2023-10-13
Fix clang tidy, remove unused dependencies

ghpvnist:
cc: @penagos


burmako (approved):
Approving to facilitate an upcoming downstream integrate

PR #1807, opened 2023-10-16
Integrate LLVM at llvm/llvm-project@97217d188469

burmako:


ghpvnist (approved):

stablehlo/reference/NumPy.cpp
ghpvnist, line 1: Changes in this file should be reflected in the downstream integrate soon.

PR #1809, opened 2023-10-17
[RFC] Add collective_broadcast to the StableHLO specification

chaserileyroberts:
This RFC proposes adding `collective_broadcast` as one of the collective communication primitives. 
Please provide any feedback you feel is valuable.

### DIFF
--- a/rfcs/20231017-collective-broadcast.md
+++ b/rfcs/20231017-collective-broadcast.md
@@ -9,7 +9,7 @@ Discussion thread: [GitHub](https://github.com/openxla/stablehlo/pull/1809)
 
 StableHLO currently has [five collective communication primitives](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#collective-ops): `collective_permute`, `all_gather`, `all_to_all`, `all_reduce`, and `reduce_scatter`. However, one of the major collective communication primitives, `broadcast`, is missing from this list. This primitive allows for a one-to-many replication of a tensor to many devices efficiently. `broadcast` is a primitive in [MPI](https://www.open-mpi.org/doc/v4.1/man3/MPI_Bcast.3.php), [NCCL](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/api/colls.html#c.ncclBroadcast), and [PyTorch](https://pytorch.org/docs/stable/distributed.html#torch.distributed.broadcast). From here on out, we will refer to this operation as `collective_broadcast` for reasons discussed later.
 
-While it technically would be possible to replicate a broadcast with a conditional mask and a `psum`, that reduces to an `all_reduce` communication primitive, which is significantly more expensive than a simple `collective_broadcast`. Additionally, when dealing with network-switch environments, the explicit use of `collective_broadcast` allows the switch to greatly optimize it's throughput when replicating to many targets simultaneously. However, XLA currently has no ability to lower directly to a mesh's `collective_broadcast` primitive, so a lot of that optimization is left on the table. 
+While it technically would be possible to replicate a broadcast with a conditional mask and a `psum`, that reduces to an `all_reduce` communication primitive, which is significantly more expensive than a simple `collective_broadcast`. Additionally, when dealing with network-switch environments, the explicit use of `collective_broadcast` allows the switch to greatly optimize it's throughput when replicating to many targets simultaneously. However, XLA currently has no ability to lower directly to a mesh's `collective_broadcast` primitive, so a lot of that optimization is left on the table.
 
 Additionally, a new compiler pass that detects usage of the old `psum` hack and replaces it with a `collective_broadcast` could be implemented only once and forever be supported by all hardware, future and current. This could have positive knock-on effects for users who don't even realize they're using it!
 
@@ -31,8 +31,8 @@ Unfortunately, the op name `broadcast` is already taken by [an op in XLA proper]
 Within each process group in the StableHLO process grid, send the value of the
 `operand` tensor from the source process to the target processes and produce a
 `result` tensor. The first process in each of the `replica_groups` is treated as the 
-source process. For any process `n` not in any replica group, an additional replica 
-group `[n]` will be implied. 
+source process. For any process `n` not in any replica group, that process will just return 
+zeros.
 
 The operation splits the StableHLO process grid into `process_groups` which is
 defined as follows:
@@ -78,12 +78,13 @@ Afterwards, within each `process_group`:
 
 ```mlir
 // num_replicas: 1
-// num_partitions: 3
+// num_partitions: 4
 // %operand@(0, 0): [[1, 2]]
 // %operand@(0, 1): [[3, 4]]
 // %operand@(0, 2): [[5, 6]]
+// %operand@(0, 3): [[7, 8]]
 %result = "stablehlo.collective_broadcast"(%operand) {
-  replica_groups = [[2, 0], [1]]
+  replica_groups = dense<[[2, 0], [1, 3]]> : tensor<2x2xi64>,
   // channel_id = 0
   channel_handle = #stablehlo.channel_handle<handle = 0, type = 0>
   // use_global_device_ids = false
@@ -91,4 +92,25 @@ Afterwards, within each `process_group`:
 // %result@(0, 0): [[5, 6]]
 // %result@(0, 1): [[3, 4]]
 // %result@(0, 2): [[5, 6]]
+// %result@(0, 4): [[3, 4]]
+```
+
+Example with missing processes
+```mlir
+// num_replicas: 1
+// num_partitions: 4
+// %operand@(0, 0): [[1, 2]]
+// %operand@(0, 1): [[3, 4]]
+// %operand@(0, 2): [[5, 6]]
+// %operand@(0, 3): [[7, 8]]
+%result = "stablehlo.collective_broadcast"(%operand) {
+  replica_groups = dense<[[2, 1]]> : tensor<1x2xi64>,
+  // channel_id = 0
+  channel_handle = #stablehlo.channel_handle<handle = 0, type = 0>
+  // use_global_device_ids = false
+} : (tensor1x2xi64>) -> tensor<1x2xi64>
+// %result@(0, 0): [[0, 0]]
+// %result@(0, 1): [[5, 6]]
+// %result@(0, 2): [[5, 6]]
+// %result@(0, 4): [[0, 0]]
 ```
--- a/rfcs/20231017-collective-broadcast.md
+++ b/rfcs/20231017-collective-broadcast.md
@@ -2,7 +2,7 @@
 
 Status: Review<br/>
 Initial version: 10/17/20223<br/>
-Last updated: 10/17/2023<br/>
+Last updated: 11/1/2023<br/>
 Discussion thread: [GitHub](https://github.com/openxla/stablehlo/pull/1809)
 
 ## Motivation
@@ -30,49 +30,44 @@ Unfortunately, the op name `broadcast` is already taken by [an op in XLA proper]
 
 Within each process group in the StableHLO process grid, send the value of the
 `operand` tensor from the source process to the target processes and produce a
-`result` tensor. The first process in each of the `replica_groups` is treated as the 
-source process. For any process `n` not in any replica group, that process will just return 
-zeros.
+`result` tensor. For any process `n` not in any replica group, that process will just
+return zeros.
 
 The operation splits the StableHLO process grid into `process_groups` which is
 defined as follows:
 
-* `cross_replica(replica_groups)`
-  if `channel_id <= 0 and use_global_device_ids = false`.
-* `cross_replica_and_partition(replica_groups)`
-  if `channel_id > 0 and use_global_device_ids = false`.
-* `flattened_ids(replica_groups)`
-  if `channel_id > 0 and use_global_device_ids = true`.
+* `cross_replica(replica_groups)` if `channel_id <= 0`.
+* `cross_partition(replica_groups)` if `channel_id > 0`.
 
-Afterwards, within each `process_group`:
+Afterwards, `result@process` is given by:
 
-* `result@process[result_index] = operand@process[0]`
+* `operand@process_groups[i, 0]` if there exists an `i` such that
+  the process is in `process_groups[i]`.
+* `broadcast_in_dim(constant(0, element_type(result)), [], type(result))`
+  if process is never found in `process_groups`.
+
 
 #### Inputs
 
 | Label | Name                    | Type                                                             | Constraints |
 |-------|-------------------------|------------------------------------------------------------------|-------------|
-| (I1)  | `operand`               | tensor                                                           | (C5)        |
-| (I2)  | `replica_groups`        | variadic number of 1-dimensional tensor constants of type `si64` | (C1-C3)     |
-| (I3)  | `channel_id`            | constant of type `si64`                                          | (C4)        |
-| (I4)  | `use_global_device_ids` | constant of type `i1`                                            | (C4)        |
+| (I1)  | `operand`               | tensor                                                           | (C3)        |
+| (I2)  | `replica_groups`        | variadic number of 1-dimensional tensor constants of type `si64` | (C1-C2)     |
+| (I3)  | `channel_id`            | constant of type `si64`                                          |             |
 
 #### Outputs
 
 | Name     | Type   | Constraints |
 |----------|--------|-------------|
-| `result` | tensor | (C5)        |
+| `result` | tensor | (C3)        |
 
 #### Constraints
 
-* (C1) `is_unique(replica_groups)`.
-* (C2) `size(replica_groups)` is defined as:
+* (C1) is_unique(replica_groups).
+* (C2) `0 <= replica_groups < N` where `N` is defined as:
   * `num_replicas` if `cross_replica` is used.
-  * `num_replicas` if `cross_replica_and_partition` is used.
-  * `num_processes` if `flattened_ids` is used.
-* (C3) `0 <= replica_groups < size(replica_groups)`.
-* (C4) If `use_global_device_ids = true`, then `channel_id > 0`.
-* (C5) `type(result) = type(operand)`.
+  * `num_partitions` if `cross_partition` is used.
+* (C3) `type(result) = type(operand)`.
 
 #### Examples
 
@@ -85,9 +80,8 @@ Afterwards, within each `process_group`:
 // %operand@(0, 3): [[7, 8]]
 %result = "stablehlo.collective_broadcast"(%operand) {
   replica_groups = dense<[[2, 0], [1, 3]]> : tensor<2x2xi64>,
-  // channel_id = 0
-  channel_handle = #stablehlo.channel_handle<handle = 0, type = 0>
-  // use_global_device_ids = false
+  // channel_id = 1
+  channel_handle = #stablehlo.channel_handle<handle = 1, type = 0>
 } : (tensor1x2xi64>) -> tensor<1x2xi64>
 // %result@(0, 0): [[5, 6]]
 // %result@(0, 1): [[3, 4]]
@@ -105,9 +99,8 @@ Example with missing processes
 // %operand@(0, 3): [[7, 8]]
 %result = "stablehlo.collective_broadcast"(%operand) {
   replica_groups = dense<[[2, 1]]> : tensor<1x2xi64>,
-  // channel_id = 0
-  channel_handle = #stablehlo.channel_handle<handle = 0, type = 0>
-  // use_global_device_ids = false
+  // channel_id = 1
+  channel_handle = #stablehlo.channel_handle<handle = 1, type = 0>
 } : (tensor1x2xi64>) -> tensor<1x2xi64>
 // %result@(0, 0): [[0, 0]]
 // %result@(0, 1): [[5, 6]]
--- a/rfcs/20231017-collective-broadcast.md
+++ b/rfcs/20231017-collective-broadcast.md
@@ -1,0 +1,94 @@
+# [RFC] Add collective_broadcast to the StableHLO specification
+
+Status: Review<br/>
+Initial version: 10/17/20223<br/>
+Last updated: 10/17/2023<br/>
+Discussion thread: TBD
+
+## Motivation
+
+StableHLO currently has [five collective communication primitives](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#collective-ops): `collective_permute`, `all_gather`, `all_to_all`, `all_reduce`, and `reduce_scatter`. However, one of the major collective communication primitives, `broadcast`, is missing from this list. This primitive allows for a one-to-many replication of a tensor to many devices efficiently. `broadcast` is a primitive in [MPI](https://www.open-mpi.org/doc/v4.1/man3/MPI_Bcast.3.php), [NCCL](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/api/colls.html#c.ncclBroadcast), and [PyTorch](https://pytorch.org/docs/stable/distributed.html#torch.distributed.broadcast). From here on out, we will refer to this operation as `collective_broadcast` for reasons discussed later.
+
+While it technically would be possible to replicate a broadcast with a conditional mask and a `psum`, that reduces to an `all_reduce` communication primitive, which is significantly more expensive than a simple `collective_broadcast`. Additionally, when dealing with network-switch environments, the explicit use of `collective_broadcast` allows the switch to greatly optimize it's throughput when replicating to many targets simultaneously. However, XLA currently has no ability to lower directly to a mesh's `collective_broadcast` primitive, so a lot of that optimization is left on the table. 
+
+Additionally, a new compiler pass that detects usage of the old `psum` hack and replaces it with a `collective_broadcast` could be implemented only once and forever be supported by all hardware, future and current. This could have positive knock-on effects for users who don't even realize they're using it!
+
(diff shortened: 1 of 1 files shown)


rfcs/20231017-collective-broadcast.md
```diff
@@ -0,0 +1,94 @@
...
+
+Additionally, a new compiler pass that detects usage of the old `psum` hack and replaces it with a `collective_broadcast` could be implemented only once and forever be supported by all hardware, future and current. This could have positive knock-on effects for users who don't even realize they're using it!
+
+`collective_broadcast` can be used to quickly replicate a tensor across an entire mesh, and would use less communication resources as compared to `all_gather` or `psum`. `collective_broadcast` is also the base primitive used in the [SUMMA](https://www.netlib.org/lapack/lawnspdf/lawn96.pdf) distributed GEMM algorithm. As AI computing grows larger, there likely will grow a need for these 2D distributed GEMM algorithms. Adding support for one of the needed primitives could help advance research in these areas. 
+
+## Alternatives considered
```
GleasonK, line 18: I'd assume "use existing `psum` to `all_reduce` lowering and handle `all_reduce`" would be an alternative as well. PT/XLA emits a similar `all_reduce` pattern for torch bradcast. Above you mention:

> While it technically would be possible to replicate a broadcast with a conditional mask and a `psum`, that reduces to an `all_reduce` communication primitive, which is significantly more expensive than a simple `collective_broadcast`.

Could you add more details to this? Is it difficult (or infeasible) to pattern match / map the existing lowering to the desired broadcast operation on the compiler side, or why can't this operation be used for the desired optimization? It sounds like lowering this `psum` pattern to `collective_broadcast` is an option, so it would be good to add details on why we'd like a new op instead.

rfcs/20231017-collective-broadcast.md
```diff
@@ -0,0 +1,94 @@
...
+
+While it technically would be possible to replicate a broadcast with a conditional mask and a `psum`, that reduces to an `all_reduce` communication primitive, which is significantly more expensive than a simple `collective_broadcast`. Additionally, when dealing with network-switch environments, the explicit use of `collective_broadcast` allows the switch to greatly optimize it's throughput when replicating to many targets simultaneously. However, XLA currently has no ability to lower directly to a mesh's `collective_broadcast` primitive, so a lot of that optimization is left on the table.
```
GleasonK, line 12: > However, XLA currently has no ability to lower directly to a mesh's `collective_broadcast` primitive, so a lot of that optimization is left on the table.

IIUC this new op is only useful if a dedicated HLO is added for it in [openxla/xla](https://github.com/openxla/xla). Prior to adding to StableHLO, would be good to get buy-in / agreement on adding a new HLO.

chaserileyroberts:
> Also - could you follow this link and sign the CLA when you get a chance.

I work at Nvidia so I need to go through the separate channel. I'll take care of it don't worry.

rfcs/20231017-collective-broadcast.md
chaserileyroberts: >  it difficult (or infeasible) to pattern match / map the existing lowering to the desired broadcast operation on the compiler side.

The issue is that doing this pattern matching without the `collective_broadcast` primitive means that the logic will need to be reimplemented for every hardware implementation (CPU, GPU, TPU, etc). If we instead had the primitive, then we could implement that compiler pass at an early, hardware agnostic stage. Each hardware specific compiler would need to only worry about implementing `collective_broadcast`, and then they would also get the `psum hack -> broadcast` transform for free forever!

I'm aware the PT emits a similar pattern when it lowers its broadcast, and that is also one of the reasons I want to add this primitive so that PT doesn't need to do emit hack anymore.

rfcs/20231017-collective-broadcast.md
chaserileyroberts: Sure, is there a similar RFC mechanism to raise to raise awareness to XLA proper?

rfcs/20231017-collective-broadcast.md
```diff
@@ -0,0 +1,94 @@
...
+#### Examples
+
+```mlir
+// num_replicas: 1
+// num_partitions: 3
+// %operand@(0, 0): [[1, 2]]
+// %operand@(0, 1): [[3, 4]]
+// %operand@(0, 2): [[5, 6]]
+%result = "stablehlo.collective_broadcast"(%operand) {
+  replica_groups = [[2, 0], [1]]
```
trevor-m, line 86: I think the replica groups need to represented this way in order to be valid stablehlo:
`replica_groups = dense<[[2, 0], [1, -1]]> : tensor<2x2xi64>,`

chaserileyroberts: Oh that's a good point. I really don't like the -1 in the list, so should we instead just require that all processes are represented and the replica_groups all need to be the same size?

chaserileyroberts: I updated this now and changed the definition of what to do when processes are missing.

rfcs/20231017-collective-broadcast.md
```diff
@@ -0,0 +1,116 @@
...
+// num_replicas: 1
+// num_partitions: 4
+// %operand@(0, 0): [[1, 2]]
+// %operand@(0, 1): [[3, 4]]
+// %operand@(0, 2): [[5, 6]]
+// %operand@(0, 3): [[7, 8]]
+%result = "stablehlo.collective_broadcast"(%operand) {
+  replica_groups = dense<[[2, 0], [1, 3]]> : tensor<2x2xi64>,
+  // channel_id = 0
+  channel_handle = #stablehlo.channel_handle<handle = 0, type = 0>
```
ghpvnist, line 89: The channel handle (alias of channel_id) should be 1 for cross_partition communication since we have 4 partitions. Similarly with the next example.
```
channel_handle = #stablehlo.channel_handle<handle = 1, type = 0>
```

rfcs/20231017-collective-broadcast.md
```diff
@@ -0,0 +1,116 @@
...
+```mlir
+// num_replicas: 1
+// num_partitions: 4
+// %operand@(0, 0): [[1, 2]]
+// %operand@(0, 1): [[3, 4]]
+// %operand@(0, 2): [[5, 6]]
+// %operand@(0, 3): [[7, 8]]
+%result = "stablehlo.collective_broadcast"(%operand) {
+  replica_groups = dense<[[2, 0], [1, 3]]> : tensor<2x2xi64>,
+  // channel_id = 0
```
ghpvnist, line 88: This comment can be removed, similarly with the next example.

rfcs/20231017-collective-broadcast.md
```diff
@@ -0,0 +1,116 @@
...
+  channel_handle = #stablehlo.channel_handle<handle = 0, type = 0>
+  // use_global_device_ids = false
```
ghpvnist, line 90: Remove this comment too with the updated semantics, similarly with the next example.

rfcs/20231017-collective-broadcast.md
```diff
@@ -0,0 +1,116 @@
...
+
+#### Outputs
+
+| Name     | Type   | Constraints |
+|----------|--------|-------------|
+| `result` | tensor | (C5)        |
+
+#### Constraints
+
+* (C1) `is_unique(replica_groups)`.
```
ghpvnist, line 68: Most of these constraints will be outdated. Here's a new set of constraints for the updated semantics:
```
* (C1) is_unique(replica_groups[:, 0]).
* (C2) is_unique(replica_groups[:, 1:]).
* (C3) `0 <= replica_groups < N` where `N` is defined as:
  * `num_partitions` if `cross_partition` is used.
* (C4) `type(result) = type(operand)`.
```

rfcs/20231017-collective-broadcast.md
chaserileyroberts: Done

rfcs/20231017-collective-broadcast.md
chaserileyroberts: Done

rfcs/20231017-collective-broadcast.md
chaserileyroberts: Done

rfcs/20231017-collective-broadcast.md
```diff
@@ -0,0 +1,116 @@
...
+* (C2) `size(replica_groups)` is defined as:
+  * `num_replicas` if `cross_replica` is used.
+  * `num_replicas` if `cross_replica_and_partition` is used.
+  * `num_processes` if `flattened_ids` is used.
+* (C3) `0 <= replica_groups < size(replica_groups)`.
+* (C4) If `use_global_device_ids = true`, then `channel_id > 0`.
```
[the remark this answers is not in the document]
chaserileyroberts: Updated.

rfcs/20231017-collective-broadcast.md
```diff
@@ -0,0 +1,116 @@
...
+* `result@process[result_index] = operand@process[0]`
+
+#### Inputs
+
+| Label | Name                    | Type                                                             | Constraints |
+|-------|-------------------------|------------------------------------------------------------------|-------------|
+| (I1)  | `operand`               | tensor                                                           | (C5)        |
+| (I2)  | `replica_groups`        | variadic number of 1-dimensional tensor constants of type `si64` | (C1-C3)     |
+| (I3)  | `channel_id`            | constant of type `si64`                                          | (C4)        |
+| (I4)  | `use_global_device_ids` | constant of type `i1`                                            | (C4)        |
```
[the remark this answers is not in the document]
chaserileyroberts: Updated. Please review the new constraint set.

rfcs/20231017-collective-broadcast.md
```diff
@@ -0,0 +1,116 @@
...
+defined as follows:
+
+* `cross_replica(replica_groups)`
+  if `channel_id <= 0 and use_global_device_ids = false`.
+* `cross_replica_and_partition(replica_groups)`
+  if `channel_id > 0 and use_global_device_ids = false`.
+* `flattened_ids(replica_groups)`
+  if `channel_id > 0 and use_global_device_ids = true`.
+
+Afterwards, within each `process_group`:
```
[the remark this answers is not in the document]
chaserileyroberts: Updated and added more wording for the zeros case.

rfcs/20231017-collective-broadcast.md
```diff
@@ -0,0 +1,116 @@
...
+Within each process group in the StableHLO process grid, send the value of the
+`operand` tensor from the source process to the target processes and produce a
+`result` tensor. The first process in each of the `replica_groups` is treated as the 
+source process. For any process `n` not in any replica group, that process will just return 
+zeros.
+
+The operation splits the StableHLO process grid into `process_groups` which is
+defined as follows:
+
+* `cross_replica(replica_groups)`
```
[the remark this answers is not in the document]
chaserileyroberts: Updated

rfcs/20231017-collective-broadcast.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Proposed Specification
+
+### collective_broadcast
+
+#### Semantics
+
+Within each process group in the StableHLO process grid, send the value of the
+`operand` tensor from the source process to the target processes and produce a
+`result` tensor. The first process in each of the `replica_groups` is treated as the
```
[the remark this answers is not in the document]
chaserileyroberts: Updated

chaserileyroberts (commented):
Made updates based on the new semantics.

rfcs/20231017-collective-broadcast.md
```diff
@@ -0,0 +1,109 @@
...
+
+Within each process group in the StableHLO process grid, send the value of the
+`operand` tensor from the source process to the target processes and produce a
+`result` tensor. For any process `n` not in any replica group, that process will just
+return zeros.
+
+The operation splits the StableHLO process grid into `process_groups` which is
+defined as follows:
+
+* `cross_replica(replica_groups)` if `channel_id <= 0`.
```
changm, line 39: Sorry I'm new here, so I'm not sure this is a comment on the doc or an actual question :)

What's does having a channel_id <= 0 here mean? This sounds maybe like an attribute / argument?

rfcs/20231017-collective-broadcast.md
```diff
@@ -0,0 +1,109 @@
...
+
+#### Constraints
+
+* (C1) is_unique(replica_groups).
+* (C2) `0 <= replica_groups < N` where `N` is defined as:
+  * `num_replicas` if `cross_replica` is used.
+  * `num_partitions` if `cross_partition` is used.
+* (C3) `type(result) = type(operand)`.
+
+#### Examples
```
changm, line 72: Do you have an example with broadcast having a handle < 0?

rfcs/20231017-collective-broadcast.md
```diff
@@ -0,0 +1,109 @@
...
+* `broadcast_in_dim(constant(0, element_type(result)), [], type(result))`
+  if process is never found in `process_groups`.
+
+
+#### Inputs
+
+| Label | Name                    | Type                                                             | Constraints |
+|-------|-------------------------|------------------------------------------------------------------|-------------|
+| (I1)  | `operand`               | tensor                                                           | (C3)        |
+| (I2)  | `replica_groups`        | variadic number of 1-dimensional tensor constants of type `si64` | (C1-C2)     |
```
ghpvnist, line 55: Nit: (C1), (C2)

rfcs/20231017-collective-broadcast.md
```diff
@@ -0,0 +1,109 @@
...
+```mlir
+// num_replicas: 1
+// num_partitions: 4
+// %operand@(0, 0): [[1, 2]]
+// %operand@(0, 1): [[3, 4]]
+// %operand@(0, 2): [[5, 6]]
+// %operand@(0, 3): [[7, 8]]
+%result = "stablehlo.collective_broadcast"(%operand) {
+  replica_groups = dense<[[2, 0], [1, 3]]> : tensor<2x2xi64>,
+  // channel_id = 1
```
ghpvnist, line 83: `channel_id = 1` is redundant from `channel_handle`, similarly for example below.

rfcs/20231017-collective-broadcast.md
```diff
@@ -0,0 +1,109 @@
...
+defined as follows:
+
+* `cross_replica(replica_groups)` if `channel_id <= 0`.
+* `cross_partition(replica_groups)` if `channel_id > 0`.
+
+Afterwards, `result@process` is given by:
+
+* `operand@process_groups[i, 0]` if there exists an `i` such that
+  the process is in `process_groups[i]`.
+* `broadcast_in_dim(constant(0, element_type(result)), [], type(result))`
```
ghpvnist, line 46: Nit: I think we can just say
```
`broadcast_in_dim(constant(0, element_type(result)), [], type(result))` otherwise.
```
This is also in line with the existing convention (see [collective_permute](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#collective_permute))

rfcs/20231017-collective-broadcast.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Proposed Specification
+
+### collective_broadcast
+
+#### Semantics
+
+Within each process group in the StableHLO process grid, send the value of the
+`operand` tensor from the source process to the target processes and produce a
+`result` tensor. For any process `n` not in any replica group, that process will just
```
ghpvnist, line 33: > For any process `n` not in any replica group, that process will just
return zeros.

This is probably not needed. The formalization is clear that processes not mentioned in the "target" of replica group will be zeroed. Plus the sentence introduces `n` which is not used elsewhere in the spec.

rfcs/20231017-collective-broadcast.md
ghpvnist: `channel_id <= 0 ` means that this attribute is not provided (See [StablehloAttrs.td](https://github.com/openxla/stablehlo/blob/654c202babbab3f4bd50cdd276de02dc17ecc358/stablehlo/dialect/StablehloAttrs.td#L113))

rfcs/20231017-collective-broadcast.md
ghpvnist: Good point. Actually, the convention in the spec is to have example with `handle = 0`.

@chaserileyroberts Can you update the spec with this example instead and delete the other two?

We are trying to keep only one example in the spec (see [reference_checklist.md](https://github.com/openxla/stablehlo/blob/main/docs/reference_checklist.md)) and additional ones linked to a test file that the interpreter can run, so these tests will be used some day as long as we have the paper trail for now. I know this type of feedback is more appropriate for the PR, but it would be nice to have the PR merged with no additional feedback.

Proposed test:
```
// num_replicas: 4
// num_partitions: 1
// %operand@(0, 0): [[1, 2]]
// %operand@(1, 0): [[3, 4]]
// %operand@(2, 0): [[5, 6]]
// %operand@(3, 0): [[7, 8]]
%result = "stablehlo.collective_broadcast"(%operand) {
  replica_groups = dense<[[2, 1]]> : tensor<1x2xi64>,
  channel_handle = #stablehlo.channel_handle<handle = 0, type = 0>
} : (tensor1x2xi64>) -> tensor<1x2xi64>
// %result@(0, 0): [[0, 0]]
// %result@(1, 0): [[5, 6]]
// %result@(2, 0): [[5, 6]]
// %result@(3, 0): [[0, 0]]
```

@changm I hope this answers your question!

rfcs/20231017-collective-broadcast.md
ghpvnist: I see, I was thinking that `[2,2]`, which is a no-op, is okay to have, but it would be good to just not allow this in the first place and simplify the constraints.

rfcs/20231017-collective-broadcast.md
```diff
@@ -0,0 +1,109 @@
...
+
+#### Outputs
+
+| Name     | Type   | Constraints |
+|----------|--------|-------------|
+| `result` | tensor | (C3)        |
+
+#### Constraints
+
+* (C1) is_unique(replica_groups).
```
ghpvnist, line 66: Nit: \`is_unique(replica_groups)\`.

rfcs/20231017-collective-broadcast.md
chaserileyroberts: Added

rfcs/20231017-collective-broadcast.md
chaserileyroberts: Done

rfcs/20231017-collective-broadcast.md
chaserileyroberts: Removed this example and replaced it with the one you provided.

rfcs/20231017-collective-broadcast.md
chaserileyroberts: Done.

rfcs/20231017-collective-broadcast.md
chaserileyroberts: Removed.

rfcs/20231017-collective-broadcast.md
chaserileyroberts: Added.

rfcs/20231017-collective-broadcast.md
chaserileyroberts: > // num_replicas: 4
> // num_partitions: 1

Are we not sure it's not reversed here? My examples before where with 4 partitions and 1 replica.

ghpvnist: `channel_id = handle = 0` in the above example, so we'd use `cross_replica` instead of `cross_partition`, so we'd want to have multiple replicas here instead.

rfcs/20231017-collective-broadcast.md
```diff
@@ -0,0 +1,87 @@
...
+* (C2) `0 <= replica_groups < N` where `N` is defined as:
+  * `num_replicas` if `cross_replica` is used.
+  * `num_partitions` if `cross_partition` is used.
+* (C3) `type(result) = type(operand)`.
+
+#### Examples
+
+```mlir
+// num_replicas: 4
+// num_partitions: 1
```
chaserileyroberts, line 74: @ghpvnist I left this the way you provided, but I'm still unsure if this is correct. My previous examples had 4 partitions and 1 replica, but this one has that explicitly swapped throughout.

rfcs/20231017-collective-broadcast.md
chaserileyroberts: Ok cool, just making sure.

ghpvnist: For clarification to whoever else may be reading this thread, if you take a look at (C2), if we use `handle = 0` and had `num_replicas = 1`, then the values of `replica_groups` can only be 0. If `handle > 0`, and `num_partitions = 4`, the `replica_groups` can be in range [0, 4).

rfcs/20231017-collective-broadcast.md
changm: What do you think about making it an error then? 

Also, can you update this doc to say channel_id <= 0 means this attribute isn't provided?

rfcs/20231017-collective-broadcast.md
```diff
@@ -0,0 +1,109 @@
...
+  if process is never found in `process_groups`.
+
+
+#### Inputs
+
+| Label | Name                    | Type                                                             | Constraints |
+|-------|-------------------------|------------------------------------------------------------------|-------------|
+| (I1)  | `operand`               | tensor                                                           | (C3)        |
+| (I2)  | `replica_groups`        | variadic number of 1-dimensional tensor constants of type `si64` | (C1-C2)     |
+| (I3)  | `channel_id`            | constant of type `si64`                                          |             |
```
changm, line 54: If channel_id should be 0, maybe make this uint64 then?

ghpvnist: Most ops in StableHLO have signed attributes (the three exceptions being `replica_id`, `partition_id`, `rng_bit_generator`). Using signed attributes has been the convention in the spec, so it makes sense to give `channel_id` similar treatment here since other collective ops with `channel_id` are also signed. Now, changing the attribute to unsigned universally over other ops with this attribute is certainly possible, but it will have compatibility implications as we are changing the spec (just like this RFC). Since we are not getting much out of it compared to the efforts required, I propose we keep it for now.

rfcs/20231017-collective-broadcast.md
ghpvnist: > What do you think about making it an error then?
> Also, can you update this doc to say channel_id <= 0 means this attribute isn't provided?

I think it's the other way around. `channel_id` is not needed for cross replica communication (e.g. happening within an accelerator, and can communicate directly with other cores) whereas it is required for cross partition (e.g. happening among multiple accelerators and thus require a unique "channel" to communicate amongst each other). So `channel_id = 0` is not because this attribute is not provided, but rather because we are doing cross replica communication, so `channel_id` is not needed and it defaults to a value like 0.

However, whether negative ids are allowed, that's a broader question outside the scope of this RFC. IIRC `channel_id` being signed is a legacy inherited from MHLO.

rfcs/20231017-collective-broadcast.md
changm: sounds good, thanks for the explanation.

rfcs/20231017-collective-broadcast.md
changm: Ack, thanks for the thorough explanation :)

rfcs/20231017-collective-broadcast.md
changm: small nit: My understanding of this thread was that an example would be added with handle = 0. That doesn't seem to be the case or was the idea that the handle = 0 case would be in the reference interpreter test cases? Doesn't block either way. Thanks!

andydavis1:
Thank you for the RFC. The current set of collective operations are currently used in the SPMD context, where each device is running the same program. With the proposed collective_broadcast, do you have an MPMD use case for this, or were you expecting other devices which were not broadcasting to produce an empty buffer as a result of the broadcast?

rfcs/20231017-collective-broadcast.md
ghpvnist: If I'm understanding correctly, yes the idea is to have handle = 0 (cross_replica) as the example. The reference interpreter would have this test plus other cases (handle > 0, i.e. cross_partition). Let me know if you meant something else.

chaserileyroberts:
>With the proposed collective_broadcast, do you have an MPMD use case for this, or were you expecting other devices which were not broadcasting to produce an empty buffer as a result of the broadcast

No, the idea is for this to still be targeting exclusively SPMD. On the cuda side, the goal is to have this op lower to the exact same `nccl.Broadcast` operation on all devices. Devices that are not in any `replica_group` of the SPMD op return zeros, as per the specification and latest example.

chaserileyroberts:
To give a very concrete example of how this operation could be used in a realistic SPMD setting, here is how this op could be used in a realistic context. Here, we have an example of how we could implement the 2D pgemm algorithm [summa](https://www.netlib.org/lapack/lawnspdf/lawn96.pdf) in JAX once this spec is in place.  
```python
@partial(shard_map, mesh=Mesh(...), 
    in_specs=(P('x', 'y'), P('x', 'y')), out_specs=P('x', 'y'))
def summa_matrix_multiply(a, b):
  for i in range(N):
     abcast = pcollective_broadcast(a, 'y', i)
     bbcast = pcollective_broadcast(b, 'x', i)
     if i == 0:
        c = abcast @ bbcast
     c += abcast @ bbcast
  return  c
  ```

rfcs/20231017-collective-broadcast.md
GleasonK: I believe sufficient notice has been given on email chains / chat threads / etc. HLO internal/external stakeholders are aware that this RFC also applies to a new HLO opcode. I will also make sure that the proper reviewers are tagged on the PR which implements the HLO.

rfcs/20231017-collective-broadcast.md
GleasonK: Sounds like this feedback is mostly resolved / non-blocking. Resolving conversation, please unresolve if I am mistaken!

GleasonK (approved):
RFC Approved. I'll send a follow-up with markdownlint fixes.

chaserileyroberts:
🥳 🎉 

Thanks again Kevin for getting this over the finish line! I'll get to work on getting this implemented in JAX and cuda xla backend.

Issue #1584, opened 2023-06-05
Adopt properties for StableHLO dialects


joker-eph:
### Request description

Upstream has now an option to assist in the migration to properties. You probably wanna to try to switch before upstream changes the default and deprecate the current settings.

In general the switch is fairly trivial, here is an example of how it was done in OpenAI Triton:

- https://github.com/openai/triton/pull/1684
- https://github.com/openai/triton/pull/1696

Some extra cleanup that is worthwhile is to replace all uses of getDictionaryAttr to more targeted APIs, this one is now quite costly.

burmako:
"Some extra cleanup that is worthwhile is to replace all uses of getDictionaryAttr to more targeted APIs, this one is now quite costly". Yeah, this is exactly why we haven't yet upgraded. StableHLO <=> MHLO and StableHLO <=> VHLO is based on getAttrs ([example](https://github.com/tensorflow/mlir-hlo/blob/78fac553dbcd2282cfca8a5fb978bc4a76f5a74d/mhlo/transforms/stablehlo_legalize_to_hlo/stablehlo_legalize_to_hlo.cc#L244-L253)), and we haven't yet had the time to think what to do instead.

"You probably wanna to try to switch before upstream changes the default and deprecate the current settings". Thank you for thinking about us! What timeline do you expect for deprecating and stopping to support old-style attributes?

joker-eph:
I don't have a timeline right now, the first milestone will be switching the default, hopefully before the end of the month. Then it'll likely take a few months at least to hammer down everything.

GleasonK:
Nice! I'll take a look at this in the near future. Will post in the discord if I hit any issues along the way!

GleasonK:
Adding compatibility label since VHLO will need to be migrated as well

PR #1827, opened 2023-11-07
Migrate VHLO to use properties

GleasonK:
[MLIR Properties](https://mlir.llvm.org/OpenMeetings/2023-02-09-Properties.pdf) are the new standard for non-discardable attributes. In the near future, all dialects must migrate to use properties by default. This PR migrates VHLO first, which is an important step since we need to ensure:
1. Backward compatibility: VHLO with properties can read older portable artifacts
2. Forward compatibility: VHLO with properties can generate portable artifacts that is readable in previous versions

(1) Had no issues, after migrating VHLO to properties, only the FileCheck lines needed for the following:
- Attrs print using `{...}` syntax and properties print using `<{...}>. 
- Attrs print after regions, properties print before regions

Note no changes to `stablehlo_legalize_to_vhlo.0_9_0.mlir.bc` for example, meaning we can still read old bytecode files and it deserializes into properties successfully.

(2) Worked fine for all versions of StableHLO **except for 0.14.0**. Note again that there were no changes to `stablehlo_legalize_to_vhlo.0_9_0.mlir.bc`, which means that bytecode files generated at HEAD with properties and targeting 0.9.0 generated _identical_ bytecode files. However, [MLIR Bytecode format v5+](https://github.com/llvm/llvm-project/blob/2e6f722b88e42bba72af5bafd8cc1e37e460d978/mlir/include/mlir/Bytecode/Encoding.h#L48) generates different bytecode for dialects that use properties, and property definitions must be present for deserialization of these. This means back-deployment from `HEAD->0.14.0` breaks, since 0.14.X (pre-properties) doesn't have the property struct definitions needed for deserialization.

Given that backward compatibility is fully supported, (2) is resolved by changing 0.14.0 to target bytecode v4, which serializes attr-dict (can be deserialized by pre/post-property VHLO), and in addition bumping the StableHLO version to 0.15.0 which uses the latest bytecode format version v6. This leads to full forward/backward compatibility with no temporary breakages.

Part of #1584


stablehlo/dialect/Version.cpp
```diff
@@ -63,7 +63,8 @@ FailureOr<int64_t> Version::getBytecodeVersion() const {
   if (*this < Version(0, 10, 0)) return 0;
   if (*this < Version(0, 12, 0)) return 1;
   if (*this < Version(0, 14, 0)) return 3;
-  if (*this <= getCurrentVersion()) return 5;
+  if (*this < Version(0, 15, 0)) return 4;
```
burmako, line 66: I think this needs a comment that explains that this used to be a 5, with an explanation why this change was made.

stablehlo/dialect/VhloDialect.td
```diff
@@ -32,11 +32,12 @@ def VHLO_Dialect : Dialect {
       0.11.0: Introduce `f8E4M3B11FNUZ` type.
       0.12.0: MLIR bytecode version 1 => 3.
       0.14.0: MLIR bytecode version 3 => 5.
```
burmako, line 34: This needs to be updated as well?

burmako (changes requested):

ghpvnist (approved):
LGTM with one question. Should we tag this version separately from the integrates?

GleasonK:
> LGTM with one question. Should we tag this version separately from the integrates?


We can, wouldn't call it a requirement since the next regular integrate will:
1. Import these changes to openxla/xla
2. Tag StableHLO 0.15.0 
3. Bump revision to 0.15.1

PR #1828, opened 2023-11-07
Serialize MLIR type when instrumenting reference interpreter values

penagos:
Recently, we introduced a way to extract intermediate tensor state from the StableHLO reference interpreter (#1784) for instrumentation/debugging purposes. As part of this instrumentation process, the interpreter will create an `index.csv` metadata file which contains all serialized tensor paths and uniquely identifying `probe_id` values in the form of:

```
probe_id,/some/absolute/path/to/numpy_0.npy
...
```

In the event that an `interpreter.probe` instruction is executed more than once (i.e. due to being within a loop), there could be `N` entries with the same `probe_id` value.

This current schema however does not serialize the `mlir::TensorType` of the data. The inclusion of this type information can make it easier for post processing tools/scripts to better interpret the metadata file, without needing to load the accompanying `npy` file into memory to extract size/type information. With these changes, the serialized metadata file format now becomes:

```
probe_id,tensor<T>,/some/absolute/path/to/numpy_0.npy
...
```

Where `tensor<T>` is the type string produced by `mlir::debugString` (i.e. `tensor<1x2xf32>`, etc). Additionally, the `expect_serialized_eq` check dialect operation can now perform a stricter check, by also locking down the serialized type vs the expected type.

### DIFF
--- a/stablehlo/tests/CheckOps.cpp
+++ b/stablehlo/tests/CheckOps.cpp
@@ -33,13 +33,15 @@ namespace stablehlo {
 namespace check {
 namespace {
 
-// Splits a given string into parts, using the given delimiter.
-template <typename OutputIterator>
-void splitString(const std::string &line, char delimiter,
-                 OutputIterator output) {
-  std::istringstream buffer(line);
-  std::string part;
-  while (std::getline(buffer, part, delimiter)) *output++ = part;
+using SerializedTensorMetadata =
+    std::pair</*type=*/std::string, /*path=*/std::string>;
+
+SerializedTensorMetadata extractMetadata(StringRef line) {
+  // Parse a CSV record in the form of: probe_id,mlir_type,serialized_path
+  constexpr int kNumFields = 3;
+  SmallVector<StringRef, kNumFields> fields;
+  line.split(fields, ',', kNumFields);
+  return std::make_pair(/*type=*/fields[1].str(), /*path=*/fields[2].str());
 }
 }  // namespace
 
@@ -95,8 +97,6 @@ llvm::Error evalExpectEqOp(const Tensor &lhs, const Tensor &rhs) {
 // Fetch a previously serialized MLIR type and data filepath given a `probeId`
 // and a `probeDir` for a specified `iteration` value from an `index.csv`
 // metadata file. If no data is found, returns an error.
-using SerializedTensorMetadata =
-    std::pair</*type=*/std::string, /*path=*/std::string>;
 static llvm::ErrorOr<SerializedTensorMetadata> getSerializedTensorMetadata(
     StringRef probeId, StringRef probeDir, uint32_t iteration) {
   if (probeDir.empty()) return llvm::errc::invalid_argument;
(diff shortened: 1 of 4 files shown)


stablehlo/tests/CheckOps.cpp
```diff
@@ -97,34 +109,45 @@ static llvm::ErrorOr<std::string> getSerializedTensorPath(StringRef probeId,
...
   }
 
   return llvm::errc::bad_address;
 }
 
 llvm::Error evalExpectSerializedEqOp(const Tensor &expected, StringRef probeId,
                                      StringRef probeDir, uint32_t iteration) {
-  auto serializedFilePathOrError =
-      getSerializedTensorPath(probeId, probeDir, iteration);
+  auto serializedMetadataOrError =
```
GleasonK, line 130: `serializedMetadata` - I didn't catch this last time, but we tend to avoid `OrFailure` / `OrError` names

stablehlo/tests/CheckOps.cpp
```diff
@@ -97,34 +109,45 @@ static llvm::ErrorOr<std::string> getSerializedTensorPath(StringRef probeId,
...
 
   for (uint32_t match = 0; metadataFile >> line && match <= iteration;
        ++match) {
     auto pos = line.find(probeId);
 
-    if (pos != std::string::npos && match == iteration)
-      return line.substr(pos + probeId.size() + 1);
+    if (pos != std::string::npos && match == iteration) {
+      // Parse the record in the form of: probe_id,mlir_type,serialized_path
+      splitString(line, ',', fields.begin());
```
GleasonK, line 120: We don't need a generic string splitting method. Let's rename this to:

```
if (pos != std::string::npos && match == iteration)
    return extractMetadata(line);
```

And can use `llvm::StringRef` and its built in `split` method to implement this.

stablehlo/tests/CheckOps.cpp
```diff
@@ -97,34 +109,45 @@ static llvm::ErrorOr<std::string> getSerializedTensorPath(StringRef probeId,
...
-      numpy::deserializeTensor(*serializedFilePathOrError, expected.getType());
+      numpy::deserializeTensor(serializedPath, expected.getType());
 
   if (!tensorOrError)
     return llvm::createStringError(tensorOrError.getError(),
                                    "Failed to verify serialized tensor.");
 
+  if (type != debugString(expected.getType()))
+    return llvm::createStringError(llvm::errc::invalid_argument,
+                                   "Type mismatch.");
```
GleasonK, line 149: Include the types in the print out?

GleasonK (changes requested):

stablehlo/tests/CheckOps.cpp
penagos: There's a couple of other `OrError` variable names in other files I introduced in a prior PR. I'll create a separate PR for those updates as well, and update the ones in the function already modified here in this PR.

stablehlo/tests/CheckOps.cpp
```diff
@@ -31,6 +31,19 @@ limitations under the License.
...
+
+using SerializedTensorMetadata =
+    std::pair</*type=*/std::string, /*path=*/std::string>;
+
+SerializedTensorMetadata extractMetadata(StringRef line) {
+  // Parse a CSV record in the form of: probe_id,mlir_type,serialized_path
+  constexpr int kNumFields = 3;
+  SmallVector<StringRef, kNumFields> fields;
+  line.split(fields, ',', kNumFields);
+  return std::make_pair(/*type=*/fields[1].str(), /*path=*/fields[2].str());
```
GleasonK, line 44:
```
llvm::ErrorOr<SerializedTensorMetadata> extractMetadata(...) {
  //...
  if (fields.size() < 3) return error(...);
}
```

GleasonK (approved):

PR #1831, opened 2023-11-08
[build] create initial CMake export configuration

ashay:
Prior to this patch, it seems that consumers of StableHLO dialects and
passes are required to build StableHLO from source.  More precisely,
such consumers cannot point to a StableHLO build tree or install tree.
This is because StableHLO's build and install trees do not create an
export configuration (StablehloConfig.cmake) that consumers can refer
to, by calling `find_package(StableHLO)`.

This patch adds a CMake script to create a basic export configuration
that lists the library targets of StableHLO, along with their
dependencies.

Since these library targets are created using the `add_mlir_library()`
CMake function, these targets get added to MLIRTargets and MLIR_EXPORTS.
A better alternative is to add the names of these targets to StableHLO's
list of exports that is separate from MLIR's, but doing so requires a
somewhat invasive change of replacing all `add_mlir_library()` with a
wrapper that adds the target names to a StableHLO-specific list of
exports.

### DIFF
--- a/cmake/modules/CMakeLists.txt
+++ b/cmake/modules/CMakeLists.txt
@@ -1,0 +1,69 @@
+# Since we just use MLIR's cmake functions (add_mlir_library, etc.),
+# MLIR_EXPORTS would have the necessary targets to export.
+get_property(STABLEHLO_EXPORTS GLOBAL PROPERTY MLIR_EXPORTS)
+
+set(STABLEHLO_CONFIG_CMAKE_DIR "${CMAKE_BINARY_DIR}/lib/cmake/stablehlo")
+
+export(TARGETS ${STABLEHLO_EXPORTS}
+  NAMESPACE Stablehlo::
+  FILE "${STABLEHLO_CONFIG_CMAKE_DIR}/StablehloTargets.cmake"
+)
+
+# Set variables and configure StablehloConfig.cmake for the _build_ tree.
+set(STABLEHLO_CONFIG_INCLUDE_DIRS
+  "${STABLEHLO_SOURCE_DIR}"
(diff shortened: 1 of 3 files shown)


cmake/modules/CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
...
+  "${CMAKE_CURRENT_SOURCE_DIR}/StablehloConfig.cmake.in"
+  "${STABLEHLO_CONFIG_CMAKE_DIR}/StablehloConfig.cmake"
+  @ONLY
+)
+
+export(EXPORT MLIRTargets
+  NAMESPACE Stablehlo::
+  FILE "${PROJECT_BINARY_DIR}/StablehloTargets.cmake"
+)
+
```
brnorris03, line 28: The targets were already exported on lines 7--10, so this export is not needed and results in an error about duplicate imports in packages using the exported targets.

cmake/modules/CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
...
+  "${CMAKE_CURRENT_BINARY_DIR}/CMakeFiles/StablehloConfig.cmake"
+  @ONLY
+)
+
+# Specify sources and destinations for install files.
+install(EXPORT MLIRTargets
+  NAMESPACE Stablehlo::
+  FILE StablehloTargets.cmake
+  DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/stablehlo"
+)
```
brnorris03, line 49: The `MLIRTargets` apparently don't include the tools (`stablehlo-opt`, `stablehlo-translate`, etc), so they don't get installed

PR #1840, opened 2023-11-13
Allow unsafe probe_id values to be serialized properly to disk

penagos:
This PR builds off of the recently submitted changes in #1784. Prior to this change, the value of `probe_id` and an internal execution counter is used to derive the filename for serializing probed tensor values (i.e. `<probe_output_dir>/<probe_id>_<execution_count>.npy`). This current design has the following disadvantage:

1. This can be unsafe as no filename sanitation is performed, relying on the compiler (or other tooling) to produce friendly `probe_id` values. Instead of having this implicit assumption on the instrumented StableHLO program, this change uses internally generated serialization filenames by using a strictly increasing counter.

This change proposes that the serialization filename now be derived as follows:
```
<probe_output_dir>/probe<i>.npy
```

Where `i` is a uniquely increasing positive `int64_t`.

### DIFF
--- a/stablehlo/reference/InterpreterOps.cpp
+++ b/stablehlo/reference/InterpreterOps.cpp
@@ -42,9 +42,6 @@ namespace stablehlo {
 namespace interpreter {
 namespace {
 
-// A strictly increasing counter to uniquely serialize instrumented probe values
-static int64_t serializedProbeId = 0;
-
 // Appends a new line item to an instrumentation metadata file, `index.json` in
 // the form: `probeId,probeOutputDir/filename`.
 llvm::Error writeProbeMetadata(StringRef probeId, StringRef filename,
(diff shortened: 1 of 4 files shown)


stablehlo/reference/InterpreterOps.cpp
```diff
@@ -42,6 +42,9 @@ namespace stablehlo {
 namespace interpreter {
 namespace {
 
+// A strictly increasing counter to uniquely serialize instrumented probe values
+static int64_t serializedProbeId = 0;
```
GleasonK, line 46: Let's avoid static values. They'll make testing more difficult, among other issues.

How about we store this either in InterpreterConfig or DefaultInterpreterFallback, and pass it to probe op? Probably default fallback is better since it is related to an interpreter dialect op.

GleasonK (changes requested):

stablehlo/reference/InterpreterOps.cpp
```diff
@@ -193,17 +193,17 @@ SmallVector<InterpreterValue> evalRunParallelOp(
...
 
-  // To properly support loops, append a suffix denoting how many times this
-  // specific probe_id has executed.
-  const int32_t numTimesExecuted = ++probeIterations[probeId];
-
+  // Use an increasing unique integer to write to disk to avoid any odd file
+  // names as a result of unsafe probe_id values.
   llvm::sys::path::append(
-      filepath, probeId + "_" + std::to_string(numTimesExecuted) + ".npy");
+      filepath, "probe" + std::to_string(serializedProbeFileId) + ".npy");
```
GleasonK, line 206: Is this preferable to enforcing a grammar on probe IDs?

Can add [let hasVerifier = 1](https://github.com/openxla/stablehlo/blob/e5ad51715a11721c78b6748ab5de7945df24b1b8/stablehlo/reference/InterpreterOps.td#L80C1-L80C1) and implement a [verify](https://github.com/openxla/stablehlo/blob/e5ad51715a11721c78b6748ab5de7945df24b1b8/stablehlo/reference/InterpreterOps.cpp#L87) method which enforces `[A-Za-z_]+` for probe name.

There's something kind of nice about the probe filename matching the probe_id

penagos: This is somewhat tangentially related to the conversation in https://github.com/openxla/stablehlo/pull/1841#discussion_r1393341371.

In short, I'm hesitant to enforce a grammar here as different compilers may populate MLIR location data differently, which could be used by the StableHLO probe pass to populate `probe_id` values in the proposed PR in #1841.

GleasonK (approved):

PR #1841, opened 2023-11-13
Introduce a compiler pass to instrument modules

penagos:
This PR adds a compilation pass to insert the recently introduced `interpreter.probe` instrumentation op (#1784) into StableHLO programs. The pass is registered as `stablehlo-instrument` in the `stablehlo-opt` target (open to other naming suggestions).

The pass will instrument all SSA values which are not the result of a `stablehlo.constant` operation (as there is no benefit in instrumenting unchanging values). The pass will attempt to use MLIR named location data to derive a suitable `probe_id`, and fallback to an increasing unsigned integer otherwise. For example, given the following simple function:

```
func.func @main(%arg0: tensor<1x2xi32>, %arg1: tensor<1x2xi32>) -> tensor<1x2xi32> {
  %0 = stablehlo.constant dense<[0, 0]> : tensor<1x2xi32>
  %1 = stablehlo.add %0, %arg0 : tensor<1x2xi32> loc("add1")
  %2 = stablehlo.add %1, %arg1 : tensor<1x2xi32>
  func.return %2 : tensor<1x2xi32>
}
```

This pass would instrument it as follows:

```
func.func @main(%arg0: tensor<1x2xi32>, %arg1: tensor<1x2xi32>) -> tensor<1x2xi32> {
  %0 = stablehlo.constant dense<[0, 0]> : tensor<1x2xi32>
  %1 = stablehlo.add %0, %arg0 : tensor<1x2xi32> loc("add1")
  %2 = interpreter.probe %1, probe_id = "add1" : tensor<1x2xi32>
  %3 = stablehlo.add %2, %arg1 : tensor<1x2xi32>
  %4 = interpreter.probe %3, probe_id = "1" : tensor<1x2xi32>
  func.return %4 : tensor<1x2xi32>
}
```

Where any operation without an explicit MLIR named location defaults to a unique, increasing `probe_id` integer value (i.e. in the example, the second add op). Note that as described above, constant operations are not instrumented. The pass walks all operations in the `mlir::ModuleOp`, so it will instrument values inside control flow structures as well (while, if, etc).


stablehlo/transforms/StablehloProbeInstrumentation.cpp
```diff
@@ -0,0 +1,113 @@
...
+}
+
+void StablehloProbeInstrumentationPass::runOnOperation() {
+  ModuleOp module = getOperation();
+
+  OpBuilder builder(module);
+  SmallVector<Value> valuesToInstrument;
+
+  module.walk([&](Operation* op) {
+    for (Value result : op->getResults()) valuesToInstrument.push_back(result);
```
GleasonK, line 88: Can we move `shouldInstumentValue` -> `shouldInstrumentOp`?

```
if (!shouldInstrumentOp(op)) return WalkResult::advance();
```

Can be implemented using `op->isa<ConstantOp>()` instead of needing to get the defining op from Value.

stablehlo/transforms/StablehloProbeInstrumentation.cpp
```diff
@@ -0,0 +1,113 @@
...
+#include "mlir/IR/Value.h"
+#include "mlir/Support/LLVM.h"
+#include "stablehlo/dialect/StablehloOps.h"
+#include "stablehlo/reference/InterpreterOps.h"
+#include "stablehlo/transforms/Passes.h"
+
+namespace mlir {
+namespace stablehlo {
+
+#define GEN_PASS_DEF_STABLEHLOPROBEINSTRUMENTATIONPASS
```
GleasonK, line 28: Lets think about pass name / command line flag.

The relation should be:

```
StablehloDoSomethingPass -->  --stablehlo-do-something
```

In this case it would be

```
StablehloProbeInstrumentationPass --> --stablehlo-probe-instrumentation
```

And I don't think that's a proper flag name so perhaps we should rename this pass as a verb.

```
StablehloInsertProbesPass --> --stablehlp-insert-probes
StablehloInstrumentProbePass --> --stablehlo-instrument-probe
StablehloInstrumentWithProbePass --> --stablehlo-instrument-with-probe
StablehloProbeInstrumentModulePass --> --stablehlo-probe-instrument-module
```

Let me know if you have any other ideas for this, but I think we should land on something more declarative

GleasonK (changes requested):
I think this is fine to have in the centralized repo since it still falls under support for the RFC of "Extract Tensor Values". I would recommend adding a notification to the openxla-discuss post of the current status of probe / ongoing work.

I'll also want to think about if `transforms` is where this should go, or if we should have a separate directory of transforms used by the reference interpreter.. I'll keep thinking, I'm leaning towards keeping it as-is.

stablehlo/tests/stablehlo_probe_instrumentation.mlir
```diff
@@ -0,0 +1,75 @@
+// RUN: stablehlo-opt --stablehlo-instrument --split-input-file --verify-diagnostics %s | FileCheck %s
+
+// CHECK-LABEL: func @instrument_basic_no_location
+func.func @instrument_basic_no_location(%arg0: tensor<1x2xi32>, %arg1: tensor<1x2xi32>) -> tensor<1x2xi32> {
+  // CHECK: [[RESULT:%.*]] = interpreter.probe %0, probe_id = "1" : tensor<1x2xi32>
```
ghpvnist, line 5: I'm wondering about the usability of this default `probe_id`. Do you think there will be a need to `ctrl+f` default `probe_id`? If so, prefixing it with `id_` like `id_1` may be an option to distinguish them from other digits in the exported file.

Maybe it's not necessary if we can `ctrl+f` the `probe_id` field instead. Thoughts?

ghpvnist (changes requested):

stablehlo/transforms/StablehloProbeInstrumentation.cpp
penagos: I like the idea of dropping `Instrumentation` as it does sound redundant given `Probe`, and naming the pass `StablehloProbePass`. Then this could fall in line with existing pass nomenclature to yield `--stablehlo-probe` as @GleasonK mentioned earlier.

stablehlo/tests/stablehlo_probe_instrumentation.mlir
penagos: Hm, I don't have a strong preference either way. But maybe changing it to `probe1` vs `probe` in this case better aligns with the proposed changes in #1840?

ghpvnist: Sounds good!

stablehlo/transforms/StablehloProbeInstrumentation.cpp
penagos: Also, I may open another PR to rename existing `probeInstrumentation` nomenclature due to other recent PRs, to better unify the pass name here with `InterpreterConfiguration` settings.

stablehlo/transforms/StablehloProbeInstrumentation.cpp
```diff
@@ -0,0 +1,113 @@
...
+          StablehloProbeInstrumentationPass> {
+ public:
+  explicit StablehloProbeInstrumentationPass() = default;
+  void runOnOperation() override;
+
+ private:
+  // Create a uniquely identifying probe ID for the given value being
+  // instrumented. Attempt to use named MLIR location data, otherwise use an
+  // increasing ID.
+  std::string getLocationNameOrUniqueId(Location location);
```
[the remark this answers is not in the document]
penagos: > I.e. don't want models being written specially for probing which then ruins the general debug experience which uses file-line-column etc.

Completely agree. If we were to require this, it would not make the debugging experience work OOB and make it much less user friendly. Perhaps we could modify the pass description and/or `instrument.probe` op descriptions to indicate this (that location data is optional)? Or perhaps make it a pass option?

> I'd rather that debuginfo (location) doesn't have semantic meaning to ProbeOp. Can we just use the incrementing number `probe1`, `probe2`..

Initially, the thought behind this was to allow the instrumented data produced by the StableHLO reference interpreter to better align with other runtimes, as I imagine ultimately it will be possible to roundtrip StableHLO programs to other representations (such as TFLite, etc) given the portability goal of StableHLO.

For example, there was a [commit](https://github.com/tensorflow/tensorflow/commit/84101d7daef2b49add3da3418076547ad7333c3e) earlier this year in Tensorflow Lite to allow subgraphs to have their intermediate tensor values serialized to disk, for which tensor names could be ostensibly used for deriving unique identifiers (i.e. `probe_id`). My understanding is that the existing [flatbuffer_translate](https://github.com/tensorflow/tensorflow/blob/6f50f70c873d60a84baf35957190d8e8ccff5dd6/tensorflow/compiler/mlir/lite/tests/mlir2flatbuffer/flex_op_with_f64.mlir#L1) tool will use these values to derive MLIR location data.

Suppose if there were a way to convert TFLite MLIR into StableHLO, and if the StableHLO probe instrumentation pass were to be run as part of a larger compiler pass pipeline, it may make it a lot harder to try and match the instrumentation locations without existing MLIR location data, as the compiler may shuffle, fuse or remove ops. I think the location data could also make it easier to discern where an instrumented value was produced in a model (i.e. a `convolution` vs `matmul`, etc). Ultimately, named MLIR location data wouldn't be required for instrumentation. It'd just be an option for maximizing usability with other runtimes.

Thoughts?

stablehlo/transforms/StablehloProbeInstrumentation.cpp
```diff
@@ -0,0 +1,113 @@
...
+  // After all values have been instrumented, update any downstream uses of old
+  // values with the new instrumentation values
+  void updateAffectedOperations();
+
+  // Map of values which have been instrumented.
+  DenseMap<Value, Value> instrumentedValues;
+
+  // Strictly increasing counter to uniquely identify probe operations when MLIR
+  // location data is not available.
+  unsigned int probeId = 0;
```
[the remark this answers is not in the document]
penagos: I updated the methods to take an `unsigned int& id`, but I think readability wise that might be a bit worse?

stablehlo/transforms/StablehloProbeInstrumentation.cpp
```diff
@@ -0,0 +1,113 @@
...
+  SmallVector<Value> valuesToInstrument;
+
+  module.walk([&](Operation* op) {
+    for (Value result : op->getResults()) valuesToInstrument.push_back(result);
+  });
+
+  for (Value value : valuesToInstrument)
+    if (shouldInstrumentValue(value)) instrumentValue(value, builder);
+
+  updateAffectedOperations();
```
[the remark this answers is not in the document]
penagos: Old habit of separating analysis from rewrites; updated.

stablehlo/transforms/Passes.td
```diff
@@ -32,6 +32,19 @@ def StablehloLegalizeToVhloPass : Pass<"stablehlo-legalize-to-vhlo", "ModuleOp">
   let dependentDialects = ["mlir::vhlo::VhloDialect"];
 }
 
+def StablehloProbeInstrumentationPass : Pass<"stablehlo-instrument", "ModuleOp"> {
+  let summary = "Inserts probe instrumentation instructions in a StableHLO "
+    "program.";
+  let dependentDialects = ["mlir::stablehlo::interpreter::InterpreterDialect"];
+  let description = [{
+    Walks through a StableHLO program and inserts a probe instrumentation
+    instruction after each operation.
```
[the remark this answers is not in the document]
penagos: Updated pass description to clarify these points (nested regions are supported, constants are not).

stablehlo/transforms/StablehloProbe.cpp
```diff
@@ -0,0 +1,99 @@
+/* Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+   Copyright 2023 The StableHLO Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
```
ghpvnist, line 3: Nit: I think we need a space between line 2 and 3 for consistency.

stablehlo/transforms/StablehloProbe.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+  module.walk([&](Operation* op) {
+    if (shouldProbeOp(*op)) probeOp(*op, probeId, builder);
+
+    return WalkResult::advance();
+  });
+}
+
+bool StablehloProbePass::shouldProbeOp(Operation& op) const {
+  if (isa<ConstantOp>(op)) return false;
```
GleasonK, line 93: Can we also add ops with no results to the list of "should not probe" ? Looks like it would be silently skipped anyway in the `probeOp` function, but makes for a more clear contract. Also could you update the code comments to include skipping ops without results

GleasonK: Specifically thinking of `ReturnOp, TraceOp` etc

stablehlo/transforms/Passes.td
```diff
@@ -32,6 +32,28 @@ def StablehloLegalizeToVhloPass : Pass<"stablehlo-legalize-to-vhlo", "ModuleOp">
...
+    with other runtimes.
+
+    All operations are considered suitable for instrumentation, except constant
+    ops, as they are effectively a no-op. Suitable operations will be
+    instrumented regardless of their level of nesting. That is, operations
+    inside loop/branch regions will also be instrumented.
+    
+    Instrumented operations will have their return values written to disk using
+    the NumPy data format as they are executed. See `interpreter.probe` for
+    additional information on how data is serialized.
```
GleasonK, line 53: There is no mention of how debuginfo is used to set the probe name. If it is going to have meaning for this pass, it should be documented at least.

It looks like the rule is "extracts `probe_id` from `NamedLoc(probe_id@<other_debug_info>)`"

stablehlo/transforms/StablehloProbe.cpp
```diff
@@ -0,0 +1,100 @@
...
+class StablehloProbePass
+    : public impl::StablehloProbePassBase<StablehloProbePass> {
+ public:
+  explicit StablehloProbePass() = default;
+  void runOnOperation() override;
+
+ private:
+  // Create a uniquely identifying probe ID for the given value being
+  // instrumented. Attempt to use named MLIR location data, otherwise use an
+  // increasing ID.
```
GleasonK, line 43: Be a little more specific on how debuginfo is used. Looks like it is before `@` ?

GleasonK (changes requested):
Mostly documentation and nits.

stablehlo/transforms/StablehloProbeInstrumentation.cpp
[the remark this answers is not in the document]
GleasonK: I understand the use case, and given that this is a pass for interpreter testing it's probably OK to use debuginfo, as it will come out explicit in the resulting probed program.

Thinking about this a different way- Let's say I have infra set up to probe a model and have gathered probeinfo already, then the debuginfo (not the structure of the program) of the next export changes and I have a bunch of different names.

If you think this is a potential problem, perhaps we can hide `useDebugInfo` behind an opt-in pass option? If you think this isn't actually invasive then perhaps we're good as-is.

stablehlo/transforms/StablehloProbe.cpp
```diff
@@ -0,0 +1,100 @@
...
+void StablehloProbePass::runOnOperation() {
+  ModuleOp module = getOperation();
+  OpBuilder builder(module);
+
+  // Strictly increasing counter to uniquely identify probe operations when MLIR
+  // location data is not available.
+  unsigned int probeId = 0;
+
+  module.walk([&](Operation* op) {
+    if (shouldProbeOp(*op)) probeOp(*op, probeId, builder);
```
GleasonK, line 86: I had something like this in mind, to avoid the need to share anything by reference, also makes it clear where the value of ID is incremented:

```
if (shouldProbeOp(*op)) {
  for (auto res : op.getResults()) probeValue(res, ++id, builder);
}
```

stablehlo/transforms/StablehloProbeInstrumentation.cpp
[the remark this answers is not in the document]
GleasonK: Left a comment on the new code, can resolve there. I think the reference can/should be avoided.

GleasonK (changes requested):

stablehlo/transforms/StablehloProbeInstrumentation.cpp
GleasonK: Re-opening, sorry to nitpick on names, I just think if a typical person saw `-stablehlo-probe` used in a file they wouldn't be able to describe whats going on.

> I think in this context probe and instrument mean the same thing

In this context one is a verb one is a noun. All of our passes follow a pattern of "noun-verb-noun", ex: `stablehlo-canonicalize-dynamism`, `stablehlo-refine-shapes`, `stablehlo-legalize-to-*`, ... The only outlier is `vhlo-to-version` which is also probably a bad name that should be revisited, `vhlo-legalize-to-version`.

With that said, I'd suggest either use `probe` as a noun referring to probe ob `ProbeOp`:

- `StablehloInjectProbeOpPass, --stablehlo-inject-probe`
- `StablehloInstrumentWithProbePass, --stablehlo-instrument-with-probe`

Or a verb, meaning "to probe the op"

- `StablehloProbeModule --stablehlo-probe-module`
- `StablehloProbeOps --stablehlo-probe-ops`

With that said, I prefer `StablehloInstrumentWithProbe` or `StablehloInjectProbeOpPass`. We don't tend to get super concise on pass names, the same is true of MHLO transformation pass names like `HloLegalizeShapeOpsToStandardPass -hlo-legalize-shapeops-to-standard`

stablehlo/transforms/StablehloProbe.cpp
GleasonK: Also I'm wondering if we need something else, like:

```
if (!shouldProbeOp(*op)) return WalkResult::advance();
for (auto res : op.getResults()) {
  if (shouldProbeValue(res)) probeValue(res, ++id, builder);
}
```

I believe we only want to probe a result if the type is tensor type. I believe pass will produce invalid IR if the type is `!stablehlo.token`, since probe requires tensor operands?

GleasonK (changes requested):

stablehlo/transforms/StablehloProbeInstrumentation.cpp
penagos: I prefer the `StablehloInstrumentWithProbe` over `StablehloInjectProbeOpPass` nomenclature; updated the pass name and flag.

stablehlo/transforms/Passes.td
penagos: Updated this and relevant comments in the pass impl.

stablehlo/transforms/StablehloProbe.cpp
penagos: > I had something like this in mind, to avoid the need to share anything by reference, also makes it clear where the value of ID is incremented:

I think this will unconditionally increment `id` even if location data is used, slightly tweaked to only increment if location data isn't used.

I also updated additional checks for `TensorType` and updated the LIT test.

ghpvnist (changes requested):

stablehlo/tests/stablehlo_probe_instrumentation.mlir
```diff
@@ -0,0 +1,100 @@
+// RUN: stablehlo-opt --stablehlo-instrument-with-probe="useDebugInfo=true" --split-input-file --verify-diagnostics %s | FileCheck %s
+
```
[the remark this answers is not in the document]
penagos: Since this side effect is unrelated to compiler changes, I'll open a new PR for that specific bug fix. Thanks for catching this!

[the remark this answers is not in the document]
penagos: Created #1864.

stablehlo/transforms/Passes.td
```diff
@@ -32,6 +32,40 @@ def StablehloLegalizeToVhloPass : Pass<"stablehlo-legalize-to-vhlo", "ModuleOp">
   let dependentDialects = ["mlir::vhlo::VhloDialect"];
 }
 
+def StablehloInstrumentWithProbePass : Pass<"stablehlo-instrument-with-probe", "ModuleOp"> {
```
ghpvnist, line 35: Nit: can you move this def above `StablehloLegalizeToVhloPass`?

ghpvnist (approved):
LGTM with one last comment!

stablehlo/transforms/StablehloProbe.cpp
GleasonK: > I think this will unconditionally increment id even if location data is used, slightly tweaked to only increment if location data isn't used.

Is this a bad thing? Should ID be appended to location name as well? I.e. `my_op.1` - it seems plausible that location metadata could use the same `opName` in `opName@something` (given that debug info is whatever you want it to be), leading to non-unique IDs

penagos: That's a good point. There could be cases where the debug location in the original model leads to split ops that share the same `NamedLoc`. Updated the pass description and pass to do the following:

If MLIR location data is to be used (i.e. the `useDebugInfo` pass option is set), the `probe_id` will be computed as follows:

```
probe_id = x.#
```
Where `x` is `NamedLoc(x@...)` and `#` is a positive increasing integer. Otherwise, the probe will be named as `probe#`. I decided to leave an interleaving `.` between the location name and ID to make it easy to extract the named location from the suffix later on if needed.

GleasonK (approved):

PR #1849, opened 2023-11-16
TableGen Headers Included Before They Are Generated

Andrew-Cain:
## TableGen Headers Included Before They Are Generated

**Issue:**

The issue is due to **`StablehloLegalizeToTosa.cpp`** attempting to include a TableGen'd file (**`StablehloEnums.h.inc`**) that hasn't been generated yet, resulting in the following build failure:

```bash
In file included from /stablehlo/stablehlo/conversions/tosa/transforms/StablehloLegalizeToTosa.cpp:25:
/stablehlo/stablehlo/dialect/StablehloOps.h:45:10: fatal error: 'stablehlo/dialect/StablehloEnums.h.inc' file not found
#include "stablehlo/dialect/StablehloEnums.h.inc"
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

**Reproduce:**

To reproduce this issue on Linux or Mac, follow these steps:

```bash
export LLVM_PROJECT_BUILD="/path/to/llvm-project/build"
git clone https://github.com/openxla/stablehlo.git
python3 -m venv stablehlo_venv
source stablehlo_venv/bin/activate
pip install ninja cmake
cd stablehlo
cmake -B build -G "Ninja" \
        -DCMAKE_C_COMPILER=/usr/bin/clang \
        -DCMAKE_CXX_COMPILER=/usr/bin/clang++ \
        -DMLIR_DIR=${LLVM_PROJECT_BUILD}/lib/cmake/mlir \
        -DCMAKE_BUILD_TYPE=Release \
        -DLLVM_ENABLE_ASSERTIONS=ON
cmake --build build -j2
```

**Resolution**:

To resolve this issue, explicitly define the dependency between `StablehloLegalizeToTosa.cpp` and `StablehloEnums.h.inc`. Otherwise it is the luck of the draw which one will compile first.

This fix  defines the StableHLO TableGen'd headers as a dependency to for the for the MLIR library that includes `StablehloLegalizeToTosa.cpp` . This ensures that the stablehlo header files are always generated before the library is compiled.

### DIFF
--- a/stablehlo/conversions/tosa/transforms/CMakeLists.txt
+++ b/stablehlo/conversions/tosa/transforms/CMakeLists.txt
@@ -29,6 +29,7 @@ add_mlir_library(StablehloTOSATransforms
   DEPENDS
   StablehloTOSATransformsPassIncGen
   StablehloTOSAPDLLPatternsIncGen
+  StablehloOpsIncGen
 
   LINK_COMPONENTS
   Core


stablehlo/conversions/tosa/transforms/CMakeLists.txt
```diff
@@ -29,6 +29,7 @@ add_mlir_library(StablehloTOSATransforms
   DEPENDS
   StablehloTOSATransformsPassIncGen
   StablehloTOSAPDLLPatternsIncGen
+  StablehloOpsIncGen
```
GleasonK, line 32: Could you also add `StablehloOps ` to `LINK_LIBS` below? Looks like this is a transitive dependency when including a file from that library, so probably best to explicitly call it our here. All other targets that include `StablehloOpsIncGen` also link `StablehloOps`

GleasonK (changes requested):
Thank you for fixing this!! I have seen occasional build failures sporadically for this but never had the chance to look into it. Always had assumed it was a corrupt build area, surprised it didn't show up more.

Minor nit then will approve/merge

stablehlo/conversions/tosa/transforms/CMakeLists.txt
GleasonK: Actually I'm working on a few minor patches, I'll loop this fix in!

GleasonK (approved):

GleasonK:
> View this [failed invocation](https://github.com/openxla/stablehlo/pull/1849/checks?check_run_id=18765476478) of the CLA check for more information.

Could you follow this link and sign the CLA when you get a chance? Then we can merge this.

Andrew-Cain:
Thanks @GleasonK! Currently in the process of obtaining a corporate CLA. It's proving to be more finicky than I anticipated. I expect it to work out, but according to the troubleshooting guide, the process will take a few days.

Andrew-Cain:
> > View this [failed invocation](https://github.com/openxla/stablehlo/pull/1849/checks?check_run_id=18765476478) of the CLA check for more information.
> 
> Could you follow this link and sign the CLA when you get a chance? Then we can merge this.

@GleasonK CLA signed, should be good to merge :)

PR #1864, opened 2023-11-30
Do not throw error when executing empty StableHLO modules

penagos:
This PR addresses a bug introduced by PR #1797 where as a result of the new interpreter API, running empty StableHLO modules would result in an error `Requested main function not found.` whereas previously, the `stablehlo-translate` utility would not throw any error.

(Found by @ghpvnist in #1841).

### DIFF
--- a/stablehlo/reference/Api.cpp
+++ b/stablehlo/reference/Api.cpp
@@ -106,6 +106,9 @@ class DefaultInterpreterFallback : public InterpreterFallback {
 llvm::ErrorOr<SmallVector<InterpreterValue>> evalModule(
     ModuleOp module, ArrayRef<InterpreterValue> inputs,
     const InterpreterConfiguration &config) {
+  if (module.getOps<func::FuncOp>().empty())
+    return (SmallVector<InterpreterValue>){};
+
   auto mainFunc = getMainFunction(module, config.mainFunction);
   if (!mainFunc) llvm::report_fatal_error("Requested main function not found.");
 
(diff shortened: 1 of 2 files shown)


GleasonK:
I don't have a strong opinion here, delegating to @ghpvnist. Not sure why someone would want to execute an empty model,  but I suppose that given the StableHLO grammar:

```
Program ::= {Func}
```

The empty module is a valid StableHLO program, so maybe this is a change for the better.

stablehlo/reference/Api.cpp
```diff
@@ -106,6 +106,9 @@ class DefaultInterpreterFallback : public InterpreterFallback {
 llvm::ErrorOr<SmallVector<InterpreterValue>> evalModule(
     ModuleOp module, ArrayRef<InterpreterValue> inputs,
     const InterpreterConfiguration &config) {
+  if (module.getOps<func::FuncOp>().empty())
+    return (SmallVector<InterpreterValue>){};
```
ghpvnist, line 110: I think this should fix the build failure: `return {};`

ghpvnist (changes requested):

stablehlo/reference/Api.cpp
penagos: There's some return type ambiguity in doing this, as it's not possible to know if that's an `llvm::Error` or a `llvm::SmallVector`, so it needs some type hint. Updated.

ghpvnist (approved):
Thank you for the fix!

PR #1856, opened 2023-11-22
Integration of collective_broadcast into spec

chaserileyroberts:
This is the first PR for RFC #1809. I did not add an interpreter implementation as @GleasonK specifically asked me to leave that for new staff joining his team.

PR #1869, opened 2023-12-04
Verifier and Type inference changes for reduction operations

sdasgup3:
Implements the specification changes at https://github.com/openxla/stablehlo/pull/1796. 

The PR adds/updates  the verifier and type inference routines for the following ops: `reduce, reduce_window, select_and_scatter, all_reduce, reduce_scatter, scatter`. Please refer to #1796 for the updated constraints which the PR implements. Note the #1796 is going to be merged soon. 

Here are the changes for each operation:

 - reduce 
     - #1796 added a new constraint C8 
     - Updated labels
     - Add positive tests and negative tests verifying reduce_c6  at verify_reduce.mlir  and type inference tests at infrer_stablehlo.mlir, for reduce_c8.

  - reduce_window: 
     - #1796 updated the C16
     - Add positive tests ; negative tests verifying reduce_window_c13  at verify_reduce_window.mlir  and type inference tests at infrer_stablehlo.mlir, for reduce_window_c16.

  - select_and_scatter
     - #1796 added a new constraint C12
     - Updated labels
     - Add positive tests ; negative tests verifying selelct_and_scatter_c10  at verify_select_and_scatter.mlir  and type inference tests at infrer_stablehlo.mlir, for select_and_scatter_c12.

  - scatter
   - #1796 added a new constraint C17
   -   Updated labels
   -  Add positive tests ; negative tests verifying scatter_c15  at verify_scatter.mlir  and type inference tests at infrer_stablehlo.mlir, for scatter_C17.

  - reduce_scatter
     - #1796 added a new constraint C9 
     - This op does not have the type inference supported. We had a trait `SameOperandsAndResultElementType` implementing the outdated constraint. For the new constraint C9, we added a check at `verifyReduceScatterOp`. 
     - Add positive tests; negative tests verifying reduce_scatter_C7  at ops_stablehlo.mlir  and type inference tests at ops_stablehlo.mlir, for reduce_scatter_C9.

  - all_reduce
     - #1796 added a new constraint C7
     - This op implemented an outdated constraint related to type inference using `inferReturnTypeComponentsFromOperands`. For the new constraint C9, we added a trait `InferTensorType` in the tablegen definition of the op. 
     - Updated labels 
     - Add positive tests; negative tests verifying all_reduce_C5  at ops_stablehlo.mlir  and type inference tests at ops_stablehlo.mlir, for all_reduce_C7.


GleasonK:
We need a means of checking if an IR downgrade is possible in terms of constraints. Let me think about how this should happen. Perhaps we can have "CompatibilityConstraintChecks" somewhere. Other option would be to add ReduceOpV2 in VHLO, but for constraint modification we may be able to do better than this.

stablehlo/dialect/TypeInference.h
```diff
@@ -306,15 +310,15 @@ LogicalResult inferRngOp(
...
+                             ValueRange inputs, Region& body,
                              SmallVectorImpl<Type>& inferredReturnTypes);
 
 LogicalResult inferSelectOp(
     std::optional<Location> location, Value pred, Value onTrue, Value onFalse,
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes);
 
 LogicalResult inferSelectAndScatterOp(
-    Value operand, SmallVectorImpl<Type>& inferredReturnTypes);
+    Value operand, Region& body, SmallVectorImpl<Type>& inferredReturnTypes);
```
ghpvnist, line 321: `Region& select` or `Region& scatter` since only the arg types are used to retrieve accumulator types.

ghpvnist (commented):
Reviewed `dialect/` and `reference/`, and glanced over at the constraint naming schemes under `tests/`.

stablehlo/dialect/TypeInference.h
```diff
@@ -306,15 +310,15 @@ LogicalResult inferRngOp(
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes);
 
 LogicalResult inferScatterOp(std::optional<Location> location,
-                             ValueRange inputs,
+                             ValueRange inputs, Region& body,
```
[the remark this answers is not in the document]
sdasgup3: done

stablehlo/dialect/TypeInference.h
sdasgup3: done

stablehlo/dialect/TypeInference.h
```diff
@@ -120,6 +120,10 @@ LogicalResult inferAllToAllOp(
     DenseIntElementsAttr replicaGroups,
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes);
 
+LogicalResult inferAllReduceOp(
+    std::optional<Location> location, Value operand, Region& body,
```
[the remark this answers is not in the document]
sdasgup3: done

ghpvnist (approved):

GleasonK:
Put together https://github.com/openxla/stablehlo/compare/main...GleasonK:stablehlo:reduce-compat

Please port those changes over and implement constraint versioning for these changed ops.

Aside from that, add compatibility tests per: https://github.com/openxla/stablehlo/blob/main/docs/vhlo.md#add--update-unit-tests - including positive and negative tests.

Can see CollectiveBroadcastOp for example: https://github.com/openxla/stablehlo/pull/1856

sdasgup3:
@GleasonK Updated the PR with the suggested changes. PTAL.

stablehlo/dialect/VhloOps.cpp
```diff
@@ -296,5 +298,99 @@ void VhloDialect::printAttribute(Attribute attr, DialectAsmPrinter& os) const {
...
+///////////////////////////
+// These could be migrated to ODS in VhloOps.td if we figured out a better way
+// to represent this sort of constraint in tablegen.
+
+namespace {
+bool checkIfOperandAndResultElementTypesMatch(TypeRange operandTypes,
+                                              TypeRange resultTypes) {
+  SmallVector<ShapedType> inputShapedTypes{
+      llvm::map_range(operandTypes, [](Type t) {
+        return convertTypeToBuiltinForPrint(t).cast<ShapedType>();
```
sdasgup3, line 312: @GleasonK Use of `convertTypeToBuiltinForPrint` might seem awkward as because of its name  "ForPrint". Let me know if we should rename it to `convertTypeToBuiltin`?

GleasonK: Would prefer if we didn't convert to builtin at all? Can we make this a function of VHLO? I believe the initial changelist I shared had a method for extracting element type

GleasonK (approved):
LGTM aside from the one open comment

stablehlo/dialect/TypeInference.cpp
```diff
@@ -97,6 +97,51 @@ bool tensorsHaveSameElType(Type type1, Type type2,
...
+
+  Type tensorEl1 = tensorTy1.getElementType();
+  Type tensorEl2 = tensorTy2.getElementType();
+
+  if (ignoreFpPrecision && tensorEl1.isa<FloatType>() &&
+      tensorEl2.isa<FloatType>())
+    return true;
+
+  bool isSameType =
+      (tensorEl1.isa<IntegerType>() and tensorEl2.isa<IntegerType>()) ||
```
GleasonK, line 126: I don't think we use `and` anywhere else in our code. Let's stick to `&&`

stablehlo/dialect/TypeInference.cpp
```diff
@@ -97,6 +97,51 @@ bool tensorsHaveSameElType(Type type1, Type type2,
...
+                             bool ignoreFpPrecision = false) {
+  auto tensorTy1 = type1.dyn_cast<ShapedType>();
+  auto tensorTy2 = type2.dyn_cast<ShapedType>();
+
+  if (!tensorTy1 || !tensorTy2) return false;
+
+  Type tensorEl1 = tensorTy1.getElementType();
+  Type tensorEl2 = tensorTy2.getElementType();
+
+  if (ignoreFpPrecision && tensorEl1.isa<FloatType>() &&
```
GleasonK, line 121: What is ignore precision for?

stablehlo/dialect/TypeInference.cpp
```diff
@@ -97,6 +97,51 @@ bool tensorsHaveSameElType(Type type1, Type type2,
...
+      tensorEl2.isa<FloatType>())
+    return true;
+
+  bool isSameType =
+      (tensorEl1.isa<IntegerType>() and tensorEl2.isa<IntegerType>()) ||
+      (tensorEl1.isa<FloatType>() and tensorEl2.isa<FloatType>()) ||
+      (tensorEl1.isa<ComplexType>() and tensorEl2.isa<ComplexType>()) ||
+      (tensorEl1.isa<quant::QuantizedType>() and
+       tensorEl2.isa<quant::QuantizedType>());
```
GleasonK, line 130: A short helper method may help for all these:

```
template <typename T>
bool matchesType(Type a, Type b) {
   return a.isa<T>() && b.isa<T>();
}
```

GleasonK (changes requested):

stablehlo/dialect/TypeInference.cpp
sdasgup3: done!

stablehlo/dialect/TypeInference.cpp
sdasgup3: done!

stablehlo/dialect/VhloOps.cpp
sdasgup3: makes sense! updated. PTAL.

stablehlo/dialect/TypeInference.cpp
sdasgup3: The ignore fl precision is previously used in `verifyReducerShape` (at https://github.com/openxla/stablehlo/blob/3bdd8593a46ababb362928187d58ccc418e15b99/stablehlo/dialect/TypeInference.cpp#L601 and https://github.com/openxla/stablehlo/blob/3bdd8593a46ababb362928187d58ccc418e15b99/stablehlo/dialect/TypeInference.cpp#L611) to match the behavior with HLO (https://github.com/openxla/xla/blob/8a0bc8775368467e37e853f4441a3be4c2610ae4/xla/service/shape_inference.cc#L148 and https://github.com/openxla/xla/blob/8a0bc8775368467e37e853f4441a3be4c2610ae4/xla/service/shape_inference.cc#L159).

The idea was to ignore the fp precision when fp types are involved. I plan was to keep the existing behavior for now and address its relevance as part of https://github.com/openxla/stablehlo/issues/369. Please let me know what do you think?

stablehlo/dialect/TypeInference.cpp
```diff
@@ -530,6 +572,17 @@ LogicalResult verifyReduceOpInputsAndInferShape(
   return success();
 }
 
+// Returns the types of the terminator arguments of the input  mlir::Block
+// 'block'.
+SmallVector<ShapedType> getAccumulatorTypes(Block& block) {
+  SmallVector<ShapedType> accumulatorSubShapes;
+  for (Value retOperand : block.getTerminator()->getOperands()) {
+    auto shapedTy = retOperand.getType().dyn_cast<ShapedType>();
```
GleasonK, line 580: Use `cast` if we're sure, error check `dyn_cast` result if we aren't. I'm not sure if we've verified that these are all shaped type, also not sure if ODS enforces any of this. Can see if this crashes with a `!stablehlo.token` type being returned I suppose

stablehlo/dialect/StablehloOps.td
```diff
@@ -1448,7 +1447,7 @@ def StableHLO_AllToAllOp : StableHLO_Op<"all_to_all",
 def StableHLO_ReduceOp: StableHLO_ShapedInterfaceOp<"reduce", [
       RecursiveMemoryEffects,
       SameVariadicOperandSize /*reduce_c3*/,
-      InferTensorTypeWithReify /*reduce_c7*/,
+      InferTensorTypeWithReify /*reduce_c7, reduce_c8*/,
       SingleBlockImplicitTerminator<"ReturnOp">
     ]> { /*reduce_c7*/
   let summary = "Reduce operation";
```
GleasonK, line 1453: Can't leave comment below this line, but ~15 lines down theres a TODO comment that I think we can delete now:

```
  // TODO(hinsu): Verify that the attached body arguments and results are
  // compatible with reduce op's operands.
```

stablehlo/dialect/TypeInference.cpp
```diff
@@ -97,6 +97,54 @@ bool tensorsHaveSameElType(Type type1, Type type2,
...
+                             bool ignoreFpPrecision = false) {
+  auto tensorTy1 = type1.dyn_cast<ShapedType>();
+  auto tensorTy2 = type2.dyn_cast<ShapedType>();
+
+  if (!tensorTy1 || !tensorTy2) return false;
+
+  Type tensorEl1 = tensorTy1.getElementType();
+  Type tensorEl2 = tensorTy2.getElementType();
+
+  if (ignoreFpPrecision && tensorEl1.isa<FloatType>() &&
```
GleasonK, line 126: Let's move this below isSameType checking. Only need to check if one of the elements is FP type as well.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -97,6 +97,54 @@ bool tensorsHaveSameElType(Type type1, Type type2,
...
+
+  if (!tensorEl1.isa<quant::QuantizedType>())
+    return potentiallyComplexBitWidth(tensorEl1) <=
+           potentiallyComplexBitWidth(tensorEl2);
+
+  auto quantType1 = tensorEl1.cast<quant::QuantizedType>();
+  auto quantType2 = tensorEl2.cast<quant::QuantizedType>();
+  return quantType1.getExpressedType() == quantType2.getExpressedType() &&
+         potentiallyComplexBitWidth(quantType1.getStorageType()) <=
+             potentiallyComplexBitWidth(quantType2.getStorageType());
```
GleasonK, line 145: Let's refactor the above function:

```
unsigned getBitWidth(Type type) {
  if (auto auto complexTy = type.dyn_cast<ComplexType>())
    return getBitWidth(complexTy.getElementType());
  if (auto auto quantTy = type.dyn_cast<quant::QuantizedType>())
    return getBitWidth(complexTy.getStorageType());
  return type.getIntOrFloatBitWidth();
}
```

Then we can simplify this function to:

```
return getBitWidth(tensorEl1) <= getBitWidth(tensorEl2);
```

stablehlo/dialect/TypeInference.cpp
```diff
@@ -97,6 +97,54 @@ bool tensorsHaveSameElType(Type type1, Type type2,
...
+unsigned potentiallyComplexBitWidth(Type type) {
+  auto complexTy = type.dyn_cast<ComplexType>();
+  return complexTy ? 2 * complexTy.getElementType().getIntOrFloatBitWidth()
+                   : type.getIntOrFloatBitWidth();
+}
+
+template <typename T>
+bool matchesType(Type a, Type b) {
+  return a.isa<T>() && b.isa<T>();
+}
```
GleasonK, line 109: Let's validate expressed type here as well

```
template <typename T>
bool matchesType(Type a, Type b) {
  bool matches = a.isa<T>() && b.isa<T>();
  if constexpr (std::is_same<T, quant::QuantizedType>::value) {
    // Check that expressed type matches for quantized types
    return matches && (a.cast<quant::QuantizedType>().getExpressedType() == b.cast<quant::QuantizedType>().getExpressedType());
  }
  return matches;
}
```

stablehlo/dialect/TypeInference.cpp
```diff
@@ -2661,8 +2741,16 @@ LogicalResult inferRngOp(
 }
 
 LogicalResult inferScatterOp(std::optional<Location>, ValueRange inputs,
+                             Region& update_computation,
```
GleasonK, line 2744: `camelCase`

stablehlo/dialect/VhloOps.cpp
```diff
@@ -296,5 +298,101 @@ void VhloDialect::printAttribute(Attribute attr, DialectAsmPrinter& os) const {
...
+    return ranked.getElementType();
+  }
+  return tensorType.cast<UnrankedTensorV1Type>().getElementType();
+}
+
+bool checkIfOperandAndResultElementTypesMatch(TypeRange operandTypes,
+                                              TypeRange resultTypes) {
+  SmallVector<Type> inputElementTypes{llvm::map_range(
+      operandTypes, [](Type t) { return getVhloElementType(t); })};
+  SmallVector<Type> resultElementTypes{llvm::map_range(
```
GleasonK, line 319: To confirm, none of the types modified allow `stablehlo.token` to be an input/output correct?

stablehlo/dialect/VhloOps.cpp
```diff
@@ -296,5 +298,101 @@ void VhloDialect::printAttribute(Attribute attr, DialectAsmPrinter& os) const {
...
+    return true;
+
+  return false;
+}
+}  // namespace
+
+LogicalResult AllReduceOpV1::validateConstraint(mlir::Operation* op,
+                                                Version targetVersion) {
+  // Allow mismatched operand and result types in v0.17.0
+  if (checkIfOperandAndResultElementTypesMatch(getOperand().getType(),
```
GleasonK, line 334: What are your thoughts on making a helper method like:

```
// Allow mismatched operand and result types in reduce ops in v0.17.0
LogicalResult verifyConstraint_0_17_0(mlir::Operation* op, Version targetVersion) {
  if (checkIfOperandAndResultElementTypesMatch(op->getOperandTypes(),
                                               op->getResultTypes()) &&
      targetVersion < Version(0, 17, 0))
    return failure();
  return success();
}
```

Then replacing all these validateConstraints bodies with:

```
return verifyConstraint_0_17_0(op, targetVersion);
```

stablehlo/transforms/VhloToVersion.cpp
```diff
@@ -181,7 +181,14 @@ bool isLegalOperation(Operation* op, const Version& targetVersion) {
...
   if (!isLegalVersion(opInterface, targetVersion)) return false;
-  LLVM_DEBUG(llvm::dbgs() << "Legal version for target. " << op << '\n');
+  LLVM_DEBUG(llvm::dbgs() << "Legal op version for target. " << op << '\n');
+
+  // Validate op constraints
+  auto constraintInterface = dyn_cast<VersionedOpConstraintInterface>(op);
+  if (constraintInterface &&
+      failed(constraintInterface.validateConstraint(op, targetVersion)))
+    return false;
+  LLVM_DEBUG(llvm::dbgs() << "Legal constraints for target. " << op << '\n');
```
GleasonK, line 191: Correction to my own code here - I'm realizing that this will be difficult to debug what's going wrong as is. Adding a debug message the denote failed constraint verification will make this simpler.

```
  if (!isLegalVersion(opInterface, targetVersion)) return false;
  auto constraintInterface = dyn_cast<VersionedOpConstraintInterface>(op);
  if (constraintInterface &&
      failed(constraintInterface.validateConstraint(op, targetVersion))) {
    LLVM_DEBUG(llvm::dbgs() << "Op failed to satisfy versioned constraints. " << op << '\n');
    return false;
  }
  LLVM_DEBUG(llvm::dbgs() << "Legal op version for target. " << op << '\n');

```

GleasonK (changes requested):

stablehlo/dialect/TypeInference.cpp
sdasgup3: Yup that is a typo. fixed.

The ODS only checks if the inputs and results are Shaped types. The check for block arguments and returns are done at the custom verifier. The check at  [[cs](https://github.com/openxla/stablehlo/blob/cb42882825af722cc108e0051f760ec670321ce7/stablehlo/dialect/TypeInference.cpp#L567)] ensures that thet are on Shaped Types. 

Verified with no crashes for the following but got the verification error. 
```
func.func @op_reduce(%arg0: tensor<f32>, %arg1: tensor<f32>) -> tensor<f32> {

  %0 = "stablehlo.reduce"(%arg0, %arg1) ({

    ^bb0(%arg4: !stablehlo.token, %arg6: !stablehlo.token):
      %1 = stablehlo.add %arg4, %arg6 : !stablehlo.token
      "stablehlo.return"(%1) : (!stablehlo.token) -> ()
  }) {
    dimensions = dense<> : tensor<0xi64>
  } : (tensor<f32>, tensor<f32>) -> (tensor<f32>)
  func.return %0 : tensor<f32>
}
```

stablehlo/dialect/VhloOps.cpp
sdasgup3: Yes, that is right. The input/result types, which `checkIfOperandAndResultElementTypesMatch` is dealing with, should be Shaped types (as enforced by ODS).

stablehlo/dialect/VhloOps.cpp
sdasgup3: This looks good to me. 
This will allow modularization of future constraint changes with clear context. 
done

stablehlo/dialect/TypeInference.cpp
```diff
@@ -97,6 +97,53 @@ bool tensorsHaveSameElType(Type type1, Type type2,
...
+  if (auto quantTy = type.dyn_cast<quant::QuantizedType>())
+    return getBitWidth(quantTy.getStorageType());
+  return type.getIntOrFloatBitWidth();
+}
+
+template <typename T>
+bool matchesType(Type a, Type b) {
+  bool matches = a.isa<T>() && b.isa<T>();
+  if constexpr (std::is_same<T, quant::QuantizedType>::
+                    value) {  // Check that expressed type matches for quantized
```
GleasonK, line 112: Move the code comment to a newline. I think this can be formatted better

GleasonK (approved):

PR #1871, opened 2023-12-05
Integrate LLVM at llvm/llvm-project@c4cebe5b4489

GleasonK:


sdasgup3 (approved):

stellaraccident:
Recommend changing TokenType to be ODS generated: that would have made it immune to the upstream break. ODS is generally more resilient to change.

GleasonK:
Yeah, thats a good call, we already do this for VHLO, probably best to do the same for StableHLO. Thanks for the suggestion @stellaraccident!

cc @mlevesquedion could you make a ticket and implement this?

PR #1658, opened 2023-06-30
Use DenseI64ArrayAttr for ops which already specialized the printing/parsing as DenseI64ArrayAttr

joker-eph:
Partially implement #1578

PR #1872, opened 2023-12-05
Use DenseI64ArrayAttr for ops which already specialized the printing/parsing as DenseI64ArrayAttr (copy and update of #1658)

mlevesquedion:
This PR is based on #1658. I rebased the commits onto main, resolved conflicts, and made necessary changes to code that had been added since the last rebase.

Since I was getting a _lot_ of compilation failures after rebasing and I'm new to the project/MLIR, I decided to port each op one at a time and I gave each op its own commit. 

Porting slice op required a lot of changes to `testdata` files. I wrote a Python script that used a regex to make the replacements:

```
r'(.*?)"stablehlo.slice"(.*?){limit_indices = dense<([^>]+?)> : tensor<(\d)xi64>, start_indices = dense<([^>]+?)> : tensor<(\d)xi64>, strides = dense<([^>]+?)> : tensor<(\d)xi64>}( : .*)'
```

This change partially implements #1578.

**Special note for reviewers:** because I split the original commit into multiple commits, I don't know if authorship will be appropriately reflected (the original change was made by @joker-eph). Please advise.

### DIFF
--- a/stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
+++ b/stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
@@ -1422,9 +1422,9 @@ struct SliceConverter final : OpConversionPattern<mlir::stablehlo::SliceOp> {
     }
 
     SmallVector<OpFoldResult, 3> offsets, sizes, strides;
-    auto startIndices = sliceOp.getStartIndices().getValues<int64_t>();
-    auto limitIndices = sliceOp.getLimitIndices().getValues<int64_t>();
-    auto sliceStrides = sliceOp.getStrides().getValues<int64_t>();
+    auto startIndices = sliceOp.getStartIndices();
+    auto limitIndices = sliceOp.getLimitIndices();
+    auto sliceStrides = sliceOp.getStrides();
 
     for (int64_t i = 0, e = argType.getRank(); i < e; ++i) {
       int64_t start = startIndices[i];
--- a/stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
+++ b/stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
@@ -1470,7 +1470,7 @@ struct DynamicSliceConverter final
         dynamicSliceOp.getStartIndices().front().getType());
     for (auto [idx, start, size] :
          llvm::enumerate(adaptor.getStartIndices(),
-                         dynamicSliceOp.getSliceSizes().getValues<int64_t>())) {
+                         dynamicSliceOp.getSliceSizes())) {
       sizes.push_back(rewriter.getI64IntegerAttr(size));
 
       // By stablehlo.DynamicSlice definition:
--- a/stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
+++ b/stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
@@ -841,8 +841,8 @@ struct TransposeOpToTransposeConverter final
     Value emptyTensor =
         getEmptyTensorFor(rewriter, loc, resultTy, op, adaptor.getOperands());
 
-    auto permutation = rewriter.getDenseI64ArrayAttr(
-        llvm::to_vector(op.getPermutation()));
+    auto permutation =
+        rewriter.getDenseI64ArrayAttr(llvm::to_vector(op.getPermutation()));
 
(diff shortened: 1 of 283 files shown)


stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
```diff
@@ -818,7 +818,7 @@ struct TransposeConverter final
     SmallVector<AffineExpr, 2> inputExprs;
     inputExprs.resize(resultType.getRank());
     for (auto [idx, value] : llvm::enumerate(op.getPermutation())) {
-      inputExprs[value.getZExtValue()] = b->getAffineDimExpr(idx);
```
GleasonK, line 821: Any idea if this change is related to using dense array?

stablehlo/conversions/linalg/transforms/StablehloToLinalgConvolution.cpp
```diff
@@ -81,10 +78,8 @@ Value applyConvolutionPadding(Location loc, Value input,
                  RankedTensorType::get({}, inputType.getElementType())));
   }
 
-  return rewriter.create<mlir::stablehlo::PadOp>(
-      loc, input, zero, DenseIntElementsAttr::get(attrType, padLow),
-      DenseIntElementsAttr::get(attrType, padHigh),
-      DenseIntElementsAttr::get(attrType, padInterior));
+  return rewriter.create<mlir::stablehlo::PadOp>(loc, input, zero, padLow,
+                                                 padHigh, padInterior);
```
GleasonK, line 82: Ooh, that's super nice that we can use arrays directly in the builder method.

stablehlo/transforms/VhloLegalizeToStablehlo.cpp
```diff
@@ -464,10 +477,39 @@ SpecialResult convertSpecial(const OpConversionPattern<VhloOpTy>& pattern,
     }
   }
   if (stablehloAttr) {
-    stablehloAttrs.emplace_back(
-        StringAttr::get(pattern.getContext(), stablehloName), stablehloAttr);
+    stablehloAttrs.emplace_back(stablehloName, stablehloAttr);
     return specialSuccess();
   }
+  if constexpr (std::is_same<VhloOpTy, vhlo::FftOpV1>::value) {
```
GleasonK, line 483: A bit more overhead on this conversion since we are ambiguously converting TensorV1 to DenseElements and DenseArray, but not too bad. Overall worth it for compatibility.

stablehlo/conversions/linalg/tests/random.mlir
```diff
@@ -22,671 +22,3 @@ func.func @rng_uniform_1d(%min: tensor<f32>, %max: tensor<f32>) -> tensor<10xf32
 // CHECK-NEXT:   linalg.yield %[[VAL5]] : f32
 // CHECK-NEXT: -> tensor<10xf32>
 
-// -----
-
-// CHECK-LABEL: func @rng_uniform_2d
-func.func @rng_uniform_2d(%min: tensor<f32>, %max: tensor<f32>) -> tensor<3x3xf32> {
```
GleasonK, line 28: Why are all these tests deleted? Did we change the semantics of `stablehlo.rng` with this fix?

GleasonK (changes requested):

stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
mlevesquedion: Yeah, because permutation is now an array of i64, looping over it uses i64 instead of APInt and we no longer need the conversion to i64.

GleasonK (approved):
LGTM once open comment is addressed.

stablehlo/conversions/linalg/transforms/StablehloToLinalgRandom.cpp
```diff
@@ -634,12 +634,12 @@ LogicalResult generateLinalgPhilox32(OpBuilder &builder, Location loc,
   // Slice to only the required results.
   collapseShape[0] = resultTy.getNumElements();
 
-  llvm::SmallVector<int64_t> offset(resultTy.getRank(), 0);
-  llvm::SmallVector<int64_t> stride(resultTy.getRank(), 1);
+  llvm::SmallVector<int64_t> offset(collapseShape.size(), 0);
```
mlevesquedion, line 635: This is where I made the fix for the index out of bounds issue. See also below in generateLinalgPhilox64.

mlevesquedion:
Thanks for doing all the heavy lifting! :)

PR #1873, opened 2023-12-06
Fix broken links

mlevesquedion:
The roadmap is in its own document (it no longer has a header in the README) and some XLA stuff was moved out of tensorflow.

The XLA MLIR interpreter was deleted in https://github.com/tensorflow/tensorflow/commit/079ab56a5b0c5aefcdffa4f20f79c7bb137c6bd0.


ghpvnist (approved):
Thanks for the edits!

Issue #1878, opened 2023-12-11
Migrate to bazel 7


mlevesquedion:
Bazel 7 (https://github.com/bazelbuild/bazel/issues/18548) seems to introduce some changes that break our tests.

We should look into what is needed to migrate to Bazel 7 and do the migration. For now, we can pin the Bazel version to avoid breakages.

[This page](https://bazel.build/external/migration) has instructions on how to migrate.

I'm not sure exactly what the source of the issue is, but I'm seeing this failure in all our tests (I don't see such failures with Bazel < 7):

```
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
Executing tests from //stablehlo/testdata:vmap_gather_from_take_indices_name__4__axis_2_enable_xla_True_mode_clip_dynamic.mlir.test
-----------------------------------------------------------------------------
lit.py: $HOME/.local/lib/python3.11/site-packages/lit/TestingConfig.py:151: fatal: unable to parse config file '/tmp/bazel-working-directory/_main/bazel-out/k8-fastbuild/bin/stablehlo/testdata/vmap_gather_from_take_indices_name__4__axis_2_enable_xla_True_mode_clip_dynamic.mlir.test.runfiles/_main/stablehlo/testdata/lit.site.cfg.py', traceback: Traceback (most recent call last):
  File "$HOME/.local/lib/python3.11/site-packages/lit/TestingConfig.py", line 139, in load_from_path
    exec(compile(data, path, "exec"), cfg_globals, None)
  File "/tmp/bazel-working-directory/_main/bazel-out/k8-fastbuild/bin/stablehlo/testdata/vmap_gather_from_take_indices_name__4__axis_2_enable_xla_True_mode_clip_dynamic.mlir.test.runfiles/_main/stablehlo/testdata/lit.site.cfg.py", line 21, in <module>
    lit_config.load_config(config, os.path.join(os.environ['TEST_SRCDIR'], 'stablehlo') + "/stablehlo/testdata/lit.cfg.py")
  File "$HOME/.local/lib/python3.11/site-packages/lit/LitConfig.py", line 136, in load_config
    config.load_from_path(path, self)
  File "$HOME/.local/lib/python3.11/site-packages/lit/TestingConfig.py", line 126, in load_from_path
    f = open(path)
        ^^^^^^^^^^
[...]

PR #1879, opened 2023-12-11
Pin bazel version to 6.1.0

mlevesquedion:
This should be an effective stopgap for migrating to Bazel 7, which currently breaks our bazel build. This version of bazel should be supported until EOY 2025: https://bazel.build/release\#support-matrix

This specific version was chosen to align with XLA: https://github.com/openxla/xla/blob/83bcebe6dd1081ab501d8d6689153c6bc3ba0344/.bazelversion#L1

https://github.com/openxla/stablehlo/issues/1878


GleasonK (commented):
Should we instead sync to `6.1.0` to keep aligned with XLA:
https://github.com/openxla/xla/blob/main/.bazelversion

mlevesquedion:
SGTM, done. Thanks!

GleasonK:
Looks like bazel fails at head with the v7 update: https://github.com/openxla/stablehlo/actions/runs/7173049573/job/19531538608

It's good to keep synced to the openxla/xla bazel build since that's a very common place for users to get their copy of StableHLO. Lgtm!

PR #1881, opened 2023-12-12
Dynamism RFC

GleasonK:
The current design of dynamism in MHLO and StableHLO has been practically useful. There are success stories of it connecting JAX, PyTorch and TensorFlow to a number of compilers, in a mix of research and production environments. This RFC aims to leverage existing support for dynamism in the StableHLO dialect, discuss improvements to the existing design and then formalize the improved design in the StableHLO opset.

The main challenge with writing this RFC was that it affects the entire opset. The current design involves a lot of corner cases, so it took about a year of practical evaluation by the author - primarily within JAX native serialization, but also correlating with other ongoing and prior projects - to distill the design into just a few general design principles. I think I'm happy with the outcome, but please take a look at the "Summary" section for what that entails.

This RFC addresses a considerable chunk of community feedback on the existing design, but some feedback is deliberately left out of scope for this RFC to enable incremental progress while areas which require additional alignment are developed in parallel. See sections "Community feedback" and "Out of scope" for details.

The only open question at the moment is interoperability with HLO with respect to bounded dynamism. The proposed representation for bounded dynamic types in StableHLO is in parity with the representation for bounded dynamic types in HLO. However, I'm not sure whether this parity covers 100% of bounded dynamism functionality. For example: 1) there appears to be a mismatch in how broadcasts are represented, 2) there is misalignment in representations of dynamic windows (HLO has a high-level representation: VALID and SAME, whereas StableHLO expects the producers to dynamically compute window sizes). Nonetheless, I think that this shouldn't block the initial review of the RFC, since there's a lot of stuff to discuss - and in the meanwhile, I'll be working on confirming interoperability with HLO.

Finally, I'd like to acknowledge Smit Hinsu's work on the [Bounded Dynamism RFC](https://github.com/openxla/stablehlo/pull/194) from Q4 2022, which was superseded by this work. The representation for bounded dynamic types in the StableHLO dialect was designed and implemented by Smit, and Smit's proposal to allow bounded dynamic types everywhere is compatible with the more general proposal from this RFC to enable dynamism for all size-related program elements. Furthermore, Smit contributed the formal spec for get_dimension_size as well as the informal spec for set_dimension_size.

### DIFF
--- a/rfcs/20230704-dynamism-101.md
+++ b/rfcs/20230704-dynamism-101.md
@@ -17,7 +17,7 @@ the following:
 * [(P2) Ratify the existing convention for shape mismatches constituting undefined behavior](#p2).
 * [(P3) Ratify the existing convention for relaxed constraints already implemented in the StableHLO dialect](#p3).
 * [(P4) Enable shape-dependent dynamism for all size-related program elements but keep all axis-related program elements static](#p4).
-* [(P5) Represent shape computations as StableHLO operations on 0-dimensional tensors and drop support for unranked dynamism](#p5).
+* [(P5) Represent shape computations as StableHLO operations on variadic 0-dimensional tensors and drop support for unranked dynamism](#p5).
 <!-- markdownlint-enable line-length -->
 
 These proposals address a considerable chunk of community feedback on the
@@ -522,7 +522,7 @@ ops like `DynamicPadOp` has some drawbacks, which are denoted in [O7](#o7) as
 out of scope for this RFC.
 
 <!-- markdownlint-disable line-length -->
-## <a id="p5" /> (P5) Represent shape computations as StableHLO operations on 0-dimensional tensors and drop support for unranked dynamism
+## <a id="p5" /> (P5) Represent shape computations as StableHLO operations on variadic 0-dimensional tensors and drop support for unranked dynamism
 <!-- markdownlint-enable line-length -->
 
 **Baseline:** The StableHLO dialect represents sizes as 1-dimensional tensors,
@@ -565,8 +565,9 @@ then reshaped to 1-dimensional tensors and then concatenated together
 ([example](https://github.com/tensorflow/mlir-hlo/blob/1fd13ef28d3f423363e59b1a80eb2c26e0c0979d/mhlo/transforms/chlo_legalize_to_hlo/chlo_legalize_to_hlo.cc#L1577-L1587)).
 
 **Proposal:** The RFC proposes to standardize on representing shape computations
-as StableHLO operations working 0-dimensional tensors. More specifically,
-the proposal is to:
+as StableHLO operations working 0-dimensional tensors, changing from using a
+single operand of `tensor<Nxi64>` to using `N` operands of `tensor<i64>`. More
+specifically, the proposal is to:
 
 1) Affirm that only shape computations that use StableHLO operations are
    supported in StableHLO portable artifacts (this can be changed in future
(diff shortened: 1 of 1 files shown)


GleasonK:
This PR is a re-opening of https://github.com/openxla/stablehlo/pull/1668 which was accidentally deleted. This PR is the exact state that #1688 was in when deleted, the remaining open feedback (`dynamic_` ops) will be addressed in a followup commit.

rfcs/20230704-dynamism-101.md
```diff
@@ -0,0 +1,826 @@
...
+   from P5 will no longer work for unranked operands).
+
+```tablegen
+// See P5 for the "before" and "after" of unifying static and dynamic ops.
+// This snippet only shows the "after" for this section.
+def StableHLO_PadOp ... {
+  let arguments = (ins
+    HLO_Tensor:$operand,
+    HLO_Tensor:$padding_value,
+    I64ElementsAttr:$static_edge_padding_low,
```
GleasonK, line 767: Porting over unresolved feedback from @loganchien, will update to fix this shortly. We have decided to remove this proposal from the RFC as we don't have sufficient proof that this is easier or harder to support in MLIR.

------

I am concerned about the usage of variadic arguments.  It can make writing [Declarative Rewrite Rule (DRR)](https://mlir.llvm.org/docs/DeclarativeRewrites/) more difficult (if not impossible).  I am not sure if this will have some impact on the [PDL dialect](https://mlir.llvm.org/docs/Dialects/PDLOps/).

If we want to go through this approach, we should think about how we can assist declarative rewrite patterns.

----

Another way I can think of to unify static/dynamic shape, is to replace both of them with `1DTensorOf<HLO_DimensionValue>` and provide a utility function that converts variadic arguments into a Tensor:

```
// Static-shape
%static_edge_padding_low = "stablehlo.constant"() {value = dense<[...]> : tensor<4xi64>} : () -> tensor<4xi64>
%static_edge_padding_high = "stablehlo.constant"() {value = dense<[...]> : tensor<4xi64>} : () -> tensor<4xi64>
%static_interior_padding = "stablehlo.constant"() {value = dense<[...]> : tensor<4xi64>} : () -> tensor<4xi64>
%pad = "stablehlo.pad"(%operand, %pad_value, %static_edge_padding_low, %static_eduge_padding_high, %static_interior_padding)

// Dynamic-shape
%edge_padding_low = "stablehlo.make_tensor"(some, value, computed, dynamically) : (i64, i64, i64, i64) -> tensor<4xi64>
%edge_padding_high = "stablehlo.make_tensor"(some, value, computed, dynamically) : (i64, i64, i64, i64) -> tensor<4xi64>
%interior_padding = "stablehlo.make_tensor"(some, value, computed, dynamically) : (i64, i64, i64, i64) -> tensor<4xi64>
%pad = "stablehlo.pad"(%operand, %pad_value, %edge_padding_low, %eduge_padding_high, %interior_padding)
```

The benefit of this approach is that the static-shape version looks similar to the prior arts (e.g. TFLite).  Though, I am not sure if this approach has been discussed or not.

GleasonK: Original comment: https://github.com/openxla/stablehlo/pull/1668#discussion_r1253483521

GleasonK: Moved "unify dynamic and static operations" to Out of Scope.

sdasgup3 (approved):

mlevesquedion (approved):

GleasonK:
All open feedback has been addressed for a few weeks now. Extensive review with both OpenXLA frameworks and compilers representation has been completed as well (any feedback from these reviews has been left publicly as PR comments and modifications to the RFC). This RFC is now approved! Any PRs relating to the proposals in this RFC can now be reviewed and merged.

PR #1885, opened 2023-12-13
Introduce StableHLO interpreter module parsing API

penagos:
This PR builds off of the recent interpreter reference API PR (#1797), by exposing an additional method `parseModule` which takes a serialized StableHLO MLIR program, and an accompanying `MLIRContext` input:

```
llvm::ErrorOr<OwningOpRef<ModuleOp>> parseModule(const std::string &mlir, MLIRContext& context);
```

This serves to augment the existing `evalModule` interface in `stablehlo/reference/Api.h` by allowing users to easily execute a serialized StableHLO program, without needing to handle all proper dialect registrations. Prior to this PR, users are responsible for providing an already parsed StableHLO `ModuleOp`, which requires they register all required MLIR/StableHLO dialects a priori (less user friendly interface).


stablehlo/reference/Api.cpp
```diff
@@ -127,5 +132,26 @@ llvm::ErrorOr<SmallVector<InterpreterValue>> evalModule(
...
+  mlir::registerAllDialects(registry);
+  mlir::stablehlo::registerAllDialects(registry);
+  context.loadDialect<mlir::stablehlo::interpreter::InterpreterDialect>();
+  context.appendDialectRegistry(registry);
+
+  mlir::OwningOpRef<mlir::ModuleOp> module(
+      mlir::parseSourceFile<mlir::ModuleOp>(source_mgr, &context));
+
+  if (!module)
+    llvm::report_fatal_error("Failed to parse StableHLO module.");
```
GleasonK, line 151: Return an llvm error instead? API already is `ErrorOr`, no need to fail fatally.

BUILD.bazel
```diff
@@ -416,9 +416,12 @@ cc_library(
         ":reference_process",
         ":reference_scope",
         ":reference_value",
+        ":register",
         "@llvm-project//llvm:Support",
+        "@llvm-project//mlir:AllPassesAndDialects",
```
GleasonK, line 421: Likely don't need per other comment. FuncDialect is already linked below.

GleasonK (changes requested):

stablehlo/reference/Api.cpp
```diff
@@ -127,5 +132,26 @@ llvm::ErrorOr<SmallVector<InterpreterValue>> evalModule(
...
 }
 
+llvm::ErrorOr<OwningOpRef<ModuleOp>> parseModule(const std::string &mlir,
+                                                 MLIRContext &context) {
+  llvm::SourceMgr source_mgr;
+  source_mgr.AddNewSourceBuffer(llvm::MemoryBuffer::getMemBuffer(mlir),
+                                llvm::SMLoc());
+
+  mlir::DialectRegistry registry;
+  mlir::registerAllDialects(registry);
```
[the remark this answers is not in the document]
penagos: Makes sense; moved the `FuncDialect` registration into `stablehlo::registerAllDialects`.

stablehlo/reference/Api.cpp
```diff
@@ -127,5 +131,24 @@ llvm::ErrorOr<SmallVector<InterpreterValue>> evalModule(
   return stablehlo::eval(mainFunc.getBody(), inputs, &fallback);
 }
 
+llvm::ErrorOr<OwningOpRef<ModuleOp>> parseModule(const std::string &mlir,
```
GleasonK, line 134: `parseStablehloModule` given that this only works on StableHLO programs?

GleasonK: I'm hesitant on the redundancy of this API with MLIR's parsing APIs, but given that it registers StableHLO dialects and parses, a more precise name clears that up.

I'm not sure if there are any issues with re-registering dialects in a context.. I'd guess its cached fine so this is all ok to do.

GleasonK (changes requested):

stablehlo/reference/Api.cpp
penagos: Updated name. I believe dialect constructors are stored in a map internally, so a re-register should be idempotent.

GleasonK (approved):

PR #1887, opened 2023-12-14
Introduce attr that can be backed by either DenseI64ArrayAttr or I64ElementsAttr

mlevesquedion:
Port `BroadcastInDim`'s `broadcast_dimensions` as an example.

This will allow us (and downstream consumers) to port tests bit by bit.


stablehlo/tests/i64densearrayorelementsattr.mlir
```diff
@@ -0,0 +1,15 @@
+// RUN: stablehlo-opt %s -verify-diagnostics -split-input-file -allow-unregistered-dialect | FileCheck %s
```
sdasgup3, line 1: use underscores as in other test files?

stablehlo/dialect/AssemblyFormat.cpp
```diff
@@ -279,6 +279,33 @@ ParseResult parseDenseI64Array(OpAsmParser& parser,
...
+
+ParseResult parseI64DenseArrayOrElements1D(OpAsmParser& parser,
+                                           Attribute& attr) {
+  DenseI64ArrayAttr arrayAttr = DenseI64ArrayAttr::parse(parser, Type{})
+                                    .dyn_cast_or_null<DenseI64ArrayAttr>();
+  if (!arrayAttr) {
+    DenseIntElementsAttr elementsAttr;
+    if (parser.parseCustomAttributeWithFallback(elementsAttr, ::mlir::Type{})) {
+      return failure();
+    } else {
```
sdasgup3, line 299: can we remove the else as the previous if is returning?

stablehlo/dialect/AssemblyFormat.cpp
```diff
@@ -279,6 +279,33 @@ ParseResult parseDenseI64Array(OpAsmParser& parser,
...
+      attr = parser.getBuilder().getDenseI64ArrayAttr(
+          llvm::to_vector(elementsAttr.getValues<int64_t>()));
+    }
```
sdasgup3, line 302: return early?
```
return attr;
```

With that we can remnove the following else with

```
attr = arrayAttr;
return success();
```

sdasgup3 (commented):
Just a few nits.

stablehlo/tests/i64densearrayorelementsattr.mlir
mlevesquedion: Do you mean in the name of the test file? I kinda just took the name of the attr and lowercased it. I'm not sure where it would make sense to insert underscores TBH. Any ideas? Or did you mean something else?

sdasgup3: Yes, in the file name. Something like `i64_dense_array_or_elements_attr`. Now as I am thinking more about this can we have the test in ops_stablehlo.mlir instead. Given that this mixed attribute is going to be deprecated in future when the migration of dense array is complete, we may not need a separate file for testing the feature. However, the test variant with dense array will still be useful to keep.

GleasonK: Also can we add:

```
// TODO(#1578) migrate DenseIntElementsAttr to DenseI64ArrayAttr and delete this test
```

GleasonK: Also this is the only test in `tests/` that doesn't follow the naming convention: `category_testname`. I.e. `interpret_add` or `verify_something`.

With that in mind, I'd say we're testing printing and parsing here, which fits into the `print_` category.

How about: `print_i64array_or_elements.mlir`

GleasonK: (I wrote these at the same time as the last commit. Putting them in StableHLO ops works just as well.)

GleasonK (approved):

PR #1891, opened 2023-12-15
Port CHLO ops to I64DenseArrayOrElements1DAttr

mlevesquedion:
This is an intermediate step in the migration to `DenseI64ArrayAttr`.

I had to move the definition for the attr to `Base.td` because it's now used in `ChloOps.td` as well as `StablehloOps.td`.

I also had to do some shenanigans to support checking if the `broadcast_dimensions` are legal in
`ReifyBroadcastBinaryOpReturnTypeShapes`. Because the attr is obtained through generic means (`op->getAttr("broadcast_dimensions")`); it doesn't go through the conversion from storage (e.g. when doing `op.getBroadcastDimensions`, which returns `SmallVector<int64_t>`), so both `DenseIntElementsAttr` and `DenseI64ArrayAttr` have to be handled.

Ideally, I would be able to do `dyn_cast<I64DenseArrayOrElements1DAttr>` and use a custom method on the attr to get the integers, but I see no way to define a custom method on an attr. If we end up needing to do this in more than one place I'll extract a function for it.

### DIFF
--- a/stablehlo/dialect/Base.cpp
+++ b/stablehlo/dialect/Base.cpp
@@ -602,7 +602,7 @@ ShapedType createShapedType(ShapedTypeComponents components) {
 
 // TODO(#1578): Remove this code once all uses of I64DenseArrayOrElements1DAttr
 // have been removed.
-SmallVector<int64_t> i64ArrayOrElementsValues(Attribute attr) {
+SmallVector<int64_t> getI64Array(Attribute attr) {
   SmallVector<int64_t> vec;
   if (!attr) return vec;
   if (auto elements = dyn_cast_or_null<DenseIntElementsAttr>(attr)) {
--- a/stablehlo/dialect/Base.cpp
+++ b/stablehlo/dialect/Base.cpp
@@ -603,19 +603,15 @@ ShapedType createShapedType(ShapedTypeComponents components) {
 // TODO(#1578): Remove this code once all uses of I64DenseArrayOrElements1DAttr
 // have been removed.
 SmallVector<int64_t> getI64Array(Attribute attr) {
-  SmallVector<int64_t> vec;
-  if (!attr) return vec;
-  if (auto elements = dyn_cast_or_null<DenseIntElementsAttr>(attr)) {
-    vec = llvm::to_vector(elements.getValues<int64_t>());
-  } else if (auto array = dyn_cast_or_null<DenseI64ArrayAttr>(attr)) {
-    vec = llvm::to_vector(array.asArrayRef());
-  } else {
-    llvm::report_fatal_error(
-        "called i64ArrayOrElementsValues on Attribute that was neither a "
-        "DenseIntElementsAttr or a DenseI64ArrayAttr",
-        false);
-  }
-  return vec;
+  if (!attr) return {};
+  if (auto elements = attr.dyn_cast<DenseIntElementsAttr>())
+    return llvm::to_vector(elements.getValues<int64_t>());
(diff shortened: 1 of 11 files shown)


stablehlo/dialect/Base.h
```diff
@@ -56,6 +56,8 @@ inline static bool isStaticDimSize(int64_t val) {
   return !isDynamicDimSize(val);
 }
 
+SmallVector<int64_t> i64ArrayOrElementsValues(Attribute);
```
GleasonK, line 59: Let's phrase this as a getter function:

```
// Returns an array ref from 1-D DenseIntElementsAttrs or DenseI64ArrayAttr
ArrayRef<int64_t> getI64Array(Attribute);
```

Also can we use this getter function in the generated code? As long as it is defined before the `inc` is included it should be viable, I believe everything in `Base.h` is already included and can be used in generated code.

stablehlo/dialect/BroadcastUtils.cpp
```diff
@@ -22,12 +22,16 @@ limitations under the License.
...
 #include "llvm/ADT/SmallVector.h"
 #include "mlir/Dialect/Shape/IR/Shape.h"
+#include "stablehlo/dialect/Base.h"
 
 namespace mlir {
 namespace hlo {
 
 bool isLegalNumpyRankedBroadcast(Value lhs, Value rhs,
-                                 DenseIntElementsAttr broadcastDims) {
+                                 Attribute broadcastDimensionsAttr) {
```
GleasonK, line 31: Can we make this an ArrayRef and lift the call to `getI64Array` to the callsite?

GleasonK (changes requested):

stablehlo/dialect/Base.td
```diff
@@ -236,4 +236,21 @@ def HLO_BoundedAttrInterface : AttrInterface<"BoundedAttrInterface"> {
...
+def I64Elements1D : And<[I64ElementsAttr.predicate,  CPred<"$_self.cast<DenseIntElementsAttr>().getType().getRank() == 1">]>;
+
+// TODO(#1578) migrate uses to DenseI64ArrayAttr and delete this attr
+def I64DenseArrayOrElements1DAttr : Attr<Or<[DenseI64ArrayAttr.predicate, I64Elements1D]>, "either a DenseI64ArrayAttr or a 1-dimensional I64ElementsAttr."> {
+  let storageType = "Attribute";
+  let returnType = "SmallVector<int64_t>";
+  let convertFromStorage = [{
+    [&]() -> SmallVector<int64_t> {
+      return hlo::getI64Array($_self);
+    }()
```
GleasonK, line 252: Would `convertFromStorage = "hlo::getI64Array"` work?

GleasonK: edit: `hlo::getI64Array($_self)`

stablehlo/dialect/Base.cpp
```diff
@@ -600,5 +600,23 @@ ShapedType createShapedType(ShapedTypeComponents components) {
   return UnrankedTensorType::get(components.getElementType());
 }
 
+// TODO(#1578): Remove this code once all uses of I64DenseArrayOrElements1DAttr
+// have been removed.
+SmallVector<int64_t> getI64Array(Attribute attr) {
+  SmallVector<int64_t> vec;
```
GleasonK, line 606: Related comment from other PR:
https://github.com/openxla/stablehlo/pull/1905#discussion_r1443310097

Can address in either PR review

GleasonK (approved):

PR #1893, opened 2023-12-16
Port DynamicBroadcastInDim to I64DenseArrayOrElements1DAttr

mlevesquedion:
This is an intermediate step in the migration to `DenseI64ArrayAttr`.

This also includes changing `dense<...>` to `array<...>` in the tests for `BroadcastInDim` and `DynamicBroadcastInDim` (both are supported, but we want to move towards `array<...>`).

### DIFF
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -1906,17 +1906,17 @@ def StableHLO_DynamicBroadcastInDimOp : StableHLO_ShapedInterfaceOp<
   let arguments = (ins
     HLO_Tensor:$operand,
     HLO_DimensionTensor:$output_dimensions,
-    BroadcastDimAttr:$broadcast_dimensions,
-    OptionalAttr<BroadcastDimAttr>:$known_expanding_dimensions,
-    OptionalAttr<BroadcastDimAttr>:$known_nonexpanding_dimensions
+    I64DenseArrayOrElements1DAttr:$broadcast_dimensions,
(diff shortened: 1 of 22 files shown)


stablehlo/dialect/StablehloOps.td
```diff
@@ -1906,27 +1906,17 @@ def StableHLO_DynamicBroadcastInDimOp : StableHLO_ShapedInterfaceOp<
...
-    OptionalAttr<BroadcastDimAttr>:$known_expanding_dimensions,
-    OptionalAttr<BroadcastDimAttr>:$known_nonexpanding_dimensions
+    I64DenseArrayOrElements1DAttr:$broadcast_dimensions,
+    OptionalAttr<I64DenseArrayOrElements1DAttr>:$known_expanding_dimensions,
+    OptionalAttr<I64DenseArrayOrElements1DAttr>:$known_nonexpanding_dimensions
   );
 
   let results = (outs HLO_Tensor);
 
-  let builders = [
```
GleasonK, line 1916: Is this safe to delete? Could be 3p relying on it? Or is this the identical signature of the generated op builder as it is?

GleasonK (approved):

stablehlo/dialect/StablehloOps.td
mlevesquedion: Ah yep good point. I think it should be fine because the attrs are marked optional and the builders handle it. But I'm not 100% sure so I reverted the change. I just changed `DenseIntElementsAttr` to `Attribute` to be coherent with the other builders.

PR #1897, opened 2024-01-02
Clarify semantics of `ReduceWindowOp` and `SelectAndScatterOp`

### LINKED ISSUE #1818: Clarify semantics of `ReduceWindowOp`
It is unclear from just looking at the spec of `reduce_window` that the reduction `body` and `init_values` must form a monoid to guarantee that the operation produces the same results for all inputs on all implementations, which is implicitly inherited from the `reduce` spec through the `reduce()` function.

ghpvnist:
closes #1818

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -4298,6 +4298,13 @@ Afterwards, within each `process_group`:
 Applies a reduction function `body` to windows of `inputs` and `init_values`
 and produces `results`.
 
+The order of reductions is implementation-defined, which means that `body` and
+`init_values` must form a monoid to guarantee that the operation produces the
+same results for all inputs on all implementations. However, this condition
+doesn't hold for many popular reductions. E.g. floating-point addition for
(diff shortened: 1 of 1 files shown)


sdasgup3:
As there are other reduction operations like `select_and_scatter` which implicitly uses reduce, should be make the clarification uniformly across all those operations?

ghpvnist:
These ops use `reduce`:
* `batch_norm_grad`
* `batch_norm_training`
* `dot_general`
* `reduce_window`
* `reshape`
* `select_and_scatter`

Of these, `batch_norm_grad`, `batch_norm_training`, `dot_general`, and `reshape` use trivial `reduce` that does not require clarification because they are already monoid. `reduce_window` and `select_and_scatter` are the only ones that may benefit from further clarification. `select_and_scatter` uses `reduce_window` which in turn uses `reduce`, so I've added clarification there.

sdasgup3 (approved):

PR #1900, opened 2024-01-02
Port SelectAndScatterOp to I64DenseArrayOrElements1DAttr

mlevesquedion:
In the same vein as #1893.

#1578

### DIFF
--- a/stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
+++ b/stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
@@ -1899,10 +1899,8 @@ struct SelectAndScatterNoOverlapConverter final
     if (!op.getWindowDimensions().has_value())
       return rewriter.notifyMatchFailure(op, "no window dimensions found");
 
-    auto strides =
-        llvm::to_vector(op.getWindowStridesAttr().getValues<int64_t>());
-    auto window =
-        llvm::to_vector(op.getWindowDimensionsAttr().getValues<int64_t>());
+    auto strides = op.getWindowStrides().value();
+    auto window = op.getWindowDimensions().value();
 
     if (static_cast<int64_t>(strides.size()) != operandTy.getRank() ||
         static_cast<int64_t>(window.size()) != operandTy.getRank())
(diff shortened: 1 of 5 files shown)


stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
```diff
@@ -1899,10 +1899,8 @@ struct SelectAndScatterNoOverlapConverter final
     if (!op.getWindowDimensions().has_value())
       return rewriter.notifyMatchFailure(op, "no window dimensions found");
 
-    auto strides =
-        llvm::to_vector(op.getWindowStridesAttr().getValues<int64_t>());
-    auto window =
-        llvm::to_vector(op.getWindowDimensionsAttr().getValues<int64_t>());
+    auto strides = op.getWindowStrides().value_or(SmallVector<int64_t>{});
```
GleasonK, line 1902: The two if-stmts above check that there is a value, can we just use `value()` here instead of `value_or` ?

stablehlo/reference/Ops.cpp
```diff
@@ -699,14 +699,13 @@ SmallVector<InterpreterValue> eval(Region &region,
...
-            windowDimensionsAttr->getValues<int64_t>().begin(),
-            windowDimensionsAttr->getValues<int64_t>().end());
+        windowDimensions.assign(windowDimensionsAttr->begin(),
+                                windowDimensionsAttr->end());
 
       Sizes windowStrides(rank, 1);
       if (auto windowStridesAttr = selectAndScatterOp.getWindowStrides())
-        windowStrides.assign(windowStridesAttr->getValues<int64_t>().begin(),
-                             windowStridesAttr->getValues<int64_t>().end());
+        windowStrides.assign(windowStridesAttr->begin(),
```
GleasonK, line 707: Is this safe? Does this create multiple vectors?

GleasonK (changes requested):

stablehlo/reference/Ops.cpp
mlevesquedion: Should be just one vector. `getWindowStrides` will return a vector and `begin`/`end` will return iterators on the vector.

Or maybe I'm misunderstanding what you are alluding to?

GleasonK (approved):

PR #1902, opened 2024-01-03
Port {Map,Reduce}Op  to I64DenseArrayOrElements1DAttr

mlevesquedion:

### DIFF
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -1563,14 +1563,14 @@ void ReduceOp::print(OpAsmPrinter& p) {
     printEscapedString(innerOp.getName().getStringRef(), p.getStream());
 
     p << " across dimensions = [";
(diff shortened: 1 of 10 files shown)


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -1640,7 +1640,7 @@ ParseResult ReduceOp::parse(OpAsmParser& parser, OperationState& result) {
         parser.parseKeyword("reducer"))
       return failure();
     OpBuilder builder(parser.getBuilder().getContext());
-    result.addAttribute("dimensions", builder.getI64TensorAttr(dimensions));
+    result.addAttribute("dimensions", builder.getDenseI64ArrayAttr(dimensions));
```
GleasonK, line 1643: Should we make this change now or when we flip the default to DenseArray? Same below.

GleasonK (approved):

PR #1903, opened 2024-01-03
Port ReduceWindowOp to I64DenseArrayOrElements1DAttr

mlevesquedion:

### DIFF
--- a/stablehlo/conversions/linalg/transforms/StablehloToLinalgReduce.cpp
+++ b/stablehlo/conversions/linalg/transforms/StablehloToLinalgReduce.cpp
@@ -337,8 +337,7 @@ struct ReduceWindowOpOnTensorsGenericConversion final
       return failure();
     auto numOperands = initValues.size();
 
-    llvm::SmallVector<int64_t> windowDimensions =
-        extract1DVector(op.getWindowDimensions());
+    llvm::SmallVector<int64_t> windowDimensions(op.getWindowDimensions());
 
     llvm::SmallVector<int64_t> padding;
     if (op.getPadding()) {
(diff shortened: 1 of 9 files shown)


stablehlo/conversions/linalg/transforms/StablehloToLinalgReduce.cpp
```diff
@@ -348,17 +347,17 @@ struct ReduceWindowOpOnTensorsGenericConversion final
 
     llvm::SmallVector<int64_t> baseDilations;
     if (op.getBaseDilations()) {
-      baseDilations = extract1DVector(*op.getBaseDilations());
+      baseDilations = llvm::SmallVector<int64_t>(*op.getBaseDilations());
```
GleasonK, line 350: Is the `llvm::SmallVector<int64_t>` needed here? Isn't that the return type as it is?

If not can we use `llvm::to_vector`?

GleasonK: Same for other places in this file

stablehlo/conversions/linalg/transforms/StablehloToLinalgReduce.cpp
```diff
@@ -557,26 +556,27 @@ struct ReduceWindowOpConversion final
       return rewriter.notifyMatchFailure(op, "require paddings are all zero");
     }
 
-    if (op.getBaseDilations() && !isSplatValue(*op.getBaseDilations(), 1)) {
-      return rewriter.notifyMatchFailure(op, "expected undilated base");
+    if (auto bd = op.getBaseDilations()) {
+      if (!std::all_of(bd->begin(), bd->end(),
```
GleasonK, line 560: Let's make a helper function `isSplatArray(arr, val)`

GleasonK: This function can live in StablehloOps.h or Base.h (probably base?)

GleasonK (changes requested):

stablehlo/dialect/Base.h
```diff
@@ -56,6 +56,8 @@ inline static bool isStaticDimSize(int64_t val) {
   return !isDynamicDimSize(val);
 }
 
+bool isSplatArray(ArrayRef<int64_t> arr, int64_t val);
```
GleasonK, line 59: Add code comment which talks about why this is needed.

Same for getI64Array

GleasonK (approved):

PR #1905, opened 2024-01-05
Port {Convolution,DynamicConv,Gather}Op to I64DenseArrayOrElements1DAttr

mlevesquedion:

I decided to port the convolution "window_reversal" attribute to array directly, since it seems rarely used and it is the only attribute that is a 1D tensor of booleans.

Not porting the MLIR source files yet.

#1578

### DIFF
--- a/stablehlo/conversions/linalg/transforms/StablehloToLinalgConvolution.cpp
+++ b/stablehlo/conversions/linalg/transforms/StablehloToLinalgConvolution.cpp
@@ -22,6 +22,7 @@ limitations under the License.
 #include "mlir/Transforms/DialectConversion.h"
 #include "stablehlo/conversions/linalg/transforms/LegalizeToLinalgUtils.h"
 #include "stablehlo/conversions/linalg/transforms/Rewriters.h"
+#include "stablehlo/dialect/Base.h"
 #include "stablehlo/dialect/StablehloOps.h"
 
 namespace mlir::stablehlo {
@@ -29,13 +30,14 @@ namespace {
 /// Apply dilation and padding to the input of a convolution.
 Value applyConvolutionPadding(Location loc, Value input,
                               DenseIntElementsAttr padding,
-                              DenseIntElementsAttr lhsDilation,
+                              Attribute lhsDilation,
                               llvm::ArrayRef<int64_t> dimMappings,
                               OpBuilder &rewriter) {
-  if ((!padding || isSplatValue(padding, 0)) &&
-      (!lhsDilation || isSplatValue(lhsDilation, 1))) {
-    return input;
-  }
+  SmallVector<int64_t> lhsDilationValues;
+  if (lhsDilation) lhsDilationValues = hlo::getI64Array(lhsDilation);
+  bool noPadding = !padding || isSplatValue(padding, 0);
+  bool noDilation = !lhsDilation || hlo::isSplatArray(lhsDilationValues, 1);
+  if (noPadding && noDilation) return input;
 
   auto inputType = cast<ShapedType>(input.getType());
   int64_t rank = inputType.getRank();
@@ -58,10 +60,10 @@ Value applyConvolutionPadding(Location loc, Value input,
   // Translate input dilation into interior padding.
   SmallVector<int64_t, 8> padInterior(rank, 0);
   if (lhsDilation) {
-    assert(rank == lhsDilation.size() + 2);
-    for (int64_t i : llvm::seq<int64_t>(0, lhsDilation.size())) {
+    assert(rank == static_cast<int64_t>(lhsDilationValues.size()) + 2);
+    for (int64_t i : llvm::seq<int64_t>(0, lhsDilationValues.size())) {
       int64_t dim = dimMappings[i];
-      padInterior[dim] = lhsDilation.getValues<int64_t>()[i] - 1;
+      padInterior[dim] = lhsDilationValues[i] - 1;
     }
   }
 
@@ -91,8 +93,7 @@ Value applyConvolutionReversal(Location loc, OpBuilder &b,
     return filter;
   }
   llvm::SmallVector<int64_t> reversedDims;
-  for (auto [idx, reversed] :
-       llvm::enumerate(reversals.value().getValues<bool>())) {
+  for (auto [idx, reversed] : llvm::enumerate(reversals.value())) {
     if (reversed) {
       reversedDims.push_back(
           op.getDimensionNumbers().getKernelSpatialDimensions()[idx]);
@@ -219,8 +220,10 @@ struct NormalConvolutionOpConversion final
         loc, resultType.getShape(), resultType.getElementType(), dynSizes);
     Value zeroTensor = fillTensorWithZeros(rewriter, loc, emptyTensor);
     linalg::LinalgOp res;
-    Attribute strides = op.getWindowStridesAttr();
-    Attribute dilations = op.getRhsDilationAttr();
+    Attribute strides;
+    if (auto s = op.getWindowStrides()) strides = rewriter.getI64TensorAttr(*s);
+    Attribute dilations;
+    if (auto d = op.getRhsDilation()) dilations = rewriter.getI64TensorAttr(*d);
 
     // Apply padding and input dilation.
     llvm::SmallVector<int64_t> spatialDimMapping(rank - 2);
@@ -512,7 +515,7 @@ struct ConvolutionOpGeneralConversion final
 
       AffineExpr stride = dim0;
       if (op.getWindowStrides().has_value())
-        stride = stride * op.getWindowStrides().value().getValues<int64_t>()[i];
+        stride = stride * op.getWindowStrides().value()[i];
       AffineExpr srcExpr = stride + dim1;
 
       srcExprs[lhsIndexMapping[inputSpatialDimensions[i]]] = srcExpr;
@@ -599,7 +602,7 @@ struct DepthwiseConvolutionOpConversion final
 
     Attribute windowStrides;
     if (op.getWindowStrides()) {
-      windowStrides = op.getWindowStrides().value();
+      windowStrides = rewriter.getI64TensorAttr(op.getWindowStrides().value());
     } else {
       windowStrides = SplatElementsAttr::get(
           VectorType::get({spatialRank}, rewriter.getI64Type()),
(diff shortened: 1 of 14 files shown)


stablehlo/conversions/linalg/transforms/StablehloToLinalgConvolution.cpp
```diff
@@ -22,19 +22,24 @@ limitations under the License.
...
                               OpBuilder &rewriter) {
-  if ((!padding || isSplatValue(padding, 0)) &&
-      (!lhsDilation || isSplatValue(lhsDilation, 1))) {
-    return input;
+  SmallVector<int64_t> lhsDilationValues;
+  if (lhsDilation)
+    lhsDilationValues = hlo::i64ArrayOrElementsValues(lhsDilation);
+  if (!padding || isSplatValue(padding, 0)) {
+    if (!lhsDilation) return input;
+    if (llvm::all_of(lhsDilationValues, [](int64_t x) { return x == 1; }))
```
GleasonK, line 41: Add helper function `isSplatArray`

stablehlo/dialect/Base.cpp
```diff
@@ -600,5 +600,23 @@ ShapedType createShapedType(ShapedTypeComponents components) {
   return UnrankedTensorType::get(components.getElementType());
 }
 
+// TODO(#1578): Remove this code once all uses of I64DenseArrayOrElements1DAttr
+// have been removed.
+SmallVector<int64_t> i64ArrayOrElementsValues(Attribute attr) {
+  SmallVector<int64_t> vec;
+  if (!attr) return vec;
+  if (auto elements = dyn_cast_or_null<DenseIntElementsAttr>(attr)) {
+    vec = llvm::to_vector(elements.getValues<int64_t>());
```
GleasonK, line 609: - No ned for `else if` statements, each if can just return.
- No need for temporary vector
- I think we can usre `dyn_cast` here instead of `dyn_cast_or_null`
- Attributes have a `dyn_cast` method
- Why the `false` arg to `report_fatal_error`?

All together that makes for:

```cpp
if (!attr) return {};
if (auto elements = attr.dyn_cast<DenseIntElementsAttr>())
  return llvm::to_vector(elements.getValues<int64_t>());
if (auto array = attr.dyn_cast<DenseI64ArrayAttr>())
  return llvm::to_vector(array.asArrayRef());
llvm::report_fatal_error(
    "called i64ArrayOrElementsValues on Attribute that was neither a "
    "DenseIntElementsAttr or a DenseI64ArrayAttr");
```

stablehlo/dialect/StablehloAttrs.td
```diff
@@ -160,17 +160,6 @@ def StableHLO_FlatSymbolRefArrayAttr :
   let constBuilderCall = "::mlir::ArrayAttr::get($_builder.getContext(), $0)";
 }
 
-def StableHLO_BoolElementsAttr :
```
GleasonK, line 163: Very nice that we don't need this anymore

stablehlo/transforms/StablehloCanonicalizeDynamism.cpp
```diff
@@ -153,7 +153,7 @@ struct CanonicalizeDynamicGatherOpPattern
       return rewriter.notifyMatchFailure(op, "expected static slice_sizes");
     rewriter.replaceOpWithNewOp<GatherOp>(
         op, op.getType(), op.getOperand(), op.getStartIndices(),
-        op.getDimensionNumbersAttr(), rewriter.getI64TensorAttr(sliceSizes),
+        op.getDimensionNumbersAttr(), rewriter.getDenseI64ArrayAttr(sliceSizes),
```
GleasonK, line 156: Lets hold off on this change as well. May cause version skew issues in JAX.

GleasonK (changes requested):
Confirmed that bytecode files do not break forward compatibility

stablehlo/conversions/linalg/transforms/StablehloToLinalgConvolution.cpp
mlevesquedion: Done.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1850,22 +1850,12 @@ LogicalResult inferConvolutionOp(
...
-      convert1DAttribute(rhsDilation, location, "rhs_dilation");
-  if (failed(rhsDilationOrErr)) return failure();
-  auto windowReversalOrErr = convertWindowReversalAttribute(
-      windowReversal, location, "window_reversal");
-  if (failed(windowReversalOrErr)) return failure();
 
   auto windowOrErr = verifyWindowAttributesAndInferWindowDimensions(
-      windowDimensions, *windowStridesOrErr, *paddingOrErr, *lhsDilationOrErr,
-      *rhsDilationOrErr, *windowReversalOrErr, location);
+      windowDimensions, windowStrides.value_or(ArrayRef<int64_t>{}),
```
GleasonK, line 1855: We should consider modeling these as `DefaultValuedOptionalAttribute` so we don't need to do `value_or` everywhere. Doesn't need to be in this PR.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -2316,30 +2306,33 @@ LogicalResult inferGatherOp(
...
+  SmallVector<int64_t, 1> ssShape{static_cast<int64_t>(sliceSizes.size())};
+  ShapedTypeComponents ssSTC{ssShape};
+  ShapeAdaptor sliceSizesShape(ssSTC);
 
   // For some reason the getType call is necessary here
   if (failed(verifyGather(location,
                           /*operandShape=*/operandShape,
                           /*startIndicesShape=*/startIndicesShape,
-                          /*sliceSizesShape=*/sliceSizes.getType(), offsetDims,
+                          /*sliceSizesShape=*/sliceSizesShape, offsetDims,
```
GleasonK, line 2321: Can we modify verifyGather to take in an `ArrayRef<int64_t>` instead of creating a ShapeAdaptor?

stablehlo/tests/stablehlo_legalize_to_vhlo.0_10_0.mlir
```diff
@@ -1211,7 +1211,7 @@ func.func @op_dynamic_conv(%arg0: tensor<1x8x8x207xf32>, %arg1: tensor<3x3x207x1
     padding = dense<1> : tensor<2x2xi64>,
     lhs_dilation = dense<2> : tensor<2xi64>,
     rhs_dilation = dense<2> : tensor<2xi64>,
-    window_reversal = dense<true> : tensor<2xi1>,
+    window_reversal = array<i1: true, true>,
```
GleasonK, line 1214: Where did you land on the round-trip issue? Should we keep these as dense for now?

GleasonK (approved):

stablehlo/tests/stablehlo_legalize_to_vhlo.0_10_0.mlir
mlevesquedion: For this attr IMO it's not worth going through an intermediate "array or elements" step since it's the only attr that's a 1D range of booleans. Also since it doesn't seem like it's used a whole lot IMO it's okay to directly port it to array. We can definitely port it in a later change though, WDYT?

stablehlo/dialect/TypeInference.cpp
mlevesquedion: I don't think I can do that given that this code is shared with the type inference for `dynamic_gather`, which uses `HLO_IntTensor` for its `slice_sizes` attr. Apparently it can't use array or elements for that because it could be dynamic.

GleasonK (approved):

stablehlo/dialect/StablehloAttrs.td
```diff
@@ -193,15 +182,15 @@ def StableHLO_ConvDimensionNumbers : AttrDef<StableHLO_Dialect, "ConvDimensionNu
...
     OptionalAttr<I64ElementsAttr>:$padding,
     // Default value: one for each of the spatial dimension.
-    OptionalAttr<I64ElementsAttr>:$lhs_dilation,
+    OptionalAttr<I64DenseArrayOrElements1DAttr>:$lhs_dilation,
     // Default value: one for each of the spatial dimension.
-    OptionalAttr<I64ElementsAttr>:$rhs_dilation,
+    OptionalAttr<I64DenseArrayOrElements1DAttr>:$rhs_dilation,
     // Default value: false for each of the spatial dimension.
-    OptionalAttr<StableHLO_BoolElementsAttr>:$window_reversal,
+    OptionalAttr<DenseBoolArrayAttr>:$window_reversal,
```
GleasonK, line 193: Let's make this an optional dense or array

stablehlo/dialect/Base.cpp
```diff
@@ -612,10 +612,26 @@ SmallVector<int64_t> getI64Array(Attribute attr) {
...
       false);
 }
 
+SmallVector<bool> getBoolArray(Attribute attr) {
+  if (!attr) return {};
+  if (auto elements = attr.dyn_cast<DenseIntOrFPElementsAttr>())
+    return llvm::to_vector(elements.getValues<bool>());
+  if (auto array = attr.dyn_cast<DenseBoolArrayAttr>()) {
+    // TODO: find a cleaner way
+    SmallVector<bool> vec;
```
GleasonK, line 626: `return array.asArrayRef().vec()` ?

GleasonK (approved):
small nit, otherwise lgtm

stablehlo/dialect/Base.cpp
mlevesquedion: Sorry, meant to fix this before re-assigning you. Thanks for the rec! That returns a `std::vector` so it can't be used as is. The root problem was that I was trying to do `llvm::to_vector(array.asArrayRef())` and that fails because of the way that's implemented:

```
llvm-project/llvm/include/llvm/ADT/SmallVector.h +1308:11: error: type 'decltype(__cont.begin())' (aka 'const bool *') cannot be narrowed to 'bool' in initializer list [-Wc++11-narrowing]
  return {std::begin(Range), std::end(Range)};
          ^~~~~~~~~~~~~~~~~
stablehlo/dialect/Base.cpp +625:18: note: in instantiation of function template specialization 'llvm::to_vector<llvm::ArrayRef<bool>>' requested here
    return llvm::to_vector(array.asArrayRef());
                 ^
llvm-project/llvm/include/llvm/ADT/SmallVector.h +1308:11: note: insert an explicit cast to silence this issue
  return {std::begin(Range), std::end(Range)};
          ^~~~~~~~~~~~~~~~~
          static_cast<bool>( )
```

(For some reason the compiler thinks the code is trying to use the `initializer_list` constructor of `SmallVector` and it "helpfully" suggests casting the iterators to `bool`.)

I decided to just be explicit: `return SmallVector<bool>(array.asArrayRef());`

Also while looking into this I realized the new attr was not tested so I added two basic tests (a convolution with dense window reversal and another with array window reversal).

Issue #1523, opened 2023-05-23
ReduceOp prettyprint is different after `--strip-debuginfo`


GleasonK:
### Request description

This happens because of:
https://github.com/openxla/stablehlo/blob/4cd6f24257a364857add0e3c1dc11b2364669d50/stablehlo/dialect/StablehloOps.cpp#L1489-L1490

Repro:

```bash
$ cat /tmp/t.mlir 
func.func @op_reduce(%arg0: tensor<16xf32>, %arg1: tensor<f32>) -> tensor<f32> {
  %0 = "stablehlo.reduce"(%arg0, %arg1) ({
    ^bb0(%arg2: tensor<f32>, %arg3: tensor<f32>):
      %1 = "stablehlo.add"(%arg2, %arg3) : (tensor<f32>, tensor<f32>) -> tensor<f32>
      "stablehlo.return"(%1) : (tensor<f32>) -> ()
  }) {
    dimensions = dense<0> : tensor<1xi64>
  } : (tensor<16xf32>, tensor<f32>) -> tensor<f32>
  func.return %0 : tensor<f32>
}

$ stablehlo-opt /tmp/t.mlir
module {
  func.func @op_reduce(%arg0: tensor<16xf32>, %arg1: tensor<f32>) -> tensor<f32> {
    %0 = stablehlo.reduce(%arg0 init: %arg1) across dimensions = [0] : (tensor<16xf32>, tensor<f32>) -> tensor<f32>
     reducer(%arg2: tensor<f32>, %arg3: tensor<f32>)  {
      %1 = stablehlo.add %arg2, %arg3 : tensor<f32>
      stablehlo.return %1 : tensor<f32>
    }
    return %0 : tensor<f32>
  }
}

$ stablehlo-opt /tmp/t.mlir --strip-debuginfo
module {
  func.func @op_reduce(%arg0: tensor<16xf32>, %arg1: tensor<f32>) -> tensor<f32> {
    %0 = stablehlo.reduce(%arg0 init: %arg1) applies stablehlo.add across dimensions = [0] : (tensor<16xf32>, tensor<f32>) -> tensor<f32>
    return %0 : tensor<f32>
  }
}
```

This is probably the primary reason most ReduceOp's aren't prettyprinted. It is unclear if this is correct/inocrrect behavior, it is very uncommon for op prettyprints to be based on debug info, but reduce is more complicated so will require some investigation.

fzakaria:
Just to add some clarification:
[...]

PR #1906, opened 2024-01-05
Remove location constraint on compact printing of ReduceOpt

fzakaria:
During discussion of #1523 we came to the conclusion that it's impossible to continously guarantee that location information is kept forwrad and backwards through transformations.

There is already precendent in other transformations that location information may be ellided.

Remove the constraint for the compact pretty printing of ReduceOpt that requires the locations be the same. Update the tests as necessary.


sdasgup3 (approved):
Thanks for the fix.
LGTM.

PR #1908, opened 2024-01-08
Integrate LLVM at llvm/llvm-project@ba4cf31facda

mlevesquedion:
Bumping LLVM caused some failures because of `arith.constant` ops in the `infer_stablehlo.mlir` test being generated in a different (albeit still valid) order. I changed the checks for those ops to `CHECK-DAG` instead of `CHECK` since we do not care about the order.


GleasonK (changes requested):
Why `CHECK-DAG` not `CHECK`? Leaves us with inconsistent file. Should all the CHECKS be CHECK-DAG, or should we settle on CHECK?

mlevesquedion:
> Why `CHECK-DAG` not `CHECK`? Leaves us with inconsistent file. If theres a reason to have a mix then LGTM, otherwise should all the CHECKS be CHECK-DAG, or should we settle on CHECK?

I think it makes sense to use `CHECK-DAG` for these, but not in other cases. Indeed we don't really care about the order that these `constant` operations are in as they are independent from each other. However most other operations actually have some ordering to them and for those `CHECK` makes sense IMO.

This change is a minimal diff to resolve the failures caused by bumping LLVM. My guess is that something in LLVM changed that cause the actual emitted order to change, but we don't actually care about the order so in a way the tests were more specific than they needed to be.

GleasonK (approved):

PR #1912, opened 2024-01-10
Fully qualify storage types of ArrayOrElements attrs

mlevesquedion:
There is no guarantee that users of the generated .inc files are working in an `mlir` namespace, so we need to include the explicit `mlir::` prefix.

I discovered this because in some (out of repo) generated code an unqualified `Attribute` caused compilation failures.


ghpvnist:
I'm curious where you see failure. I don't see any code in nearby repos using fully qualified types (tensorflow usage is unqualified; e.g.:  https://github.com/tensorflow/tensorflow/blob/3a76ff9129c35c7c6d2ceec61e4ae14079943ad9/tensorflow/core/ir/types/attributes.td#L310).

mlevesquedion:
There is no guarantee that users of the generated .inc files are working in an `mlir` namespace, so we need to include the explicit `mlir::` prefix.

This was detected by failures in our internal CI.

(Updated the PR description.)

ghpvnist:
SGTM, thanks for the clarification!

ghpvnist (approved):

PR #1913, opened 2024-01-10
Move large categories of tests to their own folders

mlevesquedion:
The `tests` folder contains a lot of tests for various things. In particular, there are a lot of tests for the interpreter and a lot of tests related to VHLO, so IMO it makes sense to give these their own folders. I find that this helps navigate the tests.

I considered creating folders for CHLO, printing related tests, and verification related tests, but IMO there aren't enough of those to really warrant it.

I find the diff produced by GitHub is not super clear. Diff TLDR: I created new folders `tests/interpret` and `tests/vhlo` and moved the relevant tests into each.


ghpvnist (approved):

sdasgup3:
>>  IMO there aren't enough of those to really warrant it.
Agreed.
In future we might have more verification tests for verifying the quantization related constraints. If needed we can dedicate a separate folder for them.

ghpvnist:
Looks like we need to update `./build_tools/github_actions/lint_version.sh` if we're going to merge this: `COMPAT_TEST_BASE="stablehlo/tests/stablehlo_legalize_to_vhlo"`

mlevesquedion:
> Looks like we need to update `./build_tools/github_actions/lint_version.sh` if we're going to merge this: `COMPAT_TEST_BASE="stablehlo/tests/stablehlo_legalize_to_vhlo"`

Ah yep, good point! And this made me realize that there are a bunch of references to tests that need to be updated. Will reassign once I've fixed those.

PR #1915, opened 2024-01-10
Format bazel files and check bazel format in pull requests

mlevesquedion:
I adapted the job definition from https://github.com/openxla/xla/blob/main/.github/workflows/buildifier.yml.

Also removing a repeated comment in `lint.yml` that seems to be inaccurate in most places.


.github/workflows/lint.yml
```diff
@@ -62,3 +64,19 @@ jobs:
...
+        shell: bash
+    steps:
+      - name: Checking out repository
+        uses: actions/checkout@v4
+
+      - name: Install buildifier
+        run: go install github.com/bazelbuild/buildtools/buildifier@433ea85 # 6.4.0
+
+      - name: Run buildifier
+        run: buildifier --mode=check -r . || (echo "Please run buildifier --mode=fix -r . in the repository's root directory" && exit 1)
```
fzakaria, line 82: should we run the lint as well?

fzakaria: You have to ignore the LLVM directory in that case though.

mlevesquedion: Yep, and yep. Thanks! Ended up writing a small script because it became kinda onerous to have it in the `run:` line.

fzakaria:
Should't you be the assignee?

mlevesquedion:
Sorry, this is ready for review now :)

BUILD.bazel
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 load("@llvm-project//mlir:tblgen.bzl", "gentbl_cc_library", "gentbl_filegroup", "td_library")
+load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library")
```
fzakaria, line 15: ?

build_tools/github_actions/lint_buildifier.sh
```diff
@@ -0,0 +1,37 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+readonly SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+readonly STABLEHLO_ROOT_DIR="${SCRIPT_DIR}/../.."
+
+cd "$STABLEHLO_ROOT_DIR"
+
+targets=(
```
fzakaria, line 28: maybe something that works forward for new files: 
https://stackoverflow.com/questions/4210042/how-do-i-exclude-a-directory-when-using-find

fzakaria (commented):
I think there will be a lot less changes after a rebase.

BUILD.bazel
mlevesquedion: This is from fixing lint warnings:

```
stablehlo/tests/BUILD.bazel:23: native-cc: Function "cc_library" is not global anymore and needs to be loaded from "@rules_cc//cc:defs.bzl". (https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#native-cc)
```

mlevesquedion:
Just rebased, PTAL.

fzakaria:
Interesting there were more changes.
I ran buildifier from /usr/bin/buildifier on my Cloudtop and it had different results.
That makes me think that there are changes across buildifier version; we should consider including buildifier in a follow-up as a hermetic dependency either in the WORKSPACE or using something like Nix (https://nixos.org/) for a shell.

fzakaria (approved):

mlevesquedion:
Good callout. At least the version is pinned in the workflow, but yeah the version was different on my machine. I ended up installing the same version as the workflow and running that locally.

I'll add a comment to the script's output so people know to watch out if they run afoul of the check.

Thanks for the review!

PR #1916, opened 2024-01-10
Layering Add support for layering_check and change default toolchain

fzakaria:
Bazel has support for layering_check identification which helps validate that sources are only pulling in header files from their direct dependencies.
    
 This is considered best practice in Bazel, since if you depend on indirect dependencies it's easy to break upstream package consumers unknowingly. At the moment layering_check is only supported by Clang. The change also changes the default CC variable to be Clang as a result.
(ref: https://bazel.build/reference/be/c-cpp)

* Changed BUILD files accordingly to support layering_check. The build fails otherwise without the changes once layering_check is enabled.
* Additionally ran Buildifier on all build files to make them consistent as a separate commit.

Additional Notes:
- We should consider adding a toolchain directly in Bazel to make it hermetic or consider using Nix to bring in workspace tooling.
-  We should add add buildifier as a github_action lint check.


.bazelrc
```diff
@@ -14,3 +14,10 @@
 
 build --cxxopt=-std=c++17
 build --host_cxxopt=-std=c++17
+# support layering_check similar to Google internal
+# this only works for Clang toolchain AFAIK
+# https://github.com/bazelbuild/bazel/pull/11440
+build --features=layering_check
+# Set the default compiler to the `clang` binary on the `PATH`.
+# TODO(fzakaria): Make this a toolchain or hermetic somehow
+build --repo_env=CC=clang
```
fzakaria, line 23: Since this is an LLVM based tool; it's probably best we build with LLVM anyways :)
Would be better to use the internal toolchain but this is as hermetic as the current approach which also uses the system toolchain.

fzakaria:
The format changes are in a separate commit if you want to view the changes without them.

mlevesquedion:
FYI I'm working on adding buildifier here (WIP): https://github.com/openxla/stablehlo/pull/1915

fzakaria:
rebased from main which was the earlier build failure.
There were new layering failures.

fzakaria:
> FYI I'm working on adding buildifier here (WIP): #1915

Nice.

mlevesquedion (approved):

PR #1918, opened 2024-01-12
Updates to shape functions enabling reuse from MHLO

sdasgup3:
The upstream change https://github.com/openxla/stablehlo/pull/1869 in StableHLO updates various API related to shape inference. MHLO shape inference functions in [hlo_ops.cc](https://github.com/openxla/xla/blob/main/xla/mlir_hlo/mhlo/IR/hlo_ops.cc) uses those APIs. The PR updates the visibility and signature of those API for a clearer integration.

Specifically, the PR does the followings:
1. **updates `getAccumulatorTypes` to return a error status when the input regions is empty**: This function is used in type inference of various reduction based operations ([eg](https://github.com/openxla/stablehlo/blob/d5b464925371092095ac934b46ba93ebd4284223/stablehlo/dialect/TypeInference.cpp#L2589)). This functions enables infering type based on the reduction block of the operation, which is something proposed in  [RFC](https://github.com/openxla/stablehlo/pull/1664). However, there could be [instances](https://github.com/openxla/xla/blob/a91877b9c9aa1edf307c5927782111b1a81cd81d/xla/mlir_hlo/mhlo/transforms/group_reduction_dimensions/group_reduction_dimensions.cc#L228) when type inference can be called with empty region in which case we would like to report a meaningful diagnostic message. 

2. **Allow `hlo::inferAllReduceOp`  to accept multiple operands information**: In stableHLO, `all_reduce` op  have a single operand ([e.g.](https://github.com/openxla/stablehlo/blob/d5b464925371092095ac934b46ba93ebd4284223/stablehlo/dialect/StablehloOps.td#L1355)), whereas in MHLO the op can take multiple operand ([e.g.](https://github.com/openxla/xla/blob/79aba0801ef75c1c2dffbb4ecc506a0d8144c9ac/xla/mlir_hlo/mhlo/IR/hlo_ops.td#L1528). The `hlo::inferAllReduceOp` signature is updated to accommodate both cases. 

3. Remove unused arguments to functions `verifyReduceOpInputsAndInferShape` and `inferReduceOp`.

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -505,8 +505,8 @@ LogicalResult verifyReplicaGroups(std::optional<Location> location,
 
 LogicalResult verifyReduceOpInputsAndInferShape(
     std::optional<Location> location, SmallVector<ShapedType> inputTypes,
-    SmallVector<ShapedType> initValueTypes, ArrayRef<int64_t> dimensions,
-    SmallVector<int64_t>& newDimensions, Attribute& encoding) {
+    ArrayRef<int64_t> dimensions, SmallVector<int64_t>& newDimensions,
+    Attribute& encoding) {
   // Check for unranked tensors in input operands.
   uint64_t numInputs = inputTypes.size();
   int64_t rankedInputIdx = -1;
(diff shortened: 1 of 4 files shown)


stablehlo/dialect/TypeInference.cpp
```diff
@@ -568,13 +568,20 @@ LogicalResult verifyReduceOpInputsAndInferShape(
 
 // Returns the types of the terminator arguments of the input  mlir::Block
 // 'block'.
-SmallVector<ShapedType> getAccumulatorTypes(Block& block) {
-  SmallVector<ShapedType> accumulatorSubShapes;
+LogicalResult getAccumulatorTypes(
```
ghpvnist, line 571: Optional nit: now that the function is not returning the accumulator types, consider renaming to `populateAccumulatorTypes`. This may or may not break places depending on how widely used this API is.

ghpvnist (approved):
Overall LGTM with minor comment.

stablehlo/dialect/TypeInference.cpp
sdasgup3: The signature of the above function is same as `inferReduceOp` [ref](https://github.com/openxla/stablehlo/blob/d5b464925371092095ac934b46ba93ebd4284223/stablehlo/dialect/TypeInference.cpp#L2572) which takes a refernece to `infereedReturnShapes` for population within function. 

With that I propose to keep the function name as is.

GleasonK: Why not return `FailureOr<SmallVector<ShapedType>>` ?

GleasonK (approved):

stablehlo/dialect/TypeInference.cpp
sdasgup3: I should take back my stance on that. done.

PR #1919, opened 2024-01-16
Leverage tablegen more for StablehloLegalizeToLinalg and other passes

mlevesquedion:
For the `StablehloLegalizeToLinalg` pass:
- Define the `dependentDialects` field instead of explicitly defining `getDependentDialects`.
- Use the generated pass declaration and registration (`GEN_PASS_DECL`, `GEN_PASS_REGISTRATION`).
- Use `GEN_PASS_DEF_*` instead of `GEN_CLASSES` (deprecated).
- A few tweaks to the way the pass struct is written.

In addition to that:
- Use `GEN_PASS_DECL` to include generated pass declarations, rather than explicitly listing the passes.
- Use `register*Passes` instead of explicitly listing the passes being registered.

In total these changes reduce duplication of functionality that can be provided by Tablegen and make our use of Tablegen more consistent across different parts of the codebase.


stablehlo/tools/StablehloOptMain.cpp
```diff
@@ -31,9 +31,8 @@ int main(int argc, char **argv) {
   mlir::hlo::registerAllTestPasses();
   mlir::stablehlo::registerPassPipelines();
   mlir::stablehlo::registerPasses();
-  mlir::stablehlo::registerStablehloLegalizeToLinalgPass();
-  mlir::tosa::registerStablehloLegalizeToTosaPassPass();
-  mlir::tosa::registerStablehloPrepareForTosaPassPass();
```
GleasonK, line 36: What happened to this one? Is it bundled in the above register method?

GleasonK (approved):
Can this cause any breakages in downstream code? Maybe f `registerStablehloPrepareForTosaPassPass`  no longer exists?

mlevesquedion:
Shouldn't cause any breakages, at least for now, e.g. here's the generated code for TOSA: 

```
// StablehloPrepareForTosaPass Registration

inline void registerStablehloPrepareForTosaPass() {
  ::mlir::registerPass([]() -> std::unique_ptr<::mlir::Pass> {
    return createStablehloPrepareForTosaPass();
  });
}

// Old registration code, kept for temporary backwards compatibility.
inline void registerStablehloPrepareForTosaPassPass() {
  ::mlir::registerPass([]() -> std::unique_ptr<::mlir::Pass> {
    return createStablehloPrepareForTosaPass();
  });
}
```

The code comment says `temporary backwards compatibility`, so not sure how long that will be. So it could cause breakages at some point when they decide to remove it. IMO that's fine since we'll be following the flow of MLIR, but LMK if you want me to keep this.

stablehlo/tools/StablehloOptMain.cpp
mlevesquedion: Eh. This is annoying. I thought clicking on "Resolve conversation" with a comment pending would send the comment.

The answer to your question is yes. Here's the relevant generated code (for TOSA, but linalg is similar):

```
inline void registerStablehloTOSATransformsPasses() {
  registerStablehloLegalizeToTosaPass();
  registerStablehloPrepareForTosaPass();
}
```

PR #1923, opened 2024-01-16
Remove separate cmake_python_api github action

fzakaria:
Remove separate cmake_python_api github action and have ci_build_cmake and ci_build_cmake_llvm always create the python bindings.
    
This should cut down on rebuilding LLVM itself which takes ~5 minutes with ccache and much longer on new builds.

PR #1926, opened 2024-01-18
Make building Python bindings optional

fzakaria:
Many developers are using the ci_build_* scripts to build LLVM and StableHLO.

A previous PR #1923 made the CI system by default building python bindings however there is an undocumented step of installing Python dependencies such as pybind.

To simplify local development, make the option configurable with the default to OFF so that the ci_build_* actions work without any additional setup.

The CI system itself (GitHub actions) set the appropriate variables to build the Python bindings.


build_tools/github_actions/ci_build_cmake.sh
```diff
@@ -27,21 +27,25 @@ fi
...
   -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
   -DCMAKE_C_COMPILER=clang \
   -DCMAKE_C_COMPILER_LAUNCHER=ccache \
   -DSTABLEHLO_ENABLE_STRICT_BUILD=ON \
-  -DSTABLEHLO_ENABLE_BINDINGS_PYTHON=ON
+  -DSTABLEHLO_ENABLE_BINDINGS_PYTHON="${STABLEHLO_ENABLE_BINDINGS_PYTHON}"
 
 # Build and Test StableHLO
-cd "$STABLEHLO_BUILD_DIR"
+cd "$STABLEHLO_BUILD_DIR" || exit
```
fzakaria, line 50: This was an error ShellCheck (https://www.shellcheck.net/) gave me in my editor.
Could be yet another lint to apply :)

mlevesquedion: Neat! I personally like setting `set -o errexit` (or `set -e` for short) at the very top of the script. In this case if `cmake` fails we probably don't want to call `ninja` either.

While we're on this, I also prefer `cmake --build "$STABLEHLO_BUILD_DIR" --target check-stablehlo-ci`, but calling `cd`ing and calling `ninja` is also fine.

mlevesquedion (approved):
Nice, thank you!

fzakaria:
image

Wanted to confirm this still does the right thing.

build_tools/github_actions/ci_build_cmake.sh
fzakaria: Yes we should probably apply best practices for shellcheck;
I also like the `cmake --build` approach but it's more "new"

PR #1928, opened 2024-01-18
Add missing copyright message

sdasgup3:
The copyright message was missing `docs/_toc.yaml`.


mlevesquedion (approved):
Nit: Did you mean "copyright"? LGTM

Note: maybe we should have a presubmit check for this :thinking:

sdasgup3:
Ah! yes that was a typo.
 
>> Note: maybe we should have a presubmit check for this
agreed. Let us begin with an issue on this one.

PR #1074, opened 2023-02-07
Add copyright tag to bazel configuration files

sdasgup3:

PR #1929, opened 2024-01-18
Add licensecheck lint to CI and add missing licenses

mlevesquedion:
For the most part, `addlicense` handles the file types that we care about. It also doesn't handle some files that we (AFAICT) don't want license headers in: `.md` and `.mlir` files, e.g. However, it doesn't handle some files for which we may want license headers: `.bazel`, some `.in` files, `.pdll` and `.td`. I think it's debatable whether or not we want license headers in the following: `.bazelignore`, `.bazelrc`, `.bazelversion`, `.gitignore`, `.clang-format`, and these are also not handled by `addlicense`.

For `.bazel`, `.pdll` and `.td`, I will see if I can contribute to `addlicense` to handle those. For `.in` files (e.g. `lit.site.cfg.py.in`), I think whether or not we want a license actually depends on what the "true" type of the file is (e.g. `.py`). Supporting that would likely require additional changes to `addlicense`. Actually, I think it would be great if `addlicense` had some kind of extensibility mechanism (https://github.com/google/addlicense/issues/50).

Anyway, although imperfect I think on the whole this should help us catch missing licenses in a lot of cases we care about.

Inspired by https://github.com/openxla/stablehlo/pull/1928

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,3 +1,16 @@
+# Copyright 2024 The StableHLO Authors.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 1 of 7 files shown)


sdasgup3:
LGTM! I agree that PR is useful on its own  irrespective of the file tyeps left out. 

>> I think it's debatable whether or not we want license headers in the following: .bazelignore, .bazelrc, .bazelversion

Some of the these add licensing requirements are coming from integration presubmit checks (which runs internally). For example  https://github.com/openxla/stablehlo/pull/1074/files add missing copyright tags to bazel* files. That could be seen as an motivation.

sdasgup3 (approved):

mlevesquedion:
Ah good to know, thanks! Maybe I can add those to `addlicense` too in a follow-up.

PR #1940, opened 2024-01-23
Emit dense arrays and use dense arrays in tests

mlevesquedion:
The purpose of this change is to ease the transition from I64DenseArrayOrElements1DAttr to DenseI64ArrayAttr. The attrs themselves will be changed in a follow-up.


GleasonK (approved):
How confident are we that we got everything? Have you tested this change after flipping the attribute from `ArrayOrElements --> Array`?

Also hold off on merge until @ghpvnist's approval since this will have integration implications.

ghpvnist:
I've completed my first integrate so this PR is good to go!

ghpvnist (approved):

mlevesquedion:
> How confident are we that we got everything? Have you tested this change after flipping the attribute from `ArrayOrElements --> Array`?
> 
> Also hold off on merge until @ghpvnist's approval since this will have integration implications.

I would say very confident. I flipped the attributes and fixed all the tests, then I split off the test changes.

PR #1954, opened 2024-01-26
add canonicalize of DynamicReshapeOp.

linuxlonelyeagle:


linuxlonelyeagle:
@ghpvnist Can you review the code for me? Or if you have any comments, let me know, I'd love to know what you guys think about this PR.

linuxlonelyeagle:
I have some questions? If this PR makes sense, do I need to add some files for testing ? If there is one in mlir-hlo, where should I find it. Besides that, the canonicalize addition I refer to mlir-hlo, but now this PR doesn't look elegant, too many files added to add canonicalize, do you have any idea?

GleasonK:
Thank you for the PR! Could you say more about your use case for StableHLO canonicalization? Does your project depend directly on the StableHLO repo and not on openxla/xla?

The reason I ask, is that currently StableHLO has no canonicalizations, as it is was decided in a previous community discussion that StableHLO is an input dialect and transformations go on MHLO. Personally, I'm in favor of having opt-in StableHLO canonicalization patterns like this in the repo, as it avoids other projects duplicating patterns - for example here are [IREE canonicalization patterns for StableHLO](https://github.com/openxla/iree/tree/f3b008c6db310f787ad76f151c21a30f72b14794/compiler/plugins/input/StableHLO/stablehlo-iree/Conversion/Preprocessing) which could also be upstreamed and shared among the StableHLO community.

In the current setup, all canonicalization is done on MHLO in [xla/mlir_hlo/mhlo/transforms](https://github.com/openxla/xla/tree/main/xla/mlir_hlo/mhlo/transforms). This assumes that a project depends on XLA, which isn't always the case. Before we submit this change, we'll need to start a discussion on the topic in [OpenXLA Discuss](https://groups.google.com/a/openxla.org/g/openxla-discuss) to ensure community alignment.

I was hoping to start up a conversation about this exact topic in the coming month, so I'm glad to see this contribution :). Perhaps we can collaborate on that openxla-discuss post!

cc @jpienaar who was also interested in upstreaming some patterns to the repo.

GleasonK (changes requested):
Adding blocking review until we establish community precedence for canonicalizations in StableHLO.

linuxlonelyeagle:
@GleasonK I apologize for being a little late in replying to you, I started a discuss [here](https://github.com/orgs/openxla/discussions/106).

GleasonK:
Great! Thank you! I'm getting the right people to take a look at it, will report back tomorrow with an update!

linuxlonelyeagle:
close by https://github.com/openxla/stablehlo/pull/1984.

PR #1956, opened 2024-01-26
Enhance rendezvous docs and fix race condition

ghpvnist:
The PR removes the use of `shared_ptr`s as it complicates the logic of `rendezvous`. Instead of relying on the use count of `shared_ptr`, I've added a variable to track the state separately instead. Now, there is no need to worry about the scope of the `shared_ptr`s and cleaning up resource before exiting.

Thanks @mlevesquedion for sharing code snippets offline to help simplify this!

### DIFF
--- a/stablehlo/reference/ProcessGrid.cpp
+++ b/stablehlo/reference/ProcessGrid.cpp
@@ -20,8 +20,6 @@ limitations under the License.
 #include <optional>
 #include <utility>
 
-#include "llvm/ADT/STLExtras.h"
-#include "llvm/Support/Error.h"
 #include "stablehlo/reference/Tensor.h"
 
 namespace mlir {
@@ -226,32 +224,28 @@ RendezvousResult ProcessGrid::rendezvous(ProcessGroup processGroup,
                                          ChannelId channelId,
                                          ProcessId processId,
                                          const Tensor &operand) {
-  std::pair<ProcessGroup, ChannelId> channelKey(processGroup, channelId);
   // Process wait/notify logic below doesn't work for single process.
   if (processGroup.size() == 1)
     return RendezvousResult({std::pair{processId, operand}});
 
+  std::pair<ProcessGroup, ChannelId> channelKey(processGroup, channelId);
   auto &state = channels_[channelKey];
 
   std::unique_lock<std::mutex> lock(state.mutex);
   state.values[processId] = operand;
-  if (state.values.size() == processGroup.size()) {
-    state.result = state.values;
-    state.values.clear();
+  state.useCount++;
 
-    // This will notify all waiting processes, but they will be blocked until
-    // the lock held by this process is released in the wait() call below.
-    channelConditions_[channelKey].notify_all();
-    return state.result;
-  } else {
-    // Two cases:
-    // 1) The condition is not true and wait.
-    // 2) This process will wake up, check that the lock is still held by the
-    //    last process, and stay in a blocked state until it can acquire the
-    //    lock. Once lock is acquired, proceed below.
+  // After each process contributes, wait for the last process to notify.
+  if (state.values.size() < processGroup.size()) {
     channelConditions_[channelKey].wait(lock);
-    return state.result;
+  } else {
+    state.result = std::move(state.values);
+    channelConditions_[channelKey].notify_all();
   }
+
+  state.useCount--;
+
+  return state.useCount > 0 ? state.result : std::move(state.result);
 }
 
 void ProcessGrid::send(ArrayRef<Tensor> inputs, ChannelId channelId,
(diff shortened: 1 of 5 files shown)


mlevesquedion:
Looks like I hit this race condition in #1960: https://github.com/openxla/stablehlo/actions/runs/7675148587/job/20920959254?pr=1960

mlevesquedion:
Hit it again: https://github.com/openxla/stablehlo/actions/runs/7703298470/job/20993390030#step:6:437

I'm really unlucky :upside_down_face:

GleasonK:
Based on Michaels comments I'm not sure if this fixes the issue?

GleasonK (approved):

GleasonK:
Ah - or did I misunderstand, perhaps Michael is hitting this because this fix isnt in.

ghpvnist:
Yes, it's the latter. Though, I'm thinking about simplifying this so I'll hold off on merging just yet.

stablehlo/reference/ProcessGrid.h
```diff
@@ -42,15 +67,20 @@ namespace detail {
 /// Once all processes have added their data, the data in `values` is moved to
 /// `result` that multiple processes can concurrently read from.
 struct RendezvousState {
+ public:
```
mlevesquedion, line 70: `public` is implied by `struct`

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -220,72 +220,32 @@ SmallVector<Tensor> ProcessGrid::recv(ChannelId channelId,
...
-    // The last process to contribute takes the result from the state to allow
-    // the process that contributed last to exit the function.
-    auto result = std::move(state.result);
-    channelConditions_[channelKey].notify_one();
-    return result;
+  state.useCount++;
+
+  // After each process contributes, wait for the last process to notify.
+  if (state.values.size() < processGroup.size()) {
+    channelConditions_[channelKey].wait(lock);
```
mlevesquedion, line 240: Should we have a limit on the wait time? Used to be 3 seconds.

stablehlo/reference/ProcessGrid.h
ghpvnist: Thanks, I missed updating this while debugging :)

stablehlo/reference/ProcessGrid.cpp
mlevesquedion: (The 3 seconds is documented in `ProcessGrid.h` too, so consider updating that if you decide to remove the wait limit.)

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -220,72 +220,32 @@ SmallVector<Tensor> ProcessGrid::recv(ChannelId channelId,
...
-    if (state.result.use_count() > static_cast<int64_t>(processGroup.size()))
-      llvm::report_fatal_error(
-          "Each process should have only one shared access to the result.");
-
-    // The last process to contribute takes the result from the state to allow
-    // the process that contributed last to exit the function.
-    auto result = std::move(state.result);
-    channelConditions_[channelKey].notify_one();
-    return result;
+  state.useCount++;
```
mlevesquedion, line 236: Doesn't `state.result` already have a `use_count` helper? Would it be simpler to use that instead?

GleasonK (changes requested):
Going to re-review with code changes. Adding blocking review until then.

stablehlo/reference/ProcessGrid.cpp
ghpvnist: With the refactoring, we don't have shared_ptrs anymore. `state.result` is a map (size only increasing) so we can't rely on that.

mlevesquedion: Ah yep my bad, didn't realize this was wrapped in a `shared_ptr` at first. (Also I didn't mean to resolve this without commenting. I really wish the GH UI would either submit the comment or at least warn when you have a pending comment and you hit "resolve".

mlevesquedion (approved):
LGTM

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -220,72 +220,35 @@ SmallVector<Tensor> ProcessGrid::recv(ChannelId channelId,
...
-
-    // The last process to contribute takes the result from the state to allow
-    // the process that contributed last to exit the function.
-    auto result = std::move(state.result);
-    channelConditions_[channelKey].notify_one();
-    return result;
+            lock, std::chrono::seconds(3),
+            [&] { return state.values.size() == processGroup.size(); }))
+      llvm::report_fatal_error("rendezvous timed out");
+  } else {
```
GleasonK, line 244: If stmt with an exit in the body, do we need the `else`?

GleasonK: Ah. Resolving. This is a nested if.

GleasonK (approved):

PR #1958, opened 2024-01-26
Avoid explicitly constructing compatible attributes in a few places

mlevesquedion:
Attrs will be implicitly copied. Avoid explicitly constructing the attrs as this introduces a dependency on the type of the attrs, meaning the code will need to be changed if the attrs' type changes.


ghpvnist (approved):
Nice find!

sdasgup3 (approved):

PR #1959, opened 2024-01-27
Add instrumentation for Google Sanitizers

fzakaria:
Google Sanitiziers (https://github.com/google/sanitizers) have been integrated into Clang & GNU.

Add support for ASan + UBSan if a CMake option is set. This should help catch some of the same errors that can be provided typicaly within internal Google development.

TODO: Add support for TSan which needs to run individually since it cannot be instrumented alongside ASan.


CMakeLists.txt
```diff
@@ -40,6 +40,7 @@ endif()
 option(STABLEHLO_BUILD_EMBEDDED "Build StableHLO as part of another project" OFF)
 option(STABLEHLO_ENABLE_BINDINGS_PYTHON "Enables StableHLO Python bindings" OFF)
 option(STABLEHLO_ENABLE_STRICT_BUILD "Build StableHLO with strict warnings and warnings as errors" OFF)
+option(STABLEHLO_ENABLE_SANITIZER "Enable a sanitizer [OFF, address, memory, thread]" OFF)
```
mlevesquedion, line 43: Nit: `memory` and `thread` aren't supported yet. Remove them?

CMakeLists.txt
```diff
@@ -130,6 +131,25 @@ include_directories(${CMAKE_CURRENT_BINARY_DIR})
...
 
+
+#-------------------------------------------------------------------------------
+# Sanitizier configuration
+#-------------------------------------------------------------------------------
+
+if(STABLEHLO_ENABLE_SANITIZER)
+  string(TOLOWER "${STABLEHLO_ENABLE_SANITIZER}" STABLEHLO_ENABLE_SANITIZER_LOWERCASE)
+  if (STABLEHLO_ENABLE_SANITIZER_LOWERCASE STREQUAL "address")
+    add_compile_options(-fsanitize=address -fsanitize=undefined -fsanitize=leak -fno-omit-frame-pointer)
```
mlevesquedion, line 142: To my untrained eye, this looks like it's turning on more than just the `address` sanitizer. Is this intended?

mlevesquedion (approved):

CMakeLists.txt
fzakaria: Yes -- they seem to be a common coupling from what I found online researching best practice.
The grouping otherwise is when sanitizers themselves can't run together in same session.
Memory & Address for instance are such a duo -- same for Thread.

fzakaria:
Need to investigate more why this is failing on GitHub CI -- some discrepancy as it builds on my machine.
I did notice that passing the python tests however need additional environment variables for two reasons:
1. ASan needs to be first in the link order list to make sure it grabs malloc. Since we aren't linking Python executable itself we have to use LD_PRELOAD
2. Python itself seems to leak objects already so it should be turned off by default.

```
ASAN_OPTIONS="detect_leaks=0" LD_PRELOAD=/lib/x86_64-linux-gnu/libasan.so.8 ninja check-stablehlo-python
```

fzakaria:
@mlevesquedion this adds 2minutes to our build :S

Let me know if the comments make sense.
It caught nothing... which is good?

fzakaria:
I'll write a mock unit test that should fail and upload the expected failure and what it looks like with ASAN enabled to prove it's working.

CMakeLists.txt
```diff
@@ -130,6 +139,14 @@ include_directories(${CMAKE_CURRENT_BINARY_DIR})
 link_directories(${LLVM_BUILD_LIBRARY_DIR})
 add_definitions(${LLVM_DEFINITIONS})
 
+
+#-------------------------------------------------------------------------------
+# Sanitizier configuration
```
mlevesquedion, line 144:
```suggestion
# Sanitizer configuration
```

cmake/SetupSanitizers.cmake
```diff
@@ -0,0 +1,56 @@
...
+        return()
+    endif ()
+
+    string(TOLOWER "${STABLEHLO_ENABLE_SANITIZER}" STABLEHLO_ENABLE_SANITIZER_LOWERCASE)
+    if (STABLEHLO_ENABLE_SANITIZER_LOWERCASE STREQUAL "OFF")
+        return()
+    endif ()
+
+    if (STABLEHLO_ENABLE_BINDINGS_PYTHON)
+        message(FATAL_ERROR "Sanitizer setting not supported for Python binding builds")
```
mlevesquedion, line 44:
```suggestion
        message(FATAL_ERROR "STABLEHLO_ENABLE_SANITIZER must be set to OFF when building Python bindings")
```

cmake/SetupSanitizers.cmake
```diff
@@ -0,0 +1,56 @@
...
+# the Python bindings for sanitization builds.
+# TODO(fzakaria): Revisit a better way or if this can be solved.
+# @see: https://github.com/google/sanitizers/wiki
+function(setup_sanitizers)
+    if (NOT STABLEHLO_ENABLE_SANITIZER)
+        return()
+    endif ()
+
+    string(TOLOWER "${STABLEHLO_ENABLE_SANITIZER}" STABLEHLO_ENABLE_SANITIZER_LOWERCASE)
+    if (STABLEHLO_ENABLE_SANITIZER_LOWERCASE STREQUAL "OFF")
```
mlevesquedion, line 39:
```suggestion
    if (STABLEHLO_ENABLE_SANITIZER_LOWERCASE STREQUAL "off")
```

mlevesquedion (approved):
LGTM and IMO adding 2 minutes to the build time is fine.

fzakaria:
Here is an example of me adding an explicit leak for instance.
```
diff --git a/stablehlo/tools/StablehloOptMain.cpp b/stablehlo/tools/StablehloOptMain.cpp
index 6e4b78e4..13a8912d 100644
--- a/stablehlo/tools/StablehloOptMain.cpp
+++ b/stablehlo/tools/StablehloOptMain.cpp
@@ -40,6 +40,11 @@ int main(int argc, char **argv) {
   mlir::stablehlo::registerAllDialects(registry);
   registry.insert<mlir::stablehlo::interpreter::InterpreterDialect>();
 
+  char *s = (char *) malloc(100);
+  strcpy(s, "Hello world!");
+  printf("%s\n", s);
+
   return failed(
       mlir::MlirOptMain(argc, argv, "StableHLO optimizer driver\n", registry));
 }
```

It produces the following:
```
❯ ./bin/stablehlo-opt ../stablehlo/tests/infer_chlo.mlir
Hello world!
../stablehlo/tests/infer_chlo.mlir:14:49: error: operation being parsed with an unregistered dialect. If this is intended, please use -allow-unregistered-dialect with the MLIR tool used
  %1 = "hlo_test_infer.reify_return_type_shapes"(%0) : (tensor<?xf32>) -> tensor<1xindex>
                                                ^

=================================================================
==2962483==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 100 byte(s) in 1 object(s) allocated from:
    #0 0x555939fb8bf2 in malloc (/usr/local/google/home/fmzakari/code/github.com/openxla/stablehlo/build/bin/stablehlo-opt+0x2057bf2) (BuildId: edd01817d5d78c3e)
    #1 0x555939fecad1 in main (/usr/local/google/home/fmzakari/code/github.com/openxla/stablehlo/build/bin/stablehlo-opt+0x208bad1) (BuildId: edd01817d5d78c3e)
    #2 0x7fb5b34456c9 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16

SUMMARY: AddressSanitizer: 100 byte(s) leaked in 1 allocation(s).
```

fzakaria:
@mlevesquedion thanks for the great feedback and looking over -- pinged you again for one final look over.
It was actually kind of hard to trigger the leak -- because I kept adding code that got optimized away :)

mlevesquedion (approved):

PR #1960, opened 2024-01-27
Use reference interpreter to eval ops in refine shapes pass

mlevesquedion:
This change removes some duplication between the interpreter and the refine shapes pass. I didn't go all the way (there is still some duplication, e.g. the slice op is reimplemented in the refine shapes pass), because I'm not sure if the purpose of the interpreter is _only_ to be used as a reference, or if we want to use it for passes too.

The implementation in the interpreter is of course more generic than what is needed here, and we likely lose some efficiency due to the type conversions. However, I think reducing duplication is more important.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -927,6 +927,8 @@ cc_library(
         ":base",
         ":chlo_ops",
         ":interpreter_ops",
+        ":reference_ops",
+        ":reference_tensor",
         ":stablehlo_ops",
         ":stablehlo_ops_inc_gen",
         ":stablehlo_pass_inc_gen",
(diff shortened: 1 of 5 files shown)


GleasonK (changes requested):
We need to think about this a bit more before we rely on the reference interpreter in production use cases. I'm thinking we table this for now. A better refactoring may be to have opt-in folder methods that both reference and this end up using.

mlevesquedion:
> We need to think about this a bit more before we rely on the reference interpreter in production use cases. I'm thinking we table this for now. A better refactoring may be to have opt-in folder methods that both reference and this end up using.

That SGTM, thanks! I'll close the PR since based on your comment I think it's unlikely that we'll adopt this. We can always reopen it later if we change our minds.

PR #1961, opened 2024-01-27
Per-axis specification for `dot-general` and `add`

sdasgup3:
closes https://github.com/openxla/stablehlo/issues/1574
The PR implement the spec changes to support per-axis quantization scheme for `add` and `dot_general` ops. Here is the rationale for the support:

  - For `add` op:  Used for bias addition following convolution. Bias and [convolution op](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#semantics-25) outputs can  both be per-axis quantized and have the same shape, and same quantization parameters, but current `stablehlo.add` only support per-tensor quantized scheme. [Relevant dicussion](https://github.com/openxla/stablehlo/issues/1574#issuecomment-1659466321)
  - For `dot_general`:  Some uses cases of quantized models requires  support for a mixed quantization scheme where it accepts per-tensor quantized `lhs` and per-axis quantized `rhs`. The quantization schemes (per-axis/per-tensor) proposed for `dot_general` op is similar to the convolution. [Relevant discussion](https://github.com/openxla/stablehlo/issues/1574#issuecomment-1676914854)
     - We do not expect the the contracting dimension `rhs_contracting_dimensions` to include the `quantization dimension(rhs)`. The multiply-accumulation step for dot_general iterate over the elements along the contracting dimensions. Having a uniform quantization parameters (scales and zero-points) for those elements allows full-integer computations (refer to section 2.2. of https://arxiv.org/pdf/1712.05877.pdf). However, if the quantization parameters are different within the  contracting dimensions, which is what will happen if we allow `quantization_dimension(rhs)` to be in `rhs_contracting_dimensions`, then the computation must rescale integers (or dequantize) while doing multiply-accumulation which is not desirable. 

Looking forward to feedback!
cc @dansuh17, @loganchien, @rewu93, @dominicsymes

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -627,22 +627,28 @@ Performs element-wise addition of two tensors `lhs` and `rhs` and produces a
 
 #### Inputs
 
-| Label | Name  | Type                       | Constraints |
-|-------|-------|----------------------------|-------------|
-| (I1)  | `lhs` | tensor or quantized tensor | (C1-C2)     |
-| (I2)  | `rhs` | tensor or quantized tensor | (C1-C2)     |
+| Label | Name  | Type                       | Constraints   |
+|-------|-------|----------------------------|---------------|
+| (I1)  | `lhs` | tensor or quantized tensor | (C1-C4)       |
+| (I2)  | `rhs` | tensor or quantized tensor | (C1-C3), (C5) |
 
 #### Outputs
 
 | Name     | Type                       | Constraints |
 |----------|----------------------------|-------------|
-| `result` | tensor or quantized tensor | (C1-C2)     |
+| `result` | tensor or quantized tensor | (C1-C5)     |
 
 #### Constraints
 
 * (C1) `baseline_type(lhs) = baseline_type(rhs) = baseline_type(result)`.
-* (C2) If the operation uses quantized tensors, then
-  `is_quantized(lhs) and is_quantized(rhs) and is_quantized(result)`.
+* If the operation uses quantized tensors:
+  * (C2) `is_quantized(lhs) and is_quantized(rhs) and is_quantized(result)`.
+  * (C3) `(is_per_axis_quantized(lhs) or is_per_axis_quantized(rhs)) =
+    is_per_axis_quantized(result)`.
+  * (C4) If `is_per_axis_quantized(lhs)`, then `quantization_dimension(lhs) =
+    quantization_dimension(result)`.
+  * (C5) If `is_per_axis_quantized(rhs)`, then `quantization_dimension(rhs) =
+    quantization_dimension(result)`.
 
 #### Examples
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -629,25 +629,28 @@ Performs element-wise addition of two tensors `lhs` and `rhs` and produces a
 
 | Label | Name  | Type                       | Constraints   |
 |-------|-------|----------------------------|---------------|
-| (I1)  | `lhs` | tensor or quantized tensor | (C1-C4)       |
-| (I2)  | `rhs` | tensor or quantized tensor | (C1-C3), (C5) |
+| (I1)  | `lhs` | tensor or quantized tensor | (C1-C6)       |
+| (I2)  | `rhs` | tensor or quantized tensor | (C1-C5), (C7) |
 
 #### Outputs
 
 | Name     | Type                       | Constraints |
 |----------|----------------------------|-------------|
-| `result` | tensor or quantized tensor | (C1-C5)     |
+| `result` | tensor or quantized tensor | (C1-C7)     |
 
 #### Constraints
 
-* (C1) `baseline_type(lhs) = baseline_type(rhs) = baseline_type(result)`.
+* If the operation uses non-quantized tensors:
+  * (C1) `type(lhs) = type(rhs) = type(result)`.
 * If the operation uses quantized tensors:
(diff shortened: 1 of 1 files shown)


docs/spec.md
```diff
@@ -627,16 +627,16 @@ Performs element-wise addition of two tensors `lhs` and `rhs` and produces a
...
 #### Inputs
 
-| Label | Name  | Type                                  | Constraints |
-|-------|-------|---------------------------------------|-------------|
-| (I1)  | `lhs` | tensor or per-tensor quantized tensor | (C1)        |
-| (I2)  | `rhs` | tensor or per-tensor quantized tensor | (C1)        |
+| Label | Name  | Type                       | Constraints |
+|-------|-------|----------------------------|-------------|
+| (I1)  | `lhs` | tensor or quantized tensor | (C1)        |
+| (I2)  | `rhs` | tensor or quantized tensor | (C1)        |
```
loganchien, line 633: Add a constraints?

* `(is_per_axis_quantized(lhs) or is_per_axis_quantized(rhs)) == is_per_axis_quantized(result)`
* If `is_per_axis_quantized(lhs)`, then `quantization_dimension(lhs) == quantization_dimension(result)`
* If `is_per_axis_quantized(rhs)`, then `quantization_dimension(rhs) == quantization_dimension(result)`

For record, here is the motivation for the suggetsion https://github.com/openxla/stablehlo/pull/1446#discussion_r1283425049

loganchien (approved):

docs/spec.md
doyeonkim0: Do we allow adding per-axis quantized tensor and per-tensor quantized tensor? 
By C1, I think such cases would be rejected by having different quantization_dimension.

Also, curious about the motivation to allow operand with different quantization parameters.

loganchien: @doyeonkim0 Thank you for pointing this out.  After reading the definition of `baseline_type` (i.e. `quantization_dimension` is guaranteed to be equal) again, I agree that `baseline_type(lhs) == baseline_type(rhs) == baseline_type(result)` can be sufficient for per-axis bias addition use case (which essentially guarantees that `lhs`, `rhs`, and `result` uses the same quantization dimension, scales, and zero points).

loganchien (approved):

docs/spec.md
sdasgup3: @doyeonkim0 Thanks for pointing it point. I have updated the spec addressing your concern. The `baseline_type` should be used only when all the operands and results under comparison are non-quantized or per-tensor quantized.  

Yes, we do allow a mix of per-tensor and per-axis cases. 

>> Also, curious about the motivation to allow operand with different quantization parameters.
This is more of a flexibly added to almost all the quantized operations to have different quantization parameters.  This was possible mostly because the dq-op-q semantics for these ops.

doyeonkim0 (approved):

rewu93 (approved):

docs/spec.md
```diff
@@ -627,20 +627,31 @@ Performs element-wise addition of two tensors `lhs` and `rhs` and produces a
 
 #### Inputs
 
-| Label | Name  | Type                                  | Constraints |
-|-------|-------|---------------------------------------|-------------|
-| (I1)  | `lhs` | tensor or per-tensor quantized tensor | (C1)        |
-| (I2)  | `rhs` | tensor or per-tensor quantized tensor | (C1)        |
+| Label | Name  | Type                       | Constraints   |
+|-------|-------|----------------------------|---------------|
+| (I1)  | `lhs` | tensor or quantized tensor | (C1-C6)       |
```
GleasonK, line 632: I think this (along with @abhigunj's work on verifying quantized types) are good motivation for reworking how we spec quantized types. Namely let's make a very clear non-terminal split for QuantizedTensor, PerTensorQuantizedTensor, and PerAxisQuantizedTensor.

This can be done in a followup, but now that "per-tensor" and "quantized tensor" have different meanings, let's match it in the spec. This also will make the spec and VHLO align more closely.

docs/spec.md
```diff
@@ -627,20 +627,31 @@ Performs element-wise addition of two tensors `lhs` and `rhs` and produces a
...
+| `result` | tensor or quantized tensor | (C1-C7)     |
 
 #### Constraints
 
+* If the operation uses non-quantized tensors:
 * (C1) `baseline_type(lhs) = baseline_type(rhs) = baseline_type(result)`.
+* If the operation uses quantized tensors:
+  * (C2) `is_quantized(lhs) and is_quantized(rhs) and is_quantized(result)`.
+  * (C3) `storage_type(lhs) = storage_type(rhs) = storage_type(result)`.
+  * (C4) `expressed_type(lhs) = expressed_type(rhs) = expressed_type(result)`.
```
GleasonK, line 648: Isn't this covered in baseline type?

docs/spec.md
```diff
@@ -627,20 +627,31 @@ Performs element-wise addition of two tensors `lhs` and `rhs` and produces a
...
-| `result` | tensor or per-tensor quantized tensor | (C1)        |
+| Name     | Type                       | Constraints |
+|----------|----------------------------|-------------|
+| `result` | tensor or quantized tensor | (C1-C7)     |
 
 #### Constraints
 
-* (C1) `baseline_type(lhs) = baseline_type(rhs) = baseline_type(result)`.
+* If the operation uses non-quantized tensors:
+* (C1) `type(lhs) = type(rhs) = type(result)`.
```
GleasonK, line 644: Add space?

GleasonK (approved):

docs/spec.md

Issue #879, opened 2023-01-04
Figure out the number of elements in precision_config


burmako:
Recently we've added a verifier for precision_config to check that it contains either 0 or 2 elements, in accordance with the spec. However, it turns out that there are some producers that create precision configs with 1 element. Let's better understand this and decide on how/if the spec should be updated.

ghpvnist:
Here are the results from my investigation for the number of elements in `precision_config`:
* 0 values mean default precision will be applied to `lhs` and `rhs`.
* 1 value means the single precision will be applied to both `lhs` and `rhs`.
* 2 values mean one precision is applied to `lhs` and the other to `rhs`.
* Anything more is not valid.

The 2 value-variant covers for 1 value-variant, so we can decide to only allow 0 or 2 `precision_config` values. However, this would mean that if `lhs` and `rhs` both have the same precision, specifying it twice would be redundant. Allowing 1 value-variant would not hurt the semantics of StableHLO while maintaining readability of this attribute with the spec clearly documenting what different number of `precision_config` means.

burmako:
Thank you for figuring out the semantics of 1 precision config! 

Given this information, I would propose that we don't allow 1 precision config. It's a tradeoff between complicating the spec and complicating producer code and StableHLO printouts, and I think it makes sense to prioritize the spec given that the latter complications are fairly minor.

Given that our opset is so big, I think it's important to be vigilant about spec complexity because 100x small additions can easily snowball into a significant increase in complexity levels. Now, if allowing 1 precision config enabled a new feature that cannot be expressed otherwise, the argument in favor would be stronger.

burmako:
There's also a question of whether we should allow 0 precision configs, and I would propose that the answer should be similarly "No" because this doesn't add expressive power but complicates the spec (so far, we've stayed away from having an opinion about default values in the spec).

As far as the StableHLO dialect goes, precision_config is already optional, so if producers want to optimize for convenience, they can skip specifying this attribute altogether:

```
def StableHLO_PrecisionConfigAttr:
    OptionalAttr<
          TypedArrayAttrBase<StableHLO_PrecisionAttr, "Precision Config attribute">>;
```

Issue #445, opened 2022-11-04
Consider changing PrecisionConfigAttr to be more strongly typed


burmako:
It looks like the intended use of `PrecisionConfigAttr` is to have either 0 or 2 `PrecisionAttr` values - one for the `lhs`, and one for the `rhs` (or none at all, and then `DEFAULT` is going to be used). Let's consider removing this attr and introducing optional attributes `lhs_precision` and `rhs_precision`.

sdasgup3:
At the moment, https://github.com/openxla/stablehlo/pull/441 will be adding the missing constraints on PrecisionConfigAttr as is. With that the verifer for affected ops (DotGeneral, and Convolution op) can be in sync with the StableHLO spec w.r.t this attribute.

sdasgup3:
The following definition
```
def StableHLO_PrecisionConfigAttr:
    OptionalAttr<
          TypedArrayAttrBase<StableHLO_PrecisionAttr, "Precision Config attribute">>;
```

allows multiple ways to represent empty `precision_config` attribute, which is confusing.

In case we chose to keep `StableHLO_PrecisionConfigAttr` as is (as part of resolving the PR), we need to make sure to avoid the representation issue.

PR #1964, opened 2024-01-29
Add interpreter for ConvolutionOp

### LINKED ISSUE #970: Add interpreter for convolution

ghpvnist:
Since the interpreter implementation is already a mouthful, I've split #1314 to separate the implementation from the remaining checklist items. Once this is merged, I'll work on the remaining to update the rest of the docs.

closes #970

### DIFF
--- a/stablehlo/reference/Index.h
+++ b/stablehlo/reference/Index.h
@@ -121,6 +121,8 @@ using Index = Sizes;
 class IndexSpaceIterator {
  public:
   /// \name Constructor
+  IndexSpaceIterator(Sizes shape) : shape_(shape) { index_ = std::nullopt; }
+
   IndexSpaceIterator(Sizes shape, std::optional<Index> index)
       : shape_(shape), index_(index) {
     if (index && !index->inBounds(shape)) index_ = std::nullopt;
--- a/stablehlo/reference/Index.h
+++ b/stablehlo/reference/Index.h
@@ -124,8 +124,8 @@ class IndexSpaceIterator {
   IndexSpaceIterator(Sizes shape) : shape_(shape) { index_ = std::nullopt; }
 
   IndexSpaceIterator(Sizes shape, std::optional<Index> index)
-      : shape_(shape), index_(index) {
-    if (index && !index->inBounds(shape)) index_ = std::nullopt;
+      : shape_(shape), index_(std::nullopt) {
+    if (index && index->inBounds(shape)) index_ = index;
   }
 
   /// Get the current index.
--- a/stablehlo/reference/Ops.cpp
+++ b/stablehlo/reference/Ops.cpp
@@ -54,6 +54,27 @@ Index evalIndex(Tensor tensor) {
   return result;
 }
 
+Tensor evalDotGeneralOp(const Tensor &lhs, const Tensor &rhs,
+                        const Axes &lhsBatchingDimensions,
+                        const Axes &rhsBatchingDimensions,
+                        const Axes &lhsContractingDimensions,
+                        const Axes &rhsContractingDimensions) {
+  SmallVector<ShapedTypeComponents> inferredDotGeneralType;
+  if (failed(hlo::inferDotGeneralOp(
+          /*location=*/{}, lhs.getType(), rhs.getType(),
+          /*lhsBatchingDimensions=*/{}, /*rhsBatchingDimensions*/ {},
+          lhsContractingDimensions, rhsContractingDimensions,
+          /*precisionConfig=*/{}, inferredDotGeneralType)))
+    report_fatal_error(
+        invalidArgument("Could not infer DotGeneralOp's return type"));
+
+  return evalDotGeneralOp(
+      lhs, rhs, lhsBatchingDimensions, rhsBatchingDimensions,
+      lhsContractingDimensions, rhsContractingDimensions,
+      RankedTensorType::get(inferredDotGeneralType[0].getDims(),
+                            lhs.getElementType()));
+}
+
 Tensor evalPadOp(const Tensor &operand, const Tensor &paddingValue,
                  const Sizes &edgePaddingLow, const Sizes &edgePaddingHigh,
                  const Sizes &interiorPadding) {
@@ -143,6 +164,12 @@ Tensor evalSliceOp(const Tensor &operand, const Index &index) {
   return evalSliceOp(operand, start, limit, strides);
 }
 
+Sizes extractElements(ArrayRef<int64_t> arr, ArrayRef<int64_t> indices) {
+  Sizes elements;
+  for (int64_t index : indices) elements.push_back(arr[index]);
+  return elements;
+}
+
 void failOnDecomposableOp(Operation &op) {
   report_fatal_error(invalidArgument(
       "Operation %s is unsupported at the moment. "
@@ -151,6 +178,13 @@ void failOnDecomposableOp(Operation &op) {
       "Visit https://github.com/openxla/stablehlo/issues/1571 to learn more "
       "about the workaround and the roadmap for supporting this operation.",
       op.getName().getStringRef().str().c_str()));
+}
+
+template <typename T>
+DenseIntElementsAttr getDenseIntElementsAttr(Type elementType, T values,
+                                             SmallVector<int64_t> valuesShape) {
+  return DenseIntElementsAttr::get(
+      RankedTensorType::get(valuesShape, elementType), values);
 }
 
 SmallVector<SmallVector<uint32_t>> getReplicaGroups(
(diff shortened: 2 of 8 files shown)


stablehlo/tests/interpret_convolution.mlir
```diff
@@ -0,0 +1,79 @@
...
+    }
+  : (tensor<1x4x4x1xi64>, tensor<1x2x1x4xi64>) -> tensor<1x1x2x4xi64>
+  check.expect_eq_const %result, dense<[[[[1, 3, 10, 12],
+                                          [5, 7, 14, 16]]]]> : tensor<1x1x2x4xi64>
+  func.return
+}
+
+// -----
+
+func.func @convolution_feature_group_count_4() {
```
mlevesquedion, line 57: The `feature_group_count` seems to be equal to 2. Typo?

stablehlo/tests/interpret_convolution.mlir
```diff
@@ -0,0 +1,79 @@
+// RUN: stablehlo-translate --interpret -split-input-file %s
+
+func.func @convolution_op_test_si64() {
```
mlevesquedion, line 3: Is the difference between `si64` and `i64` meaningful here? (I think `i64` in an MLIR type means `si64`. Is this relevant to point out in the test name?)

stablehlo/tests/interpret_convolution.mlir
```diff
@@ -0,0 +1,79 @@
...
+  %rhs = stablehlo.constant dense<1> : tensor<3x3x1x1xi64>
+  %result = stablehlo.convolution(%lhs, %rhs)
+    dim_numbers = [b, 0, 1, f]x[0, 1, i, o]->[b, 0, 1, f],
+    window = {
+      stride = [4, 4],
+      lhs_dilate = [2, 2]
+    } {
+      batch_group_count = 1 : i64,
+      feature_group_count = 1 : i64,
+      precision_config = [#stablehlo<precision DEFAULT>, #stablehlo<precision DEFAULT>]
```
mlevesquedion, line 19: Can this be elided? Since you're using the defaults. If not, we should consider using [DefaultValuedAttr](http://google3/third_party/llvm/llvm-project/mlir/include/mlir/IR/CommonAttrConstraints.td;l=93;rcl=602520726) (in a future change)

stablehlo/reference/Index.h
```diff
@@ -121,6 +121,8 @@ using Index = Sizes;
 class IndexSpaceIterator {
  public:
   /// \name Constructor
+  IndexSpaceIterator(Sizes shape) : shape_(shape) { index_ = std::nullopt; }
```
mlevesquedion, line 124: Nit: Would it be possible to use a default value of `std::nullopt` for the `index` in the constructor below? I personally think that's cleaner but up to you.

stablehlo/reference/Ops.cpp
```diff
@@ -154,6 +188,65 @@ SmallVector<SmallVector<uint32_t>> getReplicaGroups(
...
+          /*padding=*/
+          getDenseIntElementsAttr(
+              IntegerType::get(lhs.getType().getContext(), 64), paddingVector,
+              SmallVector<int64_t>(padding.size(), 2)),
+          lhsDilation, rhsDilation, windowReversal, inputBatchDimension,
+          inputFeatureDimension, ArrayRef<int64_t>(inputSpatialDimensions),
+          kernelInputFeatureDimension, kernelOutputFeatureDimension,
+          ArrayRef<int64_t>(kernelSpatialDimensions), outputBatchDimension,
+          outputFeatureDimension, ArrayRef<int64_t>(outputSpatialDimensions),
+          /*featureGroupCount=*/1, batchGroupCount,
```
mlevesquedion, line 220: Why are we hard coding the feature group count here? Add a comment?

stablehlo/reference/Ops.cpp
```diff
@@ -154,6 +188,65 @@ SmallVector<SmallVector<uint32_t>> getReplicaGroups(
...
+          /*precisionConfig=*/{}, inferredConvolutionType)))
```
mlevesquedion, line 221: I'm assuming we don't need the precision config to infer the types?

stablehlo/reference/Ops.cpp
```diff
@@ -372,6 +465,52 @@ SmallVector<InterpreterValue> eval(Region &region,
...
+          convolutionOp.getDimensionNumbers().getInputFeatureDimension(),
+          Axes(convolutionOp.getDimensionNumbers().getInputSpatialDimensions()),
+          convolutionOp.getDimensionNumbers().getKernelInputFeatureDimension(),
+          convolutionOp.getDimensionNumbers().getKernelOutputFeatureDimension(),
+          Axes(
+              convolutionOp.getDimensionNumbers().getKernelSpatialDimensions()),
+          convolutionOp.getDimensionNumbers().getOutputBatchDimension(),
+          convolutionOp.getDimensionNumbers().getOutputFeatureDimension(),
+          Axes(
+              convolutionOp.getDimensionNumbers().getOutputSpatialDimensions()),
```
mlevesquedion, line 510: Nit: extract a "dimensionNumbers" variable?

stablehlo/reference/Ops.cpp
```diff
@@ -372,6 +465,52 @@ SmallVector<InterpreterValue> eval(Region &region,
...
       auto result = evalConvertOp(operand, convertOp.getType());
       scope.add(convertOp.getResult(), result);
+    } else if (auto convolutionOp = dyn_cast<ConvolutionOp>(op)) {
+      auto lhs = scope.findTensor(convolutionOp.getLhs());
+      auto rhs = scope.findTensor(convolutionOp.getRhs());
+      auto rank = lhs.getRank();
+
+      SmallVector<int64_t> windowStrides(rank - 2, 1);
+      if (auto windowStridesAttr = convolutionOp.getWindowStridesAttr())
+        windowStrides = SmallVector<int64_t>(windowStridesAttr.asArrayRef());
```
mlevesquedion, line 475: Nit: I think `auto windowStrides = convolutionOp.getWindowStrides().value_or(SmallVector<int64_t, 0>{})` should work.

(Actually it's unclear to me that `SmallVector` is needed here, you could do something similar with `ArrayRef`.)

mlevesquedion (commented):
Took a quick look, will let someone else do a full review.

stablehlo/reference/Ops.cpp
ghpvnist: Yes that's right.

stablehlo/reference/Ops.cpp
ghpvnist: Done.

stablehlo/reference/Ops.cpp
ghpvnist: We want default value to be filled with ones. `value_or` actually returns an `ArrayRef`, so there would be a type mismatch (and creating defaults with `ArrayRef` doesn't offer more readability compared to using `SmallVector` for the type).

stablehlo/reference/Index.h
ghpvnist: Done.

stablehlo/tests/interpret_convolution.mlir
ghpvnist: The interpreter currently doesn't handle different variations of `precision_config`, so this makes it explicit for now. There's also a couple of issues (#445, #755, #879) to figure out the exact story for `precision_config`.

stablehlo/tests/interpret_convolution.mlir
ghpvnist: There's a long standing issue #22 to try to converge to one. Long story short, the types were inherited from MHLO when StableHLO was first created. To avoid ambiguity, we explicitly write `si64` in the tests.

stablehlo/tests/interpret_convolution.mlir
ghpvnist: Done.

GleasonK (approved):

stablehlo/reference/Index.h
```diff
@@ -121,9 +121,11 @@ using Index = Sizes;
...
+
   IndexSpaceIterator(Sizes shape, std::optional<Index> index)
-      : shape_(shape), index_(index) {
-    if (index && !index->inBounds(shape)) index_ = std::nullopt;
+      : shape_(shape), index_(std::nullopt) {
+    if (index && index->inBounds(shape)) index_ = index;
```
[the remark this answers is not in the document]
ghpvnist: One use case is for comparing against end of the iterator. It's useful when iterating through a tensor one by one.

stablehlo/reference/Ops.cpp
```diff
@@ -54,6 +54,27 @@ Index evalIndex(Tensor tensor) {
...
 
+Tensor evalDotGeneralOp(const Tensor &lhs, const Tensor &rhs,
+                        const Axes &lhsBatchingDimensions,
+                        const Axes &rhsBatchingDimensions,
+                        const Axes &lhsContractingDimensions,
+                        const Axes &rhsContractingDimensions) {
+  SmallVector<ShapedTypeComponents> inferredDotGeneralType;
+  if (failed(hlo::inferDotGeneralOp(
+          /*location=*/{}, lhs.getType(), rhs.getType(),
+          /*lhsBatchingDimensions=*/{}, /*rhsBatchingDimensions*/ {},
```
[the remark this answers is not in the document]
ghpvnist: The only use of this function is by convolution and convolution doesn't use batching dimension. Since they aren't used, I've removed it.

PR #1965, opened 2024-01-29
Add custom builder to reduce op allowing type inference

sdasgup3:
The PR implements the a custom `reduce` op builder similar to what we have for mhlo [code](https://github.com/openxla/xla/blob/50aec2b3b54ce7a861f45bc3b0ae9b2cc2ee2a28/xla/mlir_hlo/mhlo/IR/hlo_ops.cc#L3917). 

## Background
https://github.com/openxla/stablehlo/pull/1869 allows the block arguments of reduce op to have different element types than that of the input arguments of reduce op and the output element type of the reduce op has to equal to those block arguments. As a consequence the output type of reduce op can no longer be inferred from the operand types. The auto-generated builders creates a reduce op with empty block and, as  a result, does not allow inferring the type.

The proposed solution is to create a custom builder which takes the element-type of the block arguments as arguments allowing result type inference.

### DIFF
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -1783,6 +1783,51 @@ LogicalResult ReduceOp::inferReturnTypeComponents(
                             inferredReturnShapes);
 }
 
+void ReduceOp::build(OpBuilder&, OperationState& odsState, ValueRange inputs,
+                     ValueRange initValues, DenseIntElementsAttr dimensions,
+                     TypeRange elementTypes) {
+  odsState.addOperands(inputs);
+  odsState.addOperands(initValues);
+  odsState.addAttribute(getDimensionsAttrName(odsState.name), dimensions);
+  (void)odsState.addRegion();
+
+  SmallVector<int64_t> newDimensions;
+  Attribute encoding;
+  ReduceOp::Adaptor adaptor(
+      odsState.operands,
+      odsState.attributes.getDictionary(odsState.getContext()), {},
+      odsState.regions);
+
+  SmallVector<ShapedType> inputArgTensorTypes{
+      llvm::map_range(adaptor.getInputs().getTypes(),
+                      [](Type t) { return t.cast<ShapedType>(); })};
+  SmallVector<ShapedType> initValueTensorTypes{
+      llvm::map_range(adaptor.getInitValues().getTypes(),
+                      [](Type t) { return t.cast<ShapedType>(); })};
+
+  if (succeeded(hlo::verifyReduceOpInputsAndInferShape(
+          odsState.location, inputArgTensorTypes,
+          llvm::to_vector(dimensions.getValues<int64_t>()), newDimensions,
+          encoding))) {
+    SmallVector<Type> inferredReturnTypes;
+    for (uint64_t inputIdx = 0; inputIdx < inputArgTensorTypes.size();
+         ++inputIdx) {
+      Type elementTy = elementTypes[inputIdx];
+      ShapedType inputType = inputArgTensorTypes[inputIdx];
+      if (inputType.hasRank()) {
+        inferredReturnTypes.push_back(
+            RankedTensorType::get(newDimensions, elementTy, encoding));
+      } else {
+        assert(encoding == nullptr && "attribute not supported");
+        inferredReturnTypes.push_back(UnrankedTensorType::get(elementTy));
+      }
+    }
+    odsState.addTypes(inferredReturnTypes);
+  } else {
+    llvm::report_fatal_error("Failed to infer result type(s).");
+  }
+}
+
 LogicalResult ReduceOp::verify() {
   return hlo::verifyReduceOp(getLoc(), getInputs(), getInitValues(),
                              getDimensions(), getBody());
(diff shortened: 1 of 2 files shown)


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -1783,6 +1783,51 @@ LogicalResult ReduceOp::inferReturnTypeComponents(
                             inferredReturnShapes);
 }
 
+void ReduceOp::build(OpBuilder&, OperationState& odsState, ValueRange inputs,
+                     ValueRange initValues, DenseIntElementsAttr dimensions,
+                     TypeRange elementTypes) {
+  odsState.addOperands(inputs);
+  odsState.addOperands(initValues);
+  odsState.addAttribute(getDimensionsAttrName(odsState.name), dimensions);
```
mlevesquedion, line 1791: I think this will fail because `ReduceOp` expects `DenseI64ArrayAttr` now: https://github.com/openxla/stablehlo/blob/6661fb318a166c61ddde7867af473989bf227dc8/stablehlo/dialect/StablehloOps.td#L1475

sdasgup3: Thanks for pointing. PTAL.

mlevesquedion: LGTM! For anyone reading this, when calling the builder you may need to do e.g. `rewriter.getDenseI64ArrayAttr` instead of `rewriter.getI64TensorAttr`.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -1783,6 +1783,50 @@ LogicalResult ReduceOp::inferReturnTypeComponents(
...
+                      [](Type t) { return t.cast<ShapedType>(); })};
+
+  if (succeeded(hlo::verifyReduceOpInputsAndInferShape(
+          odsState.location, inputArgTensorTypes, dimensions, newDimensions,
+          encoding))) {
+    SmallVector<Type> inferredReturnTypes;
+    for (uint64_t inputIdx = 0; inputIdx < inputArgTensorTypes.size();
+         ++inputIdx) {
+      Type elementTy = elementTypes[inputIdx];
+      ShapedType inputType = inputArgTensorTypes[inputIdx];
```
mlevesquedion, line 1815: Would `zip` work here?

mlevesquedion (approved):
LGTM. Consider adding some kind of test for this builder.

stablehlo/dialect/StablehloOps.cpp
sdasgup3: done!

sjain-stanford:
@sdasgup3 Gently checking-in on this, and appreciate your help here. Unfortunately cherry-picking from your fork [doesn't play well](https://github.com/llvm/torch-mlir/actions/runs/7702821771/job/20991906083) with the git submodule setup in torch-mlir which expects a commit on `main`, so I can't proceed with the fix without this landing. 
```
 Error: fatal: Fetched in submodule path 'externals/stablehlo', but it did not contain f2b6877713027370f26569929fb5d841a79d3589. Direct fetching of that commit failed.
 ```

File: stablehlo/dialect/StablehloOps.cpp [hunk 3]
```diff
@@ -1783,6 +1783,48 @@ LogicalResult ReduceOp::inferReturnTypeComponents(
...
+          odsState.location, inputArgTensorTypes, dimensions, newDimensions,
+          encoding))) {
+    SmallVector<Type> inferredReturnTypes;
+    for (auto [inputTy, elementTy] :
+         llvm::zip(inputArgTensorTypes, elementTypes)) {
+      if (inputTy.hasRank()) {
+        inferredReturnTypes.push_back(
+            RankedTensorType::get(newDimensions, elementTy, encoding));
+      } else {
+        assert(encoding == nullptr && "attribute not supported");
```
GleasonK, line 1818: `llvm::report_fatal_error` here as well?

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -1783,6 +1783,48 @@ LogicalResult ReduceOp::inferReturnTypeComponents(
...
+      odsState.regions);
+
+  SmallVector<ShapedType> inputArgTensorTypes{
+      llvm::map_range(adaptor.getInputs().getTypes(),
+                      [](Type t) { return t.cast<ShapedType>(); })};
+  SmallVector<ShapedType> initValueTensorTypes{
+      llvm::map_range(adaptor.getInitValues().getTypes(),
+                      [](Type t) { return t.cast<ShapedType>(); })};
+
+  if (succeeded(hlo::verifyReduceOpInputsAndInferShape(
```
GleasonK, line 1808: Refactor this to do the opposite: Check for failure and report fatal if it fails.

Same with your if statements below. Instead of `else assert` invert the if condition and fail as a verification check.

GleasonK (approved):

stablehlo/dialect/StablehloOps.cpp
sdasgup3: refactored. PTAL.

 >> Instead of else assert invert the if condition and fail as a verification check.
We have a statement following the assert which might not help much with the suggested change. In any case, PTAL and let me know if I am missing something.

GleasonK (approved):

PR #1966, opened 2024-01-29
Perform initialization in pass initialize method

mlevesquedion:
This is unlikely to give us performance gains since most of our passes run on modules anyway (so the initialization probably already occurs only once), but it is cleaner to separate the initialization of a pass from the actual running of the pass.

The code in `initialize` will run when the pass runs regardless of whether there is at least one instance of the target operation. However, modules are pretty much always present so this is unlikely to change anything.


GleasonK (approved):
There's a tradeoff of adding state vs initializing every time. I'm not sure whats best, but there's precedence for setting up passes like this elsewhere, so seems reasonable!

sdasgup3 (approved):

Issue #1895, opened 2023-12-25
Request implementation of `chlo-legalize-to-stablehlo`


qingyunqu:
### Request description

As @burmako describe in https://github.com/openxla/stablehlo/issues/602, we need migrate `chlo-legalize-to-hlo` to `chlo-legalize-to-stablehlo`.
Here I want to ask the timeline. There are two potential benefits that I'm concerned about:
1. Reuse stablehlo's compatibility guarantee and serialization capabilities to provide compatibility between frontends and backends.
2. Convert chlo to linalg. As `stablehlo-to-linalg` is completed, `chlo-legalize-to-linag` + `stablehlo-to-linalg` maybe suitable lowering path.

I would be very happy to get a response.  **If you need more manpower to speed it up, I can offer my time to help complete it.**

qingyunqu:
Is there any response? This request is import for https://github.com/llvm/torch-mlir/pull/2712.

GleasonK:
Closed by https://github.com/openxla/stablehlo/pull/1967

PR #1967, opened 2024-01-30
Add chlo legalize to stablehlo

linuxlonelyeagle:


linuxlonelyeagle:
Regarding the [reason](https://github.com/openxla/stablehlo/issues/1895) for submitting this PR.

GleasonK:
Hello!

Thanks again for the contributions 😄! I have lumped this PR and the canonicalizers one into a single RFC:
[[RFC] Adding StableHLO transformations for CHLO, Shape, Canonicalizations](https://groups.google.com/a/openxla.org/g/openxla-discuss/c/httnobiya7U)

The early feedback on this RFC is positive, so it seem plausible that we can proceed with these changes. If you have bandwidth, I think the proper / simpler solution would be to upstream the following CHLO files from IREE:
- CHLODecompositionPatterns.td
- LegalizeCHLO.cpp
- [Build targets for CHLO](https://github.com/openxla/iree/blob/406626bd90a8e8e76946215dbc04efe09eed24f5/compiler/plugins/input/StableHLO/stablehlo-iree/Conversion/BUILD.bazel#L49)
- tests in `Conversion/test/legalize_chlo_*.mlir`

If not, I'll do my best to prioritize upstreaming in the coming weeks!

linuxlonelyeagle:
Thanks!I'll get it done.

linuxlonelyeagle:
close by [1984](https://github.com/openxla/stablehlo/pull/1984).

PR #1970, opened 2024-01-30
Disallow unranked dynamism in ops that take output shape as an operand

mlevesquedion:
This includes:
  - dynamic_broadcast_in_dim
  - dynamic_iota
  - dynamic_reshape

#1991


stablehlo/dialect/TypeInference.cpp
```diff
@@ -3470,11 +3470,7 @@ LogicalResult verifyDynamicBroadcastInDimOp(
     std::optional<ArrayRef<int64_t>> knownNonexpandingDimensions,
     Value result) {
   auto operandType = operand.getType().dyn_cast<RankedTensorType>();
-  auto resultType = result.getType().dyn_cast<RankedTensorType>();
-
-  // If either the operand or result are unranked, there is very little
-  // to verify statically.
-  if (!operandType || !resultType) return success();
+  auto resultType = result.getType().cast<RankedTensorType>();
```
GleasonK, line 3473: This impl is likely shared between MHLO/StableHLO. For the duration of any parity gaps we'll need to make sure this logic makes its way into the MHLO op verify method.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3616,8 +3611,7 @@ LogicalResult verifyDynamicReshapeOp(std::optional<Location> location,
                                      Value outputShape, Value result) {
   auto resultType = result.getType().cast<ShapedType>();
   auto outputShapeType = outputShape.getType().cast<ShapedType>();
-  if (resultType.hasRank() && outputShapeType.hasStaticShape() &&
-      outputShapeType.getDimSize(0) != resultType.getRank())
+  if (outputShapeType.getDimSize(0) != resultType.getRank())
```
GleasonK, line 3614: Do we still need the `outputShapeType.hasStaticShape()` check?

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3555,13 +3551,12 @@ LogicalResult verifyDynamicIotaOp(std::optional<Location> location,
                                   Value outputShape, int64_t iotaDimension,
                                   Value result) {
   auto shape = result.getType().cast<ShapedType>();
```
GleasonK, line 3553: We should consider moving these casts to be `RankedTensorType`. This will make it very explicit that `hasRank` would be true. Let's do this in a followup if we decide to, since it will apply to most ops eventually.

stablehlo/dialect/TypeInference.cpp
GleasonK: Edit: Saw the tablegen descriptor. That's a better way of handling this.

GleasonK (approved):
LGTM. Let me pull these changes and make sure that the new bytecode files are compatible with previous releases before submit.

stablehlo/dialect/TypeInference.cpp
mlevesquedion: Ack. The impl is indeed shared. I have a patch ready.

mlevesquedion:
Double checked bytecode compatibility; no issues detected.

Issue #1962, opened 2024-01-28
TorchToStablehlo lit test failure with scatter.mlir


sjain-stanford:
### What happened?

Hi ⁠stablehlo, we need some help on the `torch-mlir` side of things with a lit test error we hit when bumping stablehlo that hits an `unexpected error: Expects non-empty reduction block for type inference`

PR: https://github.com/llvm/torch-mlir/pull/2821
Failing CI: https://github.com/llvm/torch-mlir/actions/runs/7687361450/job/20947291128?pr=2821

Here's the error:
```
bazel run @torch-mlir//test/Conversion:TorchToStablehlo/scatter.mlir.test 

...external/torch-mlir/test/Conversion/TorchToStablehlo/scatter.mlir
within split at <stdin>:1 offset :33:8: error: unexpected error: Expects non-empty reduction block for type inference                                                                               
  %0 = torch.aten.scatter.src %arg0, %int0, %arg1, %arg2 : !torch.vtensor<[?,?],si64>, !torch.int, !torch.vtensor<[?,?],si64>, !torch.vtensor<[?,?],si64> -> !torch.vtensor<[?,?],si64>             
       ^                                                                                                             
LLVM ERROR: Failed to infer result type(s).
```

Seems like this check was added in https://github.com/openxla/stablehlo/pull/1918 but I lack the historic context to know what needs fixing. 

### Steps to reproduce your issue

1. Clone https://github.com/sjain-stanford/torch-mlir and checkout `sambhav/bazel_fix` branch
2. Fetch git submodules: `git submodule update --init --progress`
3. Launch docker: `./utils/bazel/docker/run_docker.sh`
4. Test: `bazel run @torch-mlir//test/Conversion:TorchToStablehlo/scatter.mlir.test`

sdasgup3:
Hi @sjain-stanford 
Thanks for posting the issue!

https://github.com/openxla/stablehlo/pull/1869 allows the block arguments of reduce op to have different element types than that of the input arguments of reduce op and the output type of the reduce op has to equal to those block arguments.  As a consequence the output type of reduce op can no longer be inferred from the operand types: The type inference needs to somehow have the information regarding the block-arguments's element type. The check that was added at #1918 is a guard against inferring reduce op type with empty blocks, which is what is happening at https://sourcegraph.com/github.com/llvm/torch-mlir/-/blob/lib/Conversion/TorchToStablehlo/GatherScatter.cpp?L336.

[...]

PR #1972, opened 2024-01-31
Make setup_sanitizer include optional in CMakeLists.txt

sjain-stanford:
Fixes https://github.com/openxla/stablehlo/issues/1962#issuecomment-1919576685

Context: 
This breaks torch-mlir's CMake build when bumping stablehlo. 

The CMake build [fails](https://github.com/llvm/torch-mlir/actions/runs/7729517677/job/21072878412?pr=2821) with setup_sanitizers. Probably related to https://github.com/openxla/stablehlo/pull/1959 (cc: @fzakaria , @mlevesquedion ). I see the default for `STABLEHLO_ENABLE_SANITIZER` being `OFF`, but this code probably needs to be enclosed in an `if-endif`:
```
include(SetupSanitizers)
setup_sanitizers()
```
Error:
```
  -- Building StableHLO as an external LLVM project
  -- Building with -fPIC
  CMake Error at /_work/torch-mlir/torch-mlir/externals/stablehlo/CMakeLists.txt:147 (include):
    include could not find requested file:
  
      SetupSanitizers
  
  
  CMake Error at /_work/torch-mlir/torch-mlir/externals/stablehlo/CMakeLists.txt:148 (setup_sanitizers):
    Unknown CMake command "setup_sanitizers".
  
  
  -- Configuring incomplete, errors occurred!
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -144,8 +144,10 @@ add_definitions(${LLVM_DEFINITIONS})
 # Sanitizer configuration
 #-------------------------------------------------------------------------------
 
-include(SetupSanitizers)
-setup_sanitizers()
+if(STABLEHLO_ENABLE_SANITIZER)
+  include(SetupSanitizers)
+  setup_sanitizers()
+endif()
 
 #-------------------------------------------------------------------------------
 # Python configuration
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,7 +122,7 @@ else()
 endif()
 
 # Add the CMake modules specific to StableHLO
-list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake")
 
 if(LLVM_ENABLE_ZLIB)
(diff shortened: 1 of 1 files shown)


fzakaria:
I think the change is to fix:
```
list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
```

to not be the PROJECT_SOURCE_DIR but a directory relative to the current file which would work in embedded use case.

fzakaria:
I will test a change using `CMAKE_CURRENT_LIST_DIR` so that it propagates.

sjain-stanford:
@fzakaria using `CMAKE_CURRENT_LIST_DIR` works too. Just pushed the change. Do you want me to still keep the `if-endif` (I've removed for now)?

fzakaria:
LGTM thanks!

fzakaria (approved):
LGTM -- thank you for iterating a little on it.
You also confirmed it worked for the embedded case?

I'm new to the team and wasn't aware who is using that embedded case as well.

sjain-stanford:
> You also confirmed it worked for the embedded case?

I've tested it from torch-mlir where stablehlo is an external git submodule. If that what you're referring here as the `embedded case` then yes!

PR #1973, opened 2024-01-31
Add option to separate debug information (DWARF) from executable

fzakaria:
When building Debug or RelWithDebInfo, the size of the binaries (i.e. `stablehlo-opt`) can be quite beefy.
I noticed it can be up to **2GiB** in size and attaching GDB is quite slow -- in fact loading the process to run is slow.
_This is much larger and noticeable than the Bazel build._

This change introduces a common idiom to speed up large binaries by separating the debug information (DWARF) into separate files (`*.dwo). In order to speedup GDB attachment, a gdb-index is created and stored in the file itself.

One could achieve the same effect with `.gdbinit` settings however the management of the index files are never pruned and it's a bit cumbersome to rely on developers to set it up.

For reference here is a possible `.gdbinit`
```
# History.
set history filename ~/.gdb_history
set history save on
set history size 100000

# Makes multiple invocations much faster
set index-cache on

# Allow per-project gdbinit files
set auto-load local-gdbinit on
add-auto-load-safe-path /
```

**Note**:  Looks like you also need to build MLIR with separate dwarfs as well. To be honest, the interplay between settings on LLVM and then on StableHLO are not very clear at times. Some settings seem to propagate based on how the previous object code was created and others did not.

Specifically for GDB launch, you can see a very big time saving in the below benchmark. **76s vs 1.3s**

## Inline debug info
**Size**: 3.2GiB
```shell
❯ ll -h bin/stablehlo-opt
Permissions Size User   Date Modified Name
.rwxr-xr-x  3.2G 780412 31 Jan 18:14  bin/stablehlo-opt
```

Most of the space is the debug information.
```shell
❯ /google/bin/releases/protobuf-team/bloaty/bloaty bin/stablehlo-opt --allow_unsafe_non_google3_input
    FILE SIZE        VM SIZE    
 --------------  -------------- 
  60.2%  1.80Gi   0.0%       0    .debug_info
  12.4%   378Mi   0.0%       0    .debug_str
  10.3%   314Mi   0.0%       0    .debug_loclists
   5.9%   180Mi   0.0%       0    .debug_line
```

**Benchmark**
```
# benchmark just running the build
❯ hyperfine './build/bin/stablehlo-opt --version'
Benchmark 1: ./build/bin/stablehlo-opt --version
  Time (mean ± σ):     181.6 ms ±   3.8 ms    [User: 73.4 ms, System: 108.2 ms]
  Range (min … max):   173.8 ms … 188.3 ms    16 runs


# benchark with GDB
❯ hyperfine 'gdb -ex run --args ./build/bin/stablehlo-opt --version' --warmup 1 --runs 3
Benchmark 1: gdb -ex run --args ./build/bin/stablehlo-opt --version
  Time (mean ± σ):     74.063 s ±  2.381 s    [User: 71.044 s, System: 3.958 s]
  Range (min … max):   72.361 s … 76.784 s    3 runs
```

## With separate debug info

**Size**: 847M
```shell
❯ ll -h build/bin/stablehlo-opt
Permissions Size User   Date Modified Name
.rwxr-xr-x  847M 780412 31 Jan 17:29  build/bin/stablehlo-opt
````

Much of the space is now a `gdb-index` to make attaching to the debugger much faster
```shell
❯ /google/bin/releases/protobuf-team/bloaty/bloaty build/bin/stablehlo-opt --allow_unsafe_non_google3_input
    FILE SIZE        VM SIZE    
 --------------  -------------- 
  38.0%   306Mi   0.0%       0    .gdb_index
  24.2%   195Mi   0.0%       0    .debug_addr
  20.6%   166Mi   0.0%       0    .debug_line
   6.5%  52.5Mi  55.5%  52.5Mi    .text
```
**Benchark**:
```
# benchmark just running the build
❯ hyperfine './build/bin/stablehlo-opt --version'
Benchmark 1: ./build/bin/stablehlo-opt --version
  Time (mean ± σ):      38.6 ms ±   1.6 ms    [User: 15.7 ms, System: 23.1 ms]
  Range (min … max):    36.5 ms …  44.6 ms    68 runs

# benchark with GDB
❯ hyperfine 'gdb -ex run --args ./build/bin/stablehlo-opt --version'
Benchmark 1: gdb -ex run --args ./build/bin/stablehlo-opt --version
  Time (mean ± σ):      1.401 s ±  0.037 s    [User: 1.532 s, System: 0.883 s]
  Range (min … max):    1.345 s …  1.445 s    10 runs
```


CMakeLists.txt
```diff
@@ -62,29 +62,6 @@ else()
...
-        set(CCACHE_PROGRAM "CCACHE_MAXSIZE=${LLVM_CCACHE_MAXSIZE} ${CCACHE_PROGRAM}")
-      endif()
-      if (LLVM_CCACHE_DIR)
-        set(CCACHE_PROGRAM "CCACHE_DIR=${LLVM_CCACHE_DIR} ${CCACHE_PROGRAM}")
-      endif()
-      set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ${CCACHE_PROGRAM})
-  else()
-    message(FATAL_ERROR "Unable to find the program ccache. Set LLVM_CCACHE_BUILD to OFF")
-  endif()
-endif()
```
fzakaria, line 86: This looks to be a deprecated way to setup CCACHE
https://discourse.llvm.org/t/llvm-ccache-build-is-deprecated/68431/5

In fact our build scripts use the alternative approach:
```
     -DCMAKE_C_COMPILER_LAUNCHER=ccache \
```

I find that the fact it's also prefix'd with LLVM confusing since all our other options use STABLEHLO

cmake/SetupSanitizers.cmake
```diff
@@ -47,7 +47,7 @@ function(setup_sanitizers)
 
     if (STABLEHLO_ENABLE_SANITIZER_LOWERCASE STREQUAL "address")
         add_compile_options(-fsanitize=address -fsanitize=undefined -fsanitize=leak -fno-omit-frame-pointer)
-        link_libraries(-fsanitize=address -fsanitize=undefined -fsanitize=leak)
+        add_link_options(-fsanitize=address -fsanitize=undefined -fsanitize=leak)
```
fzakaria, line 50: Small fix. It worked with `link_libraries` but it does seem like the wrong call to have made.

CMakeLists.txt
```diff
@@ -128,6 +107,22 @@ if(LLVM_ENABLE_ZLIB)
   find_package(ZLIB)
 endif()
 
+#-------------------------------------------------------------------------------
+# Performance configuration
+#-------------------------------------------------------------------------------
+
+include(CheckCXXCompilerFlag)
+include(CheckLinkerFlag)
+add_link_options("$<$<BOOL:${STABLEHLO_ENABLE_LLD}>:-fuse-ld=lld>")
```
fzakaria, line 116: I found that linking stablehlo-opt to be also much fsater with lld -- it's also recommended in LLVM.

CMakeLists.txt
```diff
@@ -128,6 +107,22 @@ if(LLVM_ENABLE_ZLIB)
...
+if(STABLEHLO_ENABLE_SPLIT_DWARF)
+    check_cxx_compiler_flag(-gsplit-dwarf STABLEHLO_SUPPORTS_SPLIT_DWARF)
+    add_compile_options("$<$<BOOL:${STABLEHLO_SUPPORTS_SPLIT_DWARF}>:-gsplit-dwarf;-ggnu-pubnames>")
+    check_linker_flag(CXX "-Wl,--gdb-index" STABLEHLO_SUPPORTS_GDB_INDEX)
+    # If we set LLD it doesn't seem to affect the check_linker_flag above.
+    # Account for it with the generator expression OR
+    add_link_options("$<$<OR:$<BOOL:${STABLEHLO_SUPPORTS_GDB_INDEX}>,$<BOOL:${STABLEHLO_ENABLE_LLD}>>:-Wl,--gdb-index>")
```
fzakaria, line 123: This was kind of confusing -- might open a bug to CMake.
The `check_linker_flag` above doesn't seem to use the previous `add_link_options` to use lld.
The default on my system was using `/bin/ld` (GNU) which didn't support gdb-index.

fzakaria:
I noticed these are settings common in many LLVM based projects such as IREE
In fact torch-mlir just hardcodes them as non-optional: https://github.com/llvm/torch-mlir/blob/54e258792c686e4bb30a8bbd1a0f1268d440d2a4/build_tools/ci/linux_default_toolchain.cmake

mlevesquedion:
Is this ready for review? I'm happy to review if so.

Nit: This change seems to be doing some cleanup in addition to the main changes (the biggest cleanup seems to be related to `ccache` changes). Consider breaking up the PR to make the change easier to review and reason about.

CMakeLists.txt
fzakaria: Moved to https://github.com/openxla/stablehlo/pull/1974/files

fzakaria:
@mlevesquedion : added you for review.
I moved the ccache to https://github.com/openxla/stablehlo/pull/1974 per your recommendation.

CMakeLists.txt
```diff
@@ -46,6 +46,8 @@ option(STABLEHLO_BUILD_EMBEDDED "Build StableHLO as part of another project" OFF
 option(STABLEHLO_ENABLE_BINDINGS_PYTHON "Enables StableHLO Python bindings" OFF)
 option(STABLEHLO_ENABLE_STRICT_BUILD "Build StableHLO with strict warnings and warnings as errors" OFF)
 option(STABLEHLO_ENABLE_SANITIZER "Enable a sanitizer [OFF, address]" OFF)
+option(STABLEHLO_ENABLE_SPLIT_DWARF "Enable split DWARF if the platform supports it" OFF)
```
mlevesquedion, line 49: Should this default to ON? Are there any downsides to using this?

CMakeLists.txt
```diff
@@ -128,6 +130,22 @@ if(LLVM_ENABLE_ZLIB)
...
 
+#-------------------------------------------------------------------------------
+# Performance configuration
+#-------------------------------------------------------------------------------
+
+include(CheckCXXCompilerFlag)
+include(CheckLinkerFlag)
+add_link_options("$<$<BOOL:${STABLEHLO_ENABLE_LLD}>:-fuse-ld=lld>")
+if(STABLEHLO_ENABLE_SPLIT_DWARF)
+    check_cxx_compiler_flag(-gsplit-dwarf STABLEHLO_SUPPORTS_SPLIT_DWARF)
```
mlevesquedion, line 141: Consider producing some kind of warning if split dwarf is not supported? Could help with debugging.

CMakeLists.txt
mlevesquedion: Looks like this is a CMake generator expression: https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html
Hadn't seen those before. Need to brush up on CMake :upside_down_face:

CMakeLists.txt
mlevesquedion: Hmm that is indeed weird. Looks like there's not really a clean way to do this: https://gitlab.kitware.com/cmake/cmake/-/issues/24254

This `OR` LGTM though.

README.md
```diff
@@ -90,6 +90,30 @@ Here's how to build the StableHLO repo on Linux or macOS:
      -DMLIR_DIR=${PWD}/../llvm-build/lib/cmake/mlir
    ```
 
+   If you are actively developing StableHLO, you may want the following additional
+   CMake settings:
+
+   ```shell
+      cmake .. -GNinja \
+     -DSTABLEHLO_ENABLE_LLD=ON \
```
mlevesquedion, line 98: Higher in this file we recommend using LLD for Linux, but not MacOS. Do we want to do the same here?

mlevesquedion (approved):
LGTM. I haven't used `gdb` yet but these are some impressive size and latency savings!

CMakeLists.txt
fzakaria: Maybe -- It might be good to get a place (maybe with Nix) where everything is controlled.
Right now I'm finding that on my machine I build with GCC but the CI uses Clang.
Removing all the variability and forcing a consistent toolchain would let us remove a lot of these options and make them default ON

CMakeLists.txt
fzakaria: https://gitlab.kitware.com/cmake/cmake/-/issues/25649 filed.

README.md
fzakaria: Will fix in subsequent change.

PR #1974, opened 2024-01-31
Add ccache by default when building MLIR

fzakaria:
* Add ccache use for build_mlir.sh script
* Remove LLVM_CCACHE_BUILD which is not used and [deprecated ](https://discourse.llvm.org/t/llvm-ccache-build-is-deprecated/68431 ) (Although our use of it is not necessarily the LLVM's version)

Note: this change was pulled out of https://github.com/openxla/stablehlo/pull/1973

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,29 +62,6 @@ else()
   set(CMAKE_CXX_STANDARD 17)
 endif()
 
-# Build with ccache if the package is present
-set(LLVM_CCACHE_BUILD OFF CACHE BOOL "Set to ON for a ccache enabled build")
-if(LLVM_CCACHE_BUILD)
-  find_program(CCACHE_PROGRAM ccache)
-  if(CCACHE_PROGRAM)
-      set(LLVM_CCACHE_MAXSIZE "" CACHE STRING "Size of ccache")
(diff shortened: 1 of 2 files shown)


mlevesquedion (approved):
Did you notice any difference in the build/rebuild speed with this change?

Also consider adding a link to https://discourse.llvm.org/t/llvm-ccache-build-is-deprecated/68431 in the PR description. It seems like the mechanism was not actually deprecated, but the `CMAKE_C/CXX_COMPILER_LAUNCHER` variables should work, and it's less config so I say let's do it.

fzakaria:
I didn't run enough benchmarks but I see ccache itself hydrating with results.
```
❯ ccache -s
Cacheable calls:    37515 / 38077 (98.52%)
  Hits:             12998 / 37515 (34.65%)
    Direct:         11561 / 12998 (88.94%)
    Preprocessed:    1437 / 12998 (11.06%)
  Misses:           24517 / 37515 (65.35%)
Uncacheable calls:    560 / 38077 ( 1.47%)
Errors:                 2 / 38077 ( 0.01%)
Local storage:
  Cache size (GiB):   6.3 /  50.0 (12.69%)
  Cleanups:          1087
  Hits:             13032 / 37749 (34.52%)
  Misses:           24717 / 37749 (65.48%)
```

PR #1976, opened 2024-02-01
Add explicit error reporting to transform passes

GleasonK:
Currently pass failures are silent (only written to `llvm::dbgs`) in these passes Adding explicit error reporting will at least make it clear what pass has failed.

### DIFF
--- a/stablehlo/transforms/StablehloCanonicalizeDynamism.cpp
+++ b/stablehlo/transforms/StablehloCanonicalizeDynamism.cpp
@@ -308,8 +308,11 @@ struct StablehloCanonicalizeDynamismPass
 
     RewritePatternSet patterns(&getContext());
     populateStablehloCanonicalizeDynamismPatterns(&patterns, &getContext());
-    if (failed(applyPatternsAndFoldGreedily(getOperation(), std::move(patterns),
-                                            config))) {
+    auto func = getOperation();
+    if (failed(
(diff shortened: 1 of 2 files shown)


build_tools/github_actions/lint_clang_format.sh
```diff
@@ -45,7 +45,7 @@ fi
 echo "Running clang-format [mode=$FORMAT_MODE]..."
 echo "  Files: $CHANGED_FILES"
 if [[ $FORMAT_MODE == 'fix' ]]; then
-  clang-format --style=google -i "$CHANGED_FILES"
+  clang-format --style=google -i $CHANGED_FILES
```
mlevesquedion, line 48: Hmm. That's annoying, sorry about that. I'll need to go re-review all the changes from https://github.com/openxla/stablehlo/pull/1971 to make sure there aren't more accidental breakages like this one.

IMO the correct solution here is to use `mapfile` when declaring `CHANGED_FILES` and to use array expansion here: `"${CHANGED_FILES[@]}"`. In practice this is unlikely to matter though, since I doubt we would ever have file names with whitespace or globs in them.

Also this `git diff` thing is making me realize that we have a few checks that run `find` needlessly.

Also the shellcheck lint should have failed with this change, but it didn't.

I'll do the array thing in a follow-up change.

mlevesquedion: Fixed in https://github.com/openxla/stablehlo/pull/1980

PR #1979, opened 2024-02-01
Remove generator expression

fzakaria:
Based on feedback from
https://gitlab.kitware.com/cmake/cmake/-/issues/25649 removed the generator expression in favor of if statement.

Example output:
```
-- Building StableHLO with an installed MLIR
-- Using MLIRConfig.cmake in: /usr/local/google/home/fmzakari/code/github.com/openxla/stablehlo/build/../llvm-build/lib/cmake/mlir
-- Using LLVMConfig.cmake in: /usr/local/google/home/fmzakari/code/github.com/openxla/stablehlo/llvm-build/lib/cmake/llvm
-- Enabling LLD as the linker
-- Enabling split-dwarf build
-- Enabling GDB index in binary
-- Building with -fPIC
-- Configuring done (0.2s)
-- Generating done (0.1s)
-- Build files have been written to: /usr/local/google/home/fmzakari/code/github.com/openxla/stablehlo/build
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -113,14 +113,24 @@ endif()
 
 include(CheckCXXCompilerFlag)
 include(CheckLinkerFlag)
-add_link_options("$<$<BOOL:${STABLEHLO_ENABLE_LLD}>:-fuse-ld=lld>")
+if (STABLEHLO_ENABLE_LLD)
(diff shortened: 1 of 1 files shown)


mlevesquedion (approved):
Cool, this does look nicer!

PR #1983, opened 2024-02-01
Add interpreter for CollectiveBroadcastOp

### LINKED ISSUE #1982: Add interpreter for CollectiveBroadcastOp

ghpvnist:
closes #1982

### DIFF
--- a/stablehlo/reference/Ops.cpp
+++ b/stablehlo/reference/Ops.cpp
@@ -1105,10 +1105,10 @@ Tensor evalCollectiveBroadcastOp(
   if (channelId <= 0) processGroups = process->crossReplica(replicaGroups);
   if (channelId > 0) processGroups = process->crossPartition(replicaGroups);
 
-  for (auto processGroup : processGroups)
-    if (llvm::find(processGroup, process->getId()) != processGroup.end())
-      return process->rendezvous(processGroup, channelId, operand)
-          .lookup(processGroup[0]);
+  auto processGroup = processGroups.findGroup(process->getId());
+  if (processGroup)
+    return process->rendezvous(*processGroup, channelId, operand)
+        .lookup((*processGroup)[0]);
 
   return evalBroadcastInDimOp(
       makeScalar(convert(operand.getElementType(), 0.0)), {},
--- a/stablehlo/reference/Ops.cpp
+++ b/stablehlo/reference/Ops.cpp
@@ -328,6 +328,25 @@ SmallVector<InterpreterValue> eval(Region &region,
       auto operand = scope.findTensor(clzOp.getOperand());
       auto result = evalClzOp(operand, clzOp.getType());
       scope.add(clzOp.getResult(), result);
+    } else if (auto collectiveBroadcastOp =
+                   dyn_cast<CollectiveBroadcastOp>(op)) {
+      auto operand = scope.findTensor(collectiveBroadcastOp.getOperand());
+
+      auto replicaGroupsAttr = collectiveBroadcastOp.getReplicaGroups();
+      auto replicaGroupsShape = replicaGroupsAttr.getShapedType().getShape();
+      SmallVector<SmallVector<uint32_t>> replicaGroups(replicaGroupsShape[0]);
+      auto replicaGroupsIt = replicaGroupsAttr.getValues<int64_t>().begin();
+      for (auto &replicaGroup : replicaGroups)
+        for (auto i = 0; i < replicaGroupsShape[1]; ++i, ++replicaGroupsIt)
+          replicaGroup.push_back(*replicaGroupsIt);
+
+      ChannelId channelId = 0;
+      if (auto channelHandle = collectiveBroadcastOp.getChannelHandle())
+        channelId = channelHandle->getHandle();
+
+      auto result =
+          evalCollectiveBroadcastOp(operand, replicaGroups, channelId, process);
+      scope.add(collectiveBroadcastOp.getResult(), result);
     } else if (auto collectivePermuteOp = dyn_cast<CollectivePermuteOp>(op)) {
       auto operand = scope.findTensor(collectivePermuteOp.getOperand());
 
@@ -1072,6 +1091,28 @@ Tensor evalClzOp(const Tensor &operand, ShapedType resultType) {
     result.set(*it, element);
   }
   return result;
+}
+
+Tensor evalCollectiveBroadcastOp(
+    const Tensor &operand, SmallVector<SmallVector<uint32_t>> replicaGroups,
+    ChannelId channelId, Process *process) {
+  if (!process)
+    llvm::report_fatal_error(
+        "collective_broadcast is only supported when run via "
+        "interpreter.run_parallel");
+
+  ProcessGroups processGroups;
+  if (channelId <= 0) processGroups = process->crossReplica(replicaGroups);
+  if (channelId > 0) processGroups = process->crossPartition(replicaGroups);
+
+  for (auto processGroup : processGroups)
+    if (llvm::find(processGroup, process->getId()) != processGroup.end())
+      return process->rendezvous(processGroup, channelId, operand)
+          .lookup(processGroup[0]);
+
+  return evalBroadcastInDimOp(
+      makeScalar(convert(operand.getElementType(), 0.0)), {},
+      operand.getType());
 }
 
 Tensor evalCollectivePermuteOp(
--- a/stablehlo/tests/interpret/collective_broadcast.mlir
+++ b/stablehlo/tests/interpret/collective_broadcast.mlir
@@ -1,14 +1,14 @@
 // RUN: stablehlo-translate --interpret -split-input-file %s
 
 module @cross_replica {
-  func.func public @collective_broadcast(%operand : tensor<1x2xi64>) -> tensor<1x2xi64> {
+  func.func @collective_broadcast(%operand : tensor<1x2xi64>) -> tensor<1x2xi64> {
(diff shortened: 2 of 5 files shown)


stablehlo/reference/Ops.cpp
```diff
@@ -328,6 +328,25 @@ SmallVector<InterpreterValue> eval(Region &region,
...
+      auto replicaGroupsShape = replicaGroupsAttr.getShapedType().getShape();
+      SmallVector<SmallVector<uint32_t>> replicaGroups(replicaGroupsShape[0]);
+      auto replicaGroupsIt = replicaGroupsAttr.getValues<int64_t>().begin();
+      for (auto &replicaGroup : replicaGroups)
+        for (auto i = 0; i < replicaGroupsShape[1]; ++i, ++replicaGroupsIt)
+          replicaGroup.push_back(*replicaGroupsIt);
+
+      ChannelId channelId = 0;
+      if (auto channelHandle = collectiveBroadcastOp.getChannelHandle())
+        channelId = channelHandle->getHandle();
```
mlevesquedion, line 345: Nit: the difference in names here and in the spec is a little bit confusing.  Should `getHandle()` be `getId()` ? (This is outside the scope of this CL, I'm just wondering.)

stablehlo/reference/Ops.cpp
```diff
@@ -328,6 +328,25 @@ SmallVector<InterpreterValue> eval(Region &region,
...
+                   dyn_cast<CollectiveBroadcastOp>(op)) {
+      auto operand = scope.findTensor(collectiveBroadcastOp.getOperand());
+
+      auto replicaGroupsAttr = collectiveBroadcastOp.getReplicaGroups();
+      auto replicaGroupsShape = replicaGroupsAttr.getShapedType().getShape();
+      SmallVector<SmallVector<uint32_t>> replicaGroups(replicaGroupsShape[0]);
+      auto replicaGroupsIt = replicaGroupsAttr.getValues<int64_t>().begin();
+      for (auto &replicaGroup : replicaGroups)
+        for (auto i = 0; i < replicaGroupsShape[1]; ++i, ++replicaGroupsIt)
+          replicaGroup.push_back(*replicaGroupsIt);
```
mlevesquedion, line 341: Would a call to `getReplicaGroups` work here? I'm noticing that `AllToAll` and `AllGather` also have this code.

stablehlo/tests/interpret/collective_broadcast.mlir
```diff
@@ -0,0 +1,223 @@
...
+    %operand3 = stablehlo.constant dense<[[7, 8]]> : tensor<1x2xi64>
+    %results:4 = "interpreter.run_parallel"(%operand0, %operand1, %operand2, %operand3) {
+      programs=[[@collective_broadcast, @collective_broadcast,
+                 @collective_broadcast, @collective_broadcast]]
+    } : (tensor<1x2xi64>, tensor<1x2xi64>, tensor<1x2xi64>, tensor<1x2xi64>) ->
+        (tensor<1x2xi64>, tensor<1x2xi64>, tensor<1x2xi64>, tensor<1x2xi64>)
+    check.expect_eq_const %results#0, dense<[[1, 2]]> : tensor<1x2xi64>
+    check.expect_eq_const %results#1, dense<[[3, 4]]> : tensor<1x2xi64>
+    check.expect_eq_const %results#2, dense<[[5, 6]]> : tensor<1x2xi64>
+    check.expect_eq_const %results#3, dense<[[7, 8]]> : tensor<1x2xi64>
```
mlevesquedion, line 80: I'm a bit confused as to why this is the expected result. Based on the previous two examples, I would have expected the output to be [[1, 2]], [[0, 0]], [[0, 0]], [[0, 0]]. What am I missing?

mlevesquedion (commented):
Change looks good but I'm confused about how replicaGroups / processGroups work. I will need to do some more digging into those before I can properly review.

stablehlo/tests/interpret/collective_broadcast.mlir
mlevesquedion: Please ignore this. I didn't notice that the `programs` argument differed between tests and I didn't understand its significance.

stablehlo/tests/interpret/collective_broadcast.mlir
```diff
@@ -0,0 +1,223 @@
+// RUN: stablehlo-translate --interpret -split-input-file %s
+
+module @cross_replica {
+  func.func public @collective_broadcast(%operand : tensor<1x2xi64>) -> tensor<1x2xi64> {
```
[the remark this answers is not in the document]
mlevesquedion: According to this documentation symbols are public by default: https://mlir.llvm.org/docs/SymbolsAndSymbolTables/#symbol-visibility

Consider removing the specifier as it doesn't seem relevant for these tests.

stablehlo/reference/Ops.cpp
```diff
@@ -1074,6 +1093,28 @@ Tensor evalClzOp(const Tensor &operand, ShapedType resultType) {
...
+    const Tensor &operand, SmallVector<SmallVector<uint32_t>> replicaGroups,
+    ChannelId channelId, Process *process) {
+  if (!process)
+    llvm::report_fatal_error(
+        "collective_broadcast is only supported when run via "
+        "interpreter.run_parallel");
+
+  ProcessGroups processGroups;
+  if (channelId <= 0) processGroups = process->crossReplica(replicaGroups);
+  if (channelId > 0) processGroups = process->crossPartition(replicaGroups);
```
mlevesquedion, line 1106: Not blocking, outside the scope of this PR: I find this way of creating the process groups unintuitive. Why not use an enum to decide which communication scheme to use?

Also, when using `crossPartition` the `replicaGroups` are actually `partitionGroups`. Would `communicationGroups` be a better name?

stablehlo/reference/Ops.cpp
```diff
@@ -1074,6 +1093,28 @@ Tensor evalClzOp(const Tensor &operand, ShapedType resultType) {
...
+
+  for (auto processGroup : processGroups)
+    if (llvm::find(processGroup, process->getId()) != processGroup.end())
+      return process->rendezvous(processGroup, channelId, operand)
+          .lookup(processGroup[0]);
```
mlevesquedion, line 1111: Would `processGroups.findGroup` work here? Perhaps you can update the implementation of that to use `llvm::find`, perhaps in a separate change.

stablehlo/tests/interpret/collective_broadcast.mlir
mlevesquedion, line 1: Nit: since the behavior of the operation depends so heavily on the way the process groups are formed, consider adding comments to each test case showing what the process groups look like.

mlevesquedion (approved):

stablehlo/reference/Ops.cpp
ghpvnist: This is one of those things that are part of the Frontend Contract workstream. We have a issue in #351 to track this, and I've added your comment.

stablehlo/reference/Ops.cpp
ghpvnist: To answer the first question, it's written like that for spec parity since the spec uses them as function calls: 
https://openxla.org/stablehlo/spec#collective_broadcast
https://openxla.org/stablehlo/spec#collective_ops

stablehlo/reference/Ops.cpp
ghpvnist: `getReplicaGroups` returns a `DenseIntElementsAttr`, but we still need to convert it to a list of lists to pass it to the eval function.

stablehlo/tests/interpret/collective_broadcast.mlir
ghpvnist: Hm, this is certainly good to have. I'll think about this and address them separately since this would require nontrivial amount of work.

stablehlo/tests/interpret/collective_broadcast.mlir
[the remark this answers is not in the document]
ghpvnist: Done.

stablehlo/reference/Ops.cpp
ghpvnist: Done.

stablehlo/reference/Ops.cpp
```diff
@@ -1105,10 +1105,10 @@ Tensor evalCollectiveBroadcastOp(
   if (channelId <= 0) processGroups = process->crossReplica(replicaGroups);
   if (channelId > 0) processGroups = process->crossPartition(replicaGroups);
 
-  for (auto processGroup : processGroups)
-    if (llvm::find(processGroup, process->getId()) != processGroup.end())
-      return process->rendezvous(processGroup, channelId, operand)
-          .lookup(processGroup[0]);
+  auto processGroup = processGroups.findGroup(process->getId());
+  if (processGroup)
```
mlevesquedion, line 1109: Nit: `if (auto processGroup = ...) ...` would be slightly cleaner, since `processGroup` only needs to be in scope inside the `if`.

stablehlo/reference/Ops.cpp
mlevesquedion: Ah good point. I notice that this code is duplicated in a few places though, consider adding a helper.

stablehlo/reference/Ops.cpp
ghpvnist: Ack. This pattern exists throughout the file, and I plan to create a PR to clean them up later.

PR #1984, opened 2024-02-05
Chlo decompose patterns

linuxlonelyeagle:
I had added Chlo conversion to Stablehlo pass, and Canonicalize @GleasonK .


linuxlonelyeagle:
Here's some more trivia, should I fix them?
```
/my_stablehlo/stablehlo/transforms/StablehloCanonicalize.cpp:45:13: warning: comparison of integers of different signs: 'std::tuple_element<0, std::tuple<unsigned long, const long long &>>::type' (aka '__type_pack_element<0UL, unsigned long, const long long &>') and 'const long long' [-Wsign-compare]
    if (idx != value) {
        ~~~ ^  ~~~~~
/my_stablehlo/stablehlo/transforms/StablehloCanonicalize.cpp:692:53: warning: comparison of integers of different signs: 'size_t' (aka 'unsigned long') and 'int64_t' (aka 'long long') [-Wsign-compare]
        op.getKnownNonexpandingDimensions()->size() != resultType.getRank()) {
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~
/my_stablehlo/stablehlo/transforms/StablehloCanonicalize.cpp:53:13: warning: unused function 'isIotaRange' [-Wunused-function]
static bool isIotaRange(ElementsAttr attr) {
```

stablehlo/transforms/CHLODecompositionPatterns.td
```diff
@@ -0,0 +1,383 @@
+// Copyright 2020 The IREE Authors
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+// This is the legalization pattern definition file for CHLO to StableHLO.
```
GleasonK, line 7: Rename file for consistency: `stablehlo/transforms/ChloDecompositionPatterns.td`

stablehlo/transforms/Passes.td
```diff
@@ -96,3 +96,13 @@ def VhloToVersionPass : Pass<"vhlo-to-version"> {
            "The target version. Must be a version of the form #.#.# or 'current'.">,
   ];
 }
+
+def LegalizeChlo :
```
GleasonK, line 100: For consistency:

`ChloLegalizeToStablehloPass` and `chlo-legalize-to-stablehlo`

stablehlo/transforms/Passes.td
```diff
@@ -96,3 +96,13 @@ def VhloToVersionPass : Pass<"vhlo-to-version"> {
...
+    InterfacePass<"stablehlo-legalize-chlo", "mlir::FunctionOpInterface"> {
+  let summary = "Legalizes from CHLO ops flow to StableHLO and Shape ops";
+}
+
+def StableHLOCanonicalize :
```
GleasonK, line 105: 1. Can we split this pass into a separate PR for review? Will make it easier to see what's a CHLO specific infra and what isn't. I.e. I'm not sure if there is a test to check that this pass is working properly yet? Would be easier to review / give feedback as a standalone PR.
2. Per feedback on: [openxla-discuss](https://groups.google.com/a/openxla.org/g/openxla-discuss/c/httnobiya7U) Let's rename this pass `StablehloAggressiveSimplificationPass` and `stablehlo-aggressive-simplification`, and we can gradually move some to native canonicalizers.

stablehlo/tests/chlo/legalize_chlo_decomposition.mlir
```diff
@@ -0,0 +1,533 @@
+// RUN: stablehlo-opt --pass-pipeline="builtin.module(func.func(stablehlo-legalize-chlo))" \
+// RUN:   --split-input-file --verify-diagnostics %s | FileCheck %s
+
+// CHECK-LABEL: func.func @asin_bf16(
```
GleasonK, line 4: I'll add more test points once this is submitted, can copy some over from CHLO->MHLO. This is a good start.

GleasonK (changes requested):
Thanks for upstreaming these passes!! 🚀 

Minor feedback on file renaming / pass naming, and lets split off the canonicalize pass to a separate PR and give it a simple test point to make sure its working, keep this PR specific to CHLO decomposition. I can add more tests once these are submitted.

GleasonK:
> Here's some more trivia, should I fix them?
> 
> ```
> /my_stablehlo/stablehlo/transforms/StablehloCanonicalize.cpp:45:13: warning: comparison of integers of different signs: 'std::tuple_element<0, std::tuple<unsigned long, const long long &>>::type' (aka '__type_pack_element<0UL, unsigned long, const long long &>') and 'const long long' [-Wsign-compare]
>     if (idx != value) {
>         ~~~ ^  ~~~~~
> /my_stablehlo/stablehlo/transforms/StablehloCanonicalize.cpp:692:53: warning: comparison of integers of different signs: 'size_t' (aka 'unsigned long') and 'int64_t' (aka 'long long') [-Wsign-compare]
>         op.getKnownNonexpandingDimensions()->size() != resultType.getRank()) {
>         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~
> /my_stablehlo/stablehlo/transforms/StablehloCanonicalize.cpp:53:13: warning: unused function 'isIotaRange' [-Wunused-function]
> static bool isIotaRange(ElementsAttr attr) {
> ```

As mentioned above, lets split canonicalize to a separate PR. Then yes, it would be great to fix these if they're simple.

stablehlo/transforms/CMakeLists.txt
```diff
@@ -25,9 +29,12 @@ add_mlir_dialect_library(StablehloPasses
   StablehloRefineShapes.cpp
   VhloLegalizeToStablehlo.cpp
   VhloToVersion.cpp
-
+  ChloLegalizeToStablehlo.cpp
```
GleasonK, line 32: We'll need `BUILD.bazel` changes as well. I'll send a diff patch you can apply with required bazel changes shortly.

stablehlo/transforms/CMakeLists.txt
```diff
@@ -16,6 +16,10 @@ set(LLVM_TARGET_DEFINITIONS Passes.td)
 mlir_tablegen(Passes.h.inc -gen-pass-decls)
 add_public_tablegen_target(PassesIncGen)
 
+set(LLVM_TARGET_DEFINITIONS ChloDecompositionPatterns.td)
```
GleasonK, line 19: This will work after renaming `CHLODecompositionPatterns.td` to ChloDecompositionPatterns.td`, but for now doesn't build on my machine.

stablehlo/transforms/Passes.td
linuxlonelyeagle: Sorry, I don't have access to Google Groups.I can't see what's in the discuss.

stablehlo/transforms/CHLODecompositionPatterns.td
linuxlonelyeagle: Because CHLO decomposition includes canonicalize,I was afraid that I didn't send a notice to test it,so I added it,sorry.

stablehlo/transforms/Passes.td
GleasonK: Ah! If you share an email address I can CC you into the thread. Feel free to send me a message from your email to gleasonk@google[dot]com, if interested :)

To sum up the feedback, there is strong support for these passes in StableHLO, but disagreement on which of these passes is a canonicalization vs an opinionated transform, so it was suggested we name it "StablehloAggressiveSimplificationPass" and gradually migrate agreed upon patterns to canonicalizations.

stablehlo/transforms/CHLODecompositionPatterns.td
GleasonK: Ahhh. I see that now. OK we can proceed with this as is. Let's just do file renames and then we can add more fine-grained tests later. Ideally I'd like for CHLO decompositions to not depend on canonicalizations, but that will be a bit of a migration for the test files so we can do it separately

stablehlo/transforms/Passes.td
linuxlonelyeagle: Thanks!

GleasonK:
Looks like we need to address those compilation warnings you mentioned to get CI working. If you can get CMakeBuild working, I'll send you a patch to make the Bazel CI work. Holding off on sending that patch until existing feedback is addressed / CMake is working to avoid any merge conflicts.

linuxlonelyeagle:
> Looks like we need to address those compilation warnings you mentioned to get CI working. If you can get CMakeBuild working, I'll send you a patch to make the Bazel CI work. Holding off on sending that patch until existing feedback is addressed / CMake is working to avoid any merge conflicts.

ok,It may be a little late to address this problem,I read the error reported by ci and I know what the problem is.

stablehlo/transforms/CMakeLists.txt
linuxlonelyeagle: Really appreciate everything you do.

BUILD.bazel
```diff
@@ -929,13 +947,18 @@ cc_library(
         ":interpreter_ops",
         ":stablehlo_ops",
         ":stablehlo_ops_inc_gen",
+        ":broadcast_utils",
+        ":chlo_rewriters_inc_gen",
```
GleasonK, line 951: This section needs to be sorted, will fix buildifier CI lint:

```
        ":base",
        ":broadcast_utils",
        ":chlo_ops",
        ":chlo_rewriters_inc_gen",
        ":interpreter_ops",
```

GleasonK: Sorry there were 2 errors in the patch I sent, luckily all the other CI is passing so should be good for submit after these fixes!

stablehlo/tests/chlo/legalize_chlo_decomposition.mlir
GleasonK: Created https://github.com/openxla/stablehlo/issues/1997 so I don't forget

GleasonK (approved):
LGTM! Thanks again for doing this! I'll add some extra test cases and might so some additional refactoring as I start trying to integrate this pass elsewhere. In the near future I plan to migrate the `CHLO->MHLO` pass to use `CHLO->StableHLO->MHLO`, this PR was an important first step. 🚀

linuxlonelyeagle:
> LGTM! Thanks again for doing this! I'll add some extra test cases and might so some additional refactoring as I start trying to integrate this pass elsewhere. In the near future I plan to migrate the `CHLO->MHLO` pass to use `CHLO->StableHLO->MHLO`, this PR was an important first step. 🚀

I appreciate everything you do. I hope to continue working together in the future :wink:.

stablehlo/transforms/StablehloAggressiveSimplification.cpp
```diff
@@ -0,0 +1,1164 @@
...
+      Region *retRegion = retOp->getParentRegion();
+      if (llvm::any_of(retOp.getResults(), [retRegion](Value result) {
+            return result.getParentRegion() == retRegion;
+          })) {
+        return failure();
+      }
+
+      rewriter.replaceOp(op, retOp.getResults());
+      return success();
+    }
```
tlsdmstn56, line 716: I have a question regarding this pattern. If reduction is no-op, doesn't returning init value make more sense? I couldn't find what it is supposed to return when reduction is no-op but returning the value from the outside of the region seems weird. Maybe I misunderstand the spec so please correct me if I am wrong.

linuxlonelyeagle: I don't know if I'm understanding this correctly, it's been a long time since I've played stablehlo.
* The pattern here should just do some checking and then some simple transformations.It mainly checks if the result of the `returnOp` is all from the region of the `reduceOp`. In other words, the operands of returnOp should all come from the region of reduceOp.
* If reduction is no-op, doesn't returning init value make more sense?Are you talking about returnOp?
  Since the check here is wrong, then the reduceOp should be wrong and just return failure.

PR #1986, opened 2024-02-05
VHLO legalization of StableHLO UniformQuantizedPerAxisType

sdasgup3:
The PR enables  
1. VHLO legalization of StableHLO UniformQuantizedPerAxisType.
2. Writing/reading from bytecode format. 
 
The mlir::quant dialect has two different Quantized types, [one](https://github.com/llvm/llvm-project/blob/e2bb91b25c8740625fecd127c1d908a2fabd0102/mlir/include/mlir/Dialect/Quant/QuantTypes.h#L255) for per-tensor and the [other](https://github.com/llvm/llvm-project/blob/e2bb91b25c8740625fecd127c1d908a2fabd0102/mlir/include/mlir/Dialect/Quant/QuantTypes.h#L315) for per-axis. Following the same analogy, we are now having two types in VHLO  `UniformQuantizedV1Type` and `UniformQuantizedPerAxisV1Type` resp. Also added the legalization of StableHLO -> VHLO and VHLO -> StableHLO for the corresponding types. 

Regarding testing the compatibility, we only added this feature (of new per-axis type) to [stablehlo_legalize_to_vhlo.0_17_0.mlir](https://github.com/openxla/stablehlo/compare/main...sdasgup3:serialize-per-axis-quantization-type?expand=1#diff-d78cbc82314e64545ee7b9f6a66a0b910f4a8ac2e7f27e2ad55c4b33d2ea409c. @GleasonK Please let me know if the feature  needs to be added to earlier version as well.


stablehlo/dialect/VhloTypes.td
```diff
@@ -228,6 +228,71 @@ def VHLO_UniformQuantizedV1 : VHLO_TypeDef<"UniformQuantizedV1", "quant_v1", "0.
...
+  }];
+  let printer = [{
+    $_printer << '[';
+    llvm::interleaveComma($_self, $_printer, [&](int64_t zeroPoint) {
+      $_printer << zeroPoint;
+    });
+    $_printer << ']';
+  }];
+}
+def VHLO_UniformQuantizedPerAxisV1 : VHLO_TypeDef<"UniformQuantizedPerAxisV1", "quant_per_axis_v1", "0.9.0", "current"> {
```
GleasonK, line 272: Needs a version bump. If we sent this attr to a binary running `0.9.0` it wouldn't know how to deserialize this.

stablehlo/dialect/VhloTypes.td
```diff
@@ -228,6 +228,71 @@ def VHLO_UniformQuantizedV1 : VHLO_TypeDef<"UniformQuantizedV1", "quant_v1", "0.
   }];
   let assemblyFormat = "`<` $storageType `` `:` `` $expressedType `,` $scale `` `:` `` $zeroPoint `,` $storageTypeMin `` `:` `` $storageTypeMax `,` $flags `>`";
 }
+def VHLO_QuantizationScalesV1 : ArrayRefParameter<"::llvm::APFloat", "array of double scales"> {
```
GleasonK, line 231: Can we just use ArrayRefParameter without the indirection?

stablehlo/dialect/VhloTypes.td
```diff
@@ -228,6 +228,71 @@ def VHLO_UniformQuantizedV1 : VHLO_TypeDef<"UniformQuantizedV1", "quant_v1", "0.
...
+  }];
+  let printer = [{
+    $_printer << '[';
+    llvm::interleaveComma($_self, $_printer, [&](APFloat scale) {
+      $_printer << scale;
+    });
+    $_printer << ']';
+  }];
+}
+def VHLO_QuantizationZeroPointsV1 : ArrayRefParameter<"int64_t", "array of int64_t zero points"> {
```
GleasonK, line 252: Same as above

stablehlo/dialect/VhloTypes.td
GleasonK, line 272: I'm still debating if we should do this 1:1 with how we spec it, vs how it's implemented in the quant dialect. Let's discuss this more, i.e. should we instead introduce `VHLO_UniformQuantizedV2`.

GleasonK (changes requested):

stablehlo/dialect/VhloTypes.td
sdasgup3: The [mlir::FieldParser](https://github.com/llvm/llvm-project/blob/6b42625b1f983f6aafb9f4fe2953970c73963603/mlir/include/mlir/IR/DialectImplementation.h#L62) does not have any specialization for `mlir::APFloat` and hence we needed to prove the parser support for that. IMO, [here](https://github.com/openxla/stablehlo/blob/b0edcdd6810551bf6aaec7d5aa00451cfe207582/stablehlo/dialect/VhloTypes.td#L196) where we cannot reply on the `FeildParser`'s existing specialization for parsing. 

For the ZeroPoints, the indirection can be removed as  `FieldParser<int>::parse` is supported.

stablehlo/dialect/VhloTypes.td
sdasgup3: Agreed. Done. PTAL.

stablehlo/dialect/VhloTypes.td
sdasgup3: done.

stablehlo/dialect/VhloDialect.td
```diff
@@ -35,6 +35,7 @@ def VHLO_Dialect : Dialect {
       0.15.0: MLIR bytecode version 5 => 6, use properties in VHLO.
       0.16.0: Introduce `collective_broadcast` operation.
       0.17.0: Allow reduce operations to promote to higher bitwidth.
+      0.18.0: Allow serialization of UniformQuantizedPerAxisType.
```
GleasonK, line 38:
```
0.18.0: Introduce `UniformQuantizedPerAxisType` type.
```

stablehlo/dialect/VhloTypes.td
GleasonK: Hmm. We have parsing/printing defined for `VHLO_APFloatV1`. I wonder if there's some way to leverage that. `ArrayRefParameter<VHLO_APFloatV1>` or `ArrayRefParameter<"vhlo::APFloatV1Type">`

stablehlo/dialect/VhloBytecode.cpp
```diff
@@ -1197,6 +1215,70 @@ void VhloBytecodeInterface::write(TupleV1Type type,
   writer.writeTypes(type.getTypes());
 }
 
+//===----------------------------------------------------------------------===//
+// UniformQuantizedPerAxisV1Type
+//===----------------------------------------------------------------------===//
+
+UniformQuantizedPerAxisV1Type
```
GleasonK, line 1222: Let's make a better use of read/writeList. Here is the bytecode for upstream quantized per axis:


```
static ::mlir::Type readUniformQuantizedPerAxisType(MLIRContext* context, DialectBytecodeReader &reader) {
  uint64_t flags;
  Type storageType, expressedType;
  uint64_t quantizedDimension;
  int64_t storageTypeMin, storageTypeMax;
  SmallVector<double> scales;
  SmallVector<int64_t> zeroPoints;
  auto readScales = [&]() -> FailureOr<double> {
    double temp;
    if (succeeded(readDoubleAPFloat(reader, temp))) {
      return temp;
    }
    return failure();
  };
  auto readZeroPoints = [&]() -> FailureOr<int64_t> {
    int64_t temp;
    if (succeeded(reader.readSignedVarInt(temp))) {
      return temp;
    }
    return failure();
  };
  if (succeeded(reader.readVarInt(flags)) &&
      succeeded(reader.readType(storageType)) &&
      succeeded(reader.readType(expressedType)) &&
      succeeded(reader.readVarInt(quantizedDimension)) &&
      succeeded(reader.readSignedVarInt(storageTypeMin)) &&
      succeeded(reader.readSignedVarInt(storageTypeMax)) &&
      succeeded(reader.readList(scales, readScales)) &&
      succeeded(reader.readList(zeroPoints, readZeroPoints))) {
    return get<UniformQuantizedPerAxisType>(context, flags, storageType, expressedType, scales,
            zeroPoints, quantizedDimension, storageTypeMin, storageTypeMax);
  }
  return UniformQuantizedPerAxisType();
}

static void write(UniformQuantizedPerAxisType type, DialectBytecodeWriter &writer) {
  writer.writeVarInt(/* UniformQuantizedPerAxisType */ 5);
  writer.writeVarInt(type.getFlags());
  writer.writeType(type.getStorageType());
  writer.writeType(type.getExpressedType());
  writer.writeVarInt(type.getQuantizedDimension());
  writer.writeSignedVarInt(type.getStorageTypeMin());
  writer.writeSignedVarInt(type.getStorageTypeMax());
  writer.writeList(type.getScales(), [&](double type) {
    writer.writeAPFloatWithKnownSemantics(APFloat(type));
  });
  writer.writeList(type.getZeroPoints(), [&](int64_t type) {
    writer.writeSignedVarInt(type);
  });
}
```

Can we make use of this? As 1:1 as we can get it would be ideal.

stablehlo/dialect/VhloTypes.cpp
```diff
@@ -120,6 +120,18 @@ void VhloTypeConverter::addBuiltinToVhloConversions() {
         convertedExpressedType, APFloat(type.getScale()), type.getZeroPoint(),
         type.getStorageTypeMin(), type.getStorageTypeMax());
   });
+  addConversion([&](quant::UniformQuantizedPerAxisType type) -> Type {
+    Type convertedStorageType = convertType(type.getStorageType());
+    Type convertedExpressedType = convertType(type.getExpressedType());
+    if (!convertedStorageType || !convertedExpressedType) return {};
+    SmallVector<APFloat> scales = llvm::to_vector(llvm::map_range(
```
GleasonK, line 127: Why is this handling different from ZP handling?

stablehlo/dialect/VhloTypes.td
```diff
@@ -228,6 +228,49 @@ def VHLO_UniformQuantizedV1 : VHLO_TypeDef<"UniformQuantizedV1", "quant_v1", "0.
   }];
   let assemblyFormat = "`<` $storageType `` `:` `` $expressedType `,` $scale `` `:` `` $zeroPoint `,` $storageTypeMin `` `:` `` $storageTypeMax `,` $flags `>`";
 }
+def VHLO_QuantizationScalesV1 : ArrayRefParameter<"::llvm::APFloat", "array of double scales"> {
+  let parser = [{
+    [&]() -> FailureOr<llvm::SmallVector<::llvm::APFloat>> {
```
GleasonK, line 233: Is there a reason we use APFloat while upstream uses double? Did this change happen after we added support? Is there a benefit to one over the other?

GleasonK (changes requested):
Close to ready!

stablehlo/dialect/VhloBytecode.cpp
sdasgup3: makes sense. updated. PTAL.

stablehlo/dialect/VhloTypes.cpp
sdasgup3: Actually, the `vhlo::UniformQuantizedPerAxisV1Type` takes APFloat for the scales whereas the `mlir::quant::UniformQuantizedPerAxisType` takes double.

sdasgup3:
>> We have parsing/printing defined for VHLO_APFloatV1. I wonder if there's some way to leverage that. ArrayRefParameter<VHLO_APFloatV1> or ArrayRefParameter<"vhlo::APFloatV1Type">

Yes, I tried to re-use the `APFloatV1`, but it did not work. Unlike the other `VHLO_Typedef` (like `VHLO_BooleanV1`), there is no `APFloatV1Type`. Similarly, no type defined with `VHLO_APFloatV1`.

stablehlo/dialect/VhloTypes.td
sdasgup3: Yes, I tried to re-use the APFloatV1, but it did not work. Unlike the other VHLO_Typedef (like VHLO_BooleanV1), there is no APFloatV1Type. Similarly, no type defined with VHLO_APFloatV1.

stablehlo/dialect/VhloTypes.td
sdasgup3: Per communication with Jacques Pienaar, `Quantization dialect was one of the first, so it predates generating types `. 

IMO having `APFLoat` for VHLO dialect here allows us to use the [APFloat::hash_value()](https://github.com/llvm/llvm-project/blob/5b780c8c6c558ec283a9eec485a4f172df0f9fe1/llvm/include/llvm/ADT/APFloat.h#L1348) which otherwise needs to be computed explicitly ([here](https://github.com/llvm/llvm-project/blob/5ce2f73b2e5e6664d74b49ee45f11505f8306577/mlir/lib/Dialect/Quant/IR/TypeDetail.h#L131), Thanks @GleasonK for pointing to that).

stablehlo/dialect/VhloTypes.td
GleasonK: This doesn't have implicaitons on bytecode I suppose, since it will get written as an APFloat so lgtm.

stablehlo/dialect/VhloBytecode.cpp
```diff
@@ -1197,6 +1215,66 @@ void VhloBytecodeInterface::write(TupleV1Type type,
...
+    DialectBytecodeReader &reader) const {
+  LOG_READ_CALL;
+  uint64_t flags;
+  Type storageType, expressedType;
+  uint64_t quantizedDimension;
+  int64_t storageTypeMin, storageTypeMax;
+  SmallVector<APFloat> scales;
+  SmallVector<int64_t> zeroPoints;
+  auto readScales = [&]() -> FailureOr<APFloat> {
+    auto valOr =
```
GleasonK, line 1233: How about `return reader.readAPFloatWithKnownSemantics(llvm::APFloat::IEEEdouble());` ?

I believe that returns `FailureOp<APFloat>` as it is?

GleasonK (approved):

stablehlo/dialect/VhloBytecode.cpp
sdasgup3: that is right! done.

stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.mlir
```diff
@@ -2387,13 +2387,20 @@ func.func @type_dynamism_unranked(%arg0: tensor<*xf32>) -> tensor<*xf32> {
   func.return %0 : tensor<*xf32>
 }
 
-// CHECK-LABEL: "type_quantization"
-func.func @type_quantization(%arg0: tensor<!quant.uniform<i8:f32, 34.0:16>>, %arg1: tensor<f32>) -> tensor<f32> {
+// CHECK-LABEL: "type_per_tensor_quantization"
```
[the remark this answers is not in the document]
sdasgup3: done

stablehlo/tests/vhlo/vhlo_to_version_downgrade_invalid.0_16_0.mlir
```diff
@@ -103,21 +103,29 @@ func.func @select_and_scatter_with_promotable_types(
...
+    padding = dense<0> : tensor<4x2xi64>
+  } : (tensor<10x24x24x64xf32>, tensor<10x12x12x64xf32>, tensor<f32>) ->
+        tensor<10x24x24x64xf64>
+  func.return
+}
+
+// -----
+
+// expected-error @+1 {{failed to legalize operation 'vhlo.func_v1' that was explicitly marked illegal}}
+func.func @type_per_axis_quantization(%arg0: tensor<2x!quant.uniform<i8:f32:0, {34.0:16, 34.0:16}>>) -> tensor<2x!quant.uniform<i8:f32:0, {34.0:16, 34.0:16}>> {
```
GleasonK, line 128: Can we put this in a new file `downgrade_invalid.0.17.0`, helps keep track of what's invalid at each step. Unclear if this downgrade would work at `0.17.0` in this file since we only test `0.16.0` in this file.

sdasgup3: The downgrade should not work on 0.17.0. Added `downgrade_invalid.0.17.0` to capture that. PTAL

build_tools/github_actions/lint_version.sh
```diff
@@ -21,8 +21,8 @@
 VERSION_H="stablehlo/dialect/Version.h"
 set_version_var() {
   # getCurrentVersion() { Version(0, X, Y); }
-  VERSION_STR=$(cat $VERSION_H | grep getCurrentVersion -A1 | grep -o 'Version([0-9], .*)')
-  REGEX="Version\(([0-9]+), ([0-9]+), ([0-9]+)\)"
+  VERSION_STR=$(cat $VERSION_H | grep getCurrentVersion -A1 | grep -o 'Version(.*[0-9])')
+  REGEX="Version\(/\*.*=\*/([0-9]+), /\*.*=\*/([0-9]+), /\*.*=\*/[^0-9]*([0-9]+)\)"
```
sdasgup3, line 25: @GleasonK This is the additional change we have to incorporate some of the readability comments. PTAL

PR #1987, opened 2024-02-05
Rename function to closely match spec

ghpvnist:
As an example, looking at the spec of `collective_broadcast`, we come across the statement:

`broadcast_in_dim(constant(0, element_type(result)), [], type(result))` otherwise.

This PR renames `makeScalar()` to `constant()` with arguments matching the spec above for parity.


GleasonK (approved):

GleasonK (changes requested):
Bazel CI failing, unclear why, looks like some change is needed?

mlevesquedion (approved):
Nice!

GleasonK (approved):

PR #1990, opened 2024-02-05
Avoid duplicating extra class declarations

mlevesquedion:
I noticed some duplication and wondered if it was possible to remove it. Then I stumbled upon this thread which provided a solution: https://discourse.llvm.org/t/how-to-add-new-content-to-extraclassdeclaration-in-subclasses/71537

The idea is to declare the common class declaration in the base class and concatenate it with the child-specific declarations.

In some cases, the field was being overriden only to be set to the same value as the parent. In such cases, I simply removed the override.


ghpvnist (approved):
LGTM, I think MHLO can also benefit from this change, so I added a tag to migrate to MHLO.

mlevesquedion:
Thanks, I'll look into it!

PR #1992, opened 2024-02-06
Add a simple example for execution order

mlevesquedion:
The text refers to "the example program above", but said example is actually near the top of the file (thousands of lines above). The example is also longer than necessary for the purposes of illustrating execution order.


test.mlir
ghpvnist, line 1: Extraneous file?

ghpvnist (approved):
LGTM with minor comment!

test.mlir
mlevesquedion: Ah yep, sorry about that. Thanks!

Issue #1991, opened 2024-02-05
Implement Dynamism RFC


mlevesquedion:
This is a work tracker for the implementation of the Dynamism 101 RFC, which was recently submitted: https://github.com/openxla/stablehlo/pull/1881

Tasks (some tasks may be missing):
- [x] Implement (P1): #2135, #2286, #2287
- [x] Implement (P2): #2136, #2192
- [x] Implement (P3): #2210
- [x] Implement (P4): #2213
- [x] Implement (P5): #2049, #2065, #2116

GleasonK:
Adding link to: https://discord.com/channels/999073994483433573/999074539138990131/1219726965783789629

Consider `dynamic_broadcast_in_dim` constraints on `?` dimensions:

```
func.func @main(%arg0: tensor<1x?x1x64xf32>, %arg1: tensor<4xindex>) -> tensor<1x?x32x64xf32> {
    %0 = stablehlo.dynamic_broadcast_in_dim %arg0, %arg1, dims = [0, 1, 2, 3] : (tensor<1x?x1x64xf32>, tensor<4xindex>) -> tensor<1x?x32x64xf32>
    return %0 : tensor<1x?x32x64xf32>
}
```

Should `(tensor<1x?x1x64xf32>, tensor<4xindex>) -> tensor<1x?x32x64xf32>` be permitted, since only dims of size `1` should be expandable, and to my knowledge, we frameworks don't resolve `?` to `1`?

srcarroll:
I think there also needs to be more constraints of `stablehlo.dynamic_reshape`. For example the following is valid
```
func.func @dynamic_reshape(%arg0: tensor<?x?x?x?xf32>, %shape: tensor<2xindex>) -> tensor<?x?xf32> {
  %0 = "stablehlo.dynamic_reshape"(%arg0, %shape) : (tensor<?x?x?x?xf32>, tensor<2xindex>) -> tensor<?x?xf32>
  func.return %0 : tensor<?x?xf32>
}
```
But this is not well defined.  It's impossible to know which input dims are collapsing to which output dims.  Unless I am missing something?

PR #1995, opened 2024-02-07
Disallow unranked dynamism in rng_op

mlevesquedion:
I missed this in https://github.com/openxla/stablehlo/pull/1970.

https://github.com/openxla/stablehlo/issues/1991


GleasonK (approved):
LGTM. Longer term we can probably make `HLO_PredIntOrFpTensor` enforce rank, no need to explicitly call everything `Ranked` once we have that rule as the default.

Maybe add a todo comment if you agree?

mlevesquedion:
> LGTM. Longer term we can probably make `HLO_PredIntOrFpTensor` enforce rank, no need to explicitly call everything `Ranked` once we have that rule as the default.
> 
> Maybe add a todo comment if you agree?

Yeah that SGTM! I'll add a TODO before merging. I'll also add the TODO to other instances of this introduced in https://github.com/openxla/stablehlo/issues/1991.

PR #1999, opened 2024-02-08
Convert lambda call to a function call

ghpvnist:
The `hasDuplicates` function was originally factored out because it is used frequently. However it was still defined as lambda, so I've updated it to be a function.

There was also an unnecessary conversion to a vector (the underlying data does not change), so I've removed the conversion too.


mlevesquedion (approved):
Looks good, thanks for cleaning up the code!

PR #2000, opened 2024-02-08
Rename function to match spec

ghpvnist:
With the recent change to rename a function to match spec in #1987, I noticed that several ops also rely on this meta function that can also benefit from name change.

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -68,9 +68,9 @@ namespace hlo {
 //===----------------------------------------------------------------------===//
 
 // Checks if the vector `nums` has duplicates.
-bool hasDuplicates(ArrayRef<int64_t> nums) {
+bool isUnique(ArrayRef<int64_t> nums) {
   llvm::SmallDenseSet<int64_t> set(nums.begin(), nums.end());
-  return set.size() != nums.size();
+  return set.size() == nums.size();
 }
 
 bool tensorsHaveSameElType(TypeRange types, bool ignoreFpPrecision = true) {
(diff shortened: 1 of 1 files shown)


mlevesquedion (commented):
Thank you for improving the parity between the spec and the implementation!

The name `is_unique` isn't super clear to me. Naively, I would think that it means "is this thing unique", which doesn't make sense (unique relative to what?). The only logical interpretation is "are the elements of this container unique", i.e. "do all elements of this container occur only once in the container", but that's not really what the name is saying. IMO something like `allUnique` would be clearer. Up to you though, if you want to make that change you would need to update the spec as well. Another option would be to keep this code has is and change the name of the predicate in the spec to `has_no_duplicates` instead of `is_unique` (IMO this would be the best option here as that would allow us to keep the very readable name `hasDuplicates` in the code and it is also the name that most clearly describes the property we are interested in).

ghpvnist:
Naming is tricky :) The other synonym I could think of is `is_distinct`, but I guess this has the same source of confusion (distinct to what?). I agree there can be source of confusion, but I still like the idea of prefixing with the canonical `is_` for boolean functions (other similar meta functions in the spec also prefix with `is_` e.g. `is_nan` or `is_sorted`). Thoughts?

mlevesquedion:
> Naming is tricky :)

Agreed!

> I still like the idea of prefixing with the canonical `is_` for boolean functions (other similar meta functions in the spec also prefix with `is_` e.g. `is_nan` or `is_sorted`). Thoughts?

I think for predicates both `is_` and `has_` are quite common as prefixes. IMO in this case `is_unique` is not the clearest name because it is a property of the elements of the container, not of the container itself, yet we are applying it to the container. `has_no_duplicates` is a property of the container, which IMO aligns with our intent more clearly.

I don't have strong feelings here though, I'm happy to approve the change. Anecdotally it seems the name `is_unique` is used in `pandas`: https://pandas.pydata.org/docs/reference/api/pandas.Series.is_unique.html

ghpvnist:
Great to see we're not coining a new term for it. If we're not the first ones, it's probably okay to keep it then. Thanks for the reference!

mlevesquedion (approved):

PR #2004, opened 2024-02-12
Fix verification of DynamicBroadcastInDimOp to handle unranked operand

sdasgup3:
Verification of `dynamic_broadcast_in_dim` does not address the event when the operand is of unranked type and crashes instead on nullptr dereference.  

The PR fixes the issue and adds a relevant test to highlight the problem.


mlevesquedion (approved):
Thanks for the fix! FYI as part of #1991 I will remove unranked dynamism completely. Thank you for handling the null case from `dynamic_cast`. I will update this code once `RankedTensor` is enforced at the Tablegen level.

PR #2006, opened 2024-02-13
Bump Bazel to 7.0

### LINKED ISSUE #1878: Migrate to bazel 7

fzakaria:
Major change: Bump Bazel to the new LTS at 7.X

* Included a default Python toolchain 3.10

  There is a problem at the moment running lit using Python311 where it cannot find the module.
  Our CI system already enforces Python310 to get around it but we should set the toolchain
  in Bazel itself to avoid this problem.
  I had already filed https://github.com/llvm/llvm-project/issues/75963

* I moved skylib to MODULE.bazel to test the MODULE portion of Bazel is working. A remaining action item is to move LLVM itself to MODULE.bazel which would preclude needing to define skylib since its a dependency of LLVM.

* Workspace names are no longer named in Bazel 7+
  
  I had to change stablehlo to _main to fix the runfiles directory.
  Please see https://github.com/bazelbuild/bazel/issues/18128

  We should consider using https://github.com/bazelbuild/rules_python/blob/main/python/runfiles/runfiles.py#L262 instead to get the runfiles directory for future proofing.

Fixes #1878


stablehlo/conversions/linalg/tests/BUILD.bazel
```diff
@@ -25,8 +25,8 @@ expand_template(
     testonly = True,
     out = "lit.site.cfg.py",
     substitutions = {
-        "\"@STABLEHLO_SOURCE_DIR@\"": "os.path.join(os.environ['TEST_SRCDIR'], 'stablehlo')",
-        "\"@STABLEHLO_TOOLS_DIR@\"": "os.path.join(os.environ['TEST_SRCDIR'], 'stablehlo')",
+        "\"@STABLEHLO_SOURCE_DIR@\"": "os.path.join(os.environ['TEST_SRCDIR'], '_main')",
```
mlevesquedion, line 28: I'm not clear on why this change is needed. Can you expand on what this `_main` thing is?

WORKSPACE.bazel
```diff
@@ -17,17 +17,6 @@ workspace(name = "stablehlo")
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
-SKYLIB_VERSION = "1.3.0"
-
-http_archive(
```
mlevesquedion, line 22: Will the other `http_archive`s in this file also need to be migrated?

mlevesquedion (approved):
LGTM, thanks for doing this!

stablehlo/conversions/linalg/tests/BUILD.bazel
fzakaria: Tried to write in the PR description but I can see it's not very clear.
Basically Bazel 7 / bazelmod changed the layout directory structure such that the repo name is `_main` in the tree structure (https://bazel.build/remote/output-directories) and not the workspace name listed in `WORKSPACE`.

The ideal thing is to use the runfiles library in Python which handles all this.
`from python.runfiles import runfiles`

WORKSPACE.bazel
fzakaria: Eventually -- but LLVM needs to change to support bazelmod.

I spoke to Bazel folks on Slack and they said for now no need to port.

https://bazelbuild.slack.com/archives/CA31HN1T3/p1707853906929909?thread_ts=1707845158.715749&cid=CA31HN1T3

MODULE.bazel
```diff
@@ -0,0 +1,19 @@
+###############################################################################
+# Bazel now uses Bzlmod by default to manage external dependencies.
+# Please consider migrating your external dependencies from WORKSPACE to MODULE.bazel.
+#
+# For more details, please check https://github.com/bazelbuild/bazel/issues/18958
+###############################################################################
```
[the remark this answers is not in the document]
mlevesquedion: SG. And it does look like we have more deps to migrate so this makes sense.

stablehlo/conversions/linalg/tests/BUILD.bazel
mlevesquedion: Okay gotcha, makes sense. Thanks!

WORKSPACE.bazel
mlevesquedion: Cool, thanks for looking into it!

stablehlo/conversions/linalg/tests/BUILD.bazel
sjain-stanford: Fly-by comment - this would break some of our downstream integrations pulling in stablehlo (and running its tests as part of downstream CI). Just a thought: Would adding `common --noenable_bzlmod` to the `.bazelrc` (like [this](https://github.com/cruise-automation/mlir-tcp/blob/main/.bazelrc#L48-L50)) resolve this, at least until stablehlo actually migrates to using bzlmod (depends on llvm migrating over as you clarified)?

stablehlo/tests/BUILD.bazel
```diff
@@ -134,8 +134,8 @@ expand_template(
     testonly = True,
     out = "lit.site.cfg.py",
     substitutions = {
-        "\"@STABLEHLO_SOURCE_DIR@\"": "os.path.join(os.environ['TEST_SRCDIR'], 'stablehlo')",
-        "\"@STABLEHLO_TOOLS_DIR@\"": "os.path.join(os.environ['TEST_SRCDIR'], 'stablehlo')",
+        "\"@STABLEHLO_SOURCE_DIR@\"": "os.path.join(os.environ['TEST_SRCDIR'], '_main')",
+        "\"@STABLEHLO_TOOLS_DIR@\"": "os.path.join(os.environ['TEST_SRCDIR'], '_main')",
```
sjain-stanford, line 138: Just a fly-by comment: I think this might break some of our downstream integrations pulling in stablehlo (and running its tests as part of downstream CI). A thought: Would adding `common --noenable_bzlmod` to the `.bazelrc` (like [this](https://github.com/cruise-automation/mlir-tcp/blob/main/.bazelrc#L48-L50)) resolve this, at least until stablehlo actually migrates to using bzlmod (depends on llvm migrating over as you clarified)?

fzakaria: I am playing around with:

```
❯ cat s\tablehlo/tests/utils.bzl

def _is_bzlmod_enabled():
    """Determine whether bzlmod mode is enabled."""
    # If bzlmod is enabled, then `str(Label(...))` returns a canonical label,
    # these start with `@@`.
    return str(Label("//:invalid")).startswith("@@")

def workspace_name():
    """Return the name of the workspace."""
    return "_main" if _is_bzlmod_enabled() else "stablehlo"
```

```
diff --git a/stablehlo/tests/BUILD.bazel b/stablehlo/tests/BUILD.bazel
index 8ce10114..d8b3046a 100644
--- a/stablehlo/tests/BUILD.bazel
+++ b/stablehlo/tests/BUILD.bazel
@@ -15,6 +15,7 @@ load("@bazel_skylib//rules:expand_template.bzl", "expand_template")
 load("@llvm-project//llvm:lit_test.bzl", "lit_test", "package_path")
 load("@llvm-project//mlir:tblgen.bzl", "gentbl_cc_library", "td_library")
 load("@rules_cc//cc:defs.bzl", "cc_library")
+load("//stablehlo/tests:utils.bzl", "workspace_name")
 
 package(
     default_visibility = ["//visibility:public"],
@@ -134,7 +135,7 @@ expand_template(
     testonly = True,
     out = "lit.site.cfg.py",
     substitutions = {
-        "\"@STABLEHLO_SOURCE_DIR@\"": "os.path.join(os.environ['TEST_SRCDIR'], '_main')",
+        "\"@STABLEHLO_SOURCE_DIR@\"": "os.path.join(os.environ['TEST_SRCDIR'], '{}')".format(workspace_name()),
         "\"@STABLEHLO_TOOLS_DIR@\"": "os.path.join(os.environ['TEST_SRCDIR'], '_main')",
         "@LIT_SITE_CFG_IN_HEADER@": "# Autogenerated, do not edit.",
         "@LLVM_TOOLS_DIR@": package_path("@llvm-project//llvm:BUILD"),
```

sjain-stanford:
> having to keep our Bazel files sync'd with every downstream seems untenable though; it means if anyone bumps we have to do it in lockstep which doesn't really make sense.

Using `workspace_name` in the paths made it so it worked in both scenarios: 1) stablehlo's standalone builds, and 2) downstream in a monorepo setting where stablehlo is added (as an `http_archive`). I'm just pointing out that the change in this PR would break 2 for all the downstream repos, and your patch on conditionally switching paths depending on `_is_bzlmod_enabled` seems to mitigate that. For context, we used to maintain a patch file for stablehlo integrations, which we could entirely retire after https://github.com/openxla/stablehlo/pull/1810.

stablehlo/tests/BUILD.bazel
sjain-stanford: > having to keep our Bazel files sync'd with every downstream seems untenable though; it means if anyone bumps we have to do it in lockstep which doesn't really make sense.

Using `workspace_name` in the paths made it so it worked in both scenarios: 1) stablehlo's standalone builds, and 2) downstream in a monorepo setting where stablehlo is added (as an `http_archive`). I'm just pointing out that the change in this PR would break 2 for all the downstream repos, and your patch on conditionally switching paths depending on `_is_bzlmod_enabled` seems to mitigate that. For context, we used to maintain a patch file for stablehlo integrations, which we could entirely retire after https://github.com/openxla/stablehlo/pull/1810.

fzakaria: Let me send up a patch PR for this so that we have backwards compatability.

fzakaria: @sjain-stanford I also have this tested and working. I'm not sure which is more idiomatic.

```

# Equivalent of configure_lit_site_cfg from CMakeLists.txt.
expand_template(
    name = "lit_site_cfg_py_gen",
    testonly = True,
    out = "lit.site.cfg.py",
    substitutions = {
        "\"@STABLEHLO_SOURCE_DIR@\"": "RUNFILES_DIR",
        "\"@STABLEHLO_TOOLS_DIR@\"": "RUNFILES_DIR",
        "@LIT_SITE_CFG_IN_HEADER@": '''# Autogenerated, do not edit.
from python.runfiles import Runfiles
from pathlib import Path

r = Runfiles.Create()
LITE_CFG_PY = Path(r.Rlocation("stablehlo/stablehlo/tests/lite.cfg.py"))
RUNFILES_DIR = LITE_CFG_PY.parent.parent.parent.absolute().as_posix()''',
        "@LLVM_TOOLS_DIR@": package_path("@llvm-project//llvm:BUILD"),
    },
    template = "lit.site.cfg.py.in",
)
```

This uses the @python_rules and then walks up the directory from a known path.

sjain-stanford: @fzakaria thanks. If you share a branch, I'd be happy to test it from our monorepo.

PR #2012, opened 2024-02-15
Use bash strict mode in all CI scripts

### LINKED ISSUE #2010: Whitespace Check did not detect - no new line at EOF
### What happened?

Q. I see whilespace-check job is not erroring out for FATAL error 
Q. why "Not a valid object name origin/main" for PR https://github.com/openxla/stablehlo/pull/2007 ?  Dont see this an issue for other recent PRs

```
Run ./build_tools/github_actions/lint_whitespace_checks.sh
fatal: Not a valid object name origin/main
Checking whitespace:
fatal: ambiguous argument '': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
  
fatal: ambiguous argument '': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
tail: cannot open 'bash' for reading: No such file or directory
fatal: ambiguous argument '': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
No whitespace issues found.

```

### Version information

_No response_

mlevesquedion:
"Bash strict mode" is a colloquial expression for a set of settings that
makes Bash stricter (safer). See e.g. https://gist.github.com/robin-a-meade/58d60124b88b60816e8349d1e3938615

TLDR: to avoid errors from being silently ignored, we should set errexit
(exit script on error), nounset (exit script when an undefined variable
is encountered) and pipefail (propagate error codes in pipelines) in all
our scripts.

Unfortunately, `errexit` is in fact not enough to exit when any error
occurs, some errors are swallowed: https://dev.to/banks/stop-ignoring-errors-in-bash-3co5

Because of this, I changed the whitespace script to gather the files in
a different way (I copied the approach from the clang-format script), such
that the script will exit when an error occurs, instead of continuing to run.

The whitespace script was missing a fetch which was causing `git
merge-base` to fail. Maybe our other scripts should also be updated to
only run on changed files.

Interestingly, GitHub already sets `-e` (`errexit`) when running
scripts, but it's best to enforce it in the script itself:

```
Run ./build_tools/github_actions/lint_whitespace_checks.sh
  ./build_tools/github_actions/lint_whitespace_checks.sh
  shell: /usr/bin/bash -e {0}
```

Fixes #2010

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -52,6 +52,7 @@ jobs:
 
       - name: Run Whitespace Checks
         run: |
+          git fetch --no-tags --prune --depth=1 origin "${GITHUB_BASE_REF?}:${GITHUB_BASE_REF?}"
           ./build_tools/github_actions/lint_whitespace_checks.sh
 
   version-checks:
--- a/build_tools/github_actions/ci_build_bazel.sh
+++ b/build_tools/github_actions/ci_build_bazel.sh
@@ -12,6 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+set -o errexit
+set -o nounset
+set -o pipefail
+
 if [[ $# -ne 0 ]] ; then
   echo "Usage: $0"
   exit 1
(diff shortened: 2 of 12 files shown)


mlevesquedion:
Converting to draft until I get a some time to fix the failures.

abhigunj (approved):

PR #2014, opened 2024-02-15
Add support for LLVM toolchain

fzakaria:
We should strive to make our Bazel builds as hermetic as possible. Bazel has support for toolchains
(https://bazel.build/extending/toolchains) which I admit are a bit more advanced than I am familiar with.

Fortunately, there seems to be a community contribution, for a toolchain that is provided for LLVM.

I picked a version that is quite old (11.1) since some of our developers at Google are on 20.04 LTS.

Alternate approach is to use something like [https://nixos.org/](https://nixos.org/) to bring in toolchains.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -18,6 +18,8 @@ build --host_cxxopt=-std=c++17
 # this only works for Clang toolchain AFAIK
 # https://github.com/bazelbuild/bazel/pull/11440
 build --features=layering_check
-# Set the default compiler to the `clang` binary on the `PATH`.
-# TODO(fzakaria): Make this a toolchain or hermetic somehow
-build --repo_env=CC=clang
+
+# Make sure we don't pickup the default CPP toolchains
+build --action_env=BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1
+# Not needed after https://github.com/bazelbuild/bazel/issues/7260 is closed
+build --incompatible_enable_cc_toolchain_resolution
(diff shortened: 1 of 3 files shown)


MODULE.bazel
```diff
@@ -4,6 +4,7 @@
 #
 # For more details, please check https://github.com/bazelbuild/bazel/issues/18958
 ###############################################################################
+bazel_dep(name = "toolchains_llvm", version = "0.10.3")
 bazel_dep(name = "bazel_skylib", version = "1.3.0")
```
sdasgup3, line 8: Should we need to change the bazel_skylib version here to 1.4.2

sdasgup3 (approved):
Thanks for the making the bazel build more predictable.

steeve:
We build stablehlo using https://github.com/uber/hermetic_cc_toolchain with great success

.bazelrc
```diff
@@ -18,6 +18,8 @@ build --host_cxxopt=-std=c++17
...
 # https://github.com/bazelbuild/bazel/pull/11440
 build --features=layering_check
-# Set the default compiler to the `clang` binary on the `PATH`.
-# TODO(fzakaria): Make this a toolchain or hermetic somehow
-build --repo_env=CC=clang
+
+# Make sure we don't pickup the default CPP toolchains
+build --action_env=BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1
+# Not needed after https://github.com/bazelbuild/bazel/issues/7260 is closed
+build --incompatible_enable_cc_toolchain_resolution
```
GleasonK, line 25: newline at eof

fzakaria:
I'm going to pause on this PR -- I should contribute for https://github.com/bazel-contrib/toolchains_llvm/issues/275 so that it's a bit more straightforward for us at Google.

PR #2015, opened 2024-02-15
Align the definition of `checkIfOperandAndResultElementTypesMatch` with its name

sdasgup3:
As pointed by @abhigunj , having this mismatch is confusing.

### DIFF
--- a/stablehlo/dialect/VhloOps.cpp
+++ b/stablehlo/dialect/VhloOps.cpp
@@ -322,16 +322,16 @@ bool checkIfOperandAndResultElementTypesMatch(TypeRange operandTypes,
   if (llvm::any_of(
           llvm::zip(inputElementTypes, resultElementTypes),
           [&](auto pair) { return std::get<0>(pair) != std::get<1>(pair); }))
-    return true;
+    return false;
 
-  return false;
+  return true;
 }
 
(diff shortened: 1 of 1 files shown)


stablehlo/dialect/VhloOps.cpp
```diff
@@ -322,16 +322,16 @@ bool checkIfOperandAndResultElementTypesMatch(TypeRange operandTypes,
   if (llvm::any_of(
           llvm::zip(inputElementTypes, resultElementTypes),
           [&](auto pair) { return std::get<0>(pair) != std::get<1>(pair); }))
-    return true;
+    return false;
 
-  return false;
+  return true;
```
mlevesquedion, line 327: Nit: I think this can just be `return !llvm::any_of(...);`

mlevesquedion (approved):

stablehlo/dialect/VhloOps.cpp
mlevesquedion: Or at that point, might be clearer to do `llvm::all_of(... == ...)`

abhigunj (approved):

stablehlo/dialect/VhloOps.cpp
sdasgup3: done.

PR #2016, opened 2024-02-16
fix copyright message for a few files

sdasgup3:
Also includes framing the version string in ways expected by some internal integration tooling.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,3 +1,17 @@
+# Copyright 2024 The StableHLO Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 ###############################################################################
 # Bazel now uses Bzlmod by default to manage external dependencies.
 # Please consider migrating your external dependencies from WORKSPACE to MODULE.bazel.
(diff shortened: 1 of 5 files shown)


mlevesquedion:
This is unfortunate. `addlicense` doesn't handle these files and it has no extensibility mechanism. Also the project seems unmaintained. Since the functionality it provides is quite simple, I might just go and rewrite this to a script with `find` and `grep`.

mlevesquedion (approved):

MODULE.bazel
```diff
@@ -1,3 +1,17 @@
+# Copyright 2024 The StableHLO Authors. All Rights Reserved.
```
fzakaria, line 1: why did this not fail on the submission?

MODULE.bazel.lock
```diff
@@ -1,3 +1,17 @@
+# Copyright 2024 The StableHLO Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
```
fzakaria, line 3: This is a generated file.

MODULE.bazel
mlevesquedion: Are you wondering about the license check we have for CI? `addlicense` doesn't support this: https://github.com/openxla/stablehlo/pull/1929

We should probably just drop `addlicense` and replace it with `find` + `grep`. I plan to send a PR for that later today.

sdasgup3: make sense Or can we add the *.bazel file extension to `addlicense `. With that any newly added bazel might should have the copyright ( note that  all the current bazel files in stablehlo/* now have the header)

MODULE.bazel.lock
sdasgup3: I see. I will remove this change. The downstream integration tool should skip any copyright check on files of this ext.

fzakaria (approved):

MODULE.bazel
mlevesquedion: > Add the .bazel file extension to addlicense

I tried: https://github.com/google/addlicense/pull/154

> Why don't we fork addlicense

It would be easy to fork and patch addlicense, but I doubt we want the maintenance burden that could come with that. Let me look into a `find` + `grep` based license check, IMO that will be much less work now and in the future.

PR #2017, opened 2024-02-16
Fix issues that were missed due to CI script issues

mlevesquedion:
I'm surprised there weren't any whitespace check misses, and I'm confused as to why there were clang-format issues, given that the check ran successfully when the code was introduced: https://github.com/openxla/stablehlo/actions/runs/7809371206/job/21301436418

Hopefully with #2012 script failures should not go unnoticed.


mlevesquedion:
Hmm. I guess the `clang-format` I have locally is different from the one that we use in CI.

PR #2020, opened 2024-02-20
Remove extra % prefix for function inputs

mlevesquedion:
The way the spec is currently written, a function input would be written e.g. `%%arg0` instead of `%arg0`.


ghpvnist (approved):
Nice catch!

PR #2024, opened 2024-02-21
Implement composite op

mlevesquedion:
This introduces the op definition, spec, verifier, compatibility, and interpreter support.

The spec has few constraints. Most are covered by ODS. Others are covered by the verifier, and tests were added for each.


docs/spec.md
```diff
@@ -1922,6 +1922,54 @@ imaginary values, `lhs` and `rhs`, and produces a `result` tensor.
...
+
+#### Outputs
+
+| Name      | Type                      |
+|-----------|---------------------------|
+| `results` | variadic number of values |
+
+#### Constraints
+
+* (C1) `is_defined_in_parent_module(decomposition)`.
```
sdasgup3, line 1956: The spec has not defined the concept of module. It is more related to how the functions in a program are organised.  Can we instead say that `decomposition` is defined in the same scope as the parent function of the op associated with decomposition.

mlevesquedion: Thanks, good point! Done.

stablehlo/transforms/VhloLegalizeToStablehlo.cpp
```diff
@@ -500,6 +500,12 @@ SpecialResult convertSpecial(const OpConversionPattern<VhloOpTy>& pattern,
       if (!stablehloAttr) return specialFailure();
     }
   }
+  if constexpr (std::is_same<VhloOpTy, vhlo::CompositeOpV1>::value) {
+    if (vhloName == "decomposition") {
+      stablehloAttr = convertSymbol(vhloAttr, pattern.getTypeConverter());
```
GleasonK, line 505: Per my comments on the spec earlier, this is the result of the hackery around function refs being stored as strings. I don't love that we have to do this..

GleasonK: To be addressed separately though

GleasonK (changes requested):

docs/spec.md
```diff
@@ -1923,6 +1923,55 @@ imaginary values, `lhs` and `rhs`, and produces a `result` tensor.
...
+
+#### Outputs
+
+| Name      | Type                      |
+|-----------|---------------------------|
+| `results` | variadic number of values |
+
+#### Constraints
+
+* (C1) `re.fullmatch(r'[a-zA-Z]+(\.[a-zA-Z][a-zA-Z0-9_]*)+', name)`
```
GleasonK, line 1957: Let's abstract this to a helper method `is_namespaced_op_name(name)`. Let's also support underscore after dot. Some existing op names in the TF world that should be considered valid:
- `tf_executor._SwitchN
- `tf._XlaHostComputeMlir`

I can also see a use case for `dialect_v1.my_op`.

How about:
- Starts with letter
- [letter, number, underscore]*
- (`.` [letter, number, underscore, `$`]+ )+

`$` is used in some custom calls as target name, so I'm hesitant to suggest changes there, in case we want to support cc target + decomposition

MODULE.bazel.lock
```diff
@@ -1,6 +1,6 @@
 {
   "lockFileVersion": 3,
-  "moduleFileHash": "836f0a7d2276ed93403f104a10008b94ec7e7f81b8d6921cea287f0a6d364efa",
+  "moduleFileHash": "9e42fa15fdae2a59a51ea764fd71c5e9f785b0fcb894b4e6007c56956b0082af",
```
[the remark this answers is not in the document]
mlevesquedion: This happened automatically, not sure why. @fzakaria any idea why this might have happened? Could it be because I changed `BUILD.bazel` ?

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -208,6 +209,56 @@ LogicalResult AfterAllOp::inferReturnTypes(
                               inferredReturnTypes);
 }
 
+//===----------------------------------------------------------------------===//
+// CompositeOp
+//===----------------------------------------------------------------------===//
+
+LogicalResult CompositeOp::verify() {
```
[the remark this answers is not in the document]
mlevesquedion: Done.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -208,6 +209,56 @@ LogicalResult AfterAllOp::inferReturnTypes(
...
+  // composite_c1
+  mlir::ModuleOp module = getOperation()->getParentOfType<ModuleOp>();
```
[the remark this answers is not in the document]
mlevesquedion: Unfortunately, as far as I can tell, we don't. Indeed e.g. custom_call doesn't do any such verification, and it isn't done by MLIR either.

stablehlo/transforms/VhloLegalizeToStablehlo.cpp
mlevesquedion: Ack, thanks for sharing your thought process on this!

mlevesquedion (commented):
Thanks for the review, Kevin!

MODULE.bazel.lock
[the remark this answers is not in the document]
fzakaria: > The moduleFileHash represents the hash of the MODULE.bazel file contents. If any changes occur in this file, the hash value differs.

I think it might have been someone else change they didn't commit.
I think from the license change to MODULE.bazel -- you have to run `bazel build` to generate the lock file.

docs/spec.md
```diff
@@ -1923,6 +1923,54 @@ imaginary values, `lhs` and `rhs`, and produces a `result` tensor.
...
+### composite
+
+#### Semantics
+
+Encapsulates an operation made up (composed) of other StableHLO operations,
+taking `inputs` and producing `results`. The decomposition is specified
+using the `decomposition` attribute. The composite op can be replaced with
+its decomposition without changing semantics.
+
+The `version` and `decomposition_attributes` enable additional information
```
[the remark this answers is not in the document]
GleasonK: Ah. I see. Hmm.. `op_attributes` may be more accurate than `decomposition_attributes`. Let's discuss offline a bit

docs/spec.md
```diff
@@ -1923,6 +1923,54 @@ imaginary values, `lhs` and `rhs`, and produces a `result` tensor.
...
+and `decomposition_attributes` defaults to an empty dictionary.
+
+#### Inputs
+
+| Label | Name                                   | Type                                          |
+|-------|----------------------------------------|-----------------------------------------------|
+| (I1)  | `inputs`                               | variadic number of values                     |
+| (I2)  | `name`                                 | constant of type `string`                     |
+| (I3)  | `decomposition`                        | constant of type `function`                   |
+| (I4)  | `version`                              | constant of type `si32`                       |
```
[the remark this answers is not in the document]
GleasonK: Nope, I left this comment before I made it to the `StablehloOps.td` file. Looks like you did make the default 0

MODULE.bazel.lock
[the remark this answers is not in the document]
GleasonK: Should this have failed bazel CI?

docs/spec.md
```diff
@@ -6485,6 +6533,12 @@ function returns `true`. If `x` is not a tensor, returns `None`.
 tensors and returns `num_results` slices of `x` along the axis `axis`.
 If `x` is not a tensor or `dim(x, axis) % num_results != 0`, returns `None`.
 
+#### Functions on non-values
+
+* `is_defined_in_parent_scope(x: Function) -> Value` is defined on functions
```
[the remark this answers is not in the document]
GleasonK: `string -> Value`? I agree that we should resolve the `Value` issue separately, and resolve the `string` vs `FunctionRef` issue separately.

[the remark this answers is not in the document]
GleasonK: Also for history - This lack of string vs function distinction has bit us before when using discardable attributes. We only support `@func` in custom_call and Func, anywhere else will fail to serialize, and it isn't trivial to add because it is ambiguous with any string.

GleasonK (commented):
LGTM

stablehlo/dialect/StablehloOps.td
```diff
@@ -2084,6 +2084,44 @@ def StableHLO_CollectivePermuteOp: StableHLO_Op<"collective_permute",
...
+    %results = stablehlo.composite "my.op" %arg0, %arg1 {
+      decomposition = @my_op,
+      decomposition_attributes = { my_attribute = "my_value" },
+      version = 1 : i32
+    } : (tensor<f32>, tensor<f32>) -> tensor<f32>
+    ```
+  }];
+
+  let arguments = (ins
+    Variadic<HLO_TensorOrTokenOrTuple>:$inputs, /*composite_i1*/
```
abhigunj, line 2113:
```suggestion
    Variadic<HLO_TensorOrPerAxisQuantizedTensorOrTokenOrTuple>:$inputs, /*composite_i1*/
```

stablehlo/dialect/StablehloOps.td
```diff
@@ -2084,6 +2084,44 @@ def StableHLO_CollectivePermuteOp: StableHLO_Op<"collective_permute",
...
+    StrAttr:$name, /*composite_i2*/
+    FlatSymbolRefAttr:$decomposition, /*composite_i3*/
+    DefaultValuedOptionalAttr<I32Attr, "0">:$version, /*composite_i4*/
+    DefaultValuedOptionalAttr<DictionaryAttr, "{}">:$decomposition_attributes /*composite_i5*/
+  );
+  let results = (outs Variadic<HLO_TensorOrTokenOrTuple>);
```
abhigunj, line 2119:
```suggestion
  let results = (outs Variadic<HLO_TensorOrPerAxisQuantizedTensorOrTokenOrTuple>);
```

MODULE.bazel.lock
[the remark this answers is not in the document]
mlevesquedion: Sent #2060 to address this.

docs/spec.md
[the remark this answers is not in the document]
mlevesquedion: Summary of offline discussion:

`stablehlo.composite` models an op, so the attributes are attributes of the op being modeled. Unfortunately we can't just use `attributes` as described above, so we went with `composite_attributes`.

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -5395,3 +5395,162 @@ func.func @dynamic_iota_output_shape_mismatching_size() -> tensor<4xf32> {
...
+  } : (tensor<f32>, tensor<f32>) -> tensor<f32>
+  func.return
+}
+
+// -----
+
+func.func @foo() { func.return }
+func.func @composite_c1() {
+  // @expected-error@+1 {{name "foo" does not match pattern}}
+  stablehlo.composite "foo" { decomposition = @foo } : () -> ()
```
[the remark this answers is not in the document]
mlevesquedion: I like the idea of referencing the spec function and giving an example. IMO giving the full regex pattern is also necessary though. Done.

stablehlo/dialect/StablehloOps.td
mlevesquedion: Done, thanks for pointing that out!

mlevesquedion:
FYI had to rebase to handle the bazel lockfile change: go/shlopr/2060

docs/spec.md
```diff
@@ -1923,6 +1923,55 @@ imaginary values, `lhs` and `rhs`, and produces a `result` tensor.
...
 &nbsp;[More Examples](../stablehlo/tests/interpret/complex.mlir)
 
+### composite
+
+#### Semantics
+
+Encapsulates an operation made up (composed) of other StableHLO operations,
+taking `inputs` and producing `results`. The decomposition is specified
+using the `decomposition` attribute. The composite op can be replaced with
+its decomposition without changing semantics.
```
fzakaria, line 1933: How does one test this in practicality?
You'd always want the inlined version to test against?

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3366,6 +3368,80 @@ LogicalResult verifyCollectivePermuteOp(
...
+      !llvm::all_of(
+          llvm::drop_begin(parts[0]),
+          [](char c) { return isalpha(c) || isdigit(c) || c == '_'; }) ||
+      // Subsequent parts must be all alpha, digit, underscore or dollar sign.
+      !llvm::all_of(llvm::drop_begin(parts), [](llvm::StringRef part) {
+        return llvm::all_of(part, [](char c) {
+          return isalpha(c) || isdigit(c) || c == '_' || c == '$';
+        });
+      })) {
+    return emitOptionalError(loc,
```
fzakaria, line 3392: personal preference: break these out into multiple conditions and you can give an error message for the more meaningful failure.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3366,6 +3368,80 @@ LogicalResult verifyCollectivePermuteOp(
   return success();
 }
 
+LogicalResult verifyCompositeOp(std::optional<Location> loc, Operation* op,
+                                StringRef name, StringRef decomposition) {
+  // composite_c1
+  llvm::SmallVector<llvm::StringRef> parts;
+  name.split(parts, '.');
```
fzakaria, line 3375: you could probably use the split that returns a pair since all you want is at most 1 split.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3366,6 +3368,80 @@ LogicalResult verifyCollectivePermuteOp(
...
+                             "must match the following regular expression: ",
+                             "[a-zA-Z][a-zA-Z0-9_]*([.][a-zA-Z0-9_$]+)+, e.g. "
+                             "\"my_namespace.my_op\"");
+  }
+
+  // composite_c2
+  auto module = op->getParentOfType<mlir::ModuleOp>();
+  auto decomp = module.lookupSymbol<mlir::func::FuncOp>(decomposition);
+  if (!decomp) {
+    return emitOptionalError(loc,
```
fzakaria, line 3403: edification: why is loc not from the op itself but comes in from the verify?

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3366,6 +3368,80 @@ LogicalResult verifyCollectivePermuteOp(
   return success();
 }
 
+LogicalResult verifyCompositeOp(std::optional<Location> loc, Operation* op,
+                                StringRef name, StringRef decomposition) {
+  // composite_c1
+  llvm::SmallVector<llvm::StringRef> parts;
```
dansuh17, line 3374: nit: Strip the `llvm::` prefixes.

docs/spec.md
mlevesquedion: I don't think this necessarily needs to be tested, although maybe once we have the inlining pass we can add a test that inlines and then runs the interpreter to validate this.

stablehlo/dialect/TypeInference.cpp
mlevesquedion: I tried breaking it out and I think the above is still preferable. I agree that by breaking it out we can give more specific error messages (e.g. "expected at least 2 non-empty parts separated by periods, but got: ..."), but IMO giving the full expected pattern in the error message makes it easiest to know what needs to be done to get to a valid name. In general I'm not a huge fan of massive conditions like this one but I think this one reads fine, it reads kind of like a regular expression thanks to the LLVM helpers.

stablehlo/dialect/TypeInference.cpp
mlevesquedion: We actually support more than 1 split though.

stablehlo/dialect/TypeInference.cpp
mlevesquedion: It actually does come from the op: 

```
LogicalResult CompositeOp::verify() {
  return hlo::verifyCompositeOp(getLoc(), getOperation(), getName(),
                                getDecomposition());
}
```

docs/spec.md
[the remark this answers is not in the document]
GleasonK: Not too late, and probably not a bad idea to do before v1.0. Just need to manage forward/backward compat. Shouldn't be too difficult

stablehlo/dialect/StablehloOps.td
```diff
@@ -2084,6 +2084,44 @@ def StableHLO_CollectivePermuteOp: StableHLO_Op<"collective_permute",
       "::mlir::DenseIntElementsAttr":$source_target_pairs)>];
 }
 
+def StableHLO_CompositeOp : StableHLO_Op<"composite", []> {
```
GleasonK, line 2087: How about implementing `SymbolUserOpInterface`:

```
def CallOp : Func_Op<"call",
    [CallOpInterface, MemRefsNormalizable,
     DeclareOpInterfaceMethods<SymbolUserOpInterface>]>
```

Makes the op have a method called:

```
LogicalResult CompositeOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
   hlo::verifyCompositeOp(<other_fields>, symbolTable);
}
```

We can do all verification in this op, including sym validation.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3366,6 +3368,80 @@ LogicalResult verifyCollectivePermuteOp(
...
+    return emitOptionalError(loc,
+                             "name must be a valid namespaced op name, i.e. it",
+                             "must match the following regular expression: ",
+                             "[a-zA-Z][a-zA-Z0-9_]*([.][a-zA-Z0-9_$]+)+, e.g. "
+                             "\"my_namespace.my_op\"");
+  }
+
+  // composite_c2
+  auto module = op->getParentOfType<mlir::ModuleOp>();
+  auto decomp = module.lookupSymbol<mlir::func::FuncOp>(decomposition);
```
GleasonK, line 3401: Let's match the CallOp impl:

```
  FuncOp fn = symbolTable.lookupNearestSymbolFrom<FuncOp>(*this, fnAttr);
  if (!fn)
    return emitOpError() <<  ...
```

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3366,6 +3368,80 @@ LogicalResult verifyCollectivePermuteOp(
...
 }
 
+LogicalResult verifyCompositeOp(std::optional<Location> loc, Operation* op,
+                                StringRef name, StringRef decomposition) {
+  // composite_c1
+  SmallVector<StringRef> parts;
+  name.split(parts, '.');
+  if (
+      // Name must have at least 2 non-empty parts separated by periods.
+      parts.size() < 2 ||
```
GleasonK, line 3378: Let's make this its own function: `LogicalResult verifyNamespacedOpName(...)`.
That way we can have a nice code comment for what a valid name is here as well.

Alternative: Any reason not to use the regex to test this as opposed to hand-parsing?

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3366,6 +3368,80 @@ LogicalResult verifyCollectivePermuteOp(
...
+  auto decomp = module.lookupSymbol<mlir::func::FuncOp>(decomposition);
+  if (!decomp) {
+    return emitOptionalError(loc,
+                             "decomposition not found in current module: \"",
+                             decomposition, "\"");
+  }
+
+  auto decompFunType = decomp.getFunctionType();
+
+  // composite_c3
```
GleasonK, line 3410: Repeating code in c3 / c4. We have a bad pattern in this file of not using helper functions.. not sure if that was by design. I suppose this duplication is OK given that..

stablehlo/reference/Ops.cpp
```diff
@@ -384,6 +384,16 @@ SmallVector<InterpreterValue> eval(Region &region,
       auto rhs = scope.findTensor(complexOp.getRhs());
       auto result = evalComplexOp(lhs, rhs, complexOp.getType());
       scope.add(complexOp.getResult(), result);
+    } else if (auto compositeOp = dyn_cast<CompositeOp>(op)) {
+      auto operands = scope.findTensors(compositeOp.getOperands());
+      SmallVector<InterpreterValue> values(operands.size());
+      std::transform(operands.begin(), operands.end(), values.begin(),
+                     [](const Tensor &t) { return InterpreterValue(t); });
```
GleasonK, line 391: More common pattern in our codebase is:

```
llvm::to_vector(llvm::map_range(operands, [](const Tensor &t) { return InterpreterValue(t); }))
```

GleasonK: Might not need to_vector given that eval wants an arrayref? Not sure how that map_range result can be used.

stablehlo/reference/Ops.cpp
```diff
@@ -384,6 +384,16 @@ SmallVector<InterpreterValue> eval(Region &region,
...
+      auto func =
+          llvm::cast<func::FuncOp>(op.getParentOfType<ModuleOp>().lookupSymbol(
```
GleasonK, line 393: Let's use that "nearestSymbol" API here as well

stablehlo/reference/Ops.cpp
```diff
@@ -384,6 +384,16 @@ SmallVector<InterpreterValue> eval(Region &region,
...
+              compositeOp.getDecomposition()));
+      auto results = eval(func.getBody(), values, fallback, process, parent);
```
GleasonK, line 395: Not sure if we want to pass `parent` or `nullptr` for scope? I believe this would mean that the decomposition could refer to variables in the parent scope..which should be invalid.

Consider adding `evalCallOp` to do the "Create empty scope, lookup func, add results to parent scope", and just call it here. This would also be adding the ability for the interpreter to make calls, which is a nice feature that you've effectively implemented here.

stablehlo/tests/interpret/composite.mlir
```diff
@@ -0,0 +1,16 @@
+// RUN: stablehlo-translate --interpret -split-input-file %s
+
+func.func @add(%arg0: tensor<i64>, %arg1: tensor<i64>) -> tensor<i64> {
+  %0 = stablehlo.add %arg0, %arg1 : tensor<i64>
+  func.return %0 : tensor<i64>
+}
+
+func.func @main() {
```
GleasonK, line 8: For the sake of using attrs / having a more "custom" composite, lets make this an AddNOp:

```
func.func @add_n.impl(%arg0: tensor<i64>) -> tensor<i64> {
  %0 = stablehlo.constant dense<2> : tensor<i64>
  %1 = stablehlo.add %arg0, %0 : tensor<i64>
  func.return %1 : tensor<i64>
}

func.func @main() {
  %0 = stablehlo.constant dense<1> : tensor<i64>
  %1 = stablehlo.composite "stablehlo.add_n" %0 {
    composite_attributes = { n = 2 : i64 },
    decomposition = @add_n.impl
  } : (tensor<i64>, tensor<i64>) -> tensor<i64>
  check.expect_eq_const %1, dense<3> : tensor<i64>
  func.return
}
```

stablehlo/dialect/StablehloOps.td
```diff
@@ -2084,6 +2084,44 @@ def StableHLO_CollectivePermuteOp: StableHLO_Op<"collective_permute",
...
+    } : (tensor<f32>, tensor<f32>) -> tensor<f32>
+    ```
+  }];
+
+  let arguments = (ins
+    Variadic<HLO_TensorOrPerAxisQuantizedTensorOrTokenOrTuple>:$inputs, /*composite_i1*/
+    StrAttr:$name, /*composite_i2*/
+    FlatSymbolRefAttr:$decomposition, /*composite_i3*/
+    DefaultValuedOptionalAttr<I32Attr, "0">:$version, /*composite_i4*/
+    DefaultValuedOptionalAttr<DictionaryAttr, "{}">:$composite_attributes /*composite_i5*/
```
GleasonK, line 2117: For reference - I don't think any special default value handling would be needed in VHLO conversions if these were just `DefaultValuedAttr` instead of `DefaultValuedOptionalAttr`. I go back and forth on if I like these. They simplify the pretty-print, may have some memory savings, but otherwise don't provide too much value.

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -5395,3 +5395,162 @@ func.func @dynamic_iota_output_shape_mismatching_size() -> tensor<4xf32> {
...
+func.func @composite_c1() {
+  // valid name
+  stablehlo.composite "f00._.$" { decomposition = @foo } : () -> ()
+  func.return
+}
+
+// -----
+
+func.func @composite_c2(%arg0: tensor<f32>) {
+  // @expected-error@+1 {{decomposition not found in current module: "nonexistent"}}
```
GleasonK, line 5492: The CallOp error is:

```
  if (!fn)
    return emitOpError() << "'" << fnAttr.getValue()
                         << "' does not reference a valid function";
```

Consider making our error closer to this message since people may be used to dealing with error messages like this. "reference a valid function" is pretty descriptive.

GleasonK (changes requested):
This should be final round of review, likely only nits to follow

docs/spec.md
```diff
@@ -1923,6 +1923,55 @@ imaginary values, `lhs` and `rhs`, and produces a `result` tensor.
 
 &nbsp;[More Examples](../stablehlo/tests/interpret/complex.mlir)
 
+### composite
+
+#### Semantics
+
+Encapsulates an operation made up (composed) of other StableHLO operations,
+taking `inputs` and producing `results`. The decomposition is specified
```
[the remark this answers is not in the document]
GleasonK: (Also update ODS summary with whatever wording we land on)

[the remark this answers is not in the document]
mlevesquedion: This wording SGTM. Done.

stablehlo/dialect/StablehloOps.td
mlevesquedion: Good callout! I did think to use `SymbolUserOpInterface`, but I'm worried it might have unintended side effects, e.g. it seems it will prevent running the `RemoveDeadValues` pass, unless we also implement `CallOpInterface`: https://github.com/llvm/llvm-project/blob/f51ade25b9205efee09a4915031848cebe772805/mlir/lib/Transforms/RemoveDeadValues.cpp#L581

`composite` isn't actually a call, so I don't think that makes sense, but I don't know if we care about the dead values thing.

I also don't think we gain a whole lot by implementing this interface. If you feel strongly about this I'm happy to do it, otherwise maybe this can be left as something to investigate in the future?

stablehlo/dialect/TypeInference.cpp
mlevesquedion: Unfortunately, this requires implementing `SymbolUserOpInterface`, which I'm not sure we should do. Let's merge the discussion into that thread: https://github.com/openxla/stablehlo/pull/2024#discussion_r1513563283

stablehlo/dialect/TypeInference.cpp
mlevesquedion: I considered extracting a helper here but because of the way the error message is constructed the helper turned out to be a bit awkward. IMO this duplication is acceptable.

stablehlo/reference/Ops.cpp
mlevesquedion: Very nice, thanks! I couldn't find an easy way to use `ArrayRef` directly here, so used `to_vector`.

stablehlo/reference/Ops.cpp
mlevesquedion: Unfortunately, this requires implementing SymbolUserOpInterface, which I'm not sure we should do. Let's merge the discussion into that thread: https://github.com/openxla/stablehlo/pull/2024#discussion_r1513563283

stablehlo/reference/Ops.cpp
mlevesquedion: > scope

Good point! Done.

> `evalCallOp`

I want to keep this PR focused on just implementing the composite op. I agree this wold be a nice feature to add, but IMO that should go into its own PR.

docs/spec.md
```diff
@@ -1923,6 +1923,55 @@ imaginary values, `lhs` and `rhs`, and produces a `result` tensor.
 
 &nbsp;[More Examples](../stablehlo/tests/interpret/complex.mlir)
 
+### composite
```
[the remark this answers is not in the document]
mlevesquedion: Good point. Created an issue: https://github.com/openxla/stablehlo/issues/2070

stablehlo/dialect/StablehloOps.td
mlevesquedion: Good to know, thanks! I find it kinda weird that both exist TBH. IMO if something has a default value, it is also optional, but I guess "optional" here means something different.

stablehlo/tests/ops_stablehlo.mlir
mlevesquedion: I agree that's a better message. Thanks, done.

mlevesquedion:
> This should be final round of review, likely only nits to follow

Thank you for your comments! I believe I've addressed everything, except this `SymbolUserOpInterface` thing. PTAL: https://github.com/openxla/stablehlo/pull/2024#discussion_r1513563283

stablehlo/dialect/StablehloOps.td
GleasonK: Do we use this pass? There was one dead-code-removal pass that didn't work for us for various reasons as-is. Something that wasn't CSE or Canonicalization-related cleanups.

That said, no strong opinion. It buys us a guarantee that the properly scoped symbol table will be passed to us even if upstream MLIR changes anything. That said if it ever became different from the sym table of the module there would probably be a lot of cleanup to do.

mlevesquedion: We don't, but I wonder if downstream users might.

In any case, I think you make a good point. And it's probably best to signal that we use a symbol. Maybe we should do this for `custom_call` too.

I'll make the change tomorrow.

docs/status.md
```diff
@@ -65,6 +65,7 @@ one of the following tracking labels.
 | collective_permute       | yes           | revisit      | yes            | no              | yes         |
 | compare                  | yes           | yes          | yes            | yes             | yes         |
 | complex                  | yes           | yes          | yes            | yes             | yes         |
+| composite                | yes           | yes          | no             | yes             | yes         |
```
GleasonK, line 68: no -> infeasible ? Like custom call

GleasonK (approved):

PR #2025, opened 2024-02-21
Exclude MLIR bytecode files from license check

mlevesquedion:
Also, don't add an extra line between each skip, and use the same message for each type of skipped file (easier to read).

I tested the script locally by creating files with various suffixes and confirmed that the skipped suffixes are indeed skipped (and files with a non-skipped suffix are not skipped).


sdasgup3 (approved):

abhigunj (approved):

ghpvnist:
We should also exclude `build_tools/llvm_version.txt` from the list: https://github.com/openxla/stablehlo/actions/runs/7996435336/job/21838895350?pr=2029

PR #2030, opened 2024-02-21
Use initialize and generate getDependentDialects for passes from IREE

mlevesquedion:
Extending work done in #1919 and #1966.


GleasonK:
Can you rebase, I think CHLO->StableHLO now uses the declarative dependencies

GleasonK (approved):

PR #2029, opened 2024-02-21
Integrate LLVM at llvm/llvm-project@4c6043de0b83

ghpvnist:

PR #2031, opened 2024-02-21
Ignore all files that shouldn't have licenses

mlevesquedion:
Unblocks #2029

I ran the license checker with `find` as input instead of `git diff` to find all the relevant files and exempted the files that don't already have licenses.


ghpvnist (approved):
Thanks for the quick fix!

mlevesquedion:
Thanks for the review! I'll keep an eye on CI but if everything is green feel free to merge this.

PR #2032, opened 2024-02-21
Print clang-format version in CI

mlevesquedion:
The version of clang-format available in CI seems to be different from the one on my machine, but I can't easily see what the version of clang-format running in CI is, so I can't really debug. Print the version so that moving forward if someone runs afoul of the check they can repro and fix.

### DIFF
--- a/build_tools/github_actions/lint_clang_format.sh
+++ b/build_tools/github_actions/lint_clang_format.sh
@@ -39,6 +39,8 @@ if [[ $# -ne 0 ]] ; then
   exit 1
 fi
 
+clang-format --version
+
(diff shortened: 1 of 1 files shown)


ghpvnist:
I believe you can run the github runners in your own fork and debug that way.

mlevesquedion:
> I believe you can run the github runners in your own fork and debug that way.

Yes, but if I send a PR and the check fails, it's onerous to have to re-run the check in my fork. Having the version printed in the check itself means if someone runs into a failure, they know right away what version of `clang-format` they need to use.

ghpvnist (approved):

PR #2033, opened 2024-02-22
Remove suffixes for files that already have a license

ghpvnist:


sdasgup3:
FYI: we are skipping copyright checks from the internal integration tool.

MODULE.bazel.lock
mlevesquedion, line 1: I'm not sure it makes sense to have a license here because this file is generated: https://bazel.build/external/lockfile#lockfile-generation

I think if we make a change that causes bazel to re-generate this file the license will be lost and we will need to add it manually again.

IMO it's okay to not have a license in this file.

mlevesquedion (approved):
LGTM, but not sure about having a license in MODULE.bazel.block since that file is generated.

PR #2036, opened 2024-02-23
Simplify creation of `TensorType`  variants using `TensorType::clone`

sdasgup3:
Let's use `TensorType::clone` which is customized to create TensorType variants (`RankedTensorType` and `UnRankedTensorType`).

### DIFF
--- a/stablehlo/dialect/Base.cpp
+++ b/stablehlo/dialect/Base.cpp
@@ -188,11 +188,7 @@ LogicalResult deriveShapeFromOperand(
 }
 
 ShapedType getSameShapeTensorType(ShapedType shapedType, Type elementType) {
-  if (auto rankedTensorTy = shapedType.dyn_cast<RankedTensorType>())
-    return RankedTensorType::get(rankedTensorTy.getShape(), elementType,
-                                 rankedTensorTy.getEncoding());
-  if (auto unrankedTensorTy = shapedType.dyn_cast<UnrankedTensorType>())
-    return UnrankedTensorType::get(elementType);
+  if (shapedType.isa<TensorType>()) return shapedType.clone(elementType);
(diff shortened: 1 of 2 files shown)


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -857,12 +857,7 @@ LogicalResult ConvolutionOp::verify() {
 
 void ConvertOp::build(OpBuilder& builder, OperationState& result, Value operand,
                       Type resultElementTy) {
-  Type resultTy;
-  Type operandTy = operand.getType();
-  if (auto rankedTy = operandTy.dyn_cast<RankedTensorType>())
-    resultTy = RankedTensorType::get(rankedTy.getShape(), resultElementTy);
-  else
-    resultTy = UnrankedTensorType::get(resultElementTy);
+  auto resultTy = operand.getType().cast<TensorType>().clone(resultElementTy);
```
mlevesquedion, line 860: Would you mind holding off on this change? I'm actually about to change this code to remove unranked dynamism support.

mlevesquedion (approved):
Neat! I didn't know about `TensorType.clone`. Thanks for cleaning up the code.

stablehlo/dialect/StablehloOps.cpp
sdasgup3: Sure. Let me know once your changes are done.

mlevesquedion: Should be done soon, but I can't provide a reliable estimate (ideally today, otherwise should be next week). WDYT about undoing these changes and committing the rest? I don't want to block you.

sdasgup3: make sense. updone!

PR #2040, opened 2024-02-23
Add StableHLO dialect version for bytecode

jpienaar:
This enables (at least) invalidating previously generated StableHLO bytecode serialization (not using VHLO) by bumping the version. The version is optionally set and if not set, not serialized - this would result in no changes to what is serialized by default today unless explicitly set.

This would enable a better error message than the existing where a loaded bytecode would fail and knowledge required as to why:

Before 

```
<unknown>:0: error: 'stablehlo.broadcast_in_dim' op attribute 'broadcast_dimensions' failed to satisfy constraint: either a DenseI64ArrayAttr or a 1-dimensional I64ElementsAttr.
<unknown>:0: note: see current operation: %4 = "stablehlo.broadcast_in_dim"(%1) {broadcast_dimensions = array<i64>} : (tensor<f32>) -> tensor<500xf32>
<unknown>:0: note: in bytecode version 6 produced by: MLIR19.0.0git
```

Approximately after:

```
reading newer dialect version than supported
```

(Note this could also be refined, but at least point to version skew vs folks being concerned about generation error)

This does not represent a policy change nor an automatic test added.

Could also add to CHLO but wanted to discuss first.

### DIFF
--- a/stablehlo/dialect/StablehloBytecode.cpp
+++ b/stablehlo/dialect/StablehloBytecode.cpp
@@ -706,8 +706,7 @@ void StablehloBytecodeInterface::write(TokenType type,
 std::unique_ptr<DialectVersion> StablehloBytecodeInterface::readVersion(
     DialectBytecodeReader &reader) const {
   int64_t major, minor, patch;
-  if (failed(reader.readVarInt(major)) ||
-      failed(reader.readVarInt(minor)) ||
+  if (failed(reader.readVarInt(major)) || failed(reader.readVarInt(minor)) ||
       failed(reader.readVarInt(patch)))
     return nullptr;
 
--- a/stablehlo/dialect/StablehloBytecode.cpp
+++ b/stablehlo/dialect/StablehloBytecode.cpp
@@ -705,27 +705,26 @@ void StablehloBytecodeInterface::write(TokenType type,
 
 std::unique_ptr<DialectVersion> StablehloBytecodeInterface::readVersion(
     DialectBytecodeReader &reader) const {
-  int64_t major, minor, patch;
+  uint64_t major, minor, patch;
   if (failed(reader.readVarInt(major)) || failed(reader.readVarInt(minor)) ||
       failed(reader.readVarInt(patch)))
     return nullptr;
 
-  Version version(/*major=*/major, /*minor=*/minor, /*patch=*/patch);
-  if (Version::getCurrentVersion() < version) {
+  auto version = std::make_unique<StablehloDialectVersion>(
+      /*major=*/major, /*minor=*/minor, /*patch=*/patch);
+  if (version && StablehloDialectVersion::getCurrentVersion() < *version) {
     return reader.emitError("reading newer dialect than supported"), nullptr;
   }
 
-  auto version = std::make_unique<StablehloDialectVersion>();
-  version->dialectVersion = version;
   return version;
 }
 
 void StablehloBytecodeInterface::writeVersion(
     DialectBytecodeWriter &writer) const {
   if (auto version = cast<StablehloDialect>(getDialect())->getVersion()) {
-    writer.writeVarInt(version.getMajor());
-    writer.writeVarInt(version.getMinor());
-    writer.writeVarInt(version.getPatch());
+    writer.writeVarInt(static_cast<uint64_t>(version->getMajor()));
+    writer.writeVarInt(static_cast<uint64_t>(version->getMinor()));
+    writer.writeVarInt(static_cast<uint64_t>(version->getPatch()));
   }
 }
 
--- a/stablehlo/dialect/StablehloBytecode.cpp
+++ b/stablehlo/dialect/StablehloBytecode.cpp
@@ -713,7 +713,7 @@ std::unique_ptr<DialectVersion> StablehloBytecodeInterface::readVersion(
   auto version = std::make_unique<StablehloDialectVersion>(
       /*major=*/major, /*minor=*/minor, /*patch=*/patch);
   if (version && StablehloDialectVersion::getCurrentVersion() < *version) {
-    return reader.emitError("reading newer dialect than supported"), nullptr;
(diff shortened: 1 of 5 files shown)


jpienaar:
Ping?

stablehlo/dialect/StablehloOps.h
```diff
@@ -52,6 +52,18 @@ limitations under the License.
 namespace mlir {
 namespace stablehlo {
 
+struct StablehloDialectVersion : public mlir::DialectVersion {
+  StablehloDialectVersion() = default;
+  StablehloDialectVersion(uint32_t dialectVersion)
+      : dialectVersion(dialectVersion){};
+
+  // The current dialect version.
+  static constexpr uint32_t kCurrentDialectVersion = 1;
```
fzakaria, line 61: how does this version differ than the other version in VHLO ?

stablehlo/dialect/StablehloBytecode.cpp
```diff
@@ -262,6 +262,14 @@ class StablehloBytecodeInterface : public BytecodeDialectInterface {
   // TO ADD TYPE: Include a write method for each type in StableHLO
   // Ex: void write(SomeType attr, DialectBytecodeWriter &writer) const;
   void write(TokenType type, DialectBytecodeWriter &writer) const;
+
+  //===--------------------------------------------------------------------===//
+  // Version
+
+  std::unique_ptr<DialectVersion> readVersion(
```
fzakaria, line 269: should we use the identifier `override` here?

fzakaria (commented):
looks okay to me ; just confusing this version is some new identifier we have to track.
I'll wait for @GleasonK to review.

stablehlo/dialect/StablehloOps.h
```diff
@@ -52,6 +52,18 @@ limitations under the License.
...
+
+  // The dialect version read from bytecode.
+  uint32_t dialectVersion = 1;
```
GleasonK, line 64: `uint32_t dialectVersion = kCurrentDialectVersion;`?

stablehlo/dialect/StablehloOps.h
```diff
@@ -52,6 +52,18 @@ limitations under the License.
 namespace mlir {
 namespace stablehlo {
 
+struct StablehloDialectVersion : public mlir::DialectVersion {
```
GleasonK, line 55: I'm thinking through the user journey of how this is used.

Today:
- Breaking change to StableHLO is made (`dynamic_broadcast_in_dim` changing operands to 0-D tensors for example).
- If I'm a user passing StableHLO (instead of VHLO) across a boundary, my bytecode still works unless I'm writing a program that uses dynamism.
- If using dynamism my program breaks and I see a cryptic error.

With this change (and assuming we would bump the version for all dialect breaking changes):
- Breaking change to StableHLO is made.
- Bump the version += 1
- If I'm a user passing StableHLO across a boundary, my bytecode breaks regardless of dynamism use

It is _safer_ I agree, but for a dialect with no stability guarantees I'm not sure which way we're better off? Perhaps we could point the error message to some documentation to say "don't pass StableHLO across boundaries with versioning, see <link to doc page>", since this should only error in scenarios where version skew is possible?

fzakaria: Isn't that the part of the version being optionally set on serialization?
So maybe what is missing is augmenting why you may **not want to** set the version and why you may want to.

stablehlo/dialect/StablehloOps.h
jpienaar: That's a good question. Given both are serialization related really and that one is updated as process already, that could resolve other process questions. I was thinking of having this be separate in case it may be slower changing or there are changes that's only related to bytecode but not there or vice versa ... but simpler is better and I'm struggling to come up with cases where it wouldn't be sufficient.

stablehlo/dialect/StablehloOps.h
jpienaar: The cryptic error message is rather cryptic though :) But there is also another case.

I was thinking of 2 cases
1. The bytecode is passed across boundary and it still works (e.g., passes verification everything runs). It could be that an attribute was changed for an op that isn't even present. So we don't need to break/error on version diff - as to suggestion above, we could semver of VHLO and use different levels to warn/error or even leave it to the consumer to decide. Simplest is warn on mismatch but continue to verify before erroring, but it has a gap in the 2nd case.
2. Change where it still parses but the semantics of the op has changed. So I don't get any error, everything compiles and runs, I just get wrong results. (I couldn't find the policies for when major, minor or patch gets bumped - I'd assume changing an op semantics would be at least a minor version bump, probably major as subtle errors). Now it could be that this never happens, and StableHLO ops never change semantics, but just how things are spelled (so that verify would catch).

With single number I can't differentiate between these two and chose more exact/invalidate approach. With semver I could see one doing a "warn but verify" flow.

Even with XLA proto (which was unstable, but still mostly followed proto development approach and so rather stable), just knowing the change at which it was produced would have been useful. The expectation was that all component interacting was all compiled at same head, but no way to check. And we did run into this (very infrequently, but I think that's mostly due to development practices rather than as an indication of skew occurrences). Of course that would have been rather frequent changes. Point being, even where there isn't expected to be a version skew, it occurs, folks invariably dump modules and try to read later. We've had 2 breaking changes in ~9 months for folks here, so even without stability guarantees, they (incl ml researchers) would have experienced a rather reusable flow before being faced by an error which required them to know MLIR attribute types to decipher :)

GleasonK: I'm in agreement with the use cases, I guess my question is:

```
  if (version && StablehloDialectVersion::getCurrentVersion() < *version) {
    return reader.emitError("reading newer dialect than supported"), nullptr;
  }
```

Should this be a warning? Or is it on the producer to set the dialect version to a newer version if the new feature is being used? Otherwise LGTM.

GleasonK (approved):

PR #2007, opened 2024-02-14
Introduce PerAxis Quantized constraint for StableHLO Quantized OPs

abhigunj:
StableHLO OPs supporting Quantization can be categorized into following three types
a. Only PerTensor Quantized Tensors
b. Only PerAxis Quantized Tensors (no OP in this category for now)
c. Both PerTensor and PerAxis Quantized Tensors
PerAxis constraint from the PR allow only PerTensorQuantized Tensor inputs to type (a) OPs  and don't allow  PerAxis Quantized Tensors. 
Also,
Added negative test cases to validate this behavior
Added positive test cases for PerTensor , PerAxis Quantized Tensor support

Excluded OPs 
* Deprecated :`BroadCast` and `Call`
* Not Specced :  
 `dynamic_iota`, `create_token`, `dynamic_broadcast_in_dim`, `cross-replica-sum`, `einsum`, `unary_einsum`, `dynamic_reshape`, `set_dimension_size`, `trace`, `return`, `torch_index_select`, `real_dynamic_slice`, `dynamic_pad`, `dynamic_gather`, `dynamic_conv`, `dynamic_reshape_shape`, `cstr_reshapable`

Issue #1812, opened 2023-10-18
stablehlo.uniform_quantize cannot be serialized to bytecode


lsy323:
### What happened?

The MLIR module containing `stablehlo.uniform_quantize/dequantize` ops failed during bytecode serializing with error
```
loc("custom-call.6"): error: failed to legalize operation 'stablehlo.uniform_quantize' that was explicitly marked illegal
```

However, the MLIR module can be serialized to readable format
```
module @IrToHlo.18 attributes {mhlo.cross_program_prefetches = [], mhlo.is_dynamic = false, mhlo.use_auto_spmd_partitioning = false} {
  func.func @main(%arg0: tensor<10xf32>, %arg1: tensor<10x3x3x3xf32>, %arg2: tensor<1x3x10x10xf32>) -> tensor<1x10x8x8xf32> {
    %0 = stablehlo.constant dense<0.000000e+00> : tensor<1x10x8x8xf32>
    %1 = stablehlo.uniform_quantize %arg2 : (tensor<1x3x10x10xf32>) -> tensor<1x3x10x10x!quant.uniform<i8:f32, 1.000000e+00>>
    %2 = stablehlo.uniform_dequantize %1 : (tensor<1x3x10x10x!quant.uniform<i8:f32, 1.000000e+00>>) -> tensor<1x3x10x10xf32>
    %3 = stablehlo.uniform_quantize %arg1 : (tensor<10x3x3x3xf32>) -> tensor<10x3x3x3x!quant.uniform<i8<-127:127>:f32, 1.000000e+00>>
    %4 = stablehlo.uniform_dequantize %3 : (tensor<10x3x3x3x!quant.uniform<i8<-127:127>:f32, 1.000000e+00>>) -> tensor<10x3x3x3xf32>
    %5 = stablehlo.convolution(%2, %4) dim_numbers = [b, f, 0, 1]x[o, i, 0, 1]->[b, f, 0, 1], window = {stride = [1, 1], pad = [[0, 0], [0, 0]], lhs_dilate = [1, 1], rhs_dilate = [1, 1], reverse = [0, 0]} {batch_group_count = 1 : i64, feature_group_count = 1 : i64, precision_config = [#stablehlo<precision DEFAULT>, #stablehlo<precision DEFAULT>]} : (tensor<1x3x10x10xf32>, tensor<10x3x3x3xf32>) -> tensor<1x10x8x8xf32>
[...]

PR #2041, opened 2024-02-23
Quantization Verifiers based on T2x set of Traits

abhigunj:
* made `isCompatibleElementTypeForHloTypeInference` stricter to return error for {not Quantize, Quantize}, {per-axis Quantized, per-tensor Quantized} cases
* `AddOp` VHLO Test failures : addressed test failures because {not Quantize, Quantize} is not allowed 
* CorrectedTraits for `CholeskyOp` and `ClampOp` to match it with the spec

~~Note: This PR is based on in review PR https://github.com/openxla/stablehlo/pull/2007~~

Follow up PR will add/update OP verifiers for OPs which need special handling


stablehlo/dialect/Base.cpp
```diff
@@ -72,31 +72,34 @@ bool isCompatibleElementTypeForHloTypeInference(Type tp1, Type tp2) {
...
         qtp1.getStorageTypeMin() != qtp2.getStorageTypeMin() ||
-        qtp1.getStorageTypeMax() != qtp2.getStorageTypeMax())
+        qtp1.getStorageTypeMax() != qtp2.getStorageTypeMax() ||
+        qtp1.getExpressedType() != qtp2.getExpressedType()) {
       return false;
+    }
+
+    auto qpatp1 = qtp1.dyn_cast<quant::UniformQuantizedPerAxisType>();
+    auto qpatp2 = qtp2.dyn_cast<quant::UniformQuantizedPerAxisType>();
+    if (qpatp1 && qpatp2) {
```
sdasgup3, line 86: replace
```
    if (qpatp1 && qpatp2) {
      // Both are also per-axis quantized
      // Don't match dimensions, Per OP verifier will do it.
      return true;
    }
    // return true if both are per-tensor quantized
    return !(qpatp1 || qpatp2);
``` 
with 
```
return qpatp1 == qpatp2;
```

stablehlo/dialect/Base.cpp
```diff
@@ -72,31 +72,34 @@ bool isCompatibleElementTypeForHloTypeInference(Type tp1, Type tp2) {
...
-  // quantized/non-quantized across any combination of operands/results,
-  // and some differences in quantization parameters across operands/results.
-  // Individual ops may introduce additional constraints.
   auto qtp1 = tp1.dyn_cast<quant::QuantizedType>();
   auto qtp2 = tp2.dyn_cast<quant::QuantizedType>();
+
+  if (!qtp1 && !qtp2) {
+    // Both are non quantized
+    return tp1 == tp2;
+  }
```
sdasgup3, line 74: Add a comment here about what we are trying to achieve:
```
For quantized types, both `tp1` and `tp2` should be quantized types, with similar quantization granularity (i.e. both per-tensor or both per-axis)  with equal storage_type, expressed_type, storage_type_min and storage_type_max
```

stablehlo/dialect/Base.cpp
```diff
@@ -72,31 +72,34 @@ bool isCompatibleElementTypeForHloTypeInference(Type tp1, Type tp2) {
   tp1 = getElementTypeOrSelf(tp1);
   tp2 = getElementTypeOrSelf(tp2);
 
-  // Quantization: In the most general case, we allow any combination of
-  // quantized/non-quantized across any combination of operands/results,
-  // and some differences in quantization parameters across operands/results.
-  // Individual ops may introduce additional constraints.
   auto qtp1 = tp1.dyn_cast<quant::QuantizedType>();
   auto qtp2 = tp2.dyn_cast<quant::QuantizedType>();
+
```
sdasgup3, line 70: Add the following comment here
```
 // Sparsity: In the most general case, we allow any combination of
  // sparsity/denseness across any combination of operands/results, as well as
  // differences in sparsity encodings for operands and results.
  // Individual ops may introduce additional constraints.
  // No additional code is needed to check this because of how sparsity is
  // currently implemented.

  // Default case: Unless dynamism, quantization and/or sparsity are involved,
  // the types are required to be exactly equal.
```

sdasgup3 (commented):
Few initial comments. Will add more related to per-op changes

stablehlo/dialect/Base.cpp
abhigunj: qpatp* are ptr.. `return qpatp1 == qpatp2;` won't return desired value.
changed it to `return ((qpatp1 && qpatp2) || (!qpatp1 && !qpatp2));` thank you.

GleasonK (changes requested):

stablehlo/dialect/Base.td
```diff
@@ -45,36 +45,46 @@ def HLO_Complex : Complex<AnyTypeOf<[F32, F64]>>;
...
+  int bitwidth = width;
+}
+
+class StableHLO_UniformQuantizedPerAxisSignedInt<int width>
+  : Type<And<[CPred<"$_self.isa<mlir::quant::UniformQuantizedPerAxisType>()">,
            CPred<"$_self.cast<mlir::quant::UniformQuantizedPerAxisType>()" #
                  ".getStorageTypeIntegralWidth() == " # width>,
            CPred<"$_self.cast<mlir::quant::UniformQuantizedPerAxisType>()" #
-                 ".isSigned()">]>]>,
+                 ".isSigned()">]>,
```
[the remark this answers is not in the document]
sdasgup3: Yup that seems right. Having a signless check here could have helped  discover the issue https://github.com/openxla/stablehlo/issues/1812 earlier during Mhlo -> StableHLO.
Will fix it is follow up PR.

sdasgup3 (commented):
Some of the unhandled cases:

 - `compare op`: Per the spec it uses `baseline_element_type(lhs) =
 baseline_element_type(rhs)`. However, this check is implemented as https://github.com/openxla/stablehlo/blob/70db15a3433318aa922c7a7d7075cfc4af17c3df/stablehlo/dialect/StablehloOps.td#L1534. How about we add `HLO_CompatibleOperandsElementType` equivant to
 `SameOperandsElementType` but work with baseline_element_types.



  - `triangular_solve`: Use `HLO_CompatibleOperandsAndResultElementType` instead
  of `SameOperandsAndResultElementType`.


  - `batch_norm_grad`, `batch_norm_training`, and `batch_norm_inference`: The
  verifier uses `AllElementTypesMatch` whereas the spec wants the operand to
  matgch using baseline type. Can we introduce `HLO_AllElementTypesCompatible`.


  - `collective_broadcast`, `collective_permute`: They are currently using
  `HLO_CompatibleOperandsAndResultType`, but per the spec the type checks should
  be implemented using `SameOperandsAndResultType`

  - `reshape`: Currently uses `HLO_CompatibleOperandsAndResultType` but per the
  spec needs stricter type checks using `SameOperandsAndResultType`.

stablehlo/dialect/StablehloOps.td
```diff
@@ -1708,19 +1709,19 @@ def StableHLO_BatchNormGradOp : StableHLO_Op<"batch_norm_grad", [Pure,
   }];
 
   let arguments = (ins
-    RankedTensorOf<[HLO_Float]>:$operand, /*batch_norm_grad_i1*/
-    1DTensorOf<[HLO_Float]>:$scale, /*batch_norm_grad_i2*/
-    1DTensorOf<[HLO_Float]>:$mean, /*batch_norm_grad_i3*/
-    1DTensorOf<[HLO_Float]>:$variance, /*batch_norm_grad_i4*/
-    RankedTensorOf<[HLO_Float]>:$grad_output, /*batch_norm_grad_i5*/
+    RankedTensorOf<[HLO_Float, HLO_QuantizedInt]>:$operand, /*batch_norm_grad_i1*/
```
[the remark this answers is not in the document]
abhigunj: thanks, created a ticket to review naming once all the Q verifier changes are in.

abhigunj:
> Some of the unhandled cases:
> 
> * `compare op`: Per the spec it uses `baseline_element_type(lhs) =  baseline_element_type(rhs)`. However, this check is implemented as https://github.com/openxla/stablehlo/blob/70db15a3433318aa922c7a7d7075cfc4af17c3df/stablehlo/dialect/StablehloOps.td#L1534
>   . How about we add `HLO_CompatibleOperandsElementType` equivant to
>   `SameOperandsElementType` but work with baseline_element_types.
> * `triangular_solve`: Use `HLO_CompatibleOperandsAndResultElementType` instead
>   of `SameOperandsAndResultElementType`.
> * `batch_norm_grad`, `batch_norm_training`, and `batch_norm_inference`: The
>   verifier uses `AllElementTypesMatch` whereas the spec wants the operand to
>   matgch using baseline type. Can we introduce `HLO_AllElementTypesCompatible`.
> * `collective_broadcast`, `collective_permute`: They are currently using
>   `HLO_CompatibleOperandsAndResultType`, but per the spec the type checks should
>   be implemented using `SameOperandsAndResultType`
> * `reshape`: Currently uses `HLO_CompatibleOperandsAndResultType` but per the
>   spec needs stricter type checks using `SameOperandsAndResultType`.

Thanks again for the review. 
`collective_broadcast` `collective_permute` `reshape` are being handled in follow up PR since existing Trait can't be directly reused for them. Added Traits for other OPs.

stablehlo/dialect/Base.td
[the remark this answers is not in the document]
sdasgup3: Here is the PR: https://github.com/openxla/stablehlo/pull/2054

stablehlo/dialect/Base.h
```diff
@@ -306,6 +306,29 @@ class CompatibleOperandsAndResultElementType
...
+    auto typeMatch = [&](Type actual) {
+      return isCompatibleElementTypeForHloTypeInference(actual, expected);
+    };
+    auto allMatch = llvm::all_of(op->getOperandTypes(), typeMatch);
+    if (!allMatch) {
+      return op->emitOpError(
+          "requires compatible element types for all operands");
+    }
+
+    return success(allMatch);
```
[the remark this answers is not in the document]
abhigunj: Done, thanks for the note.

stablehlo/dialect/Base.cpp
```diff
@@ -72,20 +65,32 @@ bool isCompatibleElementTypeForHloTypeInference(Type tp1, Type tp2) {
...
-        qtp1.getStorageTypeMax() != qtp2.getStorageTypeMax())
+        qtp1.getStorageTypeMax() != qtp2.getStorageTypeMax() ||
+        qtp1.getExpressedType() != qtp2.getExpressedType()) {
       return false;
+    }
+
+    auto qpatp1 = qtp1.dyn_cast<quant::UniformQuantizedPerAxisType>();
+    auto qpatp2 = qtp2.dyn_cast<quant::UniformQuantizedPerAxisType>();
+    bool quantizationGranularityMatches =
+        (qpatp1 && qpatp2) || (!qpatp1 && !qpatp2);
```
GleasonK, line 87: Do we have any negative test cases for this?

stablehlo/dialect/Base.cpp
```diff
@@ -72,20 +65,32 @@ bool isCompatibleElementTypeForHloTypeInference(Type tp1, Type tp2) {
...
+  //   c. with equal storage_type, storage_type_min, storage_type_max, and
+  //   expressed_type
   auto qtp1 = tp1.dyn_cast<quant::QuantizedType>();
   auto qtp2 = tp2.dyn_cast<quant::QuantizedType>();
   if (qtp1 && qtp2) {
     if (qtp1.getStorageType() != qtp2.getStorageType() ||
         qtp1.getStorageTypeMin() != qtp2.getStorageTypeMin() ||
-        qtp1.getStorageTypeMax() != qtp2.getStorageTypeMax())
+        qtp1.getStorageTypeMax() != qtp2.getStorageTypeMax() ||
+        qtp1.getExpressedType() != qtp2.getExpressedType()) {
```
GleasonK, line 80: Do we have a negative test case for this?

GleasonK (approved):
A few test recommendations to ensure this is all working as expected.

stablehlo/dialect/Base.cpp
abhigunj: thanks, yeh we do have negative tests to validate this 
https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/ops_stablehlo.mlir#L5074-L5122

stablehlo/dialect/Base.cpp
abhigunj: https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/ops_stablehlo.mlir#L5074-L5122

PR #2044, opened 2024-02-26
Free space on GitHub action

fzakaria:
Our GitHub Bazel runners seem to be failing with no space left on the device.
The size of the build graph might haven gotten larger with Bazel 7 change that introduced modules + python.
(This is a good thing as before Python was not hermetic)

This PR adds some recommended practices to free up a healthy amount of disk space on the action.

 * Add a script for Bazel builds to free up lots of space on the GitHub action

### DIFF
--- a/MODULE.bazel.lock
+++ b/MODULE.bazel.lock
@@ -1,6 +1,6 @@
 {
   "lockFileVersion": 3,
-  "moduleFileHash": "836f0a7d2276ed93403f104a10008b94ec7e7f81b8d6921cea287f0a6d364efa",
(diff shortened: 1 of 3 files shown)


MODULE.bazel.lock
```diff
@@ -1,6 +1,6 @@
 {
```
fzakaria, line 1: My lock file had changed -- I think since it tracks the line numbers; maybe the license addition changed the lockfile.

We should probably fail CI if this is changed during the build.
https://bazel.build/external/lockfile#lockfile-usage

fzakaria:
@mlevesquedion i'm not sure about this one --
We have two types of runners: large & normal

I couldn't figure out how to run this action only when the "large-runner" is **not** triggered.
It seems to take a longtime for the large-runner to clear and not particularly needed at that point.

PR #1792, opened 2023-10-05
[RFC] Add hybrid quantized convolution and dot_general for weight-only quantization

doyeonkim0:
This RFC proposes to add hybrid quantized convolution and dot_general for weight-only quantization.
Please let me know your feedback on this.

The RFC partially address the issue https://github.com/openxla/stablehlo/issues/1575 w.r.t supporting weight only quantization support in StableHLO. The remaining tasks for the above the are highlighted [here](https://github.com/openxla/stablehlo/issues/1575#issuecomment-1819957802).

Issue #1156, opened 2023-02-11
Decide on speccing numerical accuracy


burmako:
When writing the initial version of the spec, we've abstracted ourselves away from thinking about numerical accuracy, stating that it is implementation-defined. Before we release StableHLO v1.0, it would be good to decide whether we want to do more or whether we should keep the current wording.

PR #2050, opened 2024-02-26
Execution semantics of Stablehlo quantized operations

sdasgup3:
The PR provides clarification on the execution semantics of the StableHLO quantized operations.

@loganchien
@paulinesho
@doyeonkim0
@dominicsymes
@cbasile-g

Looking forward to your feedback.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -6236,6 +6236,15 @@ At the moment, StableHLO does not provide guarantees about numerical accuracy,
 but this may change in the future
 ([#1156](https://github.com/openxla/stablehlo/issues/1156)).
 
+### Execution semantics of quantized operation
+
+The interpretation of quantized StableHLO operations may vary depending on the
+hardware requirements and capabilities. For instance, some hardware may opt to
+interpret quantized operations using a "dequantize, perform floating-point
+operation, and finally quantize" strategy. Others may perform the entire
+computation in the integer domain. Consequently, the interpretation of quantized
+StableHLO operations is exclusively determined by the specific implementation.
+
 ### Errors
 
 StableHLO programs are validated through an extensive set of constraints for
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -6242,8 +6242,12 @@ The interpretation of quantized StableHLO operations may vary depending on the
 hardware requirements and capabilities. For instance, some hardware may opt to
 interpret quantized operations using a "dequantize, perform floating-point
 operation, and finally quantize" strategy. Others may perform the entire
-computation in the integer domain. Consequently, the interpretation of quantized
-StableHLO operations is exclusively determined by the specific implementation.
+computation with integer arithmetic. Consequently, the interpretation of
+quantized StableHLO operations is exclusively determined by the specific
+implementation. The interpretation of hybrid quantization
+([#1575](https://github.com/openxla/stablehlo/issues/1575)) should be based on
+the it's semantics as prescribed in the specification (via
+[1792](https://github.com/openxla/stablehlo/pull/1792)).
 
 ### Errors
 


loganchien (approved):

docs/spec.md
```diff
@@ -6185,6 +6185,15 @@ At the moment, StableHLO does not provide guarantees about numerical accuracy,
 but this may change in the future
 ([#1156](https://github.com/openxla/stablehlo/issues/1156)).
 
+### Execution semantics of quantized operation
+
+The interpretation of quantized StableHLO operations may vary depending on the
+hardware requirements and capabilities. For instance, some hardware may opt to
+interpret quantized operations using a "dequantize, perform floating-point
+operation, and finally quantize" strategy. Others may perform the entire
+computation in the integer domain. Consequently, the interpretation of quantized
```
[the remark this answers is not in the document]
sdasgup3: Sure. done.

doyeonkim0 (approved):

docs/spec.md
```diff
@@ -6185,6 +6185,16 @@ At the moment, StableHLO does not provide guarantees about numerical accuracy,
...
 ([#1156](https://github.com/openxla/stablehlo/issues/1156)).
 
+### Execution semantics of quantized operation
+
+The interpretation of quantized StableHLO operations may vary depending on the
+hardware requirements and capabilities. For instance, some hardware may opt to
+interpret quantized operations using a "dequantize, perform floating-point
+operation, and finally quantize" strategy. Others may perform the entire
+computation with integer arithmetic. Consequently, the interpretation of
+quantized StableHLO operations is exclusively determined by the specific
```
dansuh17, line 6195: How does `precision_config` in `dot_general` or `convolution` play with this flexibility? Take a quantized convolution op, for example; if the precision config is `HIGHEST` is it still ok to use integer accumulation?

Related:
* https://github.com/openxla/stablehlo/issues/755
* https://github.com/openxla/stablehlo/blob/main/docs/spec.md#dot_general

dansuh17 (changes requested):

docs/spec.md
sdasgup3: Hi @dansuh17 
That is a good point.

If the expectation is to have  `HIGHEST` precision, then the implementation can use `f32` or `f64` (as applicable) for computation (using the dequant-op-quant strategy). With that said, for quantized operations, an implementation can choose to a different implementation compared to the naive definition of  precision config (which IMO a trade-off between precison and speed) as long as the semantics of the operations is preserved. We will be able to evaluate the later part (preversing the semantics) with strong numerical accuracy guarentees defined at the StableHLO level (https://github.com/openxla/stablehlo/issues/1156, yet to be defined)

loganchien: IMHO, the concept of quantization contradicts with `precision_config`.  Representing `f32` or `f64` in 8-bit value already throws away a lot of precision bits.  I don't think there are sensible way to combine both of them and provide some kind of guarantees or set some kind of expectations.

sdasgup3: Thanks @loganchien. I can see your point. 

IMO, we should clarify the coexistence of  quantization operations with precision config. At this point there exist some open questions like (1) if quantization and `precision_config`  should co-exist, and (2) if yes, what is the desired behavior.
 
@dansuh17 
Added an AI @https://github.com/openxla/stablehlo/issues/755#issuecomment-1977199585. How about we take care of this as part of resolving the linked issue?

dansuh17: SG, thanks for addressing this. Could you also add in `spec.md` about the current status regarding `precision_config`?

dansuh17 (approved):

docs/spec.md
sdasgup3: Thanks for the approval, @dansuh17 !
 
In the spec we left a comment from the underspecified nature of `precision_config` and referred to #755 at location https://github.com/openxla/stablehlo/blob/main/docs/spec.md#semantics-30.

PTAL.

dominicsymes: For a quantized implementation of dot_general, where the accumulation is done as integer accumulation followed by a post-scale, I think it can be useful to select the width of the accumulator. For example, for quantized 16-bit by 8-bit dot product operations, both 32-bit and 48-bit accumulation range may be useful. So, I think there is a role for something like precision_config in quantized arithmetic to define the accumulator range.

sdasgup3: Hi Dominic,
That is a valid point and IMO should be captured as part of speccing the `precision_config`, which, in its current form,  may impact quantized operations in an unspecified way.
How about we take this as part of #755?

paulinesho:
Would principles from this discussion apply to dynamic range quantization vs weight-only quantization with respect to hybrid signatures as well? This direction seems to imply that we don't need to disambiguate between the two and push that responsibility down to hardware preferences, but it does mean we could have the same source model performing very differently on different hardware targets. Or is `precision_config` intended to help with this ambiguity?

docs/spec.md
dominicsymes: Hi @sdasgup3 . Yes, it seems fine to take this separately. If I understand correctly this patch is making clear that different types of implementations are possible for quantized in terms of mixed floating-point or full integer. This patch does not define further details on what level of accuracy or range is expected and this would be covered in other issues, for example on precision_config.

dominicsymes (approved):

sdasgup3:
>> Would principles from this discussion apply to dynamic range quantization vs weight-only quantization with respect to hybrid signatures as well? This direction seems to imply that we don't need to disambiguate between the two and push that responsibility down to hardware preferences, but it does mean we could have the same source model performing very differently on different hardware targets. Or is precision_config intended to help with this ambiguity?

@paulinesho Good point!

IMO, hybrid quantized ops are special in the sense that only one of the operands in quantized and other being float. Per the above direction and the following semantics from https://github.com/openxla/stablehlo/pull/1792, there exist limited freedom for implementation: de-quantize that operand and do fp computation. 

```
`hybrid_dequantize_then_op` is used to specify weight-only quantization for
hybrid op which accepts lhs in floating-point and rhs in quantized types. It
dequantizes quantized inputs into their expressed types and performs computation
in float.
```

For an implementation to choose DRQ implementation for weight-only require a calibration step for the activation (to figure out the scales and zero-points) which IMO would be an over-streatch to be inferred from the weight-only semantics as specified in https://github.com/openxla/stablehlo/pull/1792. 

With quantized operations like `add` with both operands quantized there is an additional implementation freedom where  the entire computation can be done with integer arithmetic (instead of dequant>float-op>quant), which is where the pros is relevant.

sdasgup3:
@paulinesho I have updated the patch adding clarification for the hybrid quantization cases. Please let me know if it looks good.

GleasonK (approved):

PR #2056, opened 2024-02-27
Add pass to refine dynamic function parameters into static arguments

GleasonK:
We have `StablehloRefineShapes` but using it is currently not easy. Function arguments must be replaced with static argument sizes, and a buffer must be put in place to stop the new types from creating invalid IR, then refine shapes can be used.

This CL simplifies this entire process by adding a pass which takes care of the argument refinement, preparing the program for shape refinement.

For example, consider the following program which broadcasts and computes: `add(tensor<?x10xf32>, tensor<1xf32>)`

```sh
$ cat /tmp/t.mlir 
module @IrToHlo.21 {
  func.func @main(%arg0: tensor<1xf32>, %arg1: tensor<?x10xf32>) -> tensor<?x10xf32> {
    %0 = stablehlo.constant dense<1> : tensor<2xi32>
    %1 = stablehlo.constant dense<10> : tensor<1xi32>
    %2 = stablehlo.get_dimension_size %arg1, dim = 0 : (tensor<?x10xf32>) -> tensor<i32>
    %3 = stablehlo.reshape %2 : (tensor<i32>) -> tensor<1xi32>
    %4 = stablehlo.concatenate %3, %1, dim = 0 : (tensor<1xi32>, tensor<1xi32>) -> tensor<2xi32>
    %5 = stablehlo.maximum %4, %0 : tensor<2xi32>
    %6 = stablehlo.dynamic_broadcast_in_dim %arg1, %5, dims = [0, 1] : (tensor<?x10xf32>, tensor<2xi32>) -> tensor<?x10xf32>
    %7 = stablehlo.dynamic_broadcast_in_dim %arg0, %5, dims = [1] : (tensor<1xf32>, tensor<2xi32>) -> tensor<?x10xf32>
    %8 = stablehlo.add %6, %7 : tensor<?x10xf32>
    return %8 : tensor<?x10xf32>
  }
}
```

The parameter refinement pass validates refinements, replaces the FuncOp's arguments with static operands, and wraps the operands in a custom_call `shape_refinement_operand_wrapper`, which allows the rest of the program to remain dynamic until shape refinement is called, which then removes these custom_calls:

```
$ stablehlo-opt /tmp/t.mlir --stablehlo-refine-parameters=shapes="tensor<1xf32>,tensor<5x10xf32>"
module @IrToHlo.21 {
  func.func @main(%arg0: tensor<1xf32>, %arg1: tensor<5x10xf32>) -> tensor<?x10xf32> {
    %0 = stablehlo.constant dense<[5, 10]> : tensor<2xi64>
    %1 = stablehlo.custom_call @stablehlo.shape_refinement_operand_wrapper(%arg1, %0) {indices_of_shape_operands = dense<1> : tensor<1xi64>} : (tensor<5x10xf32>, tensor<2xi64>) -> tensor<?x10xf32>
    [...]
    %8 = stablehlo.dynamic_broadcast_in_dim %1, %7, dims = [0, 1] : (tensor<?x10xf32>, tensor<2xi32>) -> tensor<?x10xf32>
    %9 = stablehlo.dynamic_broadcast_in_dim %arg0, %7, dims = [1] : (tensor<1xf32>, tensor<2xi32>) -> tensor<?x10xf32>
    %10 = stablehlo.add %8, %9 : tensor<?x10xf32>
    return %10 : tensor<?x10xf32>
  }
}
```

With this pass, and leveraging existing shape refinement / dynamism canonicalization / aggressive simplification, we can refine unbounded dynamic program into a static program which computes `add(tensor<5x10xf32>, tensor<1xf32>)`

```sh
$ stablehlo-opt /tmp/t.mlir --stablehlo-refine-parameters=shapes="tensor<1xf32>,tensor<5x10xf32>" --stablehlo-refine-shapes --stablehlo-canonicalize-dynamism --stablehlo-aggressive-simplification
module @IrToHlo.21 {
  func.func @main(%arg0: tensor<1xf32>, %arg1: tensor<5x10xf32>) -> tensor<5x10xf32> {
    %0 = stablehlo.broadcast_in_dim %arg0, dims = [1] : (tensor<1xf32>) -> tensor<5x10xf32>
    %1 = stablehlo.add %arg1, %0 : tensor<5x10xf32>
    return %1 : tensor<5x10xf32>
  }
}
```

Much of this work is a refactoring / MLIR-ification of work @mlevesquedion completed, i.e. the wrapping of operand in custom_call.

Co-authored-by: Michael Levesque-Dion <mlevesquedion@google.com>


stablehlo/transforms/StablehloRefineParameters.cpp
```diff
@@ -0,0 +1,210 @@
...
+                     "dimensions ")
+                 << type << " -> " << refinedType;
+        }
+      }
+    }
+  }
+  return success();
+}
+
+void wrapRefinedOperands(func::FuncOp func, TypeRange refinedTypes) {
```
sdasgup3, line 136: A comment would be good to understand how the "operand barrier" custom calls helps in keeping the IR valid when we update the function input types to static.

stablehlo/transforms/Passes.td
```diff
@@ -79,6 +79,20 @@ def StablehloRefineShapesPass : Pass<"stablehlo-refine-shapes", "ModuleOp"> {
...
+def StablehloRefineParametersPass : Pass<"stablehlo-refine-parameters", "ModuleOp"> {
+  let summary = "Refines the parameter shapes of the main function.";
+  let description = [{
+    Modifies the parameters of the main function using the input type signature.
+    Wraps parameters in bitcast_convert ops to keep the rest of the program the
+    same for shape refinement pass.
+  }];
+  let dependentDialects = ["mlir::stablehlo::StablehloDialect"];
+  let options = [
+    ListOption<"refinedShapesOption", "shapes", "std::string",
```
sdasgup3, line 91: It would be ideal if we limit these string arguments specify only the shape (not type) as shape is all we want to specialize. However, parsing wont be a "one-shot" as it is now. 

what do you think?

stablehlo/transforms/StablehloRefineParameters.cpp
sdasgup3: Ah just realized that we already explained that in the PR description. Thanks!
Maybe we could add a concise code comment along similar lines?

GleasonK: Added description, good call

stablehlo/transforms/Passes.td
GleasonK: I went through a few ideas, but I think I like where it is now, it's very explicit and difficult to misuse. You must specify exactly the new type signature of the `main` function to be refined.

See the test cases, some aren't tensors `!stablehlo.token`, how would we specify these? Would we skip them? Then the question becomes, do we only specify the dynamic operands? What if we only want to partially refine, how do we skip an operand, etc.

Specifying everything made for an easy contract, and the programmatic API will be to just pass the operand types to the refinement pass

GleasonK: Added some details to the pass description

GleasonK:
Sorry messed up the commit chain and had to force push, won't be able to diff the addressed feedback easily

stablehlo/transforms/Passes.td
```diff
@@ -79,6 +79,20 @@ def StablehloRefineShapesPass : Pass<"stablehlo-refine-shapes", "ModuleOp"> {
   }];
 }
 
+def StablehloRefineParametersPass : Pass<"stablehlo-refine-parameters", "ModuleOp"> {
+  let summary = "Refines the parameter shapes of the main function.";
+  let description = [{
+    Modifies the parameters of the main function using the input type signature.
+    Wraps parameters in bitcast_convert ops to keep the rest of the program the
```
sdasgup3, line 86: `Wraps parameters in bitcast_convert` Is this still true?

stablehlo/transforms/StablehloRefineParameters.cpp
```diff
@@ -0,0 +1,210 @@
+/* Copyright 2022 The StableHLO Authors.
```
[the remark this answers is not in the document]
sdasgup3: do we need a change here?

sdasgup3 (approved):

stablehlo/transforms/Passes.td
GleasonK: oof, also realizing I made all your proposed edits in the wrong client. Will update here shortly.

GleasonK: done, also made sure all previous comments were addressed

stablehlo/transforms/StablehloRefineParameters.cpp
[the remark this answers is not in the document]
GleasonK: See other comment, I edited wrong client. Should be good now.

sdasgup3 (approved):

stablehlo/transforms/StablehloRefineParameters.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+struct StablehloRefineParametersPass
+    : public impl::StablehloRefineParametersPassBase<
+          StablehloRefineParametersPass> {
+  using Super =
+      impl::StablehloRefineParametersPassBase<StablehloRefineParametersPass>;
+  StablehloRefineParametersPass() : Super() {}
+  StablehloRefineParametersPass(
+      const StablehloRefineParametersPassOptions& opts)
+      : Super(opts) {}
```
mlevesquedion, line 188: I think you can simplify this to just be `using StablehloRefineParametersPassBase::StablehloRefineParametersPassBase`, which will bring in the constructors.

stablehlo/transforms/StablehloRefineParameters.cpp
```diff
@@ -0,0 +1,226 @@
...
+  StablehloRefineParametersPass(TypeRange refinedTypes) : Super() {
+    this->refinedTypes = llvm::to_vector(refinedTypes);
```
mlevesquedion, line 190:
```suggestion
    refinedTypes = llvm::to_vector(refinedTypes);
```

stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -706,7 +707,18 @@ struct RefineCustomCallOpPattern : public OpRewritePattern<CustomCallOp> {
...
+
+    // Clean up operand buffers after refinement
+    // Must do in this pattern to avoid needing multiple refinement iterations
+    if (op.getCallTargetName().equals(kCustomCallOperandBarrierTarget)) {
+      Value operand = op.getOperand(0);
+      if (operand.getType() == op.getResult(0).getType()) {
+        op.replaceAllUsesWith(ValueRange(operand));
+      }
+    }
+    return success();
```
mlevesquedion, line 721: Consider adding some test cases to `stablehlo_refine_shapes.mlir`.

stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -1060,6 +1072,14 @@ struct UpdateRegionTypePattern : public OpRewritePattern<ReturnOp> {
   }
 };
 
+struct CleanupOperandBufferPattern : public OpRewritePattern<CustomCallOp> {
```
mlevesquedion, line 1075: I don't understand what this pattern is for. Add a comment?

stablehlo/transforms/StablehloRefineParameters.cpp
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  void runOnOperation() override {
+    auto func = getStablehloRefineShapesTarget(getOperation());
```
mlevesquedion, line 194: Maybe this should be renamed to `getMainOrUniqueFunction` or something since it's being used in more than one pass now. (I think the interpreter has similar logic, too.)

stablehlo/transforms/Passes.td
```diff
@@ -79,6 +79,25 @@ def StablehloRefineShapesPass : Pass<"stablehlo-refine-shapes", "ModuleOp"> {
   }];
 }
 
+def StablehloRefineParametersPass : Pass<"stablehlo-refine-parameters", "ModuleOp"> {
```
mlevesquedion, line 82: To my understanding running this pass on its own doesn't make a whole lot of sense, as it will introduce calls to a function that doesn't exist, and these should be cleaned up. So if you're running this pass, shape refinement should also run, probably right after this pass. I wonder if there's a way to enforce that? Or maybe we should make that explicit in the summary and description.

mlevesquedion (approved):
Neat! Some nits, mostly I think some error messages and names could be more explicit, and it would be nice to have some kind of end to end test that shows how this works together with the shape refinement pass.

stablehlo/transforms/Passes.td
```diff
@@ -79,6 +79,25 @@ def StablehloRefineShapesPass : Pass<"stablehlo-refine-shapes", "ModuleOp"> {
...
+    Wraps parameters in custom_call @stablehlo.shape_refinement_operand_wrapper
+    to keep the IR valid before shape refinement is run.
+
+    The `refinedShapesOption` can be used to specify a list of refined types.
+    This can be specified in MLIR with `--shapes='tensor<...>,tensor<...>'`, or
+    passed to the pass create method. The refinement type list must specify the
+    type of every argument to the `main` method being refined.
+  }];
+  let dependentDialects = ["mlir::stablehlo::StablehloDialect"];
+  let options = [
```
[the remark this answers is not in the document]
GleasonK: Could be, but the same would need to exist for RefineShapes. For the sake of our ecosystem, requiring a `main` function when multiple functions are needed seems to be consistent.

stablehlo/transforms/Passes.td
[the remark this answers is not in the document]
GleasonK: Agree. Done.

stablehlo/transforms/StablehloRefineParameters.cpp
GleasonK: Always uneasy on the name conflicts.. Will add underscore to the arg name

stablehlo/transforms/StablehloRefineShapes.cpp
GleasonK: Good point. Will do.

stablehlo/transforms/StablehloRefineParameters.cpp
GleasonK: Going to hold off on that rename for now as well. I.e. what if we wanted to provide the ability to refine an unused function, if there were a way to have multiple entrypoints.

It's a weak argument, but as I was making this change I saw the code comment for this function and looks like it was fairly deliberately named

stablehlo/transforms/Passes.td
GleasonK: We can provide a pass pipeline, but I think keeping the responsibilities separates is best for testability.

GleasonK: Calling that a followup, I think once we get to integrating this pass we can get a better idea on that.

stablehlo/transforms/StablehloRefineShapes.cpp
GleasonK: Should be deleted, old cruft. Initially tried to use it as a separate pattern, but refine shapes tries to complete in a single iteration over the program and this relied on pattern ordering, so bundled into the CustomCallOp pattern

stablehlo/transforms/StablehloRefineParameters.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+#define GEN_PASS_DEF_STABLEHLOREFINEPARAMETERSPASS
+#include "stablehlo/transforms/Passes.h.inc"
+
+namespace {
+
+CustomCallOp makeShapeRefinementOperandWrapper(OpBuilder& builder,
+                                               Value operand,
+                                               Type refinedType) {
+  auto rankedTensor = refinedType.cast<RankedTensorType>();
```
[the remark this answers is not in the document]
GleasonK: done

stablehlo/transforms/Passes.td
```diff
@@ -79,6 +79,25 @@ def StablehloRefineShapesPass : Pass<"stablehlo-refine-shapes", "ModuleOp"> {
...
 }
 
+def StablehloRefineParametersPass : Pass<"stablehlo-refine-parameters", "ModuleOp"> {
+  let summary = "Refines the parameter shapes of the main function.";
+  let description = [{
+    Modifies the parameters of the main function using the input type signature.
+    Wraps parameters in custom_call @stablehlo.shape_refinement_operand_wrapper
+    to keep the IR valid before shape refinement is run.
+
+    The `refinedShapesOption` can be used to specify a list of refined types.
```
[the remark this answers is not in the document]
GleasonK: shapes->types

stablehlo/transforms/StablehloRefineParameters.cpp
```diff
@@ -0,0 +1,226 @@
...
+    if (refinedType.isa<UnrankedTensorType>()) {
+      return func.emitOpError("refinement must be ranked ")
+             << type << " -> " << refinedType;
+    }
+
+    if (tensorType.hasRank()) {
+      // Validate ranks match if ranked (must allow unranked tensorType)
+      if (tensorType.getRank() != refinedTensorType.getRank()) {
+        return func.emitOpError("refinement rank must match operand rank ")
+               << type << " -> " << refinedType;
```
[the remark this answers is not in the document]
GleasonK: updated wording. see test files

stablehlo/transforms/StablehloRefineParameters.cpp
```diff
@@ -0,0 +1,226 @@
...
+       llvm::zip(func.getArgumentTypes(), refinedTypes)) {
+    // Always allow skipping refinement
+    if (type == refinedType) continue;
+
+    // If mismatched, must be tensor types
+    auto tensorType = type.dyn_cast<TensorType>();
+    auto refinedTensorType = refinedType.dyn_cast<TensorType>();
+    if (!tensorType || !refinedTensorType) {
+      return func.emitOpError("refinement type must be tensor ")
+             << type << " -> " << refinedType;
```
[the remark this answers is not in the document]
GleasonK: added index, updated wording.

stablehlo/transforms/StablehloRefineParameters.cpp
```diff
@@ -0,0 +1,226 @@
...
+                               builder.getI64TensorAttr({1}))});
+}
+
+ParseResult parseRefinedTypes(ModuleOp module,
+                              ArrayRef<std::string> shapeString,
+                              SmallVector<Type>& refinedTypes) {
+  MLIRContext* context = module.getContext();
+  for (const auto& shape : shapeString) {
+    Type type = parseType(shape, context);
+    if (!type) return module->emitOpError("Invalid type string: ") << shape;
```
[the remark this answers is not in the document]
GleasonK: done.

PR #2059, opened 2024-02-29
GetDimensionSizeOp : make arguments quantized type

abhigunj:
~~DynamicBroadcastInDimOp:  is yet to be a part of spec.~~ (see comments)
GetDimensionSizeOp: will be changed to quantized types in spec. 
Back porting from the integration as file check tests were failing for both.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -3040,14 +3040,15 @@ behavior is undefined. More formally, for all `i1 < i2` from `indices(result)`,
 #### Semantics
 
 Produces the size of the given `dimension` of the `operand`. More formally,
-`result = dim(operand, dimension)`.
+`result = dim(operand, dimension)`. The Semantics concerns only with the shape
+component of the type. The element-type could be anything.
 
 #### Inputs
 
-| Label | Name        | Type                    | Constraints |
-|-------|-------------|-------------------------|-------------|
-| (I1)  | `operand`   | tensor                  | (C1)        |
-| (I2)  | `dimension` | constant of type `si64` | (C1)        |
+| Label | Name        | Type                       | Constraints |
+|-------|-------------|----------------------------|-------------|
+| (I1)  | `operand`   | tensor or quantized tensor | (C1)        |
+| (I2)  | `dimension` | constant of type `si64`    | (C1)        |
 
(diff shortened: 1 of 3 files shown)


sdasgup3:
`DynamicBroadcastInDimOp`: Feel free to update a description with that fact that there are internal tests written with per-axis support which we would like to maintain till the op makes its place to spec (at which point the decision can be revisted)

`GetDimensionSize`: Add the spec changes in the same PR stating that its semantics concerns only with the shape component of the type: The element-type could be anything.

GleasonK (commented):
LGTM for GetDimSize. I can understand the use case here, and given that the output is not quantized, makes sense.

DynamicBroadcastInDim should match BroadcastInDim. Let's talk about the rationale for this a bit more before I approve. If this is to support a test case somewhere, perhaps we can remove that test case if it doesn't make sense?

What makes me hesitant is, what should the per-axis quantization be when the rank of the output doesn't match the rank of the input?

GleasonK:
I also think that a ticket titled "Revisit quantization support for `dynamic_*` ops" is a fair outcome of this if we truly do want to defer this work until after speccing?

sdasgup3:
>> What makes me hesitant is, what should the per-axis quantization be when the rank of the output doesn't match the rank of the input?

As the output shape is determined as by an operand for this op, it might not be possible to represent the output quantization type (unknown number of zero-point/scale pairs). Let's check if we can remove the test.

abhigunj:
Thank you folks. Taking a look if we can remove the tests and keep DynamicBroadcastInDimOp without per-axis quantization for now.

@GleasonK  This PR needs to be merged before the start of this week integration to avoid reversing GetDimensionSize type change. I will update the PR soon regarding DynamicBroadcastInDimOp test removal and merge it.

abhigunj:
Removed the DynamicBroadcastInDimOp changes form the PR; working on removing /updating tests from the g3. 

@GleasonK please flip the `HLO_Tensor` -> `HLO_TensorOrPerAxisQuantizedTensor` change during integration to avoid expected test failures. Sorry for the trouble.

GleasonK (approved):

PR #2060, opened 2024-02-29
Update bazel lockfile and check lockfile in CI

mlevesquedion:
I ran `bazel build --lockfile_mode=error` and it told me to run `bazel mod deps --lockfile_mode=update`, which led to a pretty big diff in the lockfile. I also tried just doing `bazel build` and that leads to a smaller diff (just the changes at the top of the file), but it seems more prudent to do what the tool says.


fzakaria (approved):
not 100% sure why two modes produce a different lockfile...

I like adding it to the CI action though to catch failure to update in the future.

mlevesquedion:
> not 100% sure why two modes produce a different lockfile...
> 
> I like adding it to the CI action though to catch failure to update in the future.

It seems the diff is not between modes, but between commands. According to https://bazel.build/external/lockfile, `update` is the default mode. When I ran `bazel build [--lockfile_mode=update]`, I got the header changes, when I ran `bazel mod deps [--lockfile_mode=update]`, I got all the changes. Not sure why this diff, but in the future if this fails in CI people will be instructed by the tool to run `bazel mod deps`, so might as well use that.

PR #2062, opened 2024-03-01
Add a python API for reference interpreter

GleasonK:
```py
ASM = """
func.func @test(%arg0: tensor<2x2xf16>) -> tensor<2x2xf16> {
  %0 = stablehlo.add %arg0, %arg0 : (tensor<2x2xf16>, tensor<2x2xf16>) -> tensor<2x2xf16>
  func.return %0 : tensor<2x2xf16>
}
"""

m = ir.Module.parse(ASM)
arg = np.asarray([1, 2, 3, 4], np.float16).reshape(2,2)
args = [ir.DenseIntElementsAttr.get(arg)]
res = stablehlo.eval_module(m, args)
```

- Added python API for reference interpreter, and test cases for int/float datatypes
- Added reference API which takes DenseElementsAttrs to be used by python API
- Added method to convert between `stablehlo::Tensor` and `mlir::DenseElementsAttr` more easily (only supports int/float types)
- Added more error checking at interpreter entry point to improve UX
- Changed error reporting at API boundary from `llvm::Error` to `mlir::LogicalResult` to be more consistent for external users, also improves python error experience

### DIFF
--- a/stablehlo/integrations/python/tests/stablehlo.py
+++ b/stablehlo/integrations/python/tests/stablehlo.py
@@ -257,11 +257,9 @@ def test_reference_api():
     with ir.Context() as context:
       stablehlo.register_dialect(context)
       m = ir.Module.parse(ASM_FORMAT.format(tensor_type))
-      module_str = str(m)
       args = [ir.DenseIntElementsAttr.get(arg)]
 
-    res = stablehlo.eval_module(m, args)
-    actual = np.array(res[0])
+    actual = np.array(stablehlo.eval_module(m, args)[0])
     expected = arg + arg
     assert (actual == expected).all()
 
@@ -273,11 +271,10 @@ def test_serialization_apis():
   with ir.Context() as context:
     stablehlo.register_dialect(context)
     m = ir.Module.parse(ASM_FORMAT.format("2xf32"))
-    module_str = str(m)
     assert m is not None
     serialized = stablehlo.serialize_portable_artifact(m, curr_version)
     deserialized = stablehlo.deserialize_portable_artifact(context, serialized)
-    assert module_str == str(deserialized)
+    assert str(m) == str(deserialized)
 
 @run
 def test_str_serialization_apis():
@@ -291,10 +288,9 @@ def test_str_serialization_apis():
   with ir.Context() as context:
     stablehlo.register_dialect(context)
     m = ir.Module.parse(ASM_FORMAT.format("2xf32"))
-    module_str = str(m)
     assert m is not None
     bytecode = module_to_bytecode(m)
     serialized = stablehlo.serialize_portable_artifact(bytecode, curr_version)
     deserialized = stablehlo.deserialize_portable_artifact(serialized)
     deserialized_module = ir.Module.parse(deserialized)
-    assert module_str == str(deserialized_module)
+    assert str(m) == str(deserialized_module)
--- a/stablehlo/integrations/python/tests/stablehlo.py
+++ b/stablehlo/integrations/python/tests/stablehlo.py
@@ -228,26 +228,43 @@ def test_minimum_version():
   assert is_semver_format(curr_version)
 
 
-ASM = """
-func.func @test(%arg0: tensor<2xi64>) -> tensor<2xi64> {
-  %0 = stablehlo.add %arg0, %arg0 : (tensor<2xi64>, tensor<2xi64>) -> tensor<2xi64>
-  func.return %0 : tensor<2xi64>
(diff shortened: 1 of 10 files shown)


stablehlo/integrations/python/tests/stablehlo.py
```diff
@@ -227,20 +228,51 @@ def test_minimum_version():
...
+  # Program runs arg + arg, which is used for expected value
+  tests = [
+    # No numpy types for f8
+    ("f16", np.asarray(1, np.float16)),
+    ("f32", np.asarray(2, np.float32)),
+    ("f64", np.asarray(3, np.double)),
+    ("1xi8", np.asarray([4], np.int8)),
+    ("1xi16", np.asarray([5], np.int16)),
+    ("1xi32", np.asarray([-6], np.int32)),
+    # Numpy uint types interpreter as int
```
mlevesquedion, line 251: What uint/int types is this comment referring to?

stablehlo/reference/Api.cpp
```diff
@@ -106,33 +117,77 @@ class DefaultInterpreterFallback : public InterpreterFallback {
...
+           << func.getNumArguments();
+
+  TypeRange signature = func.getArgumentTypes();
+  for (int64_t i = 0; i < func.getNumArguments(); ++i) {
+    Type sigType = signature[i];
+    Type argType = inputs[i].getType();
+    if (sigType != argType)
+      return func.emitError() << "invalid input argument type at index " << i
+                              << ", input type was " << argType
+                              << " but entry function expected " << sigType;
```
mlevesquedion, line 135: This is a great error message!

stablehlo/integrations/python/tests/stablehlo.py
GleasonK: The lack of tests for `np.uint8` etc.

mlevesquedion: Gotcha, thanks! I'm having a hard time parsing this comment then. Do you mean numpy uint types are treated as int in the interpreter? If so, IMO it could still be worth having tests for these, even if they are mostly testing the same paths as the signed int tests.

GleasonK: I don't see much value, since numpy is an implementation detail of these tests, we'd be testing upstream MLIR handling of DenseElements + numpy.

JAX defines custom lowerings for these types in: https://github.com/google/jax/blob/46913d7e2ed8e6e7839447813d9f04424d9a05b4/jaxlib/hlo_helpers.py#L28

We could add APIs similar to this, at which point I'd agree we should test them, but until then I don't see much value in testing the raw behavior of `DenseElementsAttr.get`

GleasonK: Updated the code comments to clarify this

ghpvnist (changes requested):
Mostly lgtm, few minor comments.

stablehlo/integrations/python/tests/stablehlo.py
```diff
@@ -227,20 +228,51 @@ def test_minimum_version():
...
+}}
 """
 
+
+@run
+def test_reference_api():
+  # Formatted as (tensor_type, np_value)
+  # Program runs arg + arg, which is used for expected value
+  tests = [
+    # No numpy types for f8 - skipping fp8 tests
```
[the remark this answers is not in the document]
GleasonK: Created https://github.com/openxla/stablehlo/issues/2093

None of these types map particularly cleanly from numpy, will require some more manual test cases.

ghpvnist (approved):

stablehlo/integrations/python/tests/stablehlo.py
```diff
@@ -227,20 +228,51 @@ def test_minimum_version():
...
+    ("1xi32", np.asarray([-6], np.int32)),
+    # Numpy's uint treated as int by DenseElementsAttr, skipping np.uint tests
+    ("2x2xf16", np.asarray([1, 2, 3, 4], np.float16).reshape(2,2)),
+    ("2x1x2xf16", np.asarray([1, 2, 3, 4], np.float16).reshape(2,1,2)),
+  ]
+  for test in tests:
+    tensor_type, arg = test
+    with ir.Context() as context:
+      stablehlo.register_dialect(context)
+      m = ir.Module.parse(ASM_FORMAT.format(tensor_type))
```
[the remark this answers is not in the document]
GleasonK: serialization is actually a module transformation so we need to take the string before serializing

PR #2064, opened 2024-03-04
Add shape refinement pass for DotOp

ghpvnist:
`DotOp` currently doesn't have shape refinement, so if downstream exports stablehlo with dot, and it contains unbounded dynamic shapes with `stablehlo.dot`, it fails to refine shape.

This change enables refinement, and prevents users from having to transform `dot` back to `dot_general` whenever canonicalization pass (for `dot_general` to `dot`) is run implicitly.

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -1884,40 +1884,32 @@ LogicalResult inferCreateTokenOp(HloDialectInterface* dialect,
 }
 
 LogicalResult inferDotOp(
-    std::optional<Location> location, Type lhsType, Type rhsType,
-    std::optional<ArrayAttr> precisionConfig,
+    std::optional<Location> location, RankedTensorType lhsType,
+    RankedTensorType rhsType, std::optional<ArrayAttr> precisionConfig,
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
   if (failed(verifyPrecisionConfig(location, precisionConfig)))
     return failure();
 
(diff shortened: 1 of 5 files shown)


stablehlo/dialect/TypeInference.cpp
```diff
@@ -1884,39 +1884,40 @@ LogicalResult inferCreateTokenOp(HloDialectInterface* dialect,
...
     std::optional<ArrayAttr> precisionConfig,
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
   if (failed(verifyPrecisionConfig(location, precisionConfig)))
     return failure();
 
-  auto lhsType = lhs.getType().dyn_cast<RankedTensorType>();
-  auto rhsType = rhs.getType().dyn_cast<RankedTensorType>();
-  if (!lhsType || !rhsType) {
+  auto lhsRankedType = lhsType.dyn_cast<RankedTensorType>();
+  auto rhsRankedType = rhsType.dyn_cast<RankedTensorType>();
```
GleasonK, line 1894:
```
::mlir::TypedValue<::mlir::RankedTensorType> getLhs();
::mlir::TypedValue<::mlir::RankedTensorType> getRhs();
```

Let's update this API to take `RankedTensorType`. Currently we're casting `RankedTensorType --> Type --(dyn_cast)--> RankedTensorType` and error checking.

GleasonK: Also lets revert the name to `lhsType` so that the rest of the code can stay the same

GleasonK (approved):
LGTM after addressing the one comment

PR #2069, opened 2024-03-05
Make type error checks in tests less brittle

mlevesquedion:
These error messages are automatically generated via ODS constraints so we shouldn't test the entire message. In particular, I don't think it is useful to test the part of the message that lists the possible element types a given tensor might have. This exposes us to breakages if we change the list of supported types in any way, e.g. if we add a type that wasn't supported before.

Use regular expressions instead.

Relevant documentation: https://mlir.llvm.org/docs/Diagnostics/\#sourcemgr-diagnostic-verifier-handler

### DIFF
--- a/stablehlo/tests/ops_stablehlo.mlir
+++ b/stablehlo/tests/ops_stablehlo.mlir
@@ -2038,7 +2038,7 @@ func.func @rng_normal_invalid_shape(%arg0: tensor<f32>, %arg1: tensor<f32>) {
 
 func.func @rng_normal_invalid_mu_rank(%mu: tensor<1xf32>, %sigma: tensor<f32>) -> tensor<2x3x5xf32> {
   %shape = stablehlo.constant dense<[2, 3, 5]> : tensor<3xi64>
-  // expected-error@+1 {{#0 must be 0D tensor of pred (AKA boolean or 1-bit integer) or 4/8/16/32/64-bit signless integer or 4/8/16/32/64-bit unsigned integer or f8E4M3B11FNUZ type or f8E4M3FN type or f8E4M3FNUZ type or f8E5M2 type or f8E5M2FNUZ type or 16-bit float or 32-bit float or 64-bit float or bfloat16 type values, but got 'tensor<1xf32>'}}
+  // expected-error-re@+1 {{operand #0 must be 0D tensor of {{.*}}, but got 'tensor<1xf32>'}}
   %0 = "stablehlo.rng"(%mu, %sigma, %shape) {rng_distribution = #stablehlo<rng_distribution NORMAL>}: (tensor<1xf32>, tensor<f32>, tensor<3xi64>) -> tensor<2x3x5xf32>
   func.return %0 : tensor<2x3x5xf32>
 }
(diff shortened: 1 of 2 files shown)


sdasgup3:
I am afraid that I have a mixed feeling here. Consider the case if we delete a type that was added before. We want tests to fail. I agree with you point about maintenance when a new type is added. Given how the error string is laid out it might not be difficult to batch replace to updated string.

In any case, feel free to seek for additional opinion.

mlevesquedion:
> I am afraid that I have a mixed feeling here. Consider the case if we delete a type that was added before. We want tests to fail. I agree with you point about maintenance when a new type is added. Given how the error string is laid out it might not be difficult to batch replace to updated string.
> 
> In any case, feel free to seek for additional opinion.

Thanks for taking the time to review!

> Consider the case if we delete a type that was added before. We want tests to fail.

Do we? Let's say we have a type `def MyTensor : RankedTensorOf<[A, B]>` and a test that uses `stablehlo.my_op %0 : tensor<C>`, with an `expected-error`. Now let's say I remove the type `B` from `MyTensor` such that only `A` is supported. `C` is still not supported, so the test should still fail, but now we have to edit the error message, even though the relevant behavior (whether that type is supported or not) didn't change.

IMO, we shouldn't be testing the error message, because that amounts to testing ODS. I agree that batch replacing should not be that difficult, but in my opinion we shouldn't be testing that in the first place anyway. In fact, I'm not sure it's helpful to test negative cases for type specifications, because that is once again just testing ODS. When I removed support for unranked tensors in #2045, I didn't add test cases expecting errors when unranked tensors are provided to each op, I just deleted the tests that used unranked tensors.

GleasonK:
> We want tests to fail.

I think my guy reaction was the same as Sandeep's, when I thought about it a bit more I think I've changed my mind - really we want to have _test cases_ fail, not _error messages_ fail.

### Remove support:

If we ended up removing unranked tensors, we should have a test case called `op_with_unranked_tensors` which used to pass and now fails. As opposed to having 15 tests for ranked tensor errors which now fail because the error message no longer has "or unranked tensor" in it.

### Add support

If we add per-axis support for `dynamic_broadcast_in_dim`, we should _add_ a new test that ensures that `dynamic_broadcast_in_dim` supports this feature. Not add "or per-aixis" to the existing dyanmic_broadcast_in_dim error checks.

In both these cases, the value of updating these other test case messages is low IMO.

abhigunj:
> > We want tests to fail.
> 
> I think my gut reaction was the same as Sandeep's, when I thought about it a bit more I think I've changed my mind - really we want to have _test cases_ fail, not _error messages_ fail.
> 
Agree with we want to have _test cases_ fail, not _error messages_ fail.

> Consider the following two cases where these errors would be updated:
> 
> ### Add support
> If we add per-axis support for `dynamic_broadcast_in_dim`, we should add a new test that ensures that `dynamic_broadcast_in_dim` supports this feature. Not add "or per-aixis" to the existing dyanmic_broadcast_in_dim error checks.
> 
Quantization test are of three types;
positive_per_tensor
positive_per_axis
negative_per_axis

my expectation here was, add positive_per_axis and remove negative_per_axis test when `dynamic_broadcast_in_dim` OP supports per-axis quantization. Eventually once all the OPs support per-axis quantization there won't be any negative tests at ops_stablehlo_quantized.mlir. 


> ### Remove support:
> If we ended up removing unranked tensors, we should have a test case called `op_with_unranked_tensors` which used to pass and now fails. As opposed to having 15 tests for ranked tensor errors which now fail because the error message no longer has "or unranked tensor" in it.
> 
> That said.. I don't think we test all ops with all types, and the error message will let us know if we ever unintentionally change the supported types of an op. Perhaps we keep 1 of these around for each op?

as explained above, we can't promise to  "Perhaps we keep 1 of these around for each op?"   Not all StableHLO OPs has negative test that covers all types. That is also one reason for quantization negative tests I matched with all types vs using regex to match with quantized types only.  Once we start enabling per-axis feature for more OPs, there will be fewer negative tests.

It is still unclear to me, what is general guidance for tests that test  ODS. There are negative tests that tests custom logic such as Trait verifier and we definitely need those.

mlevesquedion:
Thanks for the great discussion, folks!

> I don't think we test all ops with all types, and the error message will let us know if we ever unintentionally change the supported types of an op. Perhaps we keep 1 of these around for each op?

mlevesquedion:
We didn't reach a consensus and this is low priority, so I will close the pull request. Thanks for the discussion, everyone!

PR #2071, opened 2024-03-06
Clean up aggressive simplification and add tests

mlevesquedion:
I wanted to read this pass to familiarize myself with the code. In passing, I made a few readability improvements here and there and while making these I realized that we did not import the tests for this code. Import the tests so we can have some confidence when making changes.

The tests are from https://github.com/openxla/iree/blob/7782a414ea473c59f6d7a882cb510690ed666c79/compiler/plugins/input/StableHLO/stablehlo-iree/Conversion/Preprocessing/test/canonicalization.mlir, with minor tweaks (formatting, reordering tests to match the implementation, and adding missing test separators).

### DIFF
--- a/stablehlo/transforms/StablehloAggressiveSimplification.cpp
+++ b/stablehlo/transforms/StablehloAggressiveSimplification.cpp
@@ -41,13 +41,9 @@ namespace {
 constexpr int64_t kFoldOpEltLimit = 65536;
 
 static bool isIotaRange(ArrayRef<int64_t> dims) {
-  for (auto [idx, value] : llvm::enumerate(dims)) {
-    if (static_cast<int64_t>(idx) != value) {
-      return false;
-    }
-  }
-
-  return true;
+  return llvm::all_of(llvm::enumerate(dims), [](const auto &it) {
+    return static_cast<int64_t>(it.index()) == it.value();
+  });
 }
 
 /// Matches when either of the submatchers match.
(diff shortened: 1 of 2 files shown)


stablehlo/transforms/StablehloAggressiveSimplification.cpp
```diff
@@ -70,28 +66,19 @@ template <typename Fn>
...
-                  return folder(lhs, rhs);
-                })) {
-      return cast<TypedAttr>(res);
-    }
-    return nullptr;
+    res = constFoldBinaryOp<FloatAttr, FloatAttr::ValueType, void>(operands,
+                                                                   folder);
+  }
+  if (res) {
+    return cast<TypedAttr>(res);
```
GleasonK, line 81: Does `res.cast<TypedAttr>();` work? I always forget which support casting as a method

stablehlo/transforms/StablehloAggressiveSimplification.cpp
```diff
@@ -70,28 +66,19 @@ template <typename Fn>
...
-                operands, [&folder](const APFloat &lhs, const APFloat &rhs) {
-                  return folder(lhs, rhs);
-                })) {
-      return cast<TypedAttr>(res);
-    }
-    return nullptr;
+    res = constFoldBinaryOp<FloatAttr, FloatAttr::ValueType, void>(operands,
+                                                                   folder);
+  }
+  if (res) {
```
GleasonK, line 80: LLVM style guide allows removal of braces from single line if's, feel free to do for all these

stablehlo/transforms/StablehloAggressiveSimplification.cpp
GleasonK: revisit other uses of `cast` as well.

stablehlo/transforms/StablehloAggressiveSimplification.cpp
```diff
@@ -245,17 +226,16 @@ static mlir::stablehlo::ComparisonDirection invertDirection(
...
-      return ComparisonDirection::GE;
     case ComparisonDirection::GT:
       return ComparisonDirection::LT;
+    case ComparisonDirection::LE:
+      return ComparisonDirection::GE;
     case ComparisonDirection::LT:
       return ComparisonDirection::GT;
-    case ComparisonDirection::NE:
-      return ComparisonDirection::NE;
   }
```
GleasonK, line 240: Prefer `llvm::report_fatal_error` to `unreachable`, since unreachable may be removed in release builds I believe?

GleasonK: See: https://discourse.llvm.org/t/llvm-unreachable-is-widely-misused/60587/21

stablehlo/transforms/StablehloAggressiveSimplification.cpp
```diff
@@ -316,8 +281,7 @@ struct CompareOpCanon final : OpRewritePattern<mlir::stablehlo::CompareOp> {
 
   LogicalResult matchAndRewrite(mlir::stablehlo::CompareOp op,
                                 PatternRewriter &rewriter) const override {
-    auto type = dyn_cast<RankedTensorType>(op.getType());
-    if (!type) return failure();
+    auto type = cast<RankedTensorType>(op.getType());
```
GleasonK, line 284: Do we even need to cast this anymore?

Does `RankedTensorType type = op.getType()` work?

GleasonK: Same question for similar code cases

GleasonK (approved):
Nice! Any idea how these compare to the `Eval*` patterns in shape refinement / canonicalize dynamism? I wonder if we can expose a method here called `addConstantEvalPatterns` which includes any folders.

mlevesquedion:
> Nice! Any idea how these compare to the `Eval*` patterns in shape refinement / canonicalize dynamism? I wonder if we can expose a method here called `addConstantEvalPatterns` which includes any folders.
> 
> For additional context - we don't want to canonicalize everything when refining shapes, only certain things, hence the subset of existing patterns would be useful

Thanks! It looks like there's some overlap, e.g. the `add` folder is very similar to `evalAddOp`. I think we should consider sharing implementation in a future change.

stablehlo/transforms/StablehloAggressiveSimplification.cpp
mlevesquedion: Good point. Applied, here and elsewhere.

stablehlo/transforms/StablehloAggressiveSimplification.cpp
mlevesquedion: Thanks for sharing the link. Changed to `report_fatal_error`.

PR #2073, opened 2024-03-06
Implement composite inlining pass

mlevesquedion:


stablehlo/tests/interpret/composite.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: stablehlo-translate --interpret -split-input-file %s
+// RUN: stablehlo-translate --interpret %s
```
GleasonK: Can we interpret the inlined composite as well? I'm forgetting if the last PR added call op support:

```
// RUN: stablehlo-translate --interpret %s
// RUN: stablehlo-translate --stablehlo-replace-composites-with-calls --interpret %s
```

stablehlo/transforms/Passes.td
```diff
@@ -144,3 +144,40 @@ def ShapeLegalizeToStablehloPass : Pass<"shape-legalize-to-stablehlo", "func::Fu
   }];
   let dependentDialects = ["mlir::stablehlo::StablehloDialect"];
 }
+
+def StablehloReplaceCompositesWithCallsPass :
+    Pass<"stablehlo-replace-composites-with-calls", "func::FuncOp"> {
```
GleasonK, line 149: Brainstorming pass names:

```
--stablehlo-inline-composites
--stablehlo-decompose-composite
--stablehlo-use-composite-decompositions
--stablehlo-composite-use-decompositions
```

I'm OK with `inline-composites` even if its a bit of a misnomer. This current name is a bit verbose / implementation-detail-y. I.e. `replace-composites-with-call-to-decomposition` should be the true verbose name. I think `use-decomposition` may be more accurate?

stablehlo/transforms/StablehloReplaceCompositesWithCalls.cpp
```diff
@@ -0,0 +1,89 @@
...
+  using OpRewritePattern::OpRewritePattern;
+
+  ReplaceCompositeWithCall(MLIRContext *context,
+                           const DenseSet<StringRef> &excludedNames_)
+      : OpRewritePattern<mlir::stablehlo::CompositeOp>(context),
+        excludedNames(excludedNames_) {}
+
+  LogicalResult matchAndRewrite(CompositeOp op,
+                                PatternRewriter &rewriter) const override {
+    if (excludedNames.contains(op.getName()))
```
GleasonK, line 51: Good for a V1. I wouldn't be surprised if people wanted wildcards or `except-namespace=my_namespace` as well.

stablehlo/transforms/StablehloReplaceCompositesWithCalls.cpp
```diff
@@ -0,0 +1,89 @@
...
+struct StablehloInlineCompositesPass
+    : public impl::StablehloReplaceCompositesWithCallsPassBase<
+          StablehloInlineCompositesPass> {
+  using StablehloReplaceCompositesWithCallsPassBase::
+      StablehloReplaceCompositesWithCallsPassBase;
+
+  void runOnOperation() override {
+    MLIRContext *context = &getContext();
+    RewritePatternSet patterns(context);
+    auto excludedNames =
```
GleasonK, line 75: `DenseSet<StringRef> excludedNames(exceptListOption.begin(), exceptListOption.end());` ?

GleasonK: Use of auto here seems a bit non-standard here?

stablehlo/transforms/StablehloReplaceCompositesWithCalls.cpp
```diff
@@ -0,0 +1,89 @@
...
+        DenseSet<StringRef>(exceptListOption.begin(), exceptListOption.end());
+    patterns.add<ReplaceCompositeWithCall>(context, excludedNames);
+    if (failed(applyPatternsAndFoldGreedily(getOperation(),
```
GleasonK, line 78: Can we use `DialectLegalization` instead? `applyPartialConversion`? This will do constant folding and more things that may modify the rest of the program I believe. Not sure we want to bundle that into a simple `composite->call` pass. Can use `addDynamicallyLegalOp` to test if the op is in the exclude list, then we don't need that check in the pattern itself :)

GleasonK (changes requested):

stablehlo/transforms/StablehloReplaceCompositesWithCalls.cpp
GleasonK: Also I think the Legal/Illegal targets should be:

Legal: StableHLO, FuncOp
DynamicallyIllegal: CompositeOp

stablehlo/transforms/Passes.td
mlevesquedion: Thanks for taking the time to brainstorm names! I think `inline` and `decompose` are both misnomers, although maybe not majorly so. But as a user, if I called a pass called "inline" and it didn't actually inline the thing, I'd be confused.

I don't really understand what `use` means in this context, so e.g. `use-composite-decompositions` is a bit confusing to me.

I agree the current name might be too specific, but it's accurate. I like `replace-composites-with-call-to-decomposition` though.

So, I don't know. I would personally go with `replace-composites-with-call-to-decomposition`. WDYT?

stablehlo/tests/interpret/composite.mlir
mlevesquedion: I ended up not adding it to keep the PR more focused. I'm happy to add it, but I'm not sure if that test really adds much. Feels like testing MLIR at that point?

stablehlo/transforms/StablehloReplaceCompositesWithCalls.cpp
mlevesquedion: Agreed, let's see what the users want.

stablehlo/transforms/Passes.td
GleasonK: `--stablehlo-composite-call-decompositions` ?

mlevesquedion: Discussed offline. We will go with `stablehlo-legalize-composite-to-call`, to be consistent with existing passes with names like `mhlo-legalize-X-to-Y`, e.g. `mhlo-legalize-dot-to-dot-general`.

mlevesquedion:
Thanks for the review! I addressed your comments, except for one I'm not sure about: https://github.com/openxla/stablehlo/pull/2073#discussion_r1515023008. PTAL

stablehlo/tests/interpret/composite.mlir
mlevesquedion: Discussed offline. I ended up implementing this.

mlevesquedion:
Ended up adding a test that inlining doesn't change semantics. PTAL

stablehlo/tools/StablehloOptMain.cpp
```diff
@@ -38,6 +39,7 @@ int main(int argc, char **argv) {
   mlir::registerAllDialects(registry);
   mlir::registerAllExtensions(registry);
   mlir::stablehlo::registerAllDialects(registry);
+  registry.insert<mlir::stablehlo::check::CheckDialect>();
```
GleasonK, line 42: I'm even more OK with this now that I see that `opt` has been able to handle InterpreterDialect this entire time. Check is effectively a feature of the interpreter currently.

BUILD.bazel
```diff
@@ -1132,6 +1133,7 @@ cc_binary(
         ":register",
         ":stablehlo_passes",
         ":tosa_passes",
+        "//stablehlo/tests:check_ops",
```
GleasonK, line 1136: Does CMakeList in `tools/` need any updates?

stablehlo/transforms/StablehloLegalizeCompositeToCall.cpp
```diff
@@ -0,0 +1,91 @@
...
+#include "mlir/Dialect/Tensor/IR/Tensor.h"
+#include "mlir/IR/Attributes.h"
+#include "mlir/IR/BuiltinAttributeInterfaces.h"
+#include "mlir/IR/BuiltinAttributes.h"
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/Matchers.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/IR/TypeUtilities.h"
+#include "mlir/Interfaces/FunctionInterfaces.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
```
GleasonK, line 28: Don't need

GleasonK (approved):

stablehlo/reference/Ops.cpp
```diff
@@ -111,6 +111,20 @@ Tensor evalSliceOp(const Tensor &operand, const Sizes &startIndices,
...
+                                         Process *process, Operation *op,
+                                         StringRef funcName) {
+  SymbolTableCollection symbolTableCollection;
+  auto symbolTable =
+      symbolTableCollection.getSymbolTable(op->getParentOfType<ModuleOp>());
+  auto func = symbolTable.lookupNearestSymbolFrom<func::FuncOp>(
+      op, StringAttr::get(op->getContext(), funcName));
+  SmallVector<InterpreterValue> values = llvm::to_vector(llvm::map_range(
+      inputs, [](const Tensor &t) { return InterpreterValue(t); }));
+  return eval(func.getBody(), values, fallback, process, nullptr);
```
GleasonK, line 125: Actually wait - do we need any code which adds the result to the current scope?

stablehlo/reference/Ops.cpp
```diff
@@ -386,15 +405,8 @@ SmallVector<InterpreterValue> eval(Region &region,
...
-      SmallVector<InterpreterValue> values = llvm::to_vector(llvm::map_range(
-          operands, [](const Tensor &t) { return InterpreterValue(t); }));
-      SymbolTableCollection symbolTableCollection;
-      auto symbolTable =
-          symbolTableCollection.getSymbolTable(op.getParentOfType<ModuleOp>());
-      auto func = symbolTable.lookupNearestSymbolFrom<func::FuncOp>(
-          &op,
-          StringAttr::get(op.getContext(), compositeOp.getDecomposition()));
-      auto results = eval(func.getBody(), values, fallback, process, nullptr);
+      auto results = evalCallOp(operands, fallback, process, &op,
```
GleasonK, line 408: Would be cool in the future to enhance `fallback` to be able to specify supported composites, then this function would actually behave like the impl to composite -- is there special handling? Use it. Else call decomp.

stablehlo/reference/Ops.cpp
GleasonK: Let's add a single test for this.

BUILD.bazel
mlevesquedion: Weirdly enough, I guess not, since everything builds and runs fine. But I added check ops just in case.

stablehlo/reference/Ops.cpp
mlevesquedion: Ooh yeah that's a cool idea! I'll keep that in mind.

stablehlo/transforms/StablehloLegalizeCompositeToCall.cpp
mlevesquedion: Cleaned up the imports. There was a lot of cruft.

stablehlo/reference/Ops.cpp
mlevesquedion: Discussed offline. Getting the inputs and putting the results in the scope is handled in `eval`.

PR #2077, opened 2024-03-06
Fix stablehlo_legalize_to_vhlo.0_19_0.mlir bytecode

mlevesquedion:
Also, use same header in all tests, and avoid command substitutions since they don't work on Windows: https://github.com/openxla/stablehlo/pull/1764

### Why does `0_19_0` have a different header? 

I followed the instructions in https://github.com/openxla/stablehlo/blob/main/docs/vhlo.md, except for the part that says `Replace RUN commands`. That is, I copied `stablehlo_legalize_to_vhlo.mlir`, but I forgot to update the RUN commands to match the other versioned tests. IMO, we should avoid duplicating the RUN commands across tests and documentation, better to just instruct developers to copy the latest versioned test and add their tests in there.

### Why was the bytecode incorrect?

I serialized the bytecode a while ago. During review, I ended up changing the order of some attributes, which caused the bytecode to be invalid (e.g. the `name` attribute showed up where `composite_attributes` should be).


stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.0_10_0.mlir
```diff
@@ -1,10 +1,11 @@
 // RUN: stablehlo-opt --mlir-print-op-generic %s.bc | FileCheck %s
 // RUN: stablehlo-translate --deserialize %s.bc | stablehlo-translate --serialize --target=0.10.0 | stablehlo-opt --mlir-print-op-generic | FileCheck %s
-// RUN: stablehlo-translate --deserialize %s.bc | stablehlo-opt > %t.0
-// RUN: stablehlo-opt --strip-debuginfo %s > %t.1
-// RUN: diff %t.0 %t.1
-// RUN: stablehlo-translate --serialize --target=0.10.0 --strip-debuginfo %s > %t.2
-// RUN: diff %s.bc %t.2
+// RUN: diff <(stablehlo-translate --deserialize %s.bc | stablehlo-opt) <(stablehlo-opt --strip-debuginfo %s)
```
GleasonK, line 3: https://github.com/openxla/stablehlo/pull/1764

Can we avoid this refactoring?

mlevesquedion: Oh, good to know! Yeah I changed it back to use temp files.

stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.0_10_0.mlir
```diff
@@ -1,10 +1,11 @@
...
+// RUN: diff %s.bc <(stablehlo-translate --serialize --target=0.10.0 --strip-debuginfo %s)
+// RUN: stablehlo-opt --stablehlo-legalize-to-vhlo -emit-bytecode -debug-only=vhlo-bytecode %s 2>&1 | FileCheck --check-prefix=CHECK-WARN %s
+// RUN: stablehlo-opt --stablehlo-legalize-to-vhlo -emit-bytecode %s | stablehlo-opt -debug-only=vhlo-bytecode 2>&1 | FileCheck --check-prefix=CHECK-WARN %s
```
sdasgup3, line 9: Not sure I get this: what are we checking with the log output?

mlevesquedion: Making sure we don't hit `Not Implemented` while reading/writing the bytecode.

mlevesquedion: See `stablehlo_legalize_to_vhlo.mlir` for reference. Some of the versioned tests have the `CHECK-WARN-NOT`, but they don't actually run any commands that might check for these. IMO we should check for these in all of the tests.

sdasgup3 (approved):

stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.0_10_0.mlir
```diff
@@ -5,6 +5,10 @@
 // RUN: diff %t.0 %t.1
 // RUN: stablehlo-translate --serialize --target=0.10.0 --strip-debuginfo %s > %t.2
 // RUN: diff %s.bc %t.2
+// RUN: stablehlo-opt --stablehlo-legalize-to-vhlo -emit-bytecode -debug-only=vhlo-bytecode %s 2>&1 | FileCheck --check-prefix=CHECK-WARN %s
+// RUN: stablehlo-opt --stablehlo-legalize-to-vhlo -emit-bytecode %s | stablehlo-opt -debug-only=vhlo-bytecode 2>&1 | FileCheck --check-prefix=CHECK-WARN %s
```
GleasonK, line 9: Why not:

`// RUN: stablehlo-opt -debug-only=vhlo-bytecode %s 2>&1 | FileCheck --check-prefix=CHECK-WARN %s`

stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.0_10_0.mlir
GleasonK: +1. I think I reviewed this initial change, bit didn't fully understand how filecheck worked at that time. I think we do need this.

stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.0_10_0.mlir
```diff
@@ -5,6 +5,10 @@
 // RUN: diff %t.0 %t.1
 // RUN: stablehlo-translate --serialize --target=0.10.0 --strip-debuginfo %s > %t.2
 // RUN: diff %s.bc %t.2
+// RUN: stablehlo-opt --stablehlo-legalize-to-vhlo -emit-bytecode -debug-only=vhlo-bytecode %s 2>&1 | FileCheck --check-prefix=CHECK-WARN %s
```
GleasonK, line 8: Also can you update `vhlo.md` with the new test header? 

https://github.com/openxla/stablehlo/blob/main/docs/vhlo.md#add-versioned-serialization-test

stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.0_10_0.mlir
mlevesquedion: Sure, we can do that too. I grabbed these from `stablehlo_legalize_to_vhlo.mlir`.

stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.0_10_0.mlir
mlevesquedion: Done.

mlevesquedion: Actually, reading these instructions again I remember what happened now. I followed the instructions in the doc you linked, but I forgot to do the manual part at the end. IMO we should advise to copy the highest versioned test and add tests to it, I'll update the doc.

stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.0_10_0.mlir
mlevesquedion: Actually IMO that doesn't belong here since this is a test for VHLO legalization. But we should test this somewhere, I'll send a PR to test this in `ops_stablehlo`.

docs/vhlo.md
```diff
@@ -133,33 +133,28 @@ negative unit tests of the feature, as well as compatibility unit tests.
...
-$ build/bin/stablehlo-translate --serialize --target=$TARGET_VERSION --strip-debuginfo stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.$TARGET_FILENAME.mlir > stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.$TARGET_FILENAME.mlir.bc
-
-# Replace RUN commands in stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.0_X_0.mlir with the following for 0.X.0:
-// RUN: stablehlo-opt --mlir-print-op-generic %s.bc | FileCheck %s
-// RUN: stablehlo-translate --deserialize %s.bc | stablehlo-translate --serialize --target=0.X.0 | stablehlo-opt --mlir-print-op-generic | FileCheck %s
-// RUN: diff <(stablehlo-translate --deserialize %s.bc | stablehlo-opt) <(stablehlo-opt --strip-debuginfo %s)
-// RUN: diff %s.bc <(stablehlo-translate --serialize --target=0.X.0 --strip-debuginfo %s)
+export TARGET_VERSION=0.X.0
+export TARGET_FILENAME=${TARGET_VERSION//./_}
+stablehlo-translate --serialize --target=$TARGET_VERSION --strip-debuginfo stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.$TARGET_FILENAME.mlir > stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.$TARGET_FILENAME.mlir.bc
```
sdasgup3, line 156: I can see you have a prose at the top, but the bash instruction are useful as well (like `cp` instruction) if we are specifying the whole pipeline.

mlevesquedion: IMO `cp` is simple enough to not warrant a bash example. My fear is that if we have a script that "seems" complete, developers may apply the script blindly and miss some steps (like I did :upside_down_face:). To be honest I would personally be in favor of removing any bash examples, as it is easy to figure out how to generate the bytecode by looking at the tests. WDYT?

docs/vhlo.md
```diff
@@ -133,33 +133,28 @@ negative unit tests of the feature, as well as compatibility unit tests.
...
+`vhlo_to_version_downgrade_invalid.0_Y_0.mlir` that shows the op is unsupported
+before version `X`. If adding a new version of an op, add a test file like
+`vhlo_to_version_downgrade.0_Y_0.mlir` that shows that the op can be downgraded
+successfully.
 
-A few examples:
-
-* Backward compatibility, positive tests: [vhlo_to_version_upgrade.mlir](https://github.com/openxla/stablehlo/blob/6886b59f6cd4369674e7e3beff61301c145176e2/stablehlo/tests/vhlo_to_version_upgrade.mlir#L2)
-* Forward compatibility, positive tests: [vhlo_to_version_downgrade.mlir](https://github.com/openxla/stablehlo/blob/6886b59f6cd4369674e7e3beff61301c145176e2/stablehlo/tests/vhlo_to_version_downgrade.mlir#L1)
-* Forward compatibility, negative tests: [vhlo_to_version_downgrade_invalid.0_9_0.mlir](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/vhlo/vhlo_to_version_downgrade_invalid.0_9_0.mlir)
```
GleasonK, line 142: Should we keep these around? Or update to point at a recent invalid downgrade test?

docs/vhlo.md
sdasgup3: I am on board with that.

docs/vhlo.md
mlevesquedion: Added a pointer to a recent test. Thanks!

stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.0_10_0.mlir
mlevesquedion: Follow-up: go/shlopr/2080

PR #2079, opened 2024-03-07
Updates to ConvolutionOP verifier to support quantization constraints

abhigunj:


stablehlo/dialect/TypeInference.cpp
```diff
@@ -3453,6 +3489,59 @@ LogicalResult verifyConvolutionOp(
...
+  // convolution_c32
+  if (rhsQPAType &&
+      rhsQPAType.getQuantizedDimension() != kernelOutputFeatureDimension)
+    return emitOptionalError(
+        location, "mismatched kernel_output_feature_dimension ",
+        kernelOutputFeatureDimension, " and operand quantized dimension ",
+        rhsQPAType.getQuantizedDimension(), "");
+  // convolution_c33
+  if (resultQPAType &&
+      resultQPAType.getQuantizedDimension() != kernelOutputFeatureDimension)
```
sdasgup3, line 3539: `output_feature_dimension` per the spec
```
(C33) If is_per_axis_quantized(result), then quantization_dimension(result) = output_feature_dimension.

```

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3453,6 +3489,59 @@ LogicalResult verifyConvolutionOp(
...
+  auto expressedType = lhsQType.getExpressedType();
+  if (expressedType != rhsQType.getExpressedType() ||
+      expressedType != resultQType.getExpressedType())
+    return emitOptionalError(location,
+                             "mismatched operands and result expressed types");
+
+  llvm::SmallVector<Type, 2> typeEntriesPerAxis{rhsType, resultType};
+  if (nonePerAxisQuantized(typeEntriesPerAxis)) return success();
+  // convolution_c31
+  if (!allPerAxisQuantized(typeEntriesPerAxis)) {
```
sdasgup3, line 3521: Why is this an error? we can have `lhs` per-tensor and `rhs` per-axis, right?

sdasgup3: (C31) says if rhs is per-axis then result must be per-axis.

abhigunj: we are not checking lhs here. typeEntriesPerAxis has {rhs, result} only

stablehlo/dialect/TypeInference.cpp
abhigunj: thanks, updated.

sdasgup3 (approved):
Suggested a few nits. Otherwise LGTM.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3453,6 +3472,60 @@ LogicalResult verifyConvolutionOp(
...
+      getElementTypeOrSelf(lhsType).dyn_cast<quant::QuantizedType>();
+  auto rhsQType =
+      getElementTypeOrSelf(rhsType).dyn_cast<quant::QuantizedType>();
+  auto resultQType =
+      getElementTypeOrSelf(resultType).dyn_cast<quant::QuantizedType>();
+  // convolution_c29
+  if (lhsQType.getStorageType() != rhsQType.getStorageType())
+    return emitOptionalError(location, "mismatched operand storage types ",
+                             lhsQType.getStorageType(), " and ",
+                             rhsQType.getStorageType(), "");
```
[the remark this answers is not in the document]
abhigunj: not recall why I had to add this. Removed now. Thank you.

PR #2081, opened 2024-03-08
Quantization Verifiers : Add/update custom verifiers for quantized constraints

abhigunj:
This PR add/update verifiers for ops to verify quantization specific constraints

OPs:  `bitcast_convert` `broadcast_in_dim` `transpose` `reshape`  `sort` `select` `abs`

~`dot_general`~  dropping dot_general verifier changes due to expected internal test failures similar to https://github.com/openxla/stablehlo/pull/2122

### DIFF
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -2852,7 +2852,7 @@ def StableHLO_TraceOp: StableHLO_Op<"trace"> {
 }
 
 def StableHLO_TransposeOp: StableHLO_ShapedInterfaceOp<"transpose",
-      [Pure, HLO_CompatibleOperandsAndResultElementType,
+      [Pure, HLO_CompatibleOperandsAndResultElementType, /*transpose_c1*/
       DeclareOpInterfaceMethods<InferTypeOpInterface>]> {
   let summary = "Transpose operation";
   let description = [{
@@ -2871,7 +2871,9 @@ def StableHLO_TransposeOp: StableHLO_ShapedInterfaceOp<"transpose",
     HLO_TensorOrPerAxisQuantizedTensor:$operand,
     DenseI64ArrayAttr:$permutation
   );
-  let results = (outs HLO_TensorOrPerAxisQuantizedTensor);
+  let results = (outs HLO_TensorOrPerAxisQuantizedTensor:$result);
+
+  let hasVerifier = 1;
 
   let assemblyFormat = [{
     $operand `,` `dims` `=` $permutation
@@ -3009,7 +3011,7 @@ def StableHLO_ReturnOp : StableHLO_Op<"return", [Pure, Terminator]> {
   }];
 
   let arguments = (ins
-    Variadic<HLO_TensorOrTokenOrTuple >:$results
+    Variadic<HLO_TensorOrPerAxisQuantizedTensorOrToken>:$results
   );
 
   let assemblyFormat = "$results attr-dict (`:` type($results)^)?";
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -76,6 +76,12 @@ bool allQuantized(ArrayRef<Type> typeRange) {
 }
 
 template <typename T>
+bool allQuantized(Type tp1, Type tp2) {
+  llvm::SmallVector<Type, 2> typeEntries{tp1, tp2};
+  return allQuantized<T>(typeEntries);
+}
+
+template <typename T>
 bool noneQuantized(ArrayRef<Type> typeRange) {
   return llvm::all_of(
       typeRange, [&](Type val) { return !getElementTypeOrSelf(val).isa<T>(); });
@@ -83,6 +89,13 @@ bool noneQuantized(ArrayRef<Type> typeRange) {
 
 template <typename T>
 bool anyQuantized(ArrayRef<Type> typeRange) {
+  return llvm::any_of(
+      typeRange, [&](Type val) { return getElementTypeOrSelf(val).isa<T>(); });
+}
+
+template <typename T>
+bool anyQuantized(Type tp1, Type tp2) {
+  llvm::SmallVector<Type, 2> typeRange{tp1, tp2};
   return llvm::any_of(
       typeRange, [&](Type val) { return getElementTypeOrSelf(val).isa<T>(); });
 }
@@ -151,6 +164,24 @@ LogicalResult verifyBinaryOpQuantizationConstraints(
                              "result per_axis quantized but none from rhs "
                              "and lhs are per_axis quantized");
   return success();
+}
+
+bool isSameQuantPerAxisScaleZeroPoint(Type ty1, Type ty2) {
+  auto qty1 =
+      getElementTypeOrSelf(ty1).dyn_cast<quant::UniformQuantizedPerAxisType>();
+  auto qty2 =
+      getElementTypeOrSelf(ty2).dyn_cast<quant::UniformQuantizedPerAxisType>();
+  if (!qty1 || !qty2) return false;
+
+  if (qty1.getScales().size() != qty1.getScales().size()) return false;
+
+  for (auto [lhs, rhs] : llvm::zip(qty1.getScales(), qty2.getScales()))
+    if (lhs != rhs) return false;
+
+  for (auto [lhs, rhs] : llvm::zip(qty1.getZeroPoints(), qty2.getZeroPoints()))
+    if (lhs != rhs) return false;
+
+  return true;
 }
 
 }  // namespace
@@ -366,6 +397,39 @@ LogicalResult verifyAddOp(std::optional<Location> location, Operation* op,
         location,
         "op requires the same element type for all operands and results");
 
+  return success();
+}
+
+LogicalResult verifyTransposeOp(std::optional<Location> location,
+                                Type operandType, ArrayRef<int64_t> permutation,
+                                Type resultType) {
+  // transpose_c1
+  if (allQuantized<quant::UniformQuantizedType>(operandType, resultType))
+    if (operandType != resultType)
+      return emitOptionalError(location,
+                               "expect same quantization scale and zero_point "
+                               "for operand and result but got ",
+                               operandType, " vs ", resultType);
+
+  if (allQuantized<quant::UniformQuantizedPerAxisType>(operandType, resultType))
+    if (!isSameQuantPerAxisScaleZeroPoint(operandType, resultType))
+      return emitOptionalError(location,
+                               "expect same quantization scales and "
+                               "zero_points for operand and result but got ",
+                               operandType, " vs ", resultType);
+
+  // transpose_c4
+  if (auto resultQType = getElementTypeOrSelf(resultType)
+                             .dyn_cast<quant::UniformQuantizedPerAxisType>()) {
+    auto resultQDim = resultQType.getQuantizedDimension();
+    auto operandQDim = getElementTypeOrSelf(operandType)
+                           .cast<quant::UniformQuantizedPerAxisType>()
+                           .getQuantizedDimension();
+    if (operandQDim != permutation[resultQDim])
+      return emitOptionalError(location, "operand quantization_dimension ",
+                               operandQDim, " is not same as permutation[",
+                               resultQDim, "] ", permutation[resultQDim]);
+  }
   return success();
 }
 
@@ -3183,8 +3247,15 @@ LogicalResult verifyBitcastConvertOp(std::optional<Location> location,
         location, "cannot convert between real and complex types, but got: ",
         operandShapedType, " and ", targetShapedType);
 
-  auto targetEltBitWidth = getBitWidth(targetElt);
-  auto operandEltBitWidth = getBitWidth(operandElt);
+  auto targetEltBitWidth =
+      targetElt.isa<quant::QuantizedType>()
+          ? getBitWidth(targetElt.cast<quant::QuantizedType>().getStorageType())
+          : getBitWidth(targetElt);
+  auto operandEltBitWidth =
+      operandElt.isa<quant::QuantizedType>()
+          ? getBitWidth(
+                operandElt.cast<quant::QuantizedType>().getStorageType())
+          : getBitWidth(operandElt);
 
   auto operandType = operandShapedType.dyn_cast<RankedTensorType>();
   auto targetType = targetShapedType.dyn_cast<RankedTensorType>();
@@ -3251,6 +3322,15 @@ LogicalResult verifyBroadcastInDimOp(std::optional<Location> location,
     return success();
   }
 
+  // broadcast_in_dim_c1
+  if (allQuantized<quant::UniformQuantizedType>(operand.getType(),
+                                                result.getType()))
+    if (operand.getType() != result.getType())
+      return emitOptionalError(location,
+                               "expect same quantization scale and zero_point "
+                               "for operand and result but got ",
+                               operand.getType(), " vs ", result.getType());
+
   // broadcast_in_dim_c2
   auto dimensionsSize = broadcastDimensions.size();
   auto operandRank = operandType.getRank();
@@ -3283,6 +3363,34 @@ LogicalResult verifyBroadcastInDimOp(std::optional<Location> location,
             location, "size of operand dimension ", i, " (", dimSize,
             ") is not equal to 1 or size of result dimension ", dimIndex, " (",
             resultDimSize, ")");
+    }
+  }
+
+  // broadcast_in_dim_c6
+  if (auto resultQType = getElementTypeOrSelf(result.getType())
+                             .dyn_cast<quant::UniformQuantizedPerAxisType>()) {
+    auto operandQType = getElementTypeOrSelf(operand.getType())
+                            .cast<quant::UniformQuantizedPerAxisType>();
+    auto operandQDim = operandQType.getQuantizedDimension();
+    auto resultQDim = resultQType.getQuantizedDimension();
+    if (resultQDim != broadcastDimensions[operandQDim])
+      return emitOptionalError(location, "result quantization_dimension ",
+                               resultQDim, " not same as broadcast_dimensions ",
+                               operandQDim, " (",
+                               broadcastDimensions[operandQDim], ")");
+    if (operandType.getDimSize(operandQDim) == 1) {
+      for (int64_t j = 0; j != resultType.getDimSize(resultQDim); ++j) {
+        if (resultQType.getScales()[j] != operandQType.getScales()[0])
+          return emitOptionalError(location, "mismatch result scale ", j, " (",
+                                   resultQType.getScales()[j],
+                                   ") and operand scale 0 (",
+                                   operandQType.getScales()[0], ")");
+        if (resultQType.getZeroPoints()[j] != operandQType.getZeroPoints()[0])
+          return emitOptionalError(location, "mismatch result zero_point ", j,
+                                   " (", resultQType.getZeroPoints()[j],
+                                   ") and operand zero_point 0 (",
+                                   operandQType.getZeroPoints()[0], ")");
+      }
     }
   }
 
@@ -3954,6 +4062,59 @@ LogicalResult verifyReduceWindowOp(
   return success();
 }
 
+LogicalResult verifyReshapeOpQuantizationConstraints(
+    std::optional<Location> location, Type operandTy, Type resultTy) {
+  // reshape_c1
+  if (allQuantized<quant::UniformQuantizedType>(operandTy, resultTy))
+    if (getElementTypeOrSelf(operandTy) != getElementTypeOrSelf(resultTy))
+      return emitOptionalError(location,
+                               "expect same quantization scale and zero_point "
+                               "for operand and result but got ",
+                               operandTy, " vs ", resultTy);
+
+  // reshape_c1
+  if (allQuantized<quant::UniformQuantizedPerAxisType>(operandTy, resultTy)) {
+    if (!isSameQuantPerAxisScaleZeroPoint(operandTy, resultTy)) {
+      return emitOptionalError(
+          location,
+          "expect same quantization scales and zero_points "
+          "for operand and result but got ",
+          operandTy, " vs ", resultTy);
+    }
+    // reshape_c3
+    auto operandQDim = getElementTypeOrSelf(operandTy)
+                           .cast<quant::UniformQuantizedPerAxisType>()
+                           .getQuantizedDimension();
+    auto resultQDim = getElementTypeOrSelf(resultTy)
+                          .cast<quant::UniformQuantizedPerAxisType>()
+                          .getQuantizedDimension();
+    auto operandShape = operandTy.cast<ShapedType>().getShape();
+    auto resultShape = resultTy.cast<ShapedType>().getShape();
+
+    if (operandTy.cast<ShapedType>().getDimSize(operandQDim) !=
+        resultTy.cast<ShapedType>().getDimSize(resultQDim))
+      return emitOptionalError(
+          location,
+          "expect same quantization dimension size for operand and result ",
+          operandTy, " and ", resultTy);
+
+    uint64_t operandProd = 1;
+    std::for_each(operandShape.begin(), operandShape.begin() + operandQDim,
+                  [&operandProd](int32_t dimSize) { operandProd *= dimSize; });
+    uint64_t resultProd = 1;
+    std::for_each(resultShape.begin(), resultShape.begin() + resultQDim,
+                  [&resultProd](int32_t dimSize) { resultProd *= dimSize; });
+    if (operandProd != resultProd)
+      return emitOptionalError(
+          location,
+          "product of dimensions before quantization dimension must match "
+          "between operand and result for ",
+          operandProd, " and ", resultProd);
+  }
+
+  return success();
+}
+
 LogicalResult verifyReshapeOp(std::optional<Location> location, Value operand,
                               Value result) {
   // If the operand type is dynamically shaped there is nothing to verify.
@@ -3970,6 +4131,10 @@ LogicalResult verifyReshapeOp(std::optional<Location> location, Value operand,
                              numResultElements,
                              ") doesn't match expected number of elements (",
                              numOperandElements, ")");
+
+  if (anyQuantized<quant::QuantizedType>(operand.getType(), result.getType()))
+    return verifyReshapeOpQuantizationConstraints(location, operand.getType(),
+                                                  result.getType());
 
   return success();
 }
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -184,6 +184,28 @@ bool isSameQuantPerAxisScaleZeroPoint(Type ty1, Type ty2) {
   return true;
 }
 
+LogicalResult verifyQPerTensorScaleAndZeroPointConstraints(
+    std::optional<Location> location, Type ty1, Type ty2) {
+  if (allQuantized<quant::UniformQuantizedType>(ty1, ty2)) {
+    if (getElementTypeOrSelf(ty1) != getElementTypeOrSelf(ty2))
+      return emitOptionalError(
+          location, "expect same quantization scale and zero_point but got ",
+          ty1, " vs ", ty2);
+  }
+  return success();
+}
+
+LogicalResult verifyQPerAxisScaleAndZeroPointConstraints(
+    std::optional<Location> location, Type ty1, Type ty2) {
+  if (allQuantized<quant::UniformQuantizedPerAxisType>(ty1, ty2)) {
+    if (!isSameQuantPerAxisScaleZeroPoint(ty1, ty2))
+      return emitOptionalError(
+          location, "expect same quantization scales and zero_points but got ",
+          ty1, " vs ", ty2);
+  }
+  return success();
+}
+
 }  // namespace
 
 //===----------------------------------------------------------------------===//
@@ -404,19 +426,13 @@ LogicalResult verifyTransposeOp(std::optional<Location> location,
                                 Type operandType, ArrayRef<int64_t> permutation,
                                 Type resultType) {
   // transpose_c1
-  if (allQuantized<quant::UniformQuantizedType>(operandType, resultType))
-    if (operandType != resultType)
-      return emitOptionalError(location,
-                               "expect same quantization scale and zero_point "
-                               "for operand and result but got ",
-                               operandType, " vs ", resultType);
+  if (failed(verifyQPerTensorScaleAndZeroPointConstraints(location, operandType,
+                                                          resultType)))
+    return failure();
 
-  if (allQuantized<quant::UniformQuantizedPerAxisType>(operandType, resultType))
-    if (!isSameQuantPerAxisScaleZeroPoint(operandType, resultType))
-      return emitOptionalError(location,
-                               "expect same quantization scales and "
-                               "zero_points for operand and result but got ",
-                               operandType, " vs ", resultType);
+  if (failed(verifyQPerAxisScaleAndZeroPointConstraints(location, operandType,
+                                                        resultType)))
+    return failure();
 
   // transpose_c4
   if (auto resultQType = getElementTypeOrSelf(resultType)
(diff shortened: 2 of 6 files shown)


stablehlo/dialect/TypeInference.cpp
```diff
@@ -3307,6 +3365,34 @@ LogicalResult verifyBroadcastInDimOp(std::optional<Location> location,
...
+                            .cast<quant::UniformQuantizedPerAxisType>();
+    auto operandQDim = operandQType.getQuantizedDimension();
+    auto resultQDim = resultQType.getQuantizedDimension();
+    if (resultQDim != broadcastDimensions[operandQDim])
+      return emitOptionalError(location, "result quantization_dimension ",
+                               resultQDim, " not same as broadcast_dimensions ",
+                               operandQDim, " (",
+                               broadcastDimensions[operandQDim], ")");
+    if (operandType.getDimSize(operandQDim) == 1) {
+      for (auto j = 0; j != resultType.getDimSize(resultQDim); ++j) {
```
sdasgup3, line 3381: `size_t j = 0`

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3565,13 +3651,75 @@ LogicalResult verifyDotGeneralOp(std::optional<Location> location, Value lhs,
...
+  // dot_general_c16
+  auto expressedType = lhsQType.getExpressedType();
+  if (expressedType != rhsQType.getExpressedType() ||
+      expressedType != resultQType.getExpressedType())
+    return emitOptionalError(location,
+                             "mismatched operands and result expressed types");
+
+  auto rhsQPAType = rhsQType.dyn_cast<quant::UniformQuantizedPerAxisType>();
+  // dot_general_c17
+  auto expectedZP = 0;
```
sdasgup3, line 3691: Can w do some thing this instead

```
  auto rhsQPAType = rhsQType.dyn_cast<quant::UniformQuantizedPerAxisType>();
SmallVector<int64_t> zeroPoints = rhsQPAType ? rhsQPAType.getZeroPoints() : {rhsQType.cast<quant::UniformQuantizedType>().getZeroPoint()};

if(llvm::all_of(zeroPoints, [](int64_t zp) {return zp != 0})) {
return emitOptionalError(location, "Zero point of rhs type ",rhsQType, " is expected to be 0");
}
```

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3565,13 +3651,75 @@ LogicalResult verifyDotGeneralOp(std::optional<Location> location, Value lhs,
...
+    if (rhsZP != expectedZP)
+      return emitOptionalError(location, "rhs zero_point ", rhsZP, " is not ",
+                               expectedZP);
+  }
+
+  auto resultQPAType =
+      resultQType.dyn_cast<quant::UniformQuantizedPerAxisType>();
+  // dot_general_c18
+  if (!rhsQPAType && resultQPAType)
+    return emitOptionalError(location,
```
sdasgup3, line 3709: Same as convolution error message
```
    return emitOptionalError(
        location, "per-tensor rhs expects per-tensor result but received ",
        rhsType, " and ", resultType, " respectively");
  }
```

stablehlo/dialect/TypeInference.cpp
```diff
@@ -4054,6 +4202,13 @@ LogicalResult verifyReshapeOp(std::optional<Location> location, Value operand,
                              ") doesn't match expected number of elements (",
                              numOperandElements, ")");
 
+  // reshape_c1
+  // part of reshape_c1 already verified by
+  // Trait HLO_CompatibleOperandsAndResultElementType
+  if (failed(sameQuantScaleZeroPoint(location, operand.getType(),
```
sdasgup3, line 4208: should be addressed via #2112

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3272,6 +3323,13 @@ LogicalResult verifyBroadcastInDimOp(std::optional<Location> location,
     return success();
   }
 
+  // broadcast_in_dim_c1
+  // part of broadcast_in_dim_c1 already verified by
+  // Trait HLO_CompatibleOperandsAndResultElementType
+  if (failed(sameQuantScaleZeroPoint(location, operand.getType(),
```
sdasgup3, line 3329: I note here that `broadcast_in_dim` should not use the trait introduced at #2112. At the constrainsts on scales/zps for per-tensor and per-axis are different.

stablehlo/dialect/TypeInference.cpp
abhigunj: neat! needed additional minor change.  thank you.

we decided to drop `dot_general` verifier from this PR due to internal test failures. Removed this change for now. I will park it as a draft PR.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -83,6 +83,27 @@ bool noneQuantized(ArrayRef<Type> typeRange) {
   });
 }
 
+// Try to match scale and zero_point for per_tensor quantized inputs
+LogicalResult sameQuantScaleZeroPoint(std::optional<Location> location,
```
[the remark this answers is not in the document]
abhigunj: Decided not to add New Trait from https://github.com/openxla/stablehlo/pull/2112 
refer: https://github.com/openxla/stablehlo/pull/2112#discussion_r1534546696

stablehlo/dialect/TypeInference.cpp
abhigunj: we will need this check in absence of the new Trait.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -283,6 +304,29 @@ LogicalResult verifyPairwiseCompatibleShapes(TypeRange values) {
   return success();
 }
 
+LogicalResult verifyTransposeOp(std::optional<Location> location,
+                                Type operandType, ArrayRef<int64_t> permutation,
+                                Type resultType) {
+  // transpose_c1
+  // part of transpose_c1 already verified by
```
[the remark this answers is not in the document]
sdasgup3: Thanks! now we can remove the comments except for `// transpose_c1`

stablehlo/dialect/TypeInference.cpp
```diff
@@ -71,16 +71,37 @@ namespace {
...
 template <typename T>
 bool noneQuantized(ArrayRef<Type> typeRange) {
-  return llvm::all_of(typeRange, [&](Type val) {
-    return !val.cast<ShapedType>().getElementType().isa<T>();
-  });
+  return llvm::all_of(
+      typeRange, [&](Type val) { return !getElementTypeOrSelf(val).isa<T>(); });
+}
+
+// Try to match scale and zero_point for per_tensor quantized inputs
```
sdasgup3, line 84: Try to make this generic for per-tensor and per-axis.. The name of the function is sufficient to assume that it can be applied for both the cases.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -283,6 +304,29 @@ LogicalResult verifyPairwiseCompatibleShapes(TypeRange values) {
   return success();
 }
 
+LogicalResult verifyTransposeOp(std::optional<Location> location,
+                                Type operandType, ArrayRef<int64_t> permutation,
+                                Type resultType) {
+  // transpose_c1
+  // part of transpose_c1 already verified by
+  // Trait HLO_CompatibleOperandsAndResultElementType
+  if (failed(isSameQuantScaleZeroPoint(location, operandType, resultType)))
```
sdasgup3, line 313: Transpose can be both per-axis and per-tensor. We are not matching the scales and zero-points for per-axis case?

stablehlo/dialect/TypeInference.cpp
```diff
@@ -4054,6 +4138,13 @@ LogicalResult verifyReshapeOp(std::optional<Location> location, Value operand,
                              ") doesn't match expected number of elements (",
                              numOperandElements, ")");
 
+  // reshape_c1
+  // part of reshape_c1 already verified by
```
sdasgup3, line 4142: remove comments as before

stablehlo/dialect/TypeInference.cpp
```diff
@@ -4458,6 +4551,13 @@ LogicalResult verifyWhileOp(std::optional<Location> location,
         location,
         "expect condition block return a zero-ranked tensor of i1 but got ",
         condReturnTypes[0]);
+  // while_c3
```
sdasgup3, line 4554: why do we need that? Itsn't this checked in during type inference?

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3272,6 +3323,11 @@ LogicalResult verifyBroadcastInDimOp(std::optional<Location> location,
     return success();
   }
 
+  // broadcast_in_dim_c1
+  if (failed(isSameQuantScaleZeroPoint(location, operand.getType(),
```
sdasgup3, line 3327:
```
(C1) element_type(result) is given by:
element_type(operand), if !is_per_axis_quantized(operand).
element_type(operand) except that quantization_dimension(operand), scales(operand), and zero_points(operand) may differ from quantization_dimension(result), scales(result), and zero_points(result) resp., otherwise.
```

whya re we macthing scales and zero-points here as broadcast_in_dim does not need that?

stablehlo/dialect/TypeInference.cpp
```diff
@@ -4054,6 +4138,13 @@ LogicalResult verifyReshapeOp(std::optional<Location> location, Value operand,
                              ") doesn't match expected number of elements (",
                              numOperandElements, ")");
 
+  // reshape_c1
+  // part of reshape_c1 already verified by
+  // Trait HLO_CompatibleOperandsAndResultElementType
+  if (failed(isSameQuantScaleZeroPoint(location, operand.getType(),
```
sdasgup3, line 4144: we are not macthing per-axis scale/zero-points

stablehlo/dialect/TypeInference.cpp
abhigunj: refactored isSameQuantScaleZeroPoint to include per-axis Q support, removed location logging, removed per scale or per zero_point mismatch logging. Now caller will log the error.  We won't get scale mismatch or zero_point mismatch granularity in error message instead generic error message "expect same quantization scale and zero_point"

stablehlo/dialect/TypeInference.cpp
abhigunj: yes.. type inference is checking it. validated.
Removed, kept the test.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -71,16 +71,65 @@ namespace {
...
+  if (std::is_same<QuantTy, quant::UniformQuantizedType>::value) {
+    auto qty1 =
+        getElementTypeOrSelf(ty1).dyn_cast<quant::UniformQuantizedType>();
+    auto qty2 =
+        getElementTypeOrSelf(ty2).dyn_cast<quant::UniformQuantizedType>();
+    scales1.push_back(qty1.getScale());
+    scales2.push_back(qty2.getScale());
+    zPoints1.push_back(qty1.getZeroPoint());
+    zPoints2.push_back(qty2.getZeroPoint());
+  }
```
GleasonK, line 110: `return ty1 == ty2` ?

abhigunj: removed scale and zero_point matching for `per-tensor` quant. Instead, will be using `type1 == type2`. There is overlap of checks between HLO_CompatibleOperandsAndResultElementType Trait and using `type1 == type2` here. But it improves readability and reduces use of custom logic.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3869,6 +3996,59 @@ LogicalResult verifyReduceWindowOp(
...
+    auto resultShape = resultTy.cast<ShapedType>().getShape();
+
+    if (operandTy.cast<ShapedType>().getDimSize(operandQDim) !=
+        resultTy.cast<ShapedType>().getDimSize(resultQDim))
+      return emitOptionalError(
+          location,
+          "expect same quantization dimension size for operand and result ",
+          operandTy, " and ", resultTy);
+
+    uint64_t operandProd = 1;
```
sdasgup3, line 4035: The second half of the check is not implemented where we need to compare the product of the dim-sizes past the quantization dimension.  
```
reduce(dims(operand, [quantization_dimension(operand) + 1, ..., rank(operand) - 1]), init_values=1, dimensions=[0], body=lambda x, y: x * y) = reduce(dims(result, [quantization_dimension(result) + 1, ..., rank(result) - 1]), init_values=1, dimensions=[0], body=lambda x, y: x * y).
```

sdasgup3:
```
    isCompatibleQuantizationDimention = [&](operandShape, resultShape, int64_t begin_pos, int64_end_pos) {
       uint64_t operandProd = 1;
       std::for_each(operandShape.begin() + begin_pos, operandShape.begin() + end_pos,
                  [&operandProd](int32_t dimSize) { operandProd *= dimSize; });
        uint64_t resultProd = 1;
        std::for_each(resultShape.begin() + negin_pos, resultShape.begin() + end_pos,
                  [&resultProd](int32_t dimSize) { resultProd *= dimSize; });
     };
    

   return isCompatibleQuantizationDimention(operandShape, resultShape, 0, operandQDim) && isCompatibleQuantizationDimention(operandShape, resultShape,  operandQDim, resultShape.size()-1)
```

abhigunj: I added check for the second half, however it is already covered by other checks and I was unable to simulate test condition. 

1. there is a check which validate operand element count == result element count
2. product of dims before quantization dim are equal for operand and result (first half)
3. quantization dim sizes are equal. 

after these checks, 2nd half of the check is already verified. We can add placeholder code but can't add test for it. So decided to drop the changes for second half of the constraint.

sdasgup3: Thanks for explaining. This makes sense!

sdasgup3 (approved):

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3183,8 +3248,15 @@ LogicalResult verifyBitcastConvertOp(std::optional<Location> location,
         location, "cannot convert between real and complex types, but got: ",
         operandShapedType, " and ", targetShapedType);
 
-  auto targetEltBitWidth = getBitWidth(targetElt);
-  auto operandEltBitWidth = getBitWidth(operandElt);
+  auto targetEltBitWidth =
```
GleasonK, line 3251: Update `getBitWidth` to handle this internally.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -369,6 +400,40 @@ LogicalResult verifyAddOp(std::optional<Location> location, Operation* op,
...
+                                Type operandType, ArrayRef<int64_t> permutation,
+                                Type resultType) {
+  // transpose_c1
+  if (allQuantized<quant::UniformQuantizedType>(operandType, resultType)) {
+    if (getElementTypeOrSelf(operandType) != getElementTypeOrSelf(resultType))
+      return emitOptionalError(location,
+                               "expect same quantization scale and zero_point "
+                               "for operand and result but got ",
+                               operandType, " vs ", resultType);
+  }
```
GleasonK, line 413: Lots of repeating logic. I think this PR can benefit from a few functions like:

```
// Verifies uniform quantized and per-axis scale and zero point match
LogicalResult verifyQuantizedScaleAndZeroPointConstraints();

// Verifies uniform quantized tensor type scale and zp matches
LogicalResult verifyUniformQuantizedScaleAndZeroPointConstraints();
```

stablehlo/dialect/TypeInference.cpp
```diff
@@ -369,6 +400,40 @@ LogicalResult verifyAddOp(std::optional<Location> location, Operation* op,
...
+  if (allQuantized<quant::UniformQuantizedPerAxisType>(operandType,
+                                                       resultType)) {
```
GleasonK, line 415: Is it valid for transpose to have a quantized operand but not a quantized result? Is this handled elsewhere?

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3251,6 +3323,15 @@ LogicalResult verifyBroadcastInDimOp(std::optional<Location> location,
...
 
+  // broadcast_in_dim_c1
+  if (allQuantized<quant::UniformQuantizedType>(operand.getType(),
+                                                result.getType())) {
+    if ((getElementTypeOrSelf(operand) != getElementTypeOrSelf(result)))
+      return emitOptionalError(location,
+                               "expect same quantization scale and zero_point "
+                               "for operand and result but got ",
+                               operand.getType(), " vs ", result.getType());
+  }
```
GleasonK, line 3334: Per above comment:

```
if (failed(verifyUniformQuantizedScaleAndZeroPoint()))
  return failure();
```

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3954,6 +4063,59 @@ LogicalResult verifyReduceWindowOp(
...
+
+  // reshape_c1
+  if (allQuantized<quant::UniformQuantizedPerAxisType>(operandTy, resultTy)) {
+    if (!isSameQuantPerAxisScaleZeroPoint(operandTy, resultTy)) {
+      return emitOptionalError(
+          location,
+          "expect same quantization scales and zero_points "
+          "for operand and result but got ",
+          operandTy, " vs ", resultTy);
+    }
```
GleasonK, line 4084: Per above comment:

```
if (failed(verifyQuantizedScaleAndZeroPoint()))
  return failure();
```

GleasonK (changes requested):

stablehlo/dialect/TypeInference.cpp
abhigunj: HLO_CompatibleOperandsAndResultElementType Trait is handling quant vs non quant or per-tensor vs per-axis mismatch.

stablehlo/dialect/TypeInference.cpp
abhigunj: getBitWidth is updated. this change is not required. thank you.

stablehlo/dialect/TypeInference.cpp
abhigunj: done, thanks for the suggestion.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3954,6 +4067,56 @@ LogicalResult verifyReduceWindowOp(
...
+LogicalResult verifyReshapeOpQuantizationConstraints(
+    std::optional<Location> location, Type operandTy, Type resultTy) {
+  // reshape_c1
+  if (failed(verifyQPerTensorScaleAndZeroPointConstraints(location, operandTy,
+                                                          resultTy)))
+    return failure();
+
+  // reshape_c1
+  if (allQuantized<quant::UniformQuantizedPerAxisType>(operandTy, resultTy)) {
+    if (!isSameQuantPerAxisScaleZeroPoint(operandTy, resultTy)) {
```
GleasonK, line 4079: Let's use the `verify` method you added above. Will cost an extra `if` stmt but should be cheap.

stablehlo/dialect/StablehloOps.td
```diff
@@ -2871,7 +2871,9 @@ def StableHLO_TransposeOp: StableHLO_ShapedInterfaceOp<"transpose",
     HLO_TensorOrPerAxisQuantizedTensor:$operand,
     DenseI64ArrayAttr:$permutation
   );
-  let results = (outs HLO_TensorOrPerAxisQuantizedTensor);
+  let results = (outs HLO_TensorOrPerAxisQuantizedTensor:$result);
+
+  let hasVerifier = 1;
```
GleasonK, line 2876: cc @ghpvnist - I think the new verification code should go in `TypeInference.cpp > inferTransposeOp`, per https://github.com/openxla/stablehlo/blob/main/docs/type_inference.md#p3-maintain-verification-code-in-verifiers-and-shape-functions

Does this line up with your expectations having written many of these?

GleasonK: This setup is similar to `ReverseOp` (and probably a few others) but I'm not sure if that's out of model as well

GleasonK: At a minimum lets update the verification doc linked above

abhigunj: `+ AddOp` 
what we conclude here is also applicable for AddOp

stablehlo/tests/ops_stablehlo_quantized.mlir
```diff
@@ -821,3 +824,194 @@ func.func @illegal_storage_type_for_quantized_element_type(%arg0: tensor<4x!quan
...
+func.func @transpose_c1_mismatched_zp(%arg0: tensor<1x2x2x!quant.uniform<i8<-128:127>:f32, 0.1:-30>>) {
+  // expected-error@+1 {{expect same quantization scale and zero_point}}
+  %transpose = "stablehlo.transpose"(%arg0) {permutation = array<i64: 0, 2, 1>
+  } : (tensor<1x2x2x!quant.uniform<i8<-128:127>:f32, 0.1:-30>>) -> tensor<1x2x2x!quant.uniform<i8<-128:127>:f32, 0.1:-20>>
+  func.return
+}
+
+// -----
+
+func.func @transpose_c1_mismatched_scales(%arg0: tensor<1x2x2x!quant.uniform<i8<-128:127>:f32:0, {0.1:-30, 0.5:-20}>>) {
```
ghpvnist, line 906: Does the following two tests test scales and zp that is not covered by the above two? This [guideline](https://github.com/openxla/stablehlo/blob/main/docs/reference.md#testing-guidelines) was written for the reference interpreter, but it also applies to mlir files.

stablehlo/tests/ops_stablehlo_quantized.mlir
```diff
@@ -821,3 +824,194 @@ func.func @illegal_storage_type_for_quantized_element_type(%arg0: tensor<4x!quan
...
+
+func.func @reshape_c3_mismatch_preoduct_before(%arg0: tensor<1x2x3x4x5x!quant.uniform<i8:f32:0, {1.0:17}>>){
+  // expected-error@+1 {{product of dimensions before quantization dimension must match between operand and result}}
+  %reshape = "stablehlo.reshape" (%arg0) : (tensor<1x2x3x4x5x!quant.uniform<i8:f32:0, {1.0:17}>>) -> tensor<2x1x3x20x!quant.uniform<i8:f32:1, {1.0:17}>>
+  func.return
+}
+
+// -----
+
+func.func @while_c3(%arg0: tensor<4x!quant.uniform<i8:f32, 1.0:17>>) -> tensor<?x!quant.uniform<i8:f64, 1.0:17>> {
```
ghpvnist, line 973: I think all the tests below (while, sort, abs, select) are covered by ODS and does not need to be tested ([#5 in this checklist](https://github.com/openxla/stablehlo/blob/main/docs/reference_checklist.md))

ghpvnist (changes requested):

stablehlo/dialect/StablehloOps.td
abhigunj: thanks @ghpvnist  missed your comment.. maybe didn't refreshed the page. 

`HLO_CompatibleOperandsAndResultElementTyp`e doesn't cover all of `transpose_c1` 
also `transpose_c4` is not covered.

ghpvnist: Okay makes sense. The point of labeling `HLO_CompatibleOperandsAndResultElementType` with `transpose_c1` is that it tests at least a subset of c1. We're good as long as we aren't writing tests that ODS already covers.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -369,6 +422,33 @@ LogicalResult verifyAddOp(std::optional<Location> location, Operation* op,
   return success();
 }
 
+LogicalResult verifyTransposeOp(std::optional<Location> location,
+                                Type operandType, ArrayRef<int64_t> permutation,
+                                Type resultType) {
+  // transpose_c1
+  if (failed(verifyQPerTensorScaleAndZeroPointConstraints(location, operandType,
+                                                          resultType)))
+    return failure();
```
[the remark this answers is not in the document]
abhigunj: failure()  returns the `emitOptionalError ` logged at the API.

stablehlo/dialect/StablehloOps.td
sdasgup3: As the result typeof transpose is fully inferable, it can be created suing `TransposeOp::build` which does not take result type as input and hence it makes sense to  move the verification logic of transpose to inferTransposeOp to avoid missing on any verification during op creation.

For ops like quantized version of AddOp, the result type cannot be inferred as the scales and zero-points can vary. In order to create such an op, we need to use a version of AddOP::build which takes result type as parameter and that builder will never invoke `Add::inferReturnTypes`. So thereis no point for AddOp verfication logic for quantization to move to type ingerence. 

WIth that it makes sense to move the code from verification to type inference if
1. The op has region.
2. The type inference is feasible.

Note that the reason reshapeOp and broadcast_in_dim op has separete verifioers is because their type cannot be inferred. 

We should update https://github.com/openxla/stablehlo/blob/main/docs/type_inference.md#p3-maintain-verification-code-in-verifiers-and-shape-functions with this information for future use.

abhigunj: thanks folks. 

* updated `type_inference.md` at https://github.com/openxla/stablehlo/pull/2158 as conclusion from this thread. 
* `TransposeOp` : since return type inference is feasible, will move the verification logic from custom verifier to `inferTransposeOp`
* `AddOp` : no change needed because return type inference is not feasible for quantized types. 
* I will also take a relook at existing ops and need of dedicated verifier.

reopening this comment for further  discussion if needed..

stablehlo/tests/ops_stablehlo_quantized.mlir
```diff
@@ -821,3 +824,194 @@ func.func @illegal_storage_type_for_quantized_element_type(%arg0: tensor<4x!quan
...
+  %arg0: tensor<1x2x1x!quant.uniform<i8<-128:127>:f32, 0.2:-30>>) {
+  // expected-error@+1 {{expect same quantization scale and zero_point}}
+  %broadcast_in_dim = "stablehlo.broadcast_in_dim" (%arg0) {broadcast_dimensions = array<i64: 0, 1, 2>
+  } : (tensor<1x2x1x!quant.uniform<i8<-128:127>:f32, 0.2:-30>>) -> tensor<1x2x3x2x!quant.uniform<i8<-128:127>:f32, 0.1:-30>>
+  func.return
+}
+
+// -----
+
+func.func @broadcast_in_dim_c1(
```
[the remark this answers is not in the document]
abhigunj: yes, broadcast_in_dim_c1 is custom verifier.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -75,6 +75,12 @@ bool allQuantized(ArrayRef<Type> typeRange) {
       typeRange, [&](Type val) { return getElementTypeOrSelf(val).isa<T>(); });
 }
 
+template <typename T>
+bool allQuantized(Type tp1, Type tp2) {
```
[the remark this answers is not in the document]
abhigunj: thanks, thought about it.. trying to keep names neutral because we have Ops with operands `lhstype` and `rhstype`. naming consistent with HLO trait apis.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -153,6 +166,46 @@ LogicalResult verifyBinaryOpQuantizationConstraints(
   return success();
 }
 
+bool isSameQuantPerAxisScaleZeroPoint(Type ty1, Type ty2) {
+  auto qty1 =
```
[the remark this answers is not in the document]
abhigunj: thanks, keeping it same.. same reason as above.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -153,6 +166,46 @@ LogicalResult verifyBinaryOpQuantizationConstraints(
...
+      getElementTypeOrSelf(ty1).dyn_cast<quant::UniformQuantizedPerAxisType>();
+  auto qty2 =
+      getElementTypeOrSelf(ty2).dyn_cast<quant::UniformQuantizedPerAxisType>();
+  if (!qty1 || !qty2) return false;
+
+  if (qty1.getScales().size() != qty1.getScales().size()) return false;
+
+  for (auto [lhs, rhs] : llvm::zip(qty1.getScales(), qty2.getScales()))
```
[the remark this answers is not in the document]
ghpvnist: Ah looks like it could be messier when you combine them. You may disregard this comment.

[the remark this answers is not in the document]
abhigunj: thanks, we can, then we will need to maintain iterator position like qty1.getScales()[i]...   llvm::zip is cleaner.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -369,6 +422,33 @@ LogicalResult verifyAddOp(std::optional<Location> location, Operation* op,
   return success();
 }
 
+LogicalResult verifyTransposeOp(std::optional<Location> location,
+                                Type operandType, ArrayRef<int64_t> permutation,
```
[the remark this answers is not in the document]
abhigunj: nice catch. updated declaration :)

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3954,6 +4067,53 @@ LogicalResult verifyReduceWindowOp(
   return success();
 }
 
+LogicalResult verifyReshapeOpQuantizationConstraints(
+    std::optional<Location> location, Type operandTy, Type resultTy) {
```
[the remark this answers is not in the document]
abhigunj: Thanks for suggestion.. I will leave as it is... since won't change wrt readability. usage is contained within the function. So it won't create confusion with other variables.

stablehlo/tests/ops_stablehlo_quantized.mlir
abhigunj: yeh.. we can remove these tests it is going to test same api for different op.

abhigunj: these are  per-axis and previous ones are per-axis quantization tests. we do need both of them.

stablehlo/tests/ops_stablehlo_quantized.mlir
abhigunj: done, removed. Thank you.

stablehlo/tests/ops_stablehlo_quantized.mlir
abhigunj: thanks for the ref [guideline](https://github.com/openxla/stablehlo/blob/main/docs/reference.md#testing-guidelines). 
 API is tested with per-axis quantized inputs only for `transpose` op. avoided adding new tests for each op.

stablehlo/tests/ops_stablehlo_quantized.mlir
```diff
@@ -821,3 +824,146 @@ func.func @illegal_storage_type_for_quantized_element_type(%arg0: tensor<4x!quan
...
+
+func.func @reshape_c3_mismatch_qdim_size(%arg0: tensor<1x2x3x4x5x!quant.uniform<i8:f32:0, {1.0:17}>>){
+  // expected-error@+1 {{expect same quantization dimension size for operand and result}}
+  %reshape = "stablehlo.reshape" (%arg0) : (tensor<1x2x3x4x5x!quant.uniform<i8:f32:0, {1.0:17}>>) -> tensor<2x3x20x!quant.uniform<i8:f32:1, {1.0:17}>>
+  func.return
+}
+
+// -----
+
+func.func @reshape_c3_mismatch_preoduct_before(%arg0: tensor<1x2x3x4x5x!quant.uniform<i8:f32:0, {1.0:17}>>){
```
ghpvnist, line 965: Typo: `reshape_c3_mismatch_preoduct_before`

ghpvnist (approved):
LGTM with minor comment

PR #2082, opened 2024-03-11
Fix broken links

mlevesquedion:
Most of these links are only broken when rendered out of tree, e.g. on the OpenXLA website: https://openxla.org/stablehlo. For example, the "More Examples" link lead to e.g.
https://openxla.org/stablehlo/tests/interpret/add.mlir, which is 404.

For links under `docs/` it is possible to use relative paths.

Also, add labels to images used in spec. This will materialize as the "alt" text, which is used by screen readers.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -3,7 +3,7 @@
 StableHLO is a backward compatible ML compute opset inspired by HLO/MHLO.
 This document explains the kind and the extent of the compatibility guarantees
 that StableHLO provides, based on the process established in
-[the compatibility RFC](../rfcs/20220912-compatibility.md).
+[the compatibility RFC](https://github.com/openxla/stablehlo/tree/main/rfcs/20220912-compatibility.md).
 
 ## Versions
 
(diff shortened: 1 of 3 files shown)


docs/spec.md
```diff
@@ -5054,7 +5054,7 @@ a `result` tensor.
 The following diagram shows how elements in `result` are computed from
 `operand` and `source` using a concrete example.
 
-![](images/spec/select_and_scatter.svg)
+![select_and_scatter](images/spec/select_and_scatter.svg)
```
ghpvnist, line 5057: I prefer the default description to be the same as file name like here (but up to you). Could you also update others (or update this one to append `example`) for consistency? (e.g.
`scatter example` to `scatter`)

docs/compatibility.md
```diff
@@ -3,7 +3,7 @@
 StableHLO is a backward compatible ML compute opset inspired by HLO/MHLO.
 This document explains the kind and the extent of the compatibility guarantees
 that StableHLO provides, based on the process established in
-[the compatibility RFC](../rfcs/20220912-compatibility.md).
+[the compatibility RFC](https://github.com/openxla/stablehlo/tree/main/rfcs/20220912-compatibility.md).
```
ghpvnist, line 6: Based on your description, it looks like this can also be relative path.

ghpvnist (changes requested):

docs/compatibility.md
mlevesquedion: It can't because this is under `stablehlo/rfcs` rather than `stablehlo/docs`. Only paths under `docs` can be relative. Which part of the description is inaccurate?

docs/spec.md
mlevesquedion: Good catch, changed all to match the filename.

docs/compatibility.md
ghpvnist: Nvm. I thought `rfcs/` was nested under `docs/` :)

ghpvnist (approved):

docs/compatibility.md
mlevesquedion: Ah, makes sense. No worries. TBH I think it would make sense to nest `rfcs` under `docs`, maybe that can be a follow-up change.

sdasgup3 (approved):
LGTM.
Just an FYI: There are other links at the index page of https://openxla.org/stablehlo which needs simialar fixes. I can take on that.

mlevesquedion:
> LGTM. Just an FYI: There are other links at the index page of https://openxla.org/stablehlo which needs simialar fixes. I can take on that.

Ah interesting. Good catch yeah it looks like some links in the README point to e.g. `docs/`, but on the OpenXLA website the relevant pages appear under the top-level rather than under some `docs` subdirectory.

Yes, please feel free to send a PR! :)

PR #2083, opened 2024-03-11
Add examples directory

fzakaria:
Add the beginnings of an `examples` directory structure that will help demonstrate how to use the StableHLO library.

At the moment I've only added a c++ example, and a simple one at that but the directory structure should indicate that we can support Python in the future.

* Added a meta library 'stablehlo_lib' that you can link against which should eventually include everything
* Added the add stablehlo example
* Added a simple README

```console
❯ ./build/examples/c++/stablehlo-add
module @test_module {
  func.func @main(%arg0: tensor<3x4xf32>, %arg1: tensor<3x4xf32>) -> tensor<3x4xf32> {
    %0 = stablehlo.add %arg0, %arg1 : tensor<3x4xf32>
    %1 = stablehlo.add %arg0, %arg1 : tensor<3x4xf32>
    return %1 : tensor<3x4xf32>
  }
}
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -187,14 +187,4 @@ add_custom_target(check-stablehlo-quick)
 add_dependencies(check-stablehlo-ci check-stablehlo-quick)
 
 add_subdirectory(stablehlo)
-
-# Let's create a single library here so that linking against
-add_library(stablehlo_lib INTERFACE)
-# Link the master library with the other libraries
-target_link_libraries(stablehlo_lib INTERFACE
-  ${dialect_libs}
-  ${conversion_libs}
-  StablehloOps
-  )
-
 add_subdirectory(examples)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -187,3 +187,14 @@ add_custom_target(check-stablehlo-quick)
 add_dependencies(check-stablehlo-ci check-stablehlo-quick)
 
 add_subdirectory(stablehlo)
+
+# Let's create a single library here so that linking against
+add_library(stablehlo_lib INTERFACE)
+# Link the master library with the other libraries
+target_link_libraries(stablehlo_lib INTERFACE
+  ${dialect_libs}
+  ${conversion_libs}
+  StablehloOps
+  )
+
+add_subdirectory(examples)
--- a/examples/c++/BUILD.bazel
+++ b/examples/c++/BUILD.bazel
@@ -43,4 +43,3 @@ cc_test(
         "@llvm-project//mlir:QuantOps",
     ],
 )
-
--- a/examples/c++/BUILD.bazel
+++ b/examples/c++/BUILD.bazel
@@ -1,4 +1,4 @@
-# Copyright 2024 The StableHLO Authors.
+# Copyright 2023 The StableHLO Authors. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


examples/c++/stablehlo_add.cpp
```diff
@@ -0,0 +1,54 @@
...
+  block_builder.create<mlir::stablehlo::AddOp>(loc, arguments, attributes)
+      .getOperation();
+
+  mlir::Operation* op =
+      block_builder
+          .create<mlir::stablehlo::AddOp>(loc, arg0, arguments, attributes)
+          .getOperation();
+  block_builder.create<mlir::func::ReturnOp>(loc, op->getResult(0));
+
+  // verify the module and dump
```
sdasgup3, line 49: For consistency lets use `/**/` comment style we were using for high level actions?

sdasgup3 (approved):

CMakeLists.txt
```diff
@@ -187,3 +187,14 @@ add_custom_target(check-stablehlo-quick)
...
+# Let's create a single library here so that linking against
+add_library(stablehlo_lib INTERFACE)
+# Link the master library with the other libraries
+target_link_libraries(stablehlo_lib INTERFACE
+  ${dialect_libs}
+  ${conversion_libs}
+  StablehloOps
+  )
+
+add_subdirectory(examples)
```
fzakaria, line 200: @sdasgup3 are you familiar enough with CMake to know if this pattern is a good idea?
The goal would be to make a single target library (interface) that includes everything we need to make examples easier to work with.

fzakaria:
@sdasgup3 I still need to add BUILD (Bazel support) which is why it is draft.
I will finish that this evening, much simpler.

CMakeLists.txt
sdasgup3: Not super familiar but I can see similar pattern in other places like https://github.com/openxla/stablehlo/blob/04365f85cfbffe3d95ba2fb79ff34cd929d4a9a6/stablehlo/tools/CMakeLists.txt#L49.

A few nits: 
1. Can we home this lib in examples/c++ directory instead, where it is actually consumed. 
2. Name it `stablehlo-example-lib` (with dashes) to be  consistent with other lib target names.

examples/c++/BUILD.bazel
```diff
@@ -0,0 +1,45 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_test")
+
+package(
+    default_visibility = ["//visibility:public"],
+    licenses = ["notice"],
+)
+
+cc_binary(
```
sdasgup3, line 21: Wondering if need to build this binary. `cc_test`  is sufficient here, right?

fzakaria: Just so you can `bazel run` it ?

sdasgup3: Yes, or `bazel test`

fzakaria: I mean t-- I added cc_binary so you can get a runnable binary.
This will be needed if eventually we create binaries that take any args on input -- so you can give it different arguments to play with.

sdasgup3: My point was the binary for the  target `stablehlo-add_test` will still be created if we do `bazel test examples/c++:stablehlo-add_test`. 

>> This will be needed if eventually we create binaries that take any args on input 

I see your point. In that case it will be a different target capable, with potentially different source file than what is provide now, with the provision of accepting input args.

fzakaria: in this example it doesn't make sense since there is no use of input; but it's setting a template for other examples to follow.

sdasgup3 (approved):

Issue #741, opened 2022-12-11
Unify metadata modelling in CustomCallOp


burmako:
Over years, CustomCallOp has gained a lot of miscellaneous metadata fields (for a detailed account on the HLO side, see [hlo_instruction.cc](https://github.com/tensorflow/tensorflow/blob/5e015d95cfdf67ddac0b228aa3ff2a237be7c1bc/tensorflow/compiler/xla/hlo/ir/hlo_instruction.cc#L736-L811)). StableHLO models only a small part of this metadata, but it already looks pretty disorganized.

I think that we should unify all this metadata under a single dictionary attribute, similarly to what has been done in MHLO under `API_VERSION_TYPED_FFI`, so that the API of CustomCallOp could ideally look like: 1) `call_target_name`, 2) `inputs`, 3) implementation-defined `metadata`. Depending on the outcome of #629 and #740, we might want to throw in a few more fields, but they should have clear (i.e. non implementation-defined) semantics, or otherwise they should go into `metadata`.

PR #2088, opened 2024-03-14
[RFC] Standardize TanOp, CustomCallOp with typed FFI, and Tuple-collectives ops

abhigunj:
This RFC proposes standardizing  `TanOp` , `CustomCallOp` with typed FFI, and Tuple-collectives (`AllGatherOp`, `AllReduceOp`, `AllToAllOp`) for  HLO-StableHLO parity. 
Please review and provide you feedback.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -700,8 +700,7 @@ it only exists to establish data dependencies from `result` to `inputs`.
 
 Within each process group in the StableHLO process grid, concatenates the values
 of the `operands` tensors from each process along `all_gather_dim` and produces
-`results` tensors. When multiple `operands` are specified, the `all_gather` is
-performed on each operand.
+`results` tensors.
 
 The operation splits the StableHLO process grid into `process_groups` which is
 defined as follows:
@@ -722,13 +721,13 @@ Afterwards, within each `process_group`:
 
 #### Inputs
 
-| Label | Name                    | Type                                                               | Constraints |
-|-------|-------------------------|--------------------------------------------------------------------|-------------|
-| (I1)  | `operands`               | variadic number of tensors or per-tensor quantized tensors        | (C1), (C6)  |
-| (I2)  | `all_gather_dim`        | constant of type `si64`                                            | (C1), (C6)  |
-| (I3)  | `replica_groups`        | 2-dimensional tensor constant of type `si64`                       | (C2-C4)     |
-| (I4)  | `channel_id`            | constant of type `si64`                                            | (C5)        |
-| (I5)  | `use_global_device_ids` | constant of type `i1`                                              | (C5)        |
+| Label | Name                    | Type                                                        | Constraints |
+|-------|-------------------------|-------------------------------------------------------------|-------------|
+| (I1)  | `operands`               | variadic number of tensors or per-tensor quantized tensors | (C1), (C6)  |
+| (I2)  | `all_gather_dim`        | constant of type `si64`                                     | (C1), (C6)  |
+| (I3)  | `replica_groups`        | 2-dimensional tensor constant of type `si64`                | (C2-C4)     |
+| (I4)  | `channel_id`            | constant of type `si64`                                     | (C5)        |
+| (I5)  | `use_global_device_ids` | constant of type `i1`                                       | (C5)        |
 
 #### Outputs
 
@@ -776,8 +775,7 @@ Afterwards, within each `process_group`:
 
 Within each process group in the StableHLO process grid, applies a reduction
 function `computation` to the values of the `operands` tensors from each process
-and produces `results` tensors. When multiple `operands` are specified, the
-`all_reduce` is performed on each operand.
+and produces `results` tensors.
 
 The operation splits the StableHLO process grid into `process_groups` which is
 defined as follows:
@@ -801,13 +799,13 @@ Afterwards, within each `process_group`:
 
 #### Inputs
 
-| Label | Name                    | Type                                                                                   | Constraints |
-|-------|-------------------------|----------------------------------------------------------------------------------------|-------------|
-| (I1)  | `operands`               | variadic number of tensors or per-tensor quantized tensors                            | (C5), (C6)  |
-| (I2)  | `replica_groups`        | variadic number of 1-dimensional tensor constants of type `si64`                       | (C1-C3)     |
-| (I3)  | `channel_id`            | constant of type `si64`                                                                | (C4)        |
-| (I4)  | `use_global_device_ids` | constant of type `i1`                                                                  | (C4)        |
-| (I5)  | `computation`           | function                                                                               | (C5)        |
+| Label | Name                    | Type                                                             | Constraints |
+|-------|-------------------------|------------------------------------------------------------------|-------------|
+| (I1)  | `operands`               | variadic number of tensors or per-tensor quantized tensors      | (C5), (C6)  |
+| (I2)  | `replica_groups`        | variadic number of 1-dimensional tensor constants of type `si64` | (C1-C3)     |
+| (I3)  | `channel_id`            | constant of type `si64`                                          | (C4)        |
+| (I4)  | `use_global_device_ids` | constant of type `i1`                                            | (C4)        |
+| (I5)  | `computation`           | function                                                         | (C5)        |
 
 #### Outputs
 
@@ -859,9 +857,7 @@ Afterwards, within each `process_group`:
 Within each process group in the StableHLO process grid, splits the values of
 the `operands` tensors along `split_dimension` into parts, scatters the split
 parts between the processes, concatenates the scattered parts along
-`concat_dimension` and produces `results` tensors. When multiple `operands`
-are specified, the `all_to_all` is performed on each operand.
-
+`concat_dimension` and produces `results` tensors.
 The operation splits the StableHLO process grid into `process_groups` which is
 defined as follows:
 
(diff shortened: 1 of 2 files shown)


loganchien:
It seems that there are 3 different features proposed in this RFC.  And they look unrelated to each other IMHO.  May you split this into 3 RFCs?

Also, may you elaborate the meaning of typed FFI?  For "FFI", the closest abbreviation I can think of is "Foreign Function Interface".  But what is the relation between "making `backend_config` support typed attribute and FFI?

docs/spec.md
```diff
@@ -714,16 +715,16 @@ defined as follows:
...
-* `result@process = concatenate(operands@process, all_gather_dim)` for all
+* `results...@process = concatenate(operands...@process, all_gather_dim)` for all
   `process` in `process_group`.
 
 #### Inputs
 
 | Label | Name                    | Type                                         | Constraints |
 |-------|-------------------------|----------------------------------------------|-------------|
-| (I1)  | `operand`               | tensor or per-tensor quantized tensor        | (C1), (C6)  |
+| (I1)  | `operands`               | variadic number of tensors or per-tensor quantized tensors        | (C1), (C6)  |
```
mlevesquedion, line 727: Nit: adjust the table so the columns are aligned. (Same comment for other tables.)

docs/spec.md
```diff
@@ -699,8 +699,9 @@ it only exists to establish data dependencies from `result` to `inputs`.
 #### Semantics
 
 Within each process group in the StableHLO process grid, concatenates the values
-of the `operand` tensor from each process along `all_gather_dim` and produces a
-`result` tensor.
+of the `operands` tensors from each process along `all_gather_dim` and produces a
```
mlevesquedion, line 702:
```suggestion
of the `operands` tensors from each process along `all_gather_dim` and produces
```

docs/spec.md
```diff
@@ -699,8 +699,9 @@ it only exists to establish data dependencies from `result` to `inputs`.
...
+`results` tensors. When multiple `operands` are specified, the `all_gather` is
+performed on each operand.
```
mlevesquedion, line 704: And I'm assuming a result tensor is produced for each operand? Consider mentioning that explicitly if so. (Same comment below.)

docs/spec.md
```diff
@@ -774,8 +775,9 @@ Afterwards, within each `process_group`:
 #### Semantics
 
 Within each process group in the StableHLO process grid, applies a reduction
-function `computation` to the values of the `operand` tensor from each process
-and produces a `result` tensor.
+function `computation` to the values of the `operands` tensors from each process
+and produces a `results` tensors. When multiple `operands` are specified, the
```
mlevesquedion, line 779:
```suggestion
and produces `results` tensors. When multiple `operands` are specified, the
```

docs/spec.md
```diff
@@ -855,9 +857,10 @@ Afterwards, within each `process_group`:
 ![all_to_all](images/spec/all_to_all.svg)
 
 Within each process group in the StableHLO process grid, splits the values of
-the `operand` tensor along `split_dimension` into parts, scatters the split
+the `operands` tensors along `split_dimension` into parts, scatters the split
 parts between the processes, concatenates the scattered parts along
-`concat_dimension` and produces a `result` tensor.
+`concat_dimension` and produces a `results` tensors. When multiple `operands`
```
mlevesquedion, line 862:
```suggestion
`concat_dimension` and produces `results` tensors. When multiple `operands`
```

docs/spec.md
```diff
@@ -2421,7 +2424,7 @@ the XLA compiler. In the future, we are planning to unify this metadata
 | (I1)  | `inputs`              | variadic number of values                     |
 | (I2)  | `call_target_name`    | constant of type `string`                     |
 | (I3)  | `has_side_effect`     | constant of type `i1`                         |
-| (I4)  | `backend_config`      | constant of type `string`                     |
+| (I4)  | `backend_config`      | constant of type `string` dictionary with key of type `string` and values of type `Value`                   |
```
mlevesquedion, line 2427:
```suggestion
| (I4)  | `backend_config`      | dictionary with key of type `string` and values of type `Value`                   |
```

ezhulenev:
Re: FFI - it is indeed a "foreign function interface" (https://github.com/openxla/xla/tree/main/xla/ffi): a new C/C++ API (with PjRt-style C API guarantees) for authoring custom call implementations and linking them with the XLA runtime which replaces custom calls with `void**` arguments and `char*` backend config

docs/spec.md
```diff
@@ -2437,8 +2440,8 @@ the XLA compiler. In the future, we are planning to unify this metadata
 %results = "stablehlo.custom_call"(%input0) {
   call_target_name = "foo",
   has_side_effect = false,
-  backend_config = "bar",
-  api_version = 1 : i32,
+  backend_config = {bar = 42 : i32},
+  api_version = 4 : i32,
```
mlevesquedion, line 2444: This change doesn't seem related to the backend_config change. Consider reverting it.

docs/spec.md
```diff
@@ -2437,8 +2440,8 @@ the XLA compiler. In the future, we are planning to unify this metadata
 %results = "stablehlo.custom_call"(%input0) {
   call_target_name = "foo",
   has_side_effect = false,
-  backend_config = "bar",
-  api_version = 1 : i32,
+  backend_config = {bar = 42 : i32},
```
mlevesquedion, line 2443: I'm worried that this is going to break a lot of stuff. What will our approach be for compatibility? Should we introduce a new attribute and deprecate backend_config?

mlevesquedion (commented):
Did a quick pass. I agree with comments from others that maybe this should be 3 separate RFCs? Seems like these are independent changes.

docs/spec.md
GleasonK: Could you elaborate? I don't think this poses any compatibility risks, like with any new feature, if you use the new feature it isn't forward compatible. And for backward_compat, <APIv4,dictionary> is an option, and <APIv3, string> will still work as well.

docs/spec.md
abhigunj: `4` for is new  API_VERSION_TYPED_FFI enum  at https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/StablehloEnums.td#L88  
For `API_VERSION_TYPED_FFI` custom calls `backend_config` must be a dictionary attribute

rfcs/20240312-standardize-tan-customcall-tuplecollective-ops.md
```diff
@@ -0,0 +1,69 @@
+# [RFC] Standardize TanOp, CustomCallOp with typed FFI, and tuple-collectives ops
```
GleasonK, line 1: Opening comment here since GH replies are difficult to keep track of:

@loganchien:
It seems that there are 3 different features proposed in this RFC. And they look unrelated to each other IMHO. May you split this into 3 RFCs?

@mlevesquedion:
I agree with comments from others that maybe this should be 3 separate RFCs? Seems like these are independent changes.

GleasonK: The common theme here is "Standardize successful MHLO features into StableHLO" -- aka HLO/MHLO/StableHLO parity for these features. In that regard we could split that into 3 separate RFCs, but there would be some duplication of motivation. I'm on board with either.

docs/spec.md
mlevesquedion: `API_VERSION_TYPED_FFI` is not defined in `StablehloEnums.td`. Did you mean to link to the MHLO definition?

Consider adding a comment explaining this, magic constants are hard to understand.

docs/spec.md
mlevesquedion: Agreed, didn't realize we were supporting both.

docs/spec.md
mlevesquedion: Sorry, I didn't realize that the `api_version` and `backend_config` fields were related. Maybe this should be explained in the spec and in the RFC? I agree in this case there is no concern as we are merely adding a new feature.

(My concern was that I thought we were going to just switch completely from strings to dictionaries. That would have been a breaking change, and it would have been unclear how to deserialize old bytecode.)

rfcs/20240312-standardize-tan-customcall-tuplecollective-ops.md
loganchien: I suggested to split this into 3 RFCs because they have different complexity.  Some portion can be merged faster.  Some need thorough review.

`TanOp` is trivial.  IMO, it can be on fast track.  I don't see the reason not merging it ASAP.  However, in the RFC, we should mention how to maintain forward/backward compatibility for `TanOp` (p.s. it  should be trivial as lowering `TanOp(x)` to `SinOp(x) / CosOp(x)` for backward compatibility).

I am supportive to the `CustomCallOp` change, but I would like to see more elaboration on rationale and refinement on wordings.  For example, I think instead of copying HLO terminology or define spec as existing implementation, it will be better to focus the wording on the properties of the construct.  Instead of using "FFI", we can just describe the feature as "extend `backend_config` to take a `DictionaryAttr`", etc.

For the collective ops, I don't have any idea.

docs/spec.md
abhigunj: spec says : "`backend_config` and `api_version` may be used to provide additional implementation-defined metadata"  so don't need change here? but  I will add more details in RFC to make it clear.

mlevesquedion: IMO it would be helpful to provide more information on what the `api_version` and `backend_config` are intended for and how they relate to each other. I think "backend_config and api_version may be used to provide additional implementation-defined metadata" is too broad. In MHLO there is quite a bit more information: https://github.com/tensorflow/tensorflow/blob/54a599afec6e0fc4afb48385e0cd4e0b8e512cd0/third_party/xla/xla/mlir_hlo/mhlo/IR/hlo_ops.td#L2514-L2526

docs/spec.md
abhigunj: end up removing this part, because `operands..` and `results..` by convention tell the same. I see other specs usually don't explicitly describe variadic inputs behavior.

docs/spec.md
abhigunj: changed it to "attribute dictionary"

docs/spec.md
abhigunj: Thanks for the feedback, added a comment to explain "api_version" constant.

docs/spec.md
abhigunj: that information will be added to the op def, not sure about the spec considering those are implementation details. Also, if we talk about `API_VERSION_TYPED_FFI` in the spec we will need to provide details for all `api_version`s ?

@GleasonK  / @sdasgup3  what is your opinion?

mlevesquedion: IMO this information is relevant to have in both places. I do think this belongs in the spec given that the value of `api_version` influences the meaning of `backend_config`.

Curious to hear what others think!

sdasgup3: I totally agree that the section in the spec which talks about `backend_config`, `api_version`  are  vague and not an ideal state to be in. But I would say that it is "intentional". That is mainly because at the point of writing the section there were various unknowns like https://github.com/tensorflow/mlir-hlo/blob/e7b1891ba261c2ab1e74b6979d16fd6d648a1232/mhlo/IR/hlo_ops.td#L2534 and #741.  At some point we thought about enumerating all the enums for `api_version`, but backed out for reasons like https://github.com/tensorflow/mlir-hlo/blob/e7b1891ba261c2ab1e74b6979d16fd6d648a1232/mhlo/IR/hlo_ops_enums.td#L110 (mainly) and lack on proper documentation on available enums which would have forced us to write something vague as we did in https://github.com/openxla/stablehlo/blob/main/docs/spec.md#rng_bit_generator. The outcome we decided was call out "backend_config" and "apli_version" as measn to provide implementation defined metadata together with #741.

My proposal: How about we keep the current state and in very near future  focus on  #741 to resolve the open questions.

abhigunj:
1. **TanOp** :  https://github.com/openxla/stablehlo/pull/2101

2. **CustomCallOp**: https://github.com/openxla/stablehlo/pull/2097

3. **Collectives ops**:  https://github.com/openxla/stablehlo/pull/2099 


Created separate 3 RFCs. Please, refer those to review and provide your feedback. Thank you.
closing this PR. Open comment will be addressed in respective RFC after the conclusion.

PR #2090, opened 2024-03-14
Add support of tensor types with unit dimension prefix in EvalSlice

mvpant:
Edit: Sure, sorry for missing description.

Currently our small research team experimenting with BERT model represented in number of *-hlo dialects and wants to simplify it in terms of variety of operators. This PR fixes an issue we stumbled upon:

```
%42 = stablehlo.constant dense<"..."> : tensor<1x512xi64>
...
%66 = stablehlo.slice %42 [0:1, 0:128] : (tensor<1x512xi64>) -> tensor<1x128xi64> 
```
Previous implementation of EvalSlice can't handle such case -- tensor type prefixed with unit dimension(s) i.e. 1x128.

This PR adds support of the above case and can slice from any position, e.g.

```
  %0 = stablehlo.constant dense<[[[[1, 2], [3, 4]], [[5, 6], [7, 8]], [[9, 10], [11, 12]]]]> : tensor<1x3x2x2xi64>
  %1 = "stablehlo.slice"(%0) {
    start_indices = array<i64: 0, 1, 1, 0>,
    limit_indices = array<i64: 1, 2, 2, 2>,
    strides = array<i64: 1, 1, 1, 1>
  } : (tensor<1x3x2x2xi64>) -> tensor<1x1x1x2xi64>
```
is folded to 
```
%1 = stablehlo.constant dense<[[[[7, 8]]]]> : tensor<1x1x1x2xi64>
```

### DIFF
--- a/stablehlo/tests/stablehlo_refine_shapes.mlir
+++ b/stablehlo/tests/stablehlo_refine_shapes.mlir
@@ -424,6 +424,22 @@ func.func @eval_slice() -> tensor<2xi64> {
 
 // -----
 
+// CHECK-LABEL: func @eval_slice_wild_stride
+func.func @eval_slice_wild_stride() -> tensor<1x1x1xi64> {
+  // CHECK-NOT: stablehlo.slice
+  // CHECK: [[RESULT:%.*]] = stablehlo.constant dense<2> : tensor<1x1x1xi64>
+  // CHECK: return [[RESULT]]
+  %0 = stablehlo.constant dense<[[[1, 2], [3, 4]]]> : tensor<1x2x2xi64>
+  %1 = "stablehlo.slice"(%0) {
+    start_indices = array<i64: 0, 0, 1>,
+    limit_indices = array<i64: 1, 1, 2>,
+    strides = array<i64: 99, 42, 1>
+  } : (tensor<1x2x2xi64>) -> tensor<1x1x1xi64>
+  func.return %1 : tensor<1x1x1xi64>
+}
+
+// -----
+
 // CHECK-LABEL: func @eval_slice_unit_prefix
 func.func @eval_slice_unit_prefix() -> (tensor<1x1x1x2xi64>, tensor<1x1x1x2xi64>, tensor<1x1x1x2xi64>) {
   // CHECK-NOT: stablehlo.slice
--- a/stablehlo/tests/stablehlo_refine_shapes.mlir
+++ b/stablehlo/tests/stablehlo_refine_shapes.mlir
@@ -424,6 +424,54 @@ func.func @eval_slice() -> tensor<2xi64> {
 
 // -----
 
+// CHECK-LABEL: func @eval_slice_unit_prefix
+func.func @eval_slice_unit_prefix() -> (tensor<1x1x1x2xi64>, tensor<1x1x1x2xi64>, tensor<1x1x1x2xi64>) {
+  // CHECK-NOT: stablehlo.slice
+  // CHECK: [[RESULT1:%.*]] = stablehlo.constant dense<{{\[\[\[}}[1, 2]]]]> : tensor<1x1x1x2xi64>
+  // CHECK: [[RESULT2:%.*]] = stablehlo.constant dense<{{\[\[\[}}[7, 8]]]]> : tensor<1x1x1x2xi64>
+  // CHECK: [[RESULT3:%.*]] = stablehlo.constant dense<{{\[\[\[}}[11, 12]]]]> : tensor<1x1x1x2xi64>
+  // CHECK: return [[RESULT1]], [[RESULT2]], [[RESULT3]]
+  %0 = stablehlo.constant dense<[[[[1, 2], [3, 4]], [[5, 6], [7, 8]], [[9, 10], [11, 12]]]]> : tensor<1x3x2x2xi64>
+
+  %1 = "stablehlo.slice"(%0) {
+    start_indices = array<i64: 0, 0, 0, 0>,
+    limit_indices = array<i64: 1, 1, 1, 2>,
+    strides = array<i64: 1, 1, 1, 1>
+  } : (tensor<1x3x2x2xi64>) -> tensor<1x1x1x2xi64>
+
+  %2 = "stablehlo.slice"(%0) {
+    start_indices = array<i64: 0, 1, 1, 0>,
+    limit_indices = array<i64: 1, 2, 2, 2>,
+    strides = array<i64: 1, 1, 1, 1>
+  } : (tensor<1x3x2x2xi64>) -> tensor<1x1x1x2xi64>
+
+  %3 = "stablehlo.slice"(%0) {
+    start_indices = array<i64: 0, 2, 1, 0>,
+    limit_indices = array<i64: 1, 3, 2, 2>,
+    strides = array<i64: 1, 1, 1, 1>
+  } : (tensor<1x3x2x2xi64>) -> tensor<1x1x1x2xi64>
+
+  func.return %1, %2, %3 : tensor<1x1x1x2xi64>, tensor<1x1x1x2xi64>, tensor<1x1x1x2xi64>
+}
+
+// -----
+
+// CHECK-LABEL: func @eval_slice_non_unit_prefix
+func.func @eval_slice_non_unit_prefix() -> tensor<1x2x1xi64> {
+  // CHECK: stablehlo.constant {{.*}} : tensor<1x2x2xi64>
+  // CHECK: [[RESULT:%.*]] = stablehlo.slice{{.*}}
+  // CHECK: return [[RESULT]]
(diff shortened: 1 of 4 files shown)


GleasonK:
Hello - Thank you for the contribution! Could you add a PR description with a bit of information on the feature change / use case? Will make it easier to review / for anyone to view PR and get a high level understanding.

stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -577,19 +578,48 @@ struct EvalSliceOpPattern : public OpRewritePattern<SliceOp> {
   LogicalResult matchAndRewrite(SliceOp op,
                                 PatternRewriter& rewriter) const override {
     auto resultType = op.getType();
-    if (!resultType.hasRank() || resultType.getRank() != 1)
-      return rewriter.notifyMatchFailure(op, "expected 1-dimensional type");
+    if (!resultType.hasRank() || resultType.getRank() < 1)
+      return rewriter.notifyMatchFailure(
+          op, "expected non-0 ranked tensor result type");
+
+    auto operand = op.getOperand().dyn_cast<TypedValue<RankedTensorType>>();
```
GleasonK, line 585: I believe this is enforced by op definition now (we've removed unranked tensors):

```
TypedValue<RankedTensorType> operand = op.getOperand();
RankedTensorType operandType = operand.getType();
```

Can then remove the `!operand` error check as well, and replace all uses of `op.getOperand() --> operand` and `operand.getType() --> operandType` in this function, there's a few redundant calls to each.

stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -577,19 +578,48 @@ struct EvalSliceOpPattern : public OpRewritePattern<SliceOp> {
...
+    if (!operand || !operand.getType().hasStaticShape())
+      return rewriter.notifyMatchFailure(
+          op, "expected operand with static ranked tensor type");
+
+    // A ranked tensor type with unit dimension prefix of R-1 size is physically
+    // compatible with 1-dimensional type.
+    auto resultShape = resultType.getShape();
+    for (const auto& dimSize : resultShape.drop_back(1)) {
```
GleasonK, line 593: For consistency with later code:

```
if (!llvm::all_of(resultType.getShape().drop_back(1),
                        [](int64_t s) { return s == 1; }))
   return rewriter.notifyMatchFailure(
            op, "expected 1-dimensional compatible result type");
```

stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -577,19 +578,48 @@ struct EvalSliceOpPattern : public OpRewritePattern<SliceOp> {
...
+    SmallVector<APSInt> operandData;
+    if (failed(hlo::matchInts(op.getOperand(), operandData)))
       return rewriter.notifyMatchFailure(op, "expected constant operand");
 
-    int64_t start = op.getStartIndices()[0];
-    int64_t limit = op.getLimitIndices()[0];
-    int64_t stride = op.getStrides()[0];
+    assert(llvm::all_of(op.getStrides().drop_back(),
+                        [](int64_t s) { return s == 1; }) &&
+           "all strides but last must be 1");
```
GleasonK, line 605: I need to revisit this ops verifier to see if this is guaranteed. If this is a part of the op verifier (dimSize 1 in result shape must have stride 1) then I don't think we need this assert. If this isn't verified, then I think asserting may lead to some crashes here.

GleasonK (commented):
Mostly LGTM, a few nits

stablehlo/tests/stablehlo_refine_shapes.mlir
```diff
@@ -424,6 +424,54 @@ func.func @eval_slice() -> tensor<2xi64> {
 
 // -----
 
+// CHECK-LABEL: func @eval_slice_unit_prefix
+func.func @eval_slice_unit_prefix() -> (tensor<1x1x1x2xi64>, tensor<1x1x1x2xi64>, tensor<1x1x1x2xi64>) {
```
GleasonK, line 428: Could you add a PR description with a bit of information on the feature change / use case? Will make it easier to review / for anyone to view PR and get a high level understanding.

stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -577,19 +578,48 @@ struct EvalSliceOpPattern : public OpRewritePattern<SliceOp> {
...
+    auto resultShape = resultType.getShape();
+    for (const auto& dimSize : resultShape.drop_back(1)) {
+      if (dimSize != 1)
+        return rewriter.notifyMatchFailure(
+            op, "expected 1-dimensional result type");
+    }
 
-    SmallVector<APSInt> operand;
-    if (failed(hlo::matchInts(op.getOperand(), operand)))
+    SmallVector<APSInt> operandData;
```
ghpvnist, line 599: With the above comment, this name can be reverted back to `operand`.

ghpvnist (commented):
Also lgtm, with some comments

stablehlo/transforms/StablehloRefineShapes.cpp
mvpant: does it also mean that `!resultType.hasRank()` is excessive and can be removed? (i couldn't derive information about RankedTensorType enforcement from HLO_Tensor definition, but SliceOp has the same type for operand and result, so result is also a (always) RankedTensorType?)

stablehlo/transforms/StablehloRefineShapes.cpp
mvpant: That is also my concern expressed in this assert, wanted to restrict possible cases. But i assume that writing something like this:
```
%1 = "stablehlo.slice"(%0) {
    start_indices = array<i64: 0, 0>,
    limit_indices = array<i64: 1, 1>,
    strides = array<i64: 99, 42>
  } : (tensor<1x3xi64>) -> tensor<1x1xi64>
```
is probably valid as whatever stride is it never does 2nd iteration. Therefore assert should be removed?

stablehlo/transforms/StablehloRefineShapes.cpp
mvpant: Isn't `operand` name is still taken by
```
TypedValue<RankedTensorType> operand = op.getOperand();
```

stablehlo/transforms/StablehloRefineShapes.cpp
mvpant: removed `!resultType.hasRank()` check

stablehlo/transforms/StablehloRefineShapes.cpp
GleasonK: Looks like we only use stride to enforce the result shape:

`shape(result) = ceil((limit_indices - start_indices) / strides)`

As you point out, the following is a valid program that we would (maybe?) want to fold:

```
module {
  func.func @slice_test(%arg0: tensor<1x3xi64>) -> tensor<1x1xi64> {
    %0 = stablehlo.slice %arg0 [0:1:99, 0:1:42] : (tensor<1x3xi64>) -> tensor<1x1xi64>
    return %0 : tensor<1x1xi64>
  }
}
```

With that said I'm ok with either:

1. Support folding the above program: Remove this assert.
2. Don't support folding this program: Replace the assert with a `return notifyMatchFailure`

Will leave the decision up to you, 2 is OK since it is non-standard IR. We don't _need_ to support this program until there is a use case for it. Regardless we don't want to assert/crash on valid IR that isn't supported by this pass

stablehlo/transforms/StablehloRefineShapes.cpp
GleasonK: I think this comment may be one line off:

```
if (failed(hlo::matchInts(op.getOperand(), operandData)))
==>
if (failed(hlo::matchInts(operand, operandData)))
```

GleasonK (approved):
LGTM once open comments are addressed

mvpant:
Added missing dependency MLIRDialectUtils

ghpvnist (approved):
LGTM. Thank you for your contribution!

stablehlo/transforms/CMakeLists.txt
```diff
@@ -45,6 +45,7 @@ add_mlir_dialect_library(StablehloPasses
   MLIRArithDialect
   MLIRAsmParser
   MLIRComplexDialect
+  MLIRDialectUtils
```
GleasonK, line 48: We need to port this to BUILD.bazel too, can add around here

https://github.com/openxla/stablehlo/blob/6909631c594a0ac4f26a4e9bebbaa0f30740290c/BUILD.bazel#L964


```
 "@llvm-project//mlir:DialectUtils",
```

GleasonK (approved):

PR #2094, opened 2024-03-14
ConvolutionOP verifier : allow rhs per_axis quantized and result per_tensor quantized

abhigunj:
Previous check was restrictive and causing test failures during integration.

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -3506,10 +3506,9 @@ LogicalResult verifyConvolutionOp(
   auto resultQPAType =
       resultQType.dyn_cast<quant::UniformQuantizedPerAxisType>();
   if (!rhsQPAType && resultQPAType) {
-    return emitOptionalError(location,
-                             "rhs and result are of mixed per_tensor and "
-                             "per_axis quantized tensor type ",
-                             rhsType, " and ", resultType);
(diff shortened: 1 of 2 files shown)


abhigunj:
cc : @mlevesquedion

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3502,16 +3502,16 @@ LogicalResult verifyConvolutionOp(
   if (noneQuantized<quant::UniformQuantizedPerAxisType>(typeEntriesPerAxis))
     return success();
   // convolution_c31
-  if (!allQuantized<quant::UniformQuantizedPerAxisType>(typeEntriesPerAxis)) {
+  auto rhsQPAType = rhsQType.dyn_cast<quant::UniformQuantizedPerAxisType>();
+  auto resultQPAType =
+      resultQType.dyn_cast<quant::UniformQuantizedPerAxisType>();
+  if (!rhsQPAType && resultQPAType) {
     return emitOptionalError(location,
                              "rhs and result are of mixed per_tensor and "
```
sdasgup3, line 3510: How about we update the error messae to match the constraint like 
```
"per-tensor `rhs` expects per-tensor `result`, but got ", rhsType, " and ", resultType, " respectively." 
```

sdasgup3 (approved):

PR #2096, opened 2024-03-15
[RFC] Add algorithm to dot_general's parameters in the StableHLO specification

tdanyluk:


rfcs/20240315-algorithm-support.md
```diff
@@ -0,0 +1,165 @@
...
+
+### dot_general
+
+#### Semantics
+
+*Add these lines:*
+
+`algorithm` defines the main properties of the algorithm used to implement the
+dot operation, which also defines the precision. If the algorithm value is set
+(`!= UNSET`), then the `precision_config` is ignored.
```
sdasgup3, line 38: Does that also mean that if the algorithm is not set then precision_config will take effect? If yes, can we clarify that as well.

rfcs/20240315-algorithm-support.md
```diff
@@ -0,0 +1,165 @@
...
+
+In general the first two types in the algorithm name are `AType` and `BType` -
+the precisions that the LHS and RHS of the operation are rounded to, and the
+third one is `AccumType` - the accumulation type. These types are independent
+from the storage types of the inputs and the output (but not all combinations
+are supported). Some algorithms have other suffixes too.
+
+This is the current list of algorithms:
```
sdasgup3, line 46: I am wondering why this does not cover the accumulation storage for integer type. The same questions applies for precision_config as well.

tdanyluk: I wanted to add only the algorithms which are important to us for now.
Of course if it becomes needed, we can add algorithms like INT16_INT16_INT32 (with INT32 being the accum type).

rfcs/20240315-algorithm-support.md
tdanyluk: Added the requested clarification.

rfcs/20240315-algorithm-support.md
```diff
@@ -0,0 +1,166 @@
...
+precision.
+
+Having considered the mentioned advantages, we wanted to start by adding
+something more simple for now, and just add support for a few algorithms which
+are important to disambiguate. (All earlier types and precisions would remain
+supported for now). Also, we think that algorithm names are easier to refer to
+(although this could be probably mitigated by introducing a special syntax to
+write the structs). We also think that algorithms may have more different
+properties which would need new struct members, but if we use an enum, we can
+just append them to the name.
```
GleasonK, line 160: +1, also structs may require more verification if not all combinations are valid.

rfcs/20240315-algorithm-support.md
```diff
@@ -0,0 +1,166 @@
...
+//           [[1, 2],
+//            [3, 4]],
+//           [[5, 6],
+//            [7, 8]]
+//          ]
+```
+
+## Alternatives considered
+
+### Adding new Precision values instead of the algorithm property
```
GleasonK, line 134: I'm debating on this alternative.

On one hand it's super nice to not introduce a separate attribute, on the other, I like that we don't need an array when specifying an algorithm.

I think I still prefer this alternative, if we go this route we also don't need a sentinel enum value for `<UNSET>` since that can be represented by empty array. We can handle per-op discrepancy in op-verification pretty easily.

tdanyluk: I think we can do what you recommended for now.
(In XLA, we used a separate algorithm flag, but I think we can handle this difference at the conversion from StableHLO to HLO.)
It should work for now and if anyone needs something more complex we can iterate further.

rfcs/20240315-algorithm-support.md
```diff
@@ -0,0 +1,166 @@
...
+* `UNSET`: There is no restriction on the algorithm, the precision is decided
+  based on the `precision_config`. Use this for every type combination that
+  doesn't have a specific algorithm.
+* `ANY_F8_ANY_F8_F32`: The inputs are used with the full precision of their
+  storage type (which must be an 8-bit floating point type - the same for both
+  operands). The accumulator type is F32.
+* `ANY_F8_ANY_F8_F32_FAST_ACCUM`: The inputs are used with the full precision of
+    their storage type (which must be an 8-bit floating point type - the same
+    for both operands). The accumulator type is formally F32, but the
+    intermediate results will not periodically be promoted to a higher
```
sdasgup3, line 58: >> The accumulator type is formally F32, but the intermediate results will not periodically be promoted to a higher precision. 


For a type like `ANY_F8_ANY_F8_F32`, the accumulation would happen in `F32`. Does that mean the intermediate results will always be promoted to F32. Something like, 
```
accumulation += promote_to_f32(lhs) * promote_to_f32(rhs) 
```

whereas for `ANY_F8_ANY_F8_F32`, the behavior is:

```
accumulation += lhs *rhs 
```

Also what does formally mean in `accumulator type is formally F32`? 

what is that higher precision in `...will not periodically be promoted to a higher precision`? F32?

rfcs/20240315-algorithm-support.md
```diff
@@ -0,0 +1,166 @@
...
+*Add these lines:*
+
+`algorithm` defines the main properties of the algorithm used to implement the
+dot operation, which also defines the precision. If the algorithm value is set
+(`!= UNSET`), then the `precision_config` is ignored. If the algorithm value is
+not set (`= UNSET`), then the `precision_config` will take effect.
+
+In general the first two types in the algorithm name are `AType` and `BType` -
+the precisions that the LHS and RHS of the operation are rounded to, and the
+third one is `AccumType` - the accumulation type. These types are independent
```
changm, line 43: Does it mean that the output has the precision of `AccumType`?

rfcs/20240315-algorithm-support.md
```diff
@@ -0,0 +1,166 @@
...
+from the storage types of the inputs and the output (but not all combinations
```
changm, line 44: Is there a list of combinations that are supported or explicitly unsupported?

rfcs/20240315-algorithm-support.md
```diff
@@ -0,0 +1,166 @@
...
+* `TF32_TF32_F32`: The inputs are casted to TF32 precision and the accumulator
+  type is F32.
+* `TF32_TF32_F32_X3`: An algorithm which uses 3 TF32_TF32_F32 matmuls to achieve
+  a higher precision ("similar" to F32_F32_F32).
+* `F32_F32_F32`: The inputs are used with / casted to F32 precision and the
+  accumulator type is also F32.
+* `F64_F64_F64`: The inputs are used with / casted to F64 precision and the
+  accumulator type is also F64.
+
+In general, it is not guaranteed that the each algorithm is supported on each
```
changm, line 81: Is there any METADATA in the output that has to be plumbed through to all StableHLO ops now?

As in, if the output of `dot_general` goes into a `add`, does the `add` now need to have a precision config? 

As in, does precision config have to go through all the StableHLO ops now?

rfcs/20240315-algorithm-support.md
tdanyluk: The resolution is to reuse the precision property instead of adding a new algorithm property.
But we still have to add new enum values for precision.

rfcs/20240315-algorithm-support.md
```diff
@@ -0,0 +1,166 @@
...
+third one is `AccumType` - the accumulation type. These types are independent
+from the storage types of the inputs and the output (but not all combinations
+are supported). Some algorithms have other suffixes too.
+
+This is the current list of algorithms:
+
+* `UNSET`: There is no restriction on the algorithm, the precision is decided
+  based on the `precision_config`. Use this for every type combination that
+  doesn't have a specific algorithm.
+* `ANY_F8_ANY_F8_F32`: The inputs are used with the full precision of their
```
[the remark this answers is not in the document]
tdanyluk: Is the assumption here that the inputs are already f8?
- it is. XLA would error out if we request ANY_F8_ANY_F8_F32 for anything other than F8.* storage types.

" The inputs are used with the full precision of their
  storage type (which must be an 8-bit floating point type - the same for both
  operands). The accumulator type is F32."

rfcs/20240315-algorithm-support.md
```diff
@@ -0,0 +1,166 @@
...
+platforms and even between GPU types.
+
+We also want to make the API flexible enough to support additional algorithms
+that might be faster on certain hardware
+([6xBF16](https://arxiv.org/pdf/1904.06376.pdf), 3xTF32, etc) that can't fit
+into the limited precision_config values of `DEFAULT`, `HIGH` and `HIGHEST`. For
+example "6xBF16" provides similar precision to F32, but about 2x performance on
+some GPUs.
+
+When an algorithm is not supported by an accelerator, the program should fail
```
rhundt, line 23: Maybe I miss the nuances and I apologize if I do. I feel this choice impacts fungibility. A vendor may want to totally specialize for a specific algorithm but a customer may want to run an unmodified JAX on HW from vendors A and B.   Maybe there should be a way for vendors (or JAX) to specify "fall-back" types, sort of the reverse mapping from special type to 'DEFAULT', 'HIGH' and 'HIGHEST'? (I wonder why we don't offer 'LOW' and 'LOWEST' ;-)

rfcs/20240315-algorithm-support.md
tdanyluk: This was added to support CUBLASLT_MATMUL_DESC_FAST_ACCUM, and I think that Triton's
maxNumImpreciseAcc=32 setting may be similar.

> Also what does formally mean in accumulator type is formally F32?
By "formally", I mean that we pass F32 compute type to cuBLASlt.

> what is that higher precision in ...will not periodically be promoted to a higher precision? F32?
I think so. I copied this from NVIDIA's page.

The operands are not promoted as in your example.
The difference is that ANY_F8_ANY_F8_F32 is roughly equivalent to accumulating every single value into an F32 accumulator, while ANY_F8_ANY_F8_F32_FAST_ACCUM may allow to accumulate a lot of values in F16 and promote that to F32 only rarely - so it could be noticeably less precise.

If we need more info we can follow up with NVIDIA, but I think that this is perhaps enough for now.
@reedwm maybe knows better

rfcs/20240315-algorithm-support.md
tdanyluk: I don't think so, the intention is that we would only set the precision for the dot (and later convolution) operations.
This would basically modify what types do we use in the tensor cores or other hw which does the matmul.
(So even if we have F32 storage type, we can cast it to TF32 or BF16 when calculating with it)

rfcs/20240315-algorithm-support.md
tdanyluk: @reedwm FYI. I think you (also) advocated for the new algorithm field, because it's nicer than reusing the operand precisions, but Kevin said that it's easier to get this RFC accepted if we just reuse it.
I guess it doesn't matter that much anyway. Also it may later better support algorithms where we set different precision for the 2 operands of the matmul.

rfcs/20240315-algorithm-support.md
tdanyluk: Not really, we could accumulate in f32, but save the result in f16 in the end.

It's like this:

```
AccumType accum = 0;
accum += (AType) a * (BType) b;
accum += (AType) a * (BType) b;
accum += (AType) a * (BType) b;
...
OutputStorageType output = accum;
```

So if the output storage type can represent every value of the AccumType, then it will have "the same precision",
But if not, then it will have less.

rfcs/20240315-algorithm-support.md
tdanyluk: It's not part of the RFC, but each platform can define those.
See here for GPU:
https://source.corp.google.com/piper///depot/google3/third_party/tensorflow/compiler/xla/service/algorithm_util.cc;rcl=631817139;l=141

rfcs/20240315-algorithm-support.md
sdasgup3: Thanks!

>> I think so. I got this wording from the CUBLASlt docs.

How about we make this explicit "will not periodically be promoted to a F32 precision"?

rfcs/20240315-algorithm-support.md
tdanyluk: I think that JAX's request was that on the XLA side we should require exact algorithms.
But on the JAX side, they could write some fallback logic using the current GPU/TPU type, etc.
I thought that it's not needed to specify the fallback in the stableHLO, JAX could just decide the final algorithm before writing the stableHLO.

Maybe @hawkinsp can comment on this?
(I think he did in some other forum, but I can't find it anymore.)

hawkinsp: Yes. If I specify an exact algorithm, I want that exact algorithm. Otherwise I get, e.g., unexpected precision changes when moving code between platforms. I'd much rather I get an explicit error.

But I agree it makes sense to also allow less-specific choices as well.

hawkinsp: I'll note that this was specifically requested by one of our internal users, who found that the numerics of their computation completely changed when moving from TPU to GPU or vice versa, but the problem was almost entirely resolved when using the same matmul algorithm on both.

rfcs/20240315-algorithm-support.md
changm: ack thanks.

rfcs/20240315-algorithm-support.md
changm: Sounds good thanks, maybe clarify that its up the platforms to decide which combinations are supported?

rfcs/20240315-algorithm-support.md
changm: Hmm I'm still somewhat confused. In the example you have, because `output = accum`, an implicit cast is made. But the actual output of the dot operation is still the `AccumType`.?

Or are you saying code would be like

OutputStorageType dot_result = `stable_hlo.dot...accum_type=AccumType` so the implicit conversion is there anyway?

rfcs/20240315-algorithm-support.md
rhundt: Shouldn't we allow to control the behavior with a flag. Default could be something like "strict" vs "permissive" numerical accuracy, something akin to the -ffast-math option, except for data types?

rfcs/20240315-algorithm-support.md
tdanyluk: The input/output of the HLO level dot operation will be always the input/output storage types (=the types in the HLO).
But during the computation of the result another type may be used, which is denoted by AType/BType/AccumType.
For example if we could have a dot which has F32 for both inputs and the output, but internally the GPU will convert the inputs to TF32. (TF32_TF32_F32)
So similarly, if there are casts from the accumulator type to the output type, that should happen inside the "kernel" and not as a separate HLO operation or similar.

changm: ahh ack, thanks for the explanation. That makes sense.

rfcs/20240315-algorithm-support.md
tdanyluk: Do you think this flag should be in XLA or JAX?

majnemer: The problem with a flag is that it is global and thus does not nicely.

majnemer: It also gets a bit weird if you want to go from jax2tf. I think it needs to be serialized.

GleasonK: Does the user need to define that less specific is ok, or the compilation pipeline? This "fallback" notion could exist as an mlir pass if the pipeline can determine it

rfcs/20240315-algorithm-support.md
```diff
@@ -0,0 +1,166 @@
...
+from the storage types of the inputs and the output (but not all combinations
+are supported). Some algorithms have other suffixes too.
+
+This is the current list of algorithms:
+
+* `UNSET`: There is no restriction on the algorithm, the precision is decided
+  based on the `precision_config`. Use this for every type combination that
+  doesn't have a specific algorithm.
+* `ANY_F8_ANY_F8_F32`: The inputs are used with the full precision of their
+  storage type (which must be an 8-bit floating point type - the same for both
```
jingpu, line 53: Why must a "full precision" be 8bit floating point type? Do you assume that this algorithm is only applicable to fp8 matmul? What happens if someone use this algorithm for a fp32 matmul?

rfcs/20240315-algorithm-support.md
```diff
@@ -0,0 +1,166 @@
...
+* `F16_F16_F16`: The inputs are used with / casted to F16 precision and the
+  accumulator type is also F16.
+* `F16_F16_F32`: The inputs are used with / casted to F16 precision and the
+  accumulator type is F32.
+* `BF16_BF16_BF16`: The inputs are used with / casted to BF16 precision and the
+  accumulator type is also BF16.
+* `BF16_BF16_F32`: The inputs are used with / casted to BF16 precision and the
+  accumulator type is F32.
+* `BF16_BF16_F32_X3`: An algorithm which uses three BF16_BF16_F32 dot operations
+  per "tile" to achive a higher precision.
```
loganchien, line 69: It is unclear to me what "tile" means in this context.

rfcs/20240315-algorithm-support.md
```diff
@@ -0,0 +1,166 @@
...
+* `BF16_BF16_F32_X6`: An algorithm which uses six BF16_BF16_F32 dot operations
+  per "tile" to achive a higher precision ("similar" to F32_F32_F32).
+* `TF32_TF32_F32`: The inputs are casted to TF32 precision and the accumulator
```
loganchien, line 72: If this text is proposed, IMO, we must have a separate section to describe what `TF32` means (and the differences between `F32` and `TF32`).

rfcs/20240315-algorithm-support.md
```diff
@@ -0,0 +1,166 @@
...
+    precision.
+* `F16_F16_F16`: The inputs are used with / casted to F16 precision and the
+  accumulator type is also F16.
+* `F16_F16_F32`: The inputs are used with / casted to F16 precision and the
+  accumulator type is F32.
+* `BF16_BF16_BF16`: The inputs are used with / casted to BF16 precision and the
+  accumulator type is also BF16.
+* `BF16_BF16_F32`: The inputs are used with / casted to BF16 precision and the
+  accumulator type is F32.
+* `BF16_BF16_F32_X3`: An algorithm which uses three BF16_BF16_F32 dot operations
```
loganchien, line 68: What does `_X3` mean?

sbodenstein: Shouldn't this be `_3X`?

rfcs/20240315-algorithm-support.md
cantonios: This `ANY_F8_ANY_F8_F32_FAST_ACCUM` seems very specific to cuda.  How stable are results between different cuda versions and cards?  Is this any different than the existing `HIGH` precision option?

cantonios: IMO an "algorithm" should have well-defined semantics so that vendors can actually implement it in a consistent way.  This, on the other hand, seems like it's saying "I don't quite need F32 intermediate precision, but I want relatively high precision so give me the best you can", which to me would be the same as `HIGH`.  Do you really want the program to _fail_ if a vendor doesn't have a "Fast Accum" version and instead only had `ANY_F8_ANY_F8_F32`?

Otherwise, you may as well specify the algorithm as `CUBLASLT_MATMUL_DESC_FAST_ACCUM` if that's exactly what you wanted and want the program to fail otherwise.

reedwm: I'm not opposed to giving `ANY_F8_ANY_F8_F32_FAST_ACCUM` a more cublas-specific name. But it sounds like Triton can do something similar.

@nluehr can we get a more specific definition of what `CUBLASLT_MATMUL_DESC_FAST_ACCUM` does? The description states "intermediate results will not periodically be promoted to a higher precision". Does this mean that the flag never causes any accumulator value to be F32? Does this imply that even without this flag, values are initially accumulated in lower-than-fp32 precsision, but then "periodically" added to an fp32 accumulator?

cantonios: > But it sounds like Triton can do something similar.

"Similar" does not mean the same, so if it is different you could also have a `TRITON_MAX_NUM_IMPRECISE_ACC_32`, though again that sounds like it's less of a guaranteed precision/algorithm, and more like the existing `HIGH` precision config we already have.

rfcs/20240315-algorithm-support.md
tdanyluk: We held a meeting on the 9th of May, and had a discussion with @majnemer, where he advocated for using a struct instead, and he was quite convincing.
I think his main points were that:
- This could prevent combinatorial explosion of the enum types.
- The current enum based solution could get messy if we'll have examples where the "precision" of the lhs and rhs are not the same.

I guess that I'll try to write up how would that look like.
I think that it's OK that HLO uses enums now, we could handle this in the converter for now if we'll use a struct in StableHLO.

rfcs/20240315-algorithm-support.md
tdanyluk: Whatever order you prefer is fine. I thought that 3X is the same as X3, but maybe one of them is "grammatically more correct".

It means that we use 3 BF16xBF16->F32 matmuls to approximate something like an F32xF32->F32 matmul. (See this https://arxiv.org/pdf/1904.06376.pdf)
It is similar to this:
https://github.com/openxla/xla/blob/9c30e089b1e9f2ca60f56be2b5a39af1377e6cce/xla/service/gpu/ir_emitter_triton.cc#L1830

rfcs/20240315-algorithm-support.md
tdanyluk: It was defined by NVIDIA, is it enough to link to this?
https://blogs.nvidia.com/blog/tensorfloat-32-precision-format/

rfcs/20240315-algorithm-support.md
tdanyluk: Yeah it means here the "little" submatrix that the primitive dot operation outputs. Meaning that for example instead of using one tensor-core operation to output a 16x16 "tile", we use 3 or 6 such tensor-core operations and combine the results to achieve better precision.

rfcs/20240315-algorithm-support.md
tdanyluk: I was thinking of this:

```
struct DotAlgorithm {
  mlir::Type lhs_type; // Meaning that the LHS will be "casted" to this type before doing the operation. We could also call this a_type?
  mlir::Type rhs_type;
  mlir::Type accum_type;
  int lhs_component_count; // This can be >1 if we decompose the LHS to do an approximation using multiple primitive operations.
  int rhs_component_count;
  int num_primitive_ops; // This can be >1 if we decompose the LHS to do an approximation using multiple primitive operations.
  bool can_accumulate_in_lower_precision_for_some_steps; // We can refine this later. Corresponds to CUBLASLT_MATMUL_DESC_FAST_ACCUM.
};
```

Not sure if it should be DotAlgorithm or just Algorithm, not sure how well these generalize.

Here is how the previously proposed enum values could be described as structs:

```
TF32_TF32_F32 -> 
{lhs_type: TF32, 
 rhs_type: TF32,
 accum_type: F32,
 lhs_component_count: 1,
 rhs_component_count: 1,
 num_primitive_ops: 1,
 can_accumulate_in_lower_precision_for_some_steps: false}


BF16_BF16_F32_X6 ->
{lhs_type: BF16,
 rhs_type: BF16,
 accum_type: F32,
 lhs_component_count: 3,
 rhs_component_count: 3,
 num_primitive_ops: 6,
 can_accumulate_in_lower_precision_for_some_steps: false}


ANY_F8_ANY_F8_F32_FAST_ACCUM -> For example:
{lhs_type: F8E5M2,
 rhs_type: F8E5M2,
 accum_type: F32,
 lhs_component_count: 3,
 rhs_component_count: 3,
 num_primitive_ops: 6,
 can_accumulate_in_lower_precision_for_some_steps: true}
```

Regarding the verification:
Maybe we can do it in the stable_hlo to hlo conversion?

sdasgup3: >> Regarding the verification:

There does not seem to be any constraint among the different fields of the struct. If there are then prefer does the verification checks in StableHLO.

rfcs/20240315-algorithm-support.md
tdanyluk: Well ANY_F8_ANY_F8_F32 was defined only for F8 types, but actually we may rather use structs instead of enums as described in https://github.com/openxla/stablehlo/pull/2096#discussion_r1616185205, so this problem will not come up.

rfcs/20240315-algorithm-support.md
tdanyluk: In the new struct based proposal this would map to can_accumulate_in_lower_precision_for_some_steps=true.
(feel free to reopen if needed)

rfcs/20240315-algorithm-support.md
tdanyluk: added clarification

rfcs/20240315-algorithm-support.md
tdanyluk: Updated the RFC to reflect these changes.

tdanyluk:
I updated the RFC to reflect all the requested changes.

rfcs/20240315-algorithm-support.md
```diff
@@ -0,0 +1,181 @@
...
+the implementation is allowed to accumulate in lower precision for some steps
+(as per CUBLASLT_MATMUL_DESC_FAST_ACCUM).
+
+It is up to the implementations to decide which combinations are supported.
+
+Example `DotAlgorithm` attributes:
+
+```
+// Inputs are casted to tf32, and then accumulated in f32:
+{lhs_type = tf32,
```
loganchien, line 61: To specify `tf32` here, do you have to add `tf32` to the "TensorElementType" paragraph in https://github.com/openxla/stablehlo/blob/main/docs/spec.md#types?

tdanyluk: You are correct, added it now.

PR #2098, opened 2024-03-15
Augment example test with interpreter usage

sdasgup3:
Augmented the example test with interpreter usage demo.

nit: Fixed the dependency of `MLIRParser` on `StablehloReferenceApi` as API.cc uses `parseSourceFile`.

### DIFF
--- a/examples/c++/ExampleAdd.cpp
+++ b/examples/c++/ExampleAdd.cpp
@@ -54,28 +54,24 @@ int main() {
   mlir::Location loc = block_builder.getUnknownLoc();
 
   llvm::SmallVector<mlir::NamedAttribute, 10> attributes;
-  block_builder.create<mlir::stablehlo::AddOp>(loc, arguments, attributes)
-      .getOperation();
-
-  mlir::Operation* op = block_builder
-                            .create<mlir::stablehlo::AddOp>(
-                                loc, tensorType, arguments, attributes)
-                            .getOperation();
+  mlir::Operation* op =
+      block_builder.create<mlir::stablehlo::AddOp>(loc, arguments, attributes)
+          .getOperation();
   block_builder.create<mlir::func::ReturnOp>(loc, op->getResult(0));
 
   /** verify and dump the module **/
   assert(mlir::succeeded(mlir::verify(module.get())));
 
-  /* interpret the  with concrete inputs **/
-  auto inputValue1 = mlir::DenseElementsAttr::get(
-      tensorType, block_builder.getFloatAttr(tensorType.getElementType(),
-                                             static_cast<double>(10)));
-  auto inputValue2 = mlir::DenseElementsAttr::get(
-      tensorType, block_builder.getFloatAttr(tensorType.getElementType(),
-                                             static_cast<double>(20)));
-  auto expectedValue = mlir::DenseElementsAttr::get(
-      tensorType, block_builder.getFloatAttr(tensorType.getElementType(),
-                                             static_cast<double>(30)));
+  /* interpret the function "main" with concrete inputs **/
+  auto getConstValue = [&](double val) {
+    return mlir::DenseElementsAttr::get(
+        tensorType,
+        block_builder.getFloatAttr(tensorType.getElementType(), val));
+  };
+
+  auto inputValue1 = getConstValue(10.0);
+  auto inputValue2 = getConstValue(20.0);
+  auto expectedValue = getConstValue(30.0);
 
   mlir::stablehlo::InterpreterConfiguration config;
   auto results = evalModule(*module, {inputValue1, inputValue2}, config);
--- a/examples/c++/ExampleAdd.cpp
+++ b/examples/c++/ExampleAdd.cpp
@@ -14,10 +14,12 @@ limitations under the License.
 ==============================================================================*/
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/Quant/QuantOps.h"
+#include "mlir/IR/BuiltinAttributes.h"
 #include "mlir/IR/BuiltinOps.h"
 #include "mlir/IR/BuiltinTypes.h"
 #include "mlir/IR/Verifier.h"
 #include "stablehlo/dialect/StablehloOps.h"
+#include "stablehlo/reference/Api.h"
 
 int main() {
   mlir::MLIRContext context;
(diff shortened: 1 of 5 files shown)


examples/c++/README.md
```diff
@@ -5,5 +5,6 @@ for how to use StableHLO.
 
 Note: If you have a great example to highlight, we welcome contributions!
 
-* [stablehlo_add](./stablehlo_add.cpp): A simple example that demonstrates
-how to use the StableHLO library to add two numbers.
+* [stablehlo_add](./stablehlo_add.cpp): A simple example that demonstrates how to
+  * Use the StableHLO library to add two numbers.
+  * Interpret a StableHLO program using concrete inputs.
```
fzakaria, line 10: can you add a code block here that shows an example invocation and output?

stablehlo/reference/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ add_mlir_library(StablehloReferenceApi
 
   LINK_LIBS PUBLIC
   MLIRIR
+  MLIRParser
```
fzakaria, line 21: How did this come up?

fzakaria:
LGTM --

fzakaria:
I'm wondering if we should have a simple example of add that takes input and one that uses constants?
I'd lean on more examples than in the future keep expanding our sole tests.

For this PR -- I think the contribution is good though. 👍

fzakaria (approved):

stablehlo/reference/CMakeLists.txt
sdasgup3: Before this change the library target `StablehloReferenceApi` is used to build the binary `stablehlo-translate` [cs](https://github.com/openxla/stablehlo/blob/dc4bc72de7e483adf7c3d5fac9b14bec526b10ef/stablehlo/tools/CMakeLists.txt#L60). While building  that target the need for the dependency of `MLIRParser`  is satisfied recursively via `StablehloSerialization` ([here](https://github.com/openxla/stablehlo/blob/dc4bc72de7e483adf7c3d5fac9b14bec526b10ef/stablehlo/tools/CMakeLists.txt#L68) and [here](https://github.com/openxla/stablehlo/blob/dc4bc72de7e483adf7c3d5fac9b14bec526b10ef/stablehlo/dialect/CMakeLists.txt#L118)) and hence the problem is not exposed before.

The new target `stablehlo-add-test` used in the PR does not have dependency of `StablehloSerialization` which exposes the missing dependency of `MLIRParser` on `StablehloReferenceApi` via a link time error on undefined symbol `mlir::parseSourceFile` used in API.cc.

examples/c++/README.md
sdasgup3: Added.

sdasgup3:
>> I'd lean on more examples than in the future keep expanding our sole tests.

Makes sense to me. 

For this particualr PR: One of the motivation to create example is to demo feeding constant input to a stablehlo program using the APIs introduced in https://github.com/openxla/stablehlo/pull/2062. I agree that we should keep on iterate on this to make it simpler and intuitive to follow.

examples/c++/ExampleAdd.cpp
```diff
@@ -54,15 +57,27 @@ int main() {
...
-          .create<mlir::stablehlo::AddOp>(loc, arg0, arguments, attributes)
-          .getOperation();
+  mlir::Operation* op = block_builder
+                            .create<mlir::stablehlo::AddOp>(
+                                loc, tensorType, arguments, attributes)
+                            .getOperation();
   block_builder.create<mlir::func::ReturnOp>(loc, op->getResult(0));
 
-  // verify the module and dump
+  /** verify and dump the module **/
```
mlevesquedion, line 66: Nit: prefer `//` for single-line comments.

examples/c++/ExampleAdd.cpp
```diff
@@ -54,15 +57,27 @@ int main() {
...
   assert(mlir::succeeded(mlir::verify(module.get())));
-  module->dump();
 
-  return 0;
+  /* interpret the  with concrete inputs **/
```
mlevesquedion, line 69: The comment seems to be incomplete: "interpret the ... with concrete inputs" ?

examples/c++/ExampleAdd.cpp
```diff
@@ -54,15 +57,27 @@ int main() {
...
+  auto inputValue1 = mlir::DenseElementsAttr::get(
+      tensorType, block_builder.getFloatAttr(tensorType.getElementType(),
+                                             static_cast<double>(10)));
```
mlevesquedion, line 72: Does `s/static_cast<double>(10)/10.0` work? Or failing that, `double(10)` ?

examples/c++/README.md
```diff
@@ -5,5 +5,41 @@ for how to use StableHLO.
...
+  mlir::stablehlo::InterpreterConfiguration config;
+  auto results = evalModule(module, {inputValue1, inputValue2}, config);
+  llvm::outs() << "Output: " << (*results)[0];
+```
+
+Output:
+
+```mlir
+Program:
+ module @test_module {
```
mlevesquedion, line 36: Nit: there seems to be an extra space before module?

examples/c++/ExampleAdd.cpp
```diff
@@ -54,15 +57,27 @@ int main() {
...
+  auto inputValue2 = mlir::DenseElementsAttr::get(
+      tensorType, block_builder.getFloatAttr(tensorType.getElementType(),
+                                             static_cast<double>(20)));
+  auto expectedValue = mlir::DenseElementsAttr::get(
+      tensorType, block_builder.getFloatAttr(tensorType.getElementType(),
+                                             static_cast<double>(30)));
```
mlevesquedion, line 78: Nit: extract a function to reduce duplication? Alternatively, I think it may be possible to just write `mlir::DenseElementsAttr::get(tensorType, 30.0)`.

mlevesquedion (approved):

examples/c++/ExampleAdd.cpp
sdasgup3: done

examples/c++/ExampleAdd.cpp
sdasgup3: Yup, it should work. Thanks!
updated!

examples/c++/ExampleAdd.cpp
sdasgup3: I have used both  multi-line comments (to express high level actions)  and single line comments (express the steps involved). PTAL if it looks OK.

examples/c++/ExampleAdd.cpp
sdasgup3: Extracted a function. PTAL.

PR #2101, opened 2024-03-15
[RFC] Add TanOp to the StableHLO specification

abhigunj:
This RFC proposed adding TanOp to the StableHLO specification

Please review and provide you feedback.


loganchien (approved):

rfcs/20240312-tanop.md
```diff
@@ -0,0 +1,66 @@
...
+```
+
+## TanOp and Compatibility
+
+`TanOp` can be represented in StableHLO, with some potential
+numerics differences from multiple operations (like `SinOp/CosOp`).
+If there is a use case, forward compatibility is possible via decomposition,
+and the underlying math can be pattern matched into TanOp. We propose to make
+this an opt in pass that doesn't live in the IR upgrade/downgrade pipeline.
+There may be users who want the ops to remain separate.
```
mlevesquedion, line 66: When you write `SinOp/CosOp`, do you mean applying `sin` and `cos` to the argument and dividing the results? If so, consider writing `TanOp(x) = DivOp(SineOp(x), CosineOp(x))` for clarity.

Makes sense to me to have decomposition/recomposition be in a separate pass.

mlevesquedion (approved):

rfcs/20240312-tanop.md
```diff
@@ -0,0 +1,67 @@
...
+### tan
+
+#### Semantics
+
+Performs element-wise tangent operation on `operand` tensor and
+produces a `result` tensor. Depending on the element type, does the following:
+
+* For floats: `tan` from IEEE-754.
+* For complex numbers: complex tangent.
+* For quantized types:
```
ghpvnist, line 32: Nit: this format follows spec for sine and cosine.
```
For quantized types: `dequantize_op_quantize(tan, operand, type(result))`.
```

rfcs/20240312-tanop.md
```diff
@@ -0,0 +1,67 @@
...
+| `result` | tensor of floating-point or complex type or per-tensor quantized tensor | (C1)        |
+
+#### Constraints
+
+* (C1) `baseline_type(operand) = baseline_type(result)`.
+
+#### Examples
+
+```mlir
+// %operand: [-1.0, 0.0, 1.0]
```
ghpvnist, line 54: Nit: using the same inputs for sine and cosine example (I remember writing a bullet point on the [checklist](https://github.com/openxla/stablehlo/blob/main/docs/reference.md#testing-guidelines) to use the highest precision possible for examples as default, but can't find it anymore, but let's still use f64 here):
```
// %operand: [
//            [0.0, 1.57079632],       // [0, pi/2]
//            [3.14159265, 4.71238898] // [pi, 3pi/2]
//           ]
%result = "stablehlo.tan"(%operand) : (tensor<2x2xf64>) -> tensor<2x2xf64>
// %result: [
//           [0.0, 0.0],
//           [0.0, 0.0]
//          ]
```

ghpvnist (approved):
Minor comments, but otherwise LGTM!

sdasgup3 (approved):

GleasonK (changes requested):
Gathering feedback from community members. Adding blocking review for now.

rfcs/20240312-tanop.md
```diff
@@ -0,0 +1,72 @@
...
+#### Examples
+
+```mlir
+// %operand: [
+//            [0.0, 1.57079632],       // [0, pi/2]
+//            [3.14159265, 4.71238898] // [pi, 3pi/2]
+//           ]
+%result = "stablehlo.tan"(%operand) : (tensor<2x2xf64>) -> tensor<2x2xf64>
+// %result: [
+//           [0.0, 0.0],
```
loganchien, line 59: Second element should be +inf

rfcs/20240312-tanop.md
```diff
@@ -0,0 +1,72 @@
...
+//           [0.0, 0.0]
```
loganchien, line 60: Second element should be -inf

rfcs/20240312-tanop.md
sdasgup3: Shouldn't we use 
```
tan(pi/2) =   1.63312e+16
tan(3*pi/2) =  5.44375e+15
```
as they are well within the maximum finite value representable in double-precision ?

loganchien: It looks like I overlooked the fact that `float64(pi/2)` is never the `pi/2` in mathematical sense (because `pi` is irrational).  So, we will not get infinity here.  I think we can just dump what C/C++ libm returns and leave the actual definition to IEEE 754.

sdasgup3: Thanks @loganchien!

@abhigunj Feel  free to use the number mentioned above (generated using c++ math library)

rfcs/20240312-tanop.md
```diff
@@ -51,8 +51,8 @@ produces a `result` tensor. Depending on the element type, does the following:
 
 ```mlir
 // %operand: [
-//            [0.0, 1.57079632],       // [0, pi/2]
-//            [3.14159265, 4.71238898] // [pi, 3pi/2]
```
loganchien: Not these two lines.  These two lines were correct.

It is the `%results` that requires fixes.

abhigunj: yeh.. realized that later, and updated in latest commit., thanks

rfcs/20240312-tanop.md
abhigunj: updated, thank you

loganchien (approved):

rfcs/20240312-tanop.md
```diff
@@ -0,0 +1,72 @@
...
+Last updated: 03/15/2024<br/>
+Discussion thread: [GitHub](https://github.com/openxla/stablehlo/pull/2101)
+
+## Motivation
+
+Several features have been added to MHLO in the past year, which frameworks want
+to leverage and members of the community have made requests for them as well.
+This includes `TanOp`.
+
+Frameworks and Compilers both want `tan` op. Jax has [`jnp.tan`](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.tan.html),
```
cantonios, line 14: nit: "a" `tan` op?

rfcs/20240312-tanop.md
```diff
@@ -0,0 +1,72 @@
...
+PyTorch has [`torch.tan`](https://pytorch.org/docs/stable/generated/torch.tan.html).
+On Compilers side, XLA has [`mhlo.tan`](https://github.com/tensorflow/mlir-hlo/blob/master/mhlo/IR/hlo_ops.td#L633).
```
cantonios, line 16: nit: On "the" Compilers side

rfcs/20240312-tanop.md
```diff
@@ -0,0 +1,72 @@
...
+StableHLO doesn't support `tan` op, but there is an open ticket which requests
```
cantonios, line 17: nit: "the" `tan` op

rfcs/20240312-tanop.md
```diff
@@ -0,0 +1,72 @@
...
+adding this feature in
+[#1358](https://github.com/openxla/stablehlo/issues/1358)
+
+## Proposed Specification
+
+### tan
+
+#### Semantics
+
+Performs element-wise tangent operation on `operand` tensor and
```
cantonios, line 27: nit: on "the" `operand` tensor

rfcs/20240312-tanop.md
```diff
@@ -0,0 +1,72 @@
...
+//          ]
+```
+
+## TanOp and Compatibility
+
+`TanOp` can be represented in StableHLO, with some potential
+numerics differences from multiple operations
+(like `TanOp(x) = DivOp(SineOp(x), CosineOp(x)`).
+If there is a use case, forward compatibility is possible via decomposition,
+and the underlying math can be pattern matched into TanOp. We propose to make
```
cantonios, line 70: nit: missing backticks for `TanOp`.

abhigunj: Thanks for the review @cantonios

GleasonK (approved):

PR #2104, opened 2024-03-17
fix typo of py args

qingyunqu:
because that the lambda's signature is `[](MlirContext context, std::string artifact)`.


sdasgup3 (approved):

GleasonK:
Good catch, thanks for the fix! Will merge once CI completes

PR #2109, opened 2024-03-18
Fix failed reify for ReverseOp

srcarroll:
The `ReverseOp` did not implement a `reifyReturnTypeShapes`.  This resulted in a failed assertion in the `getEmptyTensorFor` helper function that gets used during the `stablehlo-legalize-to-linalg` pass when the op has dynamic input/output.  This patch implements this function as well as adds a test for the newly supported case (input can have dynamic dims as long as they don't correspond to the `dims` attribute of the op)

Various other quality of life changes, such as having match failure notifications have been added.

### DIFF
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -2758,6 +2758,16 @@ def StableHLO_ReverseOp: StableHLO_Op<"reverse",
     $operand `,` `dims` `=` $dimensions
       attr-dict `:` custom<SameOperandsAndResultType>(type($operand), type($result))
   }];
+
+  let extraClassDeclaration = [{
+    LogicalResult reifyReturnTypeShapes(
+        OpBuilder& builder, ValueRange operands,
+        SmallVectorImpl<Value>& reifiedReturnShapes) {
+      return ::mlir::hlo::deriveShapeFromOperand(&builder, getOperation(),
+                                                operands.front(),
+                                                &reifiedReturnShapes);
+    }
+  }];
 }
 
 def StableHLO_PadOp: StableHLO_ShapedInterfaceOp<"pad",
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -2730,7 +2730,7 @@ def StableHLO_SortOp : StableHLO_Op<"sort",
   let hasVerifier = 1;
 }
 
-def StableHLO_ReverseOp: StableHLO_Op<"reverse",
+def StableHLO_ReverseOp: StableHLO_ShapedInterfaceOp<"reverse",
       [Pure, HLO_CompatibleOperandsAndResultType /*reverse_c1*/]> {
   let summary = "Reverse operation";
   let description = [{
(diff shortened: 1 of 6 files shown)


stablehlo/dialect/StablehloOps.td
```diff
@@ -2758,6 +2758,16 @@ def StableHLO_ReverseOp: StableHLO_Op<"reverse",
     $operand `,` `dims` `=` $dimensions
       attr-dict `:` custom<SameOperandsAndResultType>(type($operand), type($result))
   }];
+
+  let extraClassDeclaration = [{
```
GleasonK, line 2762: Looks like all other ops implementing this method declare it via the `InferTensorTypeWithReify` trait. Can look at `ClampOp` for example which seems to use a similar impl, let's base the implementation off that, could also use a non-conversion-based test:


1. Use trait to declare (either [`StableHLO_ShapedInterfaceOp`](https://github.com/openxla/stablehlo/blob/ec4ec7847908b06ed6170c4aba96c164a22f7396/stablehlo/dialect/StablehloOps.td#L1954) or [`InferTensorTypeWithReify`](https://github.com/openxla/stablehlo/blob/ec4ec7847908b06ed6170c4aba96c164a22f7396/stablehlo/dialect/StablehloOps.td#L1538))
2. Implement in [`StablehloOps.cpp`](https://github.com/openxla/stablehlo/blob/ec4ec7847908b06ed6170c4aba96c164a22f7396/stablehlo/dialect/StablehloOps.cpp#L1093)
3. Test in [`infer_stablehlo.mlir`](https://github.com/openxla/stablehlo/blob/ec4ec7847908b06ed6170c4aba96c164a22f7396/stablehlo/tests/infer_stablehlo.mlir#L259)

GleasonK (changes requested):

GleasonK:
cc @sdasgup3 who authored the gather spec to take a look

stablehlo/dialect/StablehloOps.td
srcarroll: sounds good. done. thanks

srcarroll:
i'm not sure what this is about https://github.com/openxla/stablehlo/actions/runs/8346638597/job/22845377545#step:7:4322

stablehlo/dialect/StablehloOps.td
sdasgup3: Thanks!
The option [StableHLO_ShapedInterfaceOp](https://github.com/openxla/stablehlo/blob/ec4ec7847908b06ed6170c4aba96c164a22f7396/stablehlo/dialect/StablehloOps.td#L1954) is recommended in this case.  [InferTensorTypeWithReify](https://github.com/openxla/stablehlo/blob/ec4ec7847908b06ed6170c4aba96c164a22f7396/stablehlo/dialect/StablehloOps.td#L1538) would have added a specialization of `ReverseOp::inferReturnTypes` which is not need in this case.

sdasgup3 (approved):

srcarroll:
Thanks for the review guys. BTW i don't have any permissions in this repo, so someone will need to merge this for me. Appreciate it

GleasonK (approved):

PR #2110, opened 2024-03-19
Fix indentation

sdasgup3:


mlevesquedion (approved):
LGTM. I don't expect this to happen often, but if it does we should consider adding a presubmit check.

sdasgup3 (commented):
looking into putting json file exempted from adding license text.

PR #2111, opened 2024-03-20
Fix bug in Horner's rule

pparuzel:
This is related to openxla/xla#5838 (fixed by cl/615457374). The bug has been producing inaccurate results for Hurwitz Zeta.


GleasonK (approved):
Thanks for the contribution. Looks like the CLA is failing, but also looks like you're from google.. might be some internal configs missing, will reach out separately.

PR #2112, opened 2024-03-20
Add SameOperandsAndResultElementType Trait for type(operand) = type (result) ops 

abhigunj:
~~New trait for StableHLO ops. Trait tries to match `scale` and `zero_point` of per-tensor quantized `operands` and `result`.~~

~~Initially I added checks at custom verifiers for few ops but there are ops without verifier. Adding a Trait is less code/custom verifier changes.~~

This PR adds SameOperandsAndResultElementType  Trait for applicable ops with `type(operand...) = type(result..) ` constraint. 
Total Three OPs  : `collective_broadcast` , `collective_permute`, `reverse`
Testing: positive tests already exist for these ops. Negative tests: since this is a  native mlir Trait and not custom verifier, skipped negative testing. 

There is a overlap of verifications  done by SameOperandsAndResultElementType and HLO_comp.. For now we will keep both the Traits. I will create a follow up tracker to see if we can remove HLO_com.. .  Tracker : #2145

### DIFF
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -2013,7 +2013,8 @@ def StableHLO_ConcatenateOp : StableHLO_ShapedInterfaceOp<"concatenate",
 
 
 def StableHLO_CollectiveBroadcastOp: StableHLO_Op<"collective_broadcast",
-    [HLO_CompatibleOperandsAndResultType /*collective_broadcast_c3*/]> {
+    [HLO_CompatibleOperandsAndResultType,
+     SameOperandsAndResultElementType /*collective_broadcast_c3*/]> {
   let summary = "CollectiveBroadcast operation";
   let description = [{
     Within each process group in the process grid, send the value of the
@@ -2048,7 +2049,8 @@ def StableHLO_CollectiveBroadcastOp: StableHLO_Op<"collective_broadcast",
 }
 
 def StableHLO_CollectivePermuteOp: StableHLO_Op<"collective_permute",
-    [HLO_CompatibleOperandsAndResultType /*collective_permute_c5*/]> {
+    [HLO_CompatibleOperandsAndResultType,
+     SameOperandsAndResultElementType /*collective_permute_c5*/]> {
   let summary = "CollectivePermute operation";
   let description = [{
     Within each process group in the process grid, sends the value of the
@@ -2731,7 +2733,8 @@ def StableHLO_SortOp : StableHLO_Op<"sort",
 }
 
 def StableHLO_ReverseOp: StableHLO_ShapedInterfaceOp<"reverse",
-      [Pure, HLO_CompatibleOperandsAndResultType /*reverse_c1*/]> {
+      [Pure, HLO_CompatibleOperandsAndResultType,
+       SameOperandsAndResultElementType /*reverse_c1*/]> {
   let summary = "Reverse operation";
   let description = [{
     Reverses the order of elements in the `operand` along the specified


stablehlo/dialect/Base.h
```diff
@@ -85,6 +85,14 @@ bool isCompatibleForHloTypeInference(TypeRange tp1, TypeRange tp2);
 // undefined behavior.
 bool isCompatibleForHloTypeInference(Value shape1, Type tp2);
 
+// Returns true if given per-tensor quantized tensors have matching
+// scale and zero_point
+bool isMatchingScaleAndZeroPoint(Type tp1, Type tp2);
+
+// Returns true if all types from a input array are of
```
sdasgup3, line 92: `... if all input types are mlir::quant::UniformQuantizedType`

stablehlo/dialect/Base.cpp
```diff
@@ -33,6 +33,24 @@ limitations under the License.
 namespace mlir {
 namespace hlo {
 
+bool allUniformQuantizedType(TypeRange typeRange) {
```
sdasgup3, line 36: For non-shaped types it would crash.

How about we resuse the functions from [TypeInference.cpp](https://github.com/openxla/stablehlo/blob/c3f456500f0f2e96fdb4a98fde4fbe48b4d624b8/stablehlo/dialect/TypeInference.cpp#L72C1-L84C2) ( with small mofication) and put them in Base.h 

```
template <typename T>
bool allQuantized(ArrayRef<Type> typeRange) {
  return llvm::all_of(typeRange, [&](Type type) {
    return getElementTypeOrSelf(type).isa<T>();
  });
}

template <typename T>
bool noneQuantized(ArrayRef<Type> typeRange) {
  return llvm::all_of(typeRange, [&](Type type) {
    return !getElementTypeOrSelf(type).isa<T>();
  });
}
```

stablehlo/dialect/Base.cpp
```diff
@@ -33,6 +33,24 @@ limitations under the License.
...
+  return llvm::all_of(typeRange, [&](Type val) {
+    return val.cast<ShapedType>()
+        .getElementType()
+        .isa<quant::UniformQuantizedType>();
+  });
+}
+
+bool isMatchingScaleAndZeroPoint(Type tp1, Type tp2) {
+  auto qtp1 = getElementTypeOrSelf(tp1).cast<quant::UniformQuantizedType>();
+  auto qtp2 = getElementTypeOrSelf(tp2).cast<quant::UniformQuantizedType>();
```
sdasgup3, line 46:
```
auto qType1 = getElementTypeOrSelf(tp1).dyn_cast<quant::UniformQuantizedType>();
  auto qType2 = getElementTypeOrSelf(tp2).dyn_cast<quant::UniformQuantizedType>();

if(!qType1 || !qType2) return false;
...
```

stablehlo/dialect/StablehloOps.td
```diff
@@ -194,7 +194,7 @@ class StableHLO_UnaryElementwiseOp<string mnemonic, list<Trait> traits,
 
 // Abs supports complex to real, so element type is not guaranteed to match.
 def StableHLO_AbsOp: StableHLO_UnaryElementwiseOp<"abs",
-    [Pure, DeclareOpInterfaceMethods<InferTypeOpInterface>],
+    [Pure, DeclareOpInterfaceMethods<InferTypeOpInterface>, HLO_MatchingOperandsAndResultScaleZeroPoint],
```
sdasgup3, line 197: why `abs`?
```
(C2) baseline_element_type(result) is defined as:
complex_element_type(element_type(operand)) if is_complex(operand).
baseline_element_type(operand) otherwise.
```

stablehlo/dialect/Base.h
```diff
@@ -85,6 +85,14 @@ bool isCompatibleForHloTypeInference(TypeRange tp1, TypeRange tp2);
...
+// quant::UniformQuantizedType
+bool allUniformQuantizedType(TypeRange typeRange);
```
sdasgup3, line 94: With the reuse mentioned above we do not need this.

stablehlo/dialect/Base.h
```diff
@@ -257,6 +265,37 @@ template <typename ConcreteType>
...
+    : public mlir::OpTrait::TraitBase<ConcreteType,
+                                      MatchingOperandsAndResultScaleZeroPoint> {
+ public:
+  static LogicalResult verifyTrait(Operation *op) {
+    if (failed(mlir::OpTrait::impl::verifyAtLeastNOperands(op, 1)))
+      return failure();
+    if (failed(mlir::OpTrait::impl::verifyAtLeastNResults(op, 1)))
+      return failure();
+
+    if (!allUniformQuantizedType(op->getOperandTypes()) ||
```
sdasgup3, line 279: `allQuantized<UniformQuantizedType>(...)`

stablehlo/dialect/Base.td
```diff
@@ -297,6 +297,9 @@ def HLO_CompatibleOperandsAndResultType : TraitList<
 def HLO_CompatibleOperandsAndResultElementType :
   HLO_NativeOpTrait<"CompatibleOperandsAndResultElementType">;
 
+def HLO_MatchingOperandsAndResultScaleZeroPoint :
```
sdasgup3, line 300: Add a comment to the functionality of the trait.

stablehlo/dialect/StablehloOps.td
```diff
@@ -1282,7 +1282,8 @@ def StableHLO_WhileOp: StableHLO_Op<"while", [
 }
 
 def StableHLO_AllGatherOp : StableHLO_Op<"all_gather",
-    [SameOperandsAndResultElementType] /*all_gather_c6*/> {
+    [SameOperandsAndResultElementType,
+     HLO_MatchingOperandsAndResultScaleZeroPoint] /*all_gather_c6*/> {
```
sdasgup3, line 1286: We do not need this here? `SameOperandsAndResultElementType` should take care of equal scales and zero-points. Please confirm before removing.

stablehlo/tests/ops_stablehlo_quantized.mlir
```diff
@@ -893,3 +893,21 @@ func.func @convolution_c33(%arg0: tensor<1x8x8x207x!quant.uniform<i8:f32, 2.0:15
     (tensor<1x8x8x207x!quant.uniform<i8:f32, 2.0:15>>, tensor<3x3x207x16x!quant.uniform<i8:f32:3, {0.1:-30}>>) -> tensor<1x8x8x16x!quant.uniform<i8:f32:0, {2.0:-30}>>
  func.return %0 : tensor<1x8x8x16x!quant.uniform<i8:f32:0, {2.0:-30}>>
 }
+
+// -----
+
+func.func @match_scale_trait_negative_test(%arg0: tensor<1x2x2x!quant.uniform<i8:f32, 0.1:-30>>){
+  // expected-error@+1 {{requires matching scale and zero_point for all operands and result}}
+  %abs_neg = "stablehlo.abs"(%arg0) : (tensor<1x2x2x!quant.uniform<i8<-128:127>:f32, 0.1:-30>>) -> tensor<1x2x2x!quant.uniform<i8<-128:127>:f32, 0.2:-30>>
```
sdasgup3, line 901: remove the `abs` example as same zp and scale is not applicable. Replace it with `collective_permute`

stablehlo/dialect/Base.cpp
```diff
@@ -33,6 +33,24 @@ limitations under the License.
...
+  if ((qtp1.getScale() != qtp2.getScale()) ||
+      (qtp1.getZeroPoint() != qtp2.getZeroPoint()))
+    return false;
+
+  return true;
```
mlevesquedion, line 51: Consider returning the expression directly: `return qtp1.getScale() == qtp2.getScale() && qtp1.getZeroPoint() == qtp2.getZeroPoint();`.

sdasgup3 (commented):
As we are planning to reuse this trait for https://github.com/openxla/stablehlo/pull/2081,
Please make sure that the traits works for both per-tensor and per-axis types. 
For example, we need that if we want to reuse it for `transpose` op.

stablehlo/dialect/StablehloOps.td
```diff
@@ -2731,7 +2734,8 @@ def StableHLO_SortOp : StableHLO_Op<"sort",
 }
 
 def StableHLO_ReverseOp: StableHLO_ShapedInterfaceOp<"reverse",
-      [Pure, HLO_CompatibleOperandsAndResultType /*reverse_c1*/]> {
+      [Pure, HLO_CompatibleOperandsAndResultType,
+       HLO_MatchingOperandsAndResultScaleZeroPoint /*reverse_c1*/]> {
```
GleasonK, line 2738: Can we just use the existing `SameOperandsAndResultElementType`?

sdasgup3 (commented):
In this PR try to handle all the ops which needs this trait.

```
transpose
while
all_to_all
dynamic_update_slice
optimization_barrier
sort
```

stablehlo/dialect/StablehloOps.td
GleasonK: resolving moving to other thread.

stablehlo/dialect/StablehloOps.td
GleasonK: Can we just use the existing `SameOperandsAndResultElementType` everywhere else? Does this trait only expects one result or something?

sdasgup3: @GleasonK That is a good point!

It seems that in most of the case we do not need this new trait and can use  the existing one `SameOperandsAndResultElementType`. This is true for all the operations with constraints similar to `type(operand...) = type(result..)`. 

@abhigunj 

How about we:
1. Try reusing the existing traits instead of the new one. Apply the change on all operations which need same operands and result quantized type.
3. If no surprises on (1), update the issue label and description.

abhigunj: Thank you all for the suggestions. 
took a relook at ops with type(operand...) = type(result..) constraints which need verifier updates: 

* `SameOperandsAndResultElementType` Trait is used for the three OPs (`collective_broadcast`, `collective_permute`, `reverse`). Validated.  Will use this PR to update it. 

* `reshape` `transpose` `broadcast_in_dim`   we can't use  `SameOperandsAndResultElementType`  due to slight variation from the `type(operand..)` constraints. I have updated their custom verifier at https://github.com/openxla/stablehlo/pull/2081 . Plan was to replace those with the new Trait. We will keep updates to custom verifier unchanged  and won't add new Trait from this PR.

abhigunj:
> As we are planning to reuse this trait for #2081, Please make sure that the traits works for both per-tensor and per-axis types. For example, we need that if we want to reuse it for `transpose` op.

closing this as no-op since we are not using Trait and keeping custom verifier changes as it is. ref https://github.com/openxla/stablehlo/pull/2112#discussion_r1532634910

GleasonK (approved):

sdasgup3:
lgtm. Please add the follow up tracker in the description.

abhigunj:
> As we are planning to reuse this trait for #2081, Please make sure that the traits works for both per-tensor and per-axis types. For example, we need that if we want to reuse it for `transpose` op.

decided to drop the changes for new Trait. Not applicable.

PR #2116, opened 2024-03-21
Remove shape operand change from Dynamism RFC

mlevesquedion:
On the whole, it seems like this change has low returns and high costs.

See https://groups.google.com/a/openxla.org/g/openxla-discuss/c/HJRvFBum65k/m/Xzld0wJaAAAJ for discussion.


GleasonK (approved):
LGTM. Thanks for driving discussion on OpenXLA discuss as well. Symbolic shapes may become a higher priority feature going forward, but I agree its best to hold off until we have a driving use case (framework or compiler) to help guide our design.

PR #2097, opened 2024-03-15
[RFC] Standardize CustomCallOp to extend backend_config to take a DictionaryAttr

abhigunj:
This RFC proposes standardizing CustomCallOp to extend backend_config to take a DictionaryAttr.

Please review and provide you feedback.

PR #2117, opened 2024-03-21
Fix formatting on spec examples

ghpvnist:
As pointed out in https://github.com/openxla/stablehlo/pull/2097#discussion_r1534620186, MLIR doesn't insert spaces between the brackets and the content. I've also made minor formatting improvements.


mlevesquedion (approved):
Nice, thanks! Similar changes may be needed in ODS?

ghpvnist:
I did a pass through `StablehloOps.td` beforehand and only found the ones proposed here. Lmk if you find additional ones.

mlevesquedion:
LGTM then, thanks for taking the time!

PR #2118, opened 2024-03-21
Introduce a single CMakeLists.txt

fzakaria:
Build MLIR & StableHLO via a single cmake configure & build command. This will make the build_mlir.sh script unecessary.

This should simplify setting cmake values to be consistent across MLIR and StableHLO builds such as PYTHON_BINDINGS.


.github/workflows/buildAndTestCMake.yml
```diff
@@ -84,16 +71,11 @@ jobs:
     - name: Build and Test StableHLO (with AddressSanitizer)
       shell: bash
       run: |
-        ./build_tools/github_actions/ci_build_cmake.sh "$LLVM_BUILD_DIR" "$STABLEHLO_BUILD_DIR"
-      env:
-          CMAKE_BUILD_TYPE: Release
-          STABLEHLO_ENABLE_BINDINGS_PYTHON: OFF
-          STABLEHLO_ENABLE_SANITIZER: address
+        cmake --preset debug
+        cmake --build ./build --target check-stablehlo-ci
```
fzakaria, line 75: CMake has a concept of a "workflow" that can do this but there is some weird history to CMakePresets [1]

[1] https://gitlab.kitware.com/cmake/cmake/-/issues/22538

fzakaria:
There is also CMakeFetchContent which may also even simplify this more.
https://cmake.org/cmake/help/latest/module/FetchContent.html

CMakeLists.txt
```diff
@@ -89,13 +89,69 @@ if(STABLEHLO_EXTERNAL_PROJECT_BUILD)
...
+  set(LLVM_ENABLE_PROJECTS "" CACHE STRING "")
+  set(LLVM_EXTERNAL_PROJECTS "" CACHE STRING "")
+
+  # Unconditionally enable mlir.
+  list(APPEND LLVM_ENABLE_PROJECTS mlir)
+
+  # Setup LLVM lib and bin directories.
+  set(LLVM_RUNTIME_OUTPUT_INTDIR ${CMAKE_BINARY_DIR}/llvm-project/bin)
+  set(LLVM_LIBRARY_OUTPUT_INTDIR ${CMAKE_BINARY_DIR}/llvm-project/lib)
+  set(LLVM_TOOLS_BINARY_DIR ${CMAKE_BINARY_DIR}/llvm-project/bin)
```
fzakaria, line 118: having to set these here are pretty ugly; we can discuss if this is an adequate trade-off.

README.md
```diff
@@ -57,81 +50,34 @@ Here's how to build the StableHLO repo on Linux or macOS:
...
 
    You should see results like this:
 
    ```txt
-   Testing Time: 5.99s
-     Passed: 47
+   Testing Time: 4.13s
+
+   Total Discovered Tests: 137
+   Passed: 137 (100.00%)
```
mlevesquedion, line 76: Nit: This would be accurate if we were running `check-stablehlo-tests`, but because we are running `check-stablehlo-ci` 4 test suites need to run.

IMO, `check-stablehlo-tests` is more appropriate because the other 3 suites (tosa, linalg, testdata) are not as relevant to day-to-day development.

Also I commented about this elsewhere, and IDK if it's because we're using a `debug` build or something else, but running these tests is usually way faster for me. On `main`, I can run all 3888 tests in under 10 seconds, but with this branch it takes over a minute.

mlevesquedion (commented):
A few thoughts:
- First of all, this is  a massive quality of life improvement and makes the project way more approachable. Even better would be if the developer didn't have to manually clone LLVM and it was just managed by the build system (I believe this is the case with the Bazel build). Regardless, this is huge.
- I don't really know CMake so take these comments with a grain of salt: 1. Some of the config seems kinda fragile, like it may be relying on implementation details of LLVM's CMake build; this could lead to issues down the road; 2. Because I don't know CMake, I'm worried about maintainability. I feel like this is making the config more complicated than before and it feels like it would be harder to change/debug. I'm also wondering if this will have an impact on building StableHLO embedded in another project. (I don't think so, but I'm not 100% confident.)

Overall, I'm in favor of this change. I think the complexity it adds for project maintainers is worth the gains for everyone.

Another thought, vaguely related to this change: I don't love that we have two build systems. I wish we could generate CMakeLists.txt from Bazel or something. There is some prior art here but no established solution unfortunately: https://groups.google.com/g/bazel-discuss/c/MlI-xk034lw, https://github.com/google/bazel-to-cmake

README.md
fzakaria: Let's add a release preset and test against that?
I think starting the debug binaries are expensive.

PR #2119, opened 2024-03-21
Clean up hasRank checks

mlevesquedion:
These checks are no longer necessary in most cases (they are still necessary in some cases, e.g. when dealing with custom call args/results or CHLO op args/results).

I plan to make further cleanups in future PRs, e.g. in ShapeRefinement.cpp (in this PR I did only the obvious ones). Also I plan to clean up casts, especially casting to `ShapedType` instead of `RankedTensorType`.

#1991

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -4214,7 +4214,7 @@ LogicalResult verifySelectAndScatterOp(
         location, "expects select-region to return single value, but got: ",
         selectResult.size());
 
-  auto selectResultType = selectResult[0].getType().cast<RankedTensorType>();
+  auto selectResultType = selectResult[0].getType().dyn_cast<RankedTensorType>();
   // select_and_scatter_c9
   if (!selectResultType || !selectResultType.getElementType().isInteger(1) ||
       selectResultType.getRank() != 0)
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -727,7 +727,6 @@ LogicalResult verifyReduceWindowOpInputsAndInferWindow(
   if (inputTypes.empty())
     return emitOptionalError(location, "requires at least 1 input value");
 
-  // Check for unranked tensors in input operands.
   auto witnessType = inputTypes[0].cast<RankedTensorType>();
   // reduce_window_c2
   for (size_t i = 1; i < inputTypes.size(); i++)
@@ -2390,7 +2389,6 @@ LogicalResult inferMapOp(
 
   // map_c3
   ArrayRef<int64_t> resultShape;
-  bool allInputsUnranked = true;
   for (auto operand : inputs) {
     auto operandType = operand.getType().cast<ShapedType>();
     if (dimensions.size() != operandType.getShape().size())
@@ -2401,15 +2399,11 @@ LogicalResult inferMapOp(
           operandType.getShape().size(),
           ", requested map dimensions size = ", dimensions.size());
     resultShape = operandType.getShape();
-    allInputsUnranked = false;
   }
 
   // map_c4
-  if (allInputsUnranked)
-    inferredReturnShapes.emplace_back(computationOutputType.getElementType());
-  else
-    inferredReturnShapes.emplace_back(resultShape,
-                                      computationOutputType.getElementType());
+  inferredReturnShapes.emplace_back(resultShape,
+                                    computationOutputType.getElementType());
   return success();
 }
 
@@ -2981,7 +2975,7 @@ LogicalResult inferUniformDequantizeOp(
   auto operandType = operand.getType().cast<ShapedType>();
   // Trait HLO_QuantizedIntTensor in ODS guarantees QuantizedType;
   auto quantType = operandType.getElementType().cast<quant::QuantizedType>();
-  auto shape = operandType.cast<ShapedType>().getShape();
+  auto shape = operandType.getShape();
   // uniform_dequantize_c1, uniform_dequantize_c2
   inferredReturnShapes.emplace_back(shape, quantType.getExpressedType());
   return success();
@@ -4220,10 +4214,10 @@ LogicalResult verifySelectAndScatterOp(
         location, "expects select-region to return single value, but got: ",
         selectResult.size());
 
-  auto selectResultType = selectResult[0].getType().dyn_cast<ShapedType>();
+  auto selectResultType = selectResult[0].getType().cast<RankedTensorType>();
   // select_and_scatter_c9
   if (!selectResultType || !selectResultType.getElementType().isInteger(1) ||
-      selectResultType.cast<RankedTensorType>().getRank() != 0)
+      selectResultType.getRank() != 0)
     return emitOptionalError(location,
                              "expects the return-type of select-region to be "
                              "tensor<i1>, but got: ",
@@ -4285,9 +4279,8 @@ LogicalResult verifySortOp(std::optional<Location> location, ValueRange inputs,
       return emitOptionalError(location,
                                "dimension attribute value must be in range [-",
                                rank, ", ", rank, "), but found ", cmpDim);
-    else
-      break;  // ODS SameOperandsAndResultShape asserts inputs have same
-              // shape
+    // ODS SameOperandsAndResultShape asserts inputs have same shape
+    break;
   }
 
   Block& block = comparator.front();
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -529,39 +529,21 @@ LogicalResult verifyReduceOpInputsAndInferShape(
     std::optional<Location> location, SmallVector<ShapedType> inputTypes,
     ArrayRef<int64_t> dimensions, SmallVector<int64_t>& newDimensions,
     Attribute& encoding) {
-  // Check for unranked tensors in input operands.
-  uint64_t numInputs = inputTypes.size();
-  int64_t rankedInputIdx = -1;
-  for (uint64_t inputIdx = 0; inputIdx < numInputs; ++inputIdx) {
-    if (inputTypes[inputIdx].hasRank()) {
-      rankedInputIdx = inputIdx;
-      break;
-    }
-  }
-  bool allInputsUnranked = (rankedInputIdx == -1);
   // reduce_c1
-  if (!allInputsUnranked) {
-    for (uint64_t inputIdx = 0; inputIdx < numInputs; ++inputIdx)
-      if (failed(mlir::verifyCompatibleShape(inputTypes[rankedInputIdx],
-                                             inputTypes[inputIdx])))
-        return emitOptionalError(
-            location, "expects all inputs to have compatible shapes. Shape at",
-            " input-index ", inputIdx,
-            " is not compatible with shape at input-index ", rankedInputIdx);
-  }
+  auto witnessType = inputTypes[0].cast<RankedTensorType>();
+  for (size_t i = 1; i < inputTypes.size(); i++)
+    if (failed(mlir::verifyCompatibleShape(witnessType, inputTypes[i])))
+      return emitOptionalError(
+          location, "expects all inputs to have compatible shapes. Shape at",
+          " input-index ", i, " is not compatible with shape at input-index 0");
 
   DenseSet<int64_t> dimensionsToReduceSet;
   for (int64_t dimension : dimensions) {
     // reduce_c4
-    if ((!allInputsUnranked &&
-         dimension >= inputTypes[rankedInputIdx].getRank()) ||
-        dimension < 0)
-      return emitOptionalError(
-          location, "Out-of-bounds dimension ", dimension, ", expected to be ",
-          allInputsUnranked
-              ? "> 0"
-              : "less than the input-tensor rank " +
-                    std::to_string(inputTypes[rankedInputIdx].getRank()));
+    if (dimension < 0 || dimension >= witnessType.getRank())
+      return emitOptionalError(location, "Out-of-bounds dimension ", dimension,
+                               ", expected to be in range [0, ",
+                               witnessType.getRank(), ')');
 
     // reduce_c5
     if (!dimensionsToReduceSet.insert(dimension).second)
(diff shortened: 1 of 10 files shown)


stablehlo/dialect/TypeInference.cpp
```diff
@@ -3175,10 +3094,10 @@ LogicalResult verifyAllReduceOp(std::optional<Location> location, Value operand,
...
-        location,
-        "channel_id must be positive when useGlobalDeviceIds is set but got: ",
-        channelId);
+    return emitOptionalError(location,
+                             "channel_id must be positive when "
+                             "useGlobalDeviceIds is set but got: ",
+                             channelId);
 
   auto operandType = operand.getType().cast<ShapedType>();
   // all_reduce_c5
```
ghpvnist, line 3103: Also in `verifyBitcastConvertOp`, let's simplify redundant cast:
```
  auto operandShapedType = operand.getType().cast<ShapedType>();
  auto targetShapedType = result.getType().cast<ShapedType>();
  ...
  auto operandType = operandShapedType.dyn_cast<RankedTensorType>();
  auto targetType = targetShapedType.dyn_cast<RankedTensorType>();
```

to:
```
auto operandType = operand.getType().cast<RankedTensorType>();
```

stablehlo/dialect/TypeInference.cpp
```diff
@@ -4309,12 +4223,11 @@ LogicalResult verifySelectAndScatterOp(
   auto selectResultType = selectResult[0].getType().dyn_cast<ShapedType>();
   // select_and_scatter_c9
   if (!selectResultType || !selectResultType.getElementType().isInteger(1) ||
-      (selectResultType.hasRank() &&
-       selectResultType.cast<RankedTensorType>().getRank() != 0))
-    return emitOptionalError(
-        location,
-        "expects the return-type of select-region to be tensor<i1>, but got: ",
-        selectResult[0].getType());
+      selectResultType.cast<RankedTensorType>().getRank() != 0)
```
ghpvnist, line 4226: Do we need the cast here?

stablehlo/dialect/TypeInference.cpp
```diff
@@ -4370,17 +4278,16 @@ LogicalResult verifySortOp(std::optional<Location> location, ValueRange inputs,
...
-    }
+    int64_t cmpDim = dimension;
+    int64_t rank = operandShapedType.getRank();
+    // sort_c4
+    if (cmpDim < -rank || cmpDim >= rank)
+      return emitOptionalError(location,
+                               "dimension attribute value must be in range [-",
+                               rank, ", ", rank, "), but found ", cmpDim);
+    else
+      break;  // ODS SameOperandsAndResultShape asserts inputs have same
```
ghpvnist, line 4289: Nit: Not directly related to your changes, but let's remove the `else` keyword here and move the comment above the `break`

ghpvnist (changes requested):
Thanks for the clean up. This was a lot to review (probably much longer to write!), so I may do another pass. Overall two comments:
1. We have a mix of `cast` and `dyn_cast` that would be good to clean up.
2. We also have a mix of `ShapedType` and `RankedTensorType`. I think `ShapedType` was used previously because we had both `UnrankedTensorType` and `RankedTensorType` and `ShapedType` was the convenient one to cast to to do type checks. However, since we explicitly disallow `UnrankedTensorType` in our dialect, wdyt about converting all `ShapedType` casts into `RankedTensorType`? Overall, I think casting to `RankedTensorType` is more readable, and there is not need to guess wonder about the differences between the two for newcomers. This is also a nontrivial change, so I think it's appropriate to address this in a separate PR.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -529,62 +529,41 @@ LogicalResult verifyReduceOpInputsAndInferShape(
...
-      if (failed(mlir::verifyCompatibleShape(inputTypes[rankedInputIdx],
-                                             inputTypes[inputIdx])))
-        return emitOptionalError(
-            location, "expects all inputs to have compatible shapes. Shape at",
-            " input-index ", inputIdx,
-            " is not compatible with shape at input-index ", rankedInputIdx);
-  }
+  auto witnessType = inputTypes[0].cast<RankedTensorType>();
+  for (size_t i = 1; i < inputTypes.size(); i++)
+    if (failed(mlir::verifyCompatibleShape(witnessType, inputTypes[i])))
```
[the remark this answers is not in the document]
mlevesquedion: I think I could remove it, but since the diff is already large and this is not directly related to the change I'm making I'd prefer to leave this for a future cleanup.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -753,34 +728,20 @@ LogicalResult verifyReduceWindowOpInputsAndInferWindow(
     return emitOptionalError(location, "requires at least 1 input value");
 
   // Check for unranked tensors in input operands.
```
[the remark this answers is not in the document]
mlevesquedion: Removed.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -2467,17 +2393,15 @@ LogicalResult inferMapOp(
   bool allInputsUnranked = true;
```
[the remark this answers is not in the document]
mlevesquedion: Done.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3072,8 +2992,7 @@ LogicalResult inferUniformQuantizeOp(
     SmallVectorImpl<ShapedTypeComponents>& inferredReturnShapes) {
   auto operandType = operand.getType().cast<ShapedType>();
```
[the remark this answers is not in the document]
mlevesquedion: I think this is a good idea and I do plan to send a PR for that, but since this PR is already large I refrained from doing it. (Actually you may have noticed I did a few in the linalg legalization code, but that was because I needed to change the code anyway.)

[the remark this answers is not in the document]
mlevesquedion: I would say that's out of scope for this PR, but since it's a small change I removed the extraneous cast. Thanks!

stablehlo/dialect/TypeInference.cpp
mlevesquedion: Thanks for pointing that out, I'll do that in a follow-up PR.

stablehlo/dialect/TypeInference.cpp
mlevesquedion: Done. Also I cleaned up the declaration of `selectResultType` since it's right there.

stablehlo/dialect/TypeInference.cpp
mlevesquedion: Done.

mlevesquedion (commented):
Thank you for the meticulous review! I do plan to clean up casts and change ShapedType to RankedTensorType where possible, but as you suggest I will do these separately since this change is already quite large.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -1676,54 +1625,39 @@ LogicalResult inferConcatenateOp(std::optional<Location> location,
...
+    for (int d = 0; d < rank; ++d) {
+      if (d != dimension && !verifyCompatibleDims(witnessShape[d], shape[d]))
         return emitOptionalError(
-            location, "shapes of operand (", firstRankedIndex, ") and (", i,
-            ") do not match at non-concat "
-            "index: (",
-            llvm::make_range(firstShape.begin(), firstShape.end()), ") != (",
-            llvm::make_range(secondShape.begin(), secondShape.end()),
-            ") at non-concat index ", d);
+            location, "shapes of operand (", 0, ") and (", i,
```
ghpvnist, line 1651:
```
        return emitOptionalError(
            location, "shapes of operand (", 0, ") and (", i,
            ") are not compatible at non-concat index ", d, ": (",
            llvm::make_range(witnessShape.begin(), witnessShape.end()),
            ") != (", llvm::make_range(shape.begin(), shape.end()), ")");
```

stablehlo/dialect/TypeInference.cpp
```diff
@@ -2464,28 +2389,21 @@ LogicalResult inferMapOp(
...
-            location,
-            "applied to a subset of dimensions currently not supported: "
-            "operand dimensions = ",
-            operandType.getShape().size(),
-            ", requested map dimensions size = ", dimensions.size());
-      resultShape = operandType.getShape();
-      allInputsUnranked = false;
-    }
+    if (dimensions.size() != operandType.getShape().size())
+      return emitOptionalError(
```
ghpvnist, line 2395:
```
      return emitOptionalError(
          location,
          "applied to a subset of dimensions currently not supported: operand "
          "dimensions = ",
          operandType.getShape().size(),
          ", requested map dimensions size = ", dimensions.size());
```

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3175,10 +3088,10 @@ LogicalResult verifyAllReduceOp(std::optional<Location> location, Value operand,
 
   // all_reduce_c4
   if (useGlobalDeviceIds && channelId <= 0)
-    return emitOptionalError(
-        location,
-        "channel_id must be positive when useGlobalDeviceIds is set but got: ",
-        channelId);
+    return emitOptionalError(location,
```
ghpvnist, line 3091: This change can be reverted.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3751,11 +3661,11 @@ LogicalResult verifyInfeedOp(HloDialectInterface* dialect,
   // infeed_c2
   for (auto resultType : results.drop_back().getTypes())
     if (!resultType.isa<TensorType>())
-      return emitOptionalError(
-          location,
-          "all elements of result types, except the last element, are expected "
-          "to be of tensor type, but got ",
-          resultType);
+      return emitOptionalError(location,
```
ghpvnist, line 3664: This change can be reverted.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3952,10 +3860,10 @@ LogicalResult verifyReduceScatterOp(std::optional<Location> location,
 
   // reduce_scatter_c6
   if (useGlobalDeviceIds && channelId <= 0)
-    return emitOptionalError(
-        location,
-        "channel_id must be positive when useGlobalDeviceIds is set but got: ",
-        channelId);
+    return emitOptionalError(location,
```
ghpvnist, line 3863: This change can be reverted.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -4085,10 +3993,10 @@ LogicalResult verifyRngBitGeneratorOp(std::optional<Location> location,
   auto outputShape = outputState.getType().dyn_cast<RankedTensorType>();
   if (failed(verifyCompatibleShape(initialShape.getShape(),
                                    outputShape.getShape())))
-    return emitOptionalError(
-        location,
-        "output state shape must be compatible with initial state shape. Got: ",
-        initialShape, " and ", outputShape);
+    return emitOptionalError(location,
```
ghpvnist, line 3996: This change can be reverted.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -4306,15 +4214,15 @@ LogicalResult verifySelectAndScatterOp(
...
   // select_and_scatter_c9
   if (!selectResultType || !selectResultType.getElementType().isInteger(1) ||
-      (selectResultType.hasRank() &&
-       selectResultType.cast<RankedTensorType>().getRank() != 0))
-    return emitOptionalError(
-        location,
-        "expects the return-type of select-region to be tensor<i1>, but got: ",
-        selectResult[0].getType());
+      selectResultType.getRank() != 0)
+    return emitOptionalError(location,
```
ghpvnist, line 4222:
```
    return emitOptionalError(
        location,
        "expects the return-type of select-region to be tensor<i1>, but got: ",
        selectResult[0].getType());
```

stablehlo/dialect/TypeInference.cpp
```diff
@@ -4325,16 +4233,14 @@ LogicalResult verifySelectAndScatterOp(
...
-      return emitOptionalError(
-          location,
-          "expects window-dimensions size == operand rank, but got "
-          "window-dimensions size: ",
-          windowDims.size(), " and operand-type: ", operandType,
-          " with rank = ", operandType.getRank(), ".");
-  }
+  // select_and_scatter_c4
+  if (operandType.getRank() != static_cast<int64_t>(windowDims.size()))
+    return emitOptionalError(
```
ghpvnist, line 4238:
```
                             "expects window-dimensions size == operand rank, "
                             "but got window-dimensions size: ",
                             windowDims.size(),
                             " and operand-type: ", operandType,
                             " with rank = ", operandType.getRank(), ".");
```

stablehlo/dialect/TypeInference.cpp
```diff
@@ -752,35 +727,20 @@ LogicalResult verifyReduceWindowOpInputsAndInferWindow(
...
-      if (failed(mlir::verifyCompatibleShape(inputTypes[rankedInputIdx],
-                                             inputTypes[inputIdx])))
-        return emitOptionalError(
-            location, "expects all inputs to have compatible shapes. Shape at",
-            " input-index ", inputIdx,
-            " is not compatible with shape at input-index ", rankedInputIdx);
-  }
+  for (size_t i = 1; i < inputTypes.size(); i++)
+    if (failed(mlir::verifyCompatibleShape(witnessType, inputTypes[i])))
+      return emitOptionalError(
```
ghpvnist, line 734:
```
      return emitOptionalError(
          location,
          "expects all inputs to have compatible shapes. Shape at input-index ",
          i, " is not compatible with shape at input-index 0");
```

stablehlo/dialect/TypeInference.cpp
```diff
@@ -529,62 +529,41 @@ LogicalResult verifyReduceOpInputsAndInferShape(
...
-                                             inputTypes[inputIdx])))
-        return emitOptionalError(
-            location, "expects all inputs to have compatible shapes. Shape at",
-            " input-index ", inputIdx,
-            " is not compatible with shape at input-index ", rankedInputIdx);
-  }
+  auto witnessType = inputTypes[0].cast<RankedTensorType>();
+  for (size_t i = 1; i < inputTypes.size(); i++)
+    if (failed(mlir::verifyCompatibleShape(witnessType, inputTypes[i])))
+      return emitOptionalError(
```
ghpvnist, line 536:
```
      return emitOptionalError(
          location,
          "expects all inputs to have compatible shapes. Shape at input-index ",
          inputIdx, " is not compatible with shape at input-index ",
          rankedInputIdx);
```

ghpvnist (approved):
Ah I didn't notice you already pointed that out in your PR description. Sounds good to me. Mostly LGTM with some nit comments regarding formatting (sometimes format might split the string in another way while making changes to the indent layer).

mlevesquedion:
> Ah I didn't notice you already pointed that out in your PR description. Sounds good to me. Mostly LGTM with some nit comments regarding formatting (sometimes format might split the string in another way while making changes to the indent layer).

That may be because I edited the PR description since I first uploaded the change :sweat_smile:. I was unsure whether to include this at first, I guess I should have to be clearer about my intentions. Sorry about that.

Regarding the formatting changes, I ran `clang-format` on the files and this is what I got. I tried to apply a few of the changes you suggested manually, but `clang-format` changed them back.

mlevesquedion:
> > Ah I didn't notice you already pointed that out in your PR description. Sounds good to me. Mostly LGTM with some nit comments regarding formatting (sometimes format might split the string in another way while making changes to the indent layer).
> 
> That may be because I edited the PR description since I first uploaded the change 😅. I was unsure whether to include this at first, I guess I should have been clearer about my intentions. Sorry about that.
> 
> Regarding the formatting changes, I ran `clang-format` on the files and this is what I got. I tried to apply a few of the changes you suggested manually, but `clang-format` changed them back.

@ghpvnist since we discussed this offline and you provided the reformattings, I applied the changes. However, moving forward I think we should not take extra time to fix things manually after `clang-format` has been applied. Using an automated formatting tool can save us a lot of time and energy, and manually fixing things goes against that.

ghpvnist:
Here's my rationale for fixing 80-char on the first line if possible. While developing, you may attempt to format code when code still has syntax errors and may indent all code below that to something funky. Then after fixing it, you may end up having code like this which still passes `clang-format`. An extreme example:
```
                                 "expects "
                                 "all "
                                 "inputs "
                                 "to "
                                 "have "
                                 "compatible "
                                 "shapes. "
                                 "Shape "
                                 "at",
                                 " input-index ", inputIdx,
                                 " is "
                                 "not "
                                 "compatible "
                                 "with "
                                 "shape "
                                 "at "
                                 "input-index ",
                                 rankedInputIdx);
```

However, if we make sure to unify all strings at the 80-char limit, we'd get the optimal format:
```
        return emitOptionalError(
            location, "expects all inputs to have compatible shapes. Shape at",
            " input-index ", inputIdx,
```

But yes I agree, for future format changes, I'll consider using the suggest via PR feature so the author doesn't have to manually fix it themselves.

PR #2126, opened 2024-03-26
Add terminating newline

ghpvnist:


sdasgup3 (approved):

mlevesquedion:
Do we know how this made it past CI? I believe this should fail the whitespace check.

mlevesquedion:
> Do we know how this made it past CI? I believe this should fail the whitespace check.

@ghpvnist @sdasgup3 Any thoughts on this? Thanks for adding the missing newline! I'm just wondering if we need to update the whitespace check to avoid future issues like this one.

ghpvnist:
I'm not sure either. Created #2129 to track this.

PR #2127, opened 2024-03-26
AddOp: Remove HLO_CompatibleOperandsAndResultType Trait, add custom verifier

abhigunj:
AddOP allows mix of `per-tensor` and `per-axis` inputs. Existing StableHLO AddOp def uses `HLO_CompatibleOperandsAndResultType` Trait which does not allow mix of `per-tensor` and `per-axis` inputs. 
This PR:
removed HLO_CompatibleOperandsAndResultType Trait for AddOP
Added custom Verifier to implement OP constraints

Thank you @sdasgup3  for the help with debugging builder issues after removal of `HLO_CompatibleOperandsAndResultType` suggestion for  `inferReturnTypes` placement.

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -81,6 +81,83 @@ bool noneQuantized(ArrayRef<Type> typeRange) {
       typeRange, [&](Type val) { return !getElementTypeOrSelf(val).isa<T>(); });
 }
 
+template <typename T>
+bool anyQuantized(ArrayRef<Type> typeRange) {
+  return llvm::any_of(
+      typeRange, [&](Type val) { return getElementTypeOrSelf(val).isa<T>(); });
+}
+
+LogicalResult verifyBinaryOpQuantizationConstraints(
+    std::optional<Location> location, Type lhsType, Type rhsType,
+    Type resultType) {
+  lhsType = getElementTypeOrSelf(lhsType);
+  rhsType = getElementTypeOrSelf(rhsType);
+  resultType = getElementTypeOrSelf(resultType);
+  llvm::SmallVector<Type, 3> typeEntries{lhsType, rhsType, resultType};
+  // add_c1
+  if (noneQuantized<quant::QuantizedType>(typeEntries)) {
+    if (lhsType != rhsType || lhsType != resultType)
+      return emitOptionalError(
+          location, "op requires same types for all operands and results");
+    return success();
+  }
+  // add_c2
+  if (!allQuantized<quant::QuantizedType>(typeEntries)) {
+    return emitOptionalError(location,
+                             "expects  all operands and results to be either "
+                             "quantized or non-quantized");
+  }
+  auto lhsQType = lhsType.dyn_cast<quant::QuantizedType>();
+  auto rhsQType = rhsType.dyn_cast<quant::QuantizedType>();
+  auto resultQType = resultType.dyn_cast<quant::QuantizedType>();
+  // add_c3
+  auto storageType = lhsQType.getStorageType();
+  if (storageType != rhsQType.getStorageType() ||
+      storageType != resultQType.getStorageType())
+    return emitOptionalError(
+        location, "mismatched operands and result quantization storage types");
+  // add_c4
+  auto expressedType = lhsQType.getExpressedType();
+  if (expressedType != rhsQType.getExpressedType() ||
+      expressedType != resultQType.getExpressedType())
+    return emitOptionalError(
+        location,
+        "mismatched operands and result quantization expressed types");
+
+  auto lhsQPAType = lhsType.dyn_cast<quant::UniformQuantizedPerAxisType>();
+  auto rhsQPAType = rhsType.dyn_cast<quant::UniformQuantizedPerAxisType>();
+  auto resultQPAType =
+      resultType.dyn_cast<quant::UniformQuantizedPerAxisType>();
+  if (lhsQPAType || rhsQPAType) {
+    // add_c5
+    if (!resultQPAType)
+      return emitOptionalError(
+          location, "result is not per_axis quantized but lhs or rhs are");
+    // add_c6
+    if (lhsQPAType)
+      if (resultQPAType.getQuantizedDimension() !=
+          lhsQPAType.getQuantizedDimension())
+        return emitOptionalError(
+            location, "quantization_dimension of lhs and result are not same ",
+            lhsType, " vs ", resultType);
+    // add_c7
+    if (rhsQPAType)
+      if (resultQPAType.getQuantizedDimension() !=
+          rhsQPAType.getQuantizedDimension())
+        return emitOptionalError(
+            location, "quantization_dimension of rhs and result are not same ",
+            rhsType, " vs ", resultType);
+
+    return success();
+  }
+
+  return !resultQPAType
+             ? success()
+             : emitOptionalError(location,
+                                 "result per_axis quantized but none from rhs "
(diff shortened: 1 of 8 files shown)


stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -5058,62 +5058,85 @@ func.func @is_compatible_dynamism_mix(%arg0: tensor<?xf32>, %arg1: tensor<1xf32>
...
+  // expected-error@+1 {{quantization_dimension of rhs and result are not same}}
+  %0 = "stablehlo.add"(%arg0, %arg1) : (tensor<1x2x!quant.uniform<i8:f32:0, {1.0:17}>>, tensor<1x2x!quant.uniform<i8:f32:1, {1.0:17}>>) -> tensor<1x2x!quant.uniform<i8:f32:0, {1.0:17}>>
   func.return
 }
 
 // -----
 
 func.func @is_compatible_quant_signedness_mismatch(%arg0: tensor<1x!quant.uniform<i8:f32, 1.0:17>>) {
-  // expected-error@+1 {{op requires compatible types for all operands and results}}
+  // expected-error@+2 {{'stablehlo.add' op failed to infer returned types}}
```
sdasgup3, line 5138: remove `stablehlo.add` from error check

stablehlo/tests/stablehlo_refine_shapes.mlir
```diff
@@ -3,7 +3,7 @@
 func.func @error_illformed(%arg0: tensor<3xf32>, %arg1: tensor<4xf32>) -> tensor<?xf32> {
   %0 = stablehlo.abs %arg0 : (tensor<3xf32>) -> tensor<?xf32>
   %1 = stablehlo.abs %arg1 : (tensor<4xf32>) -> tensor<?xf32>
-  // expected-error@+1{{requires compatible types for all operands and results}}
+  // expected-error@+1{{requires the same shape for all operands and results}}
```
sdasgup3, line 6: [no action needed]

liked that the error messages are more informative.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -264,6 +281,76 @@ LogicalResult verifyPairwiseCompatibleShapes(TypeRange values) {
...
 }
 
+LogicalResult verifyAddOp(std::optional<Location> location, Type lhsType,
+                          Type rhsType, Type resultType) {
+  lhsType = getElementTypeOrSelf(lhsType);
+  rhsType = getElementTypeOrSelf(rhsType);
+  resultType = getElementTypeOrSelf(resultType);
+  llvm::SmallVector<Type, 3> typeEntries{lhsType, rhsType, resultType};
+  // add_c1
+  if (noneQuantized<quant::QuantizedType>(typeEntries)) {
```
GleasonK, line 291: My concern with this change is -- not all users of StableHLO want to think about quantization. It's a subset of total users who use this feature, but they'd have to filter through a very complicated implementation of verifyAdd to understand what their uses need. In that sense, perhaps we can write this function as:

```
if (anyQuantized(lhsType,rhsType,resultType))
  return verifyBinaryOpQuantizationConstraints(...);

return OpTrait::SameOperandsAndResultType<AddOp>::verifyTrait(getOperation());
```

GleasonK (changes requested):

stablehlo/dialect/TypeInference.cpp
abhigunj: Yeh.. agree. I will put the Q Verifier checks under dedicated verifyBinaryOpQuantizationConstraints function.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -65,6 +65,100 @@ limitations under the License.
...
+
+LogicalResult verifyBinaryOpQuantizationConstraints(
+    std::optional<Location> location, Type lhsType, Type rhsType,
+    Type resultType) {
+  lhsType = getElementTypeOrSelf(lhsType);
+  rhsType = getElementTypeOrSelf(rhsType);
+  resultType = getElementTypeOrSelf(resultType);
+  llvm::SmallVector<Type, 3> typeEntries{lhsType, rhsType, resultType};
+  // add_c1
+  if (noneQuantized<quant::QuantizedType>(typeEntries)) {
```
sdasgup3, line 98: This has already been checked at `verifyAddOp`? So you can remove this if

sdasgup3 (approved):

GleasonK (approved):

stablehlo/dialect/TypeInference.cpp
abhigunj: for : `return OpTrait::SameOperandsAndResultType<AddOp>::verifyTrait(getOperation());`

AddOp is  `StableHLO_BinaryElementwiseOp` and shape check is enabled by default. Custom verifier will check  same element type only.
tried  using
`mlir::OpTrait::SameOperandsAndResultElementType<::mlir::stablehlo::AddOp>::verifyTrait(op);` but bazel build failed. There is a circular dependency if I try to add`stablehlo_ops` to `stablehlo_type_inference` target to use `::mlir::stablehlo::AddOp`  at TypeInference.cpp. 

instead trait will do element type check using
`lhsType == resultType ` `lhsType == rhsType`

GleasonK: Can we use `SameOperandsAndResultElementType <Operation*>`? Otherwise we can use equality and our own mesage

abhigunj: Thank you. yeh.. tried that too before using quality and our own message.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -65,6 +65,94 @@ limitations under the License.
...
+  auto rhsQPAType = rhsType.dyn_cast<quant::UniformQuantizedPerAxisType>();
+  auto resultQPAType =
+      resultType.dyn_cast<quant::UniformQuantizedPerAxisType>();
+  if (lhsQPAType || rhsQPAType) {
+    // add_c5
+    if (!resultQPAType)
+      return emitOptionalError(
+          location, "result is not per_axis quantized but lhs or rhs are");
+    // add_c6
+    if (lhsQPAType)
```
GleasonK, line 131: Add curly braces to outer if to avoid dangling else

stablehlo/dialect/TypeInference.cpp
```diff
@@ -65,6 +65,94 @@ limitations under the License.
...
+      if (resultQPAType.getQuantizedDimension() !=
+          lhsQPAType.getQuantizedDimension())
+        return emitOptionalError(
+            location, "quantization_dimension of lhs and result are not same ",
+            lhsType, " vs ", resultType);
+    // add_c7
+    if (rhsQPAType)
```
GleasonK, line 138: Add curly braces to outer if to avoid dangling else

stablehlo/dialect/TypeInference.cpp
```diff
@@ -65,6 +65,94 @@ limitations under the License.
...
+      if (resultQPAType.getQuantizedDimension() !=
+          rhsQPAType.getQuantizedDimension())
+        return emitOptionalError(
+            location, "quantization_dimension of rhs and result are not same ",
+            rhsType, " vs ", resultType);
+
+    return success();
+  }
+
+  return !resultQPAType
```
GleasonK, line 148: Let's just use:

```
if (resultQPAType)
  emitOptionaleError(...);

return success();
```

Easier readability

GleasonK (approved):

PR #2128, opened 2024-03-27
Fix typo in error message

ghpvnist:


sdasgup3 (approved):

mlevesquedion (approved):
Yikes, thanks for catching that! I had originally changed "do no match rank" to "have different ranks" as I thought it was a bit clearer, but then I changed it back because the change was big and I didn't want to change too many things at once. When I changed it back I introduced a typo unfortunately... (https://github.com/openxla/stablehlo/pull/2119/files#diff-b9a4bbdcc79f8a54c314d05116c0086722f73c0cce075432fa49414204f63a66)

PR #2130, opened 2024-03-27
Check whitespace in all files except bytecode and image/vector files

### LINKED ISSUE #2129: Figure out the root cause of #2126
We have a CI to prevent missing terminating newline, but recent new python file seem to have bypassed it.

mlevesquedion:
The root cause of #2129 is that Python files (ending in `.py`) were excluded from the whitespace check. I originally proposed expanding the list of included files to include Python files, but @ghpvnist pointed out that there are more files that need this check than files that don't and suggested using an exclude list, so I did that instead.

Also, use `-E` and rewrite the pattern to allow the exclude list to be expanded more easily in the future.

Fixes #2129

### DIFF
--- a/build_tools/github_actions/lint_whitespace_checks.sh
+++ b/build_tools/github_actions/lint_whitespace_checks.sh
@@ -40,7 +40,7 @@ if [[ $# -ne 0 ]] ; then
 fi
 
 echo "Gathering changed files..."
-mapfile -t CHANGED_FILES < <(git diff "$BASE_BRANCH" HEAD --name-only --diff-filter=d | grep -E '.*\.(cpp|h|md|mlir|sh|td|txt|ya?ml|py)$')
+mapfile -t CHANGED_FILES < <(git diff "$BASE_BRANCH" HEAD --name-only --diff-filter=d | grep -Ev '.*\.(bc|png|svg)$')
 if (( ${#CHANGED_FILES[@]} == 0 )); then
   echo "No files to check."
   exit 0
--- a/build_tools/github_actions/lint_whitespace_checks.sh
+++ b/build_tools/github_actions/lint_whitespace_checks.sh
@@ -40,7 +40,7 @@ if [[ $# -ne 0 ]] ; then
 fi
 
 echo "Gathering changed files..."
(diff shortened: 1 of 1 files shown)


build_tools/github_actions/lint_whitespace_checks.sh
```diff
@@ -40,7 +40,7 @@ if [[ $# -ne 0 ]] ; then
 fi
 
 echo "Gathering changed files..."
-mapfile -t CHANGED_FILES < <(git diff "$BASE_BRANCH" HEAD --name-only --diff-filter=d | grep '.*\.cpp$\|.*\.h$\|.*\.md$\|.*\.mlir$\|.*\.sh$\|.*\.td$\|.*\.txt$\|.*\.yml$\|.*\.yaml$')
+mapfile -t CHANGED_FILES < <(git diff "$BASE_BRANCH" HEAD --name-only --diff-filter=d | grep -E '.*\.(cpp|h|md|mlir|sh|td|txt|ya?ml|py)$')
```
ghpvnist, line 43: In our repo, since every file (except `.bc`) should have terminating newline, wdyt about changing this to exclude list? This would also simplify the grep script to something like:
`grep -v '\.bc$'`

ghpvnist (approved):
Thank you for the change! Left an optional comment.

GleasonK (approved):

build_tools/github_actions/lint_whitespace_checks.sh
mlevesquedion: This is a great point! Here is the list of extensions we use:

```
.bazel
.bc
.bzl
.cmake
.cpp
.h
.in
.json
.lock
.md
.mlir
.pdll
.png
.py
.sh
.svg
.td
.txt
.yaml
```

Of those, I think `.bc`, `.in` (generated files), `.png` and `.svg` can be safely excluded.

Indeed, way more files and file extensions should be checked for whitespace than not. In the future, if new file extensions come up, we will likely want them to be checked rather than excluded, so using an exclude list makes sense. I will make the change.

GleasonK: We could do allowlist or denylist. Maybe we try the denylist for a bit. Currently in our repo, filetypes that don't need to end in newline are: png, svg, bc.

If someone uploaded a jpeg during RFC process, this lint will fail. Maybe that's OK at the cost of a more sound lint?

GleasonK: Didn't notice the new comment, on the same page :)

PR #2125, opened 2024-03-25
Add support of float type in specialization of ConvertOp

mvpant:
This adds support of following specializations in ConvertOp:

Integer -> Integer (old)
Integer -> FloatType (new)
FloatType -> FloatType (new)
FloatType -> Integer (new)

I have considered approach used in `convertElementsAttr` from `mlir-hlo/utils/convert_op_folder.cc`, but `DenseIntOrFPElementsAttr::mapValues` doesn't give an option to stop conversion in case of failure.

This implementation works for my case:

```
%cst = arith.constant dense<1.000000e+00> : tensor<1xf64>
%cst_0 = arith.constant dense<128> : tensor<1xi64>
%cst_1 = arith.constant dense<9.9999999999999998E-13> : tensor<1xf64>
%61 = stablehlo.constant dense<8.000000e+00> : tensor<f64>
%73 = stablehlo.constant dense<-3.4028234663852886E+38> : tensor<f64>
%76 = stablehlo.constant dense<1> : tensor<2x128xi32> 
...
%86 = stablehlo.convert %76 : (tensor<2x128xi32>) -> tensor<2x128xf32>
...
%90 = stablehlo.convert %cst : (tensor<1xf64>) -> tensor<1xf32>
...
%94 = stablehlo.convert %73 : (tensor<f64>) -> tensor<f32>
...
%106 = stablehlo.convert %cst_0 : (tensor<1xi64>) -> tensor<1xf32>
...
%117 = stablehlo.convert %cst_1 : (tensor<1xf64>) -> tensor<1xf32>
...
%151 = stablehlo.convert %61 : (tensor<f64>) -> tensor<f32>
```

```
// other constants are equivalent/folded completely
%73 => stablehlo.constant dense<-3.40282347E+38> : tensor<2x1x1x128xf32>
%cst_1 => stablehlo.constant dense<9.99999996E-13> : tensor<2x128x1xf32>
```

but i am not sure how sound it is with [operator specification](https://github.com/tensorflow/mlir-hlo/blob/master/stablehlo/docs/spec.md#convert), mainly how conversion should be performed.

PR #2132, opened 2024-03-27
Add EvalIotaOp pattern to StablehloAggressiveFolder

mvpant:
This PR adds support of IotaOp specialization:

```
func.func @eval_iota() -> tensor<3x4x5xi32> {
  %1 = stablehlo.iota dim = 1 : tensor<3x4x5xi32>
  func.return %1 : tensor<3x4x5xi32>
}
```
is transformed into:
```
func.func @eval_iota() -> tensor<3x4x5xi32> {
  %0 = stablehlo.constant dense<
            [[[0, 0, 0, 0, 0], [1, 1, 1, 1, 1], [2, 2, 2, 2, 2], [3, 3, 3, 3, 3]],
             [[0, 0, 0, 0, 0], [1, 1, 1, 1, 1], [2, 2, 2, 2, 2], [3, 3, 3, 3, 3]],
             [[0, 0, 0, 0, 0], [1, 1, 1, 1, 1], [2, 2, 2, 2, 2], [3, 3, 3, 3, 3]]]> : tensor<3x4x5xi32>
  func.return %0 : tensor<3x4x5xi32>
}
```

### DIFF
--- a/stablehlo/tests/stablehlo_aggressive_folder.mlir
+++ b/stablehlo/tests/stablehlo_aggressive_folder.mlir
@@ -1,0 +1,28 @@
+// RUN: stablehlo-opt --stablehlo-aggressive-folder --split-input-file --verify-diagnostics %s | FileCheck %s
+
+
+// CHECK-LABEL: func @eval_iota
+func.func @eval_iota() -> (tensor<3x4x5xi32>, tensor<3x4x5xi32>, tensor<3x4x5xi32>) {
+  // CHECK-NOT: stablehlo.eval_iota
+  // CHECK: [[RESULT0:%.*]] = stablehlo.constant dense<{{\[\[}}[0, 0, 0, 0, 0], [0, 0, 0, 0, 0], [0, 0, 0, 0, 0], [0, 0, 0, 0, 0]], {{\[}}[1, 1, 1, 1, 1], [1, 1, 1, 1, 1], [1, 1, 1, 1, 1], [1, 1, 1, 1, 1]], {{\[}}[2, 2, 2, 2, 2], [2, 2, 2, 2, 2], [2, 2, 2, 2, 2], [2, 2, 2, 2, 2]]]> : tensor<3x4x5xi32>
+  // CHECK: [[RESULT1:%.*]] = stablehlo.constant dense<{{\[\[}}[0, 0, 0, 0, 0], [1, 1, 1, 1, 1], [2, 2, 2, 2, 2], [3, 3, 3, 3, 3]], {{\[}}[0, 0, 0, 0, 0], [1, 1, 1, 1, 1], [2, 2, 2, 2, 2], [3, 3, 3, 3, 3]], {{\[}}[0, 0, 0, 0, 0], [1, 1, 1, 1, 1], [2, 2, 2, 2, 2], [3, 3, 3, 3, 3]]]> : tensor<3x4x5xi32>
+  // CHECK: [[RESULT2:%.*]] = stablehlo.constant dense<{{\[\[}}[0, 1, 2, 3, 4], [0, 1, 2, 3, 4], [0, 1, 2, 3, 4], [0, 1, 2, 3, 4]], {{\[}}[0, 1, 2, 3, 4], [0, 1, 2, 3, 4], [0, 1, 2, 3, 4], [0, 1, 2, 3, 4]], {{\[}}[0, 1, 2, 3, 4], [0, 1, 2, 3, 4], [0, 1, 2, 3, 4], [0, 1, 2, 3, 4]]]> : tensor<3x4x5xi32>
+  // CHECK: return [[RESULT0]], [[RESULT1]], [[RESULT2]]
+  %0 = stablehlo.iota dim = 0 : tensor<3x4x5xi32>
+  %1 = stablehlo.iota dim = 1 : tensor<3x4x5xi32>
+  %2 = stablehlo.iota dim = 2 : tensor<3x4x5xi32>
+  func.return %0, %1, %2 : tensor<3x4x5xi32>, tensor<3x4x5xi32>, tensor<3x4x5xi32>
+}
+
+// -----
+
+// CHECK-LABEL: func @eval_iota_zero_dimension
+func.func @eval_iota_zero_dimension() -> (tensor<0xi32>, tensor<5x0x2xi32>) {
+  // CHECK-NOT: stablehlo.eval_iota
+  // CHECK: [[RESULT0:%.*]] = stablehlo.constant dense<> : tensor<0xi32>
+  // CHECK: [[RESULT1:%.*]] = stablehlo.constant dense<> : tensor<5x0x2xi32>
+  // CHECK: return [[RESULT0]], [[RESULT1]]
+  %0 = stablehlo.iota dim = 0 : tensor<0xi32>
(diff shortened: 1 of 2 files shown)


stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -465,6 +465,45 @@ struct EvalGetDimensionSizeOpPattern
...
+    auto dimension = op.getIotaDimension();
+    auto resultBitWidth = elementType.getIntOrFloatBitWidth();
+
+    llvm::SmallVector<APInt> values;
+    values.reserve(outputSize);
+
+    int64_t sequences = 1;
+    int64_t sequenceMax = resultType.getDimSize(dimension);
+    int64_t elementRepetitions = 1;
+    for (uint64_t i = 0; i < resultType.getShape().size(); i++) {
```
mlevesquedion, line 488: Can `getShape().size()` be replaced with `getRank()`?

stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -465,6 +465,45 @@ struct EvalGetDimensionSizeOpPattern
...
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(IotaOp op,
+                                PatternRewriter& rewriter) const override {
+    auto resultType = op.getType().cast<RankedTensorType>();
+    auto elementType = resultType.getElementType();
+
+    if (!elementType.isInteger())
+      return rewriter.notifyMatchFailure(op, "expected integer result type");
+
+    auto outputSize = resultType.getNumElements();
```
mlevesquedion, line 478: Consider adding an early return for when this is zero.

mlevesquedion (commented):
Thanks for the PR!

I wasn't sure what to expect from the PR name. Consider changing the name to e.g. "Add EvalIota pattern to RefineShapes" to be more specific about what the change is doing.

GleasonK:
After thinking about this more, let's start a new pass `StablehloAggressiveFolder` (someone from StableHLO team can set up the basic infra if you'd like, let me know). This pass can have all folder patterns, and I think we can add options for "fold_float" to allow potentially lossy computations.

We can then migrate the existing shape refinement pass to include these patterns and mark specific ops as illegal, and only use the integer folders.

Rationale: We only want to fold shape computations in this pass, not anything else. I'm worried that folding too much may change the behavior of some existing models / workflows using shape refinement to _only_ fold shapes with the assumption that non-shape-related computations should stay the same.

mvpant:
> After thinking about this more, let's start a new pass `StablehloAggressiveFolder` (someone from StableHLO team can set up the basic infra if you'd like, let me know). This pass can have all folder patterns, and I think we can add options for "fold_float" to allow potentially lossy computations.

Sounds great, definitely with "fold_float" option. if you have limited resources, i might try to implement this approach on the next week.

> We can then migrate the existing shape refinement pass to include these patterns and mark specific ops as illegal, and only use the integer folders.

```
  patterns->add<EvalAddOpPattern>(context);
  patterns->add<EvalAndOpPattern>(context);
  patterns->add<EvalBroadcastInDimOpPattern>(context);
  patterns->add<EvalClampOpPattern>(context);
  patterns->add<EvalCompareOpPattern>(context);
  patterns->add<EvalComputeReshapeShapeOpPattern>(context); <===
  patterns->add<EvalConcatenateOpPattern>(context);
  patterns->add<EvalConvertOpPattern>(context);
  patterns->add<EvalDivOpPattern>(context);
  patterns->add<EvalGetDimensionSizeOpPattern>(context); <===
  patterns->add<EvalIotaOpPattern>(context);
  patterns->add<EvalMaxOpPattern>(context);
  patterns->add<EvalMinOpPattern>(context);
  patterns->add<EvalMulOpPattern>(context);
  patterns->add<EvalOrOpPattern>(context);
  patterns->add<EvalRemOpPattern>(context);
  patterns->add<EvalReshapeOpPattern>(context); <===
  patterns->add<EvalSelectOpPattern>(context);
  patterns->add<EvalSignOpPattern>(context);
  patterns->add<EvalSliceOpPattern>(context);
  patterns->add<EvalSubtractOpPattern>(context);
  ...
```

From the first glance only patterns marked with `<===` should be used in shape refinement? Or everything that involves integer computations should also stay, like `EvalAddOpPattern`? 

> ... mark specific ops as illegal, ...

I didn`t catch the meaning, can you elaborate a bit?

Regarding https://github.com/openxla/stablehlo/pull/2125,  i ll stall it together with this pull request until new pass is done.

mvpant:
@GleasonK If you haven’t already begun, I`ll start working on a patch to extract folders into separate pass.

GleasonK:
Also something I learned about recently that may be of interest:
https://github.com/EnzymeAD/Enzyme-JAX/blob/main/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp

There's a lot of StableHLO canonicalizations / rewrites in the enzyme project. We'll be upstreaming some of these over the next months. Note that the file as-is includes a full copy-paste of StablehloAggressiveSimplification, but there are some new patterns in there as well!

GleasonK:
^cc @mvpant

GleasonK:
With the folders PR merged, this can be rebased and add this pattern to folders, but not shape folders (iota is not used on any common shape computations currently)

stablehlo/transforms/StablehloRefineShapes.cpp
mvpant: There is a test with 0 elements, the code works as is. Adding an early return would still require an empty `SmallVector` to create `DenseIntElementsAttr`.

stablehlo/tests/stablehlo_refine_shapes.mlir
```diff
@@ -305,6 +305,34 @@ func.func @eval_get_dimension_size(%arg0: tensor<4xf32>) -> tensor<i32> {
 
 // -----
 
+// CHECK-LABEL: func @eval_iota
+func.func @eval_iota() -> (tensor<3x4x5xi32>, tensor<3x4x5xi32>, tensor<3x4x5xi32>) {
+  // CHECK-NOT: stablehlo.eval_iota
+  // CHECK: [[RESULT0:%.*]] = stablehlo.constant dense<{{\[\[}}[0, 0, 0, 0, 0], [0, 0, 0, 0, 0], [0, 0, 0, 0, 0], [0, 0, 0, 0, 0]], {{\[}}[1, 1, 1, 1, 1], [1, 1, 1, 1, 1], [1, 1, 1, 1, 1], [1, 1, 1, 1, 1]], {{\[}}[2, 2, 2, 2, 2], [2, 2, 2, 2, 2], [2, 2, 2, 2, 2], [2, 2, 2, 2, 2]]]> : tensor<3x4x5xi32>
```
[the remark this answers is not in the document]
mvpant: have modified the expected result to make it more readable

GleasonK (approved):
LGTM! Re-requested review from @mlevesquedion as well.

stablehlo/transforms/StablehloRefineShapes.cpp
mvpant: on the second thought some artificial examples might take a while to execute without early return, e.g. 
`%1 = stablehlo.iota dim = 1 : tensor<2000000000x0x2xi32>`

PR #2135, opened 2024-03-27
Delete rank_specialization_cluster ops

mlevesquedion:
(P1) of the Dynamism RFC.

These ops are no longer used by KernelGen and can safely be deleted.

### DIFF
--- a/stablehlo/dialect/ChloOps.cpp
+++ b/stablehlo/dialect/ChloOps.cpp
@@ -413,41 +413,6 @@ LogicalResult BroadcastSelectOp::reifyReturnTypeShapes(
 }
 
 //===----------------------------------------------------------------------===//
-// RankSpecializationClusterOp
-//===----------------------------------------------------------------------===//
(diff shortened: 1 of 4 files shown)


ghpvnist (approved):
LGTM. We still plan to migrate these to `chlo_legacy` dialect right? If so, could you also update the description?

mlevesquedion:
> LGTM. We still plan to migrate these to `chlo_legacy` dialect right? If so, could you also update the description?

For these ops specifically I think it's safe to just delete them. They were only used by TF's KernelGen infrastructure, and I deleted the uses some time ago: https://github.com/openxla/xla/commit/ff8adb23f0155fc92de7b91bd2c0654e0882b476

Given how specific the ops are, I think it's very unlikely anyone else was using these, so I don't see a benefit in moving them to a new `chlo_legacy` dialect.

WDYT?

GleasonK (approved):

PR #2136, opened 2024-03-29
Add spec warning about shape mismatches

mlevesquedion:
This implements part 1 of (P2) from the Dynamism 101 RFC: https://github.com/openxla/stablehlo/blob/main/rfcs/20230704-dynamism-101.md#-p2-ratify-the-existing-convention-for-shape-mismatches-constituting-undefined-behavior

In a future change, I will implement `ConditionallySpeculatable` where relevant. Many ops have the `Pure` trait but they are not actually `Pure` in all cases because shape mismatches are undefined behavior.

#1991

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -6263,13 +6263,13 @@ future ([#1157](https://github.com/openxla/stablehlo/issues/1157)).
 #### Shape mismatches
 
 StableHLO supports dynamically-shaped tensors. However, shapes have to agree at
-runtime, otherwise the behavior is implementation-defined. StableHLO does not
-explicitly provide an op that can assert that a tensor has a given shape at
-runtime. Generating correct code is the responsibility of the producer.
+runtime, otherwise the behavior is undefined. StableHLO does not explicitly
+provide an op that can assert that a tensor has a given shape at runtime.
+Generating correct code is the responsibility of the producer.
 
 As a specific example, the below program is valid. However, at runtime, the
 exact shapes of `%arg0` and `%arg1` will have to be the same, otherwise the
-result is unspecified:
+behavior of the program is undefined:
 
 ```
 func.func @foo(%arg0: tensor<?xi32>, %arg1: tensor<?xi32>) -> tensor<?xi32> {
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -6271,7 +6271,7 @@ As a specific example, the below program is valid. However, at runtime, the
 exact shapes of `%arg0` and `%arg1` will have to be the same, otherwise the
 behavior of the program is undefined:
 
-```
+```mlir
 func.func @foo(%arg0: tensor<?xi32>, %arg1: tensor<?xi32>) -> tensor<?xi32> {
     %0 = stablehlo.add %arg0, %arg1 : tensor<?xi32>
     return %0 : tensor<?xi32>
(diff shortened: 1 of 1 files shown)


docs/spec.md
```diff
@@ -6260,6 +6260,26 @@ out-of-bounds accesses, etc. Unless explicitly called out, all these errors
 result in implementation-defined behavior, but this may change in the
 future ([#1157](https://github.com/openxla/stablehlo/issues/1157)).
 
+#### Shape mismatches
+
+StableHLO supports dynamically-shaped tensors. However, shapes have to agree at
+runtime, otherwise the behavior is implementation-defined. StableHLO does not
```
sdasgup3, line 6266: Isn't this undefined as opposed to implementation defined? 

For undefined-ness the notion is no defined behavior and the spec offers absolutely no guarantees about what might happen if these are triggered. Here we would like to delegate the ownership of providing those guarantees to the producer of stablehlo.

Moreover, implementation definiteness has some predictability within a particular implementation. This spirit comes from the fact that there are >1 possible behaviors which differ between implementations.  In this case, there is no defined behavior of addition with varying tensors.

sdasgup3 (commented):
mostly lgtm with a comment.

docs/spec.md
mlevesquedion: Thank you for explaining the difference between undefined and implementation defined! When reading some documents about (P2) I saw both terms being used and I thought they meant the same thing: if something is undefined by the spec, then the implementation can do whatever it wants, so it is implementation defined. Turns out, "implementation defined" applies to correct programs and leaves the choice to the implementation, whereas "undefined" applies to incorrect programs. (Implicitly though, what happens may depend on the implementation, which is why I used the two interchangeably at first.)

I changed it to say "undefined", thanks!

sdasgup3 (approved):

docs/spec.md
```diff
@@ -6260,6 +6260,26 @@ out-of-bounds accesses, etc. Unless explicitly called out, all these errors
 result in implementation-defined behavior, but this may change in the
 future ([#1157](https://github.com/openxla/stablehlo/issues/1157)).
 
+#### Shape mismatches
```
ghpvnist, line 6263: Could you move this section so it's in alphabetical order?

ghpvnist (approved):
LGTM with minor comment

docs/spec.md
```diff
@@ -6269,6 +6271,24 @@ continue execution without raising the corresponding status flag; similar to
...
+StableHLO supports dynamically-shaped tensors. However, shapes have to agree at
+runtime, otherwise the behavior is undefined. StableHLO does not explicitly
+provide an op that can assert that a tensor has a given shape at runtime.
+Generating correct code is the responsibility of the producer.
+
+As a specific example, the below program is valid. However, at runtime, the
+exact shapes of `%arg0` and `%arg1` will have to be the same, otherwise the
+behavior of the program is undefined:
+
+```mlir
```
GleasonK, line 6285: We have the infrastructure to make a `verifyValidShapePolymorphism` pass:

```
RefineArguments
RefineShapes
CanonicalizeDynamism
VerifyNoDynamicShapes (not a pass but can walk)
```

Not sure how to best capture that here if at all

GleasonK (approved):

PR #2141, opened 2024-03-29
Remove `RankedTensorType` verification check for `CholeskyOp`

### LINKED ISSUE #2138: Remove `RankedTensorType` verification check for `CholeskyOp`
As part of a bulk change in #2119, this seems to have slipped through the cracks.

ghpvnist:
closes #2138


GleasonK (approved):

mlevesquedion (approved):
Thanks! For context, I intentionally excluded this from #2119 as I felt the change was already big enough. I plan to do more cleanup of rankedness checks in the future, but feel free to keep helping! Just let me know what you're planning to do so we don't duplicate work.

ghpvnist:
Thanks for the context! I ran into it by chance while working with this op in another context, so I decided I'd fix it on a whim.

PR #2146, opened 2024-03-29
Implement ConditionallySpeculatable for DynamicIota

mlevesquedion:
Starting with a trivial op to get early feedback on the general direction.

In future changes I will implement the interface for additional ops. I will try to reuse as much code as possible, e.g. for binary ops I think it will be possible to share the implementation.

I am using LICM for testing because [that is what MLIR does](https://github.com/llvm/llvm-project/blob/7467dc188a0c63b32bd3068fc98ae77f86285c0d/mlir/test/Transforms/loop-invariant-code-motion.mlir#L522), but I will investigate using [test utils](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/TestUtils.td) instead as that may be simpler and more reliable.


stablehlo/tests/ops_speculatability.mlir
```diff
@@ -0,0 +1,24 @@
...
+func.func @dynamic_iota(%unknown_shape: tensor<2xi32>) {
+  %constant_shape = stablehlo.constant dense<[3, 4]> : tensor<2xi32>
+  %speculatable_0 = stablehlo.dynamic_iota %constant_shape, dim = 0 : (tensor<2xi32>) -> tensor<3x4xi64>
+  %speculatable_1 = stablehlo.dynamic_iota %constant_shape, dim = 0 : (tensor<2xi32>) -> tensor<3x?xi64>
+  %speculatable_2 = stablehlo.dynamic_iota %constant_shape, dim = 0 : (tensor<2xi32>) -> tensor<?x4xi64>
+  %speculatable_3 = stablehlo.dynamic_iota %constant_shape, dim = 0 : (tensor<2xi32>) -> tensor<?x?xi64>
+  %not_speculatable_0 = stablehlo.dynamic_iota %unknown_shape, dim = 0 : (tensor<2xi32>) -> tensor<3x4xi64>
+  %not_speculatable_1 = stablehlo.dynamic_iota %unknown_shape, dim = 0 : (tensor<2xi32>) -> tensor<3x?xi64>
+  %not_speculatable_2 = stablehlo.dynamic_iota %unknown_shape, dim = 0 : (tensor<2xi32>) -> tensor<?x4xi64>
+  %speculatable_4 = stablehlo.dynamic_iota %unknown_shape, dim = 0 : (tensor<2xi32>) -> tensor<?x?xi64>
```
sdasgup3, line 14: can we put is next to `speculatable_3`?

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -771,6 +771,25 @@ LogicalResult DynamicIotaOp::verify() {
                                   getIotaDimension(), getResult());
 }
 
+mlir::Speculation::Speculatability DynamicIotaOp::getSpeculatability() {
+  // If the output shape operand is known, for each result type dimension:
+  // 1. If it is static, the verifier has already checked it.
+  // 2. Otherwise, it is dynamic, so there cannot be a mismatch.
```
sdasgup3, line 777: "If the output shape operand is known, for each result type dimension" implies (1). 

(2) seems confusing to me. By "it is dynamic" do you mean if the output shape is dynamic of the result shape is dynamic?

sdasgup3: overall the logic lgtm.

stablehlo/tests/ops_speculatability.mlir
mlevesquedion: IMO it's more readable this way because this makes it easier to verify that all the cases are covered. In order:
1. shape argument: constant, unknown
2. dimensions: 3x4, 3x?, ?x4, ?x?

WDYT?

stablehlo/dialect/StablehloOps.cpp
mlevesquedion: > "If the output shape operand is known, for each result type dimension" implies (1).

I'm not sure I follow. Can you expand on that?

> (2) seems confusing to me. By "it is dynamic" do you mean if the output shape is dynamic of the result shape is dynamic?

I mean if a specific dimension in the result type's shape is dynamic.

Just to make sure we're on the same page:
- If the output shape operand is known, each dimension of the output shape operand is static.
- The comment makes a distinction between the *output shape operand* and the *shape of the result type*. If this is confusing, please let me know and I can expand the comment to clarify.
-

sdasgup3: Thanks, we are on the same page. (2) was not clear to me whether it talks about output shape or result type when it says "it is dynamic". It would be great if you make it clear.

sdasgup3 (approved):

stablehlo/dialect/StablehloOps.cpp
mlevesquedion: Okay cool, thanks for the feedback! I have updated the comment, please let me know if you think it should still be clarified.

GleasonK (approved):

PR #2149, opened 2024-04-02
Support sparsity through StableHLO => Linalg lowering

yinying-lisa-li:
After StableHLO => Linalg lowering was implemented in the [linalg RFC](https://github.com/openxla/stablehlo/pull/1610), most of the ops stated by @aartbik in the [sparsity RFC](https://github.com/openxla/stablehlo/pull/1143) can generate correct linalg code with sparse encodings. And it can then be lowered through mlir-opt with --sparsifier pipeline and run with mlir-runner. 

The `sparse.mlir` file in this PR checks the generated linalg code with sparse encodings. These are the same tests for the dense ops but with sparse input/output.

The only op that couldn't be lowered to correct linalg with sparsity is `stablehlo.concatenate`. Therefore, after discussing with @GleasonK, I added a flag `enable-sparse-ops` to the linalg pass and rewrite `stablehlo.concatenate` to `sparse_tensor.concatenate` when the flag is set to true and input/output have sparse encodings. If there're more linalg converters that need modification, a new pass for sparsity will be created to run before linalg pass.

### DIFF
--- a/stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
+++ b/stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
@@ -967,7 +967,7 @@ struct RealDynamicSliceConverter final
       ConversionPatternRewriter &rewriter) const override {
     Location loc = realDynamicSliceOp.getLoc();
     auto argType = llvm::cast<RankedTensorType>(adaptor.getOperand().getType());
-
+    
     Type dimElementType = getElementTypeOrSelf(adaptor.getStartIndices());
     if (getElementTypeOrSelf(adaptor.getLimitIndices()) != dimElementType ||
         getElementTypeOrSelf(adaptor.getStrides()) != dimElementType) {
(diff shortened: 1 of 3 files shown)


stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
```diff
@@ -1344,6 +1344,45 @@ struct ConcatenateConverter final
...
 
+/// Converts stablehlo.concatenate operation to a sparse_tensor.concatenate op.
+struct SparseConcatenateConverter final
+    : OpConversionPattern<mlir::stablehlo::ConcatenateOp> {
+  using OpConversionPattern::OpConversionPattern;
+
+  SparseConcatenateConverter(const TypeConverter &converter,
+                             MLIRContext *context)
+      : OpConversionPattern<mlir::stablehlo::ConcatenateOp>(converter, context,
+                                                            /*benefit=*/10) {}
```
GleasonK, line 1355: Another option from benefit actually would just be to make the other convert patter return failure on sparse operand types. I'll let you decide what's best, but IIUC the current behavior of the convert op with sparse inputs is incorrect? Probably ok to error instead of produce invalid IR

yinying-lisa-li: Yes, if the input/ouput have sparse encodings, the current convert op will produce invalid IR. Since dense converter have lower benefit, I could add `if (isResultSparse || isAnyOperandSparse) return failure();` in the dense converter to address this issue. Is that what you meant by using benefit to achieve this?

GleasonK: Initially I suggested benefit thinking of exactly what you ended up with here, to override another pattern effectively.

But yes as you suggest, add `if (isResultSparse || isAnyOperandSparse) return failure()` to the dense concat pattern is probably a better solution, can remove benefit all together since the other pattern will just fail and this one will succeed.

Can also give a reason in the other pattern:

```
if (isResultSparse || isAnyOperandSparse)
  return rewriter.notifyMatchFailure(op, "ConcatenateConverter cannot legalize sparse types");
```

yinying-lisa-li: Sounds good!

GleasonK (approved):

PR #2150, opened 2024-04-02
Implement ConditionallySpeculatable for all UnaryElementwise ops

mlevesquedion:
Note: this PR builds on top of https://github.com/openxla/stablehlo/pull/2146. Please see the 2nd commit for changes that are unique to this PR.

A somewhat unfortunate thing is that we allow programs like:

```
func.func @foo(%arg: tensor<?xi64>) {
  %0 = stablehlo.abs %arg : (tensor<?xi64>) -> tensor<2xi64>
  return
}
```

Such a program is legal, but doesn't make much sense: in what scenario would we know the output shape, but not the input shape? If it weren't for such programs, the ops would all be `Pure`, but because at runtime dynamic dimensions in the input type could differ from static dimensions in the output type, these ops have to be `ConditionallySpeculatable`. Whether or not we should allow such programs could be a future discussion.

Also note that the spec asserts that floating point exceptions are defined: https://openxla.org/stablehlo/spec#floating-point_exceptions 

https://github.com/openxla/stablehlo/issues/1991

### DIFF
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -179,7 +179,7 @@ def StableHLO_CreateTokenOp : StableHLO_Op<"create_token", [Pure,
 
 class StableHLO_UnaryElementwiseOp<string mnemonic, list<Trait> traits,
     Type OperandType, Type ResultType = OperandType> : StableHLO_Op<mnemonic, traits # [Elementwise,
-    InferShapedTypeOpInterface, SameOperandsAndResultShape]> {
+    InferShapedTypeOpInterface, SameOperandsAndResultShape, HLO_UnaryElementwiseSpeculatable, NoMemoryEffect]> {
   let arguments = (ins OperandType:$operand);
   let results = (outs ResultType:$result);
   let extraClassDeclaration = commonClassDeclaration # [{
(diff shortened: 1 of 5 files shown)


stablehlo/dialect/StablehloOps.td
```diff
@@ -194,7 +199,7 @@ class StableHLO_UnaryElementwiseOp<string mnemonic, list<Trait> traits,
 
 // Abs supports complex to real, so element type is not guaranteed to match.
 def StableHLO_AbsOp: StableHLO_UnaryElementwiseOp<"abs",
-    [Pure, DeclareOpInterfaceMethods<InferTypeOpInterface>],
+    [DeclareOpInterfaceMethods<InferTypeOpInterface>],
```
GleasonK, line 202: To clarify - `Pure` can be deleted here and after because the memory effects / speculatability is now captured in the parent `StableHLO_UnaryElementwiseOp` class?

GleasonK (approved):
LGTM with one question in comments

stablehlo/dialect/StablehloOps.td
mlevesquedion: That's right! `Pure` is `def Pure : TraitList<[AlwaysSpeculatable, NoMemoryEffect]>;` -- the parent class now specifies `HLO_UnaryElementwiseSpeculatable` and `NoMemoryEffect`, so `Pure` is no longer needed.

mlevesquedion:
> LGTM with one question in comments

Thanks for the review! I replied to your question: https://github.com/openxla/stablehlo/pull/2150#discussion_r1548687759

PR #2152, opened 2024-04-03
Avoid calling functions in asserts

penguin-wwy:
Due to a problem with the compilation options, the `shapeSource.reifyReturnTypeShapes` was not called, fix it.

### DIFF
--- a/stablehlo/conversions/linalg/transforms/LegalizeToLinalgUtils.cpp
+++ b/stablehlo/conversions/linalg/transforms/LegalizeToLinalgUtils.cpp
@@ -77,9 +77,9 @@ Value getEmptyTensorFor(OpBuilder &b, Location loc, ShapedType resultType,
     // Ask the op for its output shape.
     auto shapeSource = cast<InferShapedTypeOpInterface>(op);
     SmallVector<Value, 1> reifiedShapes;
-    assert(succeeded(
-               shapeSource.reifyReturnTypeShapes(b, operands, reifiedShapes)) &&
-           "could not reify");
+    if (failed(shapeSource.reifyReturnTypeShapes(b, operands, reifiedShapes))) {
+      llvm_unreachable("could not reify");
+    }
     assert(reifiedShapes.size() == 1 && "Expected one reified result");
     // Construct sizes for the required dimensions.
     for (const auto &en : llvm::enumerate(resultType.getShape())) {
--- a/stablehlo/conversions/linalg/transforms/LegalizeToLinalgUtils.cpp
+++ b/stablehlo/conversions/linalg/transforms/LegalizeToLinalgUtils.cpp
@@ -78,7 +78,7 @@ Value getEmptyTensorFor(OpBuilder &b, Location loc, ShapedType resultType,
     auto shapeSource = cast<InferShapedTypeOpInterface>(op);
     SmallVector<Value, 1> reifiedShapes;
     if (failed(shapeSource.reifyReturnTypeShapes(b, operands, reifiedShapes))) {
-      llvm_unreachable("could not reify");
+      llvm::report_fatal_error("could not reify");
     }
     assert(reifiedShapes.size() == 1 && "Expected one reified result");
     // Construct sizes for the required dimensions.


mlevesquedion:
> Due to a problem with the compilation options, the `shapeSource.reifyReturnTypeShapes` was not called, fix it.

Hi! Thanks for your contribution :)

It's not super clear to me why `assert` is inadequate in this case. How did you run into this? If there is a problem with the compilation options, maybe we should change the compilation options as well.

I'm also not sure about the use of `llvm_unreachable` here. It seems we use `assert` in a lot of places in our code, but I see only a few uses of `llvm_unreachable`. The style guide recommends using `llvm_unreachable for unreachable code (https://llvm.org/docs/CodingStandards.html#assert-liberally), but some people recommend `assert(0)`: https://discourse.llvm.org/t/llvm-unreachable-is-widely-misused/60587/11?page=2.

stablehlo/conversions/linalg/transforms/LegalizeToLinalgUtils.cpp
```diff
@@ -77,9 +77,9 @@ Value getEmptyTensorFor(OpBuilder &b, Location loc, ShapedType resultType,
     // Ask the op for its output shape.
     auto shapeSource = cast<InferShapedTypeOpInterface>(op);
     SmallVector<Value, 1> reifiedShapes;
-    assert(succeeded(
-               shapeSource.reifyReturnTypeShapes(b, operands, reifiedShapes)) &&
-           "could not reify");
+    if (failed(shapeSource.reifyReturnTypeShapes(b, operands, reifiedShapes))) {
+      llvm_unreachable("could not reify");
```
GleasonK, line 81: Use `llvm::report_fatal_error("could not reify");`

GleasonK (changes requested):

penguin-wwy:
> > Due to a problem with the compilation options, the `shapeSource.reifyReturnTypeShapes` was not called, fix it.
> 
> Hi! Thanks for your contribution :)
> 
> It's not super clear to me why `assert` is inadequate in this case. How did you run into this? If there is a problem with the compilation options, maybe we should change the compilation options as well.
> 
> I'm also not sure about the use of `llvm_unreachable` here. It seems we use `assert` in a lot of places in our code, but I see only a few uses of `llvm_unreachable`. The style guide recommends using `llvm_unreachable for unreachable code (https://llvm.org/docs/CodingStandards.html#assert-liberally), but some people recommend `assert(0)`: https://discourse.llvm.org/t/llvm-unreachable-is-widely-misused/60587/11?page=2.

I discovered this issue while compiling in the downstream repository torch-mlir using CMake, with the build option set to `RelWithDebInfo`.

The exported build command was for 
```
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ -DGTEST_HAS_RTTI=0 -DTORCH_MLIR_ENABLE_REFBACKEND -DTORCH_MLIR_ENABLE_STABLEHLO -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -I/code/torch-mlir/externals/llvm-project/llvm/cmake-build-relwithdebinfo/tools/torch-mlir/stablehlo/stablehlo/conversions/linalg/transforms -I/code/torch-mlir/externals/stablehlo/stablehlo/conversions/linalg/transforms -I/code/torch-mlir/externals/llvm-project/llvm/cmake-build-relwithdebinfo/include -I/code/torch-mlir/externals/llvm-project/llvm/include -I/code/torch-mlir/externals/stablehlo -I/code/torch-mlir/include -I/code/torch-mlir/externals/llvm-project/llvm/cmake-build-relwithdebinfo/tools/torch-mlir/include -I/code/torch-mlir/externals/llvm-project/llvm/cmake-build-relwithdebinfo/tools/torch-mlir/stablehlo -I/opt/homebrew/include -isystem /code/torch-mlir/externals/llvm-project/llvm/../mlir/include -isystem /code/torch-mlir/externals/llvm-project/llvm/cmake-build-relwithdebinfo/tools/mlir/include  -fPIC -fvisibility-inlines-hidden -Werror=date-time -Werror=unguarded-availability-new -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wmissing-field-initializers -pedantic -Wno-long-long -Wc++98-compat-extra-semi -Wimplicit-fallthrough -Wcovered-switch-default -Wno-noexcept-type -Wnon-virtual-dtor -Wdelete-non-virtual-dtor -Wsuggest-override -Wstring-conversion -Wmisleading-indentation -Wctad-maybe-unsupported -fdiagnostics-color -fPIC -fvisibility-inlines-hidden -Werror=date-time -Werror=unguarded-availability-new -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wmissing-field-initializers -pedantic -Wno-long-long -Wc++98-compat-extra-semi -Wimplicit-fallthrough -Wcovered-switch-default -Wno-noexcept-type -Wnon-virtual-dtor -Wdelete-non-virtual-dtor -Wsuggest-override -Wstring-conversion -Wmisleading-indentation -Wctad-maybe-unsupported -fdiagnostics-color -O2 -g -DNDEBUG -std=c++17 -arch arm64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.2.sdk -mmacosx-version-min=14.1 -fcolor-diagnostics   -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS  -fno-exceptions -funwind-tables -fno-rtti -o tools/torch-mlir/stablehlo/stablehlo/conversions/linalg/transforms/CMakeFiles/obj.StablehloLinalgTransforms.dir/LegalizeToLinalgUtils.cpp.o -c /code/torch-mlir/externals/stablehlo/stablehlo/conversions/linalg/transforms/LegalizeToLinalgUtils.cpp
```
And when I modified the build command to `-E`, I could confirm that their asserts compiled out.

GleasonK (approved):

GleasonK:
IIUC the issue is that we were placing a function call with side-effects that was required logic for function correctness inside an assert which was then optimized away.

```
assert(succeeded(shapeSource.reifyReturnTypeShapes(b, operands, reifiedShapes)) &&
       "could not reify");
```

Meaning the following would have been just as effective of an approach as the currently implemented solution:

```
auto res = shapeSource.reifyReturnTypeShapes(b, operands, reifiedShapes);
assert(succeeded(res) && "could not reify");
```

That said, I think our codebase prefers fatal errors to asserts / unreachable, so final code lgtm.

PR #2158, opened 2024-04-03
Add a note at type_inference.md for adding custom verifier if return type inference is not feasible

abhigunj:
if return type of the op is inferable, it can be created using `op::build` which does not take result type as input.  For such ops it is recommended to move verification logic to shape functions to avoid missing on any verification during op creation. If return type inference is not feasible use custom verifier.


docs/type_inference.md
```diff
@@ -67,7 +67,10 @@ a few problems:
 The solution is as follows:
 
 1. **For most ops without regions** (like `PadOp`): Put all the verification
-   code into the shape functions, and discard verifiers totally.
+   code into the shape functions, and discard verifiers totally. This is
+   applicable if return type inference is feasible. If return type inference
+   is not feasible (like `AddOp`) have a custom verifier to hold
```
sdasgup3, line 72: It is better to use examples where shape inference is not possible even with non-quantized types like  `reshape` and `broadcast_in_dims`. AddOp's type inference in-feasibility is seen oonly with quantized type and that might be difficult to follow.

abhigunj: ack, thank you

ghpvnist: May also be worth mentioning that typically inferable ops, like AddOp, may still need additional verifications when quantization is involved.

ghpvnist (approved):
LGTM with a followup comment

docs/type_inference.md
```diff
@@ -66,8 +66,13 @@ a few problems:
...
 
-1. **For most ops without regions** (like `PadOp`): Put all the verification
-   code into the shape functions, and discard verifiers totally.
+1. **For most ops without regions** (like `PadOp`): Try to put all the verification
+   code into the shape functions, and discard verifiers totally. In cases where
+   this isn't possible due to the inability to infer return types
+   (such as with`ReshapeOp` or `BroadcastInDimOp`), create a verifier to
+   contain the necessary verification logic. Typically inferable ops,
+   like `AddOp`, may still need a verifier to perform additional verifications
+   when quantization is involved.
```
GleasonK, line 75: "because they are verifying constraints of a provided return type, which is not accessible in the type/shape inference methods."

GleasonK (approved):

PR #2160, opened 2024-04-04
Implement ConditionallySpeculatable for TriangularSolve

mlevesquedion:
If `a` is constant, we could check the diagonal to confirm that it
is unit, but this may be costly and should go in the op's verifier.

### DIFF
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -2042,8 +2042,7 @@ mlir::Speculation::Speculatability TriangularSolveOp::getSpeculatability() {
   // Diagonal could be non-unit at runtime, leading to undefined behavior.
   // If `a` is constant, we could check the diagonal to confirm that it
   // is unit, but this may be costly and should go in the op's verifier.
-  if (getUnitDiagonal())
-    return mlir::Speculation::NotSpeculatable;
+  if (getUnitDiagonal()) return mlir::Speculation::NotSpeculatable;
 
   // If the inputs are statically shaped, they will be fully verified
   // statically. If the inputs are dynamic, then mismatches could occur at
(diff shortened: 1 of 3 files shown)


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2023,6 +2023,22 @@ LogicalResult TriangularSolveOp::inferReturnTypeComponents(
                                      inferredReturnShapes);
 }
 
+mlir::Speculation::Speculatability TriangularSolveOp::getSpeculatability() {
```
sdasgup3, line 2026: >> Diagonal could be non-unit at runtime, leading to undefined behavior.
This seems odd to me. IOW, it would mean that the flag always has to be true for the op to be meaningful.
What the spec says is: if the flag is true, then the implementation has to assume that the diagonal elements are 1. Failing to assume that leads to undefined behavior. 

IIUC the following check says is meant for: if the flag is true, and if  implementation  fails to assume than the diagonal is 1, then there could be undefined behavior.
>>   `if (getUnitDiagonal()) return mlir::Speculation::NotSpeculatable;`

IMO, the above check is fine.

mlevesquedion: We're on the same page. The comment only applies when `getUnitDiagonal` is `true`. I rewrote the comment to make that more obvious, thanks!

sdasgup3 (approved):

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2038,6 +2038,22 @@ LogicalResult TriangularSolveOp::inferReturnTypeComponents(
...
+  if (getUnitDiagonal()) return mlir::Speculation::NotSpeculatable;
+
+  // If the inputs are statically shaped, they will be fully verified
+  // statically. If the inputs are dynamic, then mismatches could occur at
+  // runtime.
+  auto lhsType = cast<RankedTensorType>(getOperand(0).getType());
+  auto rhsType = cast<RankedTensorType>(getOperand(1).getType());
+  if (lhsType.hasStaticShape() && rhsType.hasStaticShape())
+    return mlir::Speculation::Speculatable;
+  return mlir::Speculation::NotSpeculatable;
```
abhigunj, line 2054:
```suggestion
  
  return mlir::Speculation::NotSpeculatable;
```

abhigunj (approved):

stablehlo/dialect/StablehloOps.cpp
mlevesquedion: Done, thanks!

PR #2163, opened 2024-04-04
Mention trace in spec

mlevesquedion:
The op is slated for removal, but unlike other ops in a similar position it currently isn't mentioned in the spec.


ghpvnist (approved):
Good find!

PR #2168, opened 2024-04-06
Implement ConditionallySpeculatable for collective ops

mlevesquedion:
I'm actually not sure if speculation make sense for these ops. I don't see any indication in the spec that they could have UB except in the cases added here, so it's likely fine. Also, I'm not sure if these ops can have memory effects or not. They definitely involve interacting with some kind of global state, so they have side effects. The ops weren't marked "Pure" before this change, so there is no difference on that front being introduced with this change.

Also I slightly updated the logic in TestUtils to delete the op when the speculation check succeeds. Indeed before we were relying on DCE, but these ops don't state that they don't have side effects so DCE won't remove them.

### DIFF
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -130,6 +130,19 @@ LogicalResult ReduceScatterOp::verify() {
       channelId, getUseGlobalDeviceIds(), getComputation(), getResult());
 }
 
+mlir::Speculation::Speculatability ReduceScatterOp::getSpeculatability() {
+    auto inputType = cast<RankedTensorType>(getOperand().getType());
+    auto resultType = cast<RankedTensorType>(getResult().getType());
+    auto scatterDim = getScatterDimension();
+    if (!resultType.isDynamicDim(scatterDim)) return mlir::Speculation::NotSpeculatable;
+    for (size_t i : llvm::seq(resultType.getRank())) {
+      if (i == scatterDim) continue;
+      if (!resultType.isDynamicDim(i) && inputType.isDynamicDim(i))
+        return mlir::Speculation::NotSpeculatable;
+    }
+    return mlir::Speculation::Speculatable;
+}
+
 //===----------------------------------------------------------------------===//
 // CompatibleOperandsAndResultType
 //===----------------------------------------------------------------------===//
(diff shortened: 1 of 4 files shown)


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -130,6 +130,20 @@ LogicalResult ReduceScatterOp::verify() {
       channelId, getUseGlobalDeviceIds(), getComputation(), getResult());
 }
 
+mlir::Speculation::Speculatability ReduceScatterOp::getSpeculatability() {
+  auto inputType = cast<RankedTensorType>(getOperand().getType());
+  auto resultType = cast<RankedTensorType>(getResult().getType());
```
GleasonK, line 135: Do we still need the `cast`?

```
RankedTensorType inputType = getOperand().getType();
```

mlevesquedion: Good point, we don't! Removed, here and below.

mlevesquedion: There's some others in this file, I'll clean that up soon.

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -130,6 +130,20 @@ LogicalResult ReduceScatterOp::verify() {
       channelId, getUseGlobalDeviceIds(), getComputation(), getResult());
 }
 
+mlir::Speculation::Speculatability ReduceScatterOp::getSpeculatability() {
+  auto inputType = getOperand().getType();
+  auto resultType = getResult().getType();
+  auto scatterDim = getScatterDimension();
+  if (!resultType.isDynamicDim(scatterDim))
```
sdasgup3, line 137: I did not follow this. Why a static scatter_dim  make the op non-speculatable?

mlevesquedion: It is because of `(C8)`:

```
(C8) shape(result) = shape(operand) except:
dim(result, scatter_dimension) = dim(operand, scatter_dimension) / dim(process_groups, 1).
```

We don't know the number of processes per group until runtime. If the `scatter_dim` is dynamic, it doesn't matter, but if it's static, then we may end up with a different number at runtime, leading to UB.

Let me add some comments to this and the others.

sdasgup3: In my opinion, these ops should not be specilatable even if the scatter_dimension is dynamic. A runtime value of scatter_dimension and process_groups can also violate the C8?

mlevesquedion: But if the scatter_dimension is dynamic in the result, then the equality cannot be false.

We have `? = dim(operand, scatter_dimension) / dim(process_groups, 1)`. This is always true, in fact the `?` can be inferred from the RHS.

sdasgup3: `dim(result, scatter_dimension) = dim(operand, scatter_dimension) / dim(process_groups, 1)`
Given a constant `scatter_dimension`, either 
1. dim(result, scatter_dimension)  is static or decided at runtime if dynamic
2. dim(operand, scatter_dimension) is static or decided at runtime if dynamic
3. dim(process_groups, 1) decided at runtime

say 1 is dynamic and 2 is static, the op is still non-speculatable, right?

mlevesquedion: Are you referring to the fact that there can be a division by zero if `dim(process_groups, 1) == 0`? Can a process group be empty? If the `scatter_dimension` in the result is dynamic, I would say this is still valid.

Otherwise, I'm not sure I follow. If `dim(result, scatter_dimension) = ?`, then it does not matter what the corresponding dimension in the operands is, nor does it matter what the number of processes per group is.

sdasgup3: Based on online chat, I realized that a dynamic`dim(result, scatter_dimension)` can be deduced at runtime and thereby if does no matter if  `dim(operand, scatter_dimension)` is static or not. 

Thans for the clarification. We can close this thread now.

sdasgup3 (approved):

stablehlo/dialect/StablehloOps.cpp
mlevesquedion: Cool, thanks! :)

PR #2174, opened 2024-04-08
Fix typo in CaseOp description: branches are regions, not functions

mlevesquedion:


mlevesquedion:
After taking a closer look at the spec and at some other similar operations, it seems this may be intentional? Specifically, `function` here is a StableHLO concept, an abstraction, whereas `region` is an MLIR concept, an implementation detail?

sdasgup3:
In the spec you will find that the function is used wherever the implementation chooses a `region`. This is mainly because the spec cares only about the concept. For example, a reduction operation is associated with a reduction function. The spec deliberately avoided region because that is more of a implementation detail. 

Also the td file followins https://github.com/openxla/stablehlo/blob/main/docs/spec_checklist.md#stablehlo-specification-checklist point 6 to include the first line of the semantics in the description.

mlevesquedion:
> In the spec you will find that the function is used wherever the implementation chooses a `region`. This is mainly because the spec cares only about the concept. For example, a reduction operation is associated with a reduction function. The spec deliberately avoided region because that is more of a implementation detail.
> 
> Also the td file followins https://github.com/openxla/stablehlo/blob/main/docs/spec_checklist.md#stablehlo-specification-checklist point 6 to include the first line of the semantics in the description.

Okay, SG then! I think in the context of the spec it makes sense to use `function`, however when looking at the op's ODS definition seeing `function` instead of e.g. `region` is kinda confusing. Maybe we should revisit "Includes the first sentence of the spec.", I'm not sure we should always duplicate the spec's first sentence exactly.

PR #2175, opened 2024-04-08
Mark recursively speculatable ops as such

mlevesquedion:
This includes the regioned ops case, if and while.


sdasgup3:
Do we have plans to cover other ops with regions as well. For example, a redice op is speculatable if the all the ops in the reduction regions are speculatable as such.

sdasgup3 (approved):

mlevesquedion:
> Do we have plans to cover other ops with regions as well. For example, a redice op is speculatable if the all the ops in the reduction regions are speculatable as such.

That's next on my list, yeah! For `map` and `reduce` it's actually a bit more complicated than that because they have operands whose shapes could disagree at runtime. I plan to get to those soon.

Issue #2065, opened 2024-03-04
Remove unranked type verifications in type inference and elsewhere


ghpvnist:
There's ongoing work (e.g. #2049) on disallowing unranked types for dynamism, and while working on shape refinement, I noticed that we still have checks to assert ranked types.

There are several instances of shape inference code that checks that the tensor types are ranked types (e.g. [here](https://github.com/openxla/stablehlo/blob/c2fccec9f02daa2d709001ed5f37924d2526676f/stablehlo/dialect/TypeInference.cpp#L1893-L1898)). Since we now disallow unranked types in the ODS, we can probably remove such verifications (likely deadcode).

mlevesquedion:
Thanks! This is on my list of tasks for implementing the RFC: https://github.com/openxla/stablehlo/issues/1991#issue-2119623349. I linked to this new issue there.

mlevesquedion:
With https://github.com/openxla/stablehlo/pull/2119, https://github.com/openxla/stablehlo/issues/2138, https://github.com/openxla/stablehlo/issues/2139, https://github.com/openxla/stablehlo/pull/2183 and https://github.com/openxla/stablehlo/pull/2217, I will consider this done.

PR #2183, opened 2024-04-09
Use cast instead of dyn_cast where possible

mlevesquedion:
For type inference, if the infer/verify logic were a method on the op itself, we could remove the cast altogether, but with the current setup we can at least change dyn_cast into cast to remove any ambiguity and clean up redundant checks.

In some cases it's not possible to remove the dyn_casts, for various reasons:
- CHLO ops still allow RankedTensorType
- Some dyn_casts are used on things that may not be tensors

https://github.com/openxla/stablehlo/issues/2065


mlevesquedion:
Converting to draft, I just realized a lot of casts can be removed altogether because they occur in methods on the op, and various `getType` methods will return `RankedTensorType` directly.

mlevesquedion:
On closer look, it seems there's only one case where the cast can be removed completely.

sdasgup3 (approved):

mlevesquedion:
I re-reviewed everything and made some extra internal verifications and everything looks good.

PR #2187, opened 2024-04-09
Implement speculatability for Clamp, Compare, Complex and Select

mlevesquedion:


stablehlo/dialect/StablehloOps.td
```diff
@@ -1567,7 +1568,9 @@ def StableHLO_TupleOp : StableHLO_Op<"tuple", [Pure,
   }];
 }
 
-def StableHLO_CompareOp: StableHLO_Op<"compare", [Pure, Elementwise,
+def StableHLO_CompareOp: StableHLO_Op<"compare",
+    [HLO_SpeculatableIfAllInputsStatic, NoMemoryEffect,
+    Elementwise,
```
abhigunj, line 1573: for my clarification...  `Elementwise` Trait is unrelated to speculatability?

abhigunj (approved):

stablehlo/dialect/StablehloOps.td
mlevesquedion: It's sorta related: https://mlir.llvm.org/doxygen/structmlir_1_1OpTrait_1_1Elementwise.html

The doc says "The shape may be dynamic in which case the op's behaviour is undefined for non-matching shapes.", but it doesn't provide an implementation of `getSpeculatability`.

sdasgup3 (approved):

stablehlo/dialect/StablehloOps.td
abhigunj: got it, thank you.

PR #2193, opened 2024-04-10
Implement ConditionallySpeculatable for Map, Reduce and ReduceWindow

mlevesquedion:
All of these ops have `same(shape(inputs...))` as a constraint,
but they also have regions, so if all their inputs are static, the
op is recursively speculatable.

(Note that `scatter` also has this constraint, but it can have UB
in other circumstances so it will be handled in a follow-up change.)

### DIFF
--- a/stablehlo/dialect/Base.h
+++ b/stablehlo/dialect/Base.h
@@ -435,6 +435,20 @@ struct SpeculatableIfAllInputsStaticImplTrait
   }
 };
 
+template <typename ConcreteType>
+struct RecursivelySpeculatableIfAllInputsStaticImplTrait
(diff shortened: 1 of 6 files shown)


mlevesquedion:
I realized `reduce` has the same logic, converting to draft while I include reduce in this change (and I'll also try to see if other ops have the same logic).

mlevesquedion:
This is now ready for review. Ended up batching Reduce and ReduceWindow into this, and I have a dependent PR for Scatter: https://github.com/openxla/stablehlo/pull/2195

sdasgup3 (approved):

PR #2195, opened 2024-04-10
Implement ConditionallySpeculatable for Scatter

mlevesquedion:
I am mostly interested in shape mismatches so I left a TODO
to check if the indices are unique. In the meantime, we will do
the conservative thing, which is to assume that the op is not
speculatable (the indices may not be unique).

NOTE: This PR builds on top of https://github.com/openxla/stablehlo/pull/2193. See the 2nd commit for changes specific to this PR.

### DIFF
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -739,6 +739,20 @@ LogicalResult GatherOp::inferReturnTypeComponents(
       adaptor.getSliceSizes(), inferredReturnShapes);
 }
 
+mlir::Speculation::Speculatability GatherOp::getSpeculatability() {
+  // When indices_are_sorted is true, if the start_indices are not sorted, the
+  // behavior is undefined.
+  // A possible improvement would be to check if the start_indices are constant
+  // and if they are sorted, do not return NotSpeculatable. However, such a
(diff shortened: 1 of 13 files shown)


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2210,6 +2222,19 @@ LogicalResult ScatterOp::verify() {
       getScatterDimensionNumbers().getIndexVectorDim(), getUpdateComputation());
 }
 
+mlir::Speculation::Speculatability ScatterOp::getSpeculatability() {
+  // When unique_indices is true, if the scatter_indices are not unique, the
+  // behavior is undefined.
+  // TODO: check if the scatter_indices are constant and if they are unique,
+  // do not return NotSpeculatable.
+  if (getUniqueIndices()) return mlir::Speculation::NotSpeculatable;
```
sdasgup3, line 2230: Also addresses the UB for `indices_are_sorted`?

sdasgup3 (approved):

stablehlo/dialect/StablehloOps.cpp
mlevesquedion: Great catch, thanks I added the missing check for that.

PR #2197, opened 2024-04-10
Augment bazelrc to support asan & ubsan

fzakaria:
A prior failure in the google3 internal codebase showed that we were not executing the sanitizers like we thought we were.

This change augments bazelrc to do asan and ubsan via config. It leverages the default feature set offered by bazelrc.

I tested the commit priot to 474caa8370dccc753e4693a2ff4221864b227597 to validate that the bug was exhibited.

I've added the sanitizers to the default Bazel build which also builds debug builds.
This may come at an increase cost. If undesirable we can split the job out.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -29,12 +29,9 @@ build --repo_env=CC=clang
 
 # turn on fission https://gcc.gnu.org/wiki/DebugFission
 # this separates the debug information and can improve link times
-build --fission=yes
+build --fission=dbg
 
-build:dbg --compilation_mode=dbg
-build:opt --compilation_mode=opt
-
-build:san-common --config=dbg --strip=never --copt=-O0 --copt=-fno-omit-frame-pointer
+build:san-common --strip=never --copt=-fno-omit-frame-pointer
 
 # Some of this is from "Can I run AddressSanitizer with more aggressive diagnostics enabled?"
 # on https://github.com/google/sanitizers/wiki/AddressSanitizer#faq and some is from
--- a/.bazelrc
+++ b/.bazelrc
@@ -63,8 +63,8 @@ build:ubsan --action_env=UBSAN_OPTIONS=halt_on_error=1:print_stacktrace=1
 # Do not build runfile trees by default. If an execution strategy relies on runfile
 # symlink tree, the tree is created on-demand. See: https://github.com/bazelbuild/bazel/issues/6627
 # and https://github.com/bazelbuild/bazel/commit/03246077f948f2790a83520e7dccc2625650e6df
-#build --nobuild_runfile_links
-#test --nobuild_runfile_links
+build --nobuild_runfile_links
+test --nobuild_runfile_links
 
 # https://bazel.build/reference/command-line-reference#flag--legacy_external_runfiles
 build --nolegacy_external_runfiles
--- a/.bazelrc
+++ b/.bazelrc
@@ -14,6 +14,11 @@
 
 build --cxxopt=-std=c++17
 build --host_cxxopt=-std=c++17
+# llvm builds typically without rtti
+# disable it to be consistent with the llvm build
+# this also helps later on when using sanitizers
+build --cxxopt=-fno-rtti
+build --host_cxxopt=-fno-rtti
 # support layering_check similar to Google internal
 # this only works for Clang toolchain AFAIK
 # https://github.com/bazelbuild/bazel/pull/11440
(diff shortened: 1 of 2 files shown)


.bazelrc
```diff
@@ -14,6 +14,11 @@
 
 build --cxxopt=-std=c++17
 build --host_cxxopt=-std=c++17
+# llvm builds typically without rtti
```
mlevesquedion, line 17: Consider linking to a reference for this, would be helpful for context and in case we need to adjust this in the future.

.bazelrc
```diff
@@ -22,6 +27,38 @@ build --features=layering_check
...
+
+build:san-common --config=dbg --strip=never --copt=-O0 --copt=-fno-omit-frame-pointer
+
+# Some of this is from "Can I run AddressSanitizer with more aggressive diagnostics enabled?"
+# on https://github.com/google/sanitizers/wiki/AddressSanitizer#faq and some is from
+# https://chromium.googlesource.com/external/github.com/grpc/grpc/+/4e9206f48c91e17f43856b016b12f59dd5118293/tools/bazel.rc
+build:asan --config=san-common
+build:asan --features=asan
+build:asan --copt=-fsanitize-address-use-after-scope
+# We explicitly disable strict_init_order because LLVM is "hostile to init order"
```
mlevesquedion, line 45: Consider linking to the LLVM asan config for reference.

.bazelrc
```diff
@@ -22,6 +27,38 @@ build --features=layering_check
 # TODO(fzakaria): Make this a toolchain or hermetic somehow
 build --repo_env=CC=clang
 
+# turn on fission https://gcc.gnu.org/wiki/DebugFission
+# this separates the debug information and can improve link times
+build --fission=yes
+
+build:dbg --compilation_mode=dbg
+build:opt --compilation_mode=opt
```
mlevesquedion, line 35: These lines look like they belong in a separate change? Not a huge deal, but consider splitting this out into its own PR and/or calling this out in the commit message/PR description.

(`--fission` seems like general QOL, `--opt` seems unused, and `--dbg` is used, but I only see it used once below, so it could be inlined.)

build_tools/github_actions/ci_build_bazel.sh
```diff
@@ -22,5 +22,5 @@ if [[ $# -ne 0 ]] ; then
 fi
 
 # Build and Test StableHLO
-bazel build --lockfile_mode=error //...
-bazel test //...
+bazel build --lockfile_mode=error //... --config=asan --config=ubsan
```
mlevesquedion, line 25: Consider creating a separate config that wraps these two, e.g. `build:san --config=asan,ubsan` and then you can do `--config=san` here.

mlevesquedion (commented):
Very cool! A few high-level comments/questions:
- Consider adding a link to the documentation for sanitizers somewhere.
- Do we want `msan` as well?
- I thought we ran sanitizers [in our CMake build](https://github.com/openxla/stablehlo/blob/714d9acacd96760f4d8a9fe3898bee4f204cd419/.github/workflows/buildAndTestCMake.yml#L84)? Any insight on why that didn't catch the issues? 
- From the CI on this PR, it does look like this makes the bazel build a lot slower. Also, it seems to be producing some errors, is that expected?

fzakaria:
> I thought we ran sanitizers [in our CMake build](https://github.com/openxla/stablehlo/blob/714d9acacd96760f4d8a9fe3898bee4f204cd419/.github/workflows/buildAndTestCMake.yml#L84)? Any insight on why that didn't catch the issues?
From the CI on this PR, it does look like this makes the bazel build a lot slower. Also, it seems to be producing some errors, is that expected?

I am looking into having it run non debug.

.bazelrc
fzakaria: What do you mean ?
I'm trying to hint that the google3 config has whitelist to avoid LLVM for this check.

.bazelrc
fzakaria: I removed them for now and turned on fission only for dbg.
This should bring the code size back down to normal?

fzakaria:
If this builds and passes, I'll augment the README with a little section on how to build with Bazel & the sanitizers
(+ running with dbg)

TODO remains to look at our CMake :S

GleasonK (approved):
This seems like a generally good feature to have and shouldn't change much about the typical build process. If the fission option causes any issues we may want to hide it behind the san-common configuration

PR #2198, opened 2024-04-10
Implement ConditionallySpeculatable for Gather

mlevesquedion:


sdasgup3 (approved):

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -741,6 +741,19 @@ LogicalResult GatherOp::inferReturnTypeComponents(
       adaptor.getSliceSizes(), inferredReturnShapes);
 }
 
+mlir::Speculation::Speculatability GatherOp::getSpeculatability() {
+  // When indices_are_sorted is true, if the start_indices are not sorted, the
+  // behavior is undefined.
+  // TODO: check if the start_indices are constant and if they are sorted,
```
abhigunj, line 747: do we have a tracker for this todo?

abhigunj (approved):

stablehlo/dialect/StablehloOps.cpp
mlevesquedion: I'm not planning to do this work, and IMO there is little ROI here so I don't think we should necessarily track it either. Reworded to not use "TODO", thanks!

Issue #2074, opened 2024-03-06
Add verifier for quantized element type


sdasgup3:
### What happened?

Not all the constraints for quantized types  https://github.com/openxla/stablehlo/blob/main/docs/spec.md#types are captures in [here](https://github.com/llvm/llvm-project/blob/8b5b294ec2cf876bc5eb5bd5fcb56ef487e36d60/mlir/lib/Dialect/Quant/IR/QuantTypes.cpp#L32) or [here](https://github.com/llvm/llvm-project/blob/8b5b294ec2cf876bc5eb5bd5fcb56ef487e36d60/mlir/lib/Dialect/Quant/IR/QuantTypes.cpp#L324). 

For example, the ones not captures are 

```
For per-axis quantization:
(C12) quantization_dimension < rank(self).
(C13) dim(self, quantization_dimension) = size(scales).
```
and 

```
(C1) num_bits(storage_type) < num_bits(expressed_type).
(C5) type(scales...) = expressed_type.
(C7) is_finite(scales...).
(C8) storage_min <= zero_points <= storage_max.
(C9) type(zero_points...) = storage_type.
(C10) size(scales) = size(zero_points).
(C12) 0 <= quantization_dimension.
```

The goal of the issue to to implement the verification checks.

### Steps to reproduce your issue

Closes #2173.

PR #2203, opened 2024-04-11
Verify StableHLO Quantized Types

sdasgup3:
fixes https://github.com/openxla/stablehlo/issues/2074

It seems there are some use-cases, like quantized reduce wanting to perform the body computation in a higher bitwidth, which avoids constraining and hence is not part of the verification checks.

```
(C1) num_bits(storage_type) < num_bits(expressed_type).
```

Added a ticket to address it separately https://github.com/openxla/stablehlo/issues/2205

### DIFF
--- a/stablehlo/dialect/Base.cpp
+++ b/stablehlo/dialect/Base.cpp
@@ -688,11 +688,11 @@ bool isValidQuantizedDimension(Type type) {
 
   // quantized_type_c12, quantized_type_c13, quantized_type_c14
   int64_t quantDim = quantizedPerAxisElementType.getQuantizedDimension();
+  int64_t numScales =
+      static_cast<int64_t>(quantizedPerAxisElementType.getScales().size());
   return quantDim >= 0 && quantDim < rankedType.getRank() &&
          (!rankedType.isDynamicDim(quantDim) &&
-          static_cast<int64_t>(
-              quantizedPerAxisElementType.getScales().size()) ==
-              rankedType.getDimSize(quantDim));
+          numScales == rankedType.getDimSize(quantDim));
 }
 
 }  // namespace hlo
--- a/stablehlo/dialect/Base.cpp
+++ b/stablehlo/dialect/Base.cpp
@@ -625,5 +625,75 @@ mlir::Speculation::Speculatability getShapedSpeculatability(
              : mlir::Speculation::NotSpeculatable;
 }
 
+bool isValidStablehloQuantizedElementType(Type elementType) {
+  auto quantizedElementType = dyn_cast<mlir::quant::QuantizedType>(elementType);
+  if (!quantizedElementType) return false;
+
+  int64_t storageTypeMin = quantizedElementType.getStorageTypeMin();
+  int64_t storageTypeMax = quantizedElementType.getStorageTypeMax();
+
+  SmallVector<int64_t> zeroPoints;
+  SmallVector<double> scales;
+  if (auto quantizedPerTensorElementType =
+          dyn_cast<mlir::quant::UniformQuantizedType>(elementType)) {
+    zeroPoints.push_back(quantizedPerTensorElementType.getZeroPoint());
+    scales.push_back(quantizedPerTensorElementType.getScale());
+  } else {
+    auto quantizedPerAxisElementType =
+        cast<mlir::quant::UniformQuantizedPerAxisType>(elementType);
+    zeroPoints.insert(zeroPoints.begin(),
+                      quantizedPerAxisElementType.getZeroPoints().begin(),
+                      quantizedPerAxisElementType.getZeroPoints().end());
+    scales.insert(scales.begin(),
+                  quantizedPerAxisElementType.getScales().begin(),
+                  quantizedPerAxisElementType.getScales().end());
+  }
+
+  // quantized_type_c5
+  auto maxPosFiniteNum =
+      APFloat::getLargest(quantizedElementType.getExpressedType()
+                              .cast<FloatType>()
+                              .getFloatSemantics())
+          .convertToDouble();
+  auto minPosFiniteNum =
+      APFloat::getSmallest(quantizedElementType.getExpressedType()
+                               .cast<FloatType>()
+                               .getFloatSemantics())
+          .convertToDouble();
+  if (llvm::any_of(scales, [&](double scale) {
+        return scale < minPosFiniteNum || scale > maxPosFiniteNum;
+      })) {
+    return false;
+  }
+
+  // quantized_type_c8, quantized_type_c9
+  if (llvm::any_of(zeroPoints, [&](int64_t zeroPoint) {
+        return storageTypeMin > zeroPoint || zeroPoint > storageTypeMax;
+      })) {
+    return false;
+  }
+
+  return true;
+}
+
+bool isValidQuantizedDimension(Type type) {
+  auto rankedType = dyn_cast<RankedTensorType>(type);
+  if (!rankedType) return true;
+
+  auto quantizedPerAxisElementType =
+      dyn_cast<mlir::quant::UniformQuantizedPerAxisType>(
+          rankedType.getElementType());
+
+  if (!quantizedPerAxisElementType) return true;
+
+  // quantized_type_c12, quantized_type_c13, quantized_type_c14
+  int64_t quantDim = quantizedPerAxisElementType.getQuantizedDimension();
+  return quantDim >= 0 && quantDim < rankedType.getRank() &&
+         (!rankedType.isDynamicDim(quantDim) &&
+          static_cast<int64_t>(
+              quantizedPerAxisElementType.getScales().size()) ==
+              rankedType.getDimSize(quantDim));
+}
+
 }  // namespace hlo
 }  // namespace mlir
--- a/stablehlo/dialect/Base.td
+++ b/stablehlo/dialect/Base.td
@@ -52,6 +52,9 @@ def HLO_Complex : Complex<AnyTypeOf<[F32, F64]>>;
 // Quantized element type definitions.
 //===----------------------------------------------------------------------===//
 
+def IsValidStablehloQuantizedElementType : CPred<"mlir::hlo::isValidStablehloQuantizedElementType($_self)">;
+def IsValidQuantizedDimension : CPred<"mlir::hlo::isValidQuantizedDimension($_self)">;
+
 // TODO(b/230381284): Upstream width-specific uniform quantized element types.
 class StableHLO_UniformQuantizedSignedInt<int width>
   : Type<And<[CPred<"isa<mlir::quant::UniformQuantizedType>($_self)">,
@@ -59,7 +62,7 @@ class StableHLO_UniformQuantizedSignedInt<int width>
                  ".getStorageTypeIntegralWidth() == " # width>,
            CPred<"cast<mlir::IntegerType>(" #
                    "cast<mlir::quant::UniformQuantizedType>($_self).getStorageType()" #
-                 ").isSignless()">]>,
+                 ").isSignless()">, IsValidStablehloQuantizedElementType]>,
     "QI" # width # " type"> {
   string name = "UniformQuantizedSignedInt";
   int bitwidth = width;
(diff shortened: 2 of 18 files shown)


stablehlo/tests/ops_stablehlo_quantized.mlir
```diff
@@ -967,3 +968,59 @@ func.func @reshape_c3_mismatch_product_before(%arg0: tensor<1x2x3x4x5x!quant.uni
   %reshape = "stablehlo.reshape" (%arg0) : (tensor<1x2x3x4x5x!quant.uniform<i8:f32:0, {1.0:17}>>) -> tensor<2x1x3x20x!quant.uniform<i8:f32:1, {1.0:17}>>
   func.return
 }
+
+// -----
+
+func.func @quantized_element_type_c8(%arg0: tensor<1x2x!quant.uniform<i8<-128:127>:f32, 1.0:300>>) {
+  // expected-error@+1 {{operand #0 must be ranked tensor of f8E4M3B11FNUZ type or f8E4M3FN type or f8E4M3FNUZ type or f8E5M2 type or f8E5M2FNUZ type or 16-bit float or 32-bit float or 64-bit float or bfloat16 type or pred (AKA boolean or 1-bit integer) or 4/8/16/32/64-bit signless integer or 4/8/16/32/64-bit unsigned integer or complex type with 32-bit float or 64-bit float elements or 4/8/16/32-bit uniform quantized signed integer or 4/8/16/32-bit uniform quantized unsigned integer or 4/8/16/32-bit uniform quantized per axis signed integer or 4/8/16/32-bit uniform quantized per axis unsigned integer values, but got 'tensor<1x2x!quant.uniform<i8:f32, 1.000000e+00:300>>'}}
```
abhigunj, line 975:
```suggestion
  // expected-error-re@+1 {{operand #0 must be ranked tensor of {{.*}} 4/8/16/32-bit uniform quantized signed integer or 4/8/16/32-bit uniform quantized unsigned integer or 4/8/16/32-bit uniform quantized per axis signed integer or 4/8/16/32-bit uniform quantized per axis unsigned integer values, but got 'tensor<1x2x!quant.uniform<i8:f32, 1.000000e+00:300>>'}}
```

We can use regex match here to make tests less brittle, since these tests are validating new api.

abhigunj (approved):
Thanks for correcting the tests.

stablehlo/tests/ops_stablehlo_quantized.mlir
sdasgup3: done.

stablehlo/tests/ops_stablehlo_quantized.mlir
```diff
@@ -1051,3 +1052,59 @@ func.func @convolution_c34(%arg0: tensor<1x4x4x1xf32>, %arg1: tensor<3x3x1x1x!qu
...
+func.func @quantized_element_type_c5(%arg0: tensor<1x2x!quant.uniform<i4:f16, 10.550400e+04>>) {
+  // expected-error-re@+1 {{operand #0 must be ranked tensor of {{.*}} 4/8/16/32-bit uniform quantized signed integer or 4/8/16/32-bit uniform quantized unsigned integer or 4/8/16/32-bit uniform quantized per axis signed integer or 4/8/16/32-bit uniform quantized per axis unsigned integer values, but got 'tensor<1x2x!quant.uniform<i4:f16, 1.055040e+05>>'}}
+   %0 = stablehlo.add %arg0,  %arg0 : tensor<1x2x!quant.uniform<i4:f16, 10.550400e+04>>
+   func.return
+}
+
+// -----
+
+func.func @quantized_element_type_c5(%arg0: tensor<1x2x!quant.uniform<i4:f16, 4.960464e-08>>) {
+  // expected-error-re@+1 {{perand #0 must be ranked tensor of {{.*}} 4/8/16/32-bit uniform quantized signed integer or 4/8/16/32-bit uniform quantized unsigned integer or 4/8/16/32-bit uniform quantized per axis signed integer or 4/8/16/32-bit uniform quantized per axis unsigned integer values, but got 'tensor<1x2x!quant.uniform<i4:f16, 4.9604639999999998E-8>>'}}
```
abhigunj, line 1083: nit : operand

abhigunj (approved):

stablehlo/tests/CheckOps.cpp
```diff
@@ -23,6 +23,7 @@ limitations under the License.
 #include "llvm/Support/Path.h"
 #include "mlir/IR/BuiltinTypes.h"
 #include "mlir/Support/DebugStringHelper.h"
+#include "stablehlo/dialect/Base.h"
```
GleasonK, line 26: Is this needed? Same question for the BUILD file change. Looks like no file change? Also if needed is there a corresponding cmake change?

stablehlo/reference/InterpreterOps.cpp
```diff
@@ -30,6 +30,7 @@ limitations under the License.
 #include "mlir/IR/SymbolTable.h"
 #include "mlir/Support/DebugStringHelper.h"
 #include "mlir/Support/LLVM.h"
+#include "stablehlo/dialect/Base.h"
```
GleasonK, line 33: Is this needed? No file change? Also if it is needed, is there a corresponding bazel dep change?

GleasonK (changes requested):
LGTM, a few questions

stablehlo/dialect/Base.td
```diff
@@ -248,8 +255,8 @@ def HLO_StaticDimensionTensor : RankedTensorOf<[HLO_DimensionValue], [HasStaticS
 def HLO_StaticShapeTensor : StaticShapeTensorOf<[
       HLO_Float, HLO_Pred, HLO_Int, HLO_Complex, HLO_QuantizedInt]>;
 
-def HLO_StaticShapeTensorOrPerAxisQuantizedTensor : StaticShapeTensorOf<[
-      HLO_Float, HLO_Pred, HLO_Int, HLO_Complex, HLO_QuantizedInt, HLO_PerAxisQuantizedInt]>;
+def HLO_StaticShapeTensorOrPerAxisQuantizedTensor : RankedTensorOf<[HLO_Float, HLO_Pred, HLO_Int, HLO_Complex, HLO_QuantizedInt, HLO_PerAxisQuantizedInt],
```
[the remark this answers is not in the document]
sdasgup3: Initially we had
```
def HLO_StaticShapeTensorOrPerAxisQuantizedTensor : StaticShapeTensorOf<[
      HLO_Float, HLO_Pred, HLO_Int, HLO_Complex, HLO_QuantizedInt, HLO_PerAxisQuantizedInt]>;
```

`StaticShapeTensprOf` [cs](https://github.com/llvm/llvm-project/blob/c515c780244e3ecbb1fcfd06b3ad588d8d22c28e/mlir/include/mlir/IR/CommonTypeConstraints.td#L712) does not expose the pred which we would need to pass on `IsValidQuantizedDimension`.

This is same issue with https://github.com/openxla/stablehlo/blob/1beb2c3ad9a0650477367dabd081d232b771352e/stablehlo/dialect/Base.td#L244 resolved similarly.

One option we explored is to use something like
```
AnyTypeOf[
    StaticShapeTensorOf<[HLO_Float, HLO_Pred, HLO_Int, HLO_Complex, HLO_QuantizedInt]>,
   RankedTensorOf<[HLO_PerAxisQuantizedInt],
    [IsValidQuantizedDimension, HasStaticShapePred]>;
]
```

But that would require explicit casting in the code base at several places because  the indirection `AnyTypeOf` does not allow the type to be recognized as RankedTensor (despite the fact that  both of its constituents are RankedTensors)

stablehlo/dialect/Base.cpp
```diff
@@ -611,5 +611,75 @@ bool isSplatArray(ArrayRef<int64_t> arr, int64_t val) {
...
+      })) {
+    return false;
+  }
+
+  return true;
+}
+
+bool isValidQuantizedDimension(Type type) {
+  auto rankedType = dyn_cast<RankedTensorType>(type);
+  if (!rankedType) return true;
```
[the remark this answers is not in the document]
sdasgup3: We can have `HLO_Token` which can use used as type for certain ops via derivations like
 
```
def HLO_TensorOrToken : AnyTypeOf<[HLO_Tensor, HLO_Token]>;
```

stablehlo/dialect/Base.cpp
```diff
@@ -611,5 +611,75 @@ bool isSplatArray(ArrayRef<int64_t> arr, int64_t val) {
...
+                      quantizedPerAxisElementType.getZeroPoints().end());
+    scales.insert(scales.begin(),
+                  quantizedPerAxisElementType.getScales().begin(),
+                  quantizedPerAxisElementType.getScales().end());
+  }
+
+  // quantized_type_c5
+  auto maxPosFiniteNum =
+      APFloat::getLargest(quantizedElementType.getExpressedType()
+                              .cast<FloatType>()
```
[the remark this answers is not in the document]
sdasgup3: Yup at upstream at https://github.com/llvm/llvm-project/blob/8f07a67f9731dfcd490f8aaefac34d95f207b39c/mlir/lib/Dialect/Quant/IR/QuantTypes.cpp#L247

stablehlo/tests/CheckOps.cpp
sdasgup3: `stablehlo/tests/CheckOps.cpp.inc` needs the declaration for `isValidStablehloQuantizedElementType` which is in Base.h

Will add the cmake change. Earlier it was satisfied indirectly

stablehlo/reference/InterpreterOps.cpp
sdasgup3: `stablehlo/reference/InterpreterOps.cpp.inc` needs the declaration for isValidStablehloQuantizedElementType which is in Base.h

>> Also if it is needed, is there a corresponding bazel dep change?

You can find that in the  stablehlo/Bazel.build file.

stablehlo/dialect/Base.td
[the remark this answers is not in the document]
GleasonK: makes sense, didn't notice the `HasStaticShapePred` constraint initially.

stablehlo/dialect/Base.cpp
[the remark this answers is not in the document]
GleasonK: Not sure I follow.. looking at the ODS modifications in this CL none of them allow for token with this constraint(?)

stablehlo/dialect/Base.cpp
```diff
@@ -611,5 +611,75 @@ bool isSplatArray(ArrayRef<int64_t> arr, int64_t val) {
...
+          rankedType.getElementType());
+
+  if (!quantizedPerAxisElementType) return true;
+
+  // quantized_type_c12, quantized_type_c13, quantized_type_c14
+  int64_t quantDim = quantizedPerAxisElementType.getQuantizedDimension();
+  return quantDim >= 0 && quantDim < rankedType.getRank() &&
+         (!rankedType.isDynamicDim(quantDim) &&
+          static_cast<int64_t>(
+              quantizedPerAxisElementType.getScales().size()) ==
```
GleasonK, line 680: Let's make this a variable too, will make this return more readable.

```
int64_t numScales = static_cast<int64_t>(quantizedPerAxisElementType.getScales().size())
```

GleasonK (approved):

stablehlo/dialect/Base.cpp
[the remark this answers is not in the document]
sdasgup3: You are right, HLO_Token should not trigger the  check `isValidQuantizedDimension`. My bad!

Here is more appropriate example:
From this CL, we have

```
def HLO_AnyTensor : TensorOf<[HLO_Float, HLO_Pred, HLO_Int, HLO_Complex, HLO_QuantizedInt, HLO_PerAxisQuantizedInt],
    [IsValidQuantizedDimension]>;

def HLO_CustomCallValue : AnyTypeOf<[HLO_AnyTensor, HLO_Token, HLO_AnyTuple]>;
```

`HLO_AnyTensor` can be an input/output of a `custom_call` and an unranked tensor as an input or output of `custom_call`, like the following, 

```
func.func @custom_call_multiple_inputs_outputs(%x: tensor<*xf32>) -> tensor<*xf32> {
  %0 = "stablehlo.custom_call"(%x) {backend_config="", call_target_name = "foo", has_side_effect = false} : (tensor<*xf32>) -> tensor<*xf32>
  func.return %0 : tensor<*xf32>
}
```

should not have the check on quantization dimension.

PR #2207, opened 2024-04-12
Add a note: Type Inference for Quantization is not always feasible

abhigunj:


docs/status.md
```diff
@@ -159,3 +159,19 @@ one of the following tracking labels.
 | uniform_quantize         | yes           | revisit      | infeasible     | yes             | no          |
 | while                    | yes           | revisit      | yes            | revisit         | yes         |
 | xor                      | yes           | yes          | yes            | yes             | yes         |
+
+## Note
```
sdasgup3, line 163: ## Type inference for quantized operations

docs/status.md
```diff
@@ -159,3 +159,19 @@ one of the following tracking labels.
...
+
+- **Type inference and Quantization**: `Type Inference` column from the above
```
sdasgup3, line 165: The `Type Inference` column from the table above is intended to focus on non-quantized operations. For the majority of the quantized operations, it is not feasible to infer the result type because the quantization parameters of the result types may vary from those of the operands, with the exception of a few cases that are listed below.

docs/status.md
```diff
@@ -38,6 +38,12 @@ one of the following tracking labels.
     For example, because the result type of an op cannot be inferred from
     its operands and attributes.
 
+## Note
+
+- **Type inference for Quantization**: Result type inference is not always
+feasible with quantization. In quantization, the scales and zero-points of
```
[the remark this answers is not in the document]
abhigunj: taken care of in the latest draft, thank you

docs/status.md
```diff
@@ -159,3 +159,18 @@ one of the following tracking labels.
...
 | xor                      | yes           | yes          | yes            | yes             | yes         |
+
+## Type inference for quantized operations
+
+The `Type Inference` column from the table above is intended to focus on
+non-quantized operations. For the majority of the quantized operations, it is
+not feasible to infer the result type because the quantization parameters of
+the result types may vary from those of the operands, with the exception of
+few cases that are listed below: `all_gather`,
+`all_to_all`, `case`, `collective_permute`,
```
sdasgup3, line 170: why `dot_general` because we cannot infer the storage type of result 

```
(C15) storage_type(lhs) = storage_type(rhs).
(C16) expressed_type(lhs) = expressed_type(rhs) = expressed_type(result).
```

sdasgup3: Similarly we cannot infer the quantization parameter for result of a `uniform_quantized` operation.

abhigunj: yup..`storage_type ` can't be inferred. 

`uniform_quantized`.. forgot to remove. thanks for catching those.

GleasonK (approved):

Issue #2205, opened 2024-04-11
Evaluate the quantization type constraint `num_bits(storage_type) < num_bits(expressed_type)`


sdasgup3:
The following constraint  for quantized type

```
(C1) num_bits(storage_type) < num_bits(expressed_type).
```

seems to contradict checks like https://github.com/openxla/xla/blob/bf63042c7a659d32ae354ee6b96585c233482ff4/xla/mlir_hlo/mhlo/transforms/mhlo_quant_legalize_to_int/mhlo_quant_legalize_to_int.cc#L483.

Let get to the bottom on the relevance of the check.

sdasgup3:
<details><summary>Consider the  code snippet</summary> 

```
 func.func @main(%act: tensor<1x2x!quant.uniform<i8:f32, 0.3:-128>>, %weight: tensor<2x3x!quant.uniform<i8<-127:127>:f32:1, {0.23,0.23,0.23}>>, %bias: tensor<1x3x!quant.uniform<i32:f32:1, {0.83,0.83,0.83}>>) -> tensor<1x3x!quant.uniform<i8:f32, 0.27:-128>> {
    %0 = stablehlo.dot_general %act, %weight, contracting_dims = [1] x [0] : (tensor<1x2x!quant.uniform<i8:f32, 0.3:-128>>, tensor<2x3x!quant.uniform<i8<-127:127>:f32:1, {0.23,0.23,0.23}>>) -> tensor<1x3x!quant.uniform<i32:f32:1, {0.83,0.83,0.83}>>
    %1 = stablehlo.add %0, %bias : tensor<1x3x!quant.uniform<i32:f32:1, {0.83,0.83,0.83}>>
    %2 = stablehlo.uniform_quantize %1 : (tensor<1x3x!quant.uniform<i32:f32:1, {0.83,0.83,0.83}>>) -> tensor<1x3x!quant.uniform<i8:f32, 0.27:-128>>
    return %2 : tensor<1x3x!quant.uniform<i8:f32, 0.27:-128>>
 } 
```

 </details>

where the  quantized element-types (like `quant.uniform<i8:f32, 0.3:-128>`) are sign-extended to   `quant.uniform<i32:f32, scale:zp>` for accumulation, as part of dot semantics. while keeping the same scale (as that of the output type) which will eventually be used to scale back the accumulated result to float. 

[...]

PR #2211, opened 2024-04-13
Relax constraint `num_bits(storage_type) < num_bits(expressed_type)`

sdasgup3:
fixes https://github.com/openxla/stablehlo/issues/2205

Requesting the following for review:  
@loganchien
@paulinesho
@doyeonkim0

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -158,18 +158,17 @@ in the grammar, but have default values of `min_value(storage_type)` and
 `max_value(storage_type)` respectively. Quantized element types have the
 following constraints:
 
-* (C1) `num_bits(storage_type) < num_bits(expressed_type)`.
-* (C2) `type(storage_min) = storage_type`.
-* (C3) `type(storage_max) = storage_type`.
-* (C4) `min_value(storage_type) <= storage_min < storage_max <= max_value(storage_type)`.
-* (C5) `type(scales...) = expressed_type`.
-* (C6) `0 < scales`.
-* (C7) `is_finite(scales...)`.
-* (C8) `storage_min <= zero_points <= storage_max`.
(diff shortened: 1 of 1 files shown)


docs/spec.md
```diff
@@ -158,18 +158,17 @@ in the grammar, but have default values of `min_value(storage_type)` and
 `max_value(storage_type)` respectively. Quantized element types have the
 following constraints:
 
-* (C1) `num_bits(storage_type) < num_bits(expressed_type)`.
```
loganchien, line 161: Need to update `C_i` at Line 143-149?

sdasgup3: Thanks for pointing that. fixed.

loganchien (approved):

doyeonkim0:
Out of curiosity, do we have use case for `num_bits(storage_type) > num_bits(expressed_type)`?

GleasonK:
> Out of curiosity, do we have use case for `num_bits(storage_type) > num_bits(expressed_type)`?

+1

sdasgup3:
@doyeonkim0 
Per my offline chat with Andy, we might also have `i32` storage type and `bf16` expressed type for the same use case for matching the output type of convolution.

doyeonkim0:
That makes sense. Thanks for the answer!

doyeonkim0 (approved):

GleasonK (approved):
This feels in line with the proposal in:
https://github.com/openxla/stablehlo/blob/main/rfcs/20230622-quantized-reduction.md

Even in that proposal you use examples where the bitwidths violate this constraint as written (i.e. `!qunat.uniform<i32:f32, ...>`). Given that that proposal was approved, I think this is a reasonable spec correction.

PR #2213, opened 2024-04-15
Document that dimension sizes can be dynamic

mlevesquedion:
This satisfies P4 of the Dynamism RFC: https://github.com/openxla/stablehlo/blob/main/rfcs/20230704-dynamism-101.md#p4

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -101,7 +101,7 @@ DimensionSize ::= digit {digit} | '?'
 ```
 
 **Tensor types** represent tensors, i.e. multidimensional arrays. They have a
-**shape** and an **element type**, where a shape represents non-negative or
+**shape** and an **element tyThis is for (P3) of the dynamism RFC: https://github.com/openxla/stablehlo/blob/main/rfcs/20230704-dynamism-101.md#p3pe**, where a shape represents non-negative or
 unknown **dimension sizes** in the ascending order of the corresponding
 **dimensions** (which are also called **axes**) numbered from `0` to `R-1`. The
 number of dimensions `R` is called **rank**. For example, `tensor<2x3xf32>` is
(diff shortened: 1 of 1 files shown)


docs/spec.md
```diff
@@ -97,24 +97,23 @@ are not value types).
...
+**shape** and an **element type**, where a shape represents non-negative or
+unknown **dimension sizes** in the ascending order of the corresponding
+**dimensions** (which are also called **axes**) numbered from `0` to `R-1`. The
+number of dimensions `R` is called **rank**. For example, `tensor<2x3xf32>` is
+a tensor type with shape `2x3` and element type `f32`. It has two dimensions
+(or, in other words, two axes) - 0th dimension and 1st dimension - whose sizes
+are 2 and 3. Its rank is 2.
+
+Shapes can be partially or completely unknown (dynamic). Dynamic dimension sizes
+are represented using a `?`. Shapes may not be unranked.
```
sdasgup3, line 113: Can we be more assertive here like "Shape cannot be unranked". 

>> Shapes can be partially or completely unknown (dynamic)

I believe partially unknown means bounded shapes. Are we planning to update the spec with bounded dynamism also. If not in near future we can just write   `Shapes can be  unknown (dynamic)`

mlevesquedion: > Can we be more assertive

Yes, good idea. Done.

> I believe partially unknown means bounded shapes.

For shapes, in my understanding "partially unknown" means partially dynamic, e.g. `tensor<?x2xf64>`. I clarified the text with an example, thanks.

> Are we planning to update the spec with bounded dynamism also

We probably should? I would consider that out of scope for the RFC though.

docs/spec.md
```diff
@@ -97,24 +97,24 @@ are not value types).
...
-in other words, two axes) - 0th dimension and 1st dimension - whose sizes are
-2 and 3. Its rank is 2.
-
-This defines support for **static shapes** where dimension sizes are statically
-known. In the future, we are planning to also introduce support for
-**dynamic shapes** where dimension sizes are either partially or fully unknown
-([#8](https://github.com/openxla/stablehlo/issues/8)). Furthermore, we are
-planning to explore extending tensor types beyond dimension sizes and element
-types, for example, to include layouts
+**shape** and an **element tyThis is for (P3) of the dynamism RFC: https://github.com/openxla/stablehlo/blob/main/rfcs/20230704-dynamism-101.md#p3pe**, where a shape represents non-negative or
```
sdasgup3: typo `ty` --> `type`

sdasgup3 (approved):

docs/spec.md
mlevesquedion: Thanks, accidentally pasted a bunch of stuff in here. Fixed.

PR #2217, opened 2024-04-15
Remove unnecessary {RankedTensor,Shaped}Type casts

mlevesquedion:
I revisited all the casts to ShapedType and RankedTensorType in light of the fact that most op input/output types are ranked and removed unnecessary casts.

This PR is sort of an extension of https://github.com/openxla/stablehlo/pull/2183

Also simplified a few `op.getResult().getType()` to `op.getType()` because I was modifying that code anyway.

Fixes https://github.com/openxla/stablehlo/issues/2065

### DIFF
--- a/stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
+++ b/stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
@@ -397,8 +397,7 @@ struct BroadcastConverter final
 
   static SmallVector<AffineMap, 2> getIndexingMaps(OpTy broadcastOp,
                                                    Builder *b) {
-    ShapedType inputType =
-        llvm::cast<ShapedType>(broadcastOp.getOperand().getType());
+    ShapedType inputType = broadcastOp.getOperand().getType();
     unsigned inputRank = inputType.getRank();
     unsigned nloops = getHloOpResultType(broadcastOp).getRank();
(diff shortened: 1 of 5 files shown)


stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
```diff
@@ -1047,7 +1046,7 @@ struct ReshapeOpConverter final
     Value operand = adaptor.getOperand();
     auto operandType = llvm::cast<ShapedType>(operand.getType());
     Type elemType = operandType.getElementType();
-    auto resultType = llvm::cast<ShapedType>(reshapeOp.getType());
+    ShapedType resultType = reshapeOp.getType();
```
ghpvnist, line 1049: Could we use auto everywhere? (This goes against the rule of using auto sparingly but the file uses it quite consistently).

ghpvnist (approved):
LGTM with minor comment. Thanks!

stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
mlevesquedion: Good catch! Unfortunately there is an assignment below (on line 1061) that has to use `ShapedType`, so this one can't use `auto`.

PR #2224, opened 2024-04-16
Enable testdata for convolution ops

ghpvnist:
This is part 2 of #1964 to implement parts of #1314.

Also fix a bug on generating invalid padding shape (used to generate `N-2` number of dim size 2 instead of shape `[N-2, 2]`).

### DIFF
--- a/stablehlo/testdata/conv_general_dilated_depthwise1d_dilated_lhs_float32_2_3_9__rhs_float32_12_1_3__windowstrides__1___p4726467912149865720.mlir
+++ b/stablehlo/testdata/conv_general_dilated_depthwise1d_dilated_lhs_float32_2_3_9__rhs_float32_12_1_3__windowstrides__1___p4726467912149865720.mlir
@@ -1,4 +1,4 @@
-// RUN-DISABLED: stablehlo-opt -inline %s | stablehlo-translate --interpret
+// RUN: stablehlo-opt -inline %s | stablehlo-translate --interpret
 // RUN: stablehlo-translate --serialize --target=current %s | stablehlo-translate --deserialize | stablehlo-opt > %t.0
 // RUN: stablehlo-opt %s > %t.1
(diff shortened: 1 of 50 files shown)


GleasonK (approved):

stablehlo/testdata/conv_general_dilated_depthwise1d_dilated_lhs_float32_2_3_9__rhs_float32_12_1_3__windowstrides__1___p4726467912149865720.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN-DISABLED: stablehlo-opt -inline %s | stablehlo-translate --interpret
+// RUN: stablehlo-opt -inline %s | stablehlo-translate --interpret
```
GleasonK, line 1: Can you validate that these work on windows / mac as well -- can use internal infra to validate. I thought we had FP issues with some numerics, are we safe to enable all these?

ghpvnist: I've verified that these tests successfully pass internal CI on all platforms. FP issue we had was due to a magic number we used for fp tolerance for all data types: [link](https://github.com/openxla/stablehlo/blob/b9118430c760e1283d2e48791e80065b7dcc0db5/stablehlo/reference/Element.cpp#L174)

ghpvnist: Also passes on GH CI except one. The failure seems to be unrelated to the change though: https://github.com/tensorflow/tensorflow/pull/66020.

PR #2229, opened 2024-04-17
Implement ConditionallySpeculatable for SelectAndScatter

mlevesquedion:
The spec says (C11): shape(operand) = shape(result).

### DIFF
--- a/stablehlo/dialect/Base.h
+++ b/stablehlo/dialect/Base.h
@@ -422,6 +422,23 @@ struct SpeculatableIfStaticDimInOutputIsStaticInInputImplTrait
 };
 
 template <typename ConcreteType>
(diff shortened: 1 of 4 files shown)


ghpvnist (changes requested):
Could you provide one line comment describing the unit tests similar to how you did for other ops?

mlevesquedion:
> Could you provide one line comment describing the unit tests similar to how you did for other ops?

Sure thing, done. I think I'll need to revisit some of the tests when I'm done to improve consistency, would probably be better to have comments everywhere and use more consistent names (e.g. the `speculatable_0` etc. names are redundant).

ghpvnist (approved):

PR #2230, opened 2024-04-17
Implement ConditionallySpeculatable for SetDimensionSize

mlevesquedion:
This op isn't specced yet but the speculation logic is relatively straightforward and follows from the tablegen/type inference/XLA documentation: https://openxla.org/xla/operation_semantics#setdimensionsize.

### DIFF
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -1911,15 +1911,11 @@ LogicalResult SetDimensionSizeOp::inferReturnTypeComponents(
 }
 
 mlir::Speculation::Speculatability SetDimensionSizeOp::getSpeculatability() {
-  // If the dimension being set is constant, the verifier will have checked that
-  // it matches the corresponding dimension in the output.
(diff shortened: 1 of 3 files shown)


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -1910,6 +1910,29 @@ LogicalResult SetDimensionSizeOp::inferReturnTypeComponents(
       adaptor.getSize(), adaptor.getDimension(), inferredReturnShapes);
 }
 
+mlir::Speculation::Speculatability SetDimensionSizeOp::getSpeculatability() {
+  // If the dimension being set is constant, the verifier will have checked that
+  // it matches the corresponding dimension in the output.
+  if (matchPattern(getSize(), m_Constant()))
+    return mlir::Speculation::Speculatable;
```
sdasgup3, line 1917: Don't we have to do the check for remaining dimension?

mlevesquedion: Good point! Added the missing test case and fixed. Thanks!

sdasgup3 (approved):

PR #2231, opened 2024-04-17
Implement ConditionallySpeculatable for DynamicReshape

mlevesquedion:


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -1508,6 +1508,26 @@ LogicalResult DynamicReshapeOp::reifyReturnTypeShapes(
...
+                   [this](int64_t i) { return getType().isDynamicDim(i); }))
+    return mlir::Speculation::Speculatable;
+
+  // If the input is static and the shape operand is constant, the output
+  // shape can be inferred and any mismatch will be caught statically.
+  // If any dimension in the input is dynamic, the number of elements may
+  // disagree with either the output.
+  // If the shape operand is not constant, it could disagree with the output,
+  // which has at least 1 static dimension at this point in the function.
+  if (getOperand().getType().hasStaticShape() &&
```
ghpvnist, line 1524: Consider a few more cases:
This seems like a no-op to me so I think it should be speculatable.
```
%result = stablehlo.dynamic_reshape %operand, %output_shape : (tensor<2x2xf32>, tensor<2xi32>) -> tensor<2x?xf32>
```

I don't think we do checks like this at all, so the mismatch won't be caught.
```
%output_shape = stablehlo.constant dense<[2,2]> : tensor<2xi32>
%result = stablehlo.dynamic_reshape %operand, %output_shape : (tensor<5x?xf32>, tensor<2xi32>) -> tensor<2x2xf32>
```

ghpvnist (changes requested):

stablehlo/dialect/StablehloOps.cpp
mlevesquedion: > Consider a few more cases: This seems like a no-op to me so I think it should be speculatable.
> 
> ```
> %result = stablehlo.dynamic_reshape %operand, %output_shape : (tensor<2x2xf32>, tensor<2xi32>) -> tensor<2x?xf32>
> ```

I don't understand how this would be a no-op. Do you mean that the `?` can be inferred to be `2` because the input is `2x2`? That may be the case, but the `%output_shape` could turn out to be e.g. `dense<[3, 4]> : tensor<2xi32>` at runtime, so there would be a mismatch.

> I don't think we do checks like this at all, so the mismatch won't be caught.
> 
> ```
> %output_shape = stablehlo.constant dense<[2,2]> : tensor<2xi32>
> %result = stablehlo.dynamic_reshape %operand, %output_shape : (tensor<5x?xf32>, tensor<2xi32>) -> tensor<2x2xf32>
> ```

You are right that we don't catch this in type inference (although we could, because we know that `5x?` could never match up with `2x2`). However, because the input is not fully static, this is not speculatable. I'm not sure what "the mismatch won't be caught" is referring too. Can you expand?

Thanks for taking the time to review! :)

ghpvnist: > Do you mean that the ? can be inferred to be 2 because the input is 2x2?

Yup that's what I was referring to. Thanks for the explanation.

> "the mismatch won't be caught"

I was referring to your comment:
>If the input is static and the shape operand is constant, the output
shape can be inferred and any mismatch will be caught statically.

We don't have verifiers for dynamic ops currently, but I guess that is left for future work.

Good discussion, change looks good to me then :)

ghpvnist (approved):

stablehlo/dialect/StablehloOps.cpp
mlevesquedion: Thanks for the discussion! :)

> We don't have verifiers for dynamic ops currently, but I guess that is left for future work.

We actually do, at least for DynamicReshape and a few others. I haven't looked at all of them, but here's the code that checks that the constant shape matches the result: https://github.com/openxla/stablehlo/blob/1beb2c3ad9a0650477367dabd081d232b771352e/stablehlo/dialect/TypeInference.cpp#L3790

And here's an example that can be fed to `stablehlo-opt` to confirm that it works:

```
func.func @foo(%operand: tensor<2x3xf32>) {
  %output_shape = stablehlo.constant dense<[1, 10]> : tensor<2xi32>
  %result = stablehlo.dynamic_reshape %operand, %output_shape : (tensor<2x3xf32>, tensor<2xi32>) -> tensor<2x2xf32>
  return
}
```

The verifier is incomplete though, as the following does not error:

```
func.func @foo(%arg0: tensor<3x4xf32>) {
  %0 = stablehlo.constant dense<[11]> : tensor<1xi32>
  %1 = stablehlo.dynamic_reshape %arg0, %0 : (tensor<3x4xf32>, tensor<1xi32>) -> tensor<?xi32>
  return
}
```

I updated the verifier to handle that case as well, PTAL. (I also checked internally that it does not cause unexpected failures.)

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -3164,18 +3164,9 @@ func.func @dynamic_reshape_incompatible_shapes(%arg0: tensor<?xf32>, %shape: ten
...
-  // expected-error@+2 {{output_shape is incompatible with return type of operation 'tensor<1x4xf32>'}}
-  %0 = stablehlo.constant dense<[-1, 1]> : tensor<2xi64>
-  %1 = stablehlo.dynamic_reshape %arg0, %0 : (tensor<4xf32>, tensor<2xi64>) -> tensor<1x4xf32>
-  return %1 : tensor<1x4xf32>
-}
-
-// -----
-
 func.func @dynamic_reshape_output_shape_mismatching_size(%arg0: tensor<4xf32>) -> tensor<1x4xf32> {
   // expected-error@+2 {{output_shape is incompatible with return type of operation 'tensor<1x4xf32>'}}
```
ghpvnist, line 3168: Since you're updating the verifier, I thought I'd point this out. At first glance, `tensor<4xf32>` should be compatible with `tensor<1x4xf32>`. Including that the `output_shape` in this particular case is a constant and printing the contents of `output_shape` with the error message would help with debugging in the future. It won't be as obvious to catch the constant `output_shape` in production when MLIR is not readily available (e.g. in a SavedModel). Wdyt?

mlevesquedion: I agree, but it's a bit onerous to add this and I'd like to keep this change focused. Would it be okay if I did that in a follow-up?

ghpvnist (approved):

mlevesquedion:
Thanks for the review, @ghpvnist!

PR #2232, opened 2024-04-18
Tutorial: JAX Export to StableHLO

GleasonK:
See tutorial rendering using "View File" ([example](https://github.com/openxla/stablehlo/blob/f85d9fffc3d4e0fe4f09bfadd4739d09baaf60c1/docs/tutorials/jax-export.ipynb)).

Once merged the "Open in Colab" button should work. Also once integrated this should appear on the openxla.org documentation as well.

### DIFF
--- a/build_tools/github_actions/lint_check_license.sh
+++ b/build_tools/github_actions/lint_check_license.sh
@@ -51,6 +51,7 @@ echo
 SKIPPED_SUFFIXES=(
   .clang-format
   .gitignore
+  .ipynb
   .json
   .md
   .mlir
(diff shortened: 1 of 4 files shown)


mlevesquedion (approved):
Looks great!

A few nits:
- The title could be more evocative, e.g. "Exporting StableHLO from JAX" or even "How to Export StableHLO from JAX".
- The name "prettyprint_stablehlo" is somewhat inaccurate, given that it actually print. Maybe "render_stablehlo" or something.
- There are a few typos here and there, e.g. "JAX will generated guards" --> "JAX will generate guards"

Nothing super important. Too bad the export format for notebooks isn't nicer, otherwise I would have commented directly where appropriate.

PR #2233, opened 2024-04-18
Implement ConditionallySpeculatable for {Dynamic,}BroadcastInDim

mlevesquedion:


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -1343,6 +1343,27 @@ LogicalResult DynamicBroadcastInDimOp::reifyReturnTypeShapes(
   return success();
 }
 
+mlir::Speculation::Speculatability
+DynamicBroadcastInDimOp::getSpeculatability() {
+  auto operandType = getOperand().getType();
+
+  // If input is dynamic, not speculatable.
+  if (!operandType.hasStaticShape()) return mlir::Speculation::NotSpeculatable;
```
sdasgup3, line 1351: How about:
```
If input is dynamic, the broadcasting rules might be violated at runtime and hence not speculatable. 
```

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -1343,6 +1343,27 @@ LogicalResult DynamicBroadcastInDimOp::reifyReturnTypeShapes(
...
+
+  // If input is broadcastable (all 1's) and result is dynamic, speculatable.
+  auto resultDynamic =
```
sdasgup3, line 1354: Can't we use something like
```
auto resultDynamic = !getType().hasSaticShape();
```

stablehlo/dialect/StablehloOps.cpp
mlevesquedion: Thanks for the suggestion! Done.

sdasgup3 (approved):

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3714,6 +3714,23 @@ LogicalResult verifyDynamicBroadcastInDimOp(
                                " does not refer to a "
                                "valid operand dimension");
 
+  if (SmallVector<int64_t> shape;
+      operandType.hasStaticShape() &&
+      matchInts(outputDimensions, shape).succeeded()) {
+    for (int64_t i = 0; i != bcastDimensionsSize; ++i) {
```
ghpvnist, line 3720: I recently found out, but there's a `llvm::enumerate` to replace this.

```
for (auto [i, dimIndex] : llvm::enumerate(broadcastDimensions)
...
```

ghpvnist (approved):
LGTM with one comment.

stablehlo/dialect/TypeInference.cpp
mlevesquedion: Good catch, thanks!

PR #2240, opened 2024-04-18
Add compile_commands flag to README

ghpvnist:
We don't provide documentation on how to generate compile_commands.json and this could be a useful start. I left this flag out of CI since the CI doesn't need to generate this.


mlevesquedion (approved):

mlevesquedion:
I use this all the time, I think this is likely to be useful for most folks, and won't be very costly in general, so no downside for folks who don't use it.

PR #2253, opened 2024-04-24
Extract folders from StablehloRefineShapes into new StablehloAggressiveFolder pass

mvpant:
Extract folders from StablehloRefineShapes into new StablehloAggressiveFolder pass.

This patch addresses the problem of adding new folders and allowing potentially lossy computations in folding patterns. It preserves the behavior of existing models and workflows that rely on shape refinement to fold shapes while assuming non-shape-related computations remain unchanged.


mvpant:
It tends out that `applyPartialConversion` approach is only feasible here if we are 100% sure that folder will be applied, i.e. on the case below:
```mlir
func.func @eval_add_i32(%arg0 : tensor<i32>) -> tensor<i32> {
  // CHECK-NOT: stablehlo.add
  // CHECK: [[RESULT:%.*]] = stablehlo.constant dense<4> : tensor<i32>
  // CHECK: return [[RESULT]]
  %0 = stablehlo.constant dense<2> : tensor<i32>
  %2 = stablehlo.add %arg0, %0 : tensor<i32>
  func.return %2 : tensor<i32>
}
```
this code won't work properly as it would fail to legalize:
```c++
ConversionTarget target(getContext());
  target.addLegalDialect<stablehlo::StablehloDialect>();
  target.addDynamicallyLegalOp<stablehlo::AddOp>([&](stablehlo::AddOp op) {
    return !op.getLhs().getType().getElementType().isa<IntegerType>() &&
           !op.getRhs().getType().getElementType().isa<IntegerType>();
  });
```

Alternative we can try to use greedy folder only on a subset of `integer-like` operations and then apply shape refinement:
```c++
SmallVector<Operation*> opsToFold;
func.walk([&](Operation* op) {
  if (!isOpFoldableByStablehloAggressiveFolder(op)) return;

  if (isa<AddOp, AndOp, CompareOp, DivOp, MaxOp, MinOp, MulOp, OrOp, RemOp,
          SubtractOp>(op))
    for (const auto& operand : op->getOperands())
      if (!getElementTypeOrSelf(operand.getType()).isa<IntegerType>())
        return;

  if (auto operation = dyn_cast<BroadcastInDimOp>(op))
    if (!getElementTypeOrSelf(operation.getOperand().getType())
             .isa<IntegerType>())
      return;

  ...

  opsToFold.push_back(op);
});
GreedyRewriteConfig config;
config.strictMode = GreedyRewriteStrictness::ExistingAndNewOps;
(void)applyOpPatternsAndFold(opsToFold, foldPatterns, config);

if (failed(applyPatternsAndFoldGreedily(func, refinePatterns, config))) {
  func.emitError("Failed to converge StablehloRefineShapes in ")
      << config.maxIterations << " iterations";
  signalPassFailure();
}
...

```

stablehlo/transforms/Passes.h
```diff
@@ -61,6 +61,10 @@ void populateVhloToVersionPatterns(RewritePatternSet *patterns,
 void populateChloToStablehloPatterns(MLIRContext *context,
                                      RewritePatternSet *patterns);
 
+/// Collection of folding patterns for StableHLO.
+void populateStablehloAggressiveFolderPatterns(RewritePatternSet *patterns,
```
GleasonK, line 65: Let's also make a `populateStablehloShapeFolderPatterns` which will have the subset needed by `StablehloRefineShapes`. That way when we add more patterns to this folder pass, they won't impact shape refinement

GleasonK (commented):
Looks great! Thanks for keeping this as a NFC PR for the initial submit. Will make it easier to evaluate any followups.

stablehlo/transforms/Passes.h
GleasonK: I suppose the alternative would be to have shape refinement only classify certain ops as illegal so only those patterns apply.. I'm also OK with that approach, I'm happy to make that change in a followup.

mvpant: I like `populateStablehloShapeFolderPatterns` approach. In the next patch we can add `fold_float` option and pass it into convert pattern, but keep it constantly false for shape folders.

mvpant: I plan to add a new `tests/stablehlo_aggressive_folder.mlir` in the upcoming patches for iota and convert on floats.

mvpant:
```
diff <(sed -n '59,530p' stablehlo/transforms/StablehloAggressiveFolder.cpp) <(git show ab92aded:stablehlo/transforms/StablehloRefineShapes.cpp | sed -n '249,720p')
```
should be identical

mvpant:
```
diff <(sed -n '59,474p' stablehlo/transforms/StablehloAggressiveFolder.cpp) <(git show fdfab2e4:stablehlo/transforms/StablehloRefineShapes.cpp | sed -n '249,664p')
```
rebased, should be identical

GleasonK:
Will review this next week -- Google I/O next week and have been a bit bogged down. Apologies for the delay!

GleasonK:
> Any approach introduces the challenge of synchronizing between shape refinement and folders. If someone adds a pattern to the folders, what should be done with shape refinement? Once i add a new EvalIota pattern should i skip it in shape refinement to maintain original behavior? I'm not entirely sure about the criteria that determine which folders are suitable for shape refinement and which are not.

Shape refinement is very limited in use. It's only intended to fold specific patterns that appear in shape computations.

Mostly this is "dynamic reshape|broadcast this into the shape of another tensor" -- reshape requires multiplication/division to figure out number of elements in a dimension, and broadcast requires max to figure out which dims are broadcast, both require get_dimension_size and concat to build the shape vector. It's not a set of patterns that commonly grows.

We can figure out how to refactor this, thanks for pointing out the problem with the dynamic legalization, hadn't thought of that. Another approach could be to make a pattern that wraps a pattern and only applies when int... not sure how bad that is.. Will review this code now and see if I have any thoughts

GleasonK (approved):
This looks great. Thanks for putting this together! Let me think on better separating shape folders from other patterns. Going to hack something together hopefully tomorrow, may pick up one of your other PRs to use as an example

PR #2265, opened 2024-04-29
Clean up op speculatability tests

mlevesquedion:
Mostly, simplify redundant variable names, remove unnecessary ops in nested regions, and move ops around/add whitespace for readability.

### DIFF
--- a/stablehlo/tests/ops_speculatability.mlir
+++ b/stablehlo/tests/ops_speculatability.mlir
@@ -9,8 +9,8 @@
 // CHECK-LABEL: func @abs_multidim
 // CHECK-NEXT:  return
 func.func @abs_multidim(%dynamic_arg: tensor<?x?xf64>) {
-  %not_speculatable = stablehlo.abs %dynamic_arg : (tensor<?x?xf64>) -> tensor<?x2xf64>
-  "hlo_test_speculatability.is_not_speculatable"(%not_speculatable) : (tensor<?x2xf64>) -> ()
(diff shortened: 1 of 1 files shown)


stablehlo/tests/ops_speculatability.mlir
```diff
@@ -485,42 +485,41 @@ func.func @slice(%static_arg: tensor<2xf64>, %dynamic_arg: tensor<?xf64>) {
 // CHECK-LABEL: func @all_reduce
 // CHECK-NEXT:  return
 func.func @all_reduce(%static_arg: tensor<2xf64>, %dynamic_arg: tensor<?xf64>) {
-  %speculatable_0 = "stablehlo.all_reduce"(%static_arg) ({
+  %0 = "stablehlo.all_reduce"(%static_arg) ({
```
ghpvnist, line 488: I actually prefer this format to test specularity after each op, but since the majority of the related tests list all the ops first and then the specularity stuff, we can keep this format too, but let's use one format consistently across the tests.

ghpvnist (changes requested):

stablehlo/tests/ops_speculatability.mlir
mlevesquedion: SG, let me just change everything to test after each op then. I like having the various test cases grouped because it makes it easier to see the differences. But with ops that take more than a few lines, it becomes hard to see the speculatability checks. Since the speculatability are just one line though, it doesn't hurt visual diffability too much to always put the checks right after the ops they are checking.

Done.

ghpvnist (approved):

PR #2268, opened 2024-04-30
Add tests for uniform_{de,}quantize speculatability

mlevesquedion:
These ops are UnaryElementwise but they were missed in https://github.com/openxla/stablehlo/pull/2150


sdasgup3:
Shouldn't we add [Pure] to these ops.

mlevesquedion:
> Shouldn't we add [Pure] to these ops.

These ops are actually not pure. Because they are `StableHLO_UnaryElementwiseOp`, they get the `[HLO_SpeculatableIfStaticDimInOutputIsStaticInInput, NoMemoryEffect]` traits (the spec says `shape(operand) = shape(result)`).

sdasgup3 (approved):

PR #2269, opened 2024-04-30
Add helper to verify output shape matches result type

mlevesquedion:
And use the helper to verify shape operands in DynamicBroadcastInDim, DynamicIota and DynamicReshape.

Also, reorganize the isCompatibleForHloTypeInference logic a little bit to avoid duplication.

This is a follow-up/generalization of https://github.com/openxla/stablehlo/pull/2264.


stablehlo/dialect/TypeInference.cpp
```diff
@@ -468,6 +471,28 @@ LogicalResult verifyAddOp(std::optional<Location> location, Operation* op,
   return success();
 }
 
+// If the shape operand is constant, checks that it matches the result.
+// If not, emits an error.
+LogicalResult verifyShapeOperandMatchesResultType(std::optional<Location> loc,
```
ghpvnist, line 476: How about `verifyOperandAndResultTypesAreCompatible` or `verifyOperandTypeIsCompatibleWithResultType`? It would be more accurate to use the word "compatible".

stablehlo/tests/ops_stablehlo.mlir
ghpvnist, line 1: Do we have a positive test for compatible shapes?

ghpvnist (approved):
LGTM with few comments.

stablehlo/tests/ops_stablehlo.mlir
mlevesquedion: We don't, thanks I added some.

Issue #2267, opened 2024-04-30
Spec Dynamic ops 


abhigunj:
list of ops to be specced
- [x] #2288
- [x] #2289
- [x] #2290
- [x] #2291
- [x] #2292
- [x] #2293
- [x] real_dynamic_slice (planning to deprecate this op, spec not required)

dynamic_slice and dynamic_update_slice (already specced)

PR #2270, opened 2024-04-30
dynamic_iota op spec

abhigunj:
Constraints in the spec:
```
I1. output_shape is 1-dimensional tensor constant of type si64
I2. iota_dimension of type si64
C1. 0 <= iota_dimension < size(output_shape)
C2. rank(result) = size(output_shape)
```
Test coverage;
```
I1.1  output_shape is not a 1-dimensional tensor constant of type si64    (covered by ODS) si64 (covered by ODS)
I2.1  iota_dimension is not of type
C1.1 0 <= iota_dimension   (Already present)
C1.2  iota_dimension < size(output_shape)   (Already present)
C2.1  rank(result) = size(output_shape)   (Already present)
```

Reference interpreter:
partial implementation to support only static shapes.

ref: https://github.com/openxla/stablehlo/issues/2267


docs/spec.md
```diff
@@ -2633,6 +2633,52 @@ planning to address this in
...
+`output[result_index] = constant(is_quantized(output) ?
+quantize(result_index[iota_dimension], element_type(output)) :
+result_index[iota_dimension], element_type(output))`.
+
+#### Inputs
+
+| Label | Name             | Type                                                                               | Constraints |
+|-------|------------------|------------------------------------------------------------------------------------|-------------|
+| (I1)  | `iota_dimension` | `si64`                                                                             | (C1)        |
+| (I2)  | `output_shape`   | 1-dimensional tensor constant of type `si64`                                       | (C1), (C2)  |
```
abhigunj, line 2653: naming: `output_shape` or just  `shape` ?

docs/spec.md
```diff
@@ -2633,6 +2633,52 @@ planning to address this in
 
 &nbsp;[More Examples](https://github.com/openxla/stablehlo/tree/main/stablehlo/tests/interpret/dot_general.mlir)
 
+### dynamic_iota
+
+#### Semantics
+
+Fills an `output` tensor with values in increasing order starting from zero
+along the `iota_dimension` dimension. `output` tensor shape is specified
```
ghpvnist, line 2641: Result shape is provided dynamically via `output_shape`.

docs/spec.md
```diff
@@ -2633,6 +2633,52 @@ planning to address this in
...
+#### Outputs
+
+| Name     | Type                                                                             | Constraints |
+|----------|----------------------------------------------------------------------------------|-------------|
+| `output` | tensor of integer, floating-point or complex type or per-tensor quantized tensor | (C2)        |
+
+#### Constraints
+
+* (C1) `0 <= iota_dimension < size(output_shape)`.
+* (C2) `shape(output) = output_shape`.
```
ghpvnist, line 2664: (C2) should be removed since `output` is dynamic shape. Similarly in the Outputs Type column.

docs/spec.md
```diff
@@ -2633,6 +2633,52 @@ planning to address this in
...
+
+#### Examples
+
+```mlir
+
+// %output_shape = [4, 5]
+%output = "stablehlo.dynamic_iota"(%output_shape) {
+  iota_dimension = 0 : i64
+} : (tensor<2xi64>) -> tensor<?x?xi32>
```
ghpvnist, line 2673: `} : (tensor<2xi64>) -> tensor<?x?xi64>`

docs/spec.md
```diff
@@ -2633,6 +2633,52 @@ planning to address this in
...
+| Label | Name             | Type                                                                               | Constraints |
+|-------|------------------|------------------------------------------------------------------------------------|-------------|
+| (I1)  | `iota_dimension` | `si64`                                                                             | (C1)        |
+| (I2)  | `output_shape`   | 1-dimensional tensor constant of type `si64`                                       | (C1), (C2)  |
+
+#### Outputs
+
+| Name     | Type                                                                             | Constraints |
+|----------|----------------------------------------------------------------------------------|-------------|
+| `output` | tensor of integer, floating-point or complex type or per-tensor quantized tensor | (C2)        |
```
ghpvnist, line 2659: Let's add a comma before "or" to separate the per-tensor quantized tensor:
`tensor of integer, floating-point, or complex type or per-tensor quantized tensor`

Similarly in Iota spec.

docs/spec.md
```diff
@@ -2633,6 +2633,52 @@ planning to address this in
 
 &nbsp;[More Examples](https://github.com/openxla/stablehlo/tree/main/stablehlo/tests/interpret/dot_general.mlir)
 
+### dynamic_iota
+
+#### Semantics
+
+Fills an `output` tensor with values in increasing order starting from zero
+along the `iota_dimension` dimension. `output` tensor shape is specified
+dynamically via `output_shape` input. More formally,
```
ghpvnist, line 2642: @sdasgup3 Should we include formal calculations for dynamic ops? `result_index` is technically unknown until runtime. Maybe it's better to say something like this, which I borrowed from [StablehloOps.td](https://github.com/openxla/stablehlo/blob/205f405613a14554095d678ebd87c1ced924f7af/stablehlo/dialect/StablehloOps.td#L130-L132):
```
Informally, this operation does the same thing as IotaOp except that the
result shape is specified dynamically via `output_shape`:
https://github.com/openxla/stablehlo/blob/main/docs/spec.md#iota
```

docs/spec.md
ghpvnist: Should be [`output_shape`](https://github.com/openxla/stablehlo/blob/205f405613a14554095d678ebd87c1ced924f7af/stablehlo/dialect/StablehloOps.td#L140).

docs/spec.md
ghpvnist, line 2659: The name should be [`result`](https://github.com/openxla/stablehlo/blob/205f405613a14554095d678ebd87c1ced924f7af/stablehlo/dialect/StablehloOps.td#L141).

ghpvnist (changes requested):
Please use the [checklist](https://github.com/openxla/stablehlo/blob/main/docs/spec_checklist.md) for the remaining items.

docs/spec.md
mlevesquedion: +1

docs/spec.md
mlevesquedion: This seems to have been copied from `iota`. It seems we mostly use `result`, but a few specs use `output`. We should unify this.

docs/spec.md
mlevesquedion: Is this for consistency, or some other reason? The example for `iota` uses `i32` as the result's element type. I kinda like using different types because it emphasizes that the type of the dimension sizes is unrelated to the element type of the output tensor.

mlevesquedion (approved):
LGTM assuming comments are addressed.

docs/spec.md
ghpvnist: Changing name has implications for serialization and we have a ticket to track this https://github.com/openxla/stablehlo/issues/351. Unifying this is out of scope of this PR though.

docs/spec.md
ghpvnist: Yes for consistency and preferring to test highest bitwidth per [G1.2](https://github.com/openxla/stablehlo/blob/main/docs/reference.md#testing-guidelines)

docs/spec.md
abhigunj: thanks for confirming. Yeh, I used `output_shape` from the op def.

In general, we can always go back and update the op def as per the spec?

docs/spec.md
abhigunj: tensor element type mismatch is expected between `output_shape` and `output`

I kept it mismatched for the same reason.

abhigunj: missed previous reply. Thanks. 

will change it to highest bitwidth (i64) as per [G1.2](https://github.com/openxla/stablehlo/blob/main/docs/reference.md#testing-guidelines)

docs/spec.md
abhigunj: from other discussion - 
"Changing name has implications for serialization and we have a ticket to track this https://github.com/openxla/stablehlo/issues/351."

docs/spec.md
ghpvnist: Spec does not constrain this, so it should be clear there can be mismatches. I think this is generally the case too; I can't think of an op at the top of my head that requires attribute element type to match result element type.

docs/spec.md
abhigunj: for now removed formal calculations. 

keeping this chat open

sdasgup3: From https://github.com/openxla/stablehlo/blob/a7904883754a455bfd0c7636cbac1f99c2c24fe5/docs/spec.md?plain=1#L6605, the argument of `index_space` can be a placeholder. 
```
Placeholder` (future values, e.g. `lhs`, `rhs` or `result`; their actual
values are not known yet, only their types are known)
```

That means we can use the dynamic information of `output_shape` to bounds of `result_index` and that is all we need for the formalism.

For example,
```

More formally, for all result_index in index_space(TensorType(output_shape, si64)),

output[result_index] = constant(is_quantized(output) ?
quantize(result_index[iota_dimension], element_type(output)) :
result_index[iota_dimension], element_type(output))

```

docs/spec.md
```diff
@@ -2633,6 +2633,51 @@ planning to address this in
...
+
+#### Outputs
+
+| Name     | Type                                                                              | Constraints |
+|----------|-----------------------------------------------------------------------------------|-------------|
+| `result` | tensor of integer, floating-point, or complex type or per-tensor quantized tensor |             |
+
+#### Constraints
+
+* (C1) `0 <= iota_dimension < size(output_shape)`.
```
sdasgup3, line 2663: Add
`rank(output) = size(output_shape)`?

abhigunj: good point, added. Thanks.

docs/spec.md
ghpvnist: Btw `TensorType` is not needed since it is also defined on `Value` (`output_shape` is a `Value`):
```
Fills a `result` tensor with values in increasing order starting from zero
along the `iota_dimension` dimension. `result` shape is provided
dynamically via `output_shape`.

More formally, for all `result_index` in `index_space(output_shape)`,

`output[result_index] = constant(is_quantized(output) ?
quantize(result_index[iota_dimension], element_type(output)) :
result_index[iota_dimension], element_type(output))`.
```

abhigunj: Thanks folks, discussion was helpful to get more insight on formal calculations. Updated.

sdasgup3 (approved):

docs/spec.md
```diff
@@ -333,7 +333,7 @@ in StableHLO programs. In the meanwhile, here is the list of these operations:
 * "Dynamism" category of StableHLO operations - they were bootstrapped from
    MHLO, but we haven't specced them yet: `compute_reshape_shape`,
```
GleasonK, line 334: "in the process of speccing them:"

ghpvnist (changes requested):
`status.md` still needs to be updated.

docs/spec.md
```diff
@@ -2633,6 +2633,54 @@ planning to address this in
 
 &nbsp;[More Examples](https://github.com/openxla/stablehlo/tree/main/stablehlo/tests/interpret/dot_general.mlir)
 
+### dynamic_iota
```
GleasonK, line 2636: Modify: `stablehlo/dialect/TypeInference.cpp`

with things like `// dynamic_iota_c1` etc

docs/spec.md
```diff
@@ -2633,6 +2633,54 @@ planning to address this in
...
+
+#### Semantics
```
GleasonK, line 2638: If the corresponding parts of the ODS and StablehloOps.cpp match the spec, check that the "Verification" and "Type Inference" columns in [status.md](https://github.com/openxla/stablehlo/blob/main/docs/status.md) say "yes".

docs/spec.md
```diff
@@ -2633,6 +2633,54 @@ planning to address this in
...
+#### Constraints
+
+* (C1) `0 <= iota_dimension < size(output_shape)`.
+* (C2) `rank(result) = size(output_shape)`.
+
+#### Examples
+
+```mlir
+
+// %output_shape = [4, 5]
```
GleasonK, line 2672: Lets use a `stablehlo.constant` for this example

```
%output_shape = stablehlo.constant dense<[4, 5]> : tensor<2xi64>
```

docs/spec.md
```diff
@@ -2633,6 +2633,54 @@ planning to address this in
 
 &nbsp;[More Examples](https://github.com/openxla/stablehlo/tree/main/stablehlo/tests/interpret/dot_general.mlir)
 
+### dynamic_iota
+
+#### Semantics
+
+Fills a `result` tensor with values in increasing order starting from zero
```
GleasonK, line 2640: Fix link in StablehloOps.td to point at dynamic_iota:

`https://github.com/openxla/stablehlo/blob/main/docs/spec.md#iota`

GleasonK:
```
Check that the description in the op's ODS:
Includes the first sentence of the spec.
Then links to the corresponding section of the spec.
Then uses the same example as the spec but via pretty syntax which can be obtaining by running stablehlo-opt.
```

docs/spec.md
```diff
@@ -2633,6 +2633,54 @@ planning to address this in
...
+along the `iota_dimension` dimension. `result` shape is provided
+dynamically via `output_shape`.
+
+More formally, for all `result_index` in `index_space(output_shape)`,
```
GleasonK, line 2644: +1 to @ghpvnist comment on ensure all criteria in [spec_checklist.md](https://github.com/openxla/stablehlo/blob/main/docs/spec_checklist.md) is met.

GleasonK (changes requested):

abhigunj:
talked to @GleasonK , going to take care of implementation changes as part of this PR to keep all changes together. Will re-request approval once PR is ready.

ghpvnist:
Sounds good. Fyi interpreter can't be implemented atm, but everything else is fair game.

GleasonK:
> Fyi interpreter can't be implemented atm

Any reason for this? Could work with static shapes right?

docs/spec.md
```diff
@@ -2633,6 +2633,56 @@ planning to address this in
 
 &nbsp;[More Examples](https://github.com/openxla/stablehlo/tree/main/stablehlo/tests/interpret/dot_general.mlir)
 
+### dynamic_iota
+
+#### Semantics
+
+Fills a `result` tensor with values in increasing order starting from zero
+along the `iota_dimension` dimension. `result` shape is provided
+dynamically via `output_shape`.
```
ghpvnist, line 2642: I think it would be nice to include a link to the static version of the spec here:
```
...
along the `iota_dimension` dimension. `result` shape is provided
dynamically via `output_shape`. See [`iota`](#iota).
```

Wdyt?

ghpvnist (changes requested):

docs/spec.md
abhigunj: agree, updated

stablehlo/tests/interpret/dynamic_iota.mlir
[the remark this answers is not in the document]
abhigunj: kept two tests for dim0 and dim1. Removed remaining.

stablehlo/dialect/StablehloOps.td
```diff
@@ -124,20 +124,27 @@ def StableHLO_IotaOp : StableHLO_Op<"iota", [Pure]> {
...
+    %output_shape = stablehlo.constant dense<[4, 5]> : tensor<2xi64>
+    %0 = stablehlo.dynamic_iota %output_shape, dim = 0 : (tensor<2xi64>) -> tensor<4x5xi64>
     ```
   }];
 
-  let arguments = (ins HLO_StaticDimensionTensor:$output_shape, I64Attr:$iota_dimension);
+  let arguments = (ins
+    I64Attr:$iota_dimension /*dynamic_iota_i1*/,
+    HLO_StaticDimensionTensor:$output_shape /*dynamic_iota_i2*/
+  );
```
mlevesquedion, line 147: Why this change? StableHLO ops generally have their operands before their attributes.

docs/spec.md
```diff
@@ -2633,6 +2633,57 @@ planning to address this in
...
+Fills a `result` tensor with values in increasing order starting from zero
+along the `iota_dimension` dimension. This operation does the same thing
+as [iota](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#iota)
+op except that the result shape is specified dynamically via `output_shape`.
+
+More formally, for all `result_index` in `index_space(output_shape)`,
+
+`result[result_index] = constant(is_quantized(result) ?
+quantize(result_index[iota_dimension], element_type(result)) :
+result_index[iota_dimension], element_type(result))`.
```
mlevesquedion, line 2649: IMO we should avoid duplication if possible. IMO we should only keep:

```
This operation does the same thing
as [iota](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#iota)
op except that the result shape is specified dynamically via `output_shape`.
```

stablehlo/dialect/StablehloOps.td
ghpvnist: My [comment](https://github.com/openxla/stablehlo/pull/2270#discussion_r1589625143) wasn't clear. Michael's right; to clarify, `output_shape` should come first in the spec, that way, the `i1` and `i2` are also in order.

docs/spec.md
sdasgup3: makes sense. Let's simplify as much as possible.

ghpvnist: I agree. I'm not fond of the "does the same thing" wording in the spec. Perhaps:

```
This operation is functionally identical to
[iota](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#iota)
op, but the result shape is specified dynamically via `output_shape`.
```

abhigunj: Agree. Thank you. 
correct conclusion at our previous discussion https://github.com/openxla/stablehlo/pull/2270#discussion_r1585557343 is : formal calculations for dynamic_iota op is feasible but we can omit from the spec to avoid duplication.

stablehlo/dialect/StablehloOps.td
abhigunj: reordered operand before attribute. Thank you.

ghpvnist (approved):
LGTM. Since the label changed, consider updating the PR description too.

GleasonK (approved):

docs/spec.md
```diff
@@ -2633,6 +2633,50 @@ planning to address this in
...
+* (C2) `rank(result) = size(output_shape)`.
+
+#### Examples
+
+```mlir
+
+%output_shape = stablehlo.constant dense<[4, 5]> : tensor<2xi64>
+%result = "stablehlo.dynamic_iota"(%output_shape) {
+  iota_dimension = 0 : i64
+} : (tensor<2xi64>) -> tensor<4x5xi64>
```
abhigunj, line 2669: do we need an example with dynamic shapes (result  tensor<?x?xi64>) ? 
reference interpreter doesn't support dynamic shapes for now,  I used static shapes in example.

cc @sdasgup3 @ghpvnist

mlevesquedion: IMO that would be preferable, since that's how the op is meant to be used. IMO the example does not need to be something the interpreter supports, and when the interpreter is updated to handle dynamic shapes we are unlikely to revisit this example.

ghpvnist: Let's keep the static example for now to keep docs in sync with the reference interpreter. Once dynamic shapes are supported, we can update the spec with dynamic example and then add a test for it. We can consider keeping the static version as an extra example in the test.

sdasgup3: My 2cents:

With different dynamic ops lined up for specification, I think it would be better to spec the op with intended examples. IMO it is OK for the the spec test not to follow interpreter guidelines (1 and 4 below) before dynamic op interpretation is supported. 

For example, From https://github.com/openxla/stablehlo/blob/main/docs/reference_checklist.md (7)

```
1 Add a link to stablehlo/tests/interpret/<op_mnemonic>.mlir to the "Examples" section (e.g. [More Examples](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#add)).
2 Make sure the spec only has 1 example.
3 Make sure the spec example follows the [testing guidelines](https://github.com/openxla/stablehlo/blob/main/docs/reference.md#testing-guidelines).
4 Make sure the spec example test is interpretable.
5 Make sure the spec example is the same as what is in the ODS.
``` 

1. No "More examples" link as  as dynamic interpretation is not fully supported yet. 
4. Is it fine that we cannot do that. IMO we should not write an example just so that  (4) is honored.  


With that once the dynamic interpretation is supported we have one artifact to update (the interpreter test cases) as opposed to multiple (spec, ODS).

stablehlo/reference/Ops.cpp
```diff
@@ -1576,6 +1581,14 @@ Tensor dotGeneralOp(const Tensor &lhs, const Tensor &rhs,
   return result;
 }
 
+Tensor dynamicIotaOp(Axis iotaDimension, const Tensor &outputShape,
+                     ShapedType resultType) {
+  if (resultType.hasStaticShape()) return iotaOp(iotaDimension, resultType);
```
GleasonK, line 1586: Followup: Let's move this error check to the top of the function, check all operands and all results. Make a ticket for now.

abhigunj: tracking at https://github.com/openxla/stablehlo/issues/2294

docs/spec.md
abhigunj: Thanks folks, as per offline discussion, keeping the static shape example.

PR #2272, opened 2024-05-01
iota op spec :  replace "result_index" with "output_index"

abhigunj:
iota op spec doesn't define `result` but uses `result_index`.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -3312,9 +3312,9 @@ separate outputs to improve clarity
 Fills an `output` tensor with values in increasing order starting from zero
 along the `iota_dimension` dimension. More formally,
 
-`output[result_index] = constant(is_quantized(output) ?
-quantize(result_index[iota_dimension], element_type(output)) :
-result_index[iota_dimension], element_type(output))`.
+`output[output_index] = constant(is_quantized(output) ?
+quantize(output_index[iota_dimension], element_type(output)) :
+output_index[iota_dimension], element_type(output))`.
 
 #### Inputs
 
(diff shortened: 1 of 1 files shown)


ghpvnist (approved):

mlevesquedion (changes requested):
I think we actually want `result`, not `output`? Indeed, we use `result` for most op outputs.

Can you rename the other occurrences of `output` as well? (e.g. constant op, reduce_precision, etc.)

Actually, TBH I'm not sure we should do this change right now given that we would also need to change the ODS to be consistent with the spec, but that will break C++ APIs.

abhigunj:
> I think we actually want `result`, not `output`? Indeed, we use `result` for most op outputs.
> 
> Can you rename the other occurrences of `output` as well? (e.g. constant op, reduce_precision, etc.)
> 
> Actually, TBH I'm not sure we should do this change right now given that we would also need to change the ODS to be consistent with the spec, and that will break C++ APIs.

Yeh, eventually we need `result`. for time being, better to be consistent with other part of the iota spec. It confused me initially because spec doesn't define `result` but uses `result_index`.

abhigunj:
> > I think we actually want `result`, not `output`? Indeed, we use `result` for most op outputs.
> > Can you rename the other occurrences of `output` as well? (e.g. constant op, reduce_precision, etc.)
> > Actually, TBH I'm not sure we should do this change right now given that we would also need to change the ODS to be consistent with the spec, and that will break C++ APIs.
> 
> Yeh, eventually we need `result`. for time being, better to be consistent with other part of the iota spec. It confused me initially because spec doesn't define `result` but uses `result_index`.

https://github.com/openxla/stablehlo/issues/351 tracking the name changes which will change `output` to `result`

mlevesquedion (approved):

Issue #2121, opened 2024-03-22
Generic support for arrays in VHLO


mlevesquedion:
### Request description

Array types are used sparingly in StableHLO, mostly in the form of `DenseI64ArrayAttr`, altough there are also uses of `DenseBoolArrayAttr` and `ArrayAttr`. Support for array types in serialization/deserialization is limited.

`ArrayAttr` is handled in a [generic manner](https://github.com/openxla/stablehlo/blob/2c82fb4633e5f0c7ce21763a38ebf31989b134f2/stablehlo/transforms/StablehloLegalizeToVhlo.cpp#L129-L137), i.e. both registered and unregistered array attributes will round trip through VHLO.

However, `DenseI64ArrayAttr` and `DenseBoolArrayAttr` are supported in a specific way. [They are serialized using `vhlo::TensorV1Attr`](https://github.com/openxla/stablehlo/blob/2c82fb4633e5f0c7ce21763a38ebf31989b134f2/stablehlo/transforms/StablehloLegalizeToVhlo.cpp#L145-L166), which means that they have to be [converted back to arrays when they are deserialized](https://github.com/openxla/stablehlo/blob/2c82fb4633e5f0c7ce21763a38ebf31989b134f2/stablehlo/transforms/VhloLegalizeToStablehlo.cpp#L519-L585). This is handled with op type switches, so there is no generic support: arrays that occur outside of the specific list of handled attributes will deserialize into tensors rather than arrays:

```
$ stablehlo-translate --serialize --target=current <<<"func.func @foo() {
  stablehlo.custom_call @foo() {foo = array<i64: 1, 2, 3>} : () -> ()
  func.return
}" | stablehlo-translate --deserialize
func.func @foo() {
  stablehlo.custom_call @foo() {foo = dense<[1, 2, 3]> : tensor<3xi64>} : () -> ()
  return
}
```

These are other kinds of array attributes, such as `DenseI32ArrayAttr`, which are currently not supported:

```
$ stablehlo-translate --serialize --target=current <<<"func.func @foo() {
  stablehlo.custom_call @foo() {foo = array<i32: 1, 2, 3>} : () -> ()
  func.return
}" | stablehlo-translate --deserialize
...
Failed to convert: array<i32: 1, 2, 3>
<stdin>:2:3: error: failed to legalize operation 'stablehlo.custom_call' that was explicitly marked illegal
  stablehlo.custom_call @foo() {foo = array<i32: 1, 2, 3>} : () -> ()
[...]

PR #2274, opened 2024-05-01
Add support for SymbolRef in VHLO

GleasonK:
This is needed for arbitrary support of `@ref` in VHLO. This will enable composite operations to denote more called computations to perhaps model an op with a region.

This also reduces the complexity of the conversion, with all of the upgrade/downgrade cruft able to be cleaned up after the 6mo compatibility window passes.

Currently in VHLO function refs are strings. There's not a great reason for this, and it makes sense for a language spec to denote the difference between a string that refers to a symbol and a string that refers to arbitrary data.

Part of #2121.


stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.mlir
```diff
@@ -434,7 +434,7 @@ func.func @default_compare(%arg0: tensor<f32>, %arg1: tensor<f32>) -> tensor<i1>
 func.func @default_composite(%arg0: tensor<f32>) -> tensor<f32> {
   //               CHECK: "vhlo.composite_v1"(%arg0) <{
   //          CHECK-SAME:   composite_attributes = #vhlo.dict_v1<{}>
-  //          CHECK-SAME:   decomposition = #vhlo.string_v1<"composite_target">
+  //          CHECK-SAME:   decomposition = #vhlo.symbolref_v1<"composite_target">
```
GleasonK, line 437: Still need to add a versioned test for `.0_20_0.mlir`.

stablehlo/dialect/VhloDialect.td
```diff
@@ -37,6 +37,7 @@ def VHLO_Dialect : Dialect {
       0.17.0: Allow reduce operations to promote to higher bitwidth.
       0.18.0: Introduce `UniformQuantizedPerAxisType` type.
       0.19.0: Introduce `composite` operation.
+      0.20.0: Introduce `symbolref` attribute.
```
GleasonK, line 40: Also plan to edit `spec.md` to denote the difference between a string and a function ref.

stablehlo/dialect/VhloOps.cpp
```diff
@@ -365,5 +367,28 @@ LogicalResult SelectAndScatterOpV1::validateConstraint(mlir::Operation* op,
   return verifyConstraint_0_17_0(op, targetVersion);
 }
 
+LogicalResult CustomCallOpV1::validateConstraint(mlir::Operation* op,
+                                                 Version targetVersion) {
+  if (targetVersion < Version(0, 20, 0)) return success();
+  if (auto array = dyn_cast<ArrayV1Attr>(getCalledComputations())) {
```
mlevesquedion, line 373: I think this can be a `cast`?

stablehlo/dialect/VhloOps.cpp
```diff
@@ -365,5 +367,28 @@ LogicalResult SelectAndScatterOpV1::validateConstraint(mlir::Operation* op,
   return verifyConstraint_0_17_0(op, targetVersion);
 }
 
+LogicalResult CustomCallOpV1::validateConstraint(mlir::Operation* op,
+                                                 Version targetVersion) {
+  if (targetVersion < Version(0, 20, 0)) return success();
```
mlevesquedion, line 372: Uber-nit: consider adding a small helper `isVersionBeforeSymbolRefV1AttrIntroduced` (ideally, something less verbose), or creating an appropriately-named global constant for `Version(0, 20, 0)` to avoid having a magic constant duplicated in a few places.

mlevesquedion: More generally, I think it would be a good idea to have some place where minor versions are explicitly tagged with a short description of the relevant changes. WDYT?

stablehlo/transforms/VhloToVersion.cpp
```diff
@@ -296,15 +297,146 @@ struct VersionConversionPattern : OpConversionPattern<SourceOp> {
...
+      SourceAttr element = dyn_cast<SourceAttr>(attr);
+      if (!element) return failure();
+      elements.push_back(element);
+    }
+    return elements;
+  }
+
+  LogicalResult convertAttribute(ConversionPatternRewriter& rewriter, OpType op,
+                                 StringRef attrName) const {
+    Attribute attr = op->getAttr(attrName);
```
mlevesquedion, line 326: Check if the attribute was found?

stablehlo/transforms/VhloToVersion.cpp
```diff
@@ -296,15 +297,146 @@ struct VersionConversionPattern : OpConversionPattern<SourceOp> {
...
+
+    // Support the following:
+    //   Array<SourceAttr> -> Array<TargetAttr>
+    //   SourceAttr -> TargetAttr
+    if (auto array = matchArrayOfSource(attr); succeeded(array)) {
```
mlevesquedion, line 331: This is fine, but I feel like there should be some built-in way to tell MLIR that this attribute contains other attributes, and have MLIR do the visiting as needed? I don't know of any such facility though.

mlevesquedion (commented):
LGTM, just some ideas. Holding off on the `0.20.0` test file for formal approval. Consider marking the PR as "draft" in the meantime.

stablehlo/transforms/VhloToVersion.cpp
```diff
@@ -296,15 +297,146 @@ struct VersionConversionPattern : OpConversionPattern<SourceOp> {
...
+template <typename OpType, typename SourceAttr, typename TargetAttr>
+struct AttributeConversionPattern : public OpConversionPattern<OpType> {
+  using OpConversionPattern<OpType>::OpConversionPattern;
+  virtual SmallVector<StringRef> getAttributeNames(OpType op) const = 0;
+
+  FailureOr<SmallVector<SourceAttr>> matchArrayOfSource(Attribute attr) const {
+    vhlo::ArrayV1Attr arrayAttr = dyn_cast<ArrayV1Attr>(attr);
+    if (!arrayAttr) return failure();
+    SmallVector<SourceAttr> elements;
+    for (Attribute attr : arrayAttr.getValue()) {
```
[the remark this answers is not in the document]
GleasonK: ^from discussion with Michael

GleasonK:
Going to defer this change, closing PR for now.

Issue #2202, opened 2024-04-11
Request Interpreter to support dynamic shape basicly


qingyunqu:
### Request description

I find that Interperter doesn't support dynamic shape currently, like:
```python
from stablehlo import ir
from stablehlo._mlir_libs._stablehlo import register_dialect, eval_module

module_str = """
func.func @main(%arg0: tensor<?xf32>, %arg1: tensor<2xi64>) -> tensor<?x?xf32> {
    %0 = stablehlo.dynamic_reshape %arg0, %arg1 : (tensor<?xf32>, tensor<2xi64>) -> tensor<?x?xf32>
    return %0 : tensor<?x?xf32>
}
"""

context = ir.Context()
register_dialect(context)
with context:
    module = ir.Module.parse(module_str)
    print(module.operation.get_asm())

with context:
    a = ir.DenseFPElementsAttr(ir.Attribute.parse("dense<[0.0, 1.0, 2.0, 3.0]> : tensor<4xf32>"))
    b = ir.DenseIntElementsAttr(ir.Attribute.parse("dense<[2, 2]> : tensor<2xi64>"))
    print(a)

c = eval_module(module, [a, b])
print(c)
```

Is there any timeline to accomplish this?

qingyunqu:
We'd be happy to help expedite this.

sdasgup3:
Hi @qingyunqu 
My plan of implementation here would be to use augment the interpreter with shape refinement of dynamic arguments of the program with static shapes before interpreting the program. That should be good to address this request: The dynamic_shapes should all be folded to their static counterpart after [StablehloCanonicalizeDynamism](https://github.com/openxla/stablehlo/blob/main/stablehlo/transforms/StablehloCanonicalizeDynamism.cpp) pass. 

On a separate note:
Currently there is no support for interpreting dynamic_* stablehlo ops as such, so feel free to add that support.

PR #2275, opened 2024-05-01
Interpreter support for dynamic programs

sdasgup3:
This PR adds a pass-pipeline to refine arguments and shapes of a program followed by [StablehloCanonicalizeDynamism](https://github.com/openxla/stablehlo/blob/main/stablehlo/transforms/StablehloCanonicalizeDynamism.cpp) to replace dynamic StableHLO ops with the corresponding static counterparts if applicable* (see note below).

The goal is to interpret dynamic programs after applying the above pipeline.

partially-fixes https://github.com/openxla/stablehlo/issues/2202

Note:

For a programs like

```
func.func @main(%arg0: tensor<?xf32>, %arg1: tensor<2xi64>) -> tensor<?x?xf32> {
    %0 = stablehlo.dynamic_reshape %arg0, %arg1 : (tensor<?xf32>, tensor<2xi64>) -> tensor<?x?xf32>
    return %0 : tensor<?x?xf32>
}
```

when applied through the above pass-pipeline with  static input shapes like `tensor<6xf32>,tensor<2xi64>`, will not be able to replace  `stablehlo.dynamic_reshape`  with its static counterpart `reshape` as [StablehloCanonicalizeDynamism](https://github.com/openxla/stablehlo/blob/main/stablehlo/transforms/StablehloCanonicalizeDynamism.cpp) can only do that when the second argument of `stablehlo.dynamic_reshape`  is constant, which is not true in this case.

### DIFF
--- a/stablehlo/integrations/python/tests/stablehlo.py
+++ b/stablehlo/integrations/python/tests/stablehlo.py
@@ -251,6 +251,8 @@ def test_reference_api():
     # Numpy's uint treated as int by DenseElementsAttr, skipping np.uint tests
     ("2x2xf16", np.asarray([1, 2, 3, 4], np.float16).reshape(2,2)),
     ("2x1x2xf16", np.asarray([1, 2, 3, 4], np.float16).reshape(2,1,2)),
+    ("?x?xf16", np.asarray([1, 2, 3, 4], np.float16).reshape(2,2)),
+    ("?x2xf16", np.asarray([1, 2, 3, 4], np.float16).reshape(2,2)),
   ]
   for test in tests:
     tensor_type, arg = test
--- a/stablehlo/reference/Api.cpp
+++ b/stablehlo/reference/Api.cpp
@@ -30,6 +30,7 @@ limitations under the License.
 #include "mlir/IR/Location.h"
 #include "mlir/IR/TypeRange.h"
 #include "mlir/Parser/Parser.h"
+#include "mlir/Pass/PassManager.h"
 #include "mlir/Support/LogicalResult.h"
 #include "stablehlo/dialect/Register.h"
 #include "stablehlo/reference/Configuration.h"
@@ -41,6 +42,7 @@ limitations under the License.
 #include "stablehlo/reference/Scope.h"
 #include "stablehlo/reference/Tensor.h"
 #include "stablehlo/reference/Value.h"
+#include "stablehlo/transforms/Passes.h"
 
 namespace mlir {
 namespace stablehlo {
@@ -136,6 +138,39 @@ LogicalResult validateEntrySignature(func::FuncOp func,
   return success();
 }
 
+// Specializes the shapes of arguments in function 'func' based on runtime input
+// values. If all argument types already have static shapes, this function does
+// nothing. Otherwise, it constructs a pipeline of MLIR passes to refine
+// argument shapes using the provided `inputs`.
+//
+// Args:
+//   module: The MLIR module containing the function.
+//   func: The function whose argument shapes need specialization.
+//   inputs: The runtime input values.
+//
+// Returns:
+//   A `LogicalResult` indicating success or failure of the shape
+//   refinement pipeline.
+LogicalResult refinePolymorphicModule(ModuleOp module, func::FuncOp func,
+                                      ArrayRef<InterpreterValue> inputs) {
+  if (llvm::all_of(func.getArgumentTypes(), [](Type type) {
+        return llvm::cast<ShapedType>(type).hasStaticShape();
+      })) {
+    return success();
+  }
+
+  SmallVector<Type> refinedTypes = llvm::to_vector(llvm::map_range(
+      inputs, [](InterpreterValue input) { return input.getType(); }));
+
+  PassManager pm(module.getContext());
+  stablehlo::createStablehloRefinePolymorphicModule(pm, refinedTypes);
+  if (failed(pm.run(module))) {
+    return func.emitError("Failed to refine dynamic shape in function: ")
+           << func.getName();
+  }
+  return success();
+}
+
 }  // namespace
 
 FailureOr<SmallVector<InterpreterValue>> evalModule(
@@ -148,8 +183,11 @@ FailureOr<SmallVector<InterpreterValue>> evalModule(
     return SmallVector<InterpreterValue>();
 
   auto mainFunc = getMainFunction(module, config.mainFunction);
-  if (failed(mainFunc) || failed(validateEntrySignature(*mainFunc, inputs)))
+  if (failed(mainFunc) ||
+      failed(refinePolymorphicModule(module, *mainFunc, inputs)) ||
+      failed(validateEntrySignature(*mainFunc, inputs))) {
     return failure();
+  }
 
   if (!config.probeInstrumentationDir.empty()) {
     llvm::SmallString<128> instrumentationMetadataFile(
--- a/stablehlo/reference/Api.cpp
+++ b/stablehlo/reference/Api.cpp
@@ -151,8 +151,8 @@ LogicalResult validateEntrySignature(func::FuncOp func,
 // Returns:
 //   A `LogicalResult` indicating success or failure of the shape
 //   refinement pipeline.
-LogicalResult refinePolymorphicModule(ModuleOp module, func::FuncOp func,
-                                      ArrayRef<InterpreterValue> inputs) {
+LogicalResult removeDynamism(ModuleOp module, func::FuncOp func,
+                             ArrayRef<InterpreterValue> inputs) {
   if (llvm::all_of(func.getArgumentTypes(), [](Type type) {
         return llvm::cast<ShapedType>(type).hasStaticShape();
       })) {
@@ -163,7 +163,7 @@ LogicalResult refinePolymorphicModule(ModuleOp module, func::FuncOp func,
       inputs, [](InterpreterValue input) { return input.getType(); }));
 
   PassManager pm(module.getContext());
-  stablehlo::createStablehloRefinePolymorphicModule(pm, refinedTypes);
+  stablehlo::createStablehloRemoveDynamismPipeline(pm, refinedTypes);
   if (failed(pm.run(module))) {
     return func.emitError("Failed to refine dynamic shape in function: ")
            << func.getName();
@@ -183,8 +183,7 @@ FailureOr<SmallVector<InterpreterValue>> evalModule(
     return SmallVector<InterpreterValue>();
 
   auto mainFunc = getMainFunction(module, config.mainFunction);
-  if (failed(mainFunc) ||
-      failed(refinePolymorphicModule(module, *mainFunc, inputs)) ||
+  if (failed(mainFunc) || failed(removeDynamism(module, *mainFunc, inputs)) ||
       failed(validateEntrySignature(*mainFunc, inputs))) {
     return failure();
   }
--- a/stablehlo/reference/Api.cpp
+++ b/stablehlo/reference/Api.cpp
@@ -68,7 +68,7 @@ FailureOr<func::FuncOp> getMainFunction(ModuleOp module, StringRef mainName) {
 class DefaultInterpreterFallback : public InterpreterFallback {
  public:
   DefaultInterpreterFallback(const InterpreterConfiguration &config)
-      : config(config) {};
+      : config(config){};
 
   virtual llvm::Error operator()(Operation &op, Scope &scope,
                                  Process *process) final {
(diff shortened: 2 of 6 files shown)


stablehlo/transforms/Passes.h
```diff
@@ -86,6 +86,17 @@ std::unique_ptr<OperationPass<ModuleOp>> createStablehloRefineArgumentsPass(
...
 
+// Creates a pipeline of StableHLO-specific MLIR passes to refine
+// argument types and shapes within an MLIR module.
+//
+// This pipeline focuses on:
+//   1. Refining function argument types based on provided `refinedTypes`.
+//   2. Refining shape information of operations within functions.
+//   3. Replaces dynamic StableHLO ops with the corresponding static
+//   counterparts if applicable.
+void createStablehloRefineArgumentsAndShapesPipeline(OpPassManager &pm,
```
GleasonK, line 97: "refines the `main` functions arguments and propagates new shapes throughout the program" ?

GleasonK: Doesn't mention what function is used currently.

stablehlo/integrations/python/tests/stablehlo.py
```diff
@@ -251,6 +251,7 @@ def test_reference_api():
     # Numpy's uint treated as int by DenseElementsAttr, skipping np.uint tests
     ("2x2xf16", np.asarray([1, 2, 3, 4], np.float16).reshape(2,2)),
     ("2x1x2xf16", np.asarray([1, 2, 3, 4], np.float16).reshape(2,1,2)),
+    ("?x?xf16", np.asarray([1, 2, 3, 4], np.float16).reshape(2,2)),
```
GleasonK, line 254: How about a `?x2xf16` test case as well

stablehlo/reference/Api.cpp
```diff
@@ -137,6 +139,41 @@ LogicalResult validateEntrySignature(func::FuncOp func,
...
+//   module: The MLIR module containing the function.
+//   func: The function whose argument shapes need specialization.
+//   inputs: The runtime input values.
+//
+// Returns:
+//   A `LogicalResult` indicating success or failure of the shape
+//   refinement pipeline.
+//
+// Assumption: 'func' is already validated using validateEntrySignature.
+LogicalResult shapeSpecialize(ModuleOp module, func::FuncOp func,
```
GleasonK, line 157: `refinePolymorphicModule` maybe?

GleasonK (approved):
Looks great!

stablehlo/integrations/python/tests/stablehlo.py
sdasgup3: done

stablehlo/reference/Api.cpp
sdasgup3: done

stablehlo/transforms/PassPipelines.cpp
```diff
@@ -28,6 +28,14 @@ void createStablehloDeserializePipeline(OpPassManager &pm) {
   pm.addPass(stablehlo::createVhloLegalizeToStablehloPass());
 }
 
+void createStablehloRefineArgumentsAndShapesPipeline(OpPassManager &pm,
```
mlevesquedion, line 31: I feel like this name is a bit too specific, but also does not fully describe what the pipeline is for. To my understanding, the ultimate goal here is to remove dynamism. So maybe something like `createStablehloRemoveDynamismPipeline`? And the corresponding function in `Api.cpp` could be called `removeDynamism`?

stablehlo/reference/Api.cpp
```diff
@@ -137,6 +139,39 @@ LogicalResult validateEntrySignature(func::FuncOp func,
...
+        return llvm::cast<ShapedType>(type).hasStaticShape();
+      })) {
+    return success();
+  }
+
+  SmallVector<Type> refinedTypes = llvm::to_vector(llvm::map_range(
+      inputs, [](InterpreterValue input) { return input.getType(); }));
+
+  PassManager pm(module.getContext());
+  stablehlo::createStablehloRefineArgumentsAndShapesPipeline(pm, refinedTypes);
```
mlevesquedion, line 167: What happens if the number of function arguments does not match the number of shapes? Consider adding a test case for that. I think we should probably validate that before applying any passes.

stablehlo/transforms/Passes.h
```diff
@@ -86,6 +86,20 @@ std::unique_ptr<OperationPass<ModuleOp>> createStablehloRefineArgumentsPass(
...
+// function's arguments and propagates new shapes throughout the program
+// argument types and shapes within an MLIR module. The main function is either
+// a function with name "main", if there are multiple functions, or the single
+// function within the module.
+//
+// This pipeline focuses on:
+//   1. Refining function argument types based on provided `refinedTypes`.
+//   2. Refining shape information of operations within functions.
+//   3. Replaces dynamic StableHLO ops with the corresponding static
+//   counterparts if applicable.
```
mlevesquedion, line 99: I think this is a good description, but consider opening with a direct description of the goal: to remove dynamism from the program.

mlevesquedion (commented):
This is great stuff! Just a few comments, PTAL.

stablehlo/transforms/Passes.h
sdasgup3: done

stablehlo/reference/Api.cpp
sdasgup3: The first pass in the pipeline has the check for that https://github.com/openxla/stablehlo/blob/e2201e2af69c2b0d32370cfe0deafe2681c7b3e9/stablehlo/transforms/StablehloRefineArguments.cpp#L93 and we have tests available as well. Please let me know if that would suffice. 
 
I note that [validateEntrySignature](https://github.com/openxla/stablehlo/blob/e2201e2af69c2b0d32370cfe0deafe2681c7b3e9/stablehlo/reference/Api.cpp#L120) in APIs.cpp is already doing that check, but I reorder it to be after the the shape specialization mainly because that function deals with static shapes only and with that it is very easy to match types using `a == b` as opposed to do compatibility check ignoring dynamic dimensions for all supported interpreter `Tensor` types (Tuples and Tenor).

mlevesquedion: That sounds reasonable. I'm just wondering what the error message will look like and if it will be clear to the user what the issue is?

LGTM about `validateEntrySignature`. Earlier on I think some comment mentioned that check ran first, which I think couldn't have worked because the types would have differed. Doing the check after removing dynamism makes sense to me. Thanks!

stablehlo/transforms/Passes.h
sdasgup3: done

stablehlo/reference/Api.cpp
sdasgup3: Fair point. Here is an example.
https://github.com/openxla/stablehlo/blob/e2201e2af69c2b0d32370cfe0deafe2681c7b3e9/stablehlo/tests/stablehlo_refine_parameters.mlir#L39

stablehlo/transforms/PassPipelines.cpp
sdasgup3: Thanks for the suggestion!

I agree with the point that the name is too specific and does not convey the overall goal. Thanks!

Given that the pass-pipeline can only remove dynamism in certain cases (e.g. as mentioned in the description there canse dynamic_ops* which cannot be removed just by using the pipeline), with that my impression is that the name would over-commit to what can be achieved.

What do you think about 
`createStablehloRefinePolymorphicModule` it has what it does?

stablehlo/reference/Api.cpp
mlevesquedion: This looks fine, I just wonder if it will make sense to someone who's just using the interpreter and has no knowledge of what's going on under the hood. IMO it would be preferable to do this check before running the pipeline as that would allow us to produce a more helpful error message.

Up to you though, I think what we have now is good enough.

stablehlo/transforms/PassPipelines.cpp
mlevesquedion: I think it's okay to "over-commit": most of our passes do it, including the passes that are included in this pipeline. "refine-shapes" can leave some shapes unrefined. "canonicalize-dynamism" can fail to remove some dynamic ops. More generally, passes are named after their intent, but they rarely promise anything, e.g. the CSE and DCE passes in MLIR do their best, but they can't possible handle everything (since it's undecidable anyway).

Ideally, we would have checks at the end of these passes such that if we fail to refine all shapes we fail the pass, and similarly if we fail to remove all dynamic ops we fail the pass. Currently, I believe dynamic shapes/ops can still make it to the interpreter, which will cause obtuse errors?

mlevesquedion: > What do you think about
createStablehloRefinePolymorphicModule it has what it does?

I don't love the name "refine polymorphic module", because the term "polymorphism" is quite overloaded, and I feel like it doesn't really capture the intent of the pass. I don't think it's a huge deal, but IMO `removeDynamism` is a better name.

stablehlo/reference/Api.cpp
sdasgup3: Sure. Lets check is in as in and we can refine later.

stablehlo/transforms/PassPipelines.cpp
sdasgup3: Fair point. Thanks!
 
>> Ideally, we would have checks at the end of these passes such that if we fail to refine all shapes we fail the pass, and similarly if we fail to remove all dynamic ops we fail the pass. 

Agreed.

>>  Currently, I believe dynamic shapes/ops can still make it to the interpreter, which will cause obtuse errors?

That is right. With dynamic_ops it is op not supported by interpreter and dynamic_shapes which will fail on type mismatch with interpreterValue whle allocating resources for result type. 


>> I don't love the name "refine polymorphic module", because the term "polymorphism" i
I agree with some of it. The shape qualifier makes it less ambiguous. 

In any-case, I am fine with both option. Using `removeDyanmism` now.

sdasgup3: PTAL.

stablehlo/reference/Api.cpp
```diff
@@ -137,6 +139,39 @@ LogicalResult validateEntrySignature(func::FuncOp func,
...
+// Returns:
+//   A `LogicalResult` indicating success or failure of the shape
+//   refinement pipeline.
+LogicalResult removeDynamism(ModuleOp module, func::FuncOp func,
+                             ArrayRef<InterpreterValue> inputs) {
+  if (llvm::all_of(func.getArgumentTypes(), [](Type type) {
+        return llvm::cast<ShapedType>(type).hasStaticShape();
+      })) {
+    return success();
+  }
```
mlevesquedion, line 161: Could there be dynamism in other parts of the program? If so, maybe we should still run the pipeline? Likely won't be too costly if it turns out that there is no refinement to be done.

mlevesquedion (approved):

stablehlo/reference/Api.cpp
sdasgup3: Thanks for bringing this. Yup that is possible.

```
func.func @main(%arg0: tensor<3x4xf32>, %arg1: tensor<3x4xf32>) -> tensor<?x?xf32> {
    %0 = stablehlo.add %arg0, %arg1 : (tensor<3x4xf32>, tensor<3x4xf32>) -> tensor<?x?xf32>
    return %0 : tensor<?x?xf32>
}
```

This currently will not be refined. 
The check for eligibility to call the pass-pipiline would be a iteration over the whole graph to check for any traces of dynamism.  That includes identifying ops like dynamic_*op with static operand/result type and const shape operand (This is not usual for someone to use such operations like that but theoretically possible). That would allow us to remove as many traces of dynamism as possible. 

Other  alternative is to call the pipeline regardless given the shape refinement pass already has the logic when to skip the refinement.

Thoughts?

mlevesquedion: Right, checking for dynamism is kind of involved.

I think both approaches are fine, though I would tend to lean towards just calling the pipeline every time.

sdasgup3: done

sdasgup3:
Looking into the failures.

sdasgup3:
Note: Invoking the pass-pipeline unconditionally uncovers a few issues in the folding logic of shape refinement pass. https://github.com/openxla/stablehlo/issues/2278 


Per conversation with @GleasonK,
Let us invoke the pass-pipeline conditionally based on the input arguments being dyanmic. For programs  where dynamism can appear elsewhere, as discussed in https://github.com/openxla/stablehlo/pull/2275#discussion_r1586650963, are  not well formed mostly from the POV of real use-cases. Also, there are different ways to handle "programs  where dynamism can appear elsewhere" which should be explored separately. For example, the cases mentioned at https://github.com/openxla/stablehlo/pull/2275#discussion_r1586650963 can very well be handled while interpreting the program. 

Let us address these cases separately as part of improving the interpreter. A related goal (based on #2278 ) could be for the interpreter and shape-refinement folders to share logic.

stablehlo/reference/Api.cpp
sdasgup3: Note: Invoking the pass-pipeline unconditionally uncovers a few issues in the folding logic of shape refinement pass which can be addressed separately via  https://github.com/openxla/stablehlo/issues/2278

PR #2280, opened 2024-05-02
Use free function instead of method for dyn_cast_or_null

mlevesquedion:
The method variant is deprecated.


ghpvnist (approved):
This change lgtm. I think we have other uses of deprecated casts. Are you tracking them by any chance?

mlevesquedion:
> This change lgtm. I think we have other uses of deprecated casts. Are you tracking them by any chance?

I believe this should be the last one, unless we add more. I grepped the codebase and didn't find anything else.

(This one was added after I removed most occurrences in #2180. Another one was removed in #2260.)

PR #2284, opened 2024-05-03
dynamic_reshape op spec

abhigunj:
* Constraints in the spec:
```
I1 operand is tensor or quantized tensor
I2 output_shape is 1-dimensional tensor constant of type si64
```
* Test coverage;
```
I1.1  operand is tensor or quantized tensor.   (covered by ODS)
I2.1  output_shape is not a 1-dimensional tensor constant of type si64 .   (covered by ODS)
C1.1 if !is_per_axis_quantized(operand),  element_type(result) = element_type(operand).    (Added a verifier logic and a  test)
C1.2 if is_per_axis_quantized(operand),  element_type(result) = element_type(operand) except
       quantization_dimension(operand) and quantization_dimension(result) may differ.    (Validated; but test can't be added at this moment because interpreter doesn't support dynamic shapes) 
C2.1  size(operand) = size(result).  (can't be verified at static time for dynamic ops)
C3.1  quantization constraints.  (verifier logic is common for reshape and dynamic_reshape op. Tests are already present for reshape op. Not needed for dynamic_reshape.)
C4.1  rank(result) = size(output_shape)   (Already present, validated)
```

* Reference interpreter:
partial implementation to support only static shapes.

ref:#2267
fixes:https://github.com/openxla/stablehlo/issues/2293


stablehlo/dialect/StablehloOps.td
```diff
@@ -2656,20 +2656,26 @@ def StableHLO_DynamicReshapeOp: StableHLO_ShapedInterfaceOp<"dynamic_reshape",
       [HLO_SpeculatableIfAllInputsStaticAndShapeConstant, NoMemoryEffect]> {
   let summary = "DynamicReshape operation";
   let description = [{
-    This operation is a work in progress, so it is not yet included in
-    the StableHLO specification: https://github.com/openxla/stablehlo/issues/8.
+    Performs reshape of `operand` tensor to a `result` tensor.
 
     Informally, this operation does the same thing as ReshapeOp except that the
```
ghpvnist, line 2661: We can remove this paragraph.

docs/spec.md
```diff
@@ -2636,6 +2636,57 @@ planning to address this in
...
+
+| Label | Name           | Type                                         | Constraints |
+|-------|----------------|----------------------------------------------|-------------|
+| (I1)  | `operand`      | tensor or quantized tensor                   | (C1-C3)     |
+| (I2)  | `output_shape` | 1-dimensional tensor constant of type `si64` | (C2), (C3)  |
+
+#### Outputs
+
+| Name     | Type                       | Constraints   |
+|----------|----------------------------|---------------|
```
ghpvnist, line 2657: Fix table formatting.

ghpvnist (changes requested):

docs/spec.md
```diff
@@ -2636,6 +2636,57 @@ planning to address this in
...
+| `result` | tensor or quantized tensor | (C1-C3)       |
+
+#### Constraints
+
+* (C1) `element_type(result)` is given by:
+  * `element_type(operand)`, if `!is_per_axis_quantized(operand)`.
+  * `element_type(operand)` except that `quantization_dimension(operand)` and
+    `quantization_dimension(result)` may differ, otherwise.
+* (C2) `size(operand) = size(output_shape) = rank(result)`.
```
sdasgup3, line 2666: `rank(result) = size(output_shape)`

No constraints on size of result.

docs/spec.md
```diff
@@ -2636,6 +2636,57 @@ planning to address this in
...
+* (C3) If `is_per_axis_quantized(operand)`:
```
sdasgup3, line 2667: We cannot have any of the constraints in C3 as `outut_shape` is unknown

GleasonK (commented):
LGTM, please add the remaining spec checklist items then should be good to go

stablehlo/dialect/StablehloOps.td
```diff
@@ -2663,21 +2663,23 @@ def StableHLO_DynamicReshapeOp: StableHLO_ShapedInterfaceOp<"dynamic_reshape",
...
-    %0 = stablehlo.dynamic_reshape %arg0, %shape : (tensor<?xf32>, tensor<2xindex>) -> tensor<?x?xf32>
+    %0 = stablehlo.constant dense<[3, 2]> : tensor<2xi64>
+    %1 = stablehlo.dynamic_reshape %arg0, %0 : (tensor<2x3xi64>, tensor<2xi64>) -> tensor<3x2xi64>
     ```
   }];
 
-  let arguments = (ins HLO_Tensor:$operand, HLO_StaticDimensionTensor:$output_shape);
-  let results = (outs HLO_Tensor:$result);
+  let arguments = (ins
+    HLO_TensorOrPerAxisQuantizedTensor:$operand /*dynamic_reshape_i1*/,
```
GleasonK, line 2679: Do we want per-axis support? Doesn't that require knowing the value of the `?` dim?

cc @sdasgup3

GleasonK: Or are there cases where the mixture of unbounded dynamism and per-axis are actually useful?

stablehlo/dialect/TypeInference.cpp
```diff
@@ -255,6 +255,57 @@ LogicalResult verifyQPerAxisScaleAndZeroPointConstraints(
...
+                                                          resultTy)))
+    return failure();
+
+  // reshape_c1, dynamic_reshape_c1
+  if (failed(verifyQPerAxisScaleAndZeroPointConstraints(location, operandTy,
+                                                        resultTy)))
+    return failure();
+
+  // reshape_c3
+  if (allQuantized<quant::UniformQuantizedPerAxisType>(operandTy, resultTy)) {
```
GleasonK, line 271: Does this current logic allow for per-axis operand but not per-axis result? Is that expected, or is that handled elsewhere?

stablehlo/dialect/StablehloOps.td
sdasgup3: >> Do we want per-axis support? Doesn't that require knowing the value of the ? dim?

Being a dynamic counterpart to the reshape op, I would imagine same quantization constraint on this op.  I have seens a few uses of stablehlo dynamic ops with quantzed element-type in stablehlo quantier https://github.com/tensorflow/tensorflow/blob/5f1eb4d0a5ad87fded028e4f433692349148c9e0/tensorflow/compiler/mlir/quantization/stablehlo/tests/passes/quantize_composite_functions.mlir#L172.




However, as you said, there are things that cannot be verified statically. For e.g. https://github.com/openxla/stablehlo/blob/main/docs/spec.md#constraints-66  C3 cannot be verified: The quantization dimension for per-axis case can be determined, but size for result might be unknown.

stablehlo/dialect/TypeInference.cpp
abhigunj: op constraints do not allow `per-axis operand` and `not per-axis result`.  only `quantization_dimension` can differ.

```
(C1)
element_type(result) is given by:
element_type(operand), if !is_per_axis_quantized(operand).
element_type(operand) except that quantization_dimension(operand) and quantization_dimension(result) may differ, otherwise.

```

stablehlo/dialect/StablehloOps.td
```diff
@@ -2660,24 +2660,27 @@ def StableHLO_ReshapeOp: StableHLO_Op<"reshape",
 }
 
 def StableHLO_DynamicReshapeOp: StableHLO_ShapedInterfaceOp<"dynamic_reshape",
-      [HLO_SpeculatableIfAllInputsStaticAndShapeConstant, NoMemoryEffect]> {
+      [HLO_SpeculatableIfAllInputsStaticAndShapeConstant, NoMemoryEffect,
+       AllElementTypesMatch<["operand", "result"]> /*dynamic_reshape_c1*/]> {
```
abhigunj, line 2664: finding a best way to relax this check for per-axis quantized element type. 

`HLO_CompatibleOperandsAndResultElementType or SameOperandsAndResultElementType ` Traits try to match all  operands including `output_shape`, which is expected to mismatch.

docs/spec.md
```diff
@@ -2679,6 +2679,46 @@ op, but the result shape is specified dynamically via `output_shape`.
...
+|----------|----------------------------|-------------|
+| `result` | tensor or quantized tensor | (C1), (C2)  |
+
+#### Constraints
+
+* (C1) `element_type(result)` is given by:
+  * `element_type(operand)`, if `!is_per_axis_quantized(operand)`.
+  * `element_type(operand)` except that `quantization_dimension(operand)` and
+    `quantization_dimension(result)` may differ, otherwise.
+* (C2) `size(output_shape) = rank(result)`.
```
sdasgup3, line 2709: Please add the constraints C2, and C3 from reshape op back and add C4. `size(output_shape) = rank(result)`

stablehlo/dialect/StablehloOps.td
abhigunj: since existing Traits are not useful,  moving `element_type(result)  =  element_type(operand` verification logic  to the op verifier.

stablehlo/reference/Ops.cpp
```diff
@@ -1589,6 +1594,11 @@ Tensor dynamicIotaOp(Axis iotaDimension, const Tensor &outputShape,
       "dynamic result types are not supported at the moment");
 }
 
+Tensor dynamicReshapeOp(const Tensor &operand, const Tensor &outputShape,
+                        ShapedType resultType) {
+  return reshapeOp(operand, resultType);
```
sdasgup3, line 1599: Can we add a check at https://github.com/openxla/stablehlo/blob/8ba7728d3fdc3ea882e893ee7e53255c95ee0e5a/stablehlo/reference/Ops.cpp#L310 to error on dynamic output shape. 

Also remove the error at https://github.com/openxla/stablehlo/blob/8ba7728d3fdc3ea882e893ee7e53255c95ee0e5a/stablehlo/reference/Ops.cpp#L1588

stablehlo/dialect/TypeInference.cpp
sdasgup3: >> Does this current logic allow for per-axis operand but not per-axis result?
That is not allowed.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -255,6 +255,61 @@ LogicalResult verifyQPerAxisScaleAndZeroPointConstraints(
...
+  if (failed(verifyQPerTensorScaleAndZeroPointConstraints(location, operandTy,
+                                                          resultTy)))
+    return failure();
+
+  // reshape_c1, dynamic_reshape_c1
+  if (failed(verifyQPerAxisScaleAndZeroPointConstraints(location, operandTy,
+                                                        resultTy)))
+    return failure();
+
+  // reshape_c3
```
sdasgup3, line 270: Once we restore the constrains from reshape, aslo add dynamic_reshape_c3 here.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3852,9 +3907,15 @@ LogicalResult verifyDynamicReshapeOp(std::optional<Location> location,
                                      Value result) {
   auto resultType = cast<ShapedType>(result.getType());
   auto outputShapeType = cast<ShapedType>(outputShape.getType());
+
+  if (failed(verifyShapeOperandIsCompatibleWithResultType(location, outputShape,
```
sdasgup3, line 3911: add comment `dynamic_reshape c4` 

where c4 is `size(ouput_shape) = rank(result)`

stablehlo/dialect/StablehloOps.td
```diff
@@ -2663,21 +2663,23 @@ def StableHLO_DynamicReshapeOp: StableHLO_ShapedInterfaceOp<"dynamic_reshape",
       [HLO_SpeculatableIfAllInputsStaticAndShapeConstant, NoMemoryEffect]> {
   let summary = "DynamicReshape operation";
   let description = [{
-    This operation is a work in progress, so it is not yet included in
-    the StableHLO specification: https://github.com/openxla/stablehlo/issues/8.
+    Performs reshape of `operand` tensor to a `result` tensor.
```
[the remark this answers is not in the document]
abhigunj: done. 

`iota` description matches with the spec.

stablehlo/dialect/StablehloOps.td
```diff
@@ -2660,24 +2660,27 @@ def StableHLO_ReshapeOp: StableHLO_Op<"reshape",
...
 
 def StableHLO_DynamicReshapeOp: StableHLO_ShapedInterfaceOp<"dynamic_reshape",
-      [HLO_SpeculatableIfAllInputsStaticAndShapeConstant, NoMemoryEffect]> {
+      [HLO_SpeculatableIfAllInputsStaticAndShapeConstant, NoMemoryEffect,
+       AllElementTypesMatch<["operand", "result"]> /*dynamic_reshape_c1*/]> {
   let summary = "DynamicReshape operation";
   let description = [{
-    This operation is a work in progress, so it is not yet included in
-    the StableHLO specification: https://github.com/openxla/stablehlo/issues/8.
+    Performs reshape of `operand` tensor to a `result` tensor.
```
sdasgup3, line 2667: let this description to match the spec as well.

stablehlo/reference/Ops.cpp
abhigunj: https://github.com/openxla/stablehlo/issues/2294 will fix it. Before I merge reshape op PR.

docs/spec.md
abhigunj: Added C3 back to the spec to match with reshape and applicable for dynamic_reshape.

dynamic_reshape op specific constraint added as C4

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3850,11 +3905,26 @@ LogicalResult verifyDynamicPadOp(std::optional<Location> location,
...
+      !isCompatibleElementTypeForHloTypeInference(operand.getType(),
+                                                  result.getType()))
+    return emitOptionalError(
+        location,
+        "expects operand and result to have compatible element type. Got: ",
+        operand.getType(), " and ", result.getType());
+
   auto resultType = cast<ShapedType>(result.getType());
   auto outputShapeType = cast<ShapedType>(outputShape.getType());
+  // dynamic_reshape_c4
```
sdasgup3, line 3920: nit: add a space before

sdasgup3: where is dynamic_c2 defined?

sdasgup3 (commented):
Mostly LGTM with a few comments.

docs/spec.md
```diff
@@ -2679,6 +2679,60 @@ op, but the result shape is specified dynamically via `output_shape`.
 
 &nbsp;[More Examples](https://github.com/openxla/stablehlo/tree/main/stablehlo/tests/interpret/dynamic_iota.mlir)
 
+### dynamic_reshape
+
+#### Semantics
+
+This operation is functionally identical to
+[reshape](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#reshape)
+op, except the result shape is specified dynamically via `output_shape`.
```
ghpvnist, line 2688: Let's use the same wording as dynamic_iota spec
```
*but* the result shape is specified dynamically via `output_shape`.
```

stablehlo/dialect/StablehloOps.td
```diff
@@ -2663,21 +2663,25 @@ def StableHLO_DynamicReshapeOp: StableHLO_ShapedInterfaceOp<"dynamic_reshape",
       [HLO_SpeculatableIfAllInputsStaticAndShapeConstant, NoMemoryEffect]> {
   let summary = "DynamicReshape operation";
   let description = [{
-    This operation is a work in progress, so it is not yet included in
-    the StableHLO specification: https://github.com/openxla/stablehlo/issues/8.
+    This operation is functionally identical to
+    [reshape](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#reshape)
+    op, except the result shape is specified dynamically via `output_shape`.
```
ghpvnist, line 2668: Let's update this similar to spec comment.

stablehlo/tests/interpret/dynamic_reshape.mlir
```diff
@@ -0,0 +1,9 @@
+// RUN: stablehlo-translate --interpret -split-input-file %s
+
+func.func @dynamic_reshape_op_test_si64() {
+  %0 = stablehlo.constant dense<[[1,2,3,4,5,6]]> : tensor<1x6xi64>
```
ghpvnist, line 4: Nit: spaces between comma and renaming variables:
```
%operand = stablehlo.constant dense<[[1, 2, 3, 4, 5, 6]]> : tensor<1x6xi64>
%output_shape = stablehlo.constant dense<[6]> : tensor<1xi64>
%result = stablehlo.dynamic_reshape %operand, %output_shape : (tensor<1x6xi64>, tensor<1xi64>) -> tensor<6xi64>
check.expect_eq_const %result, dense<[1, 2, 3, 4, 5, 6]> : tensor<6xi64>

File: stablehlo/dialect/StablehloOps.td [hunk 18]
```diff
@@ -2663,21 +2663,25 @@ def StableHLO_DynamicReshapeOp: StableHLO_ShapedInterfaceOp<"dynamic_reshape",
...
-    Informally, this operation does the same thing as ReshapeOp except that the
-    result shape is specified dynamically via `output_shape`:
-    https://github.com/openxla/stablehlo/blob/main/docs/spec.md#reshape
+    See:
+    https://github.com/openxla/stablehlo/blob/main/docs/spec.md#dynamic_reshape
 
     Example:
     ```mlir
-    %0 = stablehlo.dynamic_reshape %arg0, %shape : (tensor<?xf32>, tensor<2xindex>) -> tensor<?x?xf32>
+    %0 = stablehlo.constant dense<[3, 2]> : tensor<2xi64>
```
ghpvnist, line 2675: Keeping names consistent.
```
%output_shape = stablehlo.constant dense<[3, 2]> : tensor<2xi64>
%result = stablehlo.dynamic_reshape %operand, %output_shape : (tensor<2x3xi64>, tensor<2xi64>) -> tensor<3x2xi64>
```

ghpvnist (approved):
LGTM once all comments are addressed.

stablehlo/dialect/TypeInference.cpp
abhigunj: updated marker and `dynamic_c2`  verifier logic.

stablehlo/dialect/TypeInference.cpp
```diff
@@ -255,6 +255,61 @@ LogicalResult verifyQPerAxisScaleAndZeroPointConstraints(
   return success();
 }
 
+LogicalResult verifyReshapeOpQuantizationConstraints(
+    std::optional<Location> location, Type operandTy, Type resultTy) {
+  // reshape_c1, dynamic_reshape_c1
```
ghpvnist, line 260: `dynamic_reshape_c1` comment should be listed first (alphabetical order)

stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -3181,8 +3181,24 @@ func.func @dynamic_reshape(%arg0: tensor<?xf32>, %shape: tensor<2xindex>) -> ten
...
+
+// -----
+
+func.func @dynamic_reshape_c2(%arg0: tensor<11xf32>, %shape: tensor<2xindex>) -> tensor<2x5xf32> {
+  // expected-error @+1 {{number of output elements (10) doesn't match expected number of elements}}
+  %0 = "stablehlo.dynamic_reshape"(%arg0, %shape) : (tensor<11xf32>, tensor<2xindex>) -> tensor<2x5xf32>
+  func.return %0 : tensor<2x5xf32>
+}
+
+// -----
```
ghpvnist, line 3198: Can you update the name of the test that has shared verifier logic to something like:
`@dynamic_reshape_c3_reshape_c#`

ghpvnist (approved):
Did another pass and added few more comments. Otherwise, I think this is good to go.

stablehlo/tests/ops_stablehlo.mlir
abhigunj: those are at ops_stablehlo_quantized.mlir. updated. Thanks.

sdasgup3 (approved):

stablehlo/tests/ops_stablehlo_quantized.mlir
mlevesquedion, line 1: Should "dynamic_reshape" in the function names below actually be "reshape"?

abhigunj: `dynamic_reshape_c1_reshape_c1` to indicate - the test validates a verifier logic added for `dynamic_reshape_c1` and `reshape_c1` constraints.  Since we are not adding per op negative test for the same code path,  name will help to know `dynamic_reshape_c1` verifier logic is tested.

mlevesquedion: IMO it would be preferable to add tests for dynamic_reshape in addition to reshape. Such tests should not be too onerous to add AFAIK. The fact that the underlying logic is the same does not matter as that is an implementation detail.

abhigunj: I am align with adding more per op -ve tests. 
I can't find the previous discussion during quantization verifiers, my understanding was we don't need to add per op -ve tests for the same verifier logic. 

cc @ghpvnist

ghpvnist: If both ops share the same code, it shouldn't be necessary to add duplicate tests per our [testing guideline](https://github.com/openxla/stablehlo/blob/main/docs/reference.md#testing-guidelines) (See G2 for rationale). Also, we get very little mileage (we had very little bugs, if any, for ops we reviewed with branch coverage related to constraints) at the cost of maintainability (e.g. updating tests after making breaking changes to ops), so we instead decided this is the best balance for correctness, test coverage, and maintainability.

mlevesquedion: Hmm. Did you mean `(G4) If the implementation of an op depends on other ops, should we write tests for the latter?` ?

I would argue that the maintainability cost is small: the behavior under test shouldn't change, and the burden of updating a few more ops if the syntax changes is small. Since this is high-level, user-facing behavior, I personally would prefer to see tests for it, regardless of how it's implemented. Having better test coverage gives freedom to refactor things with confidence.

All that being said, I'm okay with not adding the tests.

PR #2283, opened 2024-05-03
[RFC] StableHLO v1.0 Opset Deprecations & Cleanups

GleasonK:
A proposal to remove redundant operations from StableHLO before long-term compatibility guarantees go into place.

High level summary:
- Remove `CreateTokenOp`, `TraceOp`, `BroadcastOp`, `DotOp`, `UnaryEinsumOp`, `RealDynamicSliceOp`.
- Enhance `DynamicSliceOp`.
- Move `CrossReplicaSumOp` to CHLO.
- Hopefully remove/move to CHLO (need feedback) `MapOp`, `RngOp`, `EinsumOp`, `TorchIndexSelectOp`, `GetTupleElementOp`, and `tuple` type.

OpenXLA Discuss post: https://groups.google.com/a/openxla.org/g/openxla-discuss/c/sBAkvnd2bcA

Related tickets: #2176, #3

PR #2295, opened 2024-05-07
Remove TraceOp from StableHLO

GleasonK:
Part of #2283. This op doesn't seem to have any uses in frameworks or compilers. It shouldn't cause any large issues to remove this op all together, also since this op was never specced, it is exempt from compatibility guarantees.

In the case that it is used somewhere, I would recommend migration to a custom_call.


mlevesquedion (commented):
I think there's a few more files that have references to `trace`:
- docs/spec.md
- docs/interpreter_status.md

Besides that, LGTM.

mlevesquedion (approved):

PR #2306, opened 2024-05-08
Spec dynamic_pad

### LINKED ISSUE #2292: dynamic_pad

mlevesquedion:
This includes:
- Adding partial support to the interpreter (only when the result is static)
- Cleaning up and completing the verifier
- Updating the ODS
- Adding relevant tests

I made a few drive-by fixes to related logic/ops as well.

Constraint coverage:

```
I1. operand is tensor or per-tensor quantized tensor: ODS
I2. padding_value is 0-dimensional tensor or per-tensor quantized tensor: ODS
I3. edge_padding_low is integer tensor: ODS
I4. edge_padding_high is integer tensor: ODS
I5. interior_padding is integer tensor: ODS
C1. element_type(operand) = element_type(padding_value) = element_type(result): ODS
C2. size(edge_padding_low) = size(edge_padding_high) = size(interior_padding) = rank(operand): ODS + verifier
C3. 0 <= interior_padding: verifier
C4. shape(result) = shape(operand) + edge_padding_low + max(shape(operand) - 1, 0) * interior_padding + edge_padding_high: verifier
```
ref https://github.com/openxla/stablehlo/issues/2267
Fixes #2292

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -2647,7 +2647,7 @@ op, but the result shape is specified dynamically via `output_shape`.
 
 | Label | Name             | Type                                         | Constraints |
 |-------|------------------|----------------------------------------------|-------------|
-| (I1)  | `output_shape`   | 1-dimensional tensor of type `si64` | (C1), (C2)  |
+| (I1)  | `output_shape`   | 1-dimensional tensor of type `si64`          | (C1), (C2)  |
 | (I2)  | `iota_dimension` | `si64`                                       | (C1)        |
 
 #### Outputs
@@ -2685,8 +2685,8 @@ op, but the result shape is specified dynamically via `output_shape`.
 
 This operation is functionally identical to
 [pad](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#pad)
-op, but the amounts of padding are specified dynamically via `edge_padding_low`,
-`edge_padding_high` and `interior_padding`.
+op, but with `edge_padding_low`, `edge_padding_high` and `interior_padding`
+specified dynamically as values.
 
 #### Inputs
 
@@ -2722,9 +2722,9 @@ op, but the amounts of padding are specified dynamically via `edge_padding_low`,
 //            [4, 5, 6]
 //           ]
 // %padding_value: 0
-// %edge_padding_low = [0, 1]
-// %edge_padding_high = [2, 1]
-// %interior_padding = [1, 2]
+// %edge_padding_low: [0, 1]
+// %edge_padding_high: [2, 1]
+// %interior_padding: [1, 2]
 %result = "stablehlo.dynamic_pad"(%operand, %padding_value,
   %edge_padding_low, %edge_padding_high, %interior_padding
 ) : (tensor<2x3xi32>, tensor<i32>, tensor<2xi32>, tensor<2xi32>, tensor<2xi32>) -> tensor<5x9xi32>
@@ -2737,7 +2737,7 @@ op, but the amounts of padding are specified dynamically via `edge_padding_low`,
 //          ]
 ```
 
-&nbsp;[More Examples](https://github.com/openxla/stablehlo/tree/main/stablehlo/tests/interpret/pad.mlir)
+&nbsp;[More Examples](https://github.com/openxla/stablehlo/tree/main/stablehlo/tests/interpret/dynamic_pad.mlir)
 
 ### dynamic_reshape
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -2647,7 +2647,7 @@ op, but the result shape is specified dynamically via `output_shape`.
 
 | Label | Name             | Type                                         | Constraints |
 |-------|------------------|----------------------------------------------|-------------|
-| (I1)  | `output_shape`   | 1-dimensional tensor of type `si64`          | (C1), (C2)  |
+| (I1)  | `output_shape`   | 1-dimensional tensor of integer type         | (C1), (C2)  |
 | (I2)  | `iota_dimension` | `si64`                                       | (C1)        |
 
 #### Outputs
@@ -2694,9 +2694,9 @@ specified dynamically as values.
 |-------|---------------------|-----------------------------------------------------|------------------|
 | (I1)  | `operand`           | tensor or per-tensor quantized tensor               | (C1), (C2), (C4) |
 | (I2)  | `padding_value`     | 0-dimensional tensor or per-tensor quantized tensor | (C1)             |
-| (I3)  | `edge_padding_low`  | 1-dimensional tensor of type `si64`                 | (C1), (C4)       |
-| (I4)  | `edge_padding_high` | 1-dimensional tensor of type `si64`                 | (C1), (C4)       |
-| (I5)  | `interior_padding`  | 1-dimensional tensor of type `si64`                 | (C2-C4)          |
+| (I3)  | `edge_padding_low`  | 1-dimensional tensor of integer type                | (C1), (C4)       |
+| (I4)  | `edge_padding_high` | 1-dimensional tensor of integer type                | (C1), (C4)       |
+| (I5)  | `interior_padding`  | 1-dimensional tensor of integer type                | (C2-C4)          |
 
 #### Outputs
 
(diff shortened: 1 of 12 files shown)


docs/spec.md
```diff
@@ -2680,6 +2679,66 @@ op, but the result shape is specified dynamically via `output_shape`.
...
 &nbsp;[More Examples](https://github.com/openxla/stablehlo/tree/main/stablehlo/tests/interpret/dynamic_iota.mlir)
 
+### dynamic_pad
+
+#### Semantics
+
+This operation is functionally identical to
+[pad](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#pad)
+op, but the amounts of padding are specified dynamically via `edge_padding_low`,
+`edge_padding_high` and `interior_padding`.
```
GleasonK, line 2689:
```
but with `edge_padding_low`, `edge_padding_high` and `interior_padding` specified dynamically as operands.
```

Let's use a consistent wording for all dynamic ops, and I like the notion of:

This operation is semantically equivalent to `<OP>`, with `<attribute_names>` specified dynamically as operands.

docs/spec.md
```diff
@@ -2680,6 +2679,66 @@ op, but the result shape is specified dynamically via `output_shape`.
...
+//            [1, 2, 3],
+//            [4, 5, 6]
+//           ]
+// %padding_value: 0
+// %edge_padding_low = [0, 1]
+// %edge_padding_high = [2, 1]
+// %interior_padding = [1, 2]
+%result = "stablehlo.dynamic_pad"(%operand, %padding_value,
+  %edge_padding_low, %edge_padding_high, %interior_padding
+) : (tensor<2x3xi32>, tensor<i32>, tensor<2xi32>, tensor<2xi32>, tensor<2xi32>) -> tensor<5x9xi32>
```
GleasonK, line 2730: Do we want to use `constant` ops in the examples? Copy-pastable? Not sure where we ended on this.

stablehlo/reference/Ops.cpp
```diff
@@ -321,6 +321,14 @@ SmallVector<InterpreterValue> eval(Region &region,
   scope.add(block.getArguments(), args);
 
   for (Operation &operation : block) {
+    if (!llvm::all_of(operation.getResults(), [](OpResult r) {
```
GleasonK, line 324: Nice. Should we also check for operands, or is that handled elsewhere?

docs/spec.md
```diff
@@ -2680,6 +2679,66 @@ op, but the result shape is specified dynamically via `output_shape`.
...
+// %result: [
+//           [0, 1, 0, 0, 2, 0, 0, 3, 0],
+//           [0, 0, 0, 0, 0, 0, 0, 0, 0],
+//           [0, 4, 0, 0, 5, 0, 0, 6, 0],
+//           [0, 0, 0, 0, 0, 0, 0, 0, 0],
+//           [0, 0, 0, 0, 0, 0, 0, 0, 0]
+//          ]
+```
+
+&nbsp;[More Examples](https://github.com/openxla/stablehlo/tree/main/stablehlo/tests/interpret/pad.mlir)
```
abhigunj, line 2740: link to `/dynamic_pad.mlir`

GleasonK (approved):

stablehlo/reference/Ops.cpp
abhigunj, line 324: Thanks for taking care of this change.
I was reconsidering adding this check vs per op check because there are only 6 dynamic_ops and we are going to update the reference interpreter to support dynamic reshape. If we end up updating  interpreter to allow dynamic shape one op by one, then this check needs to be relaxed per op basis.

For now, we can keep the check, no change needed here.

abhigunj (approved):

stablehlo/reference/Ops.cpp
mlevesquedion: IMO checking the results covers this because the operands will have to be defined as the result of some op so they will be checked as well.

docs/spec.md
mlevesquedion: Good catch, thanks. Done.

stablehlo/reference/Ops.cpp
mlevesquedion: This is relevant for all ops, though. All ops can have dynamic results (except for e.g. `constant`). Supporting dynamic types in the interpreter is a global change so IMO it makes sense to have a global check.

docs/spec.md
```diff
@@ -2680,6 +2679,66 @@ op, but the result shape is specified dynamically via `output_shape`.
...
+op, but the amounts of padding are specified dynamically via `edge_padding_low`,
+`edge_padding_high` and `interior_padding`.
+
+#### Inputs
+
+| Label | Name                | Type                                                | Constraints      |
+|-------|---------------------|-----------------------------------------------------|------------------|
+| (I1)  | `operand`           | tensor or per-tensor quantized tensor               | (C1), (C2), (C4) |
+| (I2)  | `padding_value`     | 0-dimensional tensor or per-tensor quantized tensor | (C1)             |
+| (I3)  | `edge_padding_low`  | 1-dimensional tensor of type `si64`                 | (C1), (C4)       |
```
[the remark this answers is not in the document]
sdasgup3: sgtm.

sdasgup3 (approved):

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3871,33 +3863,46 @@ LogicalResult verifyDynamicPadOp(std::optional<Location> location,
   auto inputType = cast<RankedTensorType>(operand.getType());
   int inputRank = inputType.getRank();
 
-  auto padType = cast<RankedTensorType>(paddingValue.getType());
-  if (padType.getRank() != 0)
-    return emitOptionalError(location, "padding value type should be a rank-0");
-
+  /*dynamic_pad_c2*/
+  // edgePaddingLow, edgePaddingHigh and interiorPadding are enforced to have
```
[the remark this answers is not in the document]
mlevesquedion: Sorry, I'm not sure which part of `2.` you're referring to?

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3871,33 +3863,46 @@ LogicalResult verifyDynamicPadOp(std::optional<Location> location,
...
-  int outputRank = outputType.getRank();
-  if (inputRank != outputRank)
-    return emitOptionalError(location, "operand rank(", inputRank,
-                             ") must match result(", outputRank, ").");
+  if (!inputType.hasStaticShape() || !outputType.hasStaticShape() ||
+      failed(interiorPaddingMatched))
+    return success();
+
+  SmallVector<int64_t> edgePaddingLowValues;
+  if (failed(matchInts(edgePaddingLow, edgePaddingLowValues))) return success();
```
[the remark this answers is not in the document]
mlevesquedion: C4 requires the input and output to be static and all shape-related operands to be constant.

docs/spec.md
```diff
@@ -2680,6 +2679,66 @@ op, but the result shape is specified dynamically via `output_shape`.
 
 &nbsp;[More Examples](https://github.com/openxla/stablehlo/tree/main/stablehlo/tests/interpret/dynamic_iota.mlir)
 
+### dynamic_pad
+
+#### Semantics
+
+This operation is functionally identical to
+[pad](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#pad)
+op, but with `edge_padding_low`, `edge_padding_high` and `interior_padding`
```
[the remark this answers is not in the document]
mlevesquedion: I don't think the comma really clarifies the sentence, but since you asked, I added it.

stablehlo/dialect/TypeInference.cpp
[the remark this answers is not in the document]
ghpvnist: I meant 2.i. It doesn't apply directly, but the same idea applies here in that we have comments next to the td file to indicate that those are being enforced by ODS.

stablehlo/dialect/TypeInference.cpp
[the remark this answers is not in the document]
ghpvnist: You're right, I should've done a bit more careful reading. Thanks for clarifying!

PR #2308, opened 2024-05-09
Accumulate shape product in int64 to prevent overflow

mlevesquedion:


abhigunj (approved):

stablehlo/dialect/TypeInference.cpp
```diff
@@ -3942,7 +3942,7 @@ LogicalResult verifyDynamicReshapeOp(std::optional<Location> location,
   if (SmallVector<int64_t> shape; operandType.hasStaticShape() &&
                                   matchInts(outputShape, shape).succeeded()) {
     int64_t operandCount = operandType.getNumElements();
-    int64_t shapeCount = std::accumulate(shape.begin(), shape.end(), 1,
+    int64_t shapeCount = std::accumulate(shape.begin(), shape.end(), int64_t{1},
```
sdasgup3, line 3945: More clearer may be? However, based on the posible implementation shown in https://en.cppreference.com/w/cpp/algorithm/accumulate#Version_2, the init will get promited to `int64_t` when the functor is applied.

sdasgup3 (approved):

stablehlo/dialect/TypeInference.cpp
mlevesquedion: Hmm. In practice, it doesn't seem like a promotion is happening, as this can overflow in a way that suggests that the accumulator is 32 bits.

tomwardio: > More clearer may be? However, based on the posible implementation shown in https://en.cppreference.com/w/cpp/algorithm/accumulate#Version_2, the init will get promited to `int64_t` when the functor is applied.

The type T is taken from the init value type, which is currently `int32_t`, the accumulation is then done using this type before being up-cast to `int64_t` on assignment.

We hit this exact issue when increasing our batch size to shard of multiple devices, causing the shape product to overflow.

mlevesquedion: Thanks, Tom!

Issue #2053, opened 2024-02-27
Auto-generate pass documentation


mlevesquedion:
### Request description

Our number of passes is growing and it may be useful to auto-generate documentation for them, like MLIR does: https://mlir.llvm.org/docs/Passes/

Documentation on how to do this: https://mlir.llvm.org/docs/PassManagement/#tablegen-specification:~:text=Using%20the%20gen%2Dpass%2Ddoc%20generator

mlevesquedion:
As an additional example, see: https://heir.dev/docs/passes/memreftoarith/

PR #2310, opened 2024-05-09
Add script to generate pass documentation

mlevesquedion:
Ideally, we would have some fully automated process to generate the
documentation, which would run as a GitHub action on submit. However,
since we currently store all our docs in the repo itself (and they are
mirrored on the OpenXLA website by some external process), I think it
makes sense to use a manually-run script to generate the docs and check
that the docs are up to date in CI. Indeed, we rarely add new passes or
update existing pass documentation, so this shouldn't be too onerous.

Fixes https://github.com/openxla/stablehlo/issues/2053


GleasonK (approved):
lgtm. Would be great to make as much machinery in bazel going forward. Then it can be shared with openxla/xla more easily if needed.

.github/workflows/buildAndTestCMake.yml
```diff
@@ -94,3 +94,8 @@ jobs:
           CMAKE_BUILD_TYPE: Release
           STABLEHLO_ENABLE_BINDINGS_PYTHON: ON
           STABLEHLO_ENABLE_PYTHON_TF_TESTS: ON
+
```
GleasonK, line 97: consider changing `_toc.yaml` to point at the new pages.. I'm not sure how this will work in practice so can do after, but I'd assume it'll just work?

sdasgup3 (approved):
Looks great! Thanks

.github/workflows/buildAndTestCMake.yml
mlevesquedion: Thanks for pointing that out, I was planning to do that in a follow-up if that's okay.

mlevesquedion:
> lgtm. Would be great to make as much machinery in bazel going forward. Then it can be shared with openxla/xla more easily if needed.

Can you expand on why using bazel makes sharing with openxla/xla easier?

Setting this up with CMake seemed more straightforward, which is why I went with that, but using bazel shouldn't be too bad. Just need to create the appropriate build rules and copy from bazel genfiles instead of build/.

mlevesquedion:
> > lgtm. Would be great to make as much machinery in bazel going forward. Then it can be shared with openxla/xla more easily if needed.
> 
> Can you expand on why using bazel makes sharing with openxla/xla easier?
> 
> Setting this up with CMake seemed more straightforward, which is why I went with that, but using bazel shouldn't be too bad. Just need to create the appropriate build rules and copy from bazel genfiles instead of build/.

Per our out of band chat, I will rework this to use bazel instead of CMake.

sdasgup3 (approved):

Issue #2311, opened 2024-05-09
Add helpers to identify ops derived from deprecated ops and preserve their lowerings


mlevesquedion:
To help with the deprecation of Dot in favor of DotGeneral and Broadcast in favor of BroadcastInDim, aiming to minimize disruption to users of the deprecated ops.

PR #2314, opened 2024-05-09
Use DotOp linalg lowerings for simple DotGeneral ops

mlevesquedion:
We are in the process of deprecating DotOp (#2296).

To ease the deprecation, we want DotGeneral ops that are equivalent to simple DotOp ops to be lowered in the same way.

In this PR:
- Add a "isSimpleDot" method to DotGeneralOp which determines if a DotGeneralOp can be expressed as a DotOp.
- Reuse the DotOp linalg lowerings for DotGeneralOp when isSimpleDot.
- Remove "DotOperationType" as it is redundant with the linalg op types.

https://github.com/openxla/stablehlo/issues/2311

### DIFF
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -635,6 +635,21 @@ mlir::Speculation::Speculatability DotGeneralOp::getSpeculatability() {
   return mlir::Speculation::Speculatable;
 }
 
+DotDimensionNumbersAttr getDefaultDotDimensionNumbers(mlir::Value lhs) {
+  return DotDimensionNumbersAttr::get(
+      lhs.getContext(),
+      /*lhsBatchingDimensions=*/{},
+      /*rhsBatchingDimensions=*/{},
+      /*lhsContractingDimensions=*/
+      {cast<ShapedType>(lhs.getType()).getRank() - 1},
+      /*rhsContractingDimensions=*/{0});
+}
+
+bool DotGeneralOp::isSimpleDot() {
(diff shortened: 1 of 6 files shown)


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -635,6 +635,21 @@ mlir::Speculation::Speculatability DotGeneralOp::getSpeculatability() {
...
+  return DotDimensionNumbersAttr::get(
+      lhs.getContext(),
+      /*lhsBatchingDimensions=*/{},
+      /*rhsBatchingDimensions=*/{},
+      /*lhsContractingDimensions=*/
+      {cast<ShapedType>(lhs.getType()).getRank() - 1},
+      /*rhsContractingDimensions=*/{0});
+}
+
+bool DotGeneralOp::isSimpleDot() {
```
GleasonK, line 648: Attrs are stored in a context for the context's lifetime, can we implement this method without the additional DotDimAttr overhead?

If you see benefits to implementing this way then we can keep, the overhead is just one of these will live in the context forever, since there's good caching.

GleasonK (approved):

stablehlo/dialect/StablehloOps.cpp
mlevesquedion: My main concern was avoiding duplication. If we do all the checks explicitly, we are duplicating the knowledge of what it means to be a "simple dot" across two methods. Since the default attr is mostly empty, the memory overhead should not be too big? I think the overhead should be acceptable, but thanks for pointing this out!

PR #2319, opened 2024-05-10
Use BroadcastOp linalg lowerings for simple BroadcastInDimOps

mlevesquedion:
In a similar vein as #2314, this PR eases the process of deprecating BroadcastOp by reusing the BroadcastOp lowerings for BroadcastInDimOp ops derived from BroadcastOp ops.

Unfortunately, BroadcastInDimOp requires a static result type, so not all Broadcast ops can be converted through the deprecated op legalization.

#2311

### DIFF
--- a/stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
+++ b/stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
@@ -419,23 +419,18 @@ struct BroadcastConverter final
   }
 };
 
-template <typename OpTy>
-int64_t getBroadcastSizes(OpTy op) {
-  if constexpr (std::is_same<OpTy, BroadcastOp>::value)
-    return op.getBroadcastSizes().size();
-  if constexpr (std::is_same<OpTy, BroadcastInDimOp>::value)
-    return op.getType().getRank() - op.getBroadcastDimensions().size();
-  std::string str;
-  llvm::raw_string_ostream os(str);
-  os << "getBroadcastSizes called on non-broadcast op: ";
-  op.print(os);
-  llvm::report_fatal_error(str.c_str());
+int64_t getBroadcastSizes(BroadcastOp op) {
+  return op.getBroadcastSizes().size();
 }
 
-template <typename OpTy, typename OpAdaptor>
-LogicalResult lowerBroadcastOp(ConversionPatternRewriter &rewriter,
+int64_t getBroadcastSizes(BroadcastInDimOp op) {
+  return op.getType().getRank() - op.getBroadcastDimensions().size();
+}
+
+template <typename OpTy>
+LogicalResult lowerSimpleBroadcast(ConversionPatternRewriter &rewriter,
                                const TypeConverter *typeConverter, OpTy op,
-                               OpAdaptor adaptor) {
+                               typename OpTy::Adaptor adaptor) {
   auto resultTy = typeConverter->convertType<ShapedType>(op.getType());
   if (!resultTy)
     return rewriter.notifyMatchFailure(op, "type conversion failed");
@@ -461,7 +456,7 @@ struct BroadcastOpToBroadcastConverter final
   LogicalResult matchAndRewrite(
       mlir::stablehlo::BroadcastOp op, OpAdaptor adaptor,
       ConversionPatternRewriter &rewriter) const override {
-    return lowerBroadcastOp(rewriter, getTypeConverter(), op, adaptor);
+    return lowerSimpleBroadcast(rewriter, getTypeConverter(), op, adaptor);
   }
 };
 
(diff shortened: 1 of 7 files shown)


stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
```diff
@@ -419,29 +419,49 @@ struct BroadcastConverter final
...
+template <typename OpTy>
+int64_t getBroadcastSizes(OpTy op) {
+  if constexpr (std::is_same<OpTy, BroadcastOp>::value)
+    return op.getBroadcastSizes().size();
+  if constexpr (std::is_same<OpTy, BroadcastInDimOp>::value)
+    return op.getType().getRank() - op.getBroadcastDimensions().size();
+  std::string str;
+  llvm::raw_string_ostream os(str);
+  os << "getBroadcastSizes called on non-broadcast op: ";
+  op.print(os);
```
GleasonK, line 431: Is this any different than `<< op`? Or maybe `<< *op`, I always forget which

stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
```diff
@@ -419,29 +419,49 @@ struct BroadcastConverter final
   }
 };
 
+template <typename OpTy>
+int64_t getBroadcastSizes(OpTy op) {
+  if constexpr (std::is_same<OpTy, BroadcastOp>::value)
+    return op.getBroadcastSizes().size();
+  if constexpr (std::is_same<OpTy, BroadcastInDimOp>::value)
+    return op.getType().getRank() - op.getBroadcastDimensions().size();
```
GleasonK, line 427: Let's just define this method twice? Or use:

```
typename std::enable_if<std::is_same<OpTy, BroadcastOp>::value ||
                            std::is_same<OpTy, BroadcastInDimOp>::value,
                            int64_t>::type;
```

Let's keep this as a compiler error instead of a runtime error

stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
```diff
@@ -419,29 +419,49 @@ struct BroadcastConverter final
...
+  llvm::raw_string_ostream os(str);
+  os << "getBroadcastSizes called on non-broadcast op: ";
+  op.print(os);
+  llvm::report_fatal_error(str.c_str());
+}
+
+template <typename OpTy, typename OpAdaptor>
+LogicalResult lowerBroadcastOp(ConversionPatternRewriter &rewriter,
+                               const TypeConverter *typeConverter, OpTy op,
+                               OpAdaptor adaptor) {
```
GleasonK, line 438: I think you can use `OpTy::Adaptor` instead of taking a second template argument?

stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
```diff
@@ -419,29 +419,49 @@ struct BroadcastConverter final
...
+    return op.getType().getRank() - op.getBroadcastDimensions().size();
+  std::string str;
+  llvm::raw_string_ostream os(str);
+  os << "getBroadcastSizes called on non-broadcast op: ";
+  op.print(os);
+  llvm::report_fatal_error(str.c_str());
+}
+
+template <typename OpTy, typename OpAdaptor>
+LogicalResult lowerBroadcastOp(ConversionPatternRewriter &rewriter,
```
GleasonK, line 436: Let's give this a longer term name. Like `lowerSimpleLhsBroadcast`

stablehlo/dialect/StablehloOps.td
```diff
@@ -1944,6 +1944,11 @@ def StableHLO_BroadcastInDimOp : StableHLO_Op<"broadcast_in_dim",
     $operand `,` `dims` `=` custom<DenseI64Array>($broadcast_dimensions)
       attr-dict `:` functional-type(operands, results)
   }];
+
+  let extraClassDeclaration = commonClassDeclaration # [{
+    /// Determines if this BroadcastInDim instance can be expressed as a Broadcast.
+    bool isSimpleBroadcast();
```
GleasonK, line 1950: Same here, what's a "simple broadcast"? Maybe `isSimpleLhsBroadcast`?

GleasonK (approved):

stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
mlevesquedion: I ended up just rewriting this function with 2 overloads. I think I went for the template solution initially out of cognitive availability from having just done some template stuff, but overloading is simpler in this case.

stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
mlevesquedion: I don't think `lhs` is relevant here, but `lowerSimpleBroadcast` is a better name for sure. Renamed.

stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
mlevesquedion: Yep, good call! I'll apply that to the dot lowering as well (in a separate PR).

stablehlo/dialect/StablehloOps.td
mlevesquedion: Per offline discussion, let's keep "simple broadcast". Thanks!

PR #2312, opened 2024-05-09
Add spec, interpreter, and verifier for DynamicConvOp

ghpvnist:

PR #2327, opened 2024-05-11
Address remaining feedbacks from #2312

ghpvnist:


ghpvnist:
Added @GleasonK for VHLO changes
Added @sdasgup3 for Spec changes
Added @mlevesquedion for remaining changes

stablehlo/dialect/StablehloOps.td
```diff
@@ -3492,19 +3492,21 @@ def StableHLO_DynamicConvOp : StableHLO_Op<"dynamic_conv",
       [HLO_SpeculatableIfAllInputsStaticAndShapeConstant, NoMemoryEffect]> {
   let summary = "DynamicConv operation";
   let description = [{
-    Computes dot products between windows of `lhs` and slices of `rhs` and
-    produces `result`.
+    This operation is functionally identical to
+    [iota](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#iota)
+    op, but the result shape is specified dynamically via `output_shape`.
```
GleasonK, line 3497: Wrong copy-paste? `iota`?

GleasonK (approved):

stablehlo/dialect/StablehloOps.td
ghpvnist: Thanks!

stablehlo/transforms/VhloToVersionPatterns.td
```diff
@@ -16,12 +16,12 @@ limitations under the License.
...
-      (VHLO_DynamicConvOpV2 $lhs, $rhs, $d_padding, $window_strides, $lhs_dilation, $rhs_dilation, $window_reversal, $input_batch_dimension, $input_feature_dimension, $input_spatial_dimensions, $kernel_input_feature_dimension, $kernel_output_feature_dimension, $kernel_spatial_dimensions, $output_batch_dimension, $output_feature_dimension, $output_spatial_dimensions, $feature_group_count, $batch_group_count, $precision_config)>;
+      (VHLO_DynamicConvOpV2 $lhs, $rhs, $d_padding, $window_strides, /*******/ $lhs_dilation, $rhs_dilation, $window_reversal, $input_batch_dimension, $input_feature_dimension, $input_spatial_dimensions, $kernel_input_feature_dimension, $kernel_output_feature_dimension, $kernel_spatial_dimensions, $output_batch_dimension, $output_feature_dimension, $output_spatial_dimensions, $feature_group_count, $batch_group_count, $precision_config)>;
```
sdasgup3, line 23: what are these comment lines for?

docs/spec.md
```diff
@@ -2720,79 +2720,9 @@ If not specified, all dimensions are assumed to be possibly expanding.
...
-    lambda lhs, rhs: convolution(lhs, rhs, d_padding, window_strides,
-        lhs_dilation, rhs_dilation, window_reversal, input_batch_dimension,
-        input_feature_dimension, input_spatial_dimensions,
-        kernel_input_feature_dimension, kernel_output_feature_dimension,
-        kernel_spatial_dimensions, output_batch_dimension,
-        output_feature_dimension, output_spatial_dimensions,
-        feature_group_count, batch_group_count, precision_config), lhs, rhs)`.
+This operation is functionally identical to
+[convolution](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#convolution)
+op, but the padding is specified dynamically via `d_padding`.
```
sdasgup3, line 2725: Can you also address the comment at https://github.com/openxla/stablehlo/pull/2312#discussion_r1596933457?

stablehlo/dialect/StablehloOps.td
```diff
@@ -3492,19 +3492,21 @@ def StableHLO_DynamicConvOp : StableHLO_Op<"dynamic_conv",
...
+    %d_padding = stablehlo.constant dense<2> : tensor<2x2xi64>
     %result = "stablehlo.dynamic_conv"(%lhs, %rhs, %d_padding) {
       window_strides = array<i64: 4, 4>,
       lhs_dilation = array<i64: 2, 2>,
       rhs_dilation = array<i64: 1, 1>,
       window_reversal = array<i1: false, false>,
       dimension_numbers = #stablehlo.conv<[b, 0, 1, f]x[0, 1, i, o]->[b, 0, 1, f]>,
-      feature_group_count = 1 : i64,
       batch_group_count = 1 : i64,
+      feature_group_count = 1 : i64,
```
sdasgup3, line 3509: Can you please address https://github.com/openxla/stablehlo/pull/2312#discussion_r1597004287?

stablehlo/transforms/VhloToVersionPatterns.td
ghpvnist: Addressing feedback from https://github.com/openxla/stablehlo/pull/2312/files#r1597061119 to align the variables to see the diff easier. It does look a bit silly now that you mention it @mlevesquedion any suggestions?

stablehlo/dialect/StablehloOps.td
ghpvnist: I think having the static trait defeats the purpose of dynamic ops since the point of dynamic ops is to allow dynamic inputs. PTAL at https://github.com/openxla/stablehlo/pull/2327/commits/a4b5b6974e5f5b21a2cb248648580c8e060eb52d for the updated change.

sdasgup3: Let's is the change you made
```
// Any integer tensor type with rank 2.
def HLO_2DIntTensor : TensorRankOf<[HLO_Int], [2]>;
```

and here is my proposal
```
def HLO_Static2DIntTensor : RankedTensorOf<[HLO_Int], [HasStaticShapePred, HasAnyRankOfPred<[2]>], "statically shaped 2-dimensional integer tensor">;
```

>>  the static trait defeats the purpose of dynamic ops since the point of dynamic ops is to allow dynamic inputs

WIth my suggestion the d_padding will still be dynamic input only the size of axis 0 will be statically know. The padding is applied to all the acis of the operand and we always know the number of axis  (as unranked dynamism in not supported). WIth that framworks should always be able to statically generate dynamic_conv with d_padiing size.

Also, the ask here is also related to the uniform handling of padding (in dynamic_pad) and slice_sizes (dynamic_gather). In fact I am ok with any treatment (keeping static sizes for padding or not) if applied uniformly.

wdyt?

stablehlo/transforms/VhloToVersionPatterns.td
mlevesquedion: I agree this doesn't look great. Maybe a comment like: `/* (no padding) */` or `/* (padding removed) */` would help? Otherwise feel free to drop this.

mlevesquedion (approved):
Thanks for taking the time to address the extra feedback! LGTM

stablehlo/transforms/VhloToVersionPatterns.td
ghpvnist: We now have a comment on the `VhloOps.td` file explaining the change and the original purpose was to try aligning them to spot the difference easier; adding `* (no padding) */` or the other variant doesn't align them anyway, so I'll just drop it.

stablehlo/dialect/StablehloOps.td
```diff
@@ -3532,15 +3534,6 @@ def StableHLO_DynamicConvOp : StableHLO_Op<"dynamic_conv",
...
   }];
-
-  let assemblyFormat = [{
-    `(`operands`)`
-       `dim_numbers` `=` custom<ConvolutionDimensions>($dimension_numbers) `,`
-       `window` `=` `{` custom<WindowAttributes>($window_strides,
-                                                 $lhs_dilation, $rhs_dilation,
-                                                 $window_reversal) `}`
-       attr-dict `:` functional-type(operands, results)
-  }];
```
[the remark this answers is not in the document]
ghpvnist: Pretty-printing is not on the radar for v1.0 and currently needs duplication of a few code pieces. We can always add custom pretty-print later (with no compatibility impact), and `dimension_numbers` attribute is already pretty printed, so it's still fairly readable. To get this PR in time for v1.0, I proposed to remove pretty printing for now.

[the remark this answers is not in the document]
mlevesquedion: SG, thanks for explaining!

stablehlo/reference/Ops.cpp
```diff
@@ -519,9 +526,8 @@ SmallVector<InterpreterValue> eval(Region &region,
       auto rhs = scope.findTensor(op.getRhs());
       auto rank = lhs.getRank();
 
-      SmallVector<int64_t> windowStrides(rank - 2, 1);
-      if (auto windowStridesAttr = op.getWindowStrides())
-        windowStrides = SmallVector<int64_t>(windowStridesAttr.value());
+      SmallVector<int64_t> windowStrides = extractAttributeOrDefault<int64_t>(
+          op.getWindowStrides(), rank - 2, 1);
```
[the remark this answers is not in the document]
ghpvnist: `windowReversal` is another one that is affected by this, and these variables are being passed around multiple times, so it's better to keep them as `SmallVector`. I propose to keep the current formulation for consistency across different variables (since we custom handle this via `extractAttributeOrDefault`, it can be written in 1 line too).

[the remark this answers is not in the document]
mlevesquedion: Gotcha. I mostly proposed this because I find `extractAttributeOrDefault` kinda hard to understand without looking at the implementation. (For me, this isn't really about code size, just trying to make the logic as straightforward as possible.) Especially, the fact that the last two arguments are going to be fed directly into the constructor for SmallVector is not obvious.

IMO the following would be clearer, if slightly more verbose:

```
auto windowStrides = op.getWindowStrides() ? llvm::to_vector(op.getWindowStrides.value()) : SmallVector<int64_t>(rank - 2, 1);
```

Anyway, this is a minor thing, feel free to keep the current code.

PR #2328, opened 2024-05-12
Update TypeInference.cpp

xiuhu17:
add static cast to int64_t type

Context: I was trying to setup the environment of XLA, and find there was a syntax error which caused by the non consistent type of parameters in std::max() function.

There was an error on Macos Sonoma 14.3 , using LLVM 17.0.6.

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -4118,7 +4118,8 @@ LogicalResult verifyDynamicPadOp(std::optional<Location> location,
   for (auto [i, in, out, low, high, interior] : llvm::enumerate(
            inputType.getShape(), outputType.getShape(), edgePaddingLowValues,
            edgePaddingHighValues, interiorPaddingValues)) {
-    auto want = in + low + std::max(static_cast<long>(in - 1), long(0)) * interior + high;
+    auto want = in + low +
+                std::max(static_cast<long>(in - 1), long(0)) * interior + high;
     if (out != want)
       return emitOptionalError(location, "expected output dimension at index ",
                                i, " to equal ", want, ", but got ", out);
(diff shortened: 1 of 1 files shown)


GleasonK (approved):
Thanks!

mlevesquedion (approved):
Thank you for your contribution! Would you mind adding some context to the description on what prompted this change? Also, while we're changing this code, would you mind changing `long` to ` int64_t`?

xiuhu17:
> Thank you for your contribution! Would you mind adding some context to the description on what prompted this change? Also, while we're changing this code, would you mind changing `long` to ` int64_t`?

Context: I was trying to setup the environment of XLA, and find there was a syntax error which caused by the non consistent type of parameters in std::max() function. 

Yes, I would change it to ```int64_t```

mlevesquedion:
> > Thank you for your contribution! Would you mind adding some context to the description on what prompted this change? Also, while we're changing this code, would you mind changing `long` to ` int64_t`?
> 
> Context: I was trying to setup the environment of XLA, and find there was a syntax error which caused by the non consistent type of parameters in std::max() function.
> 
> Yes, I would change it to `int64_t`

SG, thank you! Was this on Windows or some other OS and/or with a different compiler? I don't see a syntax error on Linux with clang.

xiuhu17:
> > > Thank you for your contribution! Would you mind adding some context to the description on what prompted this change? Also, while we're changing this code, would you mind changing `long` to ` int64_t`?
> > 
> > 
> > Context: I was trying to setup the environment of XLA, and find there was a syntax error which caused by the non consistent type of parameters in std::max() function.
> > Yes, I would change it to `int64_t`
> 
> SG, thank you! Was this on Windows or some other OS and/or with a different compiler? I don't see a syntax error on Linux with clang.

There was an error on Macos Sonoma 14.3 , using LLVM 17.0.6.

Issue #2223, opened 2024-04-16
Bump clang-format version in lint action


GleasonK:
### Request description

Looks like the runner is still using `clang-format-14`, which has some differences from more recent versions:

https://github.com/openxla/stablehlo/pull/2221/commits/06beb92f1aebc4892f53d5948a70512722cb3164

mlevesquedion:
I discussed this with Kevin offline and I am of the opinion that under the current circumstances it is best to simply use the version of `clang-format` that is already available in the runner. We ran into some issues because we had a different (much more recent) version of clang-format installed, however that version of clang-format is not available in most standard package repositories and I don't think adding a PPA or building from source are reasonable options.

Instead, I opted to make the error message for formatting failures more helpful: https://github.com/openxla/stablehlo/pull/2330

If you want to avoid running into formatting errors in CI, consider running clang-format (be sure to use the right version) on save or in a pre-commit or pre-push hook, e.g. using `./build_tools/github_actions/lint_clang_format.sh -f`.

PR #2330, opened 2024-05-13
Make clang-format errors more helpful

mlevesquedion:
Specifically:
- Provide a command that can be used to fix formatting problems automatically
- Make it obvious what version of clang-format to use

Also:
- Don't specify --style=google because we have a .clang-format file that already specifies that.
- Only show the version if there are formatting failures

Closes https://github.com/openxla/stablehlo/issues/2223.


build_tools/github_actions/lint_clang_format.sh
```diff
@@ -51,7 +49,12 @@ fi
 echo "Running clang-format [mode=$FORMAT_MODE]..."
 echo "  Files: " "${CHANGED_FILES[@]}"
 if [[ $FORMAT_MODE == 'fix' ]]; then
-  clang-format --style=google -i "${CHANGED_FILES[@]}"
+  clang-format -i "${CHANGED_FILES[@]}"
```
GleasonK, line 52: why get rid of the `--style`? I'm forgetting why it was added but I think there were issues when removing this.. can you try running this clang format command on the repo and see how much changes?

GleasonK (approved):

build_tools/github_actions/lint_clang_format.sh
```diff
@@ -51,7 +49,12 @@ fi
...
 else
-  clang-format --style=google --dry-run --Werror "${CHANGED_FILES[@]}"
+  clang-format --dry-run --Werror "${CHANGED_FILES[@]}" || {
+      echo "Please format your code before pushing:"
+      echo "  ./build_tools/github_actions/lint_clang_format.sh -f"
+      echo "Make sure you are using the right major version of clang-format:"
+      echo "  $(clang-format --version)"
+    }
```
GleasonK, line 59: Can you try making this fail? I'm curious if `echo` being the last command will make it seem like the job actually passed in GH Actions.

build_tools/github_actions/lint_clang_format.sh
mlevesquedion: We have a `.clang-format` file which specifies `BasedOnStyle: Google`, so the flag is not needed.

build_tools/github_actions/lint_clang_format.sh
mlevesquedion: It absolutely will! Thanks, good catch.

PR #2333, opened 2024-05-13
Add versioning note to spec

GleasonK:


docs/spec.md
```diff
@@ -19,6 +19,12 @@ these ops executing together within a program. Finally, the
 [Notation](#notation) section discusses the notation used throughout the
 specification.
 
+To view the spec from a previous release of StableHLO, open the repo at the
+[tagged release](https://github.com/openxla/stablehlo/tags) of interest.
+For example, the [StableHLO v0.19.0 Spec](https://github.com/openxla/stablehlo/blob/v0.19.0/docs/spec.md).
+To view changes that occurred at each minor version bump of StableHLO, refer to
+the version log in [VhloDialect.td](https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/VhloDialect.td#L29).
```
ghpvnist, line 26: Maybe use the permalink so the line number doesn't go out of sync?

ghpvnist (approved):

docs/spec.md
mlevesquedion: IMO just link to the file, I think we want folks viewing the file to be able to see the most up to date log, and the file is small enough that the log is easy to find.

mlevesquedion (approved):

PR #2357, opened 2024-05-20
Improve the accuracy of asin for complex inputs using modified Hull et al algorithm

pearu:
This PR implements the modified [Hull et al algorithm](https://dl.acm.org/doi/10.1145/275323.275324) for evaluating `asin` on complex inputs as a CHLO function.

The modified Hull et al algorithm template is defined in [functional_algorithms: asin](https://github.com/pearu/functional_algorithms/blob/9ab746d01538249cdee15a266ffca5cd0283f91d/functional_algorithms/algorithms.py#L215) from which `stablehlo/transforms/ChloDecompositionPatternsMath.td` file is generated using a provided Python script.

In addition, test files `stablehlo/tests/math/asin*.mlir` are also generated that include complex and float samples that magnitudes vary from minimal to maximal values of the corresponding floating point systems plus complex infinities. The generating Python scripts require [MPMath](https://docs.python.org/3/library/math.html) and [functional_algorithms](https://github.com/pearu/functional_algorithms) which both are available via `pip install` and `conda install`.

Notice that stablehlo itself does not need to depend on `mpmath` or `functional_algorithms`. These will be needed only when adding other math functions to the list of these generator scripts.

This PR introduces `check.expect_is_close_to_reference` to check if the asin results are close to reference values. The closeness is measured in terms of ULP differences of `actual` and `reference` values that is defined as the number of all floating point values that are larger than or equal to `min(actual, reference)` and smaller than `max(actual, reference)`.

Before and after this PR, evaluating asin on 1 002 001 samples ULP-equispaced over all complex plane or real line, the closeness to the corresponding reference values has the following statistics:
```
                 current main     ---> this PR
complex64:
ULP diff == 0 count is 95493      ---> 817723
ULP diff == 1 count is 9436       ---> 182678
ULP diff == 2 count is 2936       ---> 1584
ULP diff == 3 count is 2527       ---> 16
ULP diff >= 4 count is 891609     ---> 0

complex128:
ULP diff == 0 count is 115337     ---> 856449
ULP diff == 1 count is 2276       ---> 144708
ULP diff == 2 count is 116        ---> 836
ULP diff == 3 count is 1026       ---> 8
ULP diff >= 4 count is 883246     ---> 0
```
that is, before this PR, there is about 90 % chance that `asin` produces incorrect or inaccurate results when considering all possible floating point complex values. This PR reduces this chance to 0 % with the maximal ULP difference=3.


stablehlo/transforms/ChloDecompositionPatternsMath.td
```diff
@@ -0,0 +1,335 @@
...
+  (StableHLO_MulOp
+    (StableHLO_ConstantLike<"2"> $x),
+    (StableHLO_Atan2Op
+      $x,
+      (StableHLO_AddOp
+        (StableHLO_ConstantLike<"1">:$constant_1 $x),
+        (StableHLO_SqrtOp
+          (StableHLO_SubtractOp
+            $constant_1,
+            (StableHLO_MulOp $x, $x))))))>;
```
GleasonK, line 335: newline at EOF (likely to show up as lint error in CI)

GleasonK: `./build_tools/github_actions/lint_whitespace_checks.sh -f`

Should fix this issue in all files in this PR

stablehlo/reference/Element.cpp
```diff
@@ -174,6 +174,33 @@ bool areApproximatelyEqual(APFloat f, APFloat g) {
   return std::fabs(f.convertToDouble() - g.convertToDouble()) <= 0.0001;
 }
 
+bool areClose(APFloat f, APFloat g, APFloat abs_error) {
```
sdasgup3, line 177: Can't we merge `areClose` and `areApproximatelyEqual` into a single function with signature

```cc
bool areClose(APFloat f, APFloat g, APFloat, APFloat abs_error)
``` 

Depending on where it is called from the `abs_error` can be default `0.0001` or explictly specified?

stablehlo/reference/Element.cpp
```diff
@@ -619,6 +646,35 @@ Element areApproximatelyEqual(const Element &e1, const Element &e2) {
                                      debugString(type).c_str()));
 }
 
+Element areClose(const Element &e1, const Element &e2,
```
sdasgup3, line 649: Being different overloads we can keep the same name as `areApproximatelyEqual`?

stablehlo/tests/CheckOps.cpp
```diff
@@ -156,6 +157,46 @@ llvm::Error evalExpectSerializedEqOp(const Tensor &expected, StringRef probeId,
   return evalExpectEqOp(expected, *tensor);
 }
 
+inline std::string toString(const Element &elem) {
```
sdasgup3, line 160: IMO we do not need this given we have 
https://github.com/openxla/stablehlo/blob/d5a22dd393e685ffeeed952e906136f0f860e74c/stablehlo/reference/Element.cpp#L1126.

wdyt?

stablehlo/tests/CheckOps.cpp
```diff
@@ -156,6 +157,46 @@ llvm::Error evalExpectSerializedEqOp(const Tensor &expected, StringRef probeId,
...
+    c.imag().toString(S);
+    S += ")";
+    return std::string(S);
+  } else {
+    return debugString(elem);
+  }
+}
+
+llvm::Error evalExpectIsCloseOp(const Tensor &lhs, const Tensor &rhs,
+                                const Tensor &abs_error, const Tensor &input) {
```
sdasgup3, line 180: What is the relevance of `input` here given that it is used only for reporting error? IOW,

stablehlo/tests/CheckOps.td
```diff
@@ -81,6 +81,42 @@ def CHECK_ExpectAlmostEqOp :
...
+
+      abs(lhs - rhs) < abs_error
+
+    abs_error is typically
+
+      atol + rtol * abs(rhs)
+
+    where atol must be chosen carefully to match rhs magnitude.
+
+    When rhs operand is non-finite, expect_allclose is equivalent to
```
sdasgup3, line 99: `expect_allcose` --> `expect_is_close`

stablehlo/tests/CheckOps.td
```diff
@@ -81,6 +81,42 @@ def CHECK_ExpectAlmostEqOp :
...
+    expect_eq with respect to that operand.
+
+    Input argument is used only when reporting mismatches between lhs
```
sdasgup3, line 102: What does this signify? Is this some extra information that `lhs`, `rhs` cannot provode?

sdasgup3 (commented):
Looks good

stablehlo/reference/Element.cpp
pearu: yes, will do it.

stablehlo/transforms/ChloDecompositionPatternsMath.td
pearu: Fixed here and other similar issues. Thanks for the lint_whitespace_checks.sh hint!

stablehlo/reference/Element.cpp
pearu: Added optional `abs_error` argument to `areApproximatelyEqual` and eliminated `areClose`.

stablehlo/reference/Element.cpp
pearu: Yes, now using `areApproximatelyEqual` with optional `abs_error` argument and `areClose` is removed.

stablehlo/tests/CheckOps.cpp
pearu: The `input` is the input to `asin` operator, `lhs` is the value of `asin(input)`, `rhs` is the expected value. That said, when the is-close test fails for `lhs` and `rhs`, one would need to know which `input` this failure corresponds to. This is required to determine which part of the asin implementation is responsible for producing the incorrect result.

For example, when disabling the implemented algorithm, a part of a failure message from running `stablehlo/tests/math/asin_complex64.mlir`, is:
```
LLVM ERROR: Error evaluating function: main. 
	Fallback for check.expect_is_close failed: Elements values don't match:
...
  index=[54], actual=(NaN, NaN), expected=(-1.57079637, -89.4159851), atol=2.42017668E-5, input=(-3.40282326E+38, -1.53329339E-19)
  index=[56], actual=(-1.53329339E-19, -5.87747175E-39), expected=(-1.53329339E-19, -1.53329339E-19), atol=4.75334363E-26, input=(-1.53329339E-19, -1.53329339E-19)
  index=[57], actual=(-1.17549435E-38, 0), expected=(-1.17549435E-38, -1.53329339E-19), atol=4.11823295E-26, input=(-1.17549435E-38, -1.53329339E-19)
...
```
While it is true that the `index` and `input` parts have one-to-one relation, it is just a very tedious task to find the `input` value from `index` (it requires decoding `samples` data in asin_complex64.mlir) while it is much more convenient to read the `input` value from the failure message and confirm that, indeed, the actual value, say, `(-1.5E-19, -5.8E-39)` is incorrect for `asin((-1.5E-19, -1.5E-19))`.

If `input` is not shown, then having the information such as `index=[56], actual=(-1.5E-19, -5.8E-39), expected=(-1.5E-19, -1.5E-19)`, is not very useful to determine what-is-wrong/where-to-look-for-a-fix until one determines the `input` value as well.

stablehlo/tests/CheckOps.td
pearu: See https://github.com/openxla/stablehlo/pull/2357#discussion_r1608025497 .
Indeed, `lhs` and `rhs` is not sufficient to determine the cause of the test failure. There is `index` but that is inconvenient to use (see the discussion above). `input` is the most useful parameter that determines the cause why `lsh` and `rhs` do not match.

stablehlo/tests/CheckOps.cpp
pearu: `print` gives a more verbose output. For instance, includes element type that in this particular case is known (e.g. `[1.57079637 : f32, -89.4159851 : f32]`). Also, `print` outputs non-finite numbers using hex format (e.g. `0xFF800000` for `Inf`).

Considering the use case here, I am fine eliminating `toString`.

stablehlo/tests/CheckOps.td
pearu: Fixed.

stablehlo/tests/CheckOps.cpp
pearu: Now using print.

stablehlo/tests/CheckOps.cpp
```diff
@@ -156,6 +157,36 @@ llvm::Error evalExpectSerializedEqOp(const Tensor &expected, StringRef probeId,
...
+            inputIt = input.index_begin();
+       lhsIt != lhs.index_end(); ++lhsIt, ++rhsIt, ++abs_errorIt, ++inputIt) {
+    double abs_error_ =
+        abs_error.get(*abs_errorIt).getFloatValue().convertToDouble();
+    if (!areApproximatelyEqual(lhs.get(*lhsIt), rhs.get(*rhsIt), abs_error_)
+             .getBooleanValue()) {
+      output << "\n  index=" << (*lhsIt) << ", actual=" << lhs.get(*lhsIt)
+             << ", expected=" << rhs.get(*rhsIt)
+             << ", atol=" << abs_error.get(*abs_errorIt)
+             << ", input=" << input.get(*inputIt);
```
sdasgup3, line 180: It is not clear from the error message how the `input` value to be interpreted. Imagine we get this error, the user will have no idea what function  (asin in this case) this input is executed with to generate the actual value.  Without that information,  the value of providing input here is still unclear to me.

stablehlo/tests/CheckOps.cpp
sdasgup3: >> That said, when the is-close test fails for lhs and rhs, one would need to know which input this failure corresponds to. This is required to determine which part of the asin implementation is responsible for producing the incorrect result.

I agree to the overall setiment towards using the input value. However, assuming that `evalExpectIsCloseOp` is intended to be used more generally, how would users decide which function to apply to the input? 

Also, what if the `lhs` is generated by a sequence of operations before they are compared using `expect_is_close`. In that case the questions would be
(1) What is the input? The one after the previous `expect_is_close` operation?
(2) How would the user know which function to apply to this input for debuggability. 

IMO it may not be a good user experience for a user of the check* operation to care about passing the original `input` the `lhs` is generated from plus the function to transform `input` to `lhs` (in some format to capture the case of multiple operations transforming `input` to `lhs`).

wdyt?

sdasgup3: In case you agree to my point we can create an overload of `check.expect_almost_equal` op with tolerances specified by user and its eval function can resuse `evalExpectAlmostEqOp` ?

sdasgup3: On a second thought: why do we need separate tolerances for individual elements of a tensor? `asin` being an elementwise all the elements of the tensor are supposed to be generated using a unified algorithm. Or is this something specific to Hull et. al. algorithm?

My main concern here is  the lines of user experience, for example, how these individual tolerances are to be generated?

stablehlo/reference/Element.cpp
```diff
@@ -170,8 +170,9 @@ bool areApproximatelyEqual(APFloat f, APFloat g) {
   }
 
   // Both f and g are finite (zero, subnormal, or normal) values.
-  if (f.isNegative() != g.isNegative()) return false;
-  return std::fabs(f.convertToDouble() - g.convertToDouble()) <= 0.0001;
+  if (!(f.isZero() || g.isZero()) && (f.isNegative() != g.isNegative()))
```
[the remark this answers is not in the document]
pearu: In testing asin, we would like to have subnormals to be equal to zero. This aligns with the definition of the closeness of floating point numbers in general: f and g are approximately equal when
```
abs(f - g) < atol + rtol * abs(g)
```
holds. Notice that this condition will not distinguish values like `+0.0`, `-0.0`, nor subnormals when atol and rtol are chosen to be close to the eps and resolution parameters of a particular floating point system (for fp32, atol=rtol=1e-7, for instance, assuming the values of `f`, `g` to have the magnitude of unity).

If one needs to test the sign of a zero (which is relevant, say, when checking branch cut values), using the closeness test is not the most appropriate tool anyway. Instead, the sign should be tested explicitly.

stablehlo/tests/CheckOps.td
```diff
@@ -81,6 +81,42 @@ def CHECK_ExpectAlmostEqOp :
   let assemblyFormat = "$lhs `,` $rhs attr-dict `:` type($lhs)";
 }
 
+def CHECK_ExpectIsCloseOp :
+    Op<CHECK_Dialect, "expect_is_close", [AllTypesMatch<["lhs", "rhs", "input"]>]> {
```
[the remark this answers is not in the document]
pearu: Probably. Notice that I am newbie in stablehlo and this kind of feedback is very much appreciated!

stablehlo/tests/CheckOps.cpp
pearu: I assumed that functions under stablehlo/tests/ are for testing purposes and not exposed to the end-users. When testing, one always knows what functions are being tests.

If this assumption is incorrect, what would be the right place to implement this functionality of `check_is_close` that is used only for testing math functions?

Btw, when users see the above error message, the interpretation of values of `lhs`, `rhs`, and `input` would be all equally unclear/useless when the function being tested is unknown.

stablehlo/tests/CheckOps.cpp
pearu: > On a second thought: why do we need separate tolerances for individual elements of a tensor? `asin` being an elementwise all the elements of the tensor are supposed to be generated using a unified algorithm. Or is this something specific to Hull et. al. algorithm?

The non-zero finite inputs to, say, fp64 version of `asin`, can range from `2.2250738585072014e-308` to `1.7976931348623157e+308`. The absolute values of complex `asin` ranges from `2.2250738585072014e-308` to approximately `710`. Recall, a proper value for `atol` depends on the magnitude of the numbers being compared. For instance, when using `atol=1e-15` for comparing `asin` values that are, say, larger than 1, is appropriate then using the same `atol` value for comparing `asin` values in the order of, say, `1e-100`, is totally useless because `abs(f - g) < 1e-15` always holds.

In the case of `asin`, we have an upper bound to `asin` values. In the case of other math functions, the range of values may be from `1e-308` to `1e+308` and one must choose `atol` value appropriately for sensible testing.

> My main concern here is the lines of user experience, for example, how these individual tolerances are to be generated?

I think there is a confusion of context here. Testing of `asin` correctness on the whole complex plane has nothing to do with an user experience other than stablehlo being able to declare to users that `asin` evaluation for any complex input produces mathematically correct result within the restrictions of the used floating point system. Users calling `asin` do not need to deal with specifying the tolerances at all. The provided Python script that generates these asin tests take care of computing the tolerances appropriately.

pearu: > I agree to the overall setiment towards using the input value. However, assuming that evalExpectIsCloseOp is intended to be used more generally, how would users decide which function to apply to the input?

The assumption of using evalExpectIsCloseOp more generally is OT for this PR. However, I can see that "is close" may be confused with a more general problem of comparing two arbitrary values while here we are dealing with values that (i) are approximations (e.g. the value of `stablhlo.asin(input)`) and (ii) are references (e.g. the value of `mpmath.asin(input)` that is taken as the mathematical truth as it is evaluated using multi-precision fp arithmetic).

To be more explicit about the intention of the used method, I suggest replacing "expect_is_close" with "validate_against" or "expect_close_to_reference" or similar:
```
expect_close_to_reference(actual, reference, abs_allowed_error, input)
```

What do you think?

stablehlo/tools/StablehloTranslateMain.cpp
```diff
@@ -154,6 +154,18 @@ class StablehloTranslateInterpreterFallback
                                            "check.expect_serialized_eq");
     }
 
+    if (auto expectIsCloseOp =
```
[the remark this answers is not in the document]
pearu: Fixed.

stablehlo/tests/CheckOps.cpp
```diff
@@ -156,6 +157,36 @@ llvm::Error evalExpectSerializedEqOp(const Tensor &expected, StringRef probeId,
   return evalExpectEqOp(expected, *tensor);
 }
 
+llvm::Error evalExpectIsCloseOp(const Tensor &lhs, const Tensor &rhs,
+                                const Tensor &abs_error, const Tensor &input) {
+  if (!isSupportedFloatType(abs_error.getElementType())) {
+    const std::string abs_errorType = debugString(abs_error.getType());
```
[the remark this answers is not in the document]
pearu: Fixed.

stablehlo/tests/CheckOps.td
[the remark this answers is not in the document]
pearu: Fixed.

stablehlo/tests/CheckOps.cpp
```diff
@@ -156,6 +157,36 @@ llvm::Error evalExpectSerializedEqOp(const Tensor &expected, StringRef probeId,
...
+    return invalidArgument("abs_error must be a float tensor, got %s",
+                           abs_errorType.c_str());
+  }
+  std::string mismatches;
+  llvm::raw_string_ostream output(mismatches);
+  for (auto lhsIt = lhs.index_begin(), rhsIt = rhs.index_begin(),
+            abs_errorIt = abs_error.index_begin(),
+            inputIt = input.index_begin();
+       lhsIt != lhs.index_end(); ++lhsIt, ++rhsIt, ++abs_errorIt, ++inputIt) {
```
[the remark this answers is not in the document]
pearu: Fixed.

stablehlo/reference/Element.cpp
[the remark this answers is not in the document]
pearu: This change has been undone, so resolving the discussion.

stablehlo/tests/CheckOps.cpp
pearu: The recent update to this PR introduces:
```
check.expect_close_to_reference(actual, reference, input, max_ulp_difference=1)
```
that is available only within `stablehlo/tests/` and it uses ULP difference between actual and reference values to measure the closeness. This eliminated the need to introduce `abs_allowed_error` tensor argument and is actually a much better measure for closeness as it does not depend on the magnitudes of the values.

Still keeping the `input` argument as it is very useful when developing algorithms such as for `asin`.

stablehlo/tests/CheckOps.cpp
pearu: Since `check.expect_is_close_to_reference` is used only in scripts that test specific math functions, there is no ambiguity of interpreting the `input` value in the error messages.

sdasgup3:
Thanks for the irresolution. I will review this asap.

pearu:
I have rebased this PR against the main branch. @sdasgup3 , please review.

stablehlo/tests/CheckOps.cpp
```diff
@@ -156,6 +157,96 @@ llvm::Error evalExpectSerializedEqOp(const Tensor &expected, StringRef probeId,
...
+    return ULPDifference(e1.getFloatValue(), e2.getFloatValue());
+  }
+}
+
+llvm::Error evalExpectIsCloseToReferenceOp(const Tensor &actual,
+                                           const Tensor &reference,
+                                           const Tensor &input,
+                                           uint64_t max_ulp_difference) {
+  auto type = actual.getElementType();
+  if (!(isSupportedFloatType(type) || isSupportedComplexType(type)))
```
GleasonK, line 209: I didn't notice the extra paren after the `!` at first, may be easier to read/reason about as: `!isSupportedFloatType(type) && !isSupportedComplexType(type)`

GleasonK (approved):

GleasonK:
Apologies for the delay on review! PR looks great! Left a few comments, we can resolve and merge!

sdasgup3 (approved):
LGTM!

@reedwm Feel free to add your review comments that we discussed offline.

stablehlo/tests/CheckOps.cpp
```diff
@@ -156,6 +157,96 @@ llvm::Error evalExpectSerializedEqOp(const Tensor &expected, StringRef probeId,
...
+                   .bitcastToAPInt()
+                   .getLimitedValue() -
+               1;
+  // a is ULP-distance between exact 0 and abs(f):
+  uint64_t a = (f.isNegative() ? -f : f).bitcastToAPInt().getLimitedValue();
+  // b is ULP-distance between exact 0 and abs(g):
+  uint64_t b = (g.isNegative() ? -g : g).bitcastToAPInt().getLimitedValue();
+  if (f.isFinite() && g.isFinite()) {
+    // subtract subnormals contribution, round subnormals to closest normal or
+    // zero:
```
reedwm, line 172: Why round subnormals to the closest normal or 0? The ULP difference between 0 and the smallest normal number should be the number of subnormal values, not 1, right?

stablehlo/tests/CheckOps.cpp
```diff
@@ -156,6 +157,96 @@ llvm::Error evalExpectSerializedEqOp(const Tensor &expected, StringRef probeId,
...
+    else
+      b = (2 * b < z ? 0 : 1);
+    if (f.isNegative() != g.isNegative()) {
+      // in case of overflow, return uint64 maximal value
+      if (a > std::numeric_limits<uint64_t>::max() - b)
+        return std::numeric_limits<uint64_t>::max();
+      return a + b;
+    }
+  }
+  return (a > b ? a - b : b - a);
```
reedwm, line 188: This result is somewhat meaningless if `a` and `b` are not finite. It also treats NaN values with different payloads as different.

Either comment something like "Note this returns an arbitrary nonzero value for non-equal values when one value is nonfinite. NaNs with different payloads are not considered equal". Or, return -1 and handle this in evalExpectIsCloseToReferenceOp

stablehlo/tests/CheckOps.cpp
```diff
@@ -156,6 +157,96 @@ llvm::Error evalExpectSerializedEqOp(const Tensor &expected, StringRef probeId,
   return evalExpectEqOp(expected, *tensor);
 }
 
+static uint64_t ULPDifference(APFloat f, APFloat g) {
+  // z is ULP-distance between exact 0 and largest subnormal:
+  uint64_t z = APFloat::getSmallestNormalized(f.getSemantics())
+                   .bitcastToAPInt()
+                   .getLimitedValue() -
```
reedwm, line 164: Assert that the float can fit in 64-bits. This is always true in practice since no float is currently more than 64 bits, but this will currently silently fail if an FP128 type is ever introduced.

stablehlo/tests/CheckOps.cpp
```diff
@@ -156,6 +157,96 @@ llvm::Error evalExpectSerializedEqOp(const Tensor &expected, StringRef probeId,
...
+    else
+      a = (2 * a < z ? 0 : 1);
+    if (z < b)
+      b -= z;
+    else
+      b = (2 * b < z ? 0 : 1);
+    if (f.isNegative() != g.isNegative()) {
+      // in case of overflow, return uint64 maximal value
+      if (a > std::numeric_limits<uint64_t>::max() - b)
+        return std::numeric_limits<uint64_t>::max();
```
reedwm, line 184: This can never overflow, given that the most significant bit of `a` and `b` are both zero, since the sign bit is zero.

reedwm (changes requested):
Note I only reviewed the ULPDifference function.

Can you add a test for ULPDifference? Either by exposing ULPDifference and adding a C++ test or testing the `expect_is_close_to_reference` op?

stablehlo/tests/math/generate_tests.py
```diff
@@ -0,0 +1,313 @@
+"""
+A script to generate test files for math functions with complex and float inputs.
+
+Prerequisites:
```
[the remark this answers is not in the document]
pearu: The `generate*.py` scripts are used to generate sources for implementations of functions algorithms and the corresponding tests. Once the sources are generated, the generate scripts are not needed in the context of building stablehlo. So, these scripts could be considered as separate tools that are useful when new algorithms are introduced that fix the correctness issues of existing functions (asin is not unique, there are several other functions in JAX/XLA/StableHLO stack that have correctness issues, see for instance [here](https://github.com/pearu/complex_function_validation/tree/main/mpmath_jax_results_dev)).

My initial assumption was that the scripts should be close to generated sources but moving to some other location makes also sense (there is no need to distribute the scripts with built stablehlo).

I think splitting the PR will introduce complications: if asin tests are introduced before fixing asin, the asin tests will fail. On the other hand, the asin fixes should come hand in hand with the corresponding tests.

Looking around in the stablehlo tree, possible location for the generate scripts are:
1. close to generated sources as it is now
2. `build_tools` - currently it contains CI actions and build_mlir.sh
3. introduce top level `tools` or similar?

When moving the generate scripts to `build_tools`, these scripts can be merged into a single file for simplicity.

Re requirements.txt: the requirements can be specified in the scripts documentation strings that explain how to use these. There are only a couple of requirements for the environment that can be specified via `pip` or `conda` command lines.

What do you think?

stablehlo/tests/CheckOps.cpp
pearu: Nice catch! Indeed, when inputs are finite, then `a` and `b` are the uint views of the absolute values of inputs that will not exceed `std::numeric_limits<uint64_t>::max()` in the cases of `float` and `double` inputs. However, for 'long double' inputs, `a + b` could still overflow.. Resolving with the assert suggestion in https://github.com/openxla/stablehlo/pull/2357#discussion_r1635403517 and deleting this if-block.

stablehlo/tests/CheckOps.cpp
pearu: > Why round subnormals to the closest normal or 0?
> The ULP difference between 0 and the smallest normal number should be the number of subnormal values, not 1, right?

Correct. My suggestion is to flush subnormals to closest normal value so that testing closeness of math functions values would be device as well as platform agnostic (some devices/platforms have FTZ mode enabled by default).

Hmm, let me check the consequences of accounting subnormals when testing math functions and get back here, perhaps accounting subnormals ought to be optional...

stablehlo/transforms/ChloDecompositionPatternsMath.td
```diff
@@ -0,0 +1,365 @@
+// This file is generated using functional_algorithms tool (0.2.0), see
```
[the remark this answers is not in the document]
pearu: Fixed.

stablehlo/tests/CheckOps.td
```diff
@@ -118,6 +118,44 @@ def CHECK_ExpectEqOp : Op<CHECK_Dialect, "expect_eq", [SameTypeOperands]> {
...
+    point values that are greater than or equal to actual and smaller
+    than reference.
+
+    When reference operand is non-finite, expect_is_close_to_reference
+    is equivalent to expect_eq with respect to that operand.
+
+    Input is the argument to a function that produces an actual value.
+    Input is included in the report of mismatches between actual and
+    reference values.
+
```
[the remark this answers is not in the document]
pearu: Input is now eliminated.

stablehlo/tests/CheckOps.td
```diff
@@ -118,6 +118,44 @@ def CHECK_ExpectEqOp : Op<CHECK_Dialect, "expect_eq", [SameTypeOperands]> {
   let assemblyFormat = "$lhs `,` $rhs attr-dict `:` type($lhs)";
 }
 
+def CHECK_ExpectIsCloseToReferenceOp :
+    Op<CHECK_Dialect, "expect_is_close_to_reference", [
```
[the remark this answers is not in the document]
pearu: Applied the suggested naming of functions.

stablehlo/tests/CheckOps.td
```diff
@@ -118,6 +118,44 @@ def CHECK_ExpectEqOp : Op<CHECK_Dialect, "expect_eq", [SameTypeOperands]> {
...
+    reference values.
+
+    ```mlir
+    check.expect_is_close %arg0, %arg1, %arg2, max_ulp_difference = 3 : tensor<2xf32>, tensor<2xf32>, tensor<2xf32>
+    ```
+  }];
+
+  let arguments = (ins
+      TensorOf<[AnyFloat, AnyComplex]>:$actual,
+      TensorOf<[AnyFloat, AnyComplex]>:$reference,
```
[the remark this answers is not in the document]
pearu: Done.

stablehlo/tests/CheckOps.cpp
```diff
@@ -156,6 +157,96 @@ llvm::Error evalExpectSerializedEqOp(const Tensor &expected, StringRef probeId,
...
+
+static uint64_t ULPDifference(const Element &e1, const Element &e2) {
+  // caller is responsible for ensuring that e1, e2 have both the same
+  // float or complex types
+  if (isSupportedComplexType(e1.getType())) {
+    auto complexLhs = e1.getComplexValue();
+    auto complexRhs = e2.getComplexValue();
+    return std::max(ULPDifference(complexLhs.real(), complexRhs.real()),
+                    ULPDifference(complexLhs.imag(), complexRhs.imag()));
+  } else {
```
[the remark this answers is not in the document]
pearu: Done.

stablehlo/tests/CheckOps.cpp
pearu: Done.

stablehlo/tests/CheckOps.cpp
```diff
@@ -156,6 +157,96 @@ llvm::Error evalExpectSerializedEqOp(const Tensor &expected, StringRef probeId,
...
+  llvm::raw_string_ostream output(mismatches);
+  constexpr int ulp_diff_counter_size = 5;
+  int ulp_diff_counter[ulp_diff_counter_size] = {};
+  for (auto lhsIt = actual.index_begin(), rhsIt = reference.index_begin(),
+            inputIt = input.index_begin();
+       lhsIt != actual.index_end(); ++lhsIt, ++rhsIt, ++inputIt) {
+    auto e1 = actual.get(*lhsIt);
+    auto e2 = reference.get(*rhsIt);
+    uint64_t ulp_diff = ULPDifference(e1, e2);
+    if (!(ulp_diff <= max_ulp_difference)) {
```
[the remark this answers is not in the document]
pearu: Applied. Thanks for the suggestion!

stablehlo/tests/CheckOps.cpp
```diff
@@ -156,6 +157,96 @@ llvm::Error evalExpectSerializedEqOp(const Tensor &expected, StringRef probeId,
...
+      output << "\n  index=" << (*lhsIt) << ", input=" << input.get(*inputIt)
+             << ", actual=" << e1 << ", reference=" << e2
+             << ", ULP difference=" << ulp_diff;
+    }
+    // Gather ULP difference statistics:
```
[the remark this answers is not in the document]
pearu: Applied.

stablehlo/tests/CheckOps.cpp
pearu: Done.

stablehlo/tests/CheckOps.cpp
pearu: Fixed.

stablehlo/tests/math/generate_tests.py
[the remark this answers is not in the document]
pearu: Done. I have introduced build_tools/math/ that contains the generate scripts and the readme with usage instructions.

stablehlo/tests/CheckOps.cpp
pearu: Update: it turns out that the generated implementations work with subnormals just fine. I'll remove subnormals flushing..

stablehlo/transforms/ChloDecompositionPatternsMath.td
```diff
@@ -0,0 +1,386 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+// This file is generated using functional_algorithms tool (0.4.0), see
+//   https://github.com/pearu/functional_algorithms
+// for more information.
```
GleasonK, line 18: Mention `build_tools/math/README.md` ?

GleasonK (approved):

stablehlo/transforms/ChloDecompositionPatternsMath.td
pearu: Yes, fixed.

reedwm:
@pearu I think you missed this comment before. It's easy to miss since it wasn't a line comment like my other comments.

> Can you add a test for ULPDifference? Either by exposing ULPDifference and adding a C++ test or testing the `expect_is_close_to_reference` op?

stablehlo/tests/CheckOps.cpp
pearu: ULPDifference implementation is now revised: all NaNs are considered equal because reference functions typically return qNaN (per definition of special values handling, say, in C++ standard) while function implementations may return signaling NaN values depending on the used algorithm.

stablehlo/tests/CheckOps.cpp
pearu: Rounding subnormals is now removed.

GleasonK:
Also - I've put in a request to add you to the OpenXLA org on GH, keep an eye out for an email / notification. That will allow CI to run without any approval.

pearu:
> Can you add a test for ULPDifference? Either by exposing ULPDifference and adding a C++ test or testing the expect_is_close_to_reference op?

Tests for ULPDifference are added by introducing optional argument `min_ulp_difference` to `expect_close` so that

```
expect_eq(ULPDifference(x, y), d)
```

is equivalent to

```
expect_close(x, y, max_ulp_difference=d, min_ulp_difference=d)
```

without exposing `ULPDifference`.

pearu:
> That will allow CI to run without any approval.

@GleasonK, it looks like a maintainer still needs to approve the CI runs on this PR.

GleasonK:
I'll look into that later today.

As for CI failures, can run: `./build_tools/github_actions/lint_clang_format.sh -f`

Note if you run this and nothing changes, make sure you're using `clang-format-14` (we're figuring out how to smoothen this experience). Also it looks like `chlo_legalize_to_stablehlo.mlir` needs to be updated with the new lowering

pearu:
> Also it looks like `chlo_legalize_to_stablehlo.mlir` needs to be updated with the new lowering

What is the procedure for updating `chlo_legalize_to_stablehlo.mlir`? Is this usually done manually?

reedwm (approved):
The ULPDifference function and its tests looks good to me. I didn't look at the rest of the PR, but @GleasonK and @sdasgup3 looked at the other parts.

Issue #2356, opened 2024-05-20
Request ConvertToSignlessPass in Stablehlo


qingyunqu:
### Request description

When I use `stablehlo-to-linalg` in torch-mlir(https://github.com/llvm/torch-mlir/pull/3367), I find that it will emit `builtin.unrealized_conversion_cast` in the following case:
```
module attributes {torch.debug_module_name = "Uint8ToFloat"} {
  func.func @forward(%arg0: tensor<3xui8>) -> tensor<3xf32> {
    %0 = stablehlo.convert %arg0 : (tensor<3xui8>) -> tensor<3xf32>
    return %0 : tensor<3xf32>
  }
}

#map = affine_map<(d0) -> (d0)>
module attributes {torch.debug_module_name = "Uint8ToFloat"} {
  func.func @forward(%arg0: tensor<3xui8>) -> tensor<3xf32> {
    %0 = builtin.unrealized_conversion_cast %arg0 : tensor<3xui8> to tensor<3xi8>
    %1 = tensor.empty() : tensor<3xf32>
    %2 = linalg.generic {indexing_maps = [#map, #map], iterator_types = ["parallel"]} ins(%0 : tensor<3xi8>) outs(%1 : tensor<3xf32>) {
    ^bb0(%in: i8, %out: f32):
      %3 = arith.uitofp %in : i8 to f32
      linalg.yield %3 : f32
    } -> tensor<3xf32>
    return %2 : tensor<3xf32>
  }
}
```

I think I need mhlo's `ConvertToSignlessPass` to eliminate the `builtin.unrealized_conversion_cast`.

GleasonK:
Hello! If you have time to migrate that pass to StableHLO, we're happy to house it. Looking at the implementation, there is nothing specific to MHLO about that pass so it likely can be shared by both StableHLO/MHLO.

PR #2359, opened 2024-05-21
migrate convert-to-signless pass from mhlo

qingyunqu:
as title, close https://github.com/openxla/stablehlo/issues/2356


stablehlo/transforms/StablehloConvertToSignless.cpp
```diff
@@ -0,0 +1,138 @@
...
+#include "mlir/IR/BuiltinOps.h"
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/MLIRContext.h"
+#include "mlir/IR/Operation.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Support/LogicalResult.h"
+#include "mlir/Transforms/DialectConversion.h"
+#include "stablehlo/conversions/linalg/transforms/TypeConversion.h"
+#include "stablehlo/dialect/StablehloOps.h"
```
GleasonK, line 34: Do we need this dependency? I don't think we use any StableHLO specific features here. (Would be good to remove any StableHLO-specific things for the sake of sharing with MHLO)

stablehlo/transforms/CMakeLists.txt
```diff
@@ -36,6 +36,7 @@ add_mlir_dialect_library(StablehloPasses
   StablehloAggressiveFolder.cpp
   StablehloAggressiveSimplification.cpp
   StablehloCanonicalizeDynamism.cpp
+  StablehloConvertToSignless.cpp
```
GleasonK, line 39: Probably need this in `BUILD.bazel` as well, should just be adding the file name.

stablehlo/transforms/Passes.td
```diff
@@ -141,6 +141,11 @@ def StablehloAggressiveSimplificationPass
   ];
 }
 
+def StablehloConvertToSignlessPass : Pass<"stablehlo-convert-to-signless", "ModuleOp"> {
+  let summary = "Pass to transform the IR to be on signless integers.";
+  let dependentDialects = ["mlir::stablehlo::StablehloDialect"];
```
GleasonK, line 146: I don't think we need this since we aren't explicitly calling any StableHLO op build methods, using upstream generic builders for everything (MHLO's pass doesn't register a dependent dialect).

stablehlo/transforms/StablehloConvertToSignless.cpp
```diff
@@ -0,0 +1,138 @@
...
+struct StablehloConvertToSignlessPass
+    : public impl::StablehloConvertToSignlessPassBase<
+          StablehloConvertToSignlessPass> {
+  using StablehloConvertToSignlessPassBase::StablehloConvertToSignlessPassBase;
+  void runOnOperation() override {
+    auto& context = getContext();
+    ConversionTarget target(context);
+
+    stablehlo::RemoveSignTypeConverter converter;
+    target.markUnknownOpDynamicallyLegal([&](auto op) {
```
GleasonK, line 109: Looks like `unknown` and `arith::ConstantOp` are not covered by the existing test. Can we add a second function to that file which:

1. Uses tensors instead of memref (or is this memref specific, in which case we should modify the passes.td and consider moving next to the linalg lowering).
2. Uses a stablehlo op
3. Uses arith.constant

GleasonK (changes requested):

stablehlo/transforms/StablehloConvertToSignless.cpp
qingyunqu: Done

GleasonK (approved):

GleasonK:
I've sent a request to add you to the OpenXLA GH org so that future PRs don't require CI approval. If there are other names I should add please let me know!

PR #2367, opened 2024-05-31
Revert bytecode changes from #2259

mlevesquedion:
The bytecode changes are causing issues due to some StableHLO consumers using StableHLO directly rather than VHLO.

Temporarily remove the relevant roundtrip tests as well since this breaks round-tripping.

### DIFF
--- a/stablehlo/dialect/StablehloBytecode.cpp
+++ b/stablehlo/dialect/StablehloBytecode.cpp
@@ -511,8 +511,6 @@ void StablehloBytecodeInterface::write(GatherDimensionNumbersAttr attr,
   writer.writeVarInt(stablehlo_encoding::kGatherDimensionNumbers);
   writer.writeSignedVarInts(attr.getOffsetDims());
   writer.writeSignedVarInts(attr.getCollapsedSliceDims());
-  writer.writeSignedVarInts(attr.getOperandBatchingDims());
-  writer.writeSignedVarInts(attr.getStartIndicesBatchingDims());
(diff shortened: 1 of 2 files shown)


sdasgup3 (approved):

stablehlo/dialect/StablehloBytecode.cpp
```diff
@@ -491,30 +491,26 @@ GatherDimensionNumbersAttr
...
+      startIndexMap, indexVectorDim);
 }
 
 void StablehloBytecodeInterface::write(GatherDimensionNumbersAttr attr,
                                        DialectBytecodeWriter &writer) const {
   writer.writeVarInt(stablehlo_encoding::kGatherDimensionNumbers);
   writer.writeSignedVarInts(attr.getOffsetDims());
   writer.writeSignedVarInts(attr.getCollapsedSliceDims());
-  writer.writeSignedVarInts(attr.getOperandBatchingDims());
-  writer.writeSignedVarInts(attr.getStartIndicesBatchingDims());
```
GleasonK, line 517: Should we report fatal if these aren't set? Currently they're silently dropped I believe

GleasonK (approved):

stablehlo/dialect/StablehloBytecode.cpp
mlevesquedion: Good point! I need to get these changes in, but I will look into that and send a follow-up.

PR #2369, opened 2024-05-31
Make target-version required arg for `stablehlo_to_tf_saved_model` API

sdasgup3:


stablehlo/integrations/python/stablehlo/savedmodel/stablehlo_to_tf_saved_model.py
```diff
@@ -208,7 +208,7 @@ def to_tf_saved_model(
 def stablehlo_to_tf_saved_model(
     module: ir.Module,
     saved_model_dir: os.PathLike,
-    target_version: str = stablehlo.get_current_version(),
+    target_version: str,
```
mlevesquedion, line 211: WDYT about defaulting to the minimum version instead?

mlevesquedion (approved):

stablehlo/integrations/python/stablehlo/savedmodel/stablehlo_to_tf_saved_model.py
sdasgup3: That should be fine as well given [min, current] is a moving target following the compatibility window. 
Let us keep this a required field for now delegating version choice to the users.

PR #2370, opened 2024-06-01
Take rank into account when checking if a DotGeneralOp is simple

qingyunqu:
When the rank of either `lhs` or `rhs` is greater than 2, `isSimpleDot` may produce an incorrect result because of how `getDefaultDotDimensionNumbers` is implemented.

### DIFF
--- a/stablehlo/conversions/linalg/transforms/StablehloToLinalgDotProduct.cpp
+++ b/stablehlo/conversions/linalg/transforms/StablehloToLinalgDotProduct.cpp
@@ -186,6 +186,11 @@ struct DotGeneralOpConversion final
       if (succeeded(lowerDotOp<DotGeneralOp, linalg::DotOp>(
               rewriter, getTypeConverter(), op, adaptor)))
         return success();
+      std::string str;
+      llvm::raw_string_ostream os(str);
+      os << "supposedly simple DotGeneralOp could not be converted: ";
+      op.print(os);
+      llvm::report_fatal_error(str.c_str());
     }
 
     // Get various dimension iterator information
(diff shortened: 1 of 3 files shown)


stablehlo/conversions/linalg/transforms/StablehloToLinalgDotProduct.cpp
```diff
@@ -186,11 +186,6 @@ struct DotGeneralOpConversion final
       if (succeeded(lowerDotOp<DotGeneralOp, linalg::DotOp>(
               rewriter, getTypeConverter(), op, adaptor)))
         return success();
-      std::string str;
-      llvm::raw_string_ostream os(str);
-      os << "supposedly simple DotGeneralOp could not be converted: ";
```
GleasonK, line 191: This is all in a branch for `if (op.isSimpleDot())`, should that method be updated instead? Or are there cases where a simple dot lowering should fall back to dot general lowering?

qingyunqu: Maybe update `isSimpleDot()` is more suitable. Either one is fine for me.

GleasonK: Let's update `isSimpleDot`. What would the change be? Adding a rank check?

We have [legalize_dot_general_to_dot.cc](https://github.com/openxla/xla/blob/4fbe1339c2957aaa7a18a5b594d7ee2257567260/xla/mlir_hlo/mhlo/transforms/legalize_dot_general_to_dot/legalize_dot_general_to_dot.cc#L53) which seems to require this.

mlevesquedion (approved):
Thanks for the fix!

PR #2372, opened 2024-06-03
Add diff patch to doc ci error message

GleasonK:
Will help with fixing doc CI issues without need for locally running the CI script:

Note the issue from https://github.com/openxla/stablehlo/pull/2125 where this CI script had trouble running on certain linux distros.

### DIFF
--- a/build_tools/github_actions/ci_build_docs.sh
+++ b/build_tools/github_actions/ci_build_docs.sh
@@ -48,7 +48,10 @@ bazel build "${!targets[@]}"
 
 cp "${targets[@]}" docs/generated
 
-[[ "$CHECK" ]] && [[ "$(git diff)" ]] && {
+DOC_DIFF="$(git diff)"
+[[ "$CHECK" ]] && [[ $DOC_DIFF ]] && {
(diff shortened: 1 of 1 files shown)


build_tools/github_actions/ci_build_docs.sh
```diff
@@ -48,7 +48,10 @@ bazel build "${!targets[@]}"
 
 cp "${targets[@]}" docs/generated
 
-[[ "$CHECK" ]] && [[ "$(git diff)" ]] && {
+DOC_DIFF="$(git diff)"
+[[ "$CHECK" ]] && [[ $DOC_DIFF ]] && {
+  echo "$DOC_DIFF"
+  echo
   echo "Generated pass documentation is out of date. Please re-generate the documentation before pushing:"
   echo "  ./build_tools/github_actions/ci_build_docs.sh"
```
mlevesquedion, line 56: Nit: it may not be super obvious what to do when looking at the error message alone. Maybe add something like: "documentation is out of date *(see diff above)*."

mlevesquedion (approved):

PR #2377, opened 2024-06-04
Fix custom call deserialization bug

GleasonK:
Discovered an issue when converting PJRT to take VHLO -- when deserializing custom_calls with empty operand/result layouts get washed away and fail verification, since these two attrs are tied together. Also this is potentially bad verification which is another potential fix.

### DIFF
--- a/stablehlo/transforms/VhloLegalizeToStablehlo.cpp
+++ b/stablehlo/transforms/VhloLegalizeToStablehlo.cpp
@@ -745,10 +745,11 @@ LogicalResult removeDefaults(const OpConversionPattern<VhloOpTy>& pattern,
       eraseAttrs(vhloAttrs, "api_version");
     if (isEmptyArray(vhloOp.getCalledComputations()))
       eraseAttrs(vhloAttrs, "called_computations");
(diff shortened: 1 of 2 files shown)


stablehlo/transforms/VhloLegalizeToStablehlo.cpp
```diff
@@ -745,10 +745,11 @@ LogicalResult removeDefaults(const OpConversionPattern<VhloOpTy>& pattern,
       eraseAttrs(vhloAttrs, "api_version");
     if (isEmptyArray(vhloOp.getCalledComputations()))
       eraseAttrs(vhloAttrs, "called_computations");
-    if (isEmptyArray(vhloOp.getOperandLayouts()))
+    if (isEmptyArray(vhloOp.getOperandLayouts()) &&
```
sdasgup3, line 748: Does the problematic case has only one of operand and result layouts empty?

GleasonK: Our verification currently says "must specify both operand+result layout, or neither".

So if there's empty operands or results, we end up stripping the empty one, leading to verification failures

Issue #2255, opened 2024-04-24
Provide finer grained build targets for downstream usage of pass pipelines


ScottTodd:
### Request description

We've been using StableHLO downstream in the [IREE project](https://github.com/iree-org/iree), specifically these Bazel targets:
```bzl
        "@stablehlo//:chlo_ops",       # ChloOps in CMake
        "@stablehlo//:stablehlo_ops",  # StablehloOps in CMake
```

With https://github.com/iree-org/iree/pull/16999, we wanted to also include VHLO to benefit from the [compatibility aspects](https://github.com/openxla/stablehlo/blob/main/docs/compatibility.md) of that dialect.

We found two issues that make downstream usage tricky:

1. Depending on `@stablehlo//:stablehlo_passes` (`StablehloPasses` in CMake) pulls in a large number of dependencies, including `interpreter_ops`.
    * Even if we were to just depend on the API in `@stablehlo//:stablehlo_portable_api`, that pulls in the passes target by way of `stablehlo_serialization`.

2. Some parts of StableHLO have had build issues on macOS and Windows.
    * A number of these are fixed at HEAD or have open PRs to resolve them, but it's hard to take a dependency on a project that doesn't regularly test across multiple platforms and compilers as a downstream project that does

With those issues together, we found that adding a fairly small extra dependency introduced a relatively large chunk of unstable code to our build graph.

What I'd like to see as a downstream project could be a smaller footprint for the `@stablehlo//:stablehlo_portable_api` and `@stablehlo//:stablehlo_serialization` targets, perhaps by splitting the VHLO serialization passes out from `@stablehlo//:stablehlo_passes`. Improved support for Windows (recent MSVC compiler versions specifically) and macOS would be appreciated too.

ScottTodd:
The specific build errors we saw on Windows were: https://gist.github.com/ScottTodd/f93a66377454bed60a16725d9039a255, and on macOS: https://github.com/iree-org/iree/actions/runs/8821674527/job/24218079731#step:6:5701. We found patches (already merged or pending) for both of those. I'm less concerned with the specific build errors and more hoping for ways to keep the upstream/downstream environment healthy.

GleasonK:
Thanks for reporting these! A few questions:

1. What is the ideal setup of passes? Do you have a project that "does it right" and a rationale for that setup?

[...]

PR #2381, opened 2024-06-05
Move interpreter probe pass to reference

GleasonK:
Part of #2255.

Currently the `stablehlo_passes` target bundles the reference interpreter, which isn't necessary or intuitive. This splits off the interpreter probe op pass to remove the dependency from stablehlo transforms.


stablehlo/transforms/CMakeLists.txt
sdasgup3, line 1: Thanks for slimming the  dependencies on the consumer of StablehloPasses!

sdasgup3 (approved):

PR #2383, opened 2024-06-07
Port mhlo-quant-legalize-to-int pass to StableHLO

sdasgup3:
This is the port of [mhlo-quant-legalize-to-int](https://github.com/openxla/xla/blob/9449b0851c855ae3f17e8675d95d1f1b68104abd/xla/mlir_hlo/mhlo/transforms/mhlo_passes.td#L378) pass to Stablehlo.

A bunch of tests needs to be updates based on:
 - Different error message because of the differences in quantization verification for mhlo and stablehlo.
 - Update check directive to accommodate pretty printing in stablehlo vs not pretty printed ops in mhlo (like dot_general)
 - Modify tests with boradcast_in_dim to update the syntax of broadcast_dimension (from `DenseIntElementsAttr` to `I64ArrayAttr`)
 - Convert a test with per-axis stablehlo.max with per-tensor stablehlo.max as max only supports per-tensor. The intent of the tests is preserved. 

Also, updated the original code to address creation of  attributes of type `I64ArrayAttr` instead of `DenseIntElementsAttr`.


stablehlo/transforms/Passes.td
```diff
@@ -178,3 +178,15 @@ def StablehloLegalizeCompositeToCallPass :
     ListOption<"exceptListOption", "except", "std::string", "Names of composites that should not be replaced with calls.">
   ];
 }
+
+def StablehloLegalizeQuantToIntPass : Pass<"stablehlo-legalize-quant-to-int", "mlir::func::FuncOp"> {
+  let summary = "Convert from StableHLO quantized ops to StableHLO primitive ops.";
+
+  let description = [{
+    Convert from StableHLO quantized ops with StableHLO quant types to StableHLO primitive ops
+    like int ops.
```
GleasonK, line 187: "Convert StableHLO programs using UniformQuantized types the semantically equivalent integer math."

I feel like when people talk about this pass they refer to "int math" so would like to have it in the description for clarity.

stablehlo/transforms/Passes.td
```diff
@@ -178,3 +178,15 @@ def StablehloLegalizeCompositeToCallPass :
...
+  }];
+  let dependentDialects = ["mlir::chlo::ChloDialect", "mlir::stablehlo::StablehloDialect",
+                           "mlir::quant::QuantizationDialect",
```
GleasonK, line 190: Do we create ops with quantized types / Func call/return? Only need to list the dialects we _create_ ops in, I believe.

stablehlo/transforms/StablehloLegalizeQuantToInt.cpp
```diff
@@ -0,0 +1,1385 @@
...
+#include "mlir/Transforms/DialectConversion.h"
+#include "stablehlo/dialect/ChloOps.h"
+#include "stablehlo/dialect/StablehloOps.h"
+
+namespace mlir::stablehlo {
+namespace {
+
+// TODO: b/311218165 - consider extract this to common utils and better ways to
+// handle polymorphism.
+using QuantType = std::variant<quant::UniformQuantizedType,
```
GleasonK, line 53: Note for future refactoring, I don't love the variant usage. In MLIR we tend to prefer verification / getters with failure, etc. Let's keep it for now though.

GleasonK: I.e.

```
bool isPerChannelType(QuantType quantType) {
  return std::holds_alternative<quant::UniformQuantizedPerAxisType>(quantType);
}
```

MLIR type has facilities built in to make `isa<...>(type)` extremely fast.

stablehlo/tests/stablehlo_legalize_quant_to_int.mlir
```diff
@@ -0,0 +1,2229 @@
+// RUN:  stablehlo-opt --stablehlo-legalize-quant-to-int -split-input-file %s -verify-diagnostics | FileCheck %s
+
+// CHECK-LABEL: func @uniform_quantize_and_dequantize
```
GleasonK, line 3: Can we add one hand-written test to reference that uses this pass? Will help for discovery of someone with the use case "how do I interpret programs with uq types?" Could be `uniform_quantize.mlir` and have a basic UQ add DQ test

stablehlo/tests/stablehlo_legalize_quant_to_int.mlir
```diff
@@ -0,0 +1,2229 @@
...
+func.func @uniform_quantize_and_dequantize(%arg0: tensor<?x?xf32>) -> tensor<?x?xf32> {
+  // CHECK-DAG: %[[SCALES:.*]] = stablehlo.constant dense<1.000000e+00> : tensor<f32>
```
GleasonK, line 5: Also want to confirm prior to approval that we plan to remove the MHLO version of this pass? This is a lot of (hopefully temporary) duplication.

GleasonK (approved):
Awesome to see!! Thanks for porting this, very excited to have this in the repo for all to use / contribute to.

stablehlo/transforms/Passes.td
sdasgup3: done.

stablehlo/transforms/Passes.td
sdasgup3: Agreed. Removed both the dependencies.

stablehlo/transforms/StablehloLegalizeQuantToInt.cpp
sdasgup3: Absolutely.  The variant is used as a catch all for both per-tensor and per-axis types. We have the base class `quant::QuantizedType` which would serve the same purpose. I believe the pass can be cleaned up a lot w.r.t the how the Q types are handled. Let us have a follow up PR to address that.

stablehlo/tests/stablehlo_legalize_quant_to_int.mlir
sdasgup3: Agreed!

stablehlo/transforms/StablehloLegalizeQuantToInt.cpp
sdasgup3: Here is the follow up: https://github.com/openxla/stablehlo/pull/2385

stablehlo/tests/stablehlo_legalize_quant_to_int.mlir
sdasgup3: The pipeline is embedded in interpreter pipeline https://github.com/openxla/stablehlo/pull/2388 and added the necessary tests.

PR #2384, opened 2024-06-07
Lower shape.cstr_broadcastable op in ShapeLegalizeToHLO

sdasgup3:
Lower shape.cstr_broadcastable op in ShapeLegalizeToHLO

Lower the op to shape_assertion custom_calls. And replace the result shape.witness with a const_witness of true. This allows the subsequent shape.assuming regions to be removed.

Also add an option to the pass to control whether to lower cstr_xxx ops. This makes the impl easier as we can reuse util functions in the file. And it  makes sure existing use cases are not modified.

## Why needed
The evaluation of quantized program depends bout lowering the stablehlo quantized types/ops to their integer equivalents. This is accomplished by pass like https://github.com/openxla/stablehlo/pull/2383 which introduces chlo boradcast operations. These chlo operations needs to be lowered to stablehlo ops as well using a pass which the current PR upgrades. Without this upgrade  shape.cstr_broadcastable will not be lowered to stablehlo for evaluation.  

Inspired from https://github.com/openxla/xla/blob/25f38dd29b8442e45072fc097d0f1ebe3163bf9d/xla/mlir_hlo/mhlo/transforms/mhlo_passes.td#L361

### DIFF
--- a/stablehlo/tests/shape_cstr_legalize_to_stablehlo.mlir
+++ b/stablehlo/tests/shape_cstr_legalize_to_stablehlo.mlir
@@ -1,0 +1,110 @@
+// RUN: stablehlo-opt --shape-legalize-to-stablehlo=legalize-constraints=true --split-input-file --verify-diagnostics %s | FileCheck %s
+
+// -----
+
+// CHECK-LABEL: func.func @shape_cstr_broadcastable
+func.func @shape_cstr_broadcastable(%arg0: tensor<2xindex>, %arg1: tensor<2xindex>) {
+  %0 = shape.cstr_broadcastable %arg0, %arg1 : tensor<2xindex>, tensor<2xindex>
+  shape.assuming %0 {
+  }
+  func.return
+  //      CHECK: %[[DIMS1:.*]] = builtin.unrealized_conversion_cast %arg0 : tensor<2xindex> to tensor<2xi32>
+  // CHECK-NEXT: %[[DIMS2:.*]] = builtin.unrealized_conversion_cast %arg1 : tensor<2xindex> to tensor<2xi32>
+  // CHECK-NEXT: %[[ONES:.*]] = stablehlo.constant dense<1> : tensor<2xi32>
+  // CHECK-NEXT: %[[DIMS1_IS_1:.*]] = stablehlo.compare  EQ, %[[DIMS1]], %[[ONES:.*]],  NOTYPE : (tensor<2xi32>, tensor<2xi32>) -> tensor<2xi1>
+  // CHECK-NEXT: %[[DIMS2_IS_1:.*]] = stablehlo.compare  EQ, %[[DIMS2]], %[[ONES:.*]],  NOTYPE : (tensor<2xi32>, tensor<2xi32>) -> tensor<2xi1>
+  // CHECK-NEXT: %[[EITHER_DIM_IS_1:.*]] = stablehlo.or %[[DIMS1_IS_1]], %[[DIMS2_IS_1]] : tensor<2xi1>
+  // CHECK-NEXT: %[[DIMS_EQ:.*]] = stablehlo.compare  EQ, %[[DIMS1]], %[[DIMS2]],  NOTYPE : (tensor<2xi32>, tensor<2xi32>) -> tensor<2xi1>
+  // CHECK-NEXT: %[[DIMS_BROADCASTABLE:.*]] = stablehlo.or %[[EITHER_DIM_IS_1]], %[[DIMS_EQ]] : tensor<2xi1>
+  // CHECK-NEXT: %[[TRUE:.*]] = stablehlo.constant dense<true> : tensor<1xi1>
+  // CHECK-NEXT: %[[DIM1_BROADCASTABLE:.*]] = stablehlo.slice %[[DIMS_BROADCASTABLE]] [0:1] : (tensor<2xi1>) -> tensor<1xi1>
+  // CHECK-NEXT: %[[BROADCASTABLE_TEMP:.*]] = stablehlo.and %[[TRUE]], %[[DIM1_BROADCASTABLE]] : tensor<1xi1>
+  // CHECK-NEXT: %[[DIM2_BROADCASTABLE:.*]] = stablehlo.slice %[[DIMS_BROADCASTABLE]] [1:2] : (tensor<2xi1>) -> tensor<1xi1>
+  // CHECK-NEXT: %[[ALL_BROADCASTABLE:.*]] = stablehlo.and %[[BROADCASTABLE_TEMP]], %[[DIM2_BROADCASTABLE]] : tensor<1xi1>
+  // CHECK-NEXT: %[[ALL_BROADCASTABLE_SCALAR:.*]] = stablehlo.reshape %[[ALL_BROADCASTABLE]] : (tensor<1xi1>) -> tensor<i1>
+  // CHECK-NEXT: stablehlo.custom_call @shape_assertion(%[[ALL_BROADCASTABLE_SCALAR]]) {error_message = "Shape assertion failed", has_side_effect = true} : (tensor<i1>) -> ()
+  // CHECK-NEXT: %[[WITNESS:.*]] = shape.const_witness true
+  // CHECK-NEXT: shape.assuming %[[WITNESS]] {
+  // CHECK-NEXT: }
+  // CHECK-NEXT: return
+}
+
+// -----
+
+func.func @shape_cstr_broadcastable_input_shape(%arg0: !shape.shape, %arg1: !shape.shape) {
+  // expected-error@+1 {{failed to legalize operation 'shape.cstr_broadcastable' that was explicitly marked illegal}}
+  %0 = shape.cstr_broadcastable %arg0, %arg1 : !shape.shape, !shape.shape
+  shape.assuming %0 {
+  }
+  func.return
+}
+
+// -----
+
+func.func @shape_cstr_broadcastable_different_dims_1(%arg0: tensor<2xindex>, %arg1: tensor<1xindex>) {
+  %0 = shape.cstr_broadcastable %arg0, %arg1 : tensor<2xindex>, tensor<1xindex>
+  shape.assuming %0 {
+  }
+  func.return
+  //      CHECK: %[[DIMS1:.*]] = builtin.unrealized_conversion_cast %arg0 : tensor<2xindex> to tensor<2xi32>
+  // CHECK-NEXT: %[[DIMS2:.*]] = builtin.unrealized_conversion_cast %arg1 : tensor<1xindex> to tensor<1xi32>
+  // CHECK-NEXT: %[[PAD:.*]] = stablehlo.constant dense<1> : tensor<1xi32>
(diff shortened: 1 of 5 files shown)


stablehlo/transforms/ShapeLegalizeToStablehlo.cpp
```diff
@@ -559,6 +648,10 @@ struct ShapeLegalizeToStablehloPass
         });
     target->addLegalOp<tensor::CastOp>();
     target->addLegalOp<UnrealizedConversionCastOp>();
+    if (this->legalize_constraints_) {
+      target->addLegalOp<shape::ConstWitnessOp, shape::AssumingOp,
+                         shape::AssumingYieldOp>();
+    }
```
mlevesquedion, line 654: IMO it would be cleaner to consolidate the handling of `this->legalize_constraints_` into a single if. If it's true, add the pattern to the set of patterns and add the legal ops.

stablehlo/transforms/ShapeLegalizeToStablehlo.cpp
```diff
@@ -491,6 +501,75 @@ struct ConvertTensorFromElementsPattern
...
+        op.getLoc(), DenseIntElementsAttr::get<bool>(boolType, true));
+    for (auto i = 0; i < rank; ++i) {
+      Value broadcastable =
+          rewriter.create<SliceOp>(op.getLoc(), dimBroadcastable, i, i + 1, 1);
+      allBroadcastable =
+          rewriter.create<AndOp>(op.getLoc(), allBroadcastable, broadcastable);
+    }
+    Value allBroadcastableScalar = rewriter.create<ReshapeOp>(
+        op.getLoc(), RankedTensorType::get({}, rewriter.getI1Type()),
+        allBroadcastable);
```
mlevesquedion, line 562: Would using `reduce` be cleaner?

stablehlo/transforms/ShapeLegalizeToStablehlo.cpp
```diff
@@ -118,6 +118,16 @@ Value castToIndex(PatternRewriter& rewriter, Location loc, Value value) {
...
 }
 
+void insertShapeAssertionCustomCall(OpBuilder builder, Location loc,
+                                    Value assert) {
+  auto customCall = builder.create<stablehlo::CustomCallOp>(loc, TypeRange{},
+                                                            ValueRange{assert});
+  customCall.setCallTargetName("shape_assertion");
+  customCall.setHasSideEffect(true);
+  customCall->setAttr("error_message",
+                      builder.getStringAttr("Shape assertion failed"));
```
mlevesquedion, line 128: Who/what consumes this? How do we know that this is the format expected by downstream consumers?

mlevesquedion (approved):
LGTM

stablehlo/tests/shape_cstr_legalize_to_stablehlo.mlir
```diff
@@ -0,0 +1,110 @@
...
+  // CHECK-NEXT: %[[DIM1_BROADCASTABLE:.*]] = stablehlo.slice %[[DIMS_BROADCASTABLE]] [0:1] : (tensor<2xi1>) -> tensor<1xi1>
+  // CHECK-NEXT: %[[BROADCASTABLE_TEMP:.*]] = stablehlo.and %[[TRUE]], %[[DIM1_BROADCASTABLE]] : tensor<1xi1>
+  // CHECK-NEXT: %[[DIM2_BROADCASTABLE:.*]] = stablehlo.slice %[[DIMS_BROADCASTABLE]] [1:2] : (tensor<2xi1>) -> tensor<1xi1>
+  // CHECK-NEXT: %[[ALL_BROADCASTABLE:.*]] = stablehlo.and %[[BROADCASTABLE_TEMP]], %[[DIM2_BROADCASTABLE]] : tensor<1xi1>
+  // CHECK-NEXT: %[[ALL_BROADCASTABLE_SCALAR:.*]] = stablehlo.reshape %[[ALL_BROADCASTABLE]] : (tensor<1xi1>) -> tensor<i1>
+  // CHECK-NEXT: stablehlo.custom_call @shape_assertion(%[[ALL_BROADCASTABLE_SCALAR]]) {error_message = "Shape assertion failed", has_side_effect = true} : (tensor<i1>) -> ()
+  // CHECK-NEXT: %[[WITNESS:.*]] = shape.const_witness true
+  // CHECK-NEXT: shape.assuming %[[WITNESS]] {
+  // CHECK-NEXT: }
+  // CHECK-NEXT: return
```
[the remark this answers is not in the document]
sdasgup3: This is to test the check in stablehlo/transforms/ShapeLegalizeToStablehlo.cpp

```
    // We only support inputs of two 1D tensors.
    if (op.getShapes().size() != 2) return failure();
```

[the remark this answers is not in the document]
mlevesquedion: Sorry, I didn't mean the test case, I meant the shape witness and assuming part.

stablehlo/transforms/ShapeLegalizeToStablehlo.cpp
sdasgup3: I agree! There could be alternative implementation for what is in the source https://github.com/openxla/xla/blob/25f38dd29b8442e45072fc097d0f1ebe3163bf9d/xla/mlir_hlo/mhlo/transforms/mhlo_passes.td#L361 which does this:

Once the individual dims are checked for broadcast safe, we have a bunch of `slice` ops which and `and`d individually followed by reshape to craete a `tensor<i1>` to be provided as input to the custom_call. The reduce would help cut down on the explicit `and`s.

This PR being a port I would prefer keeping it as is for now. We can make the enhancements as follow ups.

mlevesquedion: Sure, that's totally reasonable!

stablehlo/transforms/ShapeLegalizeToStablehlo.cpp
sdasgup3: This particular signature of the custom_call is used for TF shape refinement pass.
[Here](https://github.com/openxla/xla/blob/91713eed21f5cfd58c64a68c13bb89cc40b67229/xla/python/refine_polymorphic_shapes.cc#L58) we can find the reliance on this particular custom_call

and 

[Here]() we can see that in action. 
(RefinePolymorphicShapes)[https://github.com/openxla/xla/blob/91713eed21f5cfd58c64a68c13bb89cc40b67229/xla/python/refine_polymorphic_shapes.cc#L244)   is used to refine dynamic shape in StableHLO embedded in [XlaCallModule]
(https://github.com/tensorflow/tensorflow/blob/a632c89dd7789b62aa69ada8f1bb689fc44cbb8d/tensorflow/compiler/tf2xla/kernels/xla_call_module_loader.cc#L409). 

The main interface being [InferModuleShape](https://github.com/tensorflow/tensorflow/blob/a632c89dd7789b62aa69ada8f1bb689fc44cbb8d/tensorflow/compiler/mlir/tensorflow/transforms/shape_inference.h#L49)

stablehlo/transforms/ShapeLegalizeToStablehlo.cpp
```diff
@@ -491,6 +501,75 @@ struct ConvertTensorFromElementsPattern
   }
 };
 
+struct ConvertCstrBroadcastableOp
+    : public OpRewritePattern<shape::CstrBroadcastableOp> {
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(shape::CstrBroadcastableOp op,
+                                PatternRewriter& rewriter) const override {
+    // As defined, op inputs must be 1D tensor or !shape.shape.
```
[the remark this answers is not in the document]
sdasgup3: done.

stablehlo/transforms/ShapeLegalizeToStablehlo.cpp
```diff
@@ -491,6 +501,75 @@ struct ConvertTensorFromElementsPattern
...
+    auto shape2 = castToI32(rewriter, op.getLoc(), op.getShapes().back());
+    if (!shape1 || !shape2) return failure();
+    auto tensorType1 = dyn_cast<RankedTensorType>(shape1.getType());
+    auto tensorType2 = dyn_cast<RankedTensorType>(shape2.getType());
+    if (!tensorType1 || !tensorType2) return failure();
+
+    // If the two operand shapes are of different sizes, the smaller one is
+    // padded with 1's from the left.
+    int32_t rank =
+        std::max(tensorType1.getDimSize(0), tensorType2.getDimSize(0));
```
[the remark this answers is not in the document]
sdasgup3: done.

stablehlo/transforms/ShapeLegalizeToStablehlo.cpp
sdasgup3: I agree. The current choice is driven by how the code is currenty structured: addition of legal ops handled separately from adding the pattern. Uisng a single if could unify would create an outlier for `legalize_contriants_` case. wdyt?

stablehlo/tests/shape_cstr_legalize_to_stablehlo.mlir
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// -----
+
+func.func @shape_cstr_broadcastable_input_shape(%arg0: !shape.shape, %arg1: !shape.shape) {
+  // expected-error@+1 {{failed to legalize operation 'shape.cstr_broadcastable' that was explicitly marked illegal}}
+  %0 = shape.cstr_broadcastable %arg0, %arg1 : !shape.shape, !shape.shape
+  func.return
```
sdasgup3, line 37: removed `shape.assuming` from tests which anyway would be illegal at the first `shape.cstr_broadcastable` op.
here and below

stablehlo/tests/shape_cstr_legalize_to_stablehlo.mlir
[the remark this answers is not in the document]
sdasgup3: This and the one above it are testing essential the same thing with different dimensions only the order of operands in the `cstr_broadcastable` op are different. Updated the test to test scalar broadcasting. 

Still kept the `shape_assuming` part in the CHECK directive as in all test we are checking the whole output till the return. Propose to keep the same norm for all the tests.

stablehlo/transforms/ShapeLegalizeToStablehlo.cpp
mlevesquedion: Right, I think both takes are valid, but IMO in this case it would be more readable to have a single `if` since that would make it clearer what happens when `legalize_constraints_` is true. Up to you though.

stablehlo/transforms/ShapeLegalizeToStablehlo.cpp
mlevesquedion: Cool, makes sense, thanks! Feels brittle, but I don't necessarily have a better suggestion.

sdasgup3 (commented):
This is a blocker comment for merging this PR.

It seems that the current use-case of lowering statically shaped quantized program to stablehlo math operations can be achieve even without this pass.  

With that it might be the case that we do not prioritize this PR for now. 
I will get back with the resolution of this PR asap. 

@mlevesquedion In anycase thanks for the review. They are exptremly valuable.

PR #2385, opened 2024-06-07
Fix quant type handling in stablehlo-legalize-quant-to-int pass

sdasgup3:
Originally brought up at https://github.com/openxla/stablehlo/pull/2383#discussion_r1630510134

The `stablehlo-legalize-quant-to-int` pass propagated from mho (#2383 ) uses std::variant to handle per-tensor and per-axis quantized type. However, the base class quant::QuantizedType provides a better polymorphic way to accomplish the same. 

## Direction to reviewers

Please review the change only in the following commit: https://github.com/openxla/stablehlo/pull/2385/commits/ee8b499fa691bbb7b6b3948bdc581dce05abaa39

Reason: This PR is based on https://github.com/openxla/stablehlo/pull/2383 which is not merged yet and hence has many files which are already reviewed as part of #2383.


mlevesquedion:
Can you merge main and push? Will make reviewing a bit easier.

stablehlo/transforms/StablehloLegalizeQuantToInt.cpp
```diff
@@ -0,0 +1,1338 @@
...
+    return quantType;
+  return failure();
+}
+
+bool isPerTensorType(Type type) {
+  return isa<quant::UniformQuantizedType>(getElementTypeOrSelf(type));
+}
+
+bool isPerChannelType(Type type) {
+  return isa<quant::UniformQuantizedPerAxisType>(getElementTypeOrSelf(type));
```
mlevesquedion, line 64: Should we align on one term, either "per channel" or "per axis"?

stablehlo/transforms/StablehloLegalizeQuantToInt.cpp
```diff
@@ -0,0 +1,1338 @@
...
+
+// Extracts storage type of a UQ type. Return original type if it is no UQ type.
+Type getQuantStorageType(Type type) {
+  if (auto shaped = dyn_cast<ShapedType>(type)) {
+    return shaped.clone(getQuantStorageType(shaped.getElementType()));
+  }
+
+  auto quantizeType = getQuantType(type);
+  if (succeeded(quantizeType)) {
+    return getQuantStorageType(*quantizeType);
```
mlevesquedion, line 157: Nit: s/quantize/quant or s/quantize/quantized

mlevesquedion (approved):

stablehlo/transforms/StablehloLegalizeQuantToInt.cpp
```diff
@@ -48,36 +48,28 @@ limitations under the License.
 namespace mlir::stablehlo {
 namespace {
 
-// TODO: b/311218165 - consider extract this to common utils and better ways to
-// handle polymorphism.
-using QuantType = std::variant<quant::UniformQuantizedType,
-                               quant::UniformQuantizedPerAxisType>;
+using QuantType = quant::QuantizedType;
```
mlevesquedion, line 51: IMO not worth aliasing, but up to you.

stablehlo/transforms/StablehloLegalizeQuantToInt.cpp
sdasgup3: done

stablehlo/transforms/StablehloLegalizeQuantToInt.cpp
sdasgup3: done.

stablehlo/transforms/StablehloLegalizeQuantToInt.cpp
sdasgup3: let keep it as is mainly because of the extend it is used in the fine.

GleasonK (approved):

PR #2386, opened 2024-06-07
Add verifier for missing constraint for UniformQuantize Op

sdasgup3:
Here is the constraint this PR is adding verifier for
```
(C2) expressed_type(result) = is_float(operand) ? element_type(operand) : expressed_type(operand).
```

[link](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#constraints-96)

Note: Per the guidelines [(P3) Maintain verification code in verifiers and shape functions
](https://github.com/openxla/stablehlo/blob/main/docs/type_inference.md#p3-maintain-verification-code-in-verifiers-and-shape-functions), we are adding the logic in the verifier (not in the type inference) as the element type for this op cannot be inferred.


mlevesquedion (approved):
LGTM

Issue #2373, opened 2024-06-03
Interpreter support for quantized type 


sdasgup3:
The interpreter in its current form does not support quantized type and operations on that type.
The issue is to (1) explore the use-cases and (2) enable support for the same. 

Re (2): One option could be to decompose uisng [mhlo-quant-legalize-to-int](https://github.com/openxla/xla/blob/40d850deb23874e0c8ad6f76e18c2f621c3c53d7/xla/mlir_hlo/mhlo/transforms/mhlo_passes.td#L378) the quantized type to primitive math using before doing interpreter evaluation.

sdasgup3:
- [ ] Port mhlo-quant-legalize-to-int pass to StableHLO (#2383)
- [ ] Lower shape.cstr_broadcastable op in ShapeLegalizeToHLO (#2384)
- [ ] Add a pass-pileline line to convert quantized StableHLO program to math equivalent before evaluation

PR #2388, opened 2024-06-08
Interpreter support for quantized type

sdasgup3:
fixes https://github.com/openxla/stablehlo/issues/2373

The PR is rebased on top on https://github.com/openxla/stablehlo/pull/2383 and cherry-pick  changes from https://github.com/openxla/stablehlo/pull/2384.

### Direction to reviewer

Please review the commit https://github.com/openxla/stablehlo/pull/2388/commits/4d7dc1ae715ba0f0fb3671404441df80902dadcd **excluding** the following files 
  - docs/generated/stablehlo_passes.md
  - stablehlo/transforms/Passes.td
  - stablehlo/transforms/ShapeLegalizeToStablehlo.cpp

### DIFF
--- a/stablehlo/reference/Api.cpp
+++ b/stablehlo/reference/Api.cpp
@@ -68,7 +68,7 @@ FailureOr<func::FuncOp> getMainFunction(ModuleOp module, StringRef mainName) {
 class DefaultInterpreterFallback : public InterpreterFallback {
  public:
   DefaultInterpreterFallback(const InterpreterConfiguration &config)
-      : config(config){};
+      : config(config) {};
 
   virtual llvm::Error operator()(Operation &op, Scope &scope,
                                  Process *process) final {
@@ -229,6 +229,7 @@ LogicalResult lowerQuantization(ModuleOp module, func::FuncOp func) {
     return func.emitError("Failed to lower quantized types/ops in function: ")
            << func.getName();
   }
+  module.dump();
   return success();
 }
 
--- a/stablehlo/reference/Api.cpp
+++ b/stablehlo/reference/Api.cpp
@@ -68,7 +68,7 @@ FailureOr<func::FuncOp> getMainFunction(ModuleOp module, StringRef mainName) {
 class DefaultInterpreterFallback : public InterpreterFallback {
  public:
   DefaultInterpreterFallback(const InterpreterConfiguration &config)
-      : config(config) {};
+      : config(config){};
 
   virtual llvm::Error operator()(Operation &op, Scope &scope,
                                  Process *process) final {
--- a/stablehlo/reference/Api.cpp
+++ b/stablehlo/reference/Api.cpp
@@ -170,19 +170,82 @@ LogicalResult removeDynamism(ModuleOp module, func::FuncOp func,
   return success();
 }
 
+bool isQuantizedTypes(TypeRange types) {
+  return llvm::any_of(types, [](Type type) {
+    return isa<quant::QuantizedType>(getElementTypeOrSelf(type));
+  });
+}
+
+// Recursively checks if an operation or any of its nested operations use
+// quantized types.
+//
+// Args:
+//   op: The operation to check for quantized type usage.
+//
+// Returns:
+//   True if the operation or any nested operation uses quantized types,
+//   false otherwise.
+bool operationUsesQuantType(Operation &op) {
+  for (Region &region : op.getRegions()) {
+    for (Block &block : region) {
+      for (Operation &op : block) {
+        TypeRange operandTypes = op.getOperandTypes();
+        TypeRange resultTypes = op.getResultTypes();
+        if (isQuantizedTypes(operandTypes) || isQuantizedTypes(resultTypes) ||
+            operationUsesQuantType(op)) {
+          return true;
+        }
+      }
+    }
+  }
+  return false;
+}
+
+// Lowers quantization-related operations and types within a function to
+// primitive math operations.
+//
+// This function checks if a function uses quantized types in its inputs,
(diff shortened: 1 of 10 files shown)


stablehlo/reference/Api.cpp
```diff
@@ -170,19 +170,82 @@ LogicalResult removeDynamism(ModuleOp module, func::FuncOp func,
   return success();
 }
 
+bool isQuantizedTypes(TypeRange types) {
+  return llvm::any_of(types, [](Type type) {
```
mlevesquedion, line 174: Should this be `all_of`? To me the name reads like "are all of these types quantized".

stablehlo/reference/Api.cpp
```diff
@@ -170,19 +170,82 @@ LogicalResult removeDynamism(ModuleOp module, func::FuncOp func,
...
+// Args:
+//   op: The operation to check for quantized type usage.
+//
+// Returns:
+//   True if the operation or any nested operation uses quantized types,
+//   false otherwise.
+bool operationUsesQuantType(Operation &op) {
+  for (Region &region : op.getRegions()) {
+    for (Block &block : region) {
+      for (Operation &op : block) {
```
mlevesquedion, line 191: What about using the `walk` method on `Operation`?

stablehlo/reference/Api.cpp
```diff
@@ -170,19 +170,82 @@ LogicalResult removeDynamism(ModuleOp module, func::FuncOp func,
...
+// Args:
+//   module: The module containing the function `func`.
+//   func The function to lower quantized types/operations in.
+//
+// Returns:
+//   A `LogicalResult` indicating success or failure of the lowering process.
+LogicalResult lowerQuantization(ModuleOp module, func::FuncOp func) {
+  if (!(isQuantizedTypes(func.getFunctionType().getInputs()) ||
+        operationUsesQuantType(*func.getOperation()) ||
+        isQuantizedTypes(func.getFunctionType().getResults()))) {
```
mlevesquedion, line 222: I think this might be redundant? I would expect the walking in `operationUsesQuantType` to hit the function inputs as the arguments to the function's body (block), and the return types as the results of the block?

mlevesquedion, line 222: Is this a performance optimization, or are we expecting that running the quantization lowering on a program without quant ops may nonetheless modify the program? If so, should we try to fix that?

mlevesquedion (approved):
I think this approach is fine, but I thought the point of the reference interpreter was to show how the various operations work. Would it be much more onerous to implement support for `uniform_quantize` and `uniform_dequantize` directly?

GleasonK (changes requested):

stablehlo/reference/Api.cpp
GleasonK: Hopefully this code can be deleted if we can just run the pass without validating

stablehlo/reference/Api.cpp
sdasgup3: renamed it to suggest if any of the parameters is of Quantized types.

stablehlo/reference/Api.cpp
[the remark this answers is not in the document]
sdasgup3: Yes the passes in the pipeline will be a no-op for non-quantized operation. However, the reason I used them conditionally is because the --canonicalize pass usage in the piepline.  

I note that  it would be desirable to keep the input as little modified as possible If the interpreter is to be used for "getting familiarity with stablehlo" or "cross validation" purposes. I agree that quantized cases is different as we will be doing the decomposition. But if we can make quant-decomposition pass independent of shape dialect (even for static cases) we can ensure minimum modification. Opened a  ticket for the same https://github.com/openxla/stablehlo/issues/2390.

Once we get rid of canocalization passes from the pipeline, we can consider making the pipeline applied unconditionally.

sdasgup3:
>> I think this approach is fine, but I thought the point of the reference interpreter was to show how the various operations work. Would it be much more onerous to implement support for uniform_quantize and uniform_dequantize directly?

That  is a very good point! 
Even if we support uniform_{de}quantize operations we still need support for other quantized operations, like add with quantized type, in the interpreter.  

The idea here is to handle all the quantized operation uniformly. That means, either 

(A)  Support the execution semantics of all the quantized operations (uniform_quantize, uniform_dequantize, and any other operations which support quantized types) natively in the interpreter, OR
(B) Apply the pass uniformly to lower any quantized operations. 


(B) is the fastest path towards evaluating quantize program.

stablehlo/reference/Api.cpp
sdasgup3: >> are we expecting that running the quantization lowering on a program without quant ops may nonetheless modify the program
Yes, that is the case based on the choice of passes in the pipeline. Per https://github.com/openxla/stablehlo/issues/2390 we believe that the pass pipeline can be simplified so that non-quantized programs will not get affected in which case we can use the pipeline unconditionally.

For more info: https://github.com/openxla/stablehlo/pull/2388#discussion_r1633735620

sdasgup3: PTAL at the revised code. 
Do you mean checking the block arguments and return types for all the operations?
Given that in stablehlo most of the operations (except for if/case/while) are single block operations, I skipped that using  a special check for function op

stablehlo/reference/Api.cpp
sdasgup3: revised the code. PTAL together with https://github.com/openxla/stablehlo/pull/2388#discussion_r1634017820

GleasonK (approved):

PR #2387, opened 2024-06-07
fix StablehloConvertToSignless linkage

Connor-XY:
`StablehloPasses` needs to link `StablehloLinalgTransforms` since `StablehloConvertToSignless` needs `RemoveSignTypeConverter` from `linalg/transforms/TypeConversion`

PR #2389, opened 2024-06-10
fix linalg_passes link

Connor-XY:
As a follow-up on my previous [PR](https://github.com/openxla/stablehlo/pull/2387), The linalg_passes also needs to link `ChloOps` and `StablehloOps`. They are already included in [`BUILD.bazel`](https://github.com/openxla/stablehlo/blob/main/BUILD.bazel#L477,L479) so only the `CMake` file needs modification.

### DIFF
--- a/stablehlo/conversions/linalg/transforms/CMakeLists.txt
+++ b/stablehlo/conversions/linalg/transforms/CMakeLists.txt
@@ -35,6 +35,7 @@ add_mlir_library(StablehloLinalgTransforms
 
   LINK_LIBS PUBLIC
   ChloOps
+  StablehloBase
   StablehloOps
   MLIRArithDialect
   MLIRBufferizationDialect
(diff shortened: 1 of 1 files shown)


stablehlo/conversions/linalg/transforms/CMakeLists.txt
```diff
@@ -34,6 +34,8 @@ add_mlir_library(StablehloLinalgTransforms
   Core
 
   LINK_LIBS PUBLIC
+  ChloOps
```
sdasgup3, line 37: Thanks! Just wondering if this revelation part of the same experiment in https://github.com/openxla/stablehlo/pull/2387#issuecomment-2155661581


Nit: Comparing with the bazel file, `StablehloBase` also seems to be missing!

sdasgup3 (approved):

stablehlo/conversions/linalg/transforms/CMakeLists.txt
Connor-XY: > Thanks! Just wondering if this revelation part of the same experiment in [#2387 (comment)](https://github.com/openxla/stablehlo/pull/2387#issuecomment-2155661581)

Yes, this is the case.

GleasonK (approved):

PR #2394, opened 2024-06-14
Legalize quantized stablehlo operation using uniform_quantize/uniform_dequantize

sdasgup3:
This PR provides with a pass to decompose StableHLO quantized programs using uniform quantize/dequantize
    operations. For example, the following program

    ```mlir
    func.func @add(%arg0: tensor<!quant.uniform<i8:f32,1.0:0>>, %arg1: tensor<!quant.uniform<i8:f32,2.0:1>>) ->  tensor<!quant.uniform<i8:f32,3.0:2>> {
      %0 = "stablehlo.add"(%arg0, %arg1) : (tensor<!quant.uniform<i8:f32,1.0:0>>, tensor<!quant.uniform<i8:f32,2.0:1>>) -> tensor<!quant.uniform<i8:f32,3.0:2>>
      func.return %0 : tensor<!quant.uniform<i8:f32,3.0:2>>
    }
    ```

    Will become:

    ```mlir
    func.func @add(%arg0: tensor<!quant.uniform<i8:f32, 1.000000e+00>>, %arg1: tensor<!quant.uniform<i8:f32, 2.000000e+00:1>>) -> tensor<!quant.uniform<i8:f32, 3.000000e+00:2>> {
      %0 = stablehlo.uniform_dequantize %arg0 : (tensor<!quant.uniform<i8:f32, 1.000000e+00>>) -> tensor<f32>
      %1 = stablehlo.uniform_dequantize %arg1 : (tensor<!quant.uniform<i8:f32, 2.000000e+00:1>>) -> tensor<f32>
      %2 = stablehlo.add %0, %1 : tensor<f32>
      %3 = stablehlo.uniform_quantize %2 : (tensor<f32>) -> tensor<!quant.uniform<i8:f32, 3.000000e+00:2>>
      return %3 : tensor<!quant.uniform<i8:f32, 3.000000e+00:2>>
    }



Per the [docs/spec.md](https://github.com/openxla/stablehlo/blob/main/docs/spec.md), following is the exhaustive list of ops which can be interpreted `using dq-op-q` strategy. The current PR handles all these ops except for un-bolded ones (`DotGeneralOp`, `ConvolutionOp`,  `DynamicConvOp`, and `AddOp`) which are already lowered to integer match using `--stablehlo-legalize-quant-to-int` [pass](https://github.com/openxla/stablehlo/blob/main/stablehlo/transforms/StablehloLegalizeQuantToInt.cpp).

1. **AbsOp**
1. AddOp
1. **Atan2Op**
1. **BatchNormGradOp**
1. **BatchNormInferenceOp**
1. **BatchNormTrainingOp**
1. **CbrtOp**
1. **CeilOp**
1. **CholeskyOp**
1. **ClampOp**
1. **CompareOp**
1. ConvolutionOp
1. **CosineOp**
1. **DivOp**
1. DotGeneralOp
1. DynamicConvOp
1. **Expm1Op**
1. **ExpOp**
1. **FloorOp**
1. **Log1pOp**
1. **LogisticOp**
1. **LogOp**
1. **MaxOp**
1. **MinOp**
1. **MulOp**
1. **NegOp**
1. **PowOp**
1. **ReducePrecisionOp**
1. **RemOp**
1. **RoundOp**
1. **RoundNearestEvenOp**
1. **RsqrtOp**
1. **SelectOp**
1. **SignOp**
1. **SineOp**
1. **SqrtOp**
1. **SubtractOp**
1. **TanhOp**
1. **TriangularSolveOp**


BUILD.bazel
```diff
@@ -1030,6 +1030,7 @@ cc_library(
         "stablehlo/transforms/StablehloLegalizeCompositeToCall.cpp",
         "stablehlo/transforms/StablehloLegalizeDeprecatedOps.cpp",
         "stablehlo/transforms/StablehloLegalizeQuantToInt.cpp",
+        "stablehlo/transforms/StablehloLegalizeQuantizedOpUsingQDQ.cpp",
```
GleasonK, line 1033: `StablehloLegalizeQuantizedOpToQDQ`

stablehlo/transforms/Passes.h
```diff
@@ -66,6 +66,11 @@ void populateStablehloAggressiveFolderPatterns(RewritePatternSet *patterns,
                                                MLIRContext *context,
                                                bool foldFloat);
 
+/// Collection of rewrite patterns for lowering quantized StableHLO operations
+/// using uniform dequantize/quantize operations.
+void populateStablehloLegalizeQuantizedOpUsingQDQPatterns(
```
GleasonK, line 71: ToQDQ

stablehlo/transforms/Passes.td
```diff
@@ -191,3 +191,34 @@ def StablehloLegalizeQuantToIntPass : Pass<"stablehlo-legalize-quant-to-int", "m
     "mlir::stablehlo::StablehloDialect",
   ];
 }
+
+def StablehloLegalizeQuantizedOpUsingQDQPass : Pass<"stablehlo-legalize-quantized-op-using-qdq", "mlir::func::FuncOp"> {
```
GleasonK, line 195: ToQDQ

stablehlo/transforms/StablehloLegalizeQuantizedOpUsingQDQ.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+namespace {
+
+bool isAnyQuantizedTypes(TypeRange types) {
+  return llvm::any_of(types, [](Type type) {
+    return isa<quant::QuantizedType>(getElementTypeOrSelf(type));
+  });
+}
+
+#define DEFINE_QUANT_REWRITE_TO_QDQ_PATTERN(OpName)                            \
```
GleasonK, line 40: Let's write this as a template class then either use a very short macro to just instantiate the macro, or no macro at all if it's short.

GleasonK: Take a look at `populateHloToStablehloPatterns`, perhaps we can do a templated populate method like this which takes in a list of `populate<AddOp, DivOp,...>()`

BUILD.bazel
sdasgup3: done

stablehlo/transforms/Passes.h
sdasgup3: done.

stablehlo/transforms/Passes.td
sdasgup3: done.

stablehlo/transforms/StablehloLegalizeQuantizedOpUsingQDQ.cpp
sdasgup3: done.

stablehlo/tests/ops_stablehlo_quantized.mlir
```diff
@@ -280,20 +584,48 @@ func.func @select_and_scatter_per_tensor_quantization(%arg0: tensor<10x24x24x64x
...
 }
 
+// CHECK-QDQ-NOT: stablehlo.uniform_dequantize
+// CHECK-QDQ: stablehlo.select_and_scatter
+// CHECK-QDQ: stablehlo.uniform_dequantize
+// CHECK-QDQ: stablehlo.uniform_dequantize
+// CHECK-QDQ: stablehlo.compare
+// CHECK-QDQ-NOT: stablehlo.uniform_dequantize
+// CHECK-QDQ: stablehlo.add
+// CHECK-QDQ-NOT: stablehlo.uniform_quantize
```
GleasonK, line 594: I'm confused by how these tests work..

```
// CHECK-QDQ-NOT: stablehlo.uniform_dequantize
// CHECK-QDQ: stablehlo.select_and_scatter
// CHECK-QDQ: stablehlo.uniform_dequantize
```

So it can't have a DQ op, but then after a select_and_scatter it can?

Also my thoughts on readability for most of the tests in this file:

1. Can we make better use of `CHECK-QDQ-NEXT:` to test locality to one another.
2. Can we add variables to enforce some graph structure:

```
// CHECK-QDQ-NEXT: %[[DQ_0:.*]] =  stablehlo.uniform_dequantize 
// CHECK-QDQ-NEXT: %[[DQ_1:.*]] =  stablehlo.uniform_dequantize 
...
// CHECK-QDQ-NEXT: stablehlo.compare %[[DQ_0]], %[[DQ_1]]
```

GleasonK (changes requested):

stablehlo/tests/ops_stablehlo_quantized.mlir
```diff
@@ -23,10 +24,50 @@ func.func @ops_per_axis_quantization(
...
-  func.return
-}
+  func.return %bitcast_convert, %broadcast_in_dim_1, %broadcast_in_dim_2, %custom_call, %dynamic_broadcast_in_dim, %get_dimension_size, %outfeed, %reshape, %send, %transpose, %tuple, %uniform_dequantize, %uniform_quantize : tensor<1x2x2x!quant.uniform<i8<-128:127>:f32:2, {0.1:-30, 0.5:-20}>>, tensor<1x2x3x2x!quant.uniform<i8<-128:127>:f32:3, {0.1:-30, 0.5:-20}>>, tensor<2x2x2x!quant.uniform<i8<-128:127>:f32:0, {0.1:-30, 0.1:-30}>>, tensor<1x2x2x!quant.uniform<i8<-128:127>:f32:2, {0.1:-30, 0.5:-20}>>, tensor<1x2x2x!quant.uniform<i8<-128:127>:f32:2, {0.1:-30, 0.5:-20}>>, tensor<i32>, !stablehlo.token, tensor<2x2x!quant.uniform<i8<-128:127>:f32:1, {0.1:-30, 0.5:-20}>>, !stablehlo.token, tensor<1x2x2x!quant.uniform<i8<-128:127>:f32:1, {0.1:-30, 0.5:-20}>>, tuple<tensor<1x2x2x!quant.uniform<i8<-128:127>:f32:0, {0.1:-30}>>, tensor<1x2x2x!quant.uniform<i8<-128:127>:f32:0, {0.1:-30}>>>, tensor<1x2x2xf32>, tensor<1x2x2x!quant.uniform<i8<-128:127>:f32:2, {0.1:-30, 0.5:-20}>>
+}
+
+// CHECK-QDQ-NOT: stablehlo.uniform_dequantize
```
GleasonK, line 30: These first two tests are monolithic and will likely be more difficult to maintain (and certainly more difficult to understand what is going on in the check lines.

I see some per-op tests below, what is the coverage of these? Can we rely on QDQ testing just using those per-op tests?

stablehlo/tests/ops_stablehlo_quantized.mlir
sdasgup3: Added context to each of the check.

stablehlo/tests/ops_stablehlo_quantized.mlir
sdasgup3: The bulkiest function in this file `ops_per_tensor_quantization` covers most of the elementwise quantized operations which which are supposed to be affected by this pass. In comparison, the following per-op tests covers very few operations  (batch_norm_*, max/min, round*, compare) actually need dq-op-q treatment.

I have added all the checks with context, hopefully this would help.

stablehlo/transforms/StablehloLegalizeQuantizedOpToQDQ.cpp
```diff
@@ -0,0 +1,135 @@
...
+}  // namespace
+
+void populateStablehloLegalizeQuantizedOpToQDQPatterns(
+    RewritePatternSet* patterns, MLIRContext* context) {
+  // The following list covers most of the operations which, according to the
+  // stablehlo spoecification document, interprets the quantized
+  // operation using dequant-op-quant strategy. The ones excluded are
+  // AddOP, ConvolutionOp, DotGeneralOp, and DynamicConvOp, which are current
+  // using `stablehlo-legalize-quant-to-int` pass for decomposituion to
+  // primitive math operations.
```
quanwanandy, line 124: Optional: I think this pass should cover all ops. It is ok to have overlaps between this pass and stablehlo-legalize-quant-to-int. The 2 passes serve different purposes. This pass exactly matches the spec while the other tries to optimize for performance.

quanwanandy (approved):

stablehlo/transforms/StablehloLegalizeQuantizedOpToQDQ.cpp
sdasgup3: Thanks for the review and approval. The main reason I kept them separate is when the passes are applied. in inder `--stablehlo-legalize-quantized-op-to-qdq --stablehlo-legalize-quant-to-int` (which is one of the use-case for converting all the quantized ops to primitive operations). If we add `AddOp` in both places then the performance optimization in later pass will never trigger. However, if we found that that the performance optimization is not desired or not in use then it should consider treating all the ops uniformly (This is something we can take up in follow up PR).

stablehlo/transforms/StablehloLegalizeQuantizedOpToQDQ.cpp
```diff
@@ -0,0 +1,135 @@
...
+    populateStablehloLegalizeQuantizedOpToQDQPatterns(&patterns_, context);
+    patterns = std::move(patterns_);
+    return success();
+  }
+
+  void runOnOperation() override {
+    auto func = getOperation();
+    if (failed(applyPatternsAndFoldGreedily(func, patterns, config))) {
+      func.emitError("Failed to converge StablehloCanonicalizeDynamism in ")
+          << config.maxIterations << " iterations";
```
GleasonK, line 100: `signalPassFailure();`

GleasonK (approved):

PR #2395, opened 2024-06-15
Fix legalization of quantized Min/Max op to Int ops 

sdasgup3:
## Problem

Consider the following quantized stablehlo max operation

```
func.func @max_per_tensor_diff_quant_parameters(%arg0: tensor<!quant.uniform<i8:f32,1.0:0>>, %arg1: tensor<!quant.uniform<i8:f32,2.0:1>>) ->  tensor<!quant.uniform<i8:f32,3.0:2>> {
  %0 = "stablehlo.maximum"(%arg0, %arg1) : (tensor<!quant.uniform<i8:f32,1.0:0>>, tensor<!quant.uniform<i8:f32,2.0:1>>) -> tensor<!quant.uniform<i8:f32,3.0:2>>
  func.return %0 : tensor<!quant.uniform<i8:f32,3.0:2>>
}
```

Note that the quantization parameters (scales and zps) for operands and results are different.  

Currently [--stablehlo-legalize-quant-to-int](https://github.com/openxla/stablehlo/blob/main/stablehlo/transforms/StablehloLegalizeQuantToInt.cpp) legalizes this as follows:

```
func.func @max_per_tensor_diff_quant_parameters(%arg0: tensor<i8>, %arg1:  tensor<i8>) ->   tensor<i8> { 
  %0 = stablehlo.maximum %arg0, %arg1:  tensor<i8>
  func.return %0 :  tensor<i8>
```

which only makes sense when the quantization parameters is same for all operands and results, otherwise it is buggy ([why](reason)). 

## Proposed solution

The PR adds a check to allow the legalization only when  "the quantization parameters is same for all operands and results". Else error out.

### Appendix: Reason that it is a problem
Consider `a` and `b` are the input quantized values with  scale/ zp as `[s1, z1]` and `[s2, z2]` resp, and `c` is the output with scal/zp as `[s3, z3]`

we have from the semantics of the operation
```
(c - z3)*s3 = max[ (a-z1)*s1, (a-z2)*s2 ] 
```

Only if `z1=z2=z3`, and `s1=s2=s3`, we can simplify the above as `c = Max(a,b)` which is what the current legalization is doing.

### DIFF
--- a/stablehlo/transforms/StablehloLegalizeQuantToInt.cpp
+++ b/stablehlo/transforms/StablehloLegalizeQuantToInt.cpp
@@ -1247,6 +1247,18 @@ class ConvertGenericOp : public ConversionPattern {
       return failure();
     }
 
+    if (isa<stablehlo::MinOp, stablehlo::MaxOp>(op)) {
+      auto lhsType = getPerTensorType(op->getOperandTypes()[0]);
+      auto rhsType = getPerTensorType(op->getOperandTypes()[1]);
+      auto resultType = getPerTensorType(op->getResultTypes()[0]);
+      if (lhsType != rhsType || lhsType != resultType) {
+        op->emitError(op->getName().getStringRef() +
+                      " with different quantization parameters for operands and"
+                      " results is not supported.");
+        return failure();
+      }
+    }
+
     // Determine new result type: use storage type for uq types; use original
     // type otherwise.
     SmallVector<Type, 4> newResultTypes;
--- a/stablehlo/transforms/StablehloLegalizeQuantToInt.cpp
+++ b/stablehlo/transforms/StablehloLegalizeQuantToInt.cpp
@@ -1248,14 +1248,15 @@ class ConvertGenericOp : public ConversionPattern {
     }
 
     if (isa<stablehlo::MinOp, stablehlo::MaxOp>(op)) {
+      // Min/max only support per-tensor quantization.
       auto lhsType = getPerTensorType(op->getOperandTypes()[0]);
       auto rhsType = getPerTensorType(op->getOperandTypes()[1]);
       auto resultType = getPerTensorType(op->getResultTypes()[0]);
       if (lhsType != rhsType || lhsType != resultType) {
-        op->emitError(op->getName().getStringRef() +
-                      " with different quantization parameters for operands and"
-                      " results is not supported.");
-        return failure();
+        return op->emitError(
+            op->getName().getStringRef() +
+            " with different quantization parameters for operands and"
+            " results is not supported.");
       }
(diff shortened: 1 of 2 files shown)


stablehlo/transforms/StablehloLegalizeQuantToInt.cpp
```diff
@@ -1247,6 +1247,18 @@ class ConvertGenericOp : public ConversionPattern {
       return failure();
     }
 
+    if (isa<stablehlo::MinOp, stablehlo::MaxOp>(op)) {
```
sdasgup3, line 1250: That this point of code we know that Max/Min op is illegal, which means one of the operands or results has a quantized type. 


Together with min/max constraints

```
`lhs` | tensor or per-tensor quantized tensor
`rhs` | tensor or per-tensor quantized tensor 
`result` | tensor or per-tensor quantized tensor 
`baseline_type(lhs) = baseline_type(rhs) = baseline_type(result)`.
```
we know that the type of lhs/rhs/result should be per-tensor quantized.

stablehlo/transforms/StablehloLegalizeQuantToInt.cpp
```diff
@@ -1247,6 +1247,18 @@ class ConvertGenericOp : public ConversionPattern {
...
+      auto lhsType = getPerTensorType(op->getOperandTypes()[0]);
+      auto rhsType = getPerTensorType(op->getOperandTypes()[1]);
+      auto resultType = getPerTensorType(op->getResultTypes()[0]);
+      if (lhsType != rhsType || lhsType != resultType) {
+        op->emitError(op->getName().getStringRef() +
+                      " with different quantization parameters for operands and"
+                      " results is not supported.");
+        return failure();
```
GleasonK, line 1258: `return op->emitError` ?

stablehlo/transforms/StablehloLegalizeQuantToInt.cpp
GleasonK: Maybe make this a code comment.

```
// Min/max only support per-tensor quantization.
```

GleasonK (approved):

stablehlo/transforms/StablehloLegalizeQuantToInt.cpp
sdasgup3: done.

stablehlo/transforms/StablehloLegalizeQuantToInt.cpp
sdasgup3: done.

quanwanandy (approved):

PR #2403, opened 2024-06-20
[RFC] Add support for int2 in StableHLO

GleasonK:
To be shared to OpenXLA Discuss shortly.

### DIFF
--- a/rfcs/20240720-int2.md
+++ b/rfcs/20240720-int2.md
@@ -1,0 +1,28 @@
+# FP8 in StableHLO
+
+Status: In review<br/>
+Initial version: 07/20/2024<br/>
+Last updated: 07/20/2024<br/>
+
+# Introduction
+
+Sub-byte integer types are commonly used for model quantization. StableHLO
+currently supports signed and unsigned 4-bit integer types. Here we propose to
+extend the specification to cover 2-bit integer types as well.
+
+# Proposal
+
+Change the Types section of the spec to include si2 (signed 2-bit integer) and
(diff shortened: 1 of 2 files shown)


rfcs/20240720-int2.md
```diff
@@ -0,0 +1,28 @@
...
+currently supports signed and unsigned 4-bit integer types. Here we propose to
+extend the specification to cover 2-bit integer types as well.
+
+# Proposal
+
+Change the Types section of the spec to include si2 (signed 2-bit integer) and
+ui2 (unsigned 2-bit integer):
+
+```
+SignedIntegerType ::= 'si2' | 'si4' | 'si8' | 'si16' | 'si32' | 'si64'
```
GleasonK, line 19: @sdasgup3 interested in your thoughts on support for these new types as storage types for uniform quantizatized type. If there is intent to represent the decomposed integer math probably makes sense to add full support for i2 storage?

sdasgup3: Looks good to me. 
>> If there is intent to represent the decomposed integer math probably makes sense to add full support for i2 storage?

Yes, that is very much the intend. WIth `si2` support in StableHLO,  `stablehlo-legalize-qunt-to-int` should be ready ([cs](https://github.com/openxla/stablehlo/blob/91ed649613e5a51926045b4fd88575ab7a20a33e/stablehlo/transforms/StablehloLegalizeQuantToInt.cpp#L1290)) to express the decomposition in that type.

sdasgup3 (approved):

PR #2409, opened 2024-06-22
Add complex acos. Improve accuracy of real acos.

pearu:
- Add complex acos support using the modified [Hull et al algorithm](https://dl.acm.org/doi/10.1145/275323.275324)
  The accuracy characteristics for complex64 dtype are as follows:
  ```
  ULP difference == 0 count is 839342
  ULP difference == 1 count is 162239
  ULP difference == 2 count is 420
  ULP difference == 3 count is 0
  ULP difference >= 4 count is 0
  ```

- Improve the accuracy of real acos as follows (float32):
  ```
                         current main  -> this PR
  ULP difference == 0 count is 973884  -> 983681
  ULP difference == 1 count is 26010   -> 16320
  ULP difference == 2 count is 58      -> 0
  ULP difference == 3 count is 12      -> 0
  ULP difference >= 4 count is 37      -> 0
  ```

PR #2411, opened 2024-06-24
Improve accuracy of complex and real acosh

pearu:
- Improve the accuracy of complex acosh as follows (complex64):
  ```
                         current main  -> this PR
  ULP difference == 0 count is 141180  -> 839342
  ULP difference == 1 count is 10145   -> 162239
  ULP difference == 2 count is 1929    -> 420
  ULP difference == 3 count is 826     -> 0
  ULP difference >= 4 count is 847921  -> 0
  ```

- Improve the accuracy of real acosh as follows (float32):
  ```
                           current main  -> this PR
  ULP difference == 0 count is 966912    -> 993686
  ULP difference == 1 count is 33008     -> 6303
  ULP difference == 2 count is 49        -> 12
  ULP difference == 3 count is 13        -> 0
  ULP difference >= 4 count is 19        -> 0
  ```

This PR is created on top of https://github.com/openxla/stablehlo/pull/2409


stablehlo/transforms/ChloDecompositionPatternsMath.td
```diff
@@ -384,3 +384,544 @@ def : Pat<(CHLO_AsinOp NonComplexElementType:$x),
...
+                 $safe_max,
+                 (StableHLO_ConstantLike<"1000000000000.0"> $signed_x)),
+                StableHLO_ComparisonDirectionValue<"LT">,
+                (STABLEHLO_DEFAULT_COMPARISON_TYPE)),
+              (StableHLO_MulOp
+                $safe_max,
+                (StableHLO_ConstantLike<"1e-06"> $signed_x)),
+              (StableHLO_MulOp
+                $safe_max,
+                (StableHLO_ConstantLike<"100.0"> $signed_x))),
```
GleasonK, line 559: Very difficult to understand what's going on in this decomposition, especially with these large/small constants which don't appear in the above formula. I'm not sure how these behave with different (smaller) datatypes?

Any way to simplify this to read more like the math? Can we generate other CHLO ops in the decomposition? I.e. I believe we can generate `CHLO_Acos(x/a)` as a part of this instead of decomposing inline (assuming that's the real portion which would then be decomposed by the following decomp).

pearu: > Very difficult to understand what's going on in this decomposition, especially with these large/small constants which don't appear in the above formula.

These constants are defined in:
https://github.com/pearu/stablehlo/blob/7c655d3b5df4c6a641590299f7626a6c9e8a5fcd/stablehlo/transforms/ChloDecompositionPatternsMath.td#L124
(`100.0` is `1e2`).

> I'm not sure how these behave with different (smaller) datatypes?

This is a good question. The validity of constant `safe_max_opt` is currently tested against float32 and float64 data types. Since the definition of `safe_max_opt` worked fine for float32 and float64, I'd expect that the used scaling works fine also for `float16`. However, I agree that the validity of `safe_max_opt` needs to be tested against float16 as well. I propose to address this in a follow-up PR.

For bfloat16, we don't have tests because numpy does not support such dtype. That said, the tests for bfloat16 could be still generated, this just needs to be addressed within functional_algorithms (issue https://github.com/pearu/functional_algorithms/issues/13) by avoiding the convenience of numpy availability...

> Any way to simplify this to read more like the math? Can we generate other CHLO ops in the decomposition? I.e. I believe we can generate CHLO_Acos(x/a) as a part of this instead of decomposing inline (assuming that's the real portion which would then be decomposed by the following decomp).

Yes, there is a way to simplify the currently generated code because the algorithm for all the four complex functions - asin, asinh, acos, acosh - use the same kernel algorithm (see below).

I agree that reading this generated code for math is difficult. Instead, I'd suggest reading the original definitions of these algorithms which expose the mathematical definitions of the algorithms in much more clearly. For instance,
- https://github.com/pearu/functional_algorithms/blob/df060980aa5347bb695a66be44232a6bbae91bfe/functional_algorithms/algorithms.py#L257 defines the kernel algorithm `complex_asin_acos(z: complex)` mentioned above
- `asin(z: complex)` is evaluated as `complex_asin_acos(z)[0]`
- `acos(z: complex)` is evaluated as `complex_asin_acos(z)[1]`
- `asinh(z: complex)` is evaluated as `-I * asin(I * z)` (see https://github.com/pearu/functional_algorithms/blob/df060980aa5347bb695a66be44232a6bbae91bfe/functional_algorithms/algorithms.py#L441)
- `acosh(z: complex)` is evaluated as `sqrt(z - 1) / sqrt(1 - z) * acos(z)` (see https://github.com/pearu/functional_algorithms/blob/df060980aa5347bb695a66be44232a6bbae91bfe/functional_algorithms/algorithms.py#L517)

It would be possible to refactor the code generation for StableHLO so that the kernel algorithm would be exposed as the only "difficult-to-read" decomposition and the decompositions of asin, acos, asinh, acosh would call the kernel operation as defined in the above in FA. Could we address this as a follow-up task as it would require refactoring the currently landed implementation of `asin` as well?

What do you think?

pearu: Meanwhile, I have implemented `asin_acos_kernel` in functional_algorithms (https://github.com/pearu/functional_algorithms/pull/14) that allows representing all functions - asin, acos, asinh, acosh - in terms of a single kernel function. It will be available in the next release of functional_algorithms.

GleasonK: > Yes, there is a way to simplify the currently generated code because the algorithm for all the four complex functions - asin, asinh, acos, acosh - use the same kernel algorithm (see below).

This would be a great refactor. The more re-use in decompositions the better, if we can generate other CHLO ops used in the intermediate phases would be great, if we want to add a CHLO_Hypot (if that would make codegen smaller/more readable) we can, etc. Just thinking about general maintenance of these patterns.

In the meantime yes, we can make this a followup.

GleasonK (approved):

stablehlo/transforms/ChloDecompositionPatternsMath.td
pearu: Resolving via followup task https://github.com/pearu/functional_algorithms/issues/15

PR #2415, opened 2024-06-25
Extend CustomCallOp backend_config to take a DictionaryAttr

abhigunj:
* updated custom_call op def
* Added logic to fail a downgrade from 1.3.0 to previous versions, if the `custom_call` op's `backend_config` is a `DictionaryAttr.`


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -477,6 +477,24 @@ LogicalResult CustomCallOp::verify() {
              << "operand part has type " << operandPart
              << " and output part has type " << outputPart;
   }
+
+  // Check backend_config attribute.
+  if (auto backendConfig = getBackendConfig()) {
+    if (getApiVersion() == CustomCallApiVersion::API_VERSION_TYPED_FFI) {
+      // Typed FFI custom calls require `backend_config` to be a DictionaryAttr.
+      if (isa<mlir::StringAttr>(*backendConfig))
```
sdasgup3, line 485:
```
if (!isa<mlir::DictionaryAttr>(*backendConfig)) {
 // error
}
```

Directly cehck what is expected here?

GleasonK: +1, and `!isa<StringAttr>` for the other branch

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -477,6 +477,24 @@ LogicalResult CustomCallOp::verify() {
...
+        return emitOpError()
+               << "unsupported user-encoded backend config,"
+                  " backend config must be a dictionary attribute.";
```
GleasonK, line 488: "backend_config for api_version " << api_version << " must be a dictionary"

and below

"backend_config for api_version " << api_version << " must be a string".

stablehlo/dialect/StablehloOps.td
```diff
@@ -2339,13 +2339,35 @@ def StableHLO_CustomCallOp: StableHLO_Op<"custom_call",
     Encapsulates an implementation-defined operation `call_target_name` that
     takes `inputs` and `called_computations` and produces `results`.
 
+    If XLA runtime is enabled for a backend, then custom calls use the runtime
+    custom call calling convention to call into the external functions. This
+    calling convention defines an ABI for encoding arguments, attributes and
+    results.
+
```
GleasonK, line 2346: Don't need any XLA-specific details in StableHLO spec, delete.

stablehlo/dialect/StablehloOps.td
```diff
@@ -2339,13 +2339,35 @@ def StableHLO_CustomCallOp: StableHLO_Op<"custom_call",
...
+    Depending on the API version there are two ways to pass extra bits of static
+    information to the external function:
```
GleasonK, line 2348: A bit of this info is XLA specific and a bit confusing - i.e. what does this mean: "`API_VERSION_TYPED_FFI` custom calls only supported if XLA uses XLA runtime."

From an IR perspective, all we need to mention is:

```
Depending on the API version there are two ways to pass extra bits of static
information to the external function:

  1. Use `API_VERSION_TYPED_FFI` which allows passing a dictionary attribute.
  2. Use a previous API version with a StringAttr to encode backend config.
```

stablehlo/dialect/StablehloOps.td
```diff
@@ -2339,13 +2339,35 @@ def StableHLO_CustomCallOp: StableHLO_Op<"custom_call",
...
+
     See:
     https://github.com/openxla/stablehlo/blob/main/docs/spec.md#custom_call
 
     Example:
     ```mlir
     %results = stablehlo.custom_call @foo(%input0) {
-      backend_config = "bar",
+      backend_config = {bar = 42 : i32},
+      // api_version 4 is to express backend_config as a dictionary attribute
```
GleasonK, line 2369: Don't need the code comment with the details above

stablehlo/dialect/StablehloOps.td
```diff
@@ -2379,6 +2401,21 @@ def StableHLO_CustomCallOp: StableHLO_Op<"custom_call",
...
+      if (apiVersion == mlir::stablehlo::CustomCallApiVersion::API_VERSION_TYPED_FFI) {
+        return DictionaryAttr::get(getContext());
+      } else {
+        return StringAttr::get(getContext(), "");
+      }
+    }
+
+    mlir::Attribute getResolvedBackendConfig() {
+      return getBackendConfig().value_or(buildDefaultValue());
+    }
```
GleasonK, line 2417: Let's just declare the method in `extraClassDeclaration`, search for "CustomCall" in `StablehloOps.cpp` and implement it along with the other custom call impls there.

Also lets consolidate this to a single method called `getBackendConfigOrDefault` which has an impl like (untested code):

```
Attribute CustomCallOp::getBackendConfigOrDefault() {
  if (auto attr = getBackendConfigAttr()) return attr;
  if (getApiVersion() == CustomCallApiVersion::API_VERSION_TYPED_FFI) {
    return DictionaryAttr::get(getContext());
  }
  return StringAttr::get(getContext(), "");
}
```

GleasonK (changes requested):

stablehlo/dialect/StablehloOps.td
```diff
@@ -2340,13 +2340,19 @@ def StableHLO_CustomCallOp: StableHLO_Op<"custom_call",
...
+
     See:
     https://github.com/openxla/stablehlo/blob/main/docs/spec.md#custom_call
 
     Example:
     ```mlir
     %results = stablehlo.custom_call @foo(%input0) {
-      backend_config = "bar",
+      backend_config = {bar = 42 : i32},
+      api_version = 4 : i32,
```
sdasgup3, line 2355: Can you also remove the comment from the spec.md file mainly for the consistency with the ODS and spec? As we are not providing any explicit specification of API version, it does not matter, from the exemplification POV, which API version we are using.

``` 
 backend_config = {bar = 42 : i32},
  // api_version 4 is to express backend_config as a dictionary attribute
  api_version = 4 : i32,
```

sdasgup3 (approved):

stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.mlir
```diff
@@ -1299,7 +1299,6 @@ func.func public @op_custom_call_empty_result_layout(%arg0: tensor<i64>) -> tens
   // CHECK-SAME: }> : (!vhlo.tensor_v1<!vhlo.i64_v1>) -> !vhlo.tuple_v1<>
   %0 = "stablehlo.custom_call"(%arg0) <{
     api_version = 2 : i32,
-    backend_config = "",
```
GleasonK, line 1302: Add a API v4 w/dict testcase

stablehlo/dialect/VhloOps.cpp
```diff
@@ -333,6 +333,17 @@ LogicalResult verifyConstraint_0_17_0(mlir::Operation* op,
...
   return success();
 }
+
+LogicalResult verifyConstraint_1_1_0(mlir::Operation* op,
+                                     Version targetVersion) {
+  auto customCallOp = dyn_cast<mlir::vhlo::CustomCallOpV1>(op);
+  if (customCallOp &&
+      isa<mlir::DictionaryAttr>(customCallOp.getBackendConfig()) &&
+      targetVersion < Version(1, 1, 0))
+    return failure();
```
GleasonK, line 343: Also check if API is less than v4 for downgrade

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -493,6 +507,16 @@ void CustomCallOp::getEffects(
   effects.emplace_back(MemoryEffects::Read::get());
 }
 
+mlir::Attribute CustomCallOp::getBackendConfigOrDefault() {
+  if (getBackendConfig().has_value()) return getBackendConfig().value();
```
GleasonK, line 511: nit: let's avoid calling the getter twice 

```
auto backendConfig = getBackendConfig();
if (backendConfig.has_value()) return backendConfig.value();
```

stablehlo/dialect/VhloOps.cpp
```diff
@@ -333,6 +333,20 @@ LogicalResult verifyConstraint_0_17_0(mlir::Operation* op,
     return failure();
   return success();
 }
+
+LogicalResult verifyConstraint_1_2_1(mlir::Operation* op,
+                                     Version targetVersion) {
+  auto customCallOp = dyn_cast<mlir::vhlo::CustomCallOpV1>(op);
```
GleasonK, line 339: Probably don't need `mlir::vhlo` I believe we're in that namespace. Also this will only ever be called on CustomCallOp so use `cast`:

```
auto customCallOp = cast<CustomCallOpV1>(op);
if (targetVersion < Version(1, 2, 1) && customCallOp...) { return failure(); }
return success();
```

Also I'm not sure how enum attrs work, do we need the cast or does the following work as an attr comparison?

```
(customCallOp.getApiVersion() == CustomCallApiVersionV1::API_VERSION_TYPED_FFI)
```

stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.mlir
```diff
@@ -178,6 +178,18 @@ func.func @attr_dict() attributes {stablehlo.attr = {attr1 = 1 : i32, attr2 = 2
   return
 }
 
+// CHECK-LABEL: "attr_custom_call_api_version_typed_ffi"
+// CHECK-NEXT: (%[[ARG0:.*]]: {{.*}})
+func.func @attr_custom_call_api_version_typed_ffi(%arg0: tensor<f32>) -> tensor<f32> {
+  %0 = "stablehlo.custom_call"(%arg0) {
+    call_target_name = "foo",
```
GleasonK, line 185: Add a check for dictionary as well?

stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.mlir
```diff
@@ -178,6 +178,18 @@ func.func @attr_dict() attributes {stablehlo.attr = {attr1 = 1 : i32, attr2 = 2
   return
 }
 
+// CHECK-LABEL: "attr_custom_call_api_version_typed_ffi"
+// CHECK-NEXT: (%[[ARG0:.*]]: {{.*}})
+func.func @attr_custom_call_api_version_typed_ffi(%arg0: tensor<f32>) -> tensor<f32> {
```
GleasonK, line 183: PTAL at [vhlo_checklist.md](https://github.com/openxla/stablehlo/blob/main/docs/vhlo_checklist.md), we're missing a `1.3.0` test file. We're missing a few pieces, I've covered most of these in other comments, but we also need step 3 and 4, there should be a 1.3.0 compatibility test.

stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.mlir
```diff
@@ -178,6 +178,18 @@ func.func @attr_dict() attributes {stablehlo.attr = {attr1 = 1 : i32, attr2 = 2
...
+  %0 = "stablehlo.custom_call"(%arg0) {
+    call_target_name = "foo",
+    backend_config= {bar = 42 : i32},
+    // CHECK: api_version = #vhlo<api_version_v1 API_VERSION_TYPED_FFI>
+    api_version = 4 : i32
+  } : (tensor<f32>) -> tensor<f32>
+  func.return %0 : tensor<f32>
+}
```
GleasonK, line 191: Can you also add a test for `attr_custom_call_api_version_typed_ffi_default` which specifies API v4 but doesn't specify `backend_config`? I believe this will fail, I think we're missing some code in StablehloLegalizeToVhlo.cpp for inserting a proper default value for `dictionary|string`


Likely requires:

- [ ] Modify [here](https://github.com/openxla/stablehlo/blob/8a78ebb89bc6c5608f6f27a7a9ce687b3578b79d/stablehlo/transforms/StablehloLegalizeToVhlo.cpp#L698) in StableHLO->VHLO to instead call get `getOrDefault` method you wrote for populating the value.
- [ ] Modify [here](https://github.com/openxla/stablehlo/blob/8a78ebb89bc6c5608f6f27a7a9ce687b3578b79d/stablehlo/transforms/VhloLegalizeToStablehlo.cpp#L739) in VHLO->StableHLO if the value is a default value (empty str or empty dict)

stablehlo/tests/vhlo/vhlo_to_version_downgrade_invalid.1_1_0.mlir
```diff
@@ -13,3 +13,15 @@ func.func @type_ui2(%arg0: tensor<ui2>) -> tensor<ui2> {
   %0 = stablehlo.add %arg0, %arg0 : tensor<ui2>
   func.return %0 : tensor<ui2>
 }
+
+// -----
+
+func.func @custom_call_dictionary_attr(%arg0: tensor<f32>) -> tensor<f32> {
```
GleasonK, line 19: Testcase looks good, but after a version bump it should go in a new file: `stablehlo/tests/vhlo/vhlo_to_version_downgrade_invalid.1_2_0.mlir`, and we need to bump the minor version again to 1.3.0 since there is an opset change in this PR (will leave separate comment to track this).

We should also add a test for default valued dict (leave off backend config, but specify v4)

stablehlo/dialect/VhloOps.cpp
```diff
@@ -333,6 +333,20 @@ LogicalResult verifyConstraint_0_17_0(mlir::Operation* op,
     return failure();
   return success();
 }
+
+LogicalResult verifyConstraint_1_2_1(mlir::Operation* op,
```
GleasonK, line 337: All opset changes require a minor version bump. This should be 1.3.0. A few changes needed for this:

- [ ] Update Version.h to 1.3.0 since this is an opset change
- [ ] Update [VhloDialect.td](https://github.com/openxla/stablehlo/blob/8a78ebb89bc6c5608f6f27a7a9ce687b3578b79d/stablehlo/dialect/VhloDialect.td#L42) with the opset change.
- [ ] Rename this method to `_1_3_0`

GleasonK (changes requested):

stablehlo/dialect/VhloOps.cpp
abhigunj: yeh, mlir::vhlo namespace path is not required here. 

`getApiVersion() returns `mlir::attribute` for vhlo op, not `CustomCallApiVersionV1` enum. Stablehlo op, does return `CustomCallApiVersion` enum for `getApiVersion()`

stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.mlir
abhigunj: I validated this during testing and and test passed without the `backend_config` for APIV4

stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.mlir
abhigunj: Thank you, I incorrectly assumes, since we are not creating new version of the op, we don't need these changes. 

I will update the PR.

stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.mlir
abhigunj: will update the PR with required StablehloLegalizeToVhlo.cpp changes

stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.mlir
abhigunj: completed remaining tasks from the checklist.

stablehlo/dialect/VhloOps.cpp
abhigunj: Done

stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.mlir
GleasonK: >  I incorrectly assumed, since we are not creating new version of the op, we don't need new version changes

Think of versioning more from the perspective of compatibility -- are we introducing something that we can't deserialize on a 1mo+ old release of StableHLO? Probably needs a new version.

stablehlo/tests/vhlo/stablehlo_legalize_to_vhlo.mlir
abhigunj: updated  StableHLO->VHLO and VHLO->StableHLO change. 
added APIv4 test without the backend_config.

stablehlo/tests/vhlo/vhlo_to_version_downgrade_invalid.1_2_0.mlir
```diff
@@ -0,0 +1,11 @@
...
+
+func.func @custom_call_dictionary_attr(%arg0: tensor<f32>) -> tensor<f32> {
+// expected-error @+1 {{failed to legalize operation 'vhlo.custom_call_v1' that was explicitly marked illegal}}
+%0 = "stablehlo.custom_call"(%arg0) {
+    call_target_name = "foo",
+    api_version = 4 : i32,
+    backend_config={foo = 42 : i32}
+  } : (tensor<f32>) -> tensor<f32>
+  func.return %0 : tensor<f32>
+}
```
GleasonK, line 11: Let's also add an invalid test for default_valued backend config

GleasonK (approved):
LGTM, a few nits left

stablehlo/tests/vhlo/vhlo_to_version_downgrade_invalid.1_1_0.mlir
GleasonK: > We should also add a test for default valued dict (leave off backend config, but specify v4)

Moved this feedback to a new comment.

abhigunj: 1. moved tests to  a new file. 
2. added similar test but without backend_config

stablehlo/dialect/VhloOps.cpp
GleasonK: > direct attr comparison did not work.

Makes sense, we'd really need to be comparing to:

```
vhlo::CustomCallApiVersionV1Attr::get(getContext(),vhlo::CustomCallApiVersionV1::API_VERSION_TYPED_FFI)
```

But I don't think we should create a new attr just for the sake of comparison, lgtm as-is!

stablehlo/tests/vhlo/vhlo_to_version_downgrade_invalid.1_2_0.mlir
abhigunj: updated, while you were reviewing :) . closing this.

stablehlo/tests/vhlo/vhlo_to_version_downgrade_invalid.1_2_0.mlir
```diff
@@ -0,0 +1,22 @@
...
+// -----
+
+func.func @custom_call_dictionary_attr(%arg0: tensor<f32>) -> tensor<f32> {
+// expected-error @+1 {{failed to legalize operation 'vhlo.custom_call_v1' that was explicitly marked illegal}}
+%0 = "stablehlo.custom_call"(%arg0) {
+    call_target_name = "foo",
+    api_version = 4 : i32
+  } : (tensor<f32>) -> tensor<f32>
+  func.return %0 : tensor<f32>
+}
```
GleasonK, line 22: Missing newline at EOF (line_whitespace failing)

run:

```
./build_tools/github_actions/lint_whitespace_checks.sh -f
```

PR #2416, opened 2024-06-25
Use ConfinedAttr to verify attr range is positve or nonnegative

ghpvnist:
I recently came across [confining attributes](https://mlir.llvm.org/docs/DefiningDialects/Operations/#confining-attributes), and it's a nice feature that we should leverage to maximize ODS strengths.

I'm still looking into whether we can use them for array attributes. If possible, we may be able to remove tons of manual verification.


GleasonK (changes requested):
LGTM, these are super nice. Two notes:

1. Lets not delete existing tests, I know we don't _add_ tests for ODS machinery generally, but no sense in removing cheap tests to ensure no behavior changes (which has happened before that upstream constraints stop verifying).
2. Make sure these are ported to MHLO (I think this will cause test failures since verifiers have changed)

GleasonK (approved):

PR #2420, opened 2024-07-01
Cleanup StableHLO bytecode patches used for temporary pjrt compatibility

### LINKED ISSUE #2216: Clean up GenericDenseArrayAttr after properties migration
### Request description

Some users of StableHLO are still migrating to use VHLO, so we temporarily have IR downgrades in PjRT for DenseArray vs DenseElements: [openxla/xla > xla/pjrt/mlir_to_hlo.cc](https://github.com/openxla/xla/blob/87b3e79832d61561c5525a2e7f08a57b2fb4839f/xla/pjrt/mlir_to_hlo.cc#L180)

These downgrades don't play well with properties, so we need to use a wrapper around DenseArray which uses Attribute as the backing storage type. Once PjRT plugins which require version skew management have migrated to use VHLO we can clean up all use of `GenericDense*ArrayAttr`.

GleasonK:
Several hacks/patches were put in place to preserve StableHLO compatibility for PJRT plugins.

Recently PJRT was switched to send VHLO, meaning we can clean up all these patch fixes.

Closes #2216

### DIFF
--- a/stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
+++ b/stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
@@ -840,7 +840,6 @@ struct TransposeOpToTransposeConverter final
     Value emptyTensor =
         getEmptyTensorFor(rewriter, loc, resultTy, op, adaptor.getOperands());
 
-    // TODO(#2216) Cleanup Attribute -> DenseArrayAttr
     rewriter.replaceOpWithNewOp<linalg::TransposeOp>(
         op, adaptor.getOperand(), emptyTensor, op.getPermutationAttr(),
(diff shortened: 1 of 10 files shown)


stablehlo/conversions/linalg/transforms/StablehloLegalizeToLinalg.cpp
```diff
@@ -842,8 +842,7 @@ struct TransposeOpToTransposeConverter final
 
     // TODO(#2216) Cleanup Attribute -> DenseArrayAttr
```
ghpvnist, line 843: Looks like this comment can be removed too since we are reverting the change back to regular arrays.

ghpvnist (approved):
Mostly LGTM. Can we also close #2216 since we are getting rid of the patch diff?

PR #2430, opened 2024-07-10
Refactor CMakeList to work for all build types

GleasonK:
This setup is based on: [How_to_Build_your_own_MLIR_Dialect.pdf](https://www.google.com/url?sa=D&q=https%3A%2F%2Farchive.fosdem.org%2F2023%2Fschedule%2Fevent%2Fmlirdialect%2Fattachments%2Fslides%2F5740%2Fexport%2Fevents%2Fattachments%2Fmlirdialect%2Fslides%2F5740%2FHow_to_Build_your_own_MLIR_Dialect.pdf)

```
# There are 3 build modes, one will be set to ON
#  - Standalone: Build MLIR as a part of StableHLO
#  - External: StableHLO built as an external LLVM project (XLA/MHLO uses this)
#  - Embedded: StableHLO built as a part of another MLIR project (torch-mlir uses this)
#
# If building as part of another project, let it handle the MLIR dependency.
# The dependent project might use a bundled version of MLIR instead of installing.
```

Tested using `mlir-hlo` repo. No changes to the `EMBEDDED` build so should not impact torch-mlir.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,16 +53,24 @@ option(STABLEHLO_ENABLE_LLD "Use LLD as the linker if available" OFF)
 #-------------------------------------------------------------------------------
 # Project setup and globals
 #-------------------------------------------------------------------------------
-set(STABLEHLO_EXTERNAL_PROJECT_BUILD OFF)
 
-if(NOT (CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR) AND NOT MLIR_BINARY_DIR)
-  # Building as part of LLVM via the external project mechanism.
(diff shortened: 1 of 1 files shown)


GleasonK:
cc @qingyunqu -- I don't think this should change the torch-mlir `STABLEHLO_BUILD_EMBEDDED` workflow, but would be good to confirm if you have a moment to patch this in

sdasgup3 (approved):

qingyunqu:
> cc @qingyunqu -- I don't think this should change the torch-mlir `STABLEHLO_BUILD_EMBEDDED` workflow, but would be good to confirm if you have a moment to patch this in

Ok, I will confirm this when update stablehlo in torch-mlir.

PR #2432, opened 2024-07-11
Add convenience method to CustomCallOp for checking empty backend config

GleasonK:
This logic is used many times in openxla/xla, so decided to turn it into a method

### DIFF
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -518,6 +518,16 @@ mlir::Attribute CustomCallOp::getBackendConfigOrDefault() {
   return StringAttr::get(getContext(), "");
 }
 
+// Returns if the backend config is unset, or if empty dict / string attribute.
+bool CustomCallOp::hasEmptyBackendConfig() {
+  if (!getBackendConfig().has_value()) return true;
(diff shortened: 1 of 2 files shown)


stablehlo/dialect/StablehloOps.cpp
```diff
@@ -518,6 +518,16 @@ mlir::Attribute CustomCallOp::getBackendConfigOrDefault() {
   return StringAttr::get(getContext(), "");
 }
 
+// Returns if the backend config is unset, or if empty dict / string attribute.
+bool CustomCallOp::hasEmptyBackendConfig() {
+  if (!getBackendConfig().has_value()) return true;
```
abhigunj, line 523: Thanks for taking care of the change. 

nit:  we can remove the line `if (!getBackendConfig().has_value()) return true;` to make API simpler and  call  `getBackendConfigOrDefault` first..

abhigunj (approved):

stablehlo/dialect/StablehloOps.cpp
GleasonK: Good point! I'm thinking to leave this as-is, the `OrDefault` method will construct an object if its empty, so in this current state its more efficient and doesn't have that null-attr overhead

PR #2438, opened 2024-07-16
Add better printing in reference interpreter, add tolerance to check almost eq

GleasonK:
Background: I was running interpreter testcases and wanted some syntax I could copy-paste into python. Now tensors in the interpreter print in a python-array-compatible syntax, wrapped in a `tensor{...}` descriptor.

```
$ cat /tmp/t.mlir 
func.func @main() -> tensor<2x2xi8> {
  %matrix = stablehlo.constant dense<[[1, 2], [3, 4]]> : tensor<2x2xi8>
  func.return %matrix : tensor<2x2xi8>
}

$ stablehlo-translate --interpret /tmp/t.mlir 
tensor<2x2xi8> {
  [
    [1, 2],
    [3, 4]
  ]
}
```

Also as a part of this work I needed to modify some testcase tolerances.

PR #2439, opened 2024-07-16
Add CHLO decomposition pipeline and new interpreter check calls

GleasonK:
These are changes in preparation for refreshing all of `testdata` and adding CHLO-based testdata files.

- Add more check ops to supported custom_calls in translate -- `@check.expect_close/almost_equal/check_eq`
- Add `--chlo-pre-serialization-pipeline` to be run on programs with CHLO ops to decompose them along with their shape computation to StableHLO ops.

Note: This is based on https://github.com/openxla/stablehlo/pull/2438 for the changes to CheckOps in that PR.


stablehlo/tools/StablehloTranslateMain.cpp
```diff
@@ -97,11 +97,35 @@ class StablehloTranslateInterpreterFallback
...
     llvm::StringRef funcName = op.getParentOfType<func::FuncOp>().getSymName();
     if (auto customCall = dyn_cast<stablehlo::CustomCallOp>(op)) {
-      if (customCall.getCallTargetName() == "check.eq") {
+      auto callTarget = customCall.getCallTargetName();
+      if (callTarget == "check.eq") {
         auto status = evalCustomCallCheckEq(customCall, scope);
         return stablehlo::wrapFallbackStatus(
             std::move(status), funcName, "stablehlo.custom_call(@check.eq)");
       }
+      if (customCall->getNumOperands() == 2) {
```
sdasgup3, line 106: Not sure why we have the num operand chck here?
Can there be `custom_calls` with `callTarget == {check.expect_close, check.expect_eq, check.expect_almost_eq}` but with `!=2` operands?

stablehlo/tools/StablehloTranslateMain.cpp
```diff
@@ -97,11 +97,35 @@ class StablehloTranslateInterpreterFallback
...
+        // TODO: Replace with custom_call --> check dialect pass
+        auto runtimeLhs = scope.findTensor(customCall->getOperand(0));
+        auto runtimeRhs = scope.findTensor(customCall->getOperand(1));
+        if (callTarget == "check.expect_close") {
+          auto status =
+              stablehlo::check::evalExpectCloseOp(runtimeLhs, runtimeRhs, 0, 3);
```
sdasgup3, line 112: Can we have the constant like `0`, `3`, and `0.0001` as default arguments of `evalCustomCallCheckEq`?

stablehlo/tools/StablehloTranslateMain.cpp
GleasonK: No, but we don't validate these things for each op.. so those without 2 operands will throw the later "unsupported custom call" message.

Not ideal but also I think this is a temporary solution until we have a CC->Check dialect pass

stablehlo/tools/StablehloTranslateMain.cpp
GleasonK: I think yes, but also think that can be a followup

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -2480,9 +2480,13 @@ LogicalResult TupleOp::inferReturnTypes(
 void CompareOp::build(OpBuilder& builder, OperationState& result, Value lhs,
                       Value rhs, ComparisonDirection comparisonDirection,
                       ComparisonType compareType) {
+  ComparisonTypeAttr comparisonTypeAttr;
```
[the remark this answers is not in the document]
GleasonK: Not sure I follow, if the type is `NOTYPE` we want it to be empty `ComparisonTypeAttr()` so that the optional attr isn't populated at all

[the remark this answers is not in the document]
sdasgup3: nvm, I though the entire code is within the then block.

sdasgup3 (approved):

PR #2099, opened 2024-03-15
[RFC] Standardize Collective (AllGatherOp, AllReduceOp, AllToAllOp) ops to support variadic operand/result

abhigunj:
This RFC proposes standardizing collective (AllGatherOp, AllReduceOp, AllToAllOp) ops to  enable support for `multi-operand` and `multi-result`

Please review and provide you feedback.

PR #2443, opened 2024-07-17
Make "Rendezvous" variadic

abhigunj:
The PR allows each Process from the ProcessGrid to contribute more than one tensor at `ProcessGrid::rendezvous`. 
This is the first set of change, will be followed by  interpreter updates to collectives.  

Note: The PR does **not** make collectives interpreter variadic. 

Tested: 
1. No existing test failures indicate no change in behavior for ops using the `rendezvous`
2. The diff is  tested with new variadic interpreter for `all_reduce` op. Will upload the PR soon. 

ref: https://github.com/openxla/stablehlo/pull/2099


stablehlo/reference/ProcessGrid.h
```diff
@@ -250,7 +250,7 @@ class ProcessGrid {
   /// returned to all callers once the barrier has been reached by all StableHLO
   /// processes.
   RendezvousResult rendezvous(ProcessGroup processGroup, ChannelId channelId,
-                              ProcessId processId, const Tensor &operand);
+                              ProcessId processId, const SmallVector<Tensor> &operand);
```
sdasgup3: `ArrayRef`

stablehlo/reference/ProcessGrid.h
```diff
@@ -41,23 +41,23 @@ struct ProcessId;
...
-  void insert(ProcessId processId, Tensor tensor);
+  void insert(ProcessId processId, SmallVector<Tensor> tensor);
 
   /// Iterates through the map and returns the value associated with the key
   /// `processId`. If key is not found, return an empty `Tensor`.
-  Tensor lookup(ProcessId processId) const;
+  SmallVector<Tensor> lookup(ProcessId processId) const;
 
  private:
   /// Internal map representation of the result of `ProcessGrid::rendezvous`.
```
sdasgup3, line 59: In the comment `results`

stablehlo/reference/ProcessGrid.h
```diff
@@ -41,23 +41,23 @@ struct ProcessId;
...
-  RendezvousResult(std::map<ProcessId, Tensor> const &result);
+  RendezvousResult(std::map<ProcessId, SmallVector<Tensor>> const &results);
 
   /// Iterates through the (ProcessId, Tensor) map entires and returns a vector
   /// of Tensors sorted by ProcessId--(replicaId, partitionId) pair--in
   /// lexicographical order.
-  SmallVector<Tensor> getSortedTensors() const;
+  SmallVector<SmallVector<Tensor>> getSortedTensors() const;
 
   /// Inserts `tensor` into the map using the key `processId`.
```
sdasgup3, line 51: update the comment here and other places in the file.

sdasgup3:
Is there a way to test the `rendezvous` function in isolation?

abhigunj:
> Is there a way to test the `rendezvous` function in isolation?

`2. The diff is tested with new variadic interpreter for all_reduce op. Will upload the PR soon.`

you mean other than ^ test?

sdasgup3:
> > Is there a way to test the `rendezvous` function in isolation?
> 
> `2. The diff is tested with new variadic interpreter for all_reduce op. Will upload the PR soon.`
> 
> you mean other than ^ test?

I was wondering if it is possible to test the changes proposed in this  PR with multiple operands feeding the rendezvous. Its fine in case this can only to tested at the operation level (via the upcoming PR).

stablehlo/reference/ProcessGrid.cpp
```diff
@@ -223,16 +225,19 @@ SmallVector<Tensor> ProcessGrid::recv(ChannelId channelId,
 RendezvousResult ProcessGrid::rendezvous(ProcessGroup processGroup,
                                          ChannelId channelId,
                                          ProcessId processId,
-                                         const Tensor &operand) {
+                                         const ArrayRef<Tensor> operands) {
```
sdasgup3, line 228: avoid `const` ? 
`ArrayRef<T>` is non-owning and immutable: It doesn't manage the lifetime of the underlying data + we cannot modify the data is encapsulates.

sdasgup3 (approved):
one nit and lgtm.

stablehlo/reference/ProcessGrid.h
```diff
@@ -245,12 +245,12 @@ class ProcessGrid {
...
   ///
-  /// At the barrier, each StableHLO process contributes a tensor, and these
-  /// tensors are accumulated in `RendezvousResult` whose shared pointer is
-  /// returned to all callers once the barrier has been reached by all StableHLO
-  /// processes.
+  /// At the barrier, each StableHLO process contribute any number of tensors,
+  /// and these tensors are accumulated in `RendezvousResult` whose shared
+  /// pointer is returned to all callers once the barrier has been reached by
+  /// all StableHLO processes.
   RendezvousResult rendezvous(ProcessGroup processGroup, ChannelId channelId,
```
ghpvnist, line 252: How to assert that each process sends the same variadic sized args?

stablehlo/reference/ProcessGrid.h
```diff
@@ -35,29 +35,29 @@ namespace stablehlo {
 struct ProcessId;
 
 /// Represents a result of a `ProcessGrid::rendezvous` where multiple processes
-/// synchronize at a barrier and contribute a Tensor each.
-/// This class is pretty much a map from ProcessId to Tensor, with the
+/// synchronize at a barrier and contribute any number of Tensors each.
```
ghpvnist, line 38: "any number of Tensors" is not entirely accurate since we want all processes to send the same number of Tensors to the rendezvous.

ghpvnist (changes requested):

stablehlo/reference/ProcessGrid.h
abhigunj: Each process will contribute either 
1. all operands  or 
2. none, if it failed to send within 3 second time limit. 
If, timeout happens, we are erroring out. So we don't need  size check? lmk if I am missing any other scenario.

stablehlo/reference/ProcessGrid.h
abhigunj:
```
/// Represents a result of a `ProcessGrid::rendezvous` where multiple processes
/// synchronize at a barrier and contribute same number of Tensors.
/// This class is pretty much a map from ProcessId to set of Tensors, with the
/// map-like API.

```
wdyt?

stablehlo/reference/ProcessGrid.h
ghpvnist: Currently, we check that all processes contributed, but it doesn't enforce that the number of variadic args between processes are the same.

E.g. two processes are participating in variadic all_reduce, What if the first process sends 2 tensors while second process sends 3?

abhigunj: added a check to validate processes contributed same number of Tensors. Thank you.

ghpvnist (approved):
LGTM with minor comments.

stablehlo/reference/ProcessGrid.h
```diff
@@ -55,6 +55,10 @@ class RendezvousResult {
   /// `processId`. If key is not found, return an empty `SmallVector<Tensor>`.
   SmallVector<Tensor> lookup(ProcessId processId) const;
 
+  /// Iterates through the (ProcessId, SmallVector<Tensor>) map entires and
+  /// return true if all processes contributed same number of operand Tensors
+  bool hasMatchingOperandsCount() const;
```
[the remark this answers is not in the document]
abhigunj: if we move to private, ProcessGrid::rendezvous can't access it.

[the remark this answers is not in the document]
ghpvnist: Oh I see, I thought it belonged to ProcessGrid. Carry on!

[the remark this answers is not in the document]
abhigunj: cool, thanks for the review.

PR #2445, opened 2024-07-19
Python bindings for registering check dialect

sdasgup3:
Check dialect is an auxiliary dialect used in StableHLO repository for validation of StableHLO program evaluation. Currently there is no cleaner way to parse a module containing check dialect operations or to create them. One way to go around this is to textually modify the check dialect ops to normalize them to generic MLIR text  and allow unregistered dialect to pars that text. However, this PR provodes a cleaner way to process check dialect. 

This PR prepares for open sourcing some of the utilities, leveraged in https://github.com/openxla/stablehlo/pull/2404,  for auto-generating testdata formatted test files.

### DIFF
--- a/stablehlo/integrations/python/tests/check.py
+++ b/stablehlo/integrations/python/tests/check.py
@@ -1,0 +1,44 @@
+# Copyright 2024 The StableHLO Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


stablehlo/integrations/python/tests/check.py
```diff
@@ -0,0 +1,44 @@
...
+# pylint: disable=wildcard-import,undefined-variable
+
+from mlir import ir
+from mlir.dialects import check as check_dialect
+from mlir.dialects import stablehlo as stablehlo_dialect
+
+
+def run(f):
+  with ir.Context() as context:
+    check.register_dialect(context)
```
abhigunj, line 26:
```suggestion
    check_dialect.register_dialect(context)
```

stablehlo/integrations/python/tests/check.py
```diff
@@ -0,0 +1,44 @@
...
+    stablehlo_dialect.register_dialect(ctx)
```
abhigunj, line 27: `stablehlo_dialect.register_dialect(context)` ?

abhigunj (approved):

PR #2449, opened 2024-07-19
Revert acos changes for non complex numbers

GleasonK:
It looks like something may be broken in the `chlo.acos` lowering from functional algorithms. The complex lowering works just fine (has `SelectOp`).

Also I'm wondering how we didn't catch this / could have caught this. Perhaps we should have special cases for ops with limits..somehow? Open to iterating on this, either by dev policy on these CLs (ensure total code coverage) or if there's a way to auto generate cases would be good too.

Current behavior prior to revert: `chlo.acos(-1) --> 0`, expected behavior is `pi`.

### DIFF
--- a/stablehlo/transforms/ChloDecompositionPatterns.td
+++ b/stablehlo/transforms/ChloDecompositionPatterns.td
@@ -46,6 +46,39 @@ def StableHLO_ConstantLikeSmallestNormalizedValue : NativeCodeCall<
 // Unary op patterns.
 //===----------------------------------------------------------------------===//
 
+// Expand acos for non-complex arguments to MHLO dialect as follows:
+//   acos(x) = 2 * atan2(sqrt(1 - x^2), (1 + x))  if x != -1
+//           = pi                                 if x == -1
+//
+//   acos(x) = -(i * log(x + i * sqrt((1 + x) * (1 - x))))
+//
+// Note: Complex decomposition is in ChloDecompositionPatternsMath.td
+def : Pat<(CHLO_AcosOp NonComplexElementType:$input),
+  (StableHLO_SelectOp
+    (StableHLO_CompareOp
+      $input,
+      (StableHLO_ConstantLike<"-1"> $input),
+      StableHLO_ComparisonDirectionValue<"NE">,
+      (STABLEHLO_DEFAULT_COMPARISON_TYPE)
+    ),
+    (StableHLO_MulOp
+      (StableHLO_ConstantLike<"2"> $input),
+      (StableHLO_Atan2Op
+        (StableHLO_SqrtOp
+          (StableHLO_SubtractOp
+            (StableHLO_ConstantLike<"1"> $input),
+            (StableHLO_MulOp $input, $input)
+          )
+        ),
+        (StableHLO_AddOp
+          (StableHLO_ConstantLike<"1"> $input),
+          $input
+        )
+      )
+    ),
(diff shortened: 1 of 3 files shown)


GleasonK:
cc @pearu

abhigunj (approved):

sdasgup3 (approved):
Same comments as in the internal PR.

stablehlo/transforms/ChloDecompositionPatterns.td
```diff
@@ -46,6 +46,39 @@ def StableHLO_ConstantLikeSmallestNormalizedValue : NativeCodeCall<
 // Unary op patterns.
 //===----------------------------------------------------------------------===//
 
+// Expand acos for non-complex arguments to MHLO dialect as follows:
+//   acos(x) = 2 * atan2(sqrt(1 - x^2), (1 + x))  if x != -1
+//           = pi                                 if x == -1
+//
+//   acos(x) = -(i * log(x + i * sqrt((1 + x) * (1 - x))))
+//
```
pearu, line 54:
```suggestion
```
as it is unrelated.

stablehlo/transforms/ChloDecompositionPatterns.td
```diff
@@ -46,6 +46,39 @@ def StableHLO_ConstantLikeSmallestNormalizedValue : NativeCodeCall<
 // Unary op patterns.
 //===----------------------------------------------------------------------===//
 
+// Expand acos for non-complex arguments to MHLO dialect as follows:
+//   acos(x) = 2 * atan2(sqrt(1 - x^2), (1 + x))  if x != -1
```
pearu, line 50: Using `1-x^2` will be inaccurate when `abs(x)` is close to `1`. The corresponding cancellation errors are avoided when using `(1-x)*(1+x)`. I can address this in a follow-up PR if needed.

stablehlo/transforms/ChloDecompositionPatternsMath.td
```diff
@@ -643,17 +643,6 @@ def : Pat<(CHLO_AcosOp ComplexElementType:$z),
 //
 //       1 - x * x == (1 - x) * (1 + x)
 //
```
pearu, line 645: The comment should be removed in full:
```suggestion
```

Since ChloDecompositionPatternsMath.td is generated, editing the file directly is not recommended. However, in this particular case, I think it is OK as a temporary solution. I'll address this as a first thing when returning from my vacation.

pearu (approved):
I have a couple of comments that can be addressed in follow-up PRs after fixing acos in functional_algorithms.

pearu:
> Current behavior prior to revert: `chlo.acos(-1) --> 0`, expected behavior is `pi`.

This is related to https://github.com/pearu/functional_algorithms/issues/11, that is, in addition to log-uniform samples we should have tests for special argument values as well.

stablehlo/transforms/ChloDecompositionPatterns.td
GleasonK: deleted

stablehlo/transforms/ChloDecompositionPatterns.td
GleasonK: Created https://github.com/openxla/stablehlo/issues/2452 to track this work, in the meantime this can be seen as a revert

Issue #2433, opened 2024-07-11
Bug report on all_to_all implementation in stablehlo interpreter


Hatsunespica:
### What happened?

Let's consider the `all_to_all` operation described in spec https://github.com/openxla/stablehlo/blob/main/docs/spec.md#all_to_all 
and use the same example in the diagram but with a different `replica_group=[[1,0]]`

In other words, we have two tensors:
```
    %inputs0 = stablehlo.constant dense<[[1, 2, 3, 4],
                                         [5, 6, 7, 8]]> : tensor<2x4xi64>
    %inputs1 = stablehlo.constant dense<[[9, 10, 11, 12],
                                         [13, 14, 15, 16]]> : tensor<2x4xi64>
 
```
and want to perform an `all_to_all` operation with following attributes:
```
      split_dimension = 1 : i64,
      concat_dimension = 0 : i64,
      split_count = 2 : i64,
      replica_groups = dense<[[1, 0]]> : tensor<1x2xi64>

```

Similarly, we can derive `process_group=[[(1,0), (0,0)]]`, and let's do steps mentioned in the spec:

```
split_parts...@sender = split(operands...@sender, split_count, split_dimension) for all sender in process_group.
scattered_parts...@receiver = [split_parts...@sender[receiver_index] for sender in process_group] where receiver_index = process_group.index(receiver).
results...@process = concatenate(scattered_parts...@process, concat_dimension).
```

```
#process_group=[[(1,0), (0,0)]]
#split_parts:

split_parts@input0=[ [[1,2],[5,6]], [[3,4], [7,8]] ]
split_parts@input1=[ [[9,10], [13,14]], [[11,12], [15,16]] ]

#scattered_parts:
scattered_parts@0 = [split_parts@sender[1] for sender in process_group]
 |
 V
scattered_parts@0=[ [[11, 12], [15, 16]], [[3, 4], [7, 8]] ]

scattered_parts@1 = [split_parts@sender[0] for sender in process_group]
 |
 V
scattered_parts@1= [ [[9, 10], [13, 14]], [[1,2], [5,6]] ] 

#results
result@0=  [[11, 12], [15, 16], [3, 4], [7, 8]]
result@1= [[9,10], [13, 14], [1, 2], [5, 6]]
```

However, when it comes to our interpreter, it becomes:
<img width="171" alt="image" src="https://github.com/user-attachments/assets/0774dc15-8d2f-424c-a89e-b74d5d7e52c3">

[...]

PR #2457, opened 2024-07-23
Update all_to_all op interpreter to fix issue#2433

abhigunj:
fixes: https://github.com/openxla/stablehlo/issues/2433

```
scattered_parts...@receiver = [split_parts...@sender[receiver_index] for sender in process_group] where receiver_index = process_group.index(receiver).
```
each `sender/process` from `process_group` contributes a `split_part` to the `scattered_parts`. This PR ensures the order of `split_part` in `scattered_parts` matches with the order of the `sender/processes` within the `process_group`.

### DIFF
--- a/stablehlo/reference/Ops.cpp
+++ b/stablehlo/reference/Ops.cpp
@@ -1202,23 +1202,17 @@ Tensor allToAllOp(const Tensor &operand, Axis splitDimension,
     llvm::report_fatal_error(invalidArgument(
         "Failed to find process group with process_id: (%d, %d)",
         process->getId().replicaId, process->getId().partitionId));
-  auto rendezvousResult =
-      process->rendezvous(*processGroup, channelId, {operand});
+  auto groupOperands =
+      process->rendezvous(processGroup.value(), channelId, {operand});
 
-  auto processIndex = 0;
-  for (auto [i, processId] : llvm::enumerate(*processGroup)) {
-    if (processId == process->getId()) {
-      processIndex = i;
-      break;
-    }
-  }
+  auto receiverIndex = llvm::find(processGroup.value(), process->getId()) -
+                       processGroup->begin();
 
   SmallVector<Tensor> scatteredParts;
-  for (const auto processId : processGroup.value()) {
-    auto splitParts =
-        split(rendezvousResult.lookup(processId).front(), splitCount,
-              splitDimension, operand.getType().getContext());
-    scatteredParts.push_back(splitParts[processIndex]);
+  for (const auto &sender : processGroup.value()) {
+    auto splitParts = split(groupOperands.lookup(sender).front(), splitCount,
+                            splitDimension, operand.getType().getContext());
+    scatteredParts.push_back(splitParts[receiverIndex]);
   }
   return concatenateOp(scatteredParts, concatDimension, resultType);
 }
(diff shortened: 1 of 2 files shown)


stablehlo/reference/Ops.cpp
```diff
@@ -1198,20 +1198,27 @@ Tensor allToAllOp(const Tensor &operand, Axis splitDimension,
...
 
   auto processGroup = processGroups.findGroup(process->getId());
-  if (!processGroup)
+  if (!processGroup.has_value())
     llvm::report_fatal_error(invalidArgument(
         "Failed to find process group with process_id: (%d, %d)",
         process->getId().replicaId, process->getId().partitionId));
-  auto groupOperands = process->rendezvous(*processGroup, channelId, {operand})
-                           .getSortedTensors();
+  auto rendezvousResult =
```
ghpvnist, line 1205: Implementation looks good. Suggestion to simplify code and make the variable names better match the spec:
```
auto operands = process->rendezvous(*processGroup, channelId, {operand});
auto receiver_index =
    llvm::find(*processGroup, process->getId()) - processGroup->begin();

SmallVector<Tensor> scatteredParts;
for (const auto& sender : *processGroup) {
  auto splitParts = split(operands.lookup(sender).front(), splitCount,
                          splitDimension, operand.getType().getContext());
  scatteredParts.push_back(splitParts[receiver_index]);
}
```

stablehlo/tests/interpret/all_to_all.mlir
```diff
@@ -32,6 +32,38 @@ module @cross_replica {
 
 // -----
 
+module @cross_replica_reverse_replica_groups {
```
ghpvnist, line 35: Let's rename it to: `@cross_replica_issue_2433`
precedent in https://github.com/openxla/stablehlo/blob/eba821aa1c54a21d70331d7926dfc8b929f988f3/stablehlo/tests/interpret/all_gather.mlir#L52.

ghpvnist (approved):
LGTM with minor touch ups. Thanks @Hatsunespica for the bug and @abhigunj for the fix!

stablehlo/reference/Ops.cpp
abhigunj: agree. Thanks for the review. updated the PR with some modifications. 

renamed `operands` -> `groupOperands` to make it future proof for variadic operands.
 `*processGroup`  ->`processGroup.value()`

stablehlo/reference/Ops.cpp
```diff
@@ -1198,20 +1198,21 @@ Tensor allToAllOp(const Tensor &operand, Axis splitDimension,
...
   auto processGroup = processGroups.findGroup(process->getId());
-  if (!processGroup)
+  if (!processGroup.has_value())
     llvm::report_fatal_error(invalidArgument(
         "Failed to find process group with process_id: (%d, %d)",
         process->getId().replicaId, process->getId().partitionId));
-  auto groupOperands = process->rendezvous(*processGroup, channelId, {operand})
-                           .getSortedTensors();
+  auto groupOperands =
+      process->rendezvous(processGroup.value(), channelId, {operand});
```
sdasgup3, line 1206: nit:Can use `*processGroup`? here and below?

sdasgup3 (approved):
lgtm with a nit.

GleasonK:
LGTM, no need for approval from me on this one, other reviewers are enough

stablehlo/reference/Ops.cpp
abhigunj: any advantage of using *? `.value()` also returns reference to the object at std::optional and throws exception for null values, I think it is preferred  vs using * here?  also, it has better readability imo.

abhigunj: sg, thanks for the review.

PR #2459, opened 2024-07-24
Support qdq decomposition of TanOp

sdasgup3:
[parent PR](https://github.com/openxla/stablehlo/pull/2458)

The PR adds patterns for qdq decomposition patterns for newly added `TanOp`  

Note to the reviewers: To may just focus on the very last commit of the chain. The rest is coming from parent PR.

[childPR](https://github.com/openxla/stablehlo/pull/2460)

PR #2460, opened 2024-07-24
Remove type-inference dependency while creating qdq pattarns

sdasgup3:
[ParentPR](https://github.com/openxla/stablehlo/pull/2459)

Previously while creating the QDQ pattern we use `create` API without using the result type and hence reply on the type inference to derive the result type. That works for most of the element-wise operations, however, for dot_general and convolution the result type can might be infeasible to infer in the presense of input quantize types. 

The PR fixes that.

Note to the reviewers: To may just focus on the very last commit of the chain. The rest is coming from parent PR.

[ChildPR](https://github.com/openxla/stablehlo/pull/2461)


stablehlo/transforms/StablehloLegalizeQuantToMath.cpp
```diff
@@ -1347,7 +1363,7 @@ void populateStablehloLegalizeQuantizedOpToQDQPatterns(
   // stablehlo spoecification document, interprets the quantized
```
GleasonK, line 1363: "spoecification" typo

stablehlo/transforms/StablehloLegalizeQuantToMath.cpp
```diff
@@ -153,7 +151,21 @@ Type getQuantStorageType(Type type) {
...
   auto quantizedType = getQuantType(type);
   if (succeeded(quantizedType)) {
-    return getQuantStorageType(*quantizedType);
+    return quantizedType->getStorageType();
+  }
+  return type;
+}
+
+// Extracts expressed type of a UQ type. Return original type if it is no UQ
+// type.
```
GleasonK, line 160:
```
// ... of a UQ type, preserving its shape.
```

stablehlo/transforms/StablehloLegalizeQuantToMath.cpp
```diff
@@ -153,7 +151,21 @@ Type getQuantStorageType(Type type) {
...
+Type getQuantExpressedType(Type type) {
+  if (auto shaped = dyn_cast<ShapedType>(type)) {
```
GleasonK, line 162: Any test case to exercise the newly supported behavior?

GleasonK (approved):

stablehlo/transforms/StablehloLegalizeQuantToMath.cpp
GleasonK: tests in child PR

stablehlo/transforms/StablehloLegalizeQuantToMath.cpp
sdasgup3: done.

stablehlo/transforms/StablehloLegalizeQuantToMath.cpp
sdasgup3: done

stablehlo/transforms/StablehloLegalizeQuantToMath.cpp
sdasgup3: Actually this new behavior is exercised  for the existing operations https://github.com/openxla/stablehlo/blob/b056d3da25c18169c14018a1e953112852a53453/stablehlo/transforms/StablehloLegalizeQuantizedOpToQDQ.cpp#L127 in which cases  `get-expressed type` is explicitly computing the return type for them (instead of auto-deriving those types: old-behavior).

PR #2467, opened 2024-08-01
Implement Dot Algorithms RFC

GleasonK:
Implementation of https://github.com/openxla/stablehlo/pull/2422.

Done:
- Attribute impl
- Spec updates
- Test cases

To-do:
- Create an XLA doc on supported algorithm pairs to link to from spec.
- Compatibility impl, will do this in a child CL to keep review simplified. 

To consider:
- Should we verify algorithms more strictly, so that only algs supported on _some_ hardware are allowed in StableHLO. Expanding support will be trivial, no RFC required, but StableHLO should try to statically enforce as much as possible, why allow nonsense pairs.
- Should dot_general on integers support floating point precision types?

Other cleanup:
- PrecisionConfig is no longer implicitly optional, but all uses are wrapped in `OptionalAttr` now.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -2613,14 +2613,14 @@ TODO: Create an XLA doc which details supported combinations of algorithms.
 | (I4)  | `rhs_batching_dimensions`      | 1-dimensional tensor constant of type `si64`                 | (C1), (C4), (C7), (C9)                         |
 | (I5)  | `lhs_contracting_dimensions`   | 1-dimensional tensor constant of type `si64`                 | (C2), (C3), (C6), (C10)                        |
 | (I6)  | `rhs_contracting_dimensions`   | 1-dimensional tensor constant of type `si64`                 | (C2), (C4), (C8), (C10), (C16)                 |
-| (I7)  | `precision_config`             | variadic number of enums of `DEFAULT`, `HIGH`, and `HIGHEST` | (C11)                                          |
-| (I8)  | `lhs_precision_type`           | FloatType or TensorFloat32                                   |                                                |
-| (I9)  | `rhs_precision_type`           | FloatType or TensorFloat32                                   |                                                |
-| (I10) | `accumulation_type`            | FloatType or TensorFloat32                                   |                                                |
-| (I11) | `lhs_component_count`          | constant of type `si32`                                      | (C21)                                          |
-| (I12) | `rhs_component_count`          | constant of type `si32`                                      | (C22)                                          |
-| (I13) | `num_primitive_operations`     | constant of type `si32`                                      | (C23)                                          |
-| (I14) | `allow_imprecise_accumulation` | constant of type `bool`                                      |                                                |
+| (I7)  | `precision_config`             | variadic number of enums of `DEFAULT`, `HIGH`, and `HIGHEST` | (C11), (C21)                                   |
+| (I8)  | `lhs_precision_type`           | FloatType or TensorFloat32                                   | (C21)                                          |
+| (I9)  | `rhs_precision_type`           | FloatType or TensorFloat32                                   | (C21)                                          |
+| (I10) | `accumulation_type`            | FloatType or TensorFloat32                                   | (C21)                                          |
+| (I11) | `lhs_component_count`          | constant of type `si32`                                      | (C21), (C22)                                   |
+| (I12) | `rhs_component_count`          | constant of type `si32`                                      | (C21), (C23)                                   |
+| (I13) | `num_primitive_operations`     | constant of type `si32`                                      | (C21), (C24)                                   |
+| (I14) | `allow_imprecise_accumulation` | constant of type `bool`                                      | (C21)                                          |
 
 
 #### Outputs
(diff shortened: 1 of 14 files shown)


stablehlo/tests/ops_stablehlo.mlir
```diff
@@ -3055,6 +3055,47 @@ func.func @dot_general(%arg0: tensor<1x?x1x?xf32>, %arg1: tensor<?x1x?x1x?xf32>)
 
 // -----
 
+// CHECK-LABEL: func @dot_general_algorithm
+func.func @dot_general_algorithm_tf32(%arg0: tensor<2x2x2xi64>, %arg1: tensor<2x2x2xi64>) -> tensor<2x2x2xi64> {
```
ghpvnist, line 3059: Let's move the `_tf32` and `_i#` tests below constraint tests for alphabetical ordering

stablehlo/tests/interpret/dot_general.mlir
```diff
@@ -17,6 +17,24 @@ func.func @dot_general_op_test_si64() {
...
+func.func @dot_general_op_test_algorithm() {
+  %lhs = stablehlo.constant dense<[[[1, 2], [3, 4]],
+                                   [[5, 6], [7, 8]]]> : tensor<2x2x2xi64>
+  %rhs = stablehlo.constant dense<[[[1, 0], [0, 1]],
+                                   [[1, 0], [0, 1]]]> : tensor<2x2x2xi64>
+  %result = stablehlo.dot_general %lhs, %rhs,
+    batching_dims = [0] x [0],
+    contracting_dims = [2] x [1],
+    precision = [DEFAULT, DEFAULT],
+    algorithm = <lhs_precision_type = tf32, rhs_precision_type = tf32, accumulation_type = tf32, lhs_component_count = 1, rhs_component_count = 1, num_primitive_operations = 1, allow_imprecise_accumulation = false>
```
ghpvnist, line 29: Let's format this for readability.

ghpvnist:
For remaining todos or cleanups, let's file issues to track them.

docs/spec.md
```diff
@@ -2593,6 +2661,14 @@ planning to address this in
   * If `!is_quantized(lhs)`:
     * (C20) `element_type(lhs) = expressed_type(rhs) = element_type(result)`.
 
+* (C21) If `precision_config... != DEFAULT`,
```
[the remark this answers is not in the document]
GleasonK: I'm struggling on deciding how to spec these. Because really we want their values to be empty, but we don't have a notion of optional in our spec... I.e. `DEFAULT` doesn't actually mean that `lhs_precision_type = element_type(lhs)` on all hardware.. it's meaning is impl defined though so its hard to give a true default value.

I don't think there's a use case for explicitly specifying the precision types to be exactly these, but still risks some ambiguity?

stablehlo/tests/ops_stablehlo.mlir
GleasonK: Actually... the `_i#` and `_c#` tests shouldn't have `algorithm` in them at all.. Removed that now the positive algorithm test is alphabetically first.

Should I put the input tests after the constraint tests to be alphabetical? `dot_general_i8` vs `dot_general_c1`?

stablehlo/conversions/linalg/tests/dot-product.mlir
```diff
@@ -369,3 +369,14 @@ func.func @not_simple_dot_general(
...
+
+// -----
+
+func.func @dot_general_algorithm(%arg0: tensor<2x2x2xi64>, %arg1: tensor<2x2x2xi64>) -> tensor<2x2x2xi64> {
+  // CHECK: stablehlo.dot_general
+  %0 = "stablehlo.dot_general"(%arg0, %arg1) <{
+    dot_dimension_numbers = #stablehlo.dot<lhs_batching_dimensions = [0], rhs_batching_dimensions = [0], lhs_contracting_dimensions = [2], rhs_contracting_dimensions = [1]>,
+    precision_config = [#stablehlo<precision DEFAULT>, #stablehlo<precision DEFAULT>],
+    algorithm = #stablehlo.dot_algorithm<lhs_precision_type = tf32, rhs_precision_type = tf32, accumulation_type = f32, lhs_component_count = 1, rhs_component_count = 1, num_primitive_operations = 1, allow_imprecise_accumulation = false>
+  }> : (tensor<2x2x2xi64>, tensor<2x2x2xi64>) -> tensor<2x2x2xi64>  return %0 : tensor<2x2x2xi64>
```
GleasonK, line 381: what happened to formatting here...

stablehlo/tests/ops_stablehlo.mlir
ghpvnist: Yeah, I also ordered them alphabetically for other test, though I prefer inputs before constraints.

docs/spec.md
```diff
@@ -2542,17 +2542,85 @@ planning to address this in
...
+ allow_imprecise_accumulation = true}
+```
+
+In general, it is not guaranteed that the each algorithm is supported on each
+accelerator type by the consumer of the StableHLO. If a given algorithm is not
+supported, an error should be raised as opposed to falling back to an
+alternative. StableHLO verification will prevent some combinations that are not
+known to be supported on any hardware.
+
+TODO: Create an XLA doc which details supported combinations of algorithms.
```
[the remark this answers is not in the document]
GleasonK: done

docs/spec.md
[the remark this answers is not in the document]
GleasonK: discussed offline, will use `None`.

docs/spec.md
```diff
@@ -2542,17 +2542,85 @@ planning to address this in
...
+to accumulate in lower precision for some steps (as per
+CUBLASLT_MATMUL_DESC_FAST_ACCUM).
+
+It is up to the implementations to decide which combinations are supported.
+
+Example `DotAlgorithm` attributes:
+
+```txt
+// Inputs are casted to tf32, and then accumulated in f32:
+{lhs_precision_type = tf32,
```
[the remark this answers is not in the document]
GleasonK: Added:

```
TensorFloat32 ::= 'tf32'

...

  * `tf32` type corresponds to the [TensorFloat32 format](https://blogs.nvidia.com/blog/tensorfloat-32-precision-format/)
    and has limited support in StableHLO.
```

stablehlo/tests/ops_stablehlo.mlir
GleasonK: ordered alphabetically as well, thanks!

ghpvnist (approved):

Issue #2440, opened 2024-07-16
Bug in interpreter implementation of sort op


GleasonK:
### What happened?

Looks like the StableHLO reference impl of sort behaves differently than XLA and numpy. Found when generating new testdata files, currently we lack broader `sort` and `chlo.top_k` coverage which will be added. Filtering tests on this ticket in the meantime.

```
$ cat /tmp/t.mlir 
module @jit_main attributes {mhlo.num_partitions = 1 : i32, mhlo.num_replicas = 1 : i32} {
  func.func public @main() -> (tensor<5x7xi64> {jax.result_info = "[0]", mhlo.layout_mode = "default"}) {
    %c = stablehlo.constant dense<[[0, 1, 1, -3, -3, -2, 0], [4, -1, 3, 0, 4, 0, -3], [1, 0, 0, 0, 0, 2, 3], [-3, -4, -4, -2, 0, 3, 3], [0, 5, 2, -2, 0, -2, 0]]> : tensor<5x7xi64>
    %0 = "stablehlo.sort"(%c) <{dimension = 0 : i64}> ({
    ^bb0(%arg0: tensor<i64>, %arg1: tensor<i64>):
      %1 = stablehlo.compare  LT, %arg0, %arg1,  SIGNED : (tensor<i64>, tensor<i64>) -> tensor<i1>
      stablehlo.return %1 : tensor<i1>
    }) : (tensor<5x7xi64>) -> tensor<5x7xi64>
    return %0 : tensor<5x7xi64>
  }
}

$ stablehlo-translate -interpret /tmp/t.mlir
tensor<5x7xi64> {
  [
    [4, -4, 0, -3, -3, -2, -3],
    [0, -1, 2, -2, 0, 2, 0],
    [-3, 0, 3, -2, 4, 3, 3],
    [0, 1, 1, 0, 0, -2, 0],
    [1, 5, -4, 0, 0, 0, 3]
  ]
}

$ run_hlo_module -- /tmp/t.mlir --platform=cpu --input_format=stablehlo --print_literals
// ...
** Result with test runner Host **
s64[5,7] {
  { -3, -4, -4, -3, -3, -2, -3 },
  { 0, -1, 0, -2, 0, -2, 0 },
  { 0, 0, 1, -2, 0, 0, 0 },
  { 1, 1, 2, 0, 0, 2, 3 },
  { 4, 5, 3, 0, 4, 3, 3 }
}
Running HLO module with runner Interpreter...
[...]

PR #2474, opened 2024-08-02
Fix SortOp interpreter

sdasgup3:
fixes https://github.com/openxla/stablehlo/issues/2440. The source and destination indices were not in correct order.

### DIFF
--- a/stablehlo/reference/Ops.cpp
+++ b/stablehlo/reference/Ops.cpp
@@ -2435,8 +2435,7 @@ SmallVector<Tensor> sortOp(ArrayRef<Tensor> inputs, Axis dimension,
 
     // After the tensor of handles has been sorted, we apply the results of
     // this sort by reshuffling input elements into result elements.
-    auto &resultsTogether = inputsTogether;
-    for (auto [inputHandle, resultHandle] : llvm::enumerate(resultsTogether)) {
+    for (auto [resultHandle, inputHandle] : llvm::enumerate(inputsTogether)) {
       for (auto [input, result] : llvm::zip(inputs, results)) {
         auto inputIndex = *resultIt;
         auto resultIndex = *resultIt;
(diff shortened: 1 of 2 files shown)


stablehlo/tests/interpret/sort.mlir
```diff
@@ -15,3 +15,16 @@ func.func @sort_stable() {
   check.expect_eq_const %result1, dense<[[1, 2, 1], [3, 2, 3]]> : tensor<2x3xi64>
   func.return
 }
+
+// -----
+
+func.func public @sort() {
```
abhigunj, line 21: can we rename to sort_issue_x as precedent? 
https://github.com/openxla/stablehlo/blob/eba821aa1c54a21d70331d7926dfc8b929f988f3/stablehlo/tests/interpret/all_gather.mlir#L52

stablehlo/reference/Ops.cpp
```diff
@@ -2435,8 +2435,7 @@ SmallVector<Tensor> sortOp(ArrayRef<Tensor> inputs, Axis dimension,
 
     // After the tensor of handles has been sorted, we apply the results of
     // this sort by reshuffling input elements into result elements.
-    auto &resultsTogether = inputsTogether;
-    for (auto [inputHandle, resultHandle] : llvm::enumerate(resultsTogether)) {
+    for (auto [resultHandle, inputHandle] : llvm::enumerate(inputsTogether)) {
```
abhigunj, line 2438: nit: `const auto&`

abhigunj (approved):

stablehlo/reference/Ops.cpp
sdasgup3: Actually `inputsTogether` container of ints. I usually avoid using a reference for iterating over a container of numbers/pointers (which is also generally followed in the mlir codebase) as it wins little to nothing.

stablehlo/tests/interpret/sort.mlir
sdasgup3: done.

GleasonK (approved):

PR #2478, opened 2024-08-07
New StablehloLegalizeQDQToQuantizedOpPass

abhigunj:
`StablehloLegalizeQDQToQuantizedOpPass` composes StableHLO quantized op using uniform quantize/dequantize ops.

    ```mlir
    func.func @add(%arg0: tensor<16x16x!quant.uniform<ui8:f32, 34.0:16>>) -> tensor<16x16x!quant.uniform<ui8:f32, 34.0:16>> {
      %0 = stablehlo.uniform_dequantize %arg0 : (tensor<16x16x!quant.uniform<ui8:f32, 34.0:16>>) -> tensor<16x16xf32>
      %1 = stablehlo.abs %0 : tensor<16x16xf32>
      %2 = stablehlo.uniform_quantize %1 : (tensor<16x16xf32>) -> tensor<16x16x!quant.uniform<ui8:f32, 34.0:16>>
      func.return %2 : tensor<16x16x!quant.uniform<ui8:f32, 34.0:16>>
    }
    ```

    Will become:

    ```mlir
    func.func @add(%arg0: tensor<16x16x!quant.uniform<u8:f32, 3.400000e+01:16>>) -> tensor<16x16x!quant.uniform<u8:f32, 3.400000e+01:16>> {
      %0 = stablehlo.abs %arg0 : tensor<16x16x!quant.uniform<u8:f32, 3.400000e+01:16>>
      return %0 : tensor<16x16x!quant.uniform<u8:f32, 3.400000e+01:16>>
    }
    ```

created https://github.com/openxla/stablehlo/issues/2485 to track extending this pass to support more patterns


stablehlo/transforms/StablehloLegalizeAToB.cpp
```diff
@@ -0,0 +1,138 @@
...
+      quantizedResults.push_back(newResult);
+    }
+    rewriter.setInsertionPointAfter(quantizingOp);
+    OperationState newState(
+        quantizingOp->getLoc(), quantizingOp->getName().getStringRef(),
+        quantizedOperands, quantizedResults, quantizingOp->getAttrs());
+    for (unsigned int i = 0; i < quantizingOp->getNumRegions(); ++i) {
+      newState.addRegion();
+    }
+    Operation* quantizedOp = rewriter.create(newState);
```
GleasonK, line 88: Can we modify the existing op in place somehow? Not sure if that will work, since I'm not sure if we can change result type.

GleasonK: Abhinav: Not sure if we can manipulate the result type.

GleasonK (changes requested):

stablehlo/transforms/StablehloLegalizeAToB.cpp
```diff
@@ -0,0 +1,138 @@
...
+    // if quantizingOp is alredy quanitzed, no work to do
+    if (isAnyQuantizedTypes(quantizingOp->getOperandTypes())) return failure();
+
+    llvm::SmallVector<Value> quantizedOperands;
+    for (const auto& operandx : quantizingOp->getOperands()) {
+      // backtrack and get the quantized operand
+      // quantizedOp -> DequantizedOp -> shloOp
+      auto* opx = operandx.getDefiningOp();
+      // It is possible an operand is input of the module.
+      if (!(opx)) {
```
sdasgup3, line 63: remove extra paren

stablehlo/transforms/StablehloLegalizeAToB.cpp
```diff
@@ -0,0 +1,138 @@
...
+        quantizedOperands.push_back(operandx);
+        continue;
+      }
+
+      if (auto deqOp = dyn_cast<stablehlo::UniformDequantizeOp>(opx)) {
```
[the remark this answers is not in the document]
sdasgup3: >> // All operands and results must be quantized
This makes sense as the first version of the pass. Its worthwhile noting here that some operands can be non-quantized for hybrid quantized ops. In that way, whichever operands are coming from dq op, we can keep them in quantized type, else we will keep the operand as-is. 

I am ok with incorporating the change in a future version.

stablehlo/transforms/StablehloLegalizeAToB.cpp
```diff
@@ -0,0 +1,138 @@
...
+    rewriter.setInsertionPointAfter(quantizingOp);
+    OperationState newState(
+        quantizingOp->getLoc(), quantizingOp->getName().getStringRef(),
+        quantizedOperands, quantizedResults, quantizingOp->getAttrs());
+    for (unsigned int i = 0; i < quantizingOp->getNumRegions(); ++i) {
+      newState.addRegion();
+    }
+    Operation* quantizedOp = rewriter.create(newState);
+
+    if (quantizingOp->getNumRegions() != 0) {
```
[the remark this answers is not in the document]
sdasgup3: Totally agree given that we have ops like reduce* for which it is not straightforward to emit the quantized block (e.g. the argument storage type might be different than the operand storage type).

stablehlo/transforms/StablehloLegalizeAToB.cpp
sdasgup3: >>  modify the existing op in place

Just to add removing the original fp op might not be desirable when 
 
```
dq = stablehlo.uniform.dequantize
op = stablehlo.abs(dq)
q = stablehlo.uniform.quantize(op)
non-q = stablehlo.add(op) // another use of op in non-quantized context
```
where we might need to preserve the original fp op. The other option of of entirely removing the fp op and instead  dequantizing the quantized op to be fed to `non-q` will not have the same semantics  mainly because of "loss-of-precision" reasons.

stablehlo/tests/transforms/stablehlo_legalize_qdq_to_quantized_op.mlir
```diff
@@ -0,0 +1,32 @@
...
+
+// CHECK-LABEL @compose_quantized_abs_op
+// CHECK: %[[abs0:.*]] = stablehlo.abs %arg0 : tensor<16x16x!quant.uniform<u8:f32, 3.400000e+01:16>>
+// CHECK-NEXT: %[[abs1:.*]] = stablehlo.abs %[[abs0]] : tensor<16x16x!quant.uniform<u8:f32, 3.400000e+01:16>>
+// CHECK-NEXT: return %[[abs1]] : tensor<16x16x!quant.uniform<u8:f32, 3.400000e+01:16>>
+func.func @compose_quantized_abs_op(%arg0: tensor<16x16x!quant.uniform<ui8:f32, 34.0:16>>) -> tensor<16x16x!quant.uniform<ui8:f32, 34.0:16>> {
+    %0 = stablehlo.uniform_dequantize %arg0 : (tensor<16x16x!quant.uniform<ui8:f32, 34.0:16>>) -> tensor<16x16xf32>
+    %1 = stablehlo.abs %0 : tensor<16x16xf32>
+    %2 = stablehlo.uniform_quantize %1 : (tensor<16x16xf32>) -> tensor<16x16x!quant.uniform<ui8:f32, 34.0:16>>
+    %3 = stablehlo.abs %2 : tensor<16x16x!quant.uniform<ui8:f32, 34.0:16>>
```
sdasgup3, line 13: You can keep the first test to showcase the functioning of the pass., just with `%0, %1, %2, return %2`. No need for the trailing quantized abs. In separate tests you can showcase a case the pass cannot successfully pattern match: These will be mostly the cases we return `notifyMatchFailure` (e.g. hybrid case, ops with regions etc.).

stablehlo/transforms/Passes.h
```diff
@@ -72,6 +72,9 @@ void populateStablehloLegalizeQuantizedOpToQDQPatterns(
     RewritePatternSet *patterns, MLIRContext *context,
     PatternBenefit benefit = 1);
 
+void populateStablehloLegalizeQDQToQuantizedOpPatterns(
```
sdasgup3, line 75: header comment

stablehlo/transforms/Passes.td
```diff
@@ -258,3 +258,34 @@ def StablehloLegalizeQuantizedOpToQDQPass : Pass<"stablehlo-legalize-quantized-o
...
+      %2 = stablehlo.uniform_quantize %1 : (tensor<16x16xf32>) -> tensor<16x16x!quant.uniform<ui8:f32, 34.0:16>>
+      func.return %2 : tensor<16x16x!quant.uniform<ui8:f32, 34.0:16>>
+    }
+    ```
+
+    Will become:
+
+    ```mlir
+    func.func @add(%arg0: tensor<16x16x!quant.uniform<u8:f32, 3.400000e+01:16>>) -> tensor<16x16x!quant.uniform<u8:f32, 3.400000e+01:16>> {
+      %0 = stablehlo.abs %arg0 : tensor<16x16x!quant.uniform<u8:f32, 3.400000e+01:16>>
```
sdasgup3, line 283: Is there a reason we are not showing the original non-deleted ops?

stablehlo/transforms/StablehloLegalizeQDQToQuantizedOp.cpp
```diff
@@ -0,0 +1,135 @@
...
+    // Matching sequence of ops:
+    //    UniformDequantizeOp -> non-quantized Op -> UniformQuantizeOp.
+    // Start matching from a UniformQuantizeOp (`quantOp`).
+    // Get the Op (`computeOp`) which defines the inputs of the `quantOp`.
+    // Verify all inputs of the `computeOp` are produced by UniformDequantizeOp.
+    // Note: The pass does not delete any prexisting op.
+    auto* computeOp = quantOp->getOperand(0).getDefiningOp();
+    if (!computeOp)
+      return rewriter.notifyMatchFailure(
+          quantOp, "requires operand to be defined by a op");
```
sdasgup3, line 54: `an operation.`

stablehlo/transforms/StablehloLegalizeQDQToQuantizedOp.cpp
```diff
@@ -0,0 +1,135 @@
...
+    for (const Value& operand : computeOp->getOperands()) {
+      auto* definingOp = operand.getDefiningOp();
+      if (!definingOp)
+        return rewriter.notifyMatchFailure(
+            computeOp, "requires operand to be defined by a op");
+
+      auto dequantOp = dyn_cast<stablehlo::UniformDequantizeOp>(definingOp);
+      if (!dequantOp)
+        return rewriter.notifyMatchFailure(
+            definingOp, "requires op to be a UniformDequantizeOp");
```
sdasgup3, line 80: `required operand to be defined by an stablehlo.uniform_dequantize op`

stablehlo/tests/transforms/stablehlo_legalize_qdq_to_quantized_op.mlir
sdasgup3: we can use the current testfile and write a test which the pattern would failed to match and then check the resulting program (which will be mostly unconverted) using check directive.

stablehlo/transforms/Passes.td
abhigunj: to keep the example short and focus on what pass does. Description notes that pass doesn't delete the preexisting ops. discussed [here ](https://github.com/openxla/stablehlo/pull/2478#discussion_r1707769311)

stablehlo/transforms/Passes.td
```diff
@@ -258,3 +258,42 @@ def StablehloLegalizeQuantizedOpToQDQPass : Pass<"stablehlo-legalize-quantized-o
...
 }
+
+def StablehloLegalizeAToBPass : Pass<"stablehlo-legalize-a-to-b", "mlir::func::FuncOp"> {
+  let summary = "Recompose StableHLO quantized ops using uniform dequantize + dequantized op";
+
+  let description = [{
+    Decompose StableHLO quantized programs using uniform quantize/dequantize
+    operations. For example, the following program
+
+    ```mlir
```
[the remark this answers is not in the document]
sdasgup3: @GleasonK 
How about we keep the unmodified ops together to explemplify that the meaningful (and potentially desired) output  is not the end product here: would need additional passes (e.g. dce) on top.

stablehlo/transforms/Passes.td
abhigunj: closing this as we are discussing on at another comment thread.

stablehlo/transforms/Passes.h
```diff
@@ -72,6 +72,10 @@ void populateStablehloLegalizeQuantizedOpToQDQPatterns(
     RewritePatternSet *patterns, MLIRContext *context,
     PatternBenefit benefit = 1);
 
+/// Collection of rewrite patterns for composing quantized StableHLO operations
```
sdasgup3, line 75: `.` at the end.

stablehlo/transforms/Passes.td
```diff
@@ -258,3 +258,34 @@ def StablehloLegalizeQuantizedOpToQDQPass : Pass<"stablehlo-legalize-quantized-o
     "mlir::stablehlo::StablehloDialect",
   ];
 }
+
+def StablehloLegalizeQDQToQuantizedOpPass : Pass<"stablehlo-legalize-qdq-to-quantized-op", "mlir::func::FuncOp"> {
+  let summary = "Compose StableHLO quantized op using uniform quantize/dequantize ops";
```
sdasgup3, line 263: This summary does not seem to imply the reverse of "Decompose StableHLO quantized ops using uniform quantize/dequantize ops." (in the pass above)

How about:
"Compose uniform quantize/dequantize ops into StableHLO quantized ops"

stablehlo/transforms/StablehloLegalizeQDQToQuantizedOp.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+    if (isAnyQuantizedTypes(computeOp->getOperandTypes()))
+      return rewriter.notifyMatchFailure(computeOp,
+                                         "requires non quantized operands");
+
+    // Collect quantized operands and result types to rewrite
+    // All operands and results must be quantized
+    llvm::SmallVector<Value> quantizedComputeOpOperands;
+    for (const Value& operand : computeOp->getOperands()) {
+      auto* definingOp = operand.getDefiningOp();
```
sdasgup3, line 72:
```
      auto dequantOp = operand.getDefiningOp<stablehlo::UniformDequantizeOp>();
```

and merge the following two checks.

stablehlo/transforms/StablehloLegalizeQDQToQuantizedOp.cpp
```diff
@@ -0,0 +1,137 @@
...
+struct QuantizedStablehloQDQToQuantizedOpConversion
+    : public OpRewritePattern<stablehlo::UniformQuantizeOp> {
+  using OpRewritePattern<stablehlo::UniformQuantizeOp>::OpRewritePattern;
+  LogicalResult matchAndRewrite(stablehlo::UniformQuantizeOp quantOp,
+                                PatternRewriter& rewriter) const override {
+    // Matching sequence of ops:
+    //    UniformDequantizeOp -> non-quantized Op -> UniformQuantizeOp.
+    // Start matching from a UniformQuantizeOp (`quantOp`).
+    // Get the Op (`computeOp`) which defines the inputs of the `quantOp`.
+    // Verify all inputs of the `computeOp` are produced by UniformDequantizeOp.
```
sdasgup3, line 49: ...produced by UniformDequantizeOp (`dequantOp`).

stablehlo/transforms/StablehloLegalizeQDQToQuantizedOp.cpp
```diff
@@ -0,0 +1,137 @@
...
+      auto dequantOp = dyn_cast<stablehlo::UniformDequantizeOp>(definingOp);
+      if (!dequantOp)
+        return rewriter.notifyMatchFailure(
+            definingOp,
+            "requires operand to be defined by an stablehlo.uniform_dequantize "
+            "op");
+
+      quantizedComputeOpOperands.push_back(dequantOp->getOperand(0));
+    }
+    SmallVector<Type> quantizedComputeOpResultTypes(quantOp->getResultTypes());
```
sdasgup3, line 86: nit:Can we just inline this single use variable.?

stablehlo/tests/transforms/stablehlo_legalize_qdq_to_quantized_op.mlir
```diff
@@ -0,0 +1,151 @@
...
+    %2 = stablehlo.uniform_quantize %1 : (tensor<16x16xf32>) -> tensor<16x16x!quant.uniform<ui8:f32, 34.0:16>>
+    func.return %2 : tensor<16x16x!quant.uniform<ui8:f32, 34.0:16>>
+}
+
+// -----
+
+// CHECK-LABEL @failure_uniform_quant_op_operand_not_defined_by_op
+// CHECK-NOT: stablehlo.abs {{.*}} : tensor<16x16x!quant.uniform<ui8:f32, 34.0:16>>
+// CHECK:     return
+func.func @failure_uniform_quant_op_operand_not_defined_by_op(%arg0: tensor<16x16xf32>) -> tensor<16x16x!quant.uniform<ui8:f32, 34.0:16>> {
```
sdasgup3, line 20: The name suggests that operand of `uniform_quantize` not defined by op.. but here its is defined by `abs`?

sdasgup3: May be `failure_to_match_compute_op_operand_not_defined_by_dequantize_op`

stablehlo/tests/transforms/stablehlo_legalize_qdq_to_quantized_op.mlir
```diff
@@ -0,0 +1,151 @@
...
+// CHECK-NEXT:    }) : (tensor<4x5xf32>, tensor<4x5xf32>) -> tensor<4x5xf32>
+// CHECK-NEXT:    %5 = stablehlo.uniform_quantize %4 : (tensor<4x5xf32>) -> tensor<4x5x!quant.uniform<u8:f32, 3.400000e+01:16>>
+// CHECK-NEXT:    return %5 : tensor<4x5x!quant.uniform<u8:f32, 3.400000e+01:16>>
+
+func.func @failure_op_with_region(%arg0: tensor<4x5xf32>, %arg1: tensor<4x5xf32>) -> tensor<4x5x!quant.uniform<ui8:f32, 34.0:16>> {
+  %0 = stablehlo.uniform_quantize %arg0 : (tensor<4x5xf32>) -> tensor<4x5x!quant.uniform<ui8:f32, 34.0:16>>
+  %1 = stablehlo.uniform_dequantize %0 : (tensor<4x5x!quant.uniform<ui8:f32, 34.0:16>>) -> tensor<4x5xf32>
+  %2 = stablehlo.uniform_quantize %arg1 : (tensor<4x5xf32>) -> tensor<4x5x!quant.uniform<ui8:f32, 34.0:16>>
+  %3 = stablehlo.uniform_dequantize %2 : (tensor<4x5x!quant.uniform<ui8:f32, 34.0:16>>) -> tensor<4x5xf32>
+  %4 = "stablehlo.map"(%1, %3) ({
```
sdasgup3, line 46: map is about to deprecate. Use reduce instead.

stablehlo/tests/transforms/stablehlo_legalize_qdq_to_quantized_op.mlir
```diff
@@ -0,0 +1,151 @@
...
+// CHECK-NEXT:    %2 = stablehlo.uniform_quantize %arg1 : (tensor<4x5xf32>) -> tensor<4x5x!quant.uniform<u8:f32, 3.400000e+01:16>>
+// CHECK-NEXT:    %3 = stablehlo.uniform_dequantize %2 : (tensor<4x5x!quant.uniform<u8:f32, 3.400000e+01:16>>) -> tensor<4x5xf32>
+// CHECK-NEXT:    %4 = "stablehlo.map"(%1, %3) <{dimensions = array<i64: 0, 1>}> ({
+// CHECK-NEXT:    ^bb0(%arg2: tensor<f32>, %arg3: tensor<f32>):
+// CHECK-NEXT:      stablehlo.return %cst : tensor<f32>
+// CHECK-NEXT:    }) : (tensor<4x5xf32>, tensor<4x5xf32>) -> tensor<4x5xf32>
+// CHECK-NEXT:    %5 = stablehlo.uniform_quantize %4 : (tensor<4x5xf32>) -> tensor<4x5x!quant.uniform<u8:f32, 3.400000e+01:16>>
+// CHECK-NEXT:    return %5 : tensor<4x5x!quant.uniform<u8:f32, 3.400000e+01:16>>
+
+func.func @failure_op_with_region(%arg0: tensor<4x5xf32>, %arg1: tensor<4x5xf32>) -> tensor<4x5x!quant.uniform<ui8:f32, 34.0:16>> {
```
sdasgup3, line 41: `failure_to_match_op_with_region`

sdasgup3: same `failure_to_match_` refix below..

stablehlo/tests/transforms/stablehlo_legalize_qdq_to_quantized_op.mlir
```diff
@@ -0,0 +1,151 @@
...
+    channel_handle = #stablehlo.channel_handle<handle = 1, type = 0>,
+    replica_groups = dense<[[0, 2, 4, 6], [1, 3, 5, 7]]> : tensor<2x4xi64>
+  } : (tensor<8x2xf32>, tensor<2x2xf32>) -> (tensor<8x8xf32>, tensor<2x4xf32>)
+  %5 = stablehlo.uniform_quantize %4#0 : (tensor<8x8xf32>) -> tensor<8x8x!quant.uniform<ui8:f32, 34.0:16>>
+  func.return %5, %4#1 : tensor<8x8x!quant.uniform<ui8:f32, 34.0:16>>, tensor<2x4xf32>
+}
+
+// -----
+
+// CHECK-LABEL @failure_all_operands_not_quantized
```
sdasgup3, line 81: CHECK-LABEL name inconsistent with the actual function name

stablehlo/tests/transforms/stablehlo_legalize_qdq_to_quantized_op.mlir
```diff
@@ -0,0 +1,151 @@
...
+// CHECK:       %[[oper0:.*]] = stablehlo.uniform_quantize %arg0 : (tensor<16x16xf32>) -> tensor<16x16x!quant.uniform<u8:f32, 3.400000e+01:16>>
+// CHECK-NEXT:  %[[oper1:.*]] = stablehlo.uniform_quantize %arg1 : (tensor<16x16xf32>) -> tensor<16x16x!quant.uniform<u8:f32, 3.400000e+01:16>>
+// CHECK-NEXT:  %[[add_result:.*]] = stablehlo.add %[[oper0]], %[[oper1]] : tensor<16x16x!quant.uniform<u8:f32, 3.400000e+01:16>>
+// CHECK-NEXT:  return %[[add_result]] : tensor<16x16x!quant.uniform<u8:f32, 3.400000e+01:16>>
+func.func @compose_quantized_add_op(%arg0: tensor<16x16xf32>, %arg1: tensor<16x16xf32>) -> tensor<16x16x!quant.uniform<ui8:f32, 34.0:16>> {
```
sdasgup3, line 86: isn't abs op sufficnt to demonstrate the function of the pass? IOW, do we need this test

stablehlo/tests/transforms/stablehlo_legalize_qdq_to_quantized_op.mlir
```diff
@@ -0,0 +1,151 @@
...
+// -----
+
+// CHECK-LABEL @failure_not_all_arguments_are_quantized
+// CHECK:        %0 = stablehlo.uniform_quantize %arg0 : (tensor<1x8x8x207xf32>) -> tensor<1x8x8x207x!quant.uniform<u8:f32, 3.400000e+01:16>>
+// CHECK-NEXT:   %1 = stablehlo.uniform_dequantize %0 : (tensor<1x8x8x207x!quant.uniform<u8:f32, 3.400000e+01:16>>) -> tensor<1x8x8x207xf32>
+// CHECK-NEXT:   %2 = stablehlo.abs %arg1 : tensor<3x3x207x16x!quant.uniform<i8:f32, 5.000000e+00:20>>
+// CHECK-NEXT:   %3 = stablehlo.convolution(%1, %2) {{.*}} : (tensor<1x8x8x207xf32>, tensor<3x3x207x16x!quant.uniform<i8:f32, 5.000000e+00:20>>) -> tensor<1x8x8x16xf32>
+// CHECK-NEXT:   %4 = stablehlo.uniform_quantize %3 : (tensor<1x8x8x16xf32>) -> tensor<1x8x8x16x!quant.uniform<u8:f32, 3.400000e+01:16>>
+// CHECK-NEXT:   return %4 : tensor<1x8x8x16x!quant.uniform<u8:f32, 3.400000e+01:16>>
+func.func @failure_not_all_arguments_are_quantized(%arg0: tensor<1x8x8x207xf32>, %arg1: tensor<3x3x207x16x!quant.uniform<i8:f32, 5.0:20>>) -> tensor<1x8x8x16x!quant.uniform<ui8:f32, 34.0:16>> {
```
sdasgup3, line 105: `failure_to_match_operand_of_compute_op_already_quantized`

stablehlo/tests/transforms/stablehlo_legalize_qdq_to_quantized_op.mlir
```diff
@@ -0,0 +1,151 @@
...
+         {batch_group_count = 1 : i64, feature_group_count = 1 : i64, precision_config = [#stablehlo<precision DEFAULT>, #stablehlo<precision DEFAULT>]} :
+       (tensor<1x8x8x207xf32>, tensor<3x3x207x16x!quant.uniform<i8:f32, 5.0:20>>) -> tensor<1x8x8x16xf32>
+    %4 = stablehlo.uniform_quantize %3 : (tensor<1x8x8x16xf32>) -> tensor<1x8x8x16x!quant.uniform<ui8:f32, 34.0:16>>
+  func.return %4 : tensor<1x8x8x16x!quant.uniform<ui8:f32, 34.0:16>>
+}
+
+
+// -----
+
+// CHECK-LABEL @failure_operand_not_defined_by_op
```
sdasgup3, line 121: With the above proposal `auto dequantOp = operand.getDefiningOp<stablehlo::UniformDequantizeOp>();`, we do not have a separate check for `operand_of_compute_op_not_defined_by_op`, so following  two tetss can be reduced into one.

stablehlo/transforms/Passes.td
abhigunj: it does imply? `compose (put together)  StableHLO quantized op using uniform quantize/dequantize ops` ?

stablehlo/transforms/StablehloLegalizeQDQToQuantizedOp.cpp
abhigunj: two separate checks is better for logging-granularity which will help in debugging..?

stablehlo/transforms/StablehloLegalizeQDQToQuantizedOp.cpp
abhigunj: done

stablehlo/tests/transforms/stablehlo_legalize_qdq_to_quantized_op.mlir
abhigunj: thanks, added prefix `failed_to_match` to test names.

stablehlo/tests/transforms/stablehlo_legalize_qdq_to_quantized_op.mlir
abhigunj: corrected label name, ty.

stablehlo/tests/transforms/stablehlo_legalize_qdq_to_quantized_op.mlir
abhigunj: assuming this is from the older snapshot. The extra test is already removed.

stablehlo/transforms/Passes.td
sdasgup3: This is what I am thinking: 

In the context of `StableHLOLegalizeQuantOpToQDQ`, the line `...using uniform quantize/dequantize` suggests that the destination artifacts. That is we decompose quant-op `using` q/dq op (the destination). 

Ideally, in the context of `StablehloLegalizeQDQToQuantizedOpPass`  the description should suggest that the dqd are in the source.  

Another option I can think of:  
quant-op -> qdq: Decompose quantized StableHLO operation to (de-quantize, floating-point operation and quantize) pattern. 
qdq->quant-op: Compose  (de-quantize, floating-point operation and quantize) pattern to StableHLO quantized operation.

This is just a minor suggestion for better readability (and hence subjective). Feel free to resolve.

stablehlo/transforms/StablehloLegalizeQDQToQuantizedOp.cpp
sdasgup3: Yes the `getDefiningOp<OpTy>` merges the two checks if operand is optype and if defining operation is of `OpTy`. This will simplify the check to "if the operand is coming from a dequant op" as compored to if "operand is defined using an op" and "defined op is deqaunt". Do we need that distinction in the code ?

 Tradeoff cost of keeping the simplified form in the code for someone to understand the failure reason comes down to debugging "why a defined op is not uniform_dequantize", which is not much. 

Again it been a topic of (code simplicity vs debuggability) is subjective, so feel free to make a choice.

stablehlo/transforms/Passes.td
abhigunj: ack, thank you.
I am onboard with 
`Decompose quantized StableHLO operation to (de-quantize, floating-point operation and quantize) pattern.`
and 
`Fuse (de-quantize, floating-point operation and quantize) pattern into StableHLO quantized operation.`

will update it for both the passes.

stablehlo/transforms/StablehloLegalizeQDQToQuantizedOp.cpp
abhigunj: I unsure either, don't see a major value of keeping separate checks but Pass itself is simple, having two separate checks won't add a lot of complexity.  Decided to keep two separate checks.  Thanks for the discussion. Not a huge tradeoff to consider in this scenario, will be helping in future decisions.

stablehlo/transforms/StablehloLegalizeAToB.cpp
abhigunj: created https://github.com/openxla/stablehlo/issues/2485 to track this as further enhancements to the pass.

stablehlo/tests/transforms/stablehlo_legalize_qdq_to_quantized_op.mlir
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// -----
+
+// CHECK-LABEL @failed_to_match_all_operands_not_quantized
+// CHECK:       %[[oper0:.*]] = stablehlo.uniform_quantize %arg0 : (tensor<16x16xf32>) -> tensor<16x16x!quant.uniform<u8:f32, 3.400000e+01:16>>
+// CHECK-NEXT:  %[[oper1:.*]] = stablehlo.uniform_quantize %arg1 : (tensor<16x16xf32>) -> tensor<16x16x!quant.uniform<u8:f32, 3.400000e+01:16>>
+// CHECK-NEXT:  %[[add_result:.*]] = stablehlo.add %[[oper0]], %[[oper1]] : tensor<16x16x!quant.uniform<u8:f32, 3.400000e+01:16>>
+// CHECK-NEXT:  return %[[add_result]] : tensor<16x16x!quant.uniform<u8:f32, 3.400000e+01:16>>
+func.func @failed_to_match_all_operands_not_quantized(%arg0: tensor<16x16xf32>, %arg1: tensor<16x16xf32>) -> tensor<16x16x!quant.uniform<ui8:f32, 34.0:16>> {
```
sdasgup3, line 83: Why is this a failed to match case?

sdasgup3 (approved):
with one final comment ...

stablehlo/tests/transforms/stablehlo_legalize_qdq_to_quantized_op.mlir
abhigunj: ah, I thought I removed the test since it was not needed. It somehow renamed it. 

removed now. thank you for the review.

GleasonK (approved):

PR #2480, opened 2024-08-08
Minor spec fix : correct variadic result notation for all_gather and all_reduce ops

abhigunj:


ghpvnist (approved):

docs/spec.md
```diff
@@ -823,7 +823,7 @@ Afterwards, within each `process_group`:
 // %operand0@(1, 0): [5, 6, 7, 8]
 // %operand1@(0, 0): [9, 10, 11, 12]
 // %operand1@(1, 0): [13, 14, 15, 16]
-%result = "stablehlo.all_reduce"(%operand0, %operand0) ({
+%result:2 = "stablehlo.all_reduce"(%operand0, %operand0) ({
```
sdasgup3, line 826: Also change the examples in StableHLOOps.td

sdasgup3 (approved):

docs/spec.md
abhigunj: StableHLOOps.td examples are non variadic, don't need the change.

sdasgup3: [Something to consider for a follow up]
Can we have the same example in both places as done for other ops?

ghpvnist: Let's keep the example same per 7.v in https://github.com/openxla/stablehlo/blob/main/docs/reference_checklist.md#after-implementing-the-op.

abhigunj: https://github.com/openxla/stablehlo/pull/2481  will match the examples

PR #2481, opened 2024-08-08
Collectives Ops : Match example from the StableHLO op description comment with example from the spec

abhigunj:
as per https://github.com/openxla/stablehlo/blob/main/docs/reference_checklist.md#after-implementing-the-op.

### DIFF
--- a/stablehlo/dialect/StablehloOps.td
+++ b/stablehlo/dialect/StablehloOps.td
@@ -1356,11 +1356,13 @@ def StableHLO_AllGatherOp : StableHLO_Op<"all_gather",
 
     Example:
     ```mlir
-    %result = "stablehlo.all_gather"(%operand) {
+    %result:2 = "stablehlo.all_gather"(%operand0, %operand1) {
       all_gather_dim = 1 : i64,
       replica_groups = dense<[[0, 1]]> : tensor<1x2xi64>,
+      // channel_id = 0
       channel_handle = #stablehlo.channel_handle<handle = 0, type = 0>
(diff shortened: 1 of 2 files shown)


stablehlo/dialect/StablehloOps.td
```diff
@@ -1356,11 +1356,13 @@ def StableHLO_AllGatherOp : StableHLO_Op<"all_gather",
 
     Example:
     ```mlir
-    %result = "stablehlo.all_gather"(%operand) {
+    %result:2 = "stablehlo.all_gather"(%operand0, %operand1) {
       all_gather_dim = 1 : i64,
       replica_groups = dense<[[0, 1]]> : tensor<1x2xi64>,
+      // channel_id = 0
```
ghpvnist, line 1362: Let's remove the default value (`channel_id` and `use_global_device_ids` from the td file).

ghpvnist (approved):
LGTM with one comment.

sdasgup3 (approved):

stablehlo/dialect/StablehloOps.td
abhigunj: why ? 
from the spec too?

ghpvnist: Just some old decision made in 7.iii https://github.com/openxla/stablehlo/blob/main/docs/spec_checklist.md. FYI `use_global_device_ids` is a `UnitAttr` in MLIR, so should be written as `use_global_device_ids` (without the `=false`).

abhigunj: Done, thanks for the reference.

Issue #2452, opened 2024-07-22
Fix acos op decomposition for non complex numbers to be more accurate


GleasonK:
### What happened?

An initial fix was made for accuracy but was reverted due to a bug in the pattern generator:
https://github.com/openxla/stablehlo/pull/2449

This ticket tracks fixing that work and submitting a more accurate decomposition for acos op.

```
Current:
// Expand acos for non-complex arguments to MHLO dialect as follows:
//   acos(x) = 2 * atan2(sqrt(1 - x^2), (1 + x))  if x != -1

Using 1-x^2 will be inaccurate when abs(x) is close to 1. The corresponding cancellation errors are avoided when using (1-x)*(1+x).
```

PR #2496, opened 2024-08-17
Refactor complex asin/acos/asinh/acosh to use asin_acos_kernel. Improve real acos accuracy.

pearu:
As in the title.

This PR introduces asin_acos_kernel operation that implements modified Hull et al algorithm for evaluating complex asin, acos, asinh, and acosh operations. This task corresponds to the refactor request in https://github.com/openxla/stablehlo/pull/2411#discussion_r1662810017 and it resolves https://github.com/pearu/functional_algorithms/issues/15.

In addition, the PR also improves the real acos accuracy as a fix to https://github.com/openxla/stablehlo/issues/2452 . As a result, the accuracy of single-precision real acos improved as follows (using 1 million samples):
```
                             main       this PR
ULP difference == 0 count is 971713  -> 999045
ULP difference == 1 count is 28158   ->    962
ULP difference == 2 count is 54      ->      0
ULP difference == 3 count is 20      ->      0
ULP difference >= 4 count is 62      ->      0
```

The cause of the revert by PR https://github.com/openxla/stablehlo/pull/2449 (`acos(-1) -> 0` while expecting `pi`) is fixed in functional_algorithms which now also generates extra samples that correspond to limiting values of real acos. For instance, `acos(-1)->pi`, `acos(nextafter(-1, -inf))->nan`, and `acos(nextafter(-1, int)-> approx pi` are now included in the tests.

The required functional_algorithms version is now 0.9 which includes a fix to `fa.utils.real_samples` to return samples that are distributed uniformly with respect to ULP differences of neighboring samples. This fix lead to updates of `stablehlo/tests/math/*.mlir`.


pearu:
^ @GleasonK

stablehlo/dialect/ChloOps.td
```diff
@@ -471,6 +471,21 @@ class CHLO_UnaryElementwiseOp<string mnemonic, list<Trait> traits,
   }];
 }
 
+def CHLO_AsinAcosKernelOp : CHLO_UnaryElementwiseOp<"asin_acos_kernel",
+    [HLO_CompatibleOperandsAndResultType], HLO_AnyComplexTensor> {
+  let summary = "AsinAcosKernel operator";
+
+  let description = [{
+    Returns `AsinAcosKernel(operand)` element-wise.
```
GleasonK, line 479: Can we add more description:

"This op is used as an intermediate value in decompositions and should never be constructed directly by frameworks or consumed by backends."

And maybe rename this to `_asin_acos_kernel` if MLIR permits this.

GleasonK: The alternative here would be to generate a CustomCall with the name `_asin_acos_kernel` and then write a pattern which decomposes that custom call. That may be more in-model, but as long as this op remains an impl detail, a refactor shouldn't be difficult if we want to go this route.

GleasonK (approved):
LGTM. Big fan of the unified decompositions!

Before merge I'm going to patch and run JAX tests before approving to make sure there are no surprises. In the future we'll have https://github.com/openxla/stablehlo/pull/2441 landed which should hopefully provide adequate CHLO / JAX coverage to avoid integration issues.

Will try to get that done now, probably merge in the morning if tests come back green.

stablehlo/dialect/ChloOps.td
pearu: I have updated asin_acos_kernel docs and renamed it to `_asin_acos_kernel`.

I like the custom call idea. IIUC, this should reduce chlo_legalize_to_stablehlo.mlir size as well.

PR #2501, opened 2024-08-19
[tosa] Make legalization pass operate top-down

jpienaar:
Some of these patterns, e.g., the reduce conversion one, assumes that the reduce op is converted before the body. But this wasn't required/enforced.


sdasgup3 (approved):

cc10512 (approved):
LGTM, it provides the right solution to converting reductions.

Would be good to also have a test case attached -- not clear to me why the test in https://github.com/openxla/stablehlo/blob/d1b9d1ef7648b224d0211aa12dc11ad6ef427def/stablehlo/conversions/tosa/tests/binary.mlir#L203 did not catch the problem.

jpienaar:
Closing as I don't have time to finish this debugging at the moment.

PR #2504, opened 2024-08-21
Move exported files behind filegroups

GleasonK:
It is problematic to include StableHLO source files from non-StableHLO projects. It makes renaming files and dependencies much more difficult.

Hiding file names behind a target that lives in StableHLO should help with the file renames case. There isn't a good way to keep a dependency list in StableHLO, but the only dependency of python bindings should be CAPI so that shouldn't be too problematic.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -21,11 +21,9 @@ package(
 
 exports_files([
     "LICENSE",
-    "stablehlo/integrations/python/ChloModule.cpp",
+    # Export to be used by MLIR-free python binding impls
     "stablehlo/integrations/python/StablehloApi.cpp",
     "stablehlo/integrations/python/StablehloApi.h",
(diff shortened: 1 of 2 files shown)


sdasgup3 (approved):

abhigunj (approved):

BUILD.bazel
```diff
@@ -1021,6 +1026,15 @@ gentbl_filegroup(
     ],
 )
 
+filegroup(
+    name = "stablehlo_py_api_files",
+    srcs = [
+        "stablehlo/integrations/python/PortableApi.cpp",
```
abhigunj, line 1032: StablehloApi.h & StablehloApi.cpp

ghpvnist (approved):

BUILD.bazel
GleasonK: good catch

PR #2506, opened 2024-08-22
Add a test for non quantized type (DotGeneralOp mismatched operand and result types scenario) 

abhigunj:
Test at `/testdata/quantized/` is for Quantized type (ref PR https://github.com/openxla/stablehlo/pull/2503). Adding a test for non quantized types.


stablehlo/tests/interpret/dot_general.mlir
```diff
@@ -56,3 +56,19 @@ func.func @dot_general_op_test_empty_dims() {
                                          [[4, 0], [0, 4]]]]> : tensor<2x2x2x2xi64>
   func.return
 }
+
+// -----
+
+func.func @dot_general_op_mismatch_operand_and_result_float_types() {
```
ghpvnist, line 62: Nit: Mismatch sounds like a negative test, how about `dot_general_op_test_different_operand_and_result_element_types`?

ghpvnist (approved):

PR #2511, opened 2024-08-27
Add best effort validation for known dot algorithms

GleasonK:
This is somewhat of ecosystem implementation details leaking into StableHLO, but that seems to be safer than allowing any arbitrary combination and enforcing that the user get some lucky combination correct.

### DIFF
--- a/stablehlo/tests/interpret/dot_general.mlir
+++ b/stablehlo/tests/interpret/dot_general.mlir
@@ -29,7 +29,7 @@ func.func @dot_general_op_test_algorithm() {
     algorithm = <
       lhs_precision_type = tf32,
       rhs_precision_type = tf32,
-      accumulation_type = tf32,
+      accumulation_type = f32,
       lhs_component_count = 1,
       rhs_component_count = 1,
       num_primitive_operations = 1,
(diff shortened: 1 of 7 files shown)


stablehlo/tests/interpret/dot_general.mlir
```diff
@@ -29,7 +29,7 @@ func.func @dot_general_op_test_algorithm() {
     algorithm = <
       lhs_precision_type = tf32,
```
sdasgup3, line 30: The dot general interpreter currently uses the result type as the accumulator type 
https://github.com/openxla/stablehlo/blob/8602e098be0aaba038868b943963dd23141eb605/stablehlo/reference/Ops.cpp#L1724

Should we aim to have the interpreter support for some basic cases like lhs/rhs/accumulator type not equal to `tf32` (as I believe types like `tf32` are more like an XLA artifact) and when the component/primitive operations counts are 1.

sdasgup3 (approved):

stablehlo/tests/interpret/dot_general.mlir
GleasonK: TensorFloat32 is an NVIDIA thing (I also accidentally thought it was tensorflow related), and has upstream MLIR/LLVM support so maybe we can support it as well. I agree that we should make these updates! Maybe in a followup, can make a ticket

PR #2518, opened 2024-08-30
Add StableHLOCreateCompatibilityExpanderPass

abhigunj:
Backporting from g3..

This pass expands backward compatibility with older StableHLO versions by decomposing newer StableHLO operations into equivalent operations supported by those older versions.

### DIFF
--- a/stablehlo/transforms/StablehloCreateCompatibilityExpander.cpp
+++ b/stablehlo/transforms/StablehloCreateCompatibilityExpander.cpp
@@ -18,6 +18,7 @@ limitations under the License.
 #include "llvm/Support/ErrorHandling.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/IR/BuiltinAttributes.h"
+#include "mlir/IR/BuiltinTypes.h"
 #include "mlir/IR/PatternMatch.h"
 #include "mlir/Support/LLVM.h"
 #include "mlir/Transforms/DialectConversion.h"
@@ -39,6 +40,9 @@ namespace {
 
 // Creates a constant with all ones.
 static Value createConstantWithAllOnes(OpBuilder &b, Location loc, Value val) {
+  if (!isa<FloatType>(getElementTypeOrSelf(val)))
+    llvm_unreachable("Unsupported element type, expecting float");
+
   auto shapedTy = dyn_cast<mlir::ShapedType>(val.getType());
   if (!shapedTy) llvm_unreachable("Unsupported shaped type.");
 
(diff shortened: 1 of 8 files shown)


stablehlo/transforms/StablehloCreateCompatibilityExpander.cpp
```diff
@@ -0,0 +1,128 @@
...
+//===----------------------------------------------------------------------===//
+// Helpers.
+//===----------------------------------------------------------------------===//
+
+// Creates a constant with all ones.
+static Value createConstantWithAllOnes(OpBuilder &b, Location loc, Value val) {
+  auto shapedTy = dyn_cast<mlir::ShapedType>(val.getType());
+  if (!shapedTy) llvm_unreachable("Unsupported shaped type.");
+
+  mlir::DenseElementsAttr elementsAttr =
```
sdasgup3, line 45: I think as `createConstantWithAllOnes` is currently being called using value as the result of `tan(f32)`, we are (safely) assuming that  `val` type is always a float. Can we codify the assumption here? For example, unsupported arror for  `getElementTypeOrSelf(value)` =1 fp, Or rename the function to  `createFloatConstantWithAllOnes` or `createFloatConstantLike(..., value, double)`?

sdasgup3 (approved):

sdasgup3 (changes requested):
The changes can go in a folllow up pr as well.

stablehlo/transforms/StablehloCreateCompatibilityExpander.cpp
abhigunj: Can we codify the assumption here? >> Done. Thanks for the review.

abhigunj:
> The changes can go in a folllow up pr as well.

took care in this PR. ty.

sdasgup3 (approved):

stablehlo/transforms/StablehloCreateCompatibilityExpander.cpp
GleasonK: Maybe we should move `StableHLO_CreateConstantLike` to somewhere shared, like StablehloOps or Base (I think would need to be ops.cpp because its StableHLO specific), and use that definition here as well, similar to CHLO.

```
// Express `tan(a + bi)` as
//   (tan(a) + i tanh(b)) / (1 - i tan(a) * tanh(b))
def : Pat<(CHLO_TanOp ComplexElementType:$input),
  (StableHLO_DivOp
    (StableHLO_ComplexOp
      (CHLO_TanOp:$tan (StableHLO_RealOp $input)),
      (StableHLO_TanhOp:$tanh (StableHLO_ImagOp $input))),
    (StableHLO_ComplexOp
      (StableHLO_ConstantLike<"1.0"> $tan),
      (StableHLO_NegOp (StableHLO_MulOp $tan, $tanh)))
  )>;
```

GleasonK: Otherwise I think this is fairly duplicate code

abhigunj: Thanks for the feedback. I will take care of moving to ops.cpp and using `StableHLO_ConstantLike`  in follow up PR.

PR #2520, opened 2024-08-31
Fix linking of stablehlo_capi_objects

wsmoses:


hawkinsp:
Looks like I hit the same issue: https://github.com/openxla/stablehlo/pull/2528

I think my fix is the correct one.

sdasgup3:
reviewed https://github.com/openxla/stablehlo/pull/2528. Feel free to close this one as duplicate?

PR #2528, opened 2024-09-05
Fix stablehlo_capi_objects to depend on :CAPIIRObjects, not :CAPIIR.

hawkinsp:
Depending on :CAPIIR from a capi_objects target causes duplicate symbol problems in JAX on Windows.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -941,7 +941,7 @@ cc_library(
         ":stablehlo_serialization",
         ":version",
         "@llvm-project//llvm:Support",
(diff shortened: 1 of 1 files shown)


sdasgup3 (approved):

wsmoses:
So I had tried a fix like this first and I recall hitting the same issue as the other APIs also were doubly linked.

If I’m misremembering and it works, great!

If not I can retry in a couple of hours

wsmoses:
never mind, this looks to work -- closing other PR

PR #2530, opened 2024-09-06
Fix JAX windows ci build error : missing stablehlo C API symbols

abhigunj:
Issue: JAX windows build is failing because of missing stablehlo C API symbols
https://github.com/google/jax/actions/runs/10739809804/job/29786443538

root cause: `StablehloApi.h` defs has function overloads. Compiler does name mangling (decorating function names with additional information).  These symbols are missing in JAX APIs,  JAX only allowlist symbol exports for symbols starting with `stablehlo`, but the mangled c++ names don't have that property.  

fix:
1. add `extern "C" `  around stablehloapi.h declarations. extern "C" instructs the compiler to suppress the mangling.
2. rename functions to avoid function overloading

thank you @hawkinsp  for help with root causing and validating the fix.

### DIFF
--- a/stablehlo/integrations/c/StablehloApi.h
+++ b/stablehlo/integrations/c/StablehloApi.h
@@ -16,6 +16,10 @@ limitations under the License.
 #include "mlir-c/IR.h"
 #include "mlir-c/Support.h"
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 // Get the current StableHLO API version.
 //
 // This value is incremented as needed to help integrate API changes.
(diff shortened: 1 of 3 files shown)


stablehlo/integrations/c/StablehloApi.h
```diff
@@ -104,12 +112,7 @@ MLIR_CAPI_EXPORTED MlirLogicalResult stablehloDeserializePortableArtifact(
 // StableHLO, e.g. if it's using incompatible features.
 //
 // Returns empty module on failure.
-MLIR_CAPI_EXPORTED MlirModule stablehloDeserializePortableArtifact(
-    MlirStringRef artifactStr, MlirContext ctx);
-
-// Call the Interpreter, returns MlirArrayAttr of dense element
-// MlirAttribute results
-MLIR_CAPI_EXPORTED MlirModule stablehloDeserializePortableArtifact(
```
abhigunj, line 112: removed duplicate `MLIR_CAPI_EXPORTED MlirModule stablehloDeserializePortableArtifact(` declaration

ghpvnist (approved):

Issue #2524, opened 2024-09-04
VhloToVersionPass to allow `current`  as a target version?


abhigunj:
VhloToVersionPass description (below) says target version can be `current` but the pass does not handle it. Should we support it or update the description to remove `current`?

```
def VhloToVersionPass : Pass<"vhlo-to-version"> {
  let summary = "Convert between versions of VHLO.";
  let options = [
    Option<"targetVersionOption", "target", "std::string", "",
           "The target version. Must be a version of the form #.#.# or 'current'.">,
  ];
}

```

abhigunj:
cc: @GleasonK

sdasgup3:
`current but does not handle it..`

Can you elaborate what you mean here?

abhigunj:
> `current but does not handle it..`
> 
> Can you elaborate what you mean here?

`validateTargetVersion` does not allow `current` version. 
https://github.com/openxla/stablehlo/blob/main/stablehlo/transforms/VhloToVersion.cpp#L72C20-L72C41

validated with simple test module. `stablehlo-opt -vhlo-to-version='target=current' t1.mlir`.  Original pass was handling it, lateron particular code was removed.

PR #2532, opened 2024-09-06
Update VhloToVersionPass  description

abhigunj:
fixes https://github.com/openxla/stablehlo/issues/2524 
discussed with Kevin,  don't need to support `'target=current'`


ghpvnist (commented):
Can we add a test to confirm this?

abhigunj:
> Can we add a test to confirm this?

That would be negative test, don't think there is a value in adding negative tests for invalid pass argument here.

> Can we add a test to confirm this?

we do have invalid argument test https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/vhlo/vhlo_to_version_invalid.mlir#L24
we don't need new test for another (`current` in this case) invalid argument.

ghpvnist (approved):

PR #2534, opened 2024-09-09
Enhance stablehlo-quant-legalize-to-tosa-rescale

Tai78641:
by adding support for following quantized operators:
  -  abs
  -  multiply and divide
  -  maximum and minimum
  -  compare

### DIFF
--- a/stablehlo/conversions/tosa/transforms/StablehloQuantLegalizeToTosaRescale.cpp
+++ b/stablehlo/conversions/tosa/transforms/StablehloQuantLegalizeToTosaRescale.cpp
@@ -96,8 +96,165 @@ Value buildRescaleFromInt32(PatternRewriter& rewriter, Location loc,
                       /*perChannel=*/false);
 }
 
+using UnaryRescaleScalesFn =
+    void (*)(const quant::UniformQuantizedType &operandQType,
+             const quant::UniformQuantizedType &resultQType,
+             double &operandRescaleScale, double &resultRescaleScale);
+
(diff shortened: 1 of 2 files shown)


stablehlo/conversions/tosa/transforms/StablehloQuantLegalizeToTosaRescale.cpp
```diff
@@ -170,20 +304,115 @@ LogicalResult matchAndRewriteAddSub(StablehloOp op, PatternRewriter& rewriter) {
...
+  if (!lhsQType || !rhsQType) {
+    return rewriter.notifyMatchFailure(
+        op,
+        "The conversion supports operands with per-tensor quantized "
+        "types only");
+  }
+
+  double lhsScale = lhsQType.getScale();
+  double rhsScale = rhsQType.getScale();
+  double maxScale = std::max(lhsScale, rhsScale);
```
sdasgup3, line 361: Just curious to know why the max scale for add/sub is twice the max scale used for compare/max,min. Is that depends on some tosa backend details?

sdasgup3 (approved):
Thanks Tai!
The PR looks good except for a query that I have.

stablehlo/conversions/tosa/transforms/StablehloQuantLegalizeToTosaRescale.cpp
Tai78641: I dont know for sure. my guess is add/sub of 2 numbers may double the values whereas min/max of two numbers would not?

Issue #2539, opened 2024-09-12
Mismatch between specification and tests for quantized types


jtristan:
### What happened?

The tests in `stablehlo/stablehlo/tests/interpret` and `stablehlo/stablehlo/testdata` use a different syntax than what is described in the specification. Here is a grammar that describes the actual syntax of quantized types in those tests:

```
QuantizedTensorType ::= 'tensor' '<' Shape QuantizedTensorElementType '>'
QuantizedTensorElementType ::= '!quant.uniform' '<'
                  QuantizationStorageType
                  ['<' QuantizationStorageMin ':' QuantizationStorageMax '>']
                  ':' QuantizationExpressedType
                  [':' QuantizationDimension]
                  ',' QuantizationParameters '>'
QuantizationStorageType ::= IntegerType
QuantizationStorageMin ::= **IntegerLiteral**
QuantizationStorageMax ::= **IntegerLiteral**
QuantizationExpressedType ::= FloatType
QuantizationDimension ::= **IntegerLiteral**
QuantizationParameters ::= QuantizationParameter
                         | '{' QuantizationParameter {',' QuantizationParameter} '}'
QuantizationParameter ::= QuantizationScale **[ ':' QuantizationZeroPoint ]**
QuantizationScale ::= **FloatLiteral**
QuantizationZeroPoint ::= **IntegerLiteral**
```

where ** and ** emphasize where the differences with the specification are

sdasgup3:
Dear @jtristan 
I am assuming that  you refer to the absence of the non terminals highlighted above. Please note that  `QuantizationStorageMin`, `QuantizationStorageMax`. If not present inferred from the storage type.
`QuantizationDimension` is relevant only for per-axis quantization and absent for per-tensor. 
`QuantizationDimension` is optional if the zero-point is zero. 

Please let me know if I am missing any discrepancies.

jtristan:
No, the discrepancies are that the spec defines the grammar of `QuantizationStorageMin`, `QuantizationStorageMax`, `QuantizationDimension `, `QuantizationScale `, and `QuantizationZeroPoint ` respectively as `IntegerConstant `, `IntegerConstant `, `IntegerConstant `, `FloatConstant `, `IntegerConstant ` but that it should be `IntegerLiteral`, `IntegerLiteral`, `IntegerLiteral`, `FloatLiteral`, `IntegerLiteral` instead. Finally, the spec doesn't capture that `QuantizationZeroPoint ` is optional.

[...]

PR #2541, opened 2024-09-12
Fix Quant grammar

sdasgup3:
fixes https://github.com/openxla/stablehlo/issues/2539


abhigunj (approved):

GleasonK (approved):
> respectively as IntegerConstant , IntegerConstant , IntegerConstant , FloatConstant , IntegerConstant but that it should be IntegerLiteral, IntegerLiteral, IntegerLiteral, FloatLiteral, IntegerLiteral instead.

Looks like this is addressed

> Finally, the spec doesn't capture that QuantizationZeroPoint is optional."

Also looks like this is addressed. LGTM.

PR #2551, opened 2024-09-17
Async RFC

chaserileyroberts:
This RFC proposes adding several async features to StableHLO, mainly `async_start`, async_done`, and the `async<...>` Non-value type.

The end goal of this RFC is to allow JAX users to define their own collective matmul schedules, and allow more control to potentially better utilize their hardware.


rfcs/20240917-async-support.md
```diff
@@ -0,0 +1,133 @@
...
+// %init_sum: 3
+%future = "stablehlo.async_start"(
+    %init_i as %arg0: tensor<i64>, 
+    %init_sum as %arg1: tensor<i64>) 
+{
+    %new_sum = stablehlo.add %arg1, %arg0 : tensor<i64>
+    stablehlo.return %new_sum : tensor<i64>
+} : (tensor<i64>, tensor<i64>) -> async<tensor<i64>>
+
+%result = "stablehlo.async_done"(%future): async<tensor<i64>> -> tensor<i64>
```
ezhulenev, line 81: Let's make `async_done` variadic and:

1. Await on all async results nicely mapped to `async-done` HLO
2. Await on one async result can be represented with `async-update` (needs fact checking), but initially we don't need to implement it

ezhulenev: nvm, I see it's already variadic

rfcs/20240917-async-support.md
```diff
@@ -0,0 +1,133 @@
...
+// %result: 5
+```
+
+
+## Execution
+
+### Async Execution
+
+Stable HLO programs are usually defined as simple sequential operations performed one after another,
```
ezhulenev, line 129: Actually StableHLO (and HLO) has a dataflow semantics, and backend is free to reorder execution as long as it respects data dependencies (at HLO we also have control deps, but they are internal implementation detail)

rfcs/20240917-async-support.md
ezhulenev: I like when type is defined on the same line as op name, I think this should be parseable in MLIR. Also for `done` return type can be inferred from arguments, no need to spell it. But these details can be refined later.

```mlir
// HLO async-start
%f0, %f1 = stablehlo.async_start(...) -> async<tensor<f32>>, async<tensor<f32>> {
  %0 = ... : tensor<f32>
  %1 = ... : tensor<f32>
  stablehlo.return %0, %1 : tensor<f32>, tensor<f32>
}

// HLO async-done
%t0, %t1 = stablehlo.async_done %f0, %f1 : async<tensor<f32>>, async<tensor<f32>>
```

chaserileyroberts: So should the the return type of `async_start` be `(async<R0>, ..., async<RM>)` instead of `async<(R0, ..., RM)>`?

ezhulenev: Yes, because conceptually it should be possible to await on just one result. On GPU that should have a straightforward lowering to streams and events

rfcs/20240917-async-support.md
chaserileyroberts: I'll reword this.

rfcs/20240917-async-support.md
chaserileyroberts: That makes sense. I'll update the spec then.

rfcs/20240917-async-support.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Ops
+
+### async_start
+
+#### Semantics
+
+Produces the output from executing the `body` function, but runs all operations on a stream separate from the main compute stream.
+
+The output of an `async_start` computation must first be processed by an `async_done` operation.
```
frgossen, line 48: Do you expect to be able to pipe this through control flow. This is significantly more complicated if you want to pipe this through while and conds.

rfcs/20240917-async-support.md
```diff
@@ -0,0 +1,133 @@
...
+is to expose this in stableHLO and have it accessible in JAX's `shard_map`.
+
+
+## Proposed Specification changes
+
+### Types
+
+
+```ebnf
+AsyncType ::= 'async' '<' ValueType '>'
```
frgossen, line 29: Why do we need an async data type here? Within XLA, we encode this through tuples that forward operands and interim result from async start to async done ops. Would the same work on the stable HLO level? 

Where I think this will be a bit special is when you want to guarantee that the values are not copied on loop boundaries, or generally around control flow. An alternative would be to introduce non-copyable values (buffers (?)).

frgossen (changes requested):

rfcs/20240917-async-support.md
ezhulenev: In HLO we use tuples because it's very hard to add new types to HLO. In MLIR adding types is very easy and natural. If extending HLO would not be that hard, I'd vote for adding async type to it as well.

frgossen: Makes sense. We will have to lower it to the tuples anyways no? So this is really some form of syntactic sugar.

rfcs/20240917-async-support.md
frgossen: Yes, for SHLO this should be simple. I was rather wondering about the lowerings through XLA

rfcs/20240917-async-support.md
ezhulenev: Yes, I **think** async type <-> async bundle (tuples) representations are isomorphic and always can be converted back and forth:

1. `stablehlo.async_start` -> `%start = (args, results, sync-flag) async_start`
2. `stablehlo.async_done %ret0, %ret1, ...` -> `async-done (get-tuple-element)`
3. Tricky case: `stablehlo.async_done %ret0` just one of the returned values -> `async-update (get-tuple-element`), effectively peels N result buffers and M argument buffers from a tuple, and allows buffer assignment to reuse them (sorry, only internal link for Frederik http://goto.google.com/async-update-peeling). This is underspecified in HLO, and we don't need to support it today, and require that `async_done` must await on ALL results of corresponding start operation.

rfcs/20240917-async-support.md
chaserileyroberts: For now we would be ok with this not needing to cross control flow boundaries, but I could see the need eventually.

chaserileyroberts:
Closing this as we move to extending `jax.compute_on` instead

PR #2570, opened 2024-10-01
Add EvalTranspose pattern to StablehloAggressiveFolder

mvpant:
This patch folds `stablehlo.transpose` operation with constant operand into `stablehlo.constant`.

I have considered doing this by iterating over source index space, i.e.
```
auto initialValue = *std::begin(data);
SmallVector<ElementType> result(resultType.getNumElements(), initialValue);

for (int64_t i = 0; i < operandType.getNumElements(); ++i) {
  auto srcDimIndex = delinearize(i, operandStrides);
  auto dstDimIndex = applyPermutation(srcDimIndex, permutation);
  auto dstLinearIndex = linearize(dstDimIndex, resultStrides);
  result[dstLinearIndex] = data[i];
}
```
but that requires preinitialization of result vector with some value, which is twice as slow on simple case:
```
func.func @eval_transpose() -> (tensor<5000x80x30xi32>) {
  %0 = stablehlo.iota dim = 0 : tensor<30x80x5000xi32>
  %1 = stablehlo.transpose %0, dims = [2, 1, 0] : (tensor<30x80x5000xi32>) -> tensor<5000x80x30xi32>
  func.return %1 : tensor<5000x80x30xi32>
}
```

### DIFF
--- a/stablehlo/tests/transforms/stablehlo_aggressive_folder.mlir
+++ b/stablehlo/tests/transforms/stablehlo_aggressive_folder.mlir
@@ -163,3 +163,14 @@ func.func @eval_transpose_zerorank() -> tensor<i32> {
   %1 = stablehlo.transpose %0, dims = [] : (tensor<i32>) -> tensor<i32>
   func.return %1 : tensor<i32>
 }
+
+// -----
+
+// CHECK-LABEL: func @eval_transpose_splat
+func.func @eval_transpose_splat() -> (tensor<10x3x1xi32>) {
+  // CHECK: [[RESULT0:%.*]] = stablehlo.constant dense<1> : tensor<10x3x1xi32>
+  // CHECK: return [[RESULT0]]
+  %0 = stablehlo.constant dense<1> : tensor<3x1x10xi32>
+  %1 = stablehlo.transpose %0, dims = [2, 0, 1] : (tensor<3x1x10xi32>) -> tensor<10x3x1xi32>
+  func.return %1 : tensor<10x3x1xi32>
+}
--- a/stablehlo/tests/transforms/stablehlo_aggressive_folder.mlir
+++ b/stablehlo/tests/transforms/stablehlo_aggressive_folder.mlir
@@ -110,3 +110,56 @@ func.func @eval_convert_f64_precision_loss() -> (tensor<1xf32>, tensor<f32>) {
   %3 = stablehlo.convert %1 : (tensor<f64>) -> tensor<f32>
   func.return %2, %3 : tensor<1xf32>, tensor<f32>
 }
+
+// -----
+
+// CHECK-LABEL: func @eval_transpose
+func.func @eval_transpose() -> (tensor<2x3x2xi32>, tensor<2x4x3xi32>, tensor<4x3x2xi32>) {
+  // CHECK: [[RESULT0:%.*]] = stablehlo.constant dense<
+  // CHECK-SAME: {{\[\[}}[1, 7], [3, 9], [5, 11]],
+  // CHECK-SAME:   {{\[}}[2, 8], [4, 10], [6, 12]]]> : tensor<2x3x2xi32>
+  //
+  // CHECK: [[RESULT1:%.*]] = stablehlo.constant dense<
+  // CHECK-SAME: {{\[\[}}[1, 3, 5], [7, 9, 11], [13, 15, 17], [19, 21, 23]],
+  // CHECK-SAME:   {{\[}}[2, 4, 6], [8, 10, 12], [14, 16, 18], [20, 22, 24]]]> : tensor<2x4x3xi32>
+  //
+  // CHECK: [[RESULT2:%.*]] = stablehlo.constant dense<
+  // CHECK-SAME: {{\[\[}}[1, 2],  [3, 4],  [5, 6]]
+  // CHECK-SAME:   {{\[}}[7, 8],  [9, 10], [11, 12]],
+  // CHECK-SAME:   {{\[}}[13, 14], [15, 16], [17, 18]],
+  // CHECK-SAME:   {{\[}}[19, 20], [21, 22], [23, 24]]]> : tensor<4x3x2xi32>
+  //
+  // CHECK: return [[RESULT0]], [[RESULT1]], [[RESULT2]]
+  %0 = stablehlo.constant dense<[[[1,2], [3,4], [5,6]],
+                                 [[7,8], [9,10], [11,12]]]> : tensor<2x3x2xi32>
+  %1 = stablehlo.constant dense<[[[1, 2],  [3, 4],  [5, 6]],
+                                 [[7, 8],  [9, 10], [11,12]],
+                                 [[13,14], [15,16], [17,18]],
+                                 [[19,20], [21,22], [23,24]]]> : tensor<4x3x2xi32>
+  %2 = stablehlo.transpose %0, dims = [2, 1, 0] : (tensor<2x3x2xi32>) -> tensor<2x3x2xi32>
+  %3 = stablehlo.transpose %1, dims = [2, 0, 1] : (tensor<4x3x2xi32>) -> tensor<2x4x3xi32>
+  %4 = stablehlo.transpose %3, dims = [1, 2, 0] : (tensor<2x4x3xi32>) -> tensor<4x3x2xi32>
+  func.return %2, %3, %4 : tensor<2x3x2xi32>, tensor<2x4x3xi32>, tensor<4x3x2xi32>
+}
+
+// -----
+
+// CHECK-LABEL: func @eval_transpose_zerodim
+func.func @eval_transpose_zerodim() -> (tensor<10x3x0xf32>) {
+  // CHECK: [[RESULT0:%.*]] = stablehlo.constant dense<> : tensor<10x3x0xf32>
+  // CHECK: return [[RESULT0]]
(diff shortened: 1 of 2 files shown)


stablehlo/transforms/StablehloAggressiveFolder.cpp
```diff
@@ -643,6 +643,58 @@ struct EvalIotaOpPattern : public OpRewritePattern<IotaOp> {
...
+      return rewriter.notifyMatchFailure(
+          op, "expected constant integer or float operand");
+
+    DenseElementsAttr resAttr;
+    if (auto data = els.tryGetValues<APInt>())
+      resAttr = transposeType(op, *data);
+    else if (auto data = els.tryGetValues<APFloat>())
+      resAttr = transposeType(op, *data);
+    else
+      return rewriter.notifyMatchFailure(op.getLoc(),
```
GleasonK, line 690: What happens when execution makes it here? Does the pass fail, or if this OK because we use a greedy pass application? I.e. the alternative would be to have some method like `return noFoldSuccess();` which won't cause the pass to fail but doesn't fold anything, similar to returning `{}` from an MLIR fold method

stablehlo/transforms/StablehloAggressiveFolder.cpp
```diff
@@ -643,6 +643,58 @@ struct EvalIotaOpPattern : public OpRewritePattern<IotaOp> {
...
+    auto srcDimIndex = applyPermutation(dstDimIndex, inversePermutation);
+    auto srcLinearIndex = linearize(srcDimIndex, operandStrides);
+    result.push_back(data[srcLinearIndex]);
+  }
+
+  return DenseElementsAttr::get(op.getResult().getType(),
+                                ArrayRef<ElementType>(result));
+}
+
+struct EvalTransposeOpPattern : public OpRewritePattern<TransposeOp> {
```
GleasonK, line 671: Let's have a comment for what this pattern does, the MHLO folder for example only handles `transpose(transpose(x))` and `transpose(splat)`, looks like this one will do everything, not sure what it does when it encounters a splat constant, probably another good test case.

stablehlo/tests/transforms/stablehlo_aggressive_folder.mlir
```diff
@@ -110,3 +110,56 @@ func.func @eval_convert_f64_precision_loss() -> (tensor<1xf32>, tensor<f32>) {
...
+  %0 = stablehlo.constant dense<> : tensor<3x0x10xf32>
+  %1 = stablehlo.transpose %0, dims = [2, 0, 1] : (tensor<3x0x10xf32>) -> tensor<10x3x0xf32>
+  func.return %1 : tensor<10x3x0xf32>
+}
+
+// -----
+
+// CHECK-LABEL: func @eval_transpose_zerorank
+func.func @eval_transpose_zerorank() -> tensor<i32> {
+  // CHECK: [[RESULT0:%.*]] = stablehlo.constant dense<1> : tensor<i32>
```
GleasonK, line 160: How about one more for a splat which gets transposed to make sure we don't expand the splat (or capture the behavior if we do):

`stablehlo.constant dense<1> : tensor<3x1x10xi32>`

GleasonK (approved):

stablehlo/tests/transforms/stablehlo_aggressive_folder.mlir
mvpant: Sure. MLIR marks newly generated data as a splat value by calculating a hash in `mlir::detail::DenseIntOrFPElementsAttrStorage::getKey (ty=..., data=..., isKnownSplat=false)` at `.../llvm-project/mlir/lib/IR/AttributeDetail.h:129` for `StorageUniquer`.

stablehlo/transforms/StablehloAggressiveFolder.cpp
```diff
@@ -643,6 +643,57 @@ struct EvalIotaOpPattern : public OpRewritePattern<IotaOp> {
...
+  RankedTensorType resultType = op.getResult().getType();
+
+  const auto operandStrides = computeStrides(operandType.getShape());
+  const auto resultStrides = computeStrides(resultType.getShape());
+  const auto inversePermutation = invertPermutationVector(op.getPermutation());
+
+  SmallVector<ElementType> result;
+  result.reserve(resultType.getNumElements());
+
+  for (int64_t i = 0; i < resultType.getNumElements(); ++i) {
```
sdasgup3, line 660: Can we check for identity permutation 
```
bool isIdentityPermutation(const std::vector<unsigned>& permutation) {
   std::vector<unsigned> expected(permutation.size());
   std::iota(expected.begin(), expected.end(), 0); // Fill with 0, 1, 2, ...
   return permutation == expected;
  }

  return true;
}
```

in which case 
```
if (isIdentityPermutation(op.getPermutation())) {
    return DenseElementsAttr::get(resultType, data); 
  }
 ```

sdasgup3 (approved):

stablehlo/transforms/StablehloAggressiveFolder.cpp
mvpant: done

stablehlo/transforms/StablehloAggressiveFolder.cpp
mvpant: That’s a good point! We can also check for splat values. However, I’m not sure if these checks should be placed in the folder; perhaps `StablehloAggressiveSimplification.cpp` would be a better location. To elaborate -- these fast paths could save some CPU time in the current implementation but in the end result would be the same. By placing them in `StablehloAggressiveSimplification.cpp`, users could simplify operations without the need to fold everything?

1) transpose(X) with an identity permutation => X
2) transpose(splat) with any permutation => splat

What do you think?

sdasgup3: The checks (including the splat one you mentioned) are equally valuable for opt-in passes like "aggressive-simplification" and folders invoked during operation creation using "OpBuilder::createOrFold," canonicalization, dialect conversion, and pattern rewriting with explicit calls to fold APIs. Unless we have a contract of "running the pass before folding," making both the pass and folder independently capable makes sense.

To achieve this, https://github.com/openxla/stablehlo/blob/main/stablehlo/transforms/PassUtils.cpp seems like a suitable place to hoist any common utilities, which in this case are the two canonicalization patterns you mentioned. The pass would utilize these canonicalization, and the folder could fold after applying them.
cc @GleasonK

GleasonK: Let's leave these separate for now. Regardless of where they're implemented I think they should be separate patterns to keep things more maintainable / readable, and as such can be a separate PR. I.e. I'd rather we have `EvalTransposeSplat` and `EvalIdentityTranspose` than just an `EvalTranspose` method which does 10 things.

As for what constitutes folding vs canonicalizing..need to think more. I'd think a "simplification of a constant" should be a fold. Whereas `transpose(transpose(X))` should be a simplification. By that logic, `transpose(splat)` would be fold, and `transpose(X) ID -> X` is a simplification.

GleasonK:
@mvpant wanted to give a heads up that I'm going to be doing some re-organizing of StablehloAggressiveSimplification in the coming days (not planning to touch folders this week), and add some patterns from MHLO canonicalizers. If you don't have any active changes..maybe hold off on starting any for a few days to avoid a nasty merge conflict.

mvpant:
> @mvpant wanted to give a heads up that I'm going to be doing some re-organizing of StablehloAggressiveSimplification in the coming days (not planning to touch folders this week), and add some patterns from MHLO canonicalizers. If you don't have any active changes..maybe hold off on starting any for a few days to avoid a nasty merge conflict.

@GleasonK sure. By the way, out of curiosity, what is the reason why stablehlo operations do not have `<builtin>` canonicalization using `let hasCanonicalizer = 1;` and folder `let hasFolder = 1;`? My guess is that this approach is intended to keep it stable for external users, ensuring that the generated output is predictable?

GleasonK:
> My guess is that this approach is intended to keep it stable for external users, ensuring that the generated output is predictable?

Exactly this. I've had too many issues with folks not being able to opt-out of certain patterns being applied and then they invent "clever" solutions like barrier ops to prevent constant folding, or mirrors of `convert` ops to avoid folding ops that would indicate that model weights are quantized, etc. In the current design point I'm trying to make everything as opt-in as possible, and have a stable input format which serialized and deserializes without changing.


Also I had a followup question, I see a few uses of `tensor::EmptyOp`, and I'm thinking to replace these with `stablehlo::ConstantOp : dense<>`. Any objections to this? I figure if tensor ops are needed it can be an extension pattern which converts empty constants to tensor::empty that lives in a part of the codebase that supports tensors, as opposed to a builtin canonicalization pattern

mvpant:
> ... invent "clever" solutions like barrier ops to prevent constant folding, or mirrors of `convert` ops to avoid folding ops that would indicate that model weights are quantized, etc.

Eh, doesn't sound good, makes much more sense now to keep emitted operations as is.
 
> Also I had a followup question, I see a few uses of `tensor::EmptyOp`, and I'm thinking to replace these with `stablehlo::ConstantOp : dense<>`. Any objections to this?

Nope. I think `tensor::EmptyOp` is handy at further stages because of the built-in bufferization stuff, but at `stablehlo` level it shouldn't make a difference. Having `stablehlo.constant` everywhere seems fine and more consistent.

PR #2573, opened 2024-10-02
Use `LINK_COMPONENTS` to link `LLVMSupport`

bjacob:
In #2494, `LLVMSupport` was added as a linked dependency to a few targets such as `ChloCAPI`, using the `add_mlir_public_c_api_library` function, passing `LLVMSupport` along with other dependencies under `LINK_LIBS`.

Trying to integrate these changes in https://github.com/iree-org/iree , I get these CMake errors:

```
CMake Error at build/lib/cmake/mlir/AddMLIR.cmake:265 (message):
  ChloCAPI specifies LINK_LIBS LLVMSupport, but LINK_LIBS cannot be used for
  LLVM libraries.  Please use LINK_COMPONENTS instead.
Call Stack (most recent call first):
  build/lib/cmake/mlir/AddMLIR.cmake:375 (_check_llvm_components_usage)
  build/lib/cmake/mlir/AddMLIR.cmake:637 (add_mlir_library)
  third_party/stablehlo/stablehlo/integrations/c/CMakeLists.txt:24 (add_mlir_public_c_api_library)
```

This PR is simply following the suggesting in that error message, and that seems to work.

### DIFF
--- a/stablehlo/integrations/c/CMakeLists.txt
+++ b/stablehlo/integrations/c/CMakeLists.txt
@@ -28,6 +28,8 @@ add_mlir_public_c_api_library(ChloCAPI
 
   LINK_LIBS PUBLIC
   ChloOps
+
+  LINK_COMPONENTS PUBLIC
   LLVMSupport
 )
 
@@ -40,7 +42,6 @@ add_mlir_public_c_api_library(StablehloCAPI
   StablehloTypes.cpp
 
   LINK_LIBS PUBLIC
-  LLVMSupport
   MLIRCAPIIR
   MLIRIR
   MLIRSupport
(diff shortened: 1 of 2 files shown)


bjacob:
@GleasonK @ScottTodd

bjacob:
This is actually failing to link in the CI.  I don't have that problem in my dependent project (IREE) where Stablehlo is external. I don't know what the proper solution would look like. For now, I can just cherry-pick this.

bjacob:
Trying https://github.com/openxla/stablehlo/pull/2575 as a simpler alternative to this, if that works.

bjacob:
Superseded by #2575.

PR #2575, opened 2024-10-03
CMake: Do not specify `LLVMSupport` in `LINK_LIBS`.

bjacob:
This is a new take on https://github.com/openxla/stablehlo/pull/2573.

There, I was trying to fix CMake errors in the external build, with the error message saying that `LINK_COMPONENTS` should be used instead of `LINK_LIBS` for  LLVM libraries.  That change fixed my external build, but broke the standalone build on CI.

Reading the `AddMLIR.cmake` code generating the error, I came across this line:
https://github.com/llvm/llvm-project/blob/ee4dd147baff8f971f3ec5aad5a216ca9837a732/mlir/cmake/modules/AddMLIR.cmake#L287-L288

```cmake
  # MLIR libraries uniformly depend on LLVMSupport.  Just specify it once here.
  list(APPEND ARG_LINK_COMPONENTS Support)
```

This looks like hardcoding always depending on `LLVMSupport`  (though I wasn't quite sure about the nuance between `Support` and `LLVMSupport`). If that's correct, then we never needed specifying `LLVMSupport` in the first place. So I tried just omitting it, and that seems to work. WDYT?

### DIFF
--- a/stablehlo/integrations/c/CMakeLists.txt
+++ b/stablehlo/integrations/c/CMakeLists.txt
@@ -41,7 +41,7 @@ add_mlir_public_c_api_library(StablehloCAPI
   LINK_LIBS PUBLIC
   MLIRCAPIIR
   MLIRIR
-  MLIRSupport
+  MLIRSupport  # LLVMSupport is automatically added in AddLLVM.cmake.
   StablehloOps
   StablehloPasses
   StablehloPortableApi
--- a/stablehlo/integrations/c/CMakeLists.txt
+++ b/stablehlo/integrations/c/CMakeLists.txt
@@ -28,7 +28,6 @@ add_mlir_public_c_api_library(ChloCAPI
 
   LINK_LIBS PUBLIC
   ChloOps
-  LLVMSupport
 )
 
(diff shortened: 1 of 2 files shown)


bjacob:
@sdasgup3, @GleasonK : this supersedes #2573.

GleasonK:
I think I added LLVMSupport for bazel build dep parity as opposed to actually needing to specify it, so yes looks like we don't need it. LGTM!

GleasonK (approved):

stablehlo/integrations/c/CMakeLists.txt
```diff
@@ -28,7 +28,6 @@ add_mlir_public_c_api_library(ChloCAPI
 
   LINK_LIBS PUBLIC
   ChloOps
-  LLVMSupport
```
GleasonK, line 31: Can we add a comment `# LLVMSupport def handled in top level CMake`

Here and in other file. Otherwise I think there's a decent chance it gets added back

bjacob: done!

bjacob:
@GleasonK , please merge :-)

Issue #2578, opened 2024-10-03
get_version_from_compatibility_requirement(WEEK_4) shows 1.5.0


apivovarov:
### What happened?

Issue:
`get_version_from_compatibility_requirement(WEEK_4)` returns 1.5.0

According to [VhloDialect.td](https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/VhloDialect.td) the versions were released on the following dates:

```
1.5.0 Jul 29 (65 days ago)
1.6.0 Aug 14 (49 days ago)
1.7.0 Sep 4 (28 days ago)
```

I expect `get_version_from_compatibility_requirement(WEEK_4)`  to show 1.7.0 or at least 1.6.0, but not 1.5.0

Why it is a problem?

JAX f8e4m3 and f8e3m4 [PR-23585](https://github.com/jax-ml/jax/pull/23585) tests fail because 

```
>     module_serialized = xla_client._xla.mlir.serialize_portable_artifact(  # type: ignore
          mlir_str, target_version)
E     jaxlib.xla_extension.XlaRuntimeError: INVALID_ARGUMENT: Failed to serialize StableHLO;
E     
E     Detailed error from MLIR: <unknown>:0: error: failed to legalize operation 'vhlo.func_v1' that was explicitly marked illegal
E     <unknown>:0: note: see current operation: 


">} : (!vhlo.tensor_v1<!vhlo.bool_v1>, !vhlo.tensor_v1<!vhlo.i32_v1>) -> ()
E       %4 = "vhlo.call_v1"(%2, %arg2) <{callee = #vhlo.string_v1<"_wrapped_jax_export_main">}> : (!vhlo.tensor_v1<!vhlo.i32_v1>, !vhlo.tensor_v1<?x!vhlo.f8E3M4_v1>) -> !vhlo.tensor_v1<?x!vhlo.f8E3M4_v1>
E       "vhlo.return_v1"(%4) : (!vhlo.tensor_v1<?x!vhlo.f8E3M4_v1>) -> ()
E     }) : () -> ()
E     <unknown>:0: error: failed to convert VHLO to v1.5.0
```

### Steps to reproduce your issue

Code snippet
```py

from mlir.dialects import stablehlo
req = stablehlo.StablehloCompatibilityRequirement.WEEK_4
stablehlo.get_version_from_compatibility_requirement(req)

1.5.0
```

### Version information

d40285ef3db0687e3f1e2bb0d716d748485a9739

PR #2579, opened 2024-10-03
Bump 4w compat to v1.7.0 release Sept 05, 2024

abhigunj:
fixes https://github.com/openxla/stablehlo/issues/2578

It is safe to bump 4w compat to `1.6.0`.  It is more that 4 weeks since it's release date (Aug 16, 2024).


apivovarov:
Can you bump it to 1.7.0? it was released 29 days ago which is more that 4 weeks.

stablehlo/dialect/Version.cpp
```diff
@@ -80,7 +80,7 @@ Version Version::fromCompatibilityRequirement(
     case CompatibilityRequirement::NONE:
       return Version::getCurrentVersion();
     case CompatibilityRequirement::WEEK_4:
-      return Version(1, 5, 0);  // v1.5.0 - Aug 1, 2024
+      return Version(1, 6, 0);  // v1.6.0 - Aug 16, 2024
```
GleasonK, line 83: Above this switch statement, can you add a comment:

```
// The time frames used are from the date that the release was tagged on, not merged.
// The tag date is when the version has been verified and exported to XLA.
// See: https://github.com/openxla/stablehlo/tags
```

GleasonK: Also bump to 1.7.0 as mentioned, that will allow new fp types to be used from JAX export, that tag date is now 4w ago.

apivovarov:
Thank you for updating it so quickly!

PR #2583, opened 2024-10-08
Fix BroadcastInDimOps linalg lowering with compatible dim=1

FruitClover:
For the case of simple broadcast check that we don't have to extend dimentions, so skip convertion to BroadcastOp when only one of them is 1.

Error before:

    test.mlir:5:8: error: 'linalg.broadcast' op input dim 1 should match init dim 2. input: 1, init: 5
      %0 = stablehlo.broadcast_in_dim %arg, dims = [1, 2, 3] : (tensor<3x1x1xf32>) -> tensor<1x3x5x7xf32>
           ^
    test.mlir:5:8: note: see current operation:
    %1 = "linalg.broadcast"(%arg0, %0) <{dimensions = array<i64: 0>}> ({
    ^bb0(%arg1: f32, %arg2: f32):
      "linalg.yield"(%arg1) : (f32) -> ()
    }) : (tensor<3x1x1xf32>, tensor<1x3x5x7xf32>) -> tensor<1x3x5x7xf32>

### DIFF
--- a/stablehlo/dialect/StablehloOps.cpp
+++ b/stablehlo/dialect/StablehloOps.cpp
@@ -1517,11 +1517,51 @@ DenseI64ArrayAttr getBroadcastDimensionsFromBroadcastSizes(
       getBroadcastDimensionsFromBroadcast(broadcastSizesSize, operandRank));
 }
 
+namespace {
+
+// Check that broadcast dimensions are suitable for isSimpleBroadcast():
+// extending rank is OK for them, but for dims where rank is not extended the
+// dim sizes must match.
+//
+// Two dimensions are compatible when:
+// - they are equal, or
+// - one of them is 1.
+// and here we reject the case when only one of them is 1.
+//
+// Examples of compatible dimensions for simple broadcast:
+// - tensor<3xf32> -> tensor<1x2x3xf32>
+// - tensor<1x1xf32> -> tensor<3x1x1xf32>
+// - tensor<5x7xf32> -> tensor<1x3x5x7xf32>
+// Examples of non-compatible dimensions:
+// - tensor<3xf32> -> tensor<3x3xf32>
+// - tensor<3xf32> -> tensor<1x3x3xf32>
+// - tensor<1x1xf32> -> tensor<3x2x2xf32>
+// - tensor<3x1x1xf32> -> tensor<1x3x5x7xf32>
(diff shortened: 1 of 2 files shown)


GleasonK:
To clarify the difference between simple broadcast / not simple broadcast, it's that extending rank is OK, but for dims where rank is not extended the dim sizes must match?

```
simple: tensor<3xf32> -> tensor<1x2x3xf32>
not simple: tensor<3x1xf32> -> tensor<1x3x3xf32>
```

stablehlo/dialect/StablehloOps.cpp
```diff
@@ -1517,11 +1517,36 @@ DenseI64ArrayAttr getBroadcastDimensionsFromBroadcastSizes(
...
+                                    RankedTensorType result) {
+  auto operandTy = cast<ShapedType>(operand);
+  auto resultTy = cast<ShapedType>(result);
+  ArrayRef<int64_t> operandShape = operandTy.getShape();
+  ArrayRef<int64_t> resultShape = resultTy.getShape();
+  // Two dimensions are compatible when:
+  // - they are equal, or
+  // - one of them is 1.
+  // For the simple broadcast check that we don't have to extend dimentions -
+  // reject the case when only one of them is 1.
```
GleasonK, line 1531: Can we move this above the function, this will help understand what "simple compatible dimensions" are, and give an example of "simple compatible" and "simple not compatible" dimensions to make this easier to understand for future reviewers/contributors

FruitClover:
> To clarify the difference between simple broadcast / not simple broadcast, it's that extending rank is OK, but for dims where rank is not extended the dim sizes must match?
> 
> ```
> simple: tensor<3xf32> -> tensor<1x2x3xf32>
> not simple: tensor<3x1xf32> -> tensor<1x3x3xf32>
> ```

basically yes, as i understood from from `BroadcastOp` semantics

stablehlo/dialect/StablehloOps.cpp
FruitClover: PTAL

GleasonK (approved):

stablehlo/dialect/StablehloOps.cpp
GleasonK: LGTM! thanks! There's more intricacies to broadcast op than I'd thought.

PR #2585, opened 2024-10-10
Automatically update WEEK_4 WEEK_12 versions

apivovarov:
This PR adds `update_version_h_cpp.sh` script

The script will:
- Automatically calculate `WEEK_4` and `WEEK_12` versions based on Version tag date and current date.
- Bump CurrentVersion by increasing patch version by 1 

This script should be executed at the beginning of the next development cycle, after performing the git tagging.

This script:
1. Updates current version in `stablehlo/dialect/Version.h`

```cpp 
  static Version getCurrentVersion() { return Version(x, y, z); }
```
New version will be (x, y, z + 1)

2. Replaces WEEK_4 and WEEK12 versions in stablehlo/dialect/Version.cpp
```
      return Version(a, b, c);  // WEEK_4 ANCHOR: DO NOT MODIFY
      return Version(m, n, p);  // WEEK_12 ANCHOR: DO NOT MODIFY
```
- WEEK_4 version - The most recent git tag that was created at least 28 days ago.
- WEEK_12 version - The most recent git tag that was created at least 84 days ago.

### Usage Example:

```bash
$ build_tools/update_version_h_cpp.sh

Next Current Version: 1, 8, 3
WEEK_4 Version: 1, 7, 6
WEEK_12 Version: 1, 5, 1
Saving /local/home/user/workspace/stablehlo/build_tools/../stablehlo/dialect/Version.h...
  static Version getCurrentVersion() { return Version(1, 8, 3); }
Saving /local/home/user/workspace/stablehlo/build_tools/../stablehlo/dialect/Version.cpp...
      return Version(1, 7, 6);  // WEEK_4 ANCHOR: DO NOT MODIFY
      return Version(1, 5, 1);  // WEEK_12 ANCHOR: DO NOT MODIFY
Done
```

#### Result:

```bash
$ git diff
diff --git a/stablehlo/dialect/Version.cpp b/stablehlo/dialect/Version.cpp
index b4ea0d21..d107ba57 100644
--- a/stablehlo/dialect/Version.cpp
+++ b/stablehlo/dialect/Version.cpp
@@ -83,9 +83,9 @@ Version Version::fromCompatibilityRequirement(
-      return Version(1, 7, 5);  // WEEK_4 ANCHOR: DO NOT MODIFY
+      return Version(1, 7, 6);  // WEEK_4 ANCHOR: DO NOT MODIFY
     case CompatibilityRequirement::WEEK_12:
-      return Version(1, 5, 0);  // WEEK_12 ANCHOR: DO NOT MODIFY
+      return Version(1, 5, 1);  // WEEK_12 ANCHOR: DO NOT MODIFY
     case CompatibilityRequirement::MAX:
       return Version::getMinimumVersion();
   }
diff --git a/stablehlo/dialect/Version.h b/stablehlo/dialect/Version.h
index 2ea74427..a6fea6b3 100644
--- a/stablehlo/dialect/Version.h
+++ b/stablehlo/dialect/Version.h
@@ -38,7 +38,7 @@ class Version {
   static FailureOr<Version> fromString(llvm::StringRef versionRef);
 
   /// Return a Version representing the current VHLO dialect version.
-  static Version getCurrentVersion() { return Version(1, 8, 2); }
+  static Version getCurrentVersion() { return Version(1, 8, 3); }
 
```

### DIFF
--- a/stablehlo/dialect/Version.cpp
+++ b/stablehlo/dialect/Version.cpp
@@ -83,9 +83,9 @@ Version Version::fromCompatibilityRequirement(
     case CompatibilityRequirement::NONE:
       return Version::getCurrentVersion();
     case CompatibilityRequirement::WEEK_4:
-      return Version(1, 7, 5);  // Sep 26, 2024
+      return Version(1, 7, 5);  // WEEK_4 ANCHOR: DO NOT MODIFY
     case CompatibilityRequirement::WEEK_12:
-      return Version(1, 5, 0);  // Aug 1, 2024
+      return Version(1, 5, 0);  // WEEK_12 ANCHOR: DO NOT MODIFY
     case CompatibilityRequirement::MAX:
       return Version::getMinimumVersion();
   }
--- a/build_tools/update_version_h_cpp.sh
+++ b/build_tools/update_version_h_cpp.sh
@@ -1,0 +1,124 @@
+#!/bin/bash
+# Copyright 2024 The StableHLO Authors.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script:
+# 1. Replaces current version in stablehlo/dialect/Version.h with the next
+# patch version
+#  static Version getCurrentVersion() { return Version(x, y, z); }
+# new version will be (x, y, z + 1)
+# 2. Replaces WEEK_4 and WEEK12 versions in stablehlo/dialect/Version.cpp
+#      return Version(a, b, c);  // WEEK_4 ANCHOR: DO NOT MODIFY
+#      return Version(m, n, p);  // WEEK_12 ANCHOR: DO NOT MODIFY
+# WEEK_4 version - The most recent git tag that was created at least 28 days ago.
+# WEEK_12 version - The most recent git tag that was created at least 84 days ago.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+script_dir="$(dirname "$(realpath "$0")")"
+version_h="$script_dir/../stablehlo/dialect/Version.h"
+version_cpp="$script_dir/../stablehlo/dialect/Version.cpp"
+
+fetch_current_version() {
+  # getCurrentVersion() { Version(X, Y, Z); }
+  ver_str=$(grep -A1 getCurrentVersion "$version_h" | grep -o 'Version(.*[0-9])')
+  REGEX="Version\(([0-9]+), ([0-9]+), ([0-9]+)\)"
+  if [[ $ver_str =~ $REGEX ]]; then
+    curr_ver=("${BASH_REMATCH[1]}" "${BASH_REMATCH[2]}" "${BASH_REMATCH[3]}")
+  else
+    echo "Error: Could not find current version string in $version_h" >&2
+    exit 1
+  fi
+}
+fetch_current_version
+
+# calculate next current ver (patch ver + 1)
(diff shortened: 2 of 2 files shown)


stablehlo/dialect/Version.cpp
```diff
@@ -83,9 +133,9 @@ Version Version::fromCompatibilityRequirement(
     case CompatibilityRequirement::NONE:
       return Version::getCurrentVersion();
     case CompatibilityRequirement::WEEK_4:
-      return Version(1, 7, 0);  // v1.7.0 - Sept 05, 2024
+      return getVersionTaggedOnOrAfterDays(28);
```
GleasonK, line 136: How about hardcoding `4 * 7` and `12 * 7`. Compiler would hopefully take care of the folding and it's a bit more readable.

apivovarov:
For convenience we can use the following script to get WEEK_4 and WEEK_12 tag / tag date / days ago

```bash
python3 get_compatibility_versions.py 

WEEK_4: (tag: v1.7.1) 2024-09-06T15:27:44-07:00 34 days ago
WEEK_12: (tag: v1.4.0) 2024-07-15T14:21:55-07:00 87 days ago
```

```py
import subprocess
import time

utc = int(time.time())

command = [
  "git",
  "log",
  "--tags",
  "--simplify-by-decoration",
  "--pretty=format:%ct,%d,%cI",
]
process = subprocess.Popen(
  command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True
)

WEEK_4_tag = ""
WEEK_12_tag = ""

for line in process.stdout:
  line = line.strip().split(",")
  tag_sec = int(line[0])
  diff_days = int((utc - tag_sec) / (60 * 60 * 24))
  if WEEK_4_tag == "" and diff_days >= 7 * 4:
    WEEK_4_tag = line[1].strip()
    print("WEEK_4:", WEEK_4_tag, line[2], diff_days, "days ago")
  if WEEK_12_tag == "" and diff_days >= 7 * 12:
    WEEK_12_tag = line[1].strip()
    print("WEEK_12:", WEEK_12_tag, line[2], diff_days, "days ago")
    break
```

Maybe add it to one of the documents?

apivovarov:
> Actually this is super useful! We can add this to our weekly integrate / version bump script. Tag the latest release, check if the values should be updated, bump the `Version.h` value.
> 
> Mind sending this PR as a `.sh` file in `buid_tools`? When we get a chance we'll integrate it into the process

Added `build_tools/update_version_h.sh`

See Usage Example in the PR description above

apivovarov:
Updated the script to use the following git command to get tag_ts and tag_name
```
git for-each-ref --sort=taggerdate --format '%(taggerdate:unix),%(refname:short)' refs/tags
```
This command retrieves the actual tagger date, which represents the date the tag was created, rather than the date of the last commit in the tag.

stablehlo/dialect/Version.h
```diff
@@ -40,6 +40,12 @@ class Version {
   /// Return a Version representing the current VHLO dialect version.
   static Version getCurrentVersion() { return Version(1, 7, 9); }
 
+  /// Return CompatibilityRequirement::WEEK_4 VHLO dialect version.
+  static Version getWeek4Version() { return Version(1, 7, 1); }
```
GleasonK, line 44: Can we make these functions in the Version.cpp file, or maybe even variables within the function itself?

I don't want these methods getting explicitly called anywhere, all references should go via the WEEK_4/WEEK_12 enums, so want to avoid exposing these functions publicly / keep the header file interface clean.

GleasonK: Otherwise all this LGTM, and once submitted, we'll build it into our oncall versioning rotation, so every PR which bumps version (i.e. https://github.com/openxla/stablehlo/pull/2610) will then also call this script.

apivovarov: Changes:
- Removed getWeek4Version and getWeek12Version from Version.h
- Added comments such as `// WEEK_4 ANCHOR: DO NOT MODIFY` to `return Version(a, b, c);` in Version.cpp
- The scripts uses these anchors to update week_4 and week_12 versions in Version.cpp now
- Added validation that new values in Version.h and Version.cpp are present. sed exits with exit code 0 if replacements were not done.

GleasonK (approved):

PR #2586, opened 2024-10-11
Revamp StableHLO simplification paterns.

GleasonK:
Starting to look into porting new simplification patterns from MHLO to StableHLO and wanted to clean things up in the process to make StableHLO simplification/folders an overall better place to innovate:

- Leverage DRR when possible. This is usually possible when the pattern isn't super complicated or the op being built doesn't have a region.
- Order simplification patterns by op alphabetically, order tests alphabetically, ensure a test per simplification pattern.
- Add code comments for each pattern of the format `// Pattern: ...` which can be used in the future for generating documentation.
- Move patterns which fold constants to folder pass.

### DIFF
--- a/stablehlo/transforms/StablehloAggressiveFolder.cpp
+++ b/stablehlo/transforms/StablehloAggressiveFolder.cpp
@@ -57,6 +57,9 @@ namespace stablehlo {
 
 namespace {
 
+// This is an upper limit on how many elements can be folded by an op folder.
+// This limit doesn't apply to some special cases like adding a zero,
+// multiplying by one, doing many operations with splats.
 constexpr int64_t kFoldOpEltLimit = 65536;
 
 // DenseElementsAttr can be constructed from ArrayRef<APInt> but not from
--- a/stablehlo/transforms/StablehloAggressiveFolder.cpp
+++ b/stablehlo/transforms/StablehloAggressiveFolder.cpp
@@ -14,6 +14,7 @@ limitations under the License.
 
 #include <cstddef>
 #include <cstdint>
+#include <numeric>
 #include <utility>
 
 #include "llvm/ADT/APInt.h"
@@ -56,6 +57,8 @@ namespace stablehlo {
 
 namespace {
 
+constexpr int64_t kFoldOpEltLimit = 65536;
+
 // DenseElementsAttr can be constructed from ArrayRef<APInt> but not from
 // ArrayRef<APSInt>. This helper bridges the gap.
 DenseIntElementsAttr getTensorAttr(ShapedType type, ArrayRef<APSInt> values) {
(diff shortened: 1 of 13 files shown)


stablehlo/transforms/StablehloAggressiveFolder.cpp
```diff
@@ -56,6 +57,8 @@ namespace stablehlo {
 
 namespace {
 
+constexpr int64_t kFoldOpEltLimit = 65536;
```
mvpant, line 60: Based on my modest experience, I would say that fewer computations typically result in better performance at run-time. Therefore, any operations that can be folded at compile time should be folded. Perhaps it would be sensible to implement this limit as a pass option?

sdasgup3, line 60: Can you please add the comment as well?
```
// This is an upper limit on how many elements canonicalization patterns are
// allowed to materialize as new constants.
```

GleasonK: I agree we need to offer _a lot_ of things as folder options, lets make this a future topic. Will make a GH issue.

I'm thinking of things like `FoldConstantSize` which may only allow us to fold splats to avoid expensive operations, or `FoldElementLimit` as mentioned here may be controversial - I have had to roll back changed internally since model canonicalization took 1min+ for folding / DCE.. I think a broader design on "optimization / folding control" is needed

sdasgup3: I believe the reason we have this limit is because  folding typically introduces new constants which can  contribute to an increase in model size which can be significant for large tensors.

stablehlo/transforms/StablehloAggressiveSimplificationPatterns.td
```diff
@@ -0,0 +1,281 @@
...
+//// Utilities
+def NotConstantOp : Constraint<
+    CPred<"llvm::isa<BlockArgument>($0) || !llvm::isa<stablehlo::ConstantOp>($0.getDefiningOp())">,
+    "is not a constant.">;
+
+def OperandsEqual : Constraint<CPred<"$0 == $1">, "operands are equal">;
+
+def TypesEqual : Constraint<CPred<"$0.getType() == $1.getType()">, "operands are equal">;
+
+def NumberOfElementsEqual : Constraint<
```
sdasgup3, line 25: Should we also consider the presence of dynamic dimensions here? Or maybe this simplification is applied after making sure of that dimensions are static.

GleasonK: Simplification is not run after dims are static, but for patterns where this matters, we should be already using `AnyStaticShapeTensor` such as:

```
// Pattern: subtract(X, X) -> 0
// Must be static shape, otherwise would require broadcasting via CHLO_ConstantLike
def : Pat<(StableHLO_SubtractOp AnyStaticShapeTensor:$operand, $operand),
          (StableHLO_ConstantLike<"0"> $operand)>;
```

stablehlo/transforms/StablehloAggressiveFolder.cpp
GleasonK: done

GleasonK: Created https://github.com/openxla/stablehlo/issues/2588

Resolving this thread in the meantime

sdasgup3 (approved):

PR #2592, opened 2024-10-15
[RFC] Add result accuracy to unary ops

hanrach9:
This is a proposal to add `result_accuracy` attribute to the following transcendental unary ops to StableHLO: `sin`, `cos`, `tan`, `tanh`, `sqrt`, `rsqrt`,
`cbrt`, `exp`, `expm1`, `log`, `logp1`, `logistic` and `erf`. Any feedback would be appreciated, thanks!


rfcs/20241015-result-accuracy.md
```diff
@@ -0,0 +1,74 @@
...
+By allowing users to select implementation based on accuracy per op, we can offer more tools to weigh tradeoffs between accuracy and performance and also ensure more consistent numerical behaviors across different devices.
+
+### Caveats on targets
+The proposed result_accuracy attribute can only be supported for targets that have multiple implementations of the ops.
+For example, for XLA-CPU, the implementations of these ops are dependent on LLVM and each target may have a different implementation with a different level of accuracy.
+Thus, further analysis needs to be done to support this feature on CPUs. This analysis could be performed while building the compiler, assuming the compiler will only be used on a single type of CPU.
+
+## Proposed Specification
+
+### `result_accuracy`
```
GleasonK, line 23: @jtristan I'm curious how you feel about speccing things as structs vs the mostly-inlined representation we've been using in StableHLO, as long as the grammar is trivially parsable to you have any preference?

I.e. I could see a spec for `result_accuracy<atol,trol,ulps,mode>` being done as:
1. Adding a new "type" to the spec called `ResultAccuracy` and to each of the proposed ops (`exp`, `expm1`, `log`, `logp1`, `logistic` and `tanh`) adding a single input `(I2) result_accuracy : ResultAccuracy` and the type has a constraint `(C1) if mode != TOLERANCE: atol = rtol = ulps = 0`, or
2. Speccing these as 4 flat arguments `(I2) atol, (I3) rtol, (I4) ulps, (I5) mode` on each of the ops, and adding a constraint to each op as well `(C1) if mode != TOLERANCE: atol = rtol = ulps = 0`

Option (1) feels more correct, but it's less standard in the current spec so curious if you have thoughts on any implications of doing things this way.

GleasonK: (cc @sdasgup3, @ghpvnist, @abhigunj)

ghpvnist: As long as `mode` is optional, I don't see any implications from the IR standpoint. I also prefer option 1 since it looks cleaner.

jtristan: I'm also favorable to option 1.

rfcs/20241015-result-accuracy.md
```diff
@@ -0,0 +1,101 @@
...
+
+The users can specify the worst case numerical error they can tolerate in terms
+of absolute, relative and ULP (unit in last place) errors. If they don't care
+about the numerical accuracy, they can also choose the implementation using
+`mode`. We propose a new attribute `result_accuracy`. `result_accuracy` can be
+any combination of the following numerical tolerances `atol`, `rtol`, `ulps` or
+an enum of `HIGHEST`, `DEFAULT` or `TOLERANCE`. `TOLERANCE` enum is a default
+placeholder value for `mode` when the numerical tolerances are used. When using
+the numerical tolerances, at least one of atol, rtol or ulps should be
+specified.
```
sdasgup3, line 46: It seems to me that mode is optional with default value `TOLERANCE` which necessitates providing values for atol, rtol and ulps. IMO it seems it might be confusing to the user to call something else as default when we have a DEFAULT enum value. But anyways I do not want the proposal to block on this.  


Are the `DEFAULT` and `HIGHEST` enum value refers to the precision values defined in the existing spec.

If not, How about we 
1. remove the `TOLERANCE` value from the enum
2. The `DEFAULT` behavior is to specify the `atol`, `rtol`, and `ulps`, given that these values must tbe provided when mode != TOLERANCE.

 If yes, I have a quick question: is there a reason we do not use HIGH?

hanrach9: Thanks for the comment. `DEFAULT` would return a fast but less accurate implementation and `HIGHEST` would give the most accurate implementation. We didn't use `HIGH` because right now we only have two implementations of the ops. Also, we added `TOLERANCE` because when `atol`, `rtol` and `ulps` are set, `TOLERANCE` will disambiguate it from the `DEFAULT` mode.

rfcs/20241015-result-accuracy.md
```diff
@@ -0,0 +1,102 @@
...
+
+Status: In Review<br/>
+Initial version: 10/15/2024<br/>
+Last updated: 10/15/2024<br/>
+Discussion thread:
+
+## Overview
+
+This RFC proposes adding a new attribute `result_accuracy` to the following
+transcendental unary ops: `exp`, `expm1`, `log`, `logp1`, `logistic` and `tanh`.
```
sdasgup3, line 11: Is there a reason sqrt and rsqrt is not in the list?

rfcs/20241015-result-accuracy.md
sdasgup3: Thanks for the clarification. 
The proposal lgtm with a minor comment.

sdasgup3 (approved):

rfcs/20241015-result-accuracy.md
hanrach9: Yeah, it should be should be supported for sqrt, rsqrt, cos, sin and recp too. Will update the doc.

GleasonK: What is `recp`? I don't think we have an op for that

ghpvnist: FYI `atan2` and `tan` are the two remaining transcendental ops not in this list.

hanrach9: recp is divide. So the complete list will be sin, cos, tan, divide, sqrt, rqrt, cbrt, divide, exp, expm1, log, logp1, logistic and erf. Is erf supported?

GleasonK: This would make divide the only binary op on this list.. which feels strange. Is this right that this proposal should be extended in scope to binary ops?

rfcs/20241015-result-accuracy.md
```diff
@@ -0,0 +1,103 @@
...
+The numerical tolerances will be compared against the compiler errors according
+to the following inequality:
+
+`abs(expected(x)-actual(x)) <= max(abs(expected(x))*max(rtol, ulps*epsilon),
+atol)` for all x, where `epsilon` is the machine epsilon.
+
+The inequality will be checked against the errors of each implementation and the
+one that can satisfy the constraint will be returned. If multiple
+implementations satisfy the inequality, the faster implementation will be used.
+If none of the implementations can meet the requested tolerance, the compiler
```
changm, line 91: Naive question - Does this mean it's a runtime check that's inserted by the compiler?

Or there's some upfront guarantee / check that's promised by underlying implementations? My interpretation of this sentence makes it sound like the compiler has to first run the operation, then check that the result accuracy was correct and return if error.

hanrach9: No runtime checks, the compiler will look check the request against its implementation errors at compile time.

changm: Ack mind making it explicit that it would be a compile time error?

hanrach9: Thanks, added that it's a compile time error.

rfcs/20241015-result-accuracy.md
hanrach9: We're able to support accuracy selection for reciprocal, but stablehlo doesn't have a reciprocal op. Is adding a reciprocal op a possible option?

dimvar:
* Is DEFAULT the same as the current accuracy for these ops?
* Can you give some information on what motivated this RFC? E.g., links to recent bugs, or something else.

hanrach9:
Replying to @dimvar, 
> * Is DEFAULT the same as the current accuracy for these ops?

Yes.
> * Can you give some information on what motivated this RFC? E.g., links to recent bugs, or something else.

We have higher accuracy software versions of these functions when the hardware accuracy isn't good enough, and we want to enable the accurate versions from Jax. Also it's been difficult for Jax users to reconcile inconsistent numerical behaviors when they depend on these functions across devices. This lets you express your requirements to the compiler in a device-independent fashion.

rfcs/20241015-result-accuracy.md
hanrach9: I left out divide for now.

GleasonK:
Looks like all open comments are resolved - LGTM! Also thanks for the patience, we've has several OpenXLA events in the past month. Merging, can send a PR for the impl any time

PR #2594, opened 2024-10-16
Set `/bigobj` option on MSVC to fix object limit error.

ScottTodd:
This fixes downstream build errors tracked at https://github.com/iree-org/iree/issues/18785.

Sample error message when building on Windows with MSVC:
```
FAILED: llvm-external-projects/stablehlo/stablehlo/dialect/CMakeFiles/obj.VhloOps.dir/VhloOps.cpp.obj
C:\ProgramData\Chocolatey\bin\ccache C:\PROGRA~1\MICROS~2\2022\ENTERP~1\VC\Tools\MSVC\1441~1.341\bin\Hostx64\x64\cl.exe  /nologo /TP -DGTEST_HAS_RTTI=0 -D_HAS_EXCEPTIONS=0 -D_USE_MATH_DEFINES -ID:\a\iree\iree\third_party\llvm-project\llvm\include -ID:\a\iree\iree\build-windows\llvm-project\include -ID:\a\iree\iree\third_party\llvm-project\mlir\include -ID:\a\iree\iree\third_party\stablehlo -ID:\a\iree\iree\build-windows\llvm-external-projects\stablehlo -ID:\a\iree\iree\third_party\llvm-project\lld\include -ID:\a\iree\iree\build-windows\llvm-project\tools\lld\include -external:I\..\mlir\include -external:ID:\a\iree\iree\build-windows\llvm-project\tools\mlir\include -external:W0 /DWIN32 /D_WINDOWS /EHsc /Z7 /O2 /Ob1  -std:c++17 -MD  /EHs-c- /GR- /showIncludes /Follvm-external-projects\stablehlo\stablehlo\dialect\CMakeFiles\obj.VhloOps.dir\VhloOps.cpp.obj /Fdllvm-external-projects\stablehlo\stablehlo\dialect\CMakeFiles\obj.VhloOps.dir\ /FS -c D:\a\iree\iree\third_party\stablehlo\stablehlo\dialect\VhloOps.cpp
cl : Command line warning D9025 : overriding '/EHs' with '/EHs-'
cl : Command line warning D9025 : overriding '/EHc' with '/EHc-'
D:\a\iree\iree\third_party\stablehlo\stablehlo\dialect\VhloOps.cpp : fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj
```

I think our downstream Windows builds worked with https://github.com/openxla/stablehlo/pull/2573 but they do not work with https://github.com/openxla/stablehlo/pull/2575, without this change. Hooray for compiler options whack-a-mole.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,6 +135,10 @@ endif()
 if(MSVC)
   # Required to use M_PI. Must be set before any `#include <cmath>`.
   add_compile_definitions("_USE_MATH_DEFINES")
+
+  # Tablegen outputs from LLVM/MLIR can easily exceed the default 64k section
+  # count limit, so increase to 2^32.
+  add_compile_options("/bigobj")
 endif()
 
 #-------------------------------------------------------------------------------


CMakeLists.txt
```diff
@@ -135,6 +135,10 @@ endif()
 if(MSVC)
   # Required to use M_PI. Must be set before any `#include <cmath>`.
   add_compile_definitions("_USE_MATH_DEFINES")
+
+  # Tablegen outputs from LLVM/MLIR can easily exceed the default 64k section
+  # count limit, so increase to 2^32.
+  add_compile_options("/bigobj")
```
ScottTodd, line 141: This is also set in LLVM: https://github.com/llvm/llvm-project/blob/92ad0397f494a9895385057586cc59a908107f81/llvm/cmake/modules/HandleLLVMOptions.cmake#L666-L669 and in IREE: https://github.com/iree-org/iree/blob/c6056d197230161ea1403e88b5b8784d34e071a2/build_tools/cmake/iree_copts.cmake#L131-L136

GleasonK: Which build type gets used here? EXTERNAL or EMBEDDED? The embedded actually uses `HandleLLVMOptions` so would get this for free?

ScottTodd: External: https://github.com/iree-org/iree/blob/c6056d197230161ea1403e88b5b8784d34e071a2/CMakeLists.txt#L820-L825

ScottTodd:
cc @bjacob 

> I think our downstream Windows builds worked with #2573 but they do not work with #2575, without this change. Hooray for compiler options whack-a-mole.

CMakeLists.txt
GleasonK: Got it, maybe we play whack-a-mole this time, and if something like this comes up again we try and see if there's a better way of including `include(HandleLLVMOptions)`.

The TF/mlir-hlo build uses external as well and has already called `include(HandleLLVMOptions)` outside of StableHLO, which likely sets this flag, and calling it twice causes issues. I wonder if theres a good way to check if this has been called previously..

GleasonK (approved):

CMakeLists.txt
GleasonK: (resolve thread if sgty and I'll merge)

ScottTodd: I don't really have a strong opinion right now. The layering of these LLVM projects is definitely tricky.

In IREE we don't call `include(HandleLLVMOptions)`. We use https://github.com/iree-org/iree/blob/main/build_tools/cmake/iree_llvm.cmake and https://github.com/iree-org/iree/blob/main/build_tools/cmake/iree_copts.cmake.

What we might end up doing with StableHLO is maintaining a parallel CMake build, as we already do for torch-mlir: https://github.com/iree-org/iree/blob/main/compiler/plugins/input/Torch/torch-mlir/CMakeLists.txt. We don't support Bazel for torch-mlir though, so the maintenance burden is a bit lower. With our own build, we could also address https://github.com/openxla/stablehlo/issues/2255 ourselves.

PR #2598, opened 2024-10-22
Fix issue when targeting a patch version for serialization

GleasonK:
We want to allow targeting patch version because we allow `-target=current` so it makes sense that current can resolve to a patch version.

That said we write our compat passes against patch v0 since all opset changes require a minor version bump and therefore have patch v0.

This PR implements "version flooring" inside the vhlo-to-version pass.


abhigunj (approved):

sdasgup3 (approved):

stablehlo/tests/vhlo/vhlo_to_version_downgrade_patch.mlir
```diff
@@ -0,0 +1,15 @@
...
+//
+// This will catch issues in op `isLegal` checks:
+//   op.minVersion() <= target <= op.maxVersion()
+
+// CHECK-LABEL: vhlo.func_v1 @all_to_all
+func.func public @all_to_all(%arg0: tensor<8x8x1xui16>) -> tensor<1x8x8xui16> {
+  // CHECK: vhlo.all_to_all_v1
+  %0 = "stablehlo.all_to_all"(%arg0) <{concat_dimension = 2 : i64, replica_groups = dense<[[0, 1, 2, 3, 4, 5, 6, 7]]> : tensor<1x8xi64>, split_count = 8 : i64, split_dimension = 0 : i64}> : (tensor<8x8x1xui16>) -> tensor<1x8x8xui16>
+  return %0 : tensor<1x8x8xui16>
+}diff --ruN a/stablehlo/stablehlo/transforms/VhloToVersion.cpp b/stablehlo/stablehlo/transforms/VhloToVersion.cpp
```
apivovarov, line 15: diff --ruN ???

GleasonK: good catch, fixed. File didn't end in a new line when I was trying to patch and looks like it failed

PR #2609, opened 2024-10-30
[legalize-linalg] Simple constant propagation for elementwise operands

srcarroll:
For elementwise operations, infer splat constant operands to avoid creating corresponding operands in the converted linalg op. Moreover, to facilitate this, the rewrite patterns `PointwiseToLinalgMapConverter` and `PointwiseToLinalgConverter` have been refactored to reuse common logic.


srcarroll:
I actually think this sort of thing should be a general linalg canonicalization in upstream, but as far as I know it is surprisingly not. I will probably make contribute a change for that but the change here is simple and non-invasive enough that I figured might as well.  Happy to hear thoughts

GleasonK (approved):

PR #2621, opened 2024-11-11
Bump WEEK_12 to v1.6.0, WEEK_4 to 1.7.7

apivovarov:
```
WEEK_4: v1.7.7 2024-10-04 37 days ago
WEEK_12: v1.6.0 2024-08-16 87 days ago
```

### DIFF
--- a/stablehlo/dialect/Version.cpp
+++ b/stablehlo/dialect/Version.cpp
@@ -83,9 +83,9 @@ Version Version::fromCompatibilityRequirement(
     case CompatibilityRequirement::NONE:
       return Version::getCurrentVersion();
     case CompatibilityRequirement::WEEK_4:
-      return Version(1, 7, 5);  // WEEK_4 ANCHOR: DO NOT MODIFY
+      return Version(1, 7, 7);  // WEEK_4 ANCHOR: DO NOT MODIFY
     case CompatibilityRequirement::WEEK_12:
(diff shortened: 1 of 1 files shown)


stablehlo/dialect/Version.cpp
```diff
@@ -83,9 +83,9 @@ Version Version::fromCompatibilityRequirement(
     case CompatibilityRequirement::NONE:
       return Version::getCurrentVersion();
     case CompatibilityRequirement::WEEK_4:
-      return Version(1, 7, 5);  // WEEK_4 ANCHOR: DO NOT MODIFY
+      return Version(1, 7, 7);  // WEEK_4 ANCHOR: DO NOT MODIFY
```
abhigunj, line 86: NIT: comment "DO NOT MODIFY" can be removed?  we are, will be modifying the versions.

abhigunj (approved):

stablehlo/dialect/Version.cpp
apivovarov: Sorry for confusion
`ANCHOR: DO NOT MODIFY` - it means do not modify the comment (not the value)

`ANCHOR: DO NOT MODIFY` is used by script which will automatically bump the versions in future - https://github.com/openxla/stablehlo/blob/main/build_tools/update_version_h_cpp.sh

PR #2623, opened 2024-11-12
Add square as a StableHLO operation

pearu:
As in the title.

Providing `square` as a StableHLO operation is a step forward to resolving https://github.com/jax-ml/jax/issues/24787 that reports inaccuracy problems in `jax.numpy.square` for complex inputs with large real and imaginary parts. After landing this PR, the next step is exposing `stablehlo.square` to JAX via XLA.

The details of `square` inaccuracies is given in https://github.com/pearu/functional_algorithms/issues/52 .  Roughly speaking, the source of inaccuracies in `square(z).real` is the usage of `z.real ** 2 - z.imag ** 2` that terms will overflow when `abs(z.real/imag) > sqrt(largest)`. The overflow does not happen when using `(z.real - z.imag) * (z.real + imag)` formula instead.

As a side-effect to using the 0.11 version of `function_algorithms`, several mlir test scripts for asin/asinh/acos/... functions are updated as well.


pearu:
Marked as draft because the `square` ought to be exposed as CHLO operation.
Meantime, https://github.com/jax-ml/jax/pull/24874 resolves the issue https://github.com/jax-ml/jax/issues/24787 for JAX as temporary workaround.

GleasonK:
Resolved in https://github.com/openxla/stablehlo/pull/2625

PR #2629, opened 2024-11-18
Fuse batch normalization into convolution kernel

mvpant:
This introduces a simplification that merges the batch normalization inference operation with convolution weights (kernel). The key idea is that while the batch normalization parameters change during the training phase, but remain constant during inference. This means that the convolution kernel can be adjusted to incorporate the effects of batch normalization. This optimization is applied by default to the ResNet model in the ONNX framework.

I will provide more details about the test case and additional formulas later.

For now, I would like to know if there is interest in this?

### DIFF
--- a/stablehlo/transforms/StablehloAggressiveSimplification.cpp
+++ b/stablehlo/transforms/StablehloAggressiveSimplification.cpp
@@ -21,6 +21,7 @@
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/ADT/SmallVectorExtras.h"
 #include "llvm/Support/ErrorHandling.h"
+#include "mlir/Dialect/Utils/IndexingUtils.h"
 #include "mlir/IR/Attributes.h"
 #include "mlir/IR/Block.h"
 #include "mlir/IR/Builders.h"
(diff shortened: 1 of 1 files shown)


mvpant:
Regarding terminology, what is preferred in StableHLO for convolution rhs: `kernel` or `weight`?

ghpvnist:
I like `kernel` but both are equally well understood imo, so up to the code author :) Since this isn't affecting the spec, anything works!

mvpant:
> Is this to say - during training these values won't be `constant` ops, and this pattern won't apply, but during inference it will? This seems reasonable.

Yes, I assume that’s why there are several operations like `stablehlo.batch_norm_grad`, `stablehlo.batch_norm_inference`, and `stablehlo.batch_norm_training`. The `stablehlo.batch_norm_inference` is designed to be used during the inference phase, normalizing input data using the statistics computed during training.

GleasonK:
This fell off my radar a few weeks back - That all makes sense! Pattern LGTM if we can make the test file more targeted / shorter!

GleasonK:
> But it’s still too slow

I need to figure out why bazel builds are so much slower than cmake..this test only took a few seconds on cmake. At a bare minimum I'll figure out a way to tag tests as large and not run the bazel CI for them.

I didn't notice that this test was in `testdata`, that's totally fine to have more practical "exported from X model" tests there! It's actually probably best to have something roughly testing numerics in `testdata`.

I'm thinking about unit tests, i.e. `stablehlo/tests/transforms/stablehlo_aggressive_simplification.mlir`. A few things that test the structural changes from the comment, could use all ones and zeros / garbage data if that's feasible:

```
// X = conv(input, weight)
// Y = batch_norm_inference(X, ...)
// into ->
// X = conv(input, weight(new))
// Y = add(X, broadcast_in_dim(Bias(new)))
```

GleasonK:
Made the following PR which lets `testdata` tests to use the suffix `.large.mlir` to extend the timeout for the file

https://github.com/openxla/stablehlo/pull/2671

mvpant:
> I'm thinking about unit tests, i.e. `stablehlo/tests/transforms/stablehlo_aggressive_simplification.mlir`. A few things that test the structural changes from the comment, could use all ones and zeros / garbage data if that's feasible:
> 
> ```
> // X = conv(input, weight)
> // Y = batch_norm_inference(X, ...)
> // into ->
> // X = conv(input, weight(new))
> // Y = add(X, broadcast_in_dim(Bias(new)))
> ```

Yes, i agree that it should be fine to use dummy data as we interested in transformations.

> Made the following PR which lets testdata tests to use the suffix .large.mlir to extend the timeout for the file

Cool. I`ll try to finish up this pull request..

mvpant:
Don't understand why it takes so long to run a large test using bazel build. 
I have tested 3 combinations:
debug stablehlo + debug llvm = 3.55 min
relwithdebinfo stablehlo + debug llvm = 2.28 min
relwithdebinfo stablehlo + relwithdebinfo llvm = 12s

Could the Bazel configuration for [san-common] involve a sanitizer?

mvpant:
Reduced test size, the 'relwithdebinfo' build now executes it in ~4 seconds.

mvpant:
Any updates?

PR #2636, opened 2024-11-22
Add StableHLO complex log1p operation. Add pass stablehlo-complex-math-expander

pearu:
As in the title.

This PR introduces a new pass stablehlo-complex-math-expander that expands StableHLO complex functions in terms of StableHLO real functions. Currently, only StableHLO_Log1pOp on complex inputs is included in this expander, more will be added as follow-ups.

The provided complex `log1p` operation fixes the inaccuracy problems in jax.numpy.log1p for complex inputs `x+i*y` when `x` is close to `-0.5 * y * y` which triggers catastrophic cancellations when using straightforward definition of log1p on complex inputs. The current state of `jax.numpy.log1p` inaccuracies is given in https://github.com/pearu/functional_algorithms/issues/47 .

With this PR, the accuracy statistics of log1p is:
```
complex64:
ULP difference == 0 count is 961936
ULP difference == 1 count is 39644
ULP difference == 2 count is 455
ULP difference == 3 count is 0
ULP difference >= 4 count is 0

complex128:
ULP difference == 0 count is 988144
ULP difference == 1 count is 13891
ULP difference == 2 count is 0
ULP difference == 3 count is 0
ULP difference >= 4 count is 0
```


pearu:
@GleasonK , please review.

GleasonK:
Curious why we need a CHLO log1p, we already have a StableHLO log1p, perhaps this is a bug in the actual codegen of log1p?

GleasonK:
I think log1p is a libdevice call for GPU:
https://github.com/openxla/xla/blob/7e1735213c707353ec22a00e4cd198eaa96839c7/xla/service/gpu/elemental_ir_emitter.cc#L197

And here's the CPU impl:
https://github.com/openxla/xla/blob/7e1735213c707353ec22a00e4cd198eaa96839c7/xla/service/elemental_ir_emitter.cc#L2311

pearu:
> Curious why we need a CHLO log1p, we already have a StableHLO log1p, perhaps this is a bug in the actual codegen of log1p?

@GleasonK this PR introduces CHLO log1p on complex operands that is accurate on the whole complex plane (contrary to HLO log1p defined in xla/service/elemental_ir_emitter.cc). libdevice log1p is available only for real operands.

This PR introduces CHLO log1p on real operands so that JAX could use a single entry point `chlo.log1p` for both real and complex inputs. CHLO log1p on real operands is defined here simply as a wrapper to StableHLO log1p:
```
def : Pat<(CHLO_Log1pOp NonComplexElementType:$x),
  (StableHLO_Log1pOp $x)>;
```

GleasonK:
What are your thoughts on doing this as an expander pass? We can make it so that all log1p's with complex data use that pass in XLA

pearu:
> What are your thoughts on doing this as an expander pass? I.e. no new CHLO, just a pass that expands complex stablehlo log1p's. We can make it so that all log1p's with complex types use that pass in XLA

Do I understand your correctly that your suggestion is to implement complex log1p fixes in XLA  xla/service/elemental_ir_emitter.cc? If so, it is certainly doable. If not, could you provide an existing function that uses the "expander pass" approach.

GleasonK:
My suggestion is more so to have a `createExpandLog1pPass` pass in the StableHLO repo which will decompose any `stablehlo.log1p`s with complex types. Then we can call this pass in XLA - it sounds like we can probably call this pass without knowing the backend if the types are complex.

stablehlo/transforms/StablehloComplexMathExpander.cpp
```diff
@@ -0,0 +1,76 @@
...
+  GreedyRewriteConfig config;
+};
+
+#include "stablehlo/transforms/StablehloComplexMathExpanderPatterns.h.inc"
+
+}  // namespace
+
+void populateStablehloComplexMathExpanderPatterns(RewritePatternSet *patterns,
+                                                  MLIRContext *context) {
+  patterns->add<Log1pOp_ComplexElementType_ComplexMathExpander>(context);
```
GleasonK, line 72: Lets instead use the `populateWithGenerated` which should be included by `#include "stablehlo/transforms/StablehloComplexMathExpanderPatterns.h.inc"`

Similar to https://github.com/openxla/stablehlo/blob/9e0c9e34d02af255bfe8f896ba7dc910758c6ecf/stablehlo/transforms/VhloToVersion.cpp#L405

This way new patterns in the TD file will be auto added to the pass

GleasonK (approved):

stablehlo/transforms/StablehloComplexMathExpander.cpp
pearu: Nice! Applied.

GleasonK (approved):

Issue #2633, opened 2024-11-21
Incorrect  `stablehlo.and|or` folder patterns in StablehloAggressiveSimplificationPatterns


christopherbate:
### What happened?

There are two incorrect folders introduced in the last major overhaul to the simplification transforms:

```

def IntOne : AttrConstraint<
   CPred<"::mlir::matchPattern($_self, m_One())">, "is integer one">;


def : Pat<(StableHLO_AndOp $lhs, (StableHLO_ConstantOp:$one IntOne:$value)),
          (replaceWithValue $lhs)>;

....

def : Pat<(StableHLO_OrOp $lhs, (StableHLO_ConstantOp:$one IntOne:$value)),
          (replaceWithValue $one)>;

```

constraint `IntOne` should be replaced with something along the lines of `IntAllOnes`

will submit fix & tests later tonight or tomorrow

PR #2638, opened 2024-11-22
[transforms] Fix simplification patterns for `stablehlo.(and|or)`

christopherbate:
Fixes an issue in `stablehlo-aggressive-simplification` where `%1` in
the below would get replaced by `%arg0`:

```
  %0 = stablehlo.constant dense<1> : tensor<2xi32>
  %1 = stablehlo.and %0, %arg0 : tensor<2xi32>
```

The pattern was checking whether `%0` is equal to `0b1` and was
only tested on bools. A similar bug existed for `stablehlo.and`. Fixed
by just making sure the constant is integer with all bits set to 1.

### DIFF
--- a/stablehlo/tests/transforms/stablehlo_aggressive_simplification.mlir
+++ b/stablehlo/tests/transforms/stablehlo_aggressive_simplification.mlir
@@ -63,6 +63,25 @@ func.func @and_one(%arg0: tensor<2xi1>) -> tensor<2xi1> {
   return %1 : tensor<2xi1>
 }
 
+// CHECK-LABEL: @and_i32_one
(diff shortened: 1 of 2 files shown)


christopherbate:
Fixes #2633

sdasgup3 (approved):

christopherbate:
@sdasgup3 @GleasonK Can someone help initiate CI workflows and/or merge the PR?

GleasonK (approved):
Thanks for the fix! Just getting back from OOO and catching up. Will take a look at your GH issue on the dynamic typing tomorrow shortly, should be able to tomorrow.

PR #2651, opened 2024-11-28
Bump patch version after integrate 1.8.4 -> 1.8.5

ghpvnist:


GleasonK (approved):

apivovarov:
I think we cam bump WEEK_12 to v1.7.0 - 2024-09-05 84 days ago (as of Nov 28)

GleasonK:
I'll take care of that bump by midweek, we're going to try and settle into a routine of ~2x/wk version bumps alongside integrates. If we need to expedite 1.7.0 we can, but in general building this into a process will make it easier

PR #2652, opened 2024-11-28
Fix integrate script to properly update version

ghpvnist:
The quoted string prevents the version from updating.


ghpvnist:
Hm they seem to be semantically the same. Maybe we can figure this out in the next integrate. Closing for now.

PR #2656, opened 2024-12-03
Integrate LLVM at llvm/llvm-project@bd92e4620433

abhigunj:


stablehlo/transforms/StablehloAggressiveFolder.cpp
```diff
@@ -79,8 +79,9 @@ APSInt getAPSInt(Type type, uint64_t value) {
   } else {
     llvm::report_fatal_error("expected integer type");
   }
-  return APSInt({/*numBits=*/numBits, value},
-                /*isUnsigned=*/isUnsigned);
+  return APSInt(
+      {/*numBits=*/numBits, value, /*isSigned=*/false, /*implicitTrunc=*/true},
```
ghpvnist, line 83: @sdasgup3 Should this be `/*isSigned=integerType.isSignedInteger()*/`?

sdasgup3: That is right. Also for StableHLO, `signless`  is also treated as signed, so how about we use `/*isSigned=*/!isUnSigned`.

sdasgup3 (approved):

stablehlo/transforms/StablehloAggressiveFolder.cpp
ghpvnist: Setting that to `!isUnsigned` appears to break `EvalCompareOpPattern`: https://github.com/tensorflow/mlir-hlo/commit/5eef965834ba46c70d74255b278a9bdd3df697a0

abhigunj: This change is back-port to gh from temporary.patch and unrelated to LLVM bump. To unblock the integrate I am temporarily marking this comment resolved and will reopen to discuss it further.

sdasgup3: Based on https://github.com/tensorflow/mlir-hlo/commit/5eef965834ba46c70d74255b278a9bdd3df697a0, lets keep the status quo.

abhigunj: keeping `/*isSigned=*/false` unchanged as `!isUnsigned` can't be used.

PR #2663, opened 2024-12-09
Workflow to run tutorials in CI

pavithraes:
This PR creates a workflow to execute the Jupyter notebooks in `docs/tutorials`. Couple of notes:
- It installs dependencies as defined in the notebook with `!pip install ...`, making it a little slow. We can instead create a `requirements.txt` and [cache dependencies](https://github.com/actions/setup-python#caching-packages-dependencies) for efficient runs, but then we won't be testing the actual user workflows running the tutorials on kaggle/colab
- The workflow only _executes_ the notebooks to make sure there are no errors. The notebooks are not altered. We can also update the notebooks in-place with the freshly executed versions for consistent output. The only concern here is the workflow might fail to push commits to the PR if it were opened from forks due to permission issues.

The PR also updates `savedmodel-embed.ipynb`; `target_version` is now a required argument in `stablehlo_to_tf_saved_model()`

Tested on my fork: https://github.com/pavithraes/stablehlo/actions/runs/12239259032/job/34139262358#step:5:1

### DIFF
--- a/.github/workflows/execute_tutorials.yml
+++ b/.github/workflows/execute_tutorials.yml
@@ -19,6 +19,9 @@ on:
     paths:
       - 'docs/tutorials/**.ipynb'
       - .github/workflows/execute_tutorials.yml
+  schedule:
+    # Run once a day
+    - cron:  '0 20 * * *'
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/execute_tutorials.yml
+++ b/.github/workflows/execute_tutorials.yml
@@ -1,0 +1,29 @@
+name: Execute Tutorial Notebooks
+
+on:
+  push:
+    paths:
+      - 'docs/tutorials/**.ipynb'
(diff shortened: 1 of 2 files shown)


.github/workflows/execute_tutorials.yml
```diff
@@ -0,0 +1,43 @@
...
+# limitations under the License.
+
+name: Execute Tutorial Notebooks
+
+on:
+  push:
+    paths:
+      - 'docs/tutorials/**.ipynb'
+      - .github/workflows/execute_tutorials.yml
+  workflow_dispatch:
```
GleasonK, line 22: Lets have a nightly as well:

```
on:
  # ...
  schedule:
    # Run once every 6hr building against llvm-project@HEAD
    - cron:  '0 20 * * *'
```

GleasonK (approved):
This is awesome!

GleasonK (approved):

PR #2683, opened 2025-01-06
Integrate LLVM at llvm/llvm-project@f739aa400416

abhigunj:


abhigunj:
https://github.com/openxla/stablehlo/pull/2686 has the latest bump + fixes for the upstream change.

PR #2687, opened 2025-01-10
Add PrintOp to the Interpreter dialect

ghpvnist:
While writing an MLIR pass, I realized there's no easy way to inspect the intermediate states of tensors while debugging them. ProbeOp does something similar, but this does not provide a human readable way of inspecting them given arbitrary SSA value.

### DIFF
--- a/stablehlo/reference/InterpreterOps.td
+++ b/stablehlo/reference/InterpreterOps.td
@@ -80,6 +80,29 @@ def Interpreter_RunParallelOp : Op<Interpreter_Dialect, "run_parallel", []> {
   let hasVerifier = 1;
 }
 
+def Interpreter_PrintOp : Op<Interpreter_Dialect, "print",
+    [SameOperandsAndResultType]> {
+  let summary = "Print operation";
+  let arguments = (ins
+    HLO_Tensor:$operand
+  );
+  let results = (outs HLO_Tensor:$result);
+  let description = [{
+    Print the value to stdout.
+
+    This is useful to print intermediate states of the tensors while debugging.
+    This should only be used to debug small tensors since every instance of this
+    op and its contents are printed to stdout. To gather information in bulk for
+    larger tensors, prefer using ProbeOp.
+
+    Example:
+    ```mlir
+    %result = interpreter.print %operand : tensor<i1>
+    ```
+  }];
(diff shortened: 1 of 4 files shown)


stablehlo/reference/InterpreterOps.td
```diff
@@ -80,6 +80,29 @@ def Interpreter_RunParallelOp : Op<Interpreter_Dialect, "run_parallel", []> {
   let hasVerifier = 1;
 }
 
+def Interpreter_PrintOp : Op<Interpreter_Dialect, "print",
+    [SameOperandsAndResultType]> {
+  let summary = "Print operation";
+  let arguments = (ins
+    HLO_Tensor:$operand
```
sdasgup3, line 87: Why do we need an `in` argument here? Is the reason to create a dummy daa-dep so that the op cannot be optimized out, then we can use some side-affect like  `MemWrite`.

sdasgup3 (commented):
This is just a cursory review. I will make a full pass soon.

stablehlo/reference/InterpreterOps.td
GleasonK: Let's avoid outputs? Model it after `print` functions in other programming languages

sdasgup3: ref: https://github.com/llvm/llvm-project/blob/d2498afccb04c0f09b05827b6b9c1c6c181a4f2b/mlir/include/mlir/Dialect/Vector/IR/VectorOps.td#L2613

ghpvnist: By default not providing any "side effects" it assumes it has one. +1 to removing output. I was originally using it to get the SSA name, but I can get that via the operand instead.

sdasgup3: Thank!
How about we have the intent cleaner by specifying the exact side affect (in cases we have that knowledge) enabling the compiler to avoiding pessimistic assumptions.

ghpvnist: I think it's fine the way it is currently -- the purpose of this op is to inspect a given module and it isn't designed to run with other passes.

sdasgup3 (approved):

PR #2697, opened 2025-01-24
Add source link to awesome.md.

ScottTodd:
I couldn't find an easy link back to this source file from https://openxla.org/stablehlo/awesome.

I _think_ this style of URL should be fine, but I might be missing some website detail (for example, the hosted website has support for translations).

A site-wide solution may be available for some source files. For example:
* The mkdocs-material site generator has "edit this page" and "view source of this page" actions: https://squidfunk.github.io/mkdocs-material/setup/adding-a-git-repository/#code-actions.
* Some tensorflow examples include "View source on GitHub" buttons for notebooks: https://www.tensorflow.org/guide/keras/functional_api


GleasonK (approved):
Great addition, thanks!

PR #2698, opened 2025-01-24
Add IREE project link to awesome.md.

ScottTodd:
Happy to iterate on the specific phrasing, suggestions welcome.

I wonder if some reorganization would help too, since IREE could fit into either the "PJRT Plugins" or "Edge Compilation" sections.


GleasonK (approved):
This lgtm for wording, I'm going to ping other compiler projects that accept StableHLO as input and will likely form a separate "Compilers" section in this doc, but for now this is an OK spot.

Thanks for the addition!

PR #2702, opened 2025-01-29
Update tosa.mul op in StableHLO-to-TOSA Pass in compliance with MLIR upstream and TOSA-v1.0

bubblepipe:
This PR updates the StableHLO to TOSA legalization pass to align with the recent changes in the MLIR upstream. Specifically, [the `tosa::MulOp` operation has being modified to comply with the TOSA-v1.0 specification](https://github.com/llvm/llvm-project/pull/121953). The shift parameter of the MUL operator, which was previously an attribute, has been moved to an SSA operand. 

The upstream changes caused a compilation failure in the StableHLO to TOSA conversion pass because the `tosa.mul` operation now expects 3 operand groups instead of 2, as reflected in the error message:

```jsx
error: invalid number of operand groups for `tosa.mul`; expected 3, but got 2
     with op<tosa.mul>(input0, input1) {shift = attr<"0 : i8">};
          ^
```

This PR added a zero constant as the additional argument for the shift parameter to maintain compatibility with the updated `tosa.mul` operation.


sdasgup3 (approved):

GleasonK:
Thank you for the fix! We are still integrating these TOSA / LLVM changes throughout the rest of the OpenXLA projects, we will merge this change as a part of our next LLVM bump and ensure you are attributed as author of the fix

PR #2707, opened 2025-02-04
Add target independent optimization pass

GleasonK:


stablehlo/transforms/optimization/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
+# Copyright 2022 The StableHLO Authors.
```
sdasgup3, line 1: 2025?

stablehlo/transforms/optimization/Passes.td
```diff
@@ -0,0 +1,45 @@
...
+           "Allow for potentially lossy computations using float type.">,
+  ];
+}
+
+def StablehloAggressiveSimplificationPass
+    : Pass<"stablehlo-aggressive-simplification", "func::FuncOp"> {
+  let summary = "Canonicalizes StableHLO operations";
+  let dependentDialects = [
+    "mlir::stablehlo::StablehloDialect",
+    "mlir::arith::ArithDialect",
```
sdasgup3, line 35: [no action] seems to be missing in the original td file.

sdasgup3 (approved):

GleasonK:
Actually I'm not sure if we want this standalone directory. I may rework this. Our passes have some dependencies between eachother so it may make sense to keep them together...

stablehlo/transforms/optimization/StablehloTargetIndependentOptimization.cpp
```diff
@@ -0,0 +1,66 @@
...
+    : public impl::StablehloTargetIndependentOptimizationPassBase<
+          StablehloTargetIndependentOptimizationPass> {
+  using StablehloTargetIndependentOptimizationPassBase::
+      StablehloTargetIndependentOptimizationPassBase;
+
+  LogicalResult initialize(MLIRContext* context) override {
+    RewritePatternSet patterns_(context);
+    bool foldFloat = false;
+    populateStablehloCanonicalizationPatterns(context, &patterns_);
+    populateStablehloAggressiveFolderPatterns(&patterns_, context, foldFloat,
```
sdasgup3, line 44: Are we missing any optimization if we keep the same benefits as 1 for both canonicalization and folding? If nit then how about we keep the benefit the same unless we find a case where we want canonicalization to happen before.

abhigunj (approved):

stablehlo/transforms/optimization/StablehloTargetIndependentOptimization.cpp
GleasonK: We probably always want folding to have higher precedence than canonicalization.

There exist simplification patterns that rely on folding being applied (I have come across MLIR patterns that relied on the precedence of folding being higher than simplification in LiteRT). 

Also much of canonicalization relies on things having constants as inputs that it can reason about it, so folding first enables more canonicalization.

stablehlo/transforms/optimization/StablehloTargetIndependentOptimization.cpp
```diff
@@ -0,0 +1,66 @@
+// Copyright 2023 The IREE Authors
+//
+// Licensed under the Apache License, Version 2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+// Implements optional canonicalization patterns for StableHLO ops.
```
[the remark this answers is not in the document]
GleasonK: removed

stablehlo/transforms/optimization/Passes.td
GleasonK: Indeed it was. I noticed this as I was porting things.

stablehlo/transforms/optimization/CMakeLists.txt
GleasonK: fixed, here and elsewhere in this dir

GleasonK:
> Actually I'm not sure if we want this standalone directory. I may rework this. Our passes have some dependencies between eachother so it may make sense to keep them together...

Ok we're good, I'd initially thought things were more entangled, but seems to be OK to split this out.

PR #2709, opened 2025-02-06
Use reference to tensor data when converting to dense elements attr

GleasonK:
Found that the enzyme project was doing something like this, but didn't support complex numbers (our method does). Making this function the best of both worlds then will update enzyme.

### DIFF
--- a/stablehlo/reference/Tensor.cpp
+++ b/stablehlo/reference/Tensor.cpp
@@ -576,7 +576,7 @@ FailureOr<uint32_t> getBitWidthIfReferenceCopyable(Type elementType) {
   uint32_t bitWidth = isa<FloatType>(elementType)
                           ? cast<FloatType>(elementType).getWidth()
                           : cast<IntegerType>(elementType).getWidth();
-  if (bitWidth != 1 && bitWidth % 8 == 0) return failure();
+  if (bitWidth % 8 == 0) return failure();
   return bitWidth;
 }
 
@@ -596,6 +596,7 @@ DenseElementsAttr makeDenseElementsAttr(Tensor tensor) {
   if (isa<FloatType>(elementType)) {
     // Explicitly copy sub-byte float values
     std::vector<llvm::APFloat> values;
+    values.reserve(tensor.getNumElements());
     for (auto it = tensor.index_begin(); it != tensor.index_end(); ++it) {
       Element element = tensor.get(*it);
       values.push_back(element.getFloatValue());
@@ -613,6 +614,7 @@ DenseElementsAttr makeDenseElementsAttr(Tensor tensor) {
       return DenseElementsAttr::get(type, data);
     }
     std::vector<llvm::APInt> values;
+    values.reserve(tensor.getNumElements());
     for (auto it = tensor.index_begin(); it != tensor.index_end(); ++it) {
       Element element = tensor.get(*it);
       values.push_back(element.getIntegerValue());
--- a/stablehlo/reference/Tensor.cpp
+++ b/stablehlo/reference/Tensor.cpp
@@ -567,34 +567,57 @@ Tensor makeTensor(DenseElementsAttr attr) {
       invalidArgument("Unsupported type: ", debugString(type).c_str()));
 }
 
+FailureOr<uint32_t> getBitWidthIfReferenceCopyable(Type elementType) {
+  // Can't copy complex types by reference.
+  if (!isa<FloatType>(elementType) || !isa<IntegerType>(elementType))
+    return failure();
+
+  // Can't copy bool or sub-byte types by reference.
+  uint32_t bitWidth = isa<FloatType>(elementType)
+                          ? cast<FloatType>(elementType).getWidth()
+                          : cast<IntegerType>(elementType).getWidth();
+  if (bitWidth != 1 && bitWidth % 8 == 0) return failure();
+  return bitWidth;
+}
+
 DenseElementsAttr makeDenseElementsAttr(Tensor tensor) {
   auto type = tensor.getType();
   auto elementType = type.getElementType();
 
-  if (auto floatType = dyn_cast<FloatType>(elementType)) {
-    // Build from tensor data ref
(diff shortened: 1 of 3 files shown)


stablehlo/reference/Tensor.cpp
```diff
@@ -571,25 +571,30 @@ DenseElementsAttr makeDenseElementsAttr(Tensor tensor) {
...
-        values.push_back(APInt(1, element.getBooleanValue() ? 1 : 0));
-      } else {
-        values.push_back(element.getIntegerValue());
-      }
-    }
-    return DenseIntElementsAttr::get(tensor.getType(), values);
+    // Build from tensor data ref
+    bitWidth = bitWidth / 8;
+    auto size = tensor.getNumElements() * bitWidth;
+    auto floatValues = ArrayRef(tensor.getData(), size);
```
sdasgup3, line 596: intValues

stablehlo/reference/Tensor.cpp
```diff
@@ -571,25 +571,30 @@ DenseElementsAttr makeDenseElementsAttr(Tensor tensor) {
...
 
-  if (isa<FloatType>(elementType)) {
-    std::vector<llvm::APFloat> values;
-    for (auto it = tensor.index_begin(); it != tensor.index_end(); ++it) {
-      Element element = tensor.get(*it);
-      values.push_back(element.getFloatValue());
+  if (auto floatType = dyn_cast<FloatType>(elementType)) {
+    // Build from tensor data ref
+    uint32_t bitWidth = floatType.getWidth();
+    bitWidth = bitWidth / 8;
```
sdasgup3, line 577: how about 
```
sizeInBytes = tensor.getNumElements() * floatType.getWidth() / 8;
```
(Just for clarity)

stablehlo/reference/Tensor.cpp
```diff
@@ -571,25 +571,30 @@ DenseElementsAttr makeDenseElementsAttr(Tensor tensor) {
...
-      Element element = tensor.get(*it);
-      if (isSupportedBooleanType(elementType)) {
-        values.push_back(APInt(1, element.getBooleanValue() ? 1 : 0));
-      } else {
-        values.push_back(element.getIntegerValue());
-      }
-    }
-    return DenseIntElementsAttr::get(tensor.getType(), values);
+    // Build from tensor data ref
+    bitWidth = bitWidth / 8;
```
sdasgup3, line 594: This may not work for sub 8-bit typed integers?

sdasgup3: same for floats like `Float4E2M1FNType` with < 8 bits.

GleasonK: Modified to use explicit copies as a fallback for sub byte values. Good catch. Added tests to make sure its working.

stablehlo/reference/Tensor.cpp
GleasonK: updates names

stablehlo/reference/Tensor.cpp
```diff
@@ -567,11 +567,34 @@ Tensor makeTensor(DenseElementsAttr attr) {
       invalidArgument("Unsupported type: ", debugString(type).c_str()));
 }
 
+FailureOr<uint32_t> getBitWidthIfReferenceCopyable(Type elementType) {
+  // Can't copy complex types by reference.
+  if (!isa<FloatType>(elementType) || !isa<IntegerType>(elementType))
+    return failure();
```
sdasgup3, line 573:
```   
 return failure("Only float and integer types are copyable by reference.");
```

stablehlo/reference/Tensor.cpp
```diff
@@ -567,11 +567,34 @@ Tensor makeTensor(DenseElementsAttr attr) {
...
+
+  // Can't copy bool or sub-byte types by reference.
+  uint32_t bitWidth = isa<FloatType>(elementType)
+                          ? cast<FloatType>(elementType).getWidth()
+                          : cast<IntegerType>(elementType).getWidth();
+  if (bitWidth != 1 && bitWidth % 8 == 0) return failure();
```
sdasgup3, line 579:
```
   return failure(
        "Bool and sub-byte types are not copyable by reference due to "
        "potential alignment issues.");
```

sdasgup3: we can just have the single change `bitWidth % 8 == 0` covering all sub-byte types (including bools)

stablehlo/reference/Tensor.cpp
```diff
@@ -580,14 +603,19 @@ DenseElementsAttr makeDenseElementsAttr(Tensor tensor) {
...
+    // Explicitly copy sub-byte or bool ints
+    if (isSupportedBooleanType(elementType)) {
+      SmallVector<bool, 1> data;
+      data.reserve(tensor.getNumElements());
+      auto v = tensor.getData();
+      for (size_t i = 0; i < static_cast<size_t>(tensor.getNumElements()); ++i)
+        data.push_back(v[i] ? 1 : 0);
+      return DenseElementsAttr::get(type, data);
+    }
     std::vector<llvm::APInt> values;
```
sdasgup3, line 615: Can you add an `values.reserve(tensor.getNumElements())` here?

stablehlo/reference/Tensor.cpp
GleasonK: Fixed the bitwidth check. The `failure` doesn't take args, and I don't think we want to emit any diagnostic since it triggers the fallback.

```
inline LogicalResult failure(bool IsFailure = true) {
  return LogicalResult::failure(IsFailure);
}
```

stablehlo/reference/Tensor.cpp
GleasonK: Accidentally replied in other comment thread below, failure doesn't take args, and we don't want to emit a diagnostic from this fn since there is still a fallback.

sdasgup3 (approved):

PR #2711, opened 2025-02-12
Fix duplicate symbols for downstream CAPIObjects users

cerisier:
Should fix failure in https://github.com/openxla/xla/pull/22438

I need just a bit more time to validate this entirely on Linux


GleasonK (approved):

GleasonK:
Patched your change and validated using XLA CI, going to merge this to keep things in sync

PR #2722, opened 2025-02-27
Pass to wrap StableHLO ops in composite

sdasgup3:
Wraps StableHLO operations in `stablehlo.composite` operations.

    For instance, consider a simple StableHLO program:

    ```mlir
      func.func @main(%arg0 : tensor<2xf32>, %arg1 : tensor<2xf32>) -> tensor<2xf32> {
        %0 = stablehlo.add %arg0, %arg1 : tensor<2xf32>
        return %0 : tensor<2xf32>
      }
    ```

    Applying this pass to wrap `stablehlo.add` operations will result in the
    following program:

    ```mlir
      func.func @main(%arg0: tensor<2xf32>, %arg1: tensor<2xf32>) -> tensor<2xf32> {
        %0 = stablehlo.composite "stablehlo.add" %arg0, %arg1 {decomposition = @stablehlo.add.impl} : (tensor<2xf32>, tensor<2xf32>) -> tensor<2xf32>
        return %0 : tensor<2xf32>
      }
      func.func private @stablehlo.add.impl(%arg0: tensor<2xf32>, %arg1: tensor<2xf32>) -> tensor<2xf32> {
        %0 = stablehlo.add %arg0, %arg1 : tensor<2xf32>
        return %0 : tensor<2xf32>
      }
    ```

    Notes:

      - The `name` attribute of the generated `stablehlo.composite` operation
        will always be the same as the name of the original operation that was
        wrapped (e.g., if you wrap a `stablehlo.add` operation, the composite
        will also be named `"stablehlo.add"`).
      - The private function that encapsulates the original operation
        (referenced by the `decomposition` attribute of the
        `stablehlo.composite` operation) will be named using the pattern
        `<op_name>.impl[.N]`, where `<op_name>` is the name of the original
        operation, and `N` is a unique integer identifier generated to prevent
        naming conflicts within the module.

    This pass can be used in three distinct ways:

    **Mode 1: Command-line Usage**

    This mode is the simplest, using the `stablehlo-opt` utility with the
    `op-names` (a comma-separated list of operation names) and `version` (an
    integer version number) options. It wraps **all instances** of specified
    operations. The attributes of the newly created `stablehlo.composite`
    operation will be the same as the attributes of the original operation.

    **Usage Example:**

    ```bash
    stablehlo-opt input.mlir --stablehlo-wrap-in-composite=op-names='stablehlo.add,stablehlo.mul' -o output.mlir
    ```

    **Mode 2: Programmatic Single-Op Wrapping**

    This mode provides programmatic control to wrap
    **a specific operation instance** and returns a pointer to the newly
    created `stablehlo.composite` operation.

    **Example (C++):**

    ```cpp
      // To wrap a specific stablehlo.add instance

      mlir::stablehlo::AddOp addOp = ...; // The op instanced to be wrapped.
      mlir::ModuleOp module = addOp->getParentOfType<mlir::ModuleOp>();
      mlir::OpBuilder builder(addOp);
      mlir::NamedAttrList attrs = ...; // Attributes to be set on the composite op.
      int32_t version = 0; // Composite version.

      mlir::stablehlo::CompositeOp compositeOp = mlir::stablehlo::wrapOperationInComposite(builder, addOp, attrs, version, module);
      addOp.replaceAllUsesWith(compositeOp);
    ```

    **Mode 3: Programmatic Module-Wide Wrapping with Attribute Predicates**

    This mode extends programmatic wrapping to the entire module, offering
    fine-grained control over which operations are wrapped and their attributes.
    This is achieved by using the `createStablehloWrapInCompositePass` API,
    which takes an `AttributePredicateMap` as an argument.

    The `AttributePredicateMap` is a map that dictates which operations should
    be considered for wrapping and how their attributes should be handled. Its
    semantics are as follows:

    - **Keys (mlir::TypeID):** `TypeID` of an MLIR operation. If an operation's
        `TypeID` matches a key in the map, it becomes a candidate for wrapping.
    - **Values (Lambda Functions):** Lambda function of type
        `std::function<std::optional<NamedAttrList>(Operation*)>`. This function
        is applied to each candidate operation.
        - **Input:** An `mlir::Operation*`, which is an instance of the
          operation type corresponding to the `TypeID` key.
        - **Return Value:** An `std::optional<NamedAttrList>`.
          - If the lambda returns a `NamedAttrList` (wrapped in
            `std::optional`), the operation is wrapped in a
            `stablehlo::composite` operation, and the returned attributes are
            used to set the composite's attributes.
          - If the lambda returns `std::nullopt`, the operation is **not**
            wrapped. This allows for selective wrapping based on custom
            criteria.

    **Example (C++):**

    ```cpp

    // ... inside a pass or function ...

    stablehlo::AttributePredicateMap attributePredicateMap;

    attributePredicateMap[mlir::TypeID::get<mlir::stablehlo::AddOp>()] =
      [](mlir::Operation* op) -> std::optional<mlir::NamedAttrList> {
      // Custom logic to determine if and how to wrap the operation.
      // Example: Only wrap if it's on a specific type.
      if (op->getOperand(0).getType().isa<mlir::Float32Type>()) {
        return mlir::NamedAttrList(op->getAttrs());
      }
      return std::nullopt; // Do not wrap.
    };

    pm.addPass(createStablehloWrapInCompositePass(attributePredicateMap, compositeVersion));
    if (mlir::failed(pm.run(module))) {
      return;
    }
    ```

### DIFF
--- a/docs/generated/stablehlo_passes.md
+++ b/docs/generated/stablehlo_passes.md
@@ -380,15 +380,16 @@ Notes:
     operation, and `N` is a unique integer identifier generated to prevent
     naming conflicts within the module.
 
-This pass can be used in three distinct ways:
+This pass can be used in two distinct ways:
 
 **Mode 1: Command-line Usage**
 
-This mode is the simplest, using the `stablehlo-opt` utility with the
-`op-names` (a comma-separated list of operation names) and `version` (an
-integer version number) options. It wraps **all instances** of specified
-operations. The attributes of the newly created `stablehlo.composite`
-operation will be the same as the attributes of the original operation.
+This mode is intended for debugging or testing, as it offers minimal control
+over the attributes of the generated `stablehlo.composite` operations.
+It wraps **all instances** of operations specified using the `op-names`
+(a comma-separated list of operation names) options. The attributes of the
+newly created `stablehlo.composite` operation will be the same as the
+attributes of the original operation.
 
 **Usage Example:**
 
@@ -396,28 +397,7 @@ operation will be the same as the attributes of the original operation.
 stablehlo-opt input.mlir --stablehlo-wrap-in-composite=op-names='stablehlo.add,stablehlo.mul' -o output.mlir
 ```
 
-**Mode 2: Programmatic Single-Op Wrapping**
-
-This mode provides programmatic control to wrap
-**a specific operation instance** and returns a pointer to the newly
-created `stablehlo.composite` operation.
-
-**Example (C++):**
-
-```cpp
-// To wrap a specific stablehlo.add instance
-
-mlir::stablehlo::AddOp addOp = ...; // The op instanced to be wrapped.
-mlir::ModuleOp module = addOp->getParentOfType<mlir::ModuleOp>();
-mlir::OpBuilder builder(addOp);
-mlir::NamedAttrList attrs = ...; // Attributes to be set on the composite op.
-int32_t version = 0; // Composite version.
-
-mlir::stablehlo::CompositeOp compositeOp = mlir::stablehlo::wrapOperationInComposite(builder, addOp, attrs, version, module);
-addOp.replaceAllUsesWith(compositeOp);
-```
-
-**Mode 3: Programmatic Module-Wide Wrapping with customized Attribute Handling**
+**Mode 2: Programmatic Module-Wide Wrapping with customized Attribute Handling**
 
 This mode extends programmatic wrapping to the entire module, offering
 fine-grained control over which operations are wrapped and their attributes.
@@ -447,8 +427,6 @@ handled. Its semantics are as follows:
 **Example (C++):**
 
 ```cpp
-
-// ... inside a pass or function ...
 
 stablehlo::CompositeAttributeProviderMap compositeAttributeProviderMap;
 
--- a/docs/generated/stablehlo_passes.md
+++ b/docs/generated/stablehlo_passes.md
@@ -301,8 +301,8 @@ _Refines shapes across a StableHLO program._
   The flagship use case for this pass is specializing dynamically-shaped
   programs to static shapes. If a dynamically-shaped StableHLO program has the
   right structure, then updating its argument types from dynamic shapes to
-  static shapes and running this pass will propagate static shapes across
-  the program.
+  static shapes and running this pass will propagate static shapes across the
+  program.
 
   This pass removes `custom_call @shape_refinement_operand_wrapper` by
   replacing uses of the result with the operand directly, and propagates
@@ -347,24 +347,24 @@ Wraps StableHLO operations in `stablehlo.composite` operations.
 For instance, consider a simple StableHLO program:
 
 ```mlir
-  func.func @main(%arg0 : tensor<2xf32>, %arg1 : tensor<2xf32>) -> tensor<2xf32> {
-    %0 = stablehlo.add %arg0, %arg1 : tensor<2xf32>
-    return %0 : tensor<2xf32>
-  }
+func.func @main(%arg0 : tensor<2xf32>, %arg1 : tensor<2xf32>) -> tensor<2xf32> {
+  %0 = stablehlo.add %arg0, %arg1 : tensor<2xf32>
+  return %0 : tensor<2xf32>
+}
 ```
 
 Applying this pass to wrap `stablehlo.add` operations will result in the
 following program:
 
 ```mlir
-  func.func @main(%arg0: tensor<2xf32>, %arg1: tensor<2xf32>) -> tensor<2xf32> {
-    %0 = stablehlo.composite "stablehlo.add" %arg0, %arg1 {decomposition = @stablehlo.add.impl} : (tensor<2xf32>, tensor<2xf32>) -> tensor<2xf32>
-    return %0 : tensor<2xf32>
-  }
-  func.func private @stablehlo.add.impl(%arg0: tensor<2xf32>, %arg1: tensor<2xf32>) -> tensor<2xf32> {
-    %0 = stablehlo.add %arg0, %arg1 : tensor<2xf32>
-    return %0 : tensor<2xf32>
-  }
+func.func @main(%arg0: tensor<2xf32>, %arg1: tensor<2xf32>) -> tensor<2xf32> {
+  %0 = stablehlo.composite "stablehlo.add" %arg0, %arg1 {decomposition = @stablehlo.add.impl} : (tensor<2xf32>, tensor<2xf32>) -> tensor<2xf32>
+  return %0 : tensor<2xf32>
+}
+func.func private @stablehlo.add.impl(%arg0: tensor<2xf32>, %arg1: tensor<2xf32>) -> tensor<2xf32> {
+  %0 = stablehlo.add %arg0, %arg1 : tensor<2xf32>
+  return %0 : tensor<2xf32>
+}
 ```
 
 Notes:
@@ -405,28 +405,28 @@ created `stablehlo.composite` operation.
 **Example (C++):**
 
 ```cpp
-  // To wrap a specific stablehlo.add instance
+// To wrap a specific stablehlo.add instance
 
-  mlir::stablehlo::AddOp addOp = ...; // The op instanced to be wrapped.
-  mlir::ModuleOp module = addOp->getParentOfType<mlir::ModuleOp>();
-  mlir::OpBuilder builder(addOp);
-  mlir::NamedAttrList attrs = ...; // Attributes to be set on the composite op.
-  int32_t version = 0; // Composite version.
+mlir::stablehlo::AddOp addOp = ...; // The op instanced to be wrapped.
+mlir::ModuleOp module = addOp->getParentOfType<mlir::ModuleOp>();
+mlir::OpBuilder builder(addOp);
+mlir::NamedAttrList attrs = ...; // Attributes to be set on the composite op.
+int32_t version = 0; // Composite version.
 
-  mlir::stablehlo::CompositeOp compositeOp = mlir::stablehlo::wrapOperationInComposite(builder, addOp, attrs, version, module);
-  addOp.replaceAllUsesWith(compositeOp);
+mlir::stablehlo::CompositeOp compositeOp = mlir::stablehlo::wrapOperationInComposite(builder, addOp, attrs, version, module);
+addOp.replaceAllUsesWith(compositeOp);
 ```
 
-**Mode 3: Programmatic Module-Wide Wrapping with Attribute Predicates**
+**Mode 3: Programmatic Module-Wide Wrapping with customized Attribute Handling**
 
 This mode extends programmatic wrapping to the entire module, offering
 fine-grained control over which operations are wrapped and their attributes.
 This is achieved by using the `createStablehloWrapInCompositePass` API,
-which takes an `AttributePredicateMap` as an argument.
+which takes an `CompositeAttributeProviderMap` as an argument.
 
-The `AttributePredicateMap` is a map that dictates which operations should
-be considered for wrapping and how their attributes should be handled. Its
-semantics are as follows:
+The `CompositeAttributeProviderMap` is a map that dictates which operations
+should be considered for wrapping and how their attributes should be
+handled. Its semantics are as follows:
 
 - **Keys (mlir::TypeID):** `TypeID` of an MLIR operation. If an operation's
     `TypeID` matches a key in the map, it becomes a candidate for wrapping.
@@ -450,9 +450,9 @@ semantics are as follows:
 
 // ... inside a pass or function ...
 
-stablehlo::AttributePredicateMap attributePredicateMap;
+stablehlo::CompositeAttributeProviderMap compositeAttributeProviderMap;
 
-attributePredicateMap[mlir::TypeID::get<mlir::stablehlo::AddOp>()] =
+compositeAttributeProviderMap[mlir::TypeID::get<mlir::stablehlo::AddOp>()] =
   [](mlir::Operation* op) -> std::optional<mlir::NamedAttrList> {
   // Custom logic to determine if and how to wrap the operation.
   // Example: Only wrap if it's on a specific type.
@@ -462,7 +462,7 @@ attributePredicateMap[mlir::TypeID::get<mlir::stablehlo::AddOp>()] =
   return std::nullopt; // Do not wrap.
 };
 
-pm.addPass(createStablehloWrapInCompositePass(attributePredicateMap, compositeVersion));
+pm.addPass(createStablehloWrapInCompositePass(compositeAttributeProviderMap, compositeVersion));
 if (mlir::failed(pm.run(module))) {
   return;
 }
--- a/docs/generated/stablehlo_passes.md
+++ b/docs/generated/stablehlo_passes.md
@@ -338,6 +338,143 @@ Modules valid for shape refinement must have the following properties:
   * All calls to a single function resolve to the same argument shapes, and no
     recursive / co-recursive function calls are made.
 
+### `-stablehlo-wrap-in-composite`
+
+_Wraps a non-composite  StableHLO op in a composite op._
+
+Wraps StableHLO operations in `stablehlo.composite` operations.
+
+For instance, consider a simple StableHLO program:
+
+```mlir
+  func.func @main(%arg0 : tensor<2xf32>, %arg1 : tensor<2xf32>) -> tensor<2xf32> {
+    %0 = stablehlo.add %arg0, %arg1 : tensor<2xf32>
+    return %0 : tensor<2xf32>
+  }
+```
+
+Applying this pass to wrap `stablehlo.add` operations will result in the
+following program:
+
+```mlir
+  func.func @main(%arg0: tensor<2xf32>, %arg1: tensor<2xf32>) -> tensor<2xf32> {
+    %0 = stablehlo.composite "stablehlo.add" %arg0, %arg1 {decomposition = @stablehlo.add.impl} : (tensor<2xf32>, tensor<2xf32>) -> tensor<2xf32>
+    return %0 : tensor<2xf32>
+  }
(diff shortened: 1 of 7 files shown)


docs/generated/stablehlo_passes.md
```diff
@@ -338,6 +338,59 @@ Modules valid for shape refinement must have the following properties:
   * All calls to a single function resolve to the same argument shapes, and no
     recursive / co-recursive function calls are made.
 
+### `-stablehlo-wrap-in-composite`
+
+_Wraps a non-composite  StableHLO op in a composite op._
+
+Wraps StableHLO ops, as specified by the pass option flag, in a
```
GleasonK, line 345: Let's talk about the design more. It's missing a few things to be generally useful:

- What to do with ops that have attributes?
- What should the composites be named, or should the name be user specified as a part of the callback fn, etc

Need to think more, maybe we have a little brainstorm and run through some examples

GleasonK (changes requested):

docs/generated/stablehlo_passes.md
sdasgup3: All good points! SGTM.

>> What to do with ops that have attributes?

All the op attributes are preserved as composite attributes.  To demonstrate that I have used `convolution  op` as an example in `stablehlo/tests/transforms/stablehlo_wrap_in_composite.mlir`.

ghpvnist (changes requested):

stablehlo/tests/transforms/stablehlo_wrap_in_composite.mlir
```diff
@@ -0,0 +1,68 @@
...
+         dim_numbers = [b, 0, 1, f]x[0, 1, i, o]->[b, 0, 1, f],
+         window = {stride = [1, 1], pad = [[0, 1], [0, 1]], rhs_dilate = [2, 2]}
+         {batch_group_count = 1 : i64, feature_group_count = 1 : i64} :
+       (tensor<64x8x8x8xi8>, tensor<4x4x8x32xi8>) -> tensor<64x3x3x32xi32>
+  %1 = stablehlo.add %0, %arg2 : tensor<64x3x3x32xi32>
+  %2 = stablehlo.add %1, %1 : tensor<64x3x3x32xi32>
+  func.return %2 : tensor<64x3x3x32xi32>
+}
+
+// -----
```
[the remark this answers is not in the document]
sdasgup3: Added.

>> Should this pass run as part of a pass pipeline? 

Yes, that is the more practical usage of the pass.


>> If we run a decomposer pass that introduces stablehlo.convolution, they also need to get wrapped right?

If we have a decomposer pass and it emit a convolution and if the user intend to wrap that conv into a composite, then yes it will be wrapped.
Let me know if that address your concern.

docs/generated/stablehlo_passes.md
```diff
@@ -338,6 +338,59 @@ Modules valid for shape refinement must have the following properties:
   * All calls to a single function resolve to the same argument shapes, and no
     recursive / co-recursive function calls are made.
 
+### `-stablehlo-wrap-in-composite`
+
+_Wraps a non-composite  StableHLO op in a composite op._
```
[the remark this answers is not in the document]
sdasgup3: done

stablehlo/tests/transforms/stablehlo_wrap_in_composite.mlir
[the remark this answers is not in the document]
GleasonK: > If we have a decomposer pass and it emit a convolution

The decomposer would need to be run before the wrap in composite pass, then yes it would be wrapped. Otherwise no.

> Should this pass run as part of a pass pipeline?

I'm not sure I follow this?

stablehlo/transforms/StablehloWrapInComposite.cpp
```diff
@@ -0,0 +1,216 @@
...
+    counter++;
+    funcName = (baseName + std::to_string(counter));
+  }
+  return funcName;
+}
+
+// Builds a new function within the given `module` that encapsulates the
+// functionality of the provided `implOp`. The new function is named uniquely
+// and is set to private visibility.
+mlir::func::FuncOp buildStableHLOCompositeImplFunc(mlir::ModuleOp module,
```
GleasonK, line 68: `StablehloCompositeImplFunc`

stablehlo/transforms/StablehloWrapInComposite.cpp
```diff
@@ -0,0 +1,216 @@
...
+    patterns = std::move(patterns_);
+
+    if (!opPredicate) {
+      if (!opNamesOption.empty()) {
+        DenseSet<StringRef> opNames(opNamesOption.begin(), opNamesOption.end());
+        opPredicate = [opNames](Operation* op) {
+          return opNames.contains(op->getName().getStringRef());
+        };
+      } else {
+        opPredicate = [](Operation* op) { return false; };
```
GleasonK, line 182: Can we move this to the default compiler ctor?

```
  StablehloWrapInCompositePass(const StablehloWrapInCompositePassOptions& opts)
      : StablehloWrapInCompositePassBase<StablehloWrapInCompositePass>(opts), opPredicate([](){...}) {}
```

stablehlo/transforms/StablehloWrapInComposite.cpp
```diff
@@ -0,0 +1,216 @@
...
+    : public impl::StablehloWrapInCompositePassBase<
+          StablehloWrapInCompositePass> {
+ public:
+  StablehloWrapInCompositePass()
+      : StablehloWrapInCompositePassBase<StablehloWrapInCompositePass>() {}
+  StablehloWrapInCompositePass(const StablehloWrapInCompositePassOptions& opts)
+      : StablehloWrapInCompositePassBase<StablehloWrapInCompositePass>(opts) {}
+  explicit StablehloWrapInCompositePass(
+      std::function<bool(Operation*)> opPredicate) {
+    this->opPredicate = opPredicate;
```
GleasonK, line 168: Can we use list init instead? Also shouldn't we call the pass base here as well?

`: StablehloWrapInCompositePassBase<StablehloWrapInCompositePass>(opts), opPredicate(opPredicate)`

stablehlo/transforms/StablehloWrapInComposite.cpp
```diff
@@ -0,0 +1,216 @@
...
+      }
+    }
+    return success();
+  }
+
+  void runOnOperation() override {
+    ConversionTarget target(getContext());
+    target.addLegalDialect<func::FuncDialect>();
+    target.addDynamicallyLegalDialect<stablehlo::StablehloDialect>(
+        [this](Operation* op) { return !opPredicate(op); });
```
GleasonK, line 192: `shouldWrapOpFn` ? to be more clear on what an op predicate's use is

stablehlo/transforms/StablehloWrapInComposite.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+    Operation* op = getOperation();
+    if (failed(applyPartialConversion(op, target, std::move(patterns)))) {
```
GleasonK, line 195: How does this prevent recursively wrapping an op? It looks like an op in a map is illegal?

Maybe we should use a greedy rewrite instead of a dialect rewrite and only apply the patterns once? Or potentially even just a single IR `walk`?

stablehlo/transforms/Passes.h
```diff
@@ -102,6 +103,14 @@ void populateStablehloCompatibilityExpanderPatterns(
...
     TypeRange refinedTypes);
 
+/// Creates a pass that wraps StableHLO ops in CompositeOp.
+///
+/// The pass will wrap the StableHLO ops that match the given opPredicate
+/// function in CompositeOp. The opPredicate function should return true if the
+/// op should be wrapped in CompositeOp.
+std::unique_ptr<OperationPass<ModuleOp>> createStablehloWrapInCompositePass(
+    std::function<bool(Operation *)> opPredicate);
+
```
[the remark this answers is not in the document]
sdasgup3: done

stablehlo/transforms/StablehloWrapInComposite.cpp
sdasgup3: Yes, I tried to move it here, but the issue is that the context is notinitialized yet. I need the context to initialize the `OperationName` objects from the `opNamesOption`.

Currently I am wrapping the initialization logic in `void initializePredicateMap(MLIRContext* context, ArrayRef<std::string> opNames)` which is called from the pass initialize function.

stablehlo/transforms/StablehloWrapInComposite.cpp
```diff
@@ -0,0 +1,216 @@
...
+namespace {
+
+// Generates a unique function name based on the given `baseFuncName` within
+// the provided `module`. Ensures the generated name does not clash with any
+// existing symbols by appending a counter if necessary.
+std::string generateUniqueFunctionName(StringRef baseFuncName,
+                                       mlir::ModuleOp module) {
+  mlir::SymbolTable symbolTable(module);
+  int counter = 0;
+  std::string baseName = baseFuncName.str() + ".impl";
```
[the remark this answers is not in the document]
sdasgup3: Thanks! Fixed.

stablehlo/transforms/StablehloWrapInComposite.cpp
sdasgup3: After debugging this is what I got"

In the first pass of the conversion application, each illegal op gets converted to a legal `stablehlo.composite` and no further action is taken for that op. The legality check for newly created ops are based on if they are "inserted" by the `ConversionPatternRewriter`. The new private function that we inserted is based on OpBuilder.

sdasgup3: Applied `applyPatternsGreedily` with `GreedyRewriteStrictness::ExistingOps` strictness.

stablehlo/transforms/StablehloWrapInComposite.cpp
```diff
@@ -0,0 +1,219 @@
...
+    auto compositeName = op->getName().getStringRef();
+
+    auto attributePredicate =
+        attributePredicateMap.at(op->getName().getTypeID());
+    auto namedAttributes = attributePredicate(op);
+    auto compositeAttributes = rewriter.getDictionaryAttr(*namedAttributes);
+
+    auto compositeOp = rewriter.create<stablehlo::CompositeOp>(
+        op->getLoc(), op->getResultTypes(), op->getOperands(), compositeName,
+        compositeAttributes, decomposition.getSymName());
```
ghpvnist, line 121: We initialize the `compositeVersion` in the class, but not used here. Let's include this and add a test to make sure specified version appears.

stablehlo/transforms/StablehloWrapInComposite.cpp
```diff
@@ -0,0 +1,219 @@
...
+    if (module == nullptr) {
+      return rewriter.notifyMatchFailure(op, "Failed to find enclosing module");
+    }
+
+    func::FuncOp decomposition = buildFuncOpWrappingOperation(op, module);
+    auto compositeName = op->getName().getStringRef();
+
+    auto attributePredicate =
+        attributePredicateMap.at(op->getName().getTypeID());
+    auto namedAttributes = attributePredicate(op);
```
ghpvnist, line 116: `AttributePredicateMap` returns an optional value, so let's check for nullness here.

stablehlo/transforms/Passes.h
```diff
@@ -102,6 +111,24 @@ void populateStablehloCompatibilityExpanderPatterns(
 std::unique_ptr<OperationPass<ModuleOp>> createStablehloRefineArgumentsPass(
     TypeRange refinedTypes);
 
+/// Creates a pass that wraps StableHLO ops in CompositeOp.
+/// The pass takes in a map of op type to attribute predicate. The attribute
+/// predicate is a function that takes in an operation, of the given op type,
+/// and returns a list of attributes to be added to the CompositeOp.
+using AttributePredicate =
```
ghpvnist, line 118: The name `AttributePredicate` might be slightly misleading since it doesn't just check attributes but instead transforms them into a `NamedAttrList` for the `CompositeOp`. Perhaps `CompositeAttributeProvider` and  `CompositeAttributeProviderMap` to be more descriptive?

docs/generated/stablehlo_passes.md
```diff
@@ -338,6 +338,143 @@ Modules valid for shape refinement must have the following properties:
...
+
+Wraps StableHLO operations in `stablehlo.composite` operations.
+
+For instance, consider a simple StableHLO program:
+
+```mlir
+  func.func @main(%arg0 : tensor<2xf32>, %arg1 : tensor<2xf32>) -> tensor<2xf32> {
```
ghpvnist, line 350: Let's unindent all code blocks

docs/generated/stablehlo_passes.md
```diff
@@ -338,6 +338,143 @@ Modules valid for shape refinement must have the following properties:
...
+    %0 = stablehlo.add %arg0, %arg1 : tensor<2xf32>
+    return %0 : tensor<2xf32>
+  }
+```
+
+Applying this pass to wrap `stablehlo.add` operations will result in the
+following program:
```
ghpvnist, line 357: Let's wrap the sentences to 80-char limit here and below.

ghpvnist (changes requested):

stablehlo/tests/transforms/stablehlo_wrap_in_composite.mlir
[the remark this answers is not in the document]
ghpvnist: > I'm not sure I follow this?

To clarify, the question was getting at whether this pass is run as part of a larger pass pipeline or used as just a one off pass.

stablehlo/transforms/StablehloWrapInComposite.cpp
sdasgup3: The following check at the beginning of the `matchAndRewrite` function  makes sure that the `attributePredicateMap` is returning a non-null attribute for the op under processing. At this point (line 116) the `namedAttribute` is guaranteed to be non-null. 
 
```
if (!shouldWrapInComposite(op, attributePredicateMap)) {
      return failure();
    }
```

Hence skipped the nullity check.

stablehlo/transforms/Passes.h
sdasgup3: Nice suggestion. Will add.

stablehlo/transforms/StablehloWrapInComposite.cpp
sdasgup3: Good catch! Will fix it.

sdasgup3:
@ghpvnist Addressed the comments at https://github.com/openxla/stablehlo/pull/2722/commits/ebab4e1651ed3a896bbae9acc8a223d0e9ee8d92. PTAL.

docs/generated/stablehlo_passes.md
```diff
@@ -338,6 +338,143 @@ Modules valid for shape refinement must have the following properties:
...
+  - The private function that encapsulates the original operation
+    (referenced by the `decomposition` attribute of the
+    `stablehlo.composite` operation) will be named using the pattern
+    `<op_name>.impl[.N]`, where `<op_name>` is the name of the original
+    operation, and `N` is a unique integer identifier generated to prevent
+    naming conflicts within the module.
+
+This pass can be used in three distinct ways:
+
+**Mode 1: Command-line Usage**
```
GleasonK, line 385: Would add a note that this should be for debug/test only since you have minimal control over composite attrs

docs/generated/stablehlo_passes.md
```diff
@@ -338,6 +338,143 @@ Modules valid for shape refinement must have the following properties:
...
+operations. The attributes of the newly created `stablehlo.composite`
+operation will be the same as the attributes of the original operation.
+
+**Usage Example:**
+
+```bash
+stablehlo-opt input.mlir --stablehlo-wrap-in-composite=op-names='stablehlo.add,stablehlo.mul' -o output.mlir
+```
+
+**Mode 2: Programmatic Single-Op Wrapping**
```
GleasonK, line 399: I think we should remove this mode from the _pass_ doc string. This should just be an API code comment in the header file for the wrap function

docs/generated/stablehlo_passes.md
```diff
@@ -338,6 +338,143 @@ Modules valid for shape refinement must have the following properties:
...
+        used to set the composite's attributes.
+      - If the lambda returns `std::nullopt`, the operation is **not**
+        wrapped. This allows for selective wrapping based on custom
+        criteria.
+
+**Example (C++):**
+
+```cpp
+
+// ... inside a pass or function ...
```
GleasonK, line 451: Remove code comment, this is just a pass config so it can be anywhere technically. Probably _shouldn't_ be inside another pass

docs/generated/stablehlo_passes.md
```diff
@@ -338,6 +338,143 @@ Modules valid for shape refinement must have the following properties:
...
+    wrapped (e.g., if you wrap a `stablehlo.add` operation, the composite
+    will also be named `"stablehlo.add"`).
+  - The private function that encapsulates the original operation
+    (referenced by the `decomposition` attribute of the
+    `stablehlo.composite` operation) will be named using the pattern
+    `<op_name>.impl[.N]`, where `<op_name>` is the name of the original
+    operation, and `N` is a unique integer identifier generated to prevent
+    naming conflicts within the module.
+
+This pass can be used in three distinct ways:
```
GleasonK, line 383: Two, per later comment

docs/generated/stablehlo_passes.md
```diff
@@ -338,6 +338,143 @@ Modules valid for shape refinement must have the following properties:
...
+
+**Mode 1: Command-line Usage**
+
+This mode is the simplest, using the `stablehlo-opt` utility with the
```
GleasonK, line 387: Remove this paragraph, options are covered later (auto generated) and all passes work with opt

ghpvnist (approved):

sdasgup3:
@GleasonK Addressed your feedback at https://github.com/openxla/stablehlo/pull/2722/commits/4325e2de951cd4af899d88efe6c1e2ae2f6716b6

GleasonK (approved):

PR #2729, opened 2025-02-28
Bazel build only build affected targets on pull_request

GleasonK:
- It's painful that doc-only changes compile the entire codebase and run all tests.
- We can't use `paths-ignore` from github actions because it marks jobs as pending and we have branch protections enabled which require all status checks to be green.
- This seems like a generally good change to build what you change, after merge to main we will still build and test all.

Calculates diff in GH actions using https://github.com/Tinder/bazel-diff

## Example 1, this PR, GH actions files only, no impacted build files:

```sh
$ curl -Lo /tmp/bazel-diff.jar https://github.com/Tinder/bazel-diff/releases/latest/download/bazel-diff_deploy.jar
$ ./build_tools/github_actions/ci_build_bazel.sh /tmp/bazel-diff.jar $(git merge-base main HEAD) $(git rev-parse HEAD)
Generating Hashes for Revision 'd2708ad6cca730d4c17317388a7acbf8eac2c9f6'
Loading: 0 packages loaded

Generating Hashes for Revision '1565f5bac3ed5df5e0bdd38843fb84efde28e1d1'
Loading: 0 packages loaded

Determining Impacted Targets

No impacted targets from change.
```

## Example 2, modify `StablehloOps.cpp`, fairly core file, should test ~all:

```sh
$ ./build_tools/github_actions/ci_build_bazel.sh /tmp/bazel-diff.jar $(git merge-base main HEAD) $(git rev-parse HEAD)
Generating Hashes for Revision 'd2708ad6cca730d4c17317388a7acbf8eac2c9f6'
Loading: 0 packages loaded

Generating Hashes for Revision '310993d82d94b0cf7a5389e659e35ac32cd32089'
Loading: 0 packages loaded

Determining Impacted Targets

[3188] Impacted Targets between d2708ad6cca730d4c17317388a7acbf8eac2c9f6 and 310993d82d94b0cf7a5389e659e35ac32cd32089:
//:interpreter_ops
//:interpreter_passes
//:linalg_passes
//:reference_api
[...]
//stablehlo/testdata:abs_int32_20_20.mlir.test
//stablehlo/testdata:abs_int64_20_20.mlir.test
[...]
//stablehlo/tests:check_ops
//stablehlo/tests:chlo/chlo_legalize_to_stablehlo.mlir.test
//stablehlo/tests:chlo/chlo_legalize_to_stablehlo_broadcast.mlir.test
//stablehlo/tests:infer_chlo.mlir.test
//stablehlo/tests:infer_stablehlo.mlir.test
//stablehlo/tests:interpret/abs.mlir.test
[...]
```

### DIFF
--- a/build_tools/github_actions/ci_build_bazel.sh
+++ b/build_tools/github_actions/ci_build_bazel.sh
@@ -16,11 +16,76 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-if [[ $# -ne 0 ]] ; then
-  echo "Usage: $0"
+if [[ $# -ne 0 && $# -ne 3 ]] ; then
+  echo "Usage: $0 [<bazel-diff.jar> <start_commit> <end_commit>]"
+  echo "   Builds and tests all bazel targets."
+  echo "   If start_commit and end_commit are specified, uses bazel-diff to"
+  echo "   determine the set of targets to build based on code changes."
+  echo
+  echo "Example: Build all affected targets between current branch and main"
+  echo "  $ curl -Lo bazel-diff.jar https://github.com/Tinder/bazel-diff/releases/latest/download/bazel-diff_deploy.jar"
+  echo "  $ ci_build_bazel.sh bazel-diff.jar $(git merge-base main HEAD) $(git rev-parse HEAD)"
   exit 1
 fi
 
-# Build and Test StableHLO
-bazel build --lockfile_mode=error //... --config=asan --config=ubsan
-bazel test //... --config=asan --config=ubsan
+bazel-test-all() {
+  # Build and Test StableHLO
+  bazel build --lockfile_mode=error //... --config=asan --config=ubsan
+  bazel test //... --config=asan --config=ubsan
+}
+
+bazel-test-diff() {
+  set -e
+  WORKSPACE_PATH=$(git rev-parse --show-toplevel)
+  BAZEL_DIFF_JAR="$1"
+  PREVIOUS_REV="$2" # Starting Revision SHA
+  FINAL_REV="$3" # Final Revision SHA
+  BAZEL_PATH="$(which bazel)"
+
+  STARTING_HASHES_JSON="/tmp/starting_hashes.json"
+  FINAL_HASHES_JSON="/tmp/final_hashes.json"
+  IMPACTED_TARGETS_PATH="/tmp/impacted_targets.txt"
+
+  bazel-diff() {
+    java -jar "$BAZEL_DIFF_JAR" "$@"
+  }
+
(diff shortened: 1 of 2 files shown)


GleasonK:
cc @MichaelHudgins @ddunl - any thoughts on this / maybe worth doing for the XLA repo?

I had a doc-only change in XLA and thought it was strange that it resulted in compiling XLA code on various build configs:
https://github.com/openxla/xla/actions/runs/13574863537/job/37948558762?pr=23205

ddunl:
I think we could do this for XLA, I'm having trouble seeing how we could apply this to the TF/JAX tests on the XLA repo, but I think for XLA tests this should be possible

build_tools/github_actions/ci_build_bazel.sh
```diff
@@ -16,11 +16,77 @@ set -o errexit
...
+
+  echo "Generating Hashes for Revision '$FINAL_REV'"
+  bazel-diff generate-hashes -w "$WORKSPACE_PATH" -b "$BAZEL_PATH" $FINAL_HASHES_JSON
+
+  echo "Determining Impacted Targets"
+  bazel-diff get-impacted-targets -sh $STARTING_HASHES_JSON -fh $FINAL_HASHES_JSON -o $IMPACTED_TARGETS_PATH
+  echo ""
+
+  impacted_targets=()
+  IFS=$'\n' read -d '' -r -a impacted_targets < $IMPACTED_TARGETS_PATH || true
```
sdasgup3, line 68: How about we save/restore IFS 
```
IFS_OLD="$IFS"
IFS= .... || true
IFS="$IFS_OLD"
```

to keep the bash string splitting behavior predictable for any future modification of this file.

sdasgup3 (approved):
LGTM otherwise.

MichaelHudgins:
> We can't use paths-ignore from github actions because it marks jobs as pending and we have branch protections enabled which require all status checks to be green.

Yeah this is annoying, you can do some workarounds to try to get the job into a skipped state(which counts as passing) instead of just not running but its a gap in GH actions functionality for sure.

> cc @MichaelHudgins @ddunl - any thoughts on this / maybe worth doing for the XLA repo?

I think its interesting and probably worth an experiment to see how much time it saves.  There seems a good chance though that jobs using RBE or a remote cache may already be getting a lot the benefits that this would provide.

@vam-google wonder if you have any thoughts on this?

build_tools/github_actions/ci_build_bazel.sh
GleasonK: This is copy-pate-edited from the `bazel-diff` upstream script, so probably will keep as is to keep any future syncing of this file easier

Issue #2727, opened 2025-02-28
StableHLO Aggressive Simplification erroring on dynamic iota


bartchr808:
### What happened?

Dynamic iota with certain dynamic shapes fails during aggressive simplication

```
/tmp/bug.mlir:2:10: error: 'arith.index_cast' op operand type 'tensor<2xi32>' and result type 'tensor<2xi64>' are cast incompatible
```

This is from a JAX test (`tensor<?x?xi64>`):
```mlir
func.func @dynamic_iota_dynamic_shape(%arg0: tensor<2xi32>) -> tensor<?x?xi64> {
  %0 = stablehlo.dynamic_iota %arg0, dim = 0 : (tensor<2xi32>) -> tensor<?x?xi64>
  return %0 : tensor<?x?xi64>
}
```
but this also fails (`tensor<?x2xi64>`):
```mlir
func.func @main(%arg0: tensor<2xi32>) -> tensor<?x2xi64> {
  %0 = stablehlo.dynamic_iota %arg0, dim = 0 : (tensor<2xi32>) -> tensor<?x2xi64>
  return %0 : tensor<?x2xi64>
}
```
But this passes (`tensor<2x?xi64>`):
```
func.func @main(%arg0: tensor<2xi32>) -> tensor<2x?xi64> {
  %0 = stablehlo.dynamic_iota %arg0, dim = 0 : (tensor<2xi32>) -> tensor<2x?xi64>
  return %0 : tensor<2x?xi64>
}
```

It seems `DynamicIotaOpToBroadcast` can handle some dynamic shapes, but not all. E.g. some dynamic shapes are [already tested](https://github.com/openxla/stablehlo/blob/main/stablehlo/tests/transforms/stablehlo_aggressive_simplification.mlir#L430).

### Steps to reproduce your issue

Run `stablehlo-opt` with `--stablehlo-aggressive-simplification`

mrguenther:
I have a fix that I'll upload soon.

The issue was that the transformation assumed the shape would be passed as either an `i64` tensor or an `index` tensor. It added an `index_cast` to convert `index` to `i64` and back when necessary, relying on canonicalization to remove any trivial `i64`-to`-i64` casts. However, if the shape was, say, an `i32` tensor, the `index_cast` wouldn't be trivial but would fail since that op is for converting to and from `index`. As a result, the transformation failed when it tried to `index_cast` between `i32` and `i64`.

[...]

PR #2745, opened 2025-03-13
Fix a casting bug in `DynamicIotaOpToBroadcast`.

mrguenther:
The transformation sometimes generated an invalid `index_cast` op between two non-index int types.

Issue #2727

### DIFF
--- a/stablehlo/transforms/optimization/StablehloAggressiveSimplification.cpp
+++ b/stablehlo/transforms/optimization/StablehloAggressiveSimplification.cpp
@@ -388,19 +388,18 @@ struct DynamicIotaOpToBroadcast : public OpRewritePattern<DynamicIotaOp> {
 
   LogicalResult matchAndRewrite(DynamicIotaOp iota,
                                 PatternRewriter &rewriter) const override {
-    ShapedType resultType = cast<ShapedType>(iota.getType());
+    auto resultType = cast<ShapedType>(iota.getType());
     if (resultType.getRank() < 2)
       return rewriter.notifyMatchFailure(iota, "requires rank >= 2");
 
     Location iotaLoc = iota.getLoc();
-    int64_t iotaDimension = static_cast<int64_t>(iota.getIotaDimension());
+    auto iotaDimension = static_cast<int64_t>(iota.getIotaDimension());
 
     Value iotaShape = iota.getOutputShape();
-    Value iotaShapeI64;
-    ShapedType iotaShapeType = cast<ShapedType>(iotaShape.getType());
-    ShapedType iotaShapeI64Type =
+    auto iotaShapeType = cast<ShapedType>(iotaShape.getType());
+    auto iotaShapeI64Type =
         RankedTensorType::get(iotaShapeType.getShape(), rewriter.getI64Type());
-
+    Value iotaShapeI64;
     if (iotaShapeType.getElementType().isIndex()) {
       iotaShapeI64 = rewriter.create<arith::IndexCastOp>(
           iotaLoc, iotaShapeI64Type, iotaShape);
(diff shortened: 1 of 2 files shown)


stablehlo/transforms/optimization/StablehloAggressiveSimplification.cpp
```diff
@@ -388,43 +388,41 @@ struct DynamicIotaOpToBroadcast : public OpRewritePattern<DynamicIotaOp> {
 
   LogicalResult matchAndRewrite(DynamicIotaOp iota,
                                 PatternRewriter &rewriter) const override {
-    auto resultTy = cast<ShapedType>(iota.getType());
-    if (resultTy.getRank() < 2)
+    ShapedType resultType = cast<ShapedType>(iota.getType());
```
GleasonK, line 391: We use LLVM coding standards in this repo, so keeping this as `auto` since `cast` is a "place where the type is already obvious from the context"

https://releases.llvm.org/10.0.0/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable

GleasonK: Similar feedback below for the ShapedType casts

stablehlo/transforms/optimization/StablehloAggressiveSimplification.cpp
```diff
@@ -388,43 +388,41 @@ struct DynamicIotaOpToBroadcast : public OpRewritePattern<DynamicIotaOp> {
...
+    if (resultType.getRank() < 2)
       return rewriter.notifyMatchFailure(iota, "requires rank >= 2");
 
-    auto iotaDimension = static_cast<int64_t>(iota.getIotaDimension());
+    Location iotaLoc = iota.getLoc();
+    int64_t iotaDimension = static_cast<int64_t>(iota.getIotaDimension());
+
+    Value iotaShape = iota.getOutputShape();
+    Value iotaShapeI64;
```
GleasonK, line 399: nit - maybe move this _right_ above the conditional where its assigned, I had to backtrack up to see what this was init to

stablehlo/transforms/optimization/StablehloAggressiveSimplification.cpp
```diff
@@ -726,12 +724,13 @@ struct ReduceOpUnusedResultCanon final : OpRewritePattern<ReduceOp> {
 
     const auto newNumOperandPairs = usedResults.size();
     const auto newNumOperands = newNumOperandPairs * pairSize;
-    if (newNumOperands != usedArgs.count())
+    if (newNumOperands != usedArgs.count()) {
```
GleasonK, line 727: I'm ok with this change since its multiline, but as an FYI for LLVM coding standards, if the body of an if is a single statement then skip the `{}`s, this one is a single statement, but multiline may benefit from braces for readability

stablehlo/transforms/optimization/StablehloAggressiveSimplification.cpp
GleasonK: Removing individual comments and will leave this in a top-level comment. In general I think here are a few good cases for `auto`:

```
// Trivially OpType, currently downcasting to Value in this code
auto someOp = rewriter.create<OpType>(...)

// Trivially RankedTensorType, currently casting to Shaped Type
auto someType = RankedTensorType::get(...)

// Trivially SomeType type
auto someThing = llvm::cast<SomeType>(...)
```

GleasonK (approved):

stablehlo/transforms/optimization/StablehloAggressiveSimplification.cpp
mrguenther: The linter said to add braces since it was mutiline despite being a single logical statement. Maybe the linter was misconfigured on my end?

stablehlo/transforms/optimization/StablehloAggressiveSimplification.cpp
mrguenther: Sounds good. I'll change them to `auto` where it's obvious.

mrguenther: BTW, I intentionally downcast that op to `Value` so that I could reassign it in some code paths, but that's no longer necessary since we decided to remove the cast at the end.

mrguenther: Done.

stablehlo/transforms/optimization/StablehloAggressiveSimplification.cpp
mrguenther: Done.

PR #2753, opened 2025-03-20
Add more MHLO folders to StableHLO

GleasonK:
Some of these may seem unsafe (simplifying float `X*1.0` or `X*0.0` may not account for NaNs properly), but both MHLO and XLA simplifications do this optim so probably generally safe.

### DIFF
--- a/stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp
+++ b/stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp
@@ -12,9 +12,12 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 
+#include <cmath>
 #include <cstddef>
 #include <cstdint>
+#include <functional>
 #include <numeric>
+#include <optional>
 #include <utility>
 
 #include "llvm/ADT/APInt.h"
@@ -23,11 +26,14 @@ limitations under the License.
 #include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/ADT/StringRef.h"
+#include "llvm/Support/Casting.h"
 #include "llvm/Support/Debug.h"
 #include "llvm/Support/ErrorHandling.h"
+#include "llvm/Support/LogicalResult.h"
 #include "mlir/Dialect/CommonFolders.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/Utils/IndexingUtils.h"
+#include "mlir/IR/BuiltinAttributeInterfaces.h"
 #include "mlir/IR/BuiltinAttributes.h"
 #include "mlir/IR/BuiltinTypeInterfaces.h"
 #include "mlir/IR/BuiltinTypes.h"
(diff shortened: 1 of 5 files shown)


stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp
```diff
@@ -279,6 +285,28 @@ struct FoldAddOpPattern final : OpRewritePattern<mlir::stablehlo::AddOp> {
...
+  ShapeOpRewritePattern(MLIRContext* context, PatternBenefit benefit,
+                        bool foldFloat_)
+      : OpRewritePattern<OpType>(context, benefit), foldFloat{foldFloat_} {}
+
+  using OpRewritePattern<OpType>::OpRewritePattern;
+  using OpRewritePattern<OpType>::matchAndRewrite;
+
+  LogicalResult validateShapeFold(PatternRewriter& rewriter, OpType op,
+                                  ShapedType resultType) {
+    if (!resultType.getElementType().isInteger()) {
```
mrguenther, line 301: Should this condition also check `foldFloat`?

stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp
```diff
@@ -613,16 +622,20 @@ struct EvalRemOpPattern : public OpRewritePattern<RemOp> {
...
+struct EvalReshapeOpPattern : public ShapeOpRewritePattern<ReshapeOp> {
+  using ShapeOpRewritePattern::ShapeOpRewritePattern;
+
   LogicalResult matchAndRewrite(ReshapeOp op,
                                 PatternRewriter& rewriter) const override {
     auto resultType = op.getType();
     if (failed(validateResultTypeForEval(rewriter, op, resultType)))
       return failure();
 
+    if (!foldFloat && !resultType.getElementType().isInteger())
```
mrguenther, line 634: Is it possible for an element type to be neither float nor int?

stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp
```diff
@@ -778,6 +791,37 @@ struct EvalSubtractOpPattern : public OpRewritePattern<SubtractOp> {
...
+    return rewriter.notifyMatchFailure(op, "unable to fold sqrt");
+  }
+
+  static std::optional<APFloat> foldSqrt(const APFloat& a) {
+    if (a.getSizeInBits(a.getSemantics()) == 64)
+      return APFloat(std::sqrt(a.convertToDouble()));
+
+    if (a.getSizeInBits(a.getSemantics()) == 32)
+      return APFloat(std::sqrtf(a.convertToFloat()));
+    return {};
```
mrguenther: Is there a reason we only support `f32` and `f64`? Is it because the system running the compiler might not support `std::sqrt(std::float16_t)` for folding `f16` constants?

mrguenther (approved):

mrguenther (approved):

stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp
GleasonK: It is possible (`tensor<complex<f64>>`), but I believe that will fail to match the `m_Constant(operand, DenseIntOrFPElementsAttr)` call below

GleasonK: addressed in `validateShape` function commented above

stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp
GleasonK: image

Seems like most dialects only bother with f32/f64 folding of sqrt (TFL, MHLO, and math dialects). Perhaps we could add more but I wonder if its an accuracy issue to fold other precisions

Issue #2751, opened 2025-03-19
Fix TOSA related breakages


sdasgup3:
Hi @Tai78641

The recent TOSA dialect changes introduced by[ https://github.com/llvm/llvm-project/pull/129720](https://github.com/llvm/llvm-project/pull/129720) have caused breakages, for example:

```
/usr/local/google/home/sdasgup/Github/stablehlo/stablehlo/conversions/tosa/transforms/TosaRescaleLegalizeToStablehlo.cpp:68:25: error: no member named 'getRoundingMode' in 'mlir::tosa::RescaleOp'                
  bool doubleRound = op.getRoundingMode() == "DOUBLE_ROUND";                                                                                                                                                       
                     ~~ ^                                                                                                                                                                                          
/usr/local/google/home/sdasgup/Github/stablehlo/stablehlo/conversions/tosa/transforms/TosaRescaleLegalizeToStablehlo.cpp:110:9: error: no viable conversion from '::llvm::ArrayRef<int32_t>' (aka 'ArrayRef<int>') 
to 'Value'                                                                                                                                                                                                         
  Value multiplier = op.getMultiplier();   
```

The [PR](https://github.com/openxla/stablehlo/pull/2750) is an attempt to address some of these issues, but it appears that the problem is more involving than we initially anticipated. Do you mind helping us updating the `StablehloQuantLegalizeToTosaRescale.cpp` and `TosaRescaleLegalizeToStablehlo.cpp` passes to accommodate these recent changes.

sdasgup3:
https://github.com/openxla/stablehlo/pull/2752 disables certain build and test target to unblock integration while the while issue is fixed. 

Make sure to enable those targets once the current issue is resolved.

akuegel:
I already have a fix for the compile errors, and adjusted the tests legalize_quant_ops_to_tosa_rescale.mlir and legalize_tosa_rescale_to_stablehlo.mlir
The other disabled tests are still failing and will need adjustments.

Tai78641:
I pushed a PR: https://github.com/openxla/stablehlo/pull/2756

sdasgup3:
@Tai78641 

Thanks again for your fixes! 

>> once the stablehlo llvm has been updated to include that, I will test and re-enable tosa legalization in stablehlo

https://github.com/openxla/stablehlo/pull/2758 is an attempt to test all the affected tests (including nullary.mlir,binary.mlir, and ternay.mlir) after updating the llvm. 

[...]

PR #2756, opened 2025-03-24
[stablehlo][tosa] Fix for tosa dialect updates

Tai78641:
This fixes up legalization passes for converting to/from tosa for tosa dialect changes

- tosa rescale has multiplier, shift, input_zp, output_zp changed from attribute to operands with specific type requirements
- tosa rescale attribute "double_round" changed to "rounding_mode"
- tosa negate has additional input_zp and output_zp operands
- tosa const attribute name changed from "value" to "values"


Tai78641:
this address the issue: https://github.com/openxla/stablehlo/issues/2751#issuecomment-2742702873
but note that this fixes the code and tests for recent tosa dialect changes merged into llvm up to and including:
https://github.com/llvm/llvm-project/pull/130340 (which has yet to be pulled into stablehlo's llvm-project)

GleasonK (approved):

GleasonK:
Thanks for the fixes!

PR #2767, opened 2025-04-03
Update docs dense to array

mrguenther:
Following up on PR #1872, this updates the ODS documentation comment to show the new input type of `array<i64: foo>` instead of the old `dense<foo> : tensor<Nxi64>`.


mrguenther:
I accidentally committed that on the wrong branch. I'll re-upload it correctly.

mrguenther:
See https://github.com/openxla/stablehlo/pull/2768 for the fixed PR.

PR #2770, opened 2025-04-07
Drop failure case for `stablehlo.dynamic_broadcast_in_dim`

rsuderman:
The failure to broadcast dynamically makes the assumption the input dynamic shape could be expanded by being `1`. This should be handled by an earlier trasform to materialize a known broadcast if we intend to support both cases.

### DIFF
--- a/stablehlo/conversions/linalg/tests/miscellaneous.mlir
+++ b/stablehlo/conversions/linalg/tests/miscellaneous.mlir
@@ -206,8 +206,8 @@ func.func @constant() -> tensor<i32> {
 
 // -----
 
-// CHECK-DAG: #[[MAP0:.+]] = affine_map<(d0, d1) -> (d1)>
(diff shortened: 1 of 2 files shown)


rsuderman:
@GleasonK if you do not mind looking over. I believe there is a way to safely introduce the conditional broadcasting but it depends on how `stablehlo.get_dimension_size` is lowered.

rsuderman:
> Could you say more about requirement on lowering of get_dimension_size? Also is that what's happening at the moment via the lowering to linalg path?

There is no current lowering in the `stablehlo` repo. `iree` side we just lower to `tensor.dim` and `tensor.from_elements` instead of concatting everything together.

Something worth noting is to have the conditional broadcast (e.g. `?` being a broadcasting `1`) we could materialize out an additional `1` dim then broadcast across it.

PR #2775, opened 2025-04-10
Refactor StableHLO passes to support options; add option for op-folding element limit.

mrguenther:
Instead of hard-coding the max number of elements in a folded operation, set the limit via a config option. (The default value matches the old constant's value of 65536.)


GleasonK (approved):

stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp
```diff
@@ -927,7 +935,8 @@ struct StablehloAggressiveFolderPass
 
   LogicalResult initialize(MLIRContext* context) override {
     RewritePatternSet patterns_(context);
-    populateStablehloAggressiveFolderPatterns(&patterns_, context, foldFloat);
+    populateStablehloAggressiveFolderPatterns(&patterns_, context, foldFloat,
```
GleasonK, line 938: We can make this method take in a `StablehloAggressiveFolderPassOptions` struct

stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp
```diff
@@ -817,13 +818,18 @@ struct FoldSqrtOpPattern : public OpRewritePattern<mlir::stablehlo::SqrtOp> {
 };
 
 struct EvalIotaOpPattern : public OpRewritePattern<IotaOp> {
```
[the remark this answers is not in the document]
mrguenther: Yeah, I almost did that. The only reason I didn't was because the option was only used by two patterns, but I definitely agree with creating that class if other patterns might use the option in future.

stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp
mrguenther: Good call. Will do.

stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp
```diff
@@ -923,65 +959,97 @@ struct EvalTransposeOpPattern : public OpRewritePattern<TransposeOp> {
...
+std::unique_ptr<::mlir::Pass> createStablehloAggressiveFolderPass(
+    StablehloAggressiveFolderPassOptions options,
+    GreedyRewriteConfig rewriteConfig) {
+  return std::make_unique<StablehloAggressiveFolderPass>(options,
+                                                         rewriteConfig);
 }
 
-}  // namespace stablehlo
-}  // namespace mlir
+}  // namespace mlir::stablehlo
```
GleasonK, line 1055: Let's keep the multiple namespace decls. Rationale is (1) we shouldn't do this as a single-file thing. We should update them all together once we set a standard, for the sake of consistency. (2) This is a c++20 feature, I'm not sure how much of our codebase uses c++20 yet, and don't want this to be the early introduction that breaks some oss users.

GleasonK (approved):

stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp
mrguenther: It's a C++17 feature, so we should be safe. I can still undo it for the other reason you mentioned, though.

stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp
mrguenther: Done.

stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp
```diff
@@ -947,16 +956,18 @@ struct StablehloAggressiveFolderPass
...
   patterns->add<EvalTransposeOpPattern>(context, benefit);
 
   // TODO: Consolidate FoldOp patterns
   // One is used by Shape Refinement, the other is a generic folder.
   patterns->add<FoldAddOpPattern, FoldBroadcastInDimSplatPattern,
-                FoldConcatenateOpPattern, FoldMulOpPattern,
-                FoldSubtractOpPattern, FoldSqrtOpPattern>(context);
+                FoldMulOpPattern, FoldSubtractOpPattern, FoldSqrtOpPattern>(
+      context);
+  patterns->add<FoldConcatenateOpPattern>(context, foldOpElementLimit);
```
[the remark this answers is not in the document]
mrguenther: Done.

stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp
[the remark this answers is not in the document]
mrguenther: Done.

stablehlo/transforms/optimization/StablehloTargetIndependentOptimization.cpp
```diff
@@ -58,7 +53,7 @@ struct StablehloTargetIndependentOptimizationPass
     GreedyRewriteConfig config;
     config.fold = true;
     config.cseConstants = true;
-    config.maxIterations = kFoldOpEltLimit;
+    config.maxIterations = foldOpElementLimit;
```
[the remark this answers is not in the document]
mrguenther: Updated per our offline discussion.

GleasonK (approved):
`./build_tools/github_actions/ci_build_docs.sh` needed

PR #2784, opened 2025-04-23
[RFC] Add source_target_pairs attribute to send and recv ops in stableHLO

rosiezou:


mjsML:
cc: @chaserileyroberts @nouiz @jaro-sevcik @gspschmid @Tixxx @jprabhas

rfcs/20250421-source-target-pairs.md
```diff
@@ -0,0 +1,136 @@
...
+`recv` ops. `source_target_pairs` allows users to specify peer-to-peer
+communication patterns using global device IDs (zero-indexed integers).
+Currently this feature is only available on GPUs.
+
+## Background
+
+SPMD-based pipeline parallelism relies on optimizations in XLA to pipeline
+send/recv operations in such a way that compute and communication are
+overlapped. The user expresses this through collective permutes and relies on
+XLA to decompose these into send/recv operations, which are then pipelined
```
sdasgup3, line 20: This is more for my understanding the motivation:

The hlo-level `send`/`recv`s, generated by the XLA decomposition, do carry the source/target info using attribs like ` frontend_attributes {_xla_send_recv_source_target_pairs={{3,0}}}`. I am not clear how exposing this information at a high-level Stablehlo representation would help as the only stablehlo operations, exported from Jax, are collective permutes (not the stablehlo send/recv's).

rosiezou: Per the last sentence in this paragraph, I'm also working on adding `psend` and `precv` to `jax.lax.parallel`. I didn't include it in this RFC because I figured it's not related to stableHLO.

sdasgup3: @rosiezou Is there an available RFC for the JAX side of things.

rosiezou: Not an official RFC but https://github.com/jax-ml/jax/pull/28101 is a WIP of the JAX changes.

rfcs/20250421-source-target-pairs.md
```diff
@@ -0,0 +1,136 @@
...
+  * `DEVICE_TO_HOST` if `is_host_transfer = true`,
+  * `DEVICE_TO_DEVICE` otherwise.
+
+#### Examples
+
+```mlir
+%results0, %results1 = "stablehlo.recv"(%token) {
+  source_target_pairs = dense<[[0, 1], [1, 2]]> : tensor<2x2xi64>,
+  channel_handle = #stablehlo.channel_handle<handle = 1, type = 3>,
+  is_host_transfer = true
```
sdasgup3, line 134: Can we keep it `false` to make it clear that the `source_target_pairs` will have some effect in this particular example.

rfcs/20250421-source-target-pairs.md
```diff
@@ -0,0 +1,136 @@
...
+`shard_map` API.
+
+## Proposed Specification
+
+### send
+
+#### Semantics
+
+Sends `inputs` to a channel `channel_id`. Inputs are then sent to other devices
+in the order specified by `source_target_pairs`. The operation produces a
```
sdasgup3, line 35: Similar to `recv` can we defer `in the order specified by `source_target_pairs` to the second paragraph, unless we somehow capture the event of `is_host_transfer` being `true`  in the same paragraph. 

By goal here to keep the wording consistent with recv.

rfcs/20250421-source-target-pairs.md
rosiezou: suggestion applied

rfcs/20250421-source-target-pairs.md
rosiezou: suggestion applied

rosiezou:
@GleasonK is this RFC good to go now?

rfcs/20250421-source-target-pairs.md
```diff
@@ -0,0 +1,134 @@
...
+overlapped. The user expresses this through collective permutes and relies on
+XLA to decompose these into send/recv operations, which are then pipelined
+separately, allowing for the staggering that is unique to pipeline parallelism.
+The limitation of this approach is that it encapsulates the latency hiding
+mechanism in the compiler and allows for little control by the user. When this
+mechanism fails, the user has little choice but to debug XLA itself. This RFC is
+proposed in conjunction with exposing send/recv operations through the JAX
+`shard_map` API.
+
+## Proposed Specification
```
GleasonK, line 28: Can you add a section for "A Note on Backward Compatibility", denoting that we're technically making the semantics more strict for send given that any instances of `send is_host_transfer=false` that are serialized will no longer be deserializable, but this is likely OK since this would have been undefined behavior as it is.

GleasonK (approved):

rfcs/20250421-source-target-pairs.md
GleasonK: cc @frgossen to confirm this as well

PR #2790, opened 2025-04-30
Fix an outdated pass-creation function declaration

mrguenther:
The declaration for a pass-creation function was located in the wrong header file, resulting in a linker error if someone tried to use it. This change moves the declaration to the correct header and updates its function signature per PR #2775.


ghpvnist (approved):
Thanks for the fix. Let's land this after the internal integrate lands.

GleasonK (approved):

GleasonK:
> Thanks for the fix. Let's land this after the internal integrate lands.

Apologies didn't see this comment, hopefully doesn't throw a wrench in things. Its a fairly minimal change so probably ok?

Issue #2736, opened 2025-03-07
StableHLO Missing Folders / Optimization Patterns


GleasonK:
### Request description

Using this ticket to keep a log of missing / potential patterns for aggressive simplification or folding

- [ ] `reduce(args...) ({ return cst }) -> cst`
- [ ] https://github.com/EnzymeAD/Enzyme-JAX/pull/414
- [ ] _add any more_

GleasonK:
Test file for WhileOp and ReduceOp simplifiers / folders.

```
// Requires `WhileOp::fold` pattern
func.func @dce_while_false_condition(%arg0: tensor<i64>) -> tensor<i64> {
  // CHECK-NOT: stablehlo.while
  %0 = stablehlo.while(%iterArg = %arg0) : tensor<i64>
    cond {
    %1 = stablehlo.compare  LT, %arg0, %arg0 : (tensor<i64>, tensor<i64>) -> tensor<i1>
    stablehlo.return %1 : tensor<i1>
  } do {
    %1 = stablehlo.custom_call @something(%iterArg) : (tensor<i64>) -> tensor<i64>
    stablehlo.return %1 : tensor<i64>
  }
  return %0 : tensor<i64>
}

// -----

// Requires `LowerBoolSplatConstantsIntoRegion` and `ReduceOp::fold` pattern
// Already have ReduceEmpty / UnusedResult patterns.
func.func @reduce_op_fold(%arg0: tensor<i64>) -> tensor<i1> {
  %c = stablehlo.constant dense<false> : tensor<4x32xi1>
  %c_0 = stablehlo.constant dense<false> : tensor<i1>
  // CHECK-NOT: stablehlo.reduce
  %0 = stablehlo.reduce(%c init: %c_0) applies stablehlo.or across dimensions = [0, 1] : (tensor<4x32xi1>, tensor<i1>) -> tensor<i1>
  return %0 : tensor<i1>
}
```

GleasonK:
`PadIsNoop`

```
func.func @_pad_12(%arg0: tensor<256x1024xbf16>, %arg1: tensor<i32>) -> tensor<256x1024xbf16> {
  %0 = stablehlo.convert %arg1 : (tensor<i32>) -> tensor<bf16>
  %1 = stablehlo.pad %arg0, %0, low = [0, 0], high = [0, 0], interior = [0, 0] : (tensor<256x1024xbf16>, tensor<bf16>) -> tensor<256x1024xbf16>
[...]

PR #2795, opened 2025-05-02
Add patterns for folding Reduce ops.

mrguenther:
Add four missing patterns targeting Reduce ops:
- FoldReduceOpReducingZeroDims
- FoldReduceOpToConstantInitializer
- FoldReduceOpWithRedundantResults
- LowerBoolSplatConstantsIntoReduceOpRegion

Issue: #2736

### DIFF
--- a/stablehlo/tests/transforms/stablehlo_aggressive_folder.mlir
+++ b/stablehlo/tests/transforms/stablehlo_aggressive_folder.mlir
@@ -180,6 +180,21 @@ func.func @eval_iota_zero_dimension() -> (tensor<0xi32>, tensor<5x0x2xi32>) {
 // -----
 
 ////////
+// ReduceOp
+
+// Requires `LowerBoolSplatConstantsIntoRegion` and `ReduceOp::fold` pattern
+// Already have ReduceEmpty / UnusedResult patterns.
+func.func @reduce_op_fold(%arg0: tensor<i64>) -> tensor<i1> {
+  %c = stablehlo.constant dense<false> : tensor<4x32xi1>
+  %c_0 = stablehlo.constant dense<false> : tensor<i1>
+  // CHECK-NOT: stablehlo.reduce
+  %0 = stablehlo.reduce(%c init: %c_0) applies stablehlo.or across dimensions = [0, 1] : (tensor<4x32xi1>, tensor<i1>) -> tensor<i1>
+  return %0 : tensor<i1>
+}
+
+// -----
+
+////////
 // ReshapeOp
 
 // CHECK-LABEL: func @reshape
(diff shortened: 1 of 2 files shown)


stablehlo/tests/transforms/stablehlo_aggressive_folder.mlir
```diff
@@ -179,6 +179,21 @@ func.func @eval_iota_zero_dimension() -> (tensor<0xi32>, tensor<5x0x2xi32>) {
 
 // -----
 
+////////
+// ReduceOp
+
+// Requires `LowerBoolSplatConstantsIntoRegion` and `ReduceOp::fold` pattern
```
GleasonK, line 185: Don't need this temp comment

stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp
```diff
@@ -957,6 +957,132 @@ struct EvalTransposeOpPattern : public FoldOpRewritePattern<TransposeOp> {
...
+      auto* sourceOfBlockResult = bodyResult.getDefiningOp();
+      if (!sourceOfBlockResult ||
+          !sourceOfBlockResult->hasTrait<OpTrait::ConstantLike>())
+        return failure();
+
+      DenseElementsAttr constantAttr;
+      if (!matchPattern(sourceOfBlockResult, m_Constant(&constantAttr)))
+        return failure();
+
+      auto resultShapedType = dyn_cast_or_null<ShapedType>(opResult.getType());
```
GleasonK, line 1044: Probably don't need or null, this is for when the attr or thing you're pointing at can be null.

stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp
```diff
@@ -957,6 +957,132 @@ struct EvalTransposeOpPattern : public FoldOpRewritePattern<TransposeOp> {
...
+    auto returnOp = dyn_cast<ReturnOp>(body.back());
+    if (!returnOp) return failure();
+
+    Region* returnOpParentRegion = returnOp->getParentRegion();
+
+    for (auto [reduceOpResult, returnOpResult] :
+         llvm::zip_equal(op.getResults(), returnOp.getResults())) {
+      if (returnOpResult.getParentRegion() == returnOpParentRegion ||
+          returnOpResult.getType() != reduceOpResult.getType()) {
+        return failure();
```
GleasonK, line 1078: Can we add better failure messages for some of these. Use notify match failure

GleasonK (approved):

stablehlo/tests/transforms/stablehlo_aggressive_folder.mlir
mrguenther: Oops; removed.

stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp
mrguenther: I wasn't sure why it was this way in the MHLO implementation, but I decided to play it safe at first. Anyway, I've switched to a simple `dyn_cast` now.

PR #2801, opened 2025-05-13
Optionally DCE loops more liberally (opt-in).

mrguenther:
Firstly, this fixes a bug where we assumed a missing `has_side_effect` attribute meant that a `custom_call` had every side effect even though the attribute was documented as semantically `false` if unset.

Secondly, added an option (disabled by default) to assume, for purposes of eliminating dead `while` ops, that ops are pure unless they say otherwise via the MLIR `MemoryEffects` interface. The default behavior is still to play it safe and assume ops have side effects if they don't specify one way or the other, but enabling the new option can result in significantly smaller IR. (Note that MHLO optimizations already make this assumption when removing dead `while` ops, which suggests that it's typically safe to enable the option.)


mrguenther:
I know you're OOO, Kevin, so no pressure to review this; I just thought you'd want to be in the loop.

docs/generated/stablehlo_optimization_passes.md
```diff
@@ -7,8 +7,9 @@ _Folds StableHLO operations_
 #### Options
 
 ```
--fold-op-element-limit : Upper limit on how many elements may be folded by an op folder. This limit doesn't apply in certain special cases such as addition of 0, multiplication by 1, and some splat operations.
--fold-float            : Allow for potentially lossy computations using float type.
+-assume-no-undeclared-side-effects : Allow dead code to be eliminated in some situations (e.g. dead while loops) under the assumption that ops are pure unless declared with explicit MLIR `MemoryEffects`. Notably, this means `func.call` ops will be assumed pure.
+-fold-op-element-limit             : Upper limit on how many elements may be folded by an op folder. This limit doesn't apply in certain special cases such as addition with 0, multiplication by 1, and some splat operations.
+-optimize-float                    : Allow float optimizations that, though mathematically equivalent, may result in slightly different quantization of floating-point values (e.g. `log(sqrt(x))` -> `0.5 * log(x)`). Float optimizations that can't affect numerical results are always enabled.
```
ghpvnist, line 12: Is this related to the DCE change?

ghpvnist (changes requested):

docs/generated/stablehlo_optimization_passes.md
ghpvnist: Specifically the optimize-float flag

stablehlo/dialect/StablehloOps.td
```diff
@@ -1301,8 +1317,10 @@ def StableHLO_PartitionIdOp : StableHLO_Op<"partition_id", [Pure,
 // StableHLO control flow op definitions.
 //===----------------------------------------------------------------------===//
 
-def StableHLO_AfterAllOp : StableHLO_Op<"after_all", [Pure,
-    DeclareOpInterfaceMethods<InferTypeOpInterface>]> {
+def StableHLO_AfterAllOp : StableHLO_Op<"after_all", [
+      Pure,
```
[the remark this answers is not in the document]
mrguenther: I didn't change this; I just reformatted it a bit. I'm not familiar with that op, so I'm inclined to avoid messing with it until Kevin is back.

docs/generated/stablehlo_optimization_passes.md
mrguenther: This is the generated documentation change corresponding to my changes in `Passes.td`. I added the new `assume-no-undeclared-side-effects` option, and I clarified the documentation of other options while I was at it.

mrguenther: Oh, the `optimize-float` thing. That part is mostly unrelated; Kevin and I had previously discussed renaming it so the same name could make sense when applied to simplifier patterns as well. I was already updating the options, so I went ahead and included that change. It's not directly related to DCE, though, no.

ghpvnist: Makes sense, for future changes let's split additional changes to another PR to simplify reviews.

ghpvnist (approved):

docs/generated/stablehlo_optimization_passes.md
mrguenther: Sure, will do.

stablehlo/dialect/StablehloOps.h
```diff
@@ -134,6 +134,23 @@ ParseResult parseWindowAttributes(OpAsmParser &parser,
...
+namespace side_effects {
+
+struct SendResource : ::mlir::SideEffects::Resource::Base<SendResource> {
+  StringRef getName() final { return "SendResource"; }
+};
+struct RecvResource : ::mlir::SideEffects::Resource::Base<RecvResource> {
+  StringRef getName() final { return "RecvResource"; }
+};
+struct InfeedResource : ::mlir::SideEffects::Resource::Base<InfeedResource> {
+  StringRef getName() final { return "OutfeedResource"; }
```
mrguenther, line 146: Typo fixed in https://github.com/openxla/stablehlo/pull/2802

PR #2807, opened 2025-05-28
Canonicalize `mhlo.backend_config` to `backend_config`

mrguenther:
When `mhlo.backend_config` is specified and `backend_config` is either unset or empty, move the contents of `mhlo.backend_config` into `backend_config`.

### DIFF
--- a/stablehlo/transforms/optimization/StablehloAggressiveSimplification.cpp
+++ b/stablehlo/transforms/optimization/StablehloAggressiveSimplification.cpp
@@ -284,6 +284,37 @@ class ConcatenateOpFlatten : public SimplifyOpRewritePattern<ConcatenateOp> {
   }
 };
 
+/////////////////////////////////
+// CustomCallOp
+/////////////////////////////////
+
+struct FixMhloBackendConfigAttribute final
+    : SimplifyOpRewritePattern<CustomCallOp> {
+  using SimplifyOpRewritePattern::SimplifyOpRewritePattern;
+
+  LogicalResult matchAndRewrite(CustomCallOp op,
+                                PatternRewriter &rewriter) const override {
+    if (op.getApiVersion() != CustomCallApiVersion::API_VERSION_ORIGINAL)
+      return rewriter.notifyMatchFailure(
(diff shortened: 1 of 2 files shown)


stablehlo/transforms/optimization/StablehloAggressiveSimplification.cpp
```diff
@@ -284,6 +284,37 @@ class ConcatenateOpFlatten : public SimplifyOpRewritePattern<ConcatenateOp> {
   }
 };
 
+/////////////////////////////////
+// CustomCallOp
+/////////////////////////////////
+
+struct FixMhloBackendConfigAttribute final
```
GleasonK, line 291: How about something like: `CustomCallUnregisteredBackendConfigToFfi`

More consistent with the op-first naming, also would like to leave "mhlo" out of this. Also also we should be able to delete this pattern in the near futrue if we migrate JAX after

stablehlo/transforms/optimization/StablehloAggressiveSimplification.cpp
```diff
@@ -284,6 +284,37 @@ class ConcatenateOpFlatten : public SimplifyOpRewritePattern<ConcatenateOp> {
...
+      return rewriter.notifyMatchFailure(
+          op, "No `mhlo.backend_config` attribute to fix.");
+
+    if (!dyn_cast<StringAttr>(op.getBackendConfigOrDefault()).empty())
+      return rewriter.notifyMatchFailure(
+          op, "Non-empty `backend_config` attribute shouldn't be overwritten.");
+
+    op.setBackendConfigAttr(mhloBackendConfigAttr);
+    op.setApiVersion(CustomCallApiVersion::API_VERSION_TYPED_FFI);
+    op->removeAttr("mhlo.backend_config");
```
GleasonK, line 312: Store this string as a variable above? Used more than once in the function, less chance of desync

GleasonK (approved):

stablehlo/transforms/optimization/StablehloAggressiveSimplification.cpp
mrguenther: Good call; done.

PR #2841, opened 2025-08-07
Integrate LLVM at llvm/llvm-project@6d4a0935c850

abhigunj:


mrguenther (approved):

GleasonK:
I think you just need to patch in:
https://github.com/openxla/stablehlo/pull/2844

PR #2858, opened 2025-09-22
Add SVG for `broadcast_in_dim`

mtsokol:
A missing graphic for `broadcast_in_dim` operation.


docs/images/spec/broadcast_in_dim.svg
GleasonK, line 1: Could you replace this with:
https://gist.github.com/GleasonK/05010b3a36ed32ee5203e7cd4ca55764

I copied the slide contents into a drawing.google.com page, cropped it, and then saved as SVG. There will be less empty whitespace above/below with this

GleasonK (approved):

PR #2859, opened 2025-09-25
Integrate LLVM at llvm/llvm-project@79a0bf0efce5

mrguenther:


GleasonK (approved):

mrguenther:
Reverting this and proceeding with #2860 instead.

PR #2876, opened 2025-11-15
Spec's scatter example has incorrect dimensions

kentslaney:
In scatter's example, the dimensions for `%update` don't match between the commented input and the MLIR.

https://github.com/openxla/stablehlo/blob/96acdcb7724f4a9eec6d2e5af2597b0750c13948/docs/spec.md?plain=1#L5739-L5749

https://github.com/openxla/stablehlo/blob/96acdcb7724f4a9eec6d2e5af2597b0750c13948/docs/spec.md?plain=1#L5763

```python
>>> jnp.array([
...            [
...             [[1, 1], [1, 1], [1, 1]],
...             [[1, 1], [1, 1], [1, 1]]
...            ],
...            [
...             [[1, 1], [1, 1], [1, 1]],
...             [[1, 1], [1, 1], [1, 1]]
...            ]
...           ]).shape
(2, 2, 3, 2)
```


kentslaney:
actually, it looks like this is the line that got mistranslated

https://github.com/openxla/stablehlo/blob/96acdcb7724f4a9eec6d2e5af2597b0750c13948/stablehlo/tests/interpret/scatter.mlir#L54

not sure why `%updates` got switched to `%update` but the docs are consistent for that one

Issue #2829, opened 2025-07-04
Assert when lowering 1D Convolution from StableHLO to LinAlg


pstarkcdpr:
### What happened?

I found this when reporting the following IREE issue: https://github.com/iree-org/iree/issues/21230#issuecomment-3033570610

When compiling a model in StableHLO using IREE with `iree-compile` in Debug mode, the following assert triggers.
```
~/iree-test$ iree-org/iree-compiler/tools/iree-compile --iree-hal-target-device=vulkan --iree-vulkan-target=sm_75 -o model1.vmfb model1.mlir
iree-compile: /home/paulstark/iree-test/iree-org/iree/third_party/llvm-project/mlir/include/mlir/IR/OperationSupport.h:1082: void mlir::OperationState::addAttribute(mlir::StringAttr, mlir::Attribute): Assertion `attr && "attribute cannot be null"' failed.
Please report issues to https://github.com/iree-org/iree/issues and include the crash backtrace.
Stack dump:
0.      Program arguments: iree-org/iree-compiler/tools/iree-compile --iree-hal-target-device=vulkan --iree-vulkan-target=sm_75 -o model1.vmfb model1.mlir
 #0 0x0000742db96831e0 ___interceptor_backtrace.part.0 /build/gcc-14-ig5ci0/gcc-14-14.2.0/build/x86_64-linux-gnu/libsanitizer/asan/../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:4358:28
 #1 0x0000742d746eca69 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) /home/paulstark/iree-test/iree-org/iree/third_party/llvm-project/llvm/lib/Support/Unix/Signals.inc:804:22
 #2 0x0000742d746edb50 PrintStackTraceSignalHandler(void*) /home/paulstark/iree-test/iree-org/iree/third_party/llvm-project/llvm/lib/Support/Unix/Signals.inc:888:1
 #3 0x0000742d746e4d40 llvm::sys::RunSignalHandlers() /home/paulstark/iree-test/iree-org/iree/third_party/llvm-project/llvm/lib/Support/Signals.cpp:105:20
 #4 0x0000742d746eb07a SignalHandler(int, siginfo_t*, void*) /home/paulstark/iree-test/iree-org/iree/third_party/llvm-project/llvm/lib/Support/Unix/Signals.inc:418:13
[...]

PR #2880, opened 2025-12-02
Avoid assert with convolutions with no strides or dilates. Issue #2829

pstarkcdpr:
See issue #2829. A convolution without strides or dilates was causing spurious asserts.

### DIFF
--- a/stablehlo/conversions/linalg/transforms/StablehloToLinalgConvolution.cpp
+++ b/stablehlo/conversions/linalg/transforms/StablehloToLinalgConvolution.cpp
@@ -248,10 +248,12 @@ struct NormalConvolutionOpConversion final
                                 resultType.getElementType(), dynSizes);
     Value zeroTensor = fillTensorWithZeros(rewriter, loc, emptyTensor);
     linalg::LinalgOp res;
-    Attribute strides;
-    if (auto s = op.getWindowStrides()) strides = rewriter.getI64TensorAttr(*s);
-    Attribute dilations;
-    if (auto d = op.getRhsDilation()) dilations = rewriter.getI64TensorAttr(*d);
+    auto s = op.getWindowStrides();
+    Attribute strides =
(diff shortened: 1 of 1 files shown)


stablehlo/conversions/linalg/transforms/StablehloToLinalgConvolution.cpp
```diff
@@ -268,24 +268,42 @@ struct NormalConvolutionOpConversion final
         break;
       }
       case 3: {
-        res = linalg::Conv1DNwcWcfOp::create(
-            rewriter, loc, resultType, ValueRange{input, filter},
-            ValueRange{zeroTensor}, strides, dilations,
-            linalg::getPrunedAttributeList(op));
+        if (strides && dilations) {
```
GleasonK: Alternate impl - looks like the default values for both are `{1}`. Could we instead update the initialize the code above to:

```
auto s = op.getWindowStrides();
Attribute strides = s ? rewriter.getI64TensorAttr(*s) : rewriter.getI64TensorAttr({1});
```

I'm surprised the builder method isn't resilient to null attrs being passed in though.

pstarkcdpr:
@GleasonK I'm checking if I'm allowed to agree to the CLA since I'm using a corporate account. I'll update when I know the answer. Alternately, this is such a simple change that if someone else wants to just do it and push it in, I'm fine with that and can cancel this PR if that gets done.

stablehlo/conversions/linalg/transforms/StablehloToLinalgConvolution.cpp
pstarkcdpr: You are entirely right. My implementation was wrong anyways. I have a model that does a convolution with strides but no dilates, and that gets missed by doing the `&&`. I was trying to avoid repeating the defaults, but your way `getI64TensorAttr({1})` is very concise and looks like the best approach to me (if the code gen refuses to accept null attributes). I'll change this and update the PR.

GleasonK (approved):

PR #2882, opened 2025-12-07
Convert stablehlo.dynamic_slice into tosa.slice when indices are const

vvish:
If the `stablehlo`->`tosa` transformation is applied without stablehlo dynamism canonicalization (as a part of e.g. `stablehlo-canonicalize-dynamism` or `stablehlo-aggressive-simplification`) `stablehlo.dynamic_slice` can be converted into `tosa.slice` if the indices are const.

### DIFF
--- a/stablehlo/conversions/tosa/tests/unary.mlir
+++ b/stablehlo/conversions/tosa/tests/unary.mlir
@@ -114,6 +114,28 @@ func.func @slice_rank_seven(%arg : tensor<2x3x4x5x6x7x8xf32>) -> tensor<1x2x3x4x
   return %0 : tensor<1x2x3x4x5x6x7xf32>
 }
 
+// CHECK-LABEL: @dynamic_slice_constant_start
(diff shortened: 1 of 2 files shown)


abhigunj:
Hello, thanks for the contribution, if it is ready for a review let's move it out of draft so that we can kick-start the review process.

vvish:
> Hello, thanks for the contribution, if it is ready for a review let's move it out of draft so that we can kick-start the review process.

Hello @abhigunj, @GleasonK. Sorry for the stale PR. It covers some specific case from my experiments, but I expect is quite "exotic" for a normal workflow. What do you think - does it make sense to go ahead and review this change?

abhigunj:
> It covers some specific case from my experiments, but I expect is quite "exotic" for a normal workflow. What do you think - does it make sense to go ahead and review this change?

Yeh, agree. Instead rely on stablehlo simplifications to do stablehlo dynamic_slice -> stablehlo slice before using tosa lowering.

vvish:
Closing as `stablehlo` dynamic-to-static transformations will convert `dynamic_slice` into `slice`

PR #2884, opened 2025-12-17
Integrate LLVM at llvm/llvm-project@8f264586d752

abhigunj:


stablehlo/integrations/cpp/builder/ChloBuilderTest.cpp
```diff
@@ -0,0 +1,141 @@
...
+#include "mlir/IR/Verifier.h"
+#include "mlir/Support/DebugStringHelper.h"
+#include "mlir/Support/LLVM.h"
+#include "stablehlo/dialect/Register.h"
+#include "stablehlo/integrations/cpp/builder/AttrTypeBuilderUtil.h"
+#include "stablehlo/integrations/cpp/builder/ChloBuilder.h"
+#include "stablehlo/integrations/cpp/builder/FuncBuilder.h"
+#include "stablehlo/integrations/cpp/builder/MlirBuilder.h"
+#include "stablehlo/integrations/cpp/builder/StablehloBuilder.h"
+#include "testing/base/public/gunit.h"
```
GleasonK: `#include "gtest/gtest.h"`

.bazelversion
```diff
@@ -1,4 +1,4 @@
-7.0.2
+7.4.1
```
abhigunj: Making this change to address Bazel build failure. LLVM upstream utils/bazel/configure.bzl is updated to use the Bazel .is_dir field which is introduced in 7.4.0 version. StableHLO is at 7.0.1, causing build failures. Bumping the bazel version fixes the build failure. any concerns wrt StableHLO consumers @GleasonK ?

abhigunj: build failure before the change: https://github.com/openxla/stablehlo/actions/runs/20320956992/job/58376309791

PR #2888, opened 2026-01-06
Integrate LLVM at llvm/llvm-project@125948a058dc

mrguenther:


abhigunj:
temporary.patch changes are missing? I see it is not empty.

mrguenther:
> temporary.patch changes are missing? I see it is not empty.

`temporary.patch` contained outdated/redundant diff hunks that had already been applied.

PR #2897, opened 2026-02-09
Add async collectives RFC.

mwhittaker:


felixwqp (commented):
Can I assume implicitly this will only work for shard_map-based sharding? like how ragged_all_to_all is used?

rfcs/20260209-async-collectives.md
```diff
@@ -0,0 +1,164 @@
...
+```
+%result = "stablehlo.collective_permute_done"(%future) : (future<tensor<2x2xi64>>) -> tensor<2x2xi64>
+```
+
+Start and done ops must be matched in the obvious way. It is an error to pass
+the output of an `all_gather_start`, for example, to `reduce_scatter_done`.
+
+## Alternatives
+
+### Generic Async Ops
```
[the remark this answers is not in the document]
GleasonK: I tend to prefer opset brevity, but open if XLA folks have a reason to prefer the explicit pairs.

> It's unclear which ops can be async. [...] But, the code supports fusions, calls, and custom calls.

I agree with this.. I think documentation can help mitigate that, but I'd prefer we start with just the ops you need to support from the framework side. Given that JAX can't emit fusions I don't think this is in scope, and given that we may not need async-update I'd guess custom calls are out of scope. Calls... probably the same deal and done need them.

> I have a hard time understanding what the generic async-update is used for.

Agree, we can leave this off for now if JAX doesn't need to emit these updates

> a sign that the ops are hard to read compared to explicit start/done pairs.

I think this can be mitigated by making async ops use regions not called functions (which we probably should do regardless), technically we can match the existing prettyprint we use for reduce op for many of these async ops:

```
%0 = stablehlo.async_start %arg0 using stablehlo.all_gather {replica_groups={{0, 2, 4, 6}, {1, 3, 5, 7}}, dimensions={1}} : (tensor<128x32xf32>) -> !stablehlo.async_bundle<tensor<128x32xf32>, tensor<128x128xf32>>
```

Separate from the op definition, a benefit of adding the async_bundle type is that (IIUC) the only ops that can process it are `done` ops, this encodes that invariant into the type system. I don't think the output of a start should flow into other computations(?)

GleasonK (approved):

rfcs/20260209-async-collectives.md
```diff
@@ -0,0 +1,187 @@
...
+
+We introduce a new future type as follows.
+
+```ebnf
+ValueType ::= TensorType | QuantizedTensorType | TokenType | TupleType | BufferType | FutureType
+FutureType ::= 'future' '<' FutureValueType '>'
+FutureValueType ::= TensorType | QuantizedTensorType
+```
+
+## Proposed Op Changes
```
GleasonK: We should figure out if scheduling groups needs to be explicitly represented to ensure that your ordering is honored by the compiler. We can do this during development and backport to the RFC as needed. The general direction of "expose async collectives" seems to be generally agreed upon, we can figure out precise details as we go

GleasonK (approved):

rfcs/20260209-async-collectives.md
```diff
@@ -0,0 +1,187 @@
...
+operations---`all_gather`, `all_reduce`, `all_to_all`, `collective_broadcast`,
+`collective_permute`, and `reduce_scatter`---are lowered to HLO equivalents.
+Internally, [the XLA compiler splits these operations into asynchronous
+start/done pairs][async_collective_creator]. For example, an `all-reduce`
+operation because a pair of an `all-reduce-start` and `all-reduce-done`. Then,
+the XLA scheduler---the component responsible for picking the order in which to
+run ops---can schedule local computation between a start/done pair.
+
+The XLA scheduler is not perfect. Sometimes, it picks bad schedules. That's why
+we want to allow JAX programmers to manually specify (or at least influence) how
```
joelberkeley: not just JAX! :)

rfcs/20260209-async-collectives.md
```diff
@@ -0,0 +1,187 @@
...
+```
+
+It is an error if the region contains anything other than a call to a
+collective.
+
+We also introduce an `async_done` op which takes a variadic number of futures
+and unwraps them. Here's an example.
+
+```
+"stablehlo.async_done"(%f1, %f2) : (!stablehlo.future<tensor<4x4xf32>>, !stablehlo.future<tensor<2xf32>>) -> (tensor<4x4xf32>, tensor<2xf32>)
```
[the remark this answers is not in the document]
mwhittaker: Yes, I believe so.

rfcs/20260209-async-collectives.md
mwhittaker: Yes, totally agree.

fhoushmand:
Would this RFC extend to async dynamic-slice/dynamic-update-slice?
We want to add those to jax as well. If this RFC covers it, great. If not we can talk about extending the support.

GleasonK (approved):

mattjj (approved):
Thanks!

fhoushmand (approved):

abhigunj:
Changes from the RPC are already landed, forgot to merge the RFC. Thank you for the contribution.

PR #2920, opened 2026-03-18
Add ScatterOp no-op simplification patterns

brucechanglongxu:
ScatterOp had zero simplification patterns — even trivially dead
scatters were being left in the IR. This adds two no-op patterns to
StablehloAggressiveSimplification:

1. ScatterOpEmptyIndices — scatter_indices has no scatter points (a
   non-index_vector_dim axis is 0), so nothing gets scattered. The
   inputs pass through.

2. ScatterOpZeroExtentUpdates — all update tensors have zero elements,
   same result.

These mirror the existing DynamicUpdateSliceOp_RemoveNoop /
ConcatenateOpRemoveEmpty patterns. More interesting patterns (e.g.
full-replace folding when the update body is a trivial return of arg1)
can come later, but these two cover the low-hanging fruit.

Tests included for both patterns plus a negative case to make sure we
don't over-simplify.

Toward #2736

### DIFF
--- a/stablehlo/tests/transforms/stablehlo_aggressive_simplification.mlir
+++ b/stablehlo/tests/transforms/stablehlo_aggressive_simplification.mlir
@@ -545,15 +545,12 @@ func.func @dynamic_iota_broadcast_dim0_i64(%arg0 : tensor<2xi64>) -> tensor<5x?x
   func.return %0 : tensor<5x?xi32>
 }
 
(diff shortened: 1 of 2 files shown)


GleasonK (approved):

abhigunj:
Could you please fix the lint issue and rebase on main to see if that resolves the Bazel and CMake build failures? thanks.

abhigunj (approved):

PR #2930, opened 2026-04-01
Use self hosted runners for CI

GleasonK:


abhigunj:
merged via [pr#2944](https://github.com/openxla/stablehlo/pull/2944)

PR #2934, opened 2026-04-09
Fix RefineCallOpPattern in StablehloRefineShapes to preserve side-effecting callees

christopherbate:
The constant-function replacement in RefineCallOpPattern did not check
whether the callee contains side-effecting operations. This caused calls
to functions with side effects (e.g. custom_call with
has_side_effect=true) to be silently erased. Guard the replacement with
a walk that checks for MemoryEffectOpInterface declarations.

### DIFF
--- a/stablehlo/transforms/StablehloRefineShapes.cpp
+++ b/stablehlo/transforms/StablehloRefineShapes.cpp
@@ -45,6 +45,7 @@ limitations under the License.
 #include "mlir/IR/Value.h"
 #include "mlir/IR/ValueRange.h"
 #include "mlir/Interfaces/InferTypeOpInterface.h"
+#include "mlir/Interfaces/SideEffectInterfaces.h"
 #include "mlir/Rewrite/FrozenRewritePatternSet.h"
 #include "mlir/Support/LLVM.h"
 #include "mlir/Support/LogicalResult.h"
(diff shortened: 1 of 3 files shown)


stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -566,13 +567,21 @@ struct RefineCallOpPattern : public OpRewritePattern<func::CallOp> {
     std::optional<SmallVector<DenseIntElementsAttr>> constantAttrs =
         isConstantFunction(callee);
     if (constantAttrs.has_value()) {
-      SmallVector<Value> constants;
-      for (auto constAttr : constantAttrs.value()) {
-        constants.push_back(
-            ConstantOp::create(rewriter, op.getLoc(), constAttr));
+      auto sideEffectResult = callee.walk([](Operation* nestedOp) {
```
GleasonK, line 570: nit - `bool hasSideEffects = callee.walk(...).wasInterrupted();` to make it clear that interruption indicates side effect in the same line

GleasonK: Also I'm not sure if this is a complete solution. If there's a call op with a nested call op I don't think this catches its side effects. We had this problem in simplification as well and made a similar method:
https://github.com/openxla/stablehlo/blob/a48c8e6daf52961205f4e6b39a56fd674a55118a/stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp#L358-L375

A complete solution would likely require an analysis pass that maps call->side effects and does a full pass to calculate transitive side effects

GleasonK (approved):

stablehlo/transforms/StablehloRefineShapes.cpp
christopherbate: Ok let me see if there's a way I can address this by computing it up front... I already didn't like the walk because the result should be cached

mrguenther:
I think this is obsolete now that PR #2978 is merged, correct?

PR #2935, opened 2026-04-09
Port new entries from temporary patch to GH

abhigunj:


abhigunj:
llvm bump revision https://github.com/openxla/stablehlo/pull/2938 will take care of this change.

PR #2940, opened 2026-04-18
Refactor vhlo::TensorV1Attr to store DenseElementsAttr

gb-prog:
This change updates the internal representation of vhlo::TensorV1Attr to hold a DenseElementsAttr instead of just an ArrayRef<char>. This aligns better with MLIR's built-in attribute types. Call sites are updated to access the raw data via getData().getRawData() when interacting with bytecode or other formats requiring raw buffers.


abhigunj:
this has merged separately, no longer needed.

PR #2947, opened 2026-05-07
Preserve discardable attributes when rewriting transpose to reshape

nisheethlahoti:
The `TransposeIsReshape` pattern uses `rewriter.replaceOpWithNewOp<ReshapeOp>(...)` to fold a layout-equivalent transpose (one whose permutation only shuffles size-1 dims) into a reshape. `replaceOpWithNewOp` does not copy the original op's discardable attribute dictionary, so attributes such as `mhlo.frontend_attributes` are silently dropped on the resulting reshape.

This breaks downstream consumers that group ops by frontend attribute. Concretely, a JAX program using `jax.experimental.compute_on('device_host')` to host-offload a sub-graph containing a multi-output `lax.scan` followed by `expm1(moveaxis(scan_output, 0, -1))` produces stablehlo where the post-scan transpose is tagged with `_xla_compute_type=host`. After this pattern fires, the resulting reshape no longer carries the tag, fragmenting what XLA's host-offload pass expects to be a contiguous host region. The runtime then fails with `INVALID_ARGUMENT: Result shape s32[] does not match program shape ...`, as the host_execute returns a single value while its declared signature is the full while-tuple.

The fix mirrors what tablegen-defined patterns in this file already do via the `MergeDiscardableAttributes` `NativeCodeCall` helper: snapshot the discardable attribute dictionary before the rewrite and re-apply it to the new op. A regression test is added alongside `@transpose_is_reshape`.

### DIFF
--- a/stablehlo/tests/transforms/stablehlo_aggressive_simplification.mlir
+++ b/stablehlo/tests/transforms/stablehlo_aggressive_simplification.mlir
@@ -1971,6 +1971,14 @@ func.func @transpose_is_reshape(%arg0: tensor<1x4x5x1xf32>) -> tensor<1x4x1x5xf3
   return %0 : tensor<1x4x1x5xf32>
 }
 
+// CHECK-LABEL: @transpose_is_reshape_preserves_discardable_attrs
+func.func @transpose_is_reshape_preserves_discardable_attrs(%arg0: tensor<99x1xf32>) -> tensor<1x99xf32> {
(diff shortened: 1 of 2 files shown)


nisheethlahoti:
Turns out this is actually a particular instance of a much more general pattern of functions doing this. Doesn't make sense to fix at just one point, and I'm not familiar or confident enough with the code to change _all_ functions. Closing.

PR #2950, opened 2026-05-20
Adapt to Linalg test to new upstream arith canonicalizer

ingomueller-net:
This PR adapts the CHECK statements of a linalg conversion test to the new canonicalization patterns of the arith dialect introduced in llvm-project#195588. It will only be needed (or valid) once that PR is integrated into stablehlo.


abhigunj:
This has addressed during llvm intergrate, next llvm version bump PR will include it. Thank you for the sending this separately.

PR #2953, opened 2026-06-09
Use result element type for accumulation for integer convolution

kevinbchen:
This changes the stablehlo interpreter implementation of convolution to accumulate to the result element type instead of the LHS element type (e.g. i8xi8=>i32 accumulates to i32 instead of i8).


abhigunj:
LGTM. @srishti-pm  to once confirm okay to use output type to be the accumulation type as per the spec.

srishti-pm (approved):
I confirm that this change aligns with the spec. LGTM.

abhigunj (approved):

PR #2957, opened 2026-06-22
Fix chlo.zeta lowering returning NaN for q == +inf

### LINKED ISSUE #121501: GPU tf.math.zeta returns NaN for finite s and infinite q
### Issue type

Bug

### Have you reproduced the bug with TensorFlow Nightly?

Yes

### Source

binary

### TensorFlow version

tf-nightly 2.22.0.dev20260616

### Custom code

Yes

### OS platform and distribution

Linux-6.17.0-29-generic-x86_64-with-glibc2.39

### Mobile device

_No response_

### Python version

3.11.15

### Bazel version

_No response_

### GCC/compiler version

_No response_

### CUDA/cuDNN version

_No response_

### GPU model and memory

_No response_

### Current behavior?

On GPU, `tf.math.zeta` returns `NaN` when `s = 2.0` and `q = inf`.

The problematic input is `tf.math.zeta([2.0], [inf])`. TensorFlow GPU returns `NaN`, but the expected result is `0.0`.

## Expected behavior

`tf.math.zeta([2.0], [inf])` should return `0.0`.

Expected full result:

```text
[0.0]
```

### Standalone code to reproduce the issue

```shell
#!/usr/bin/env python3
import os
import sys

os.environ.setdefault("TF_CPP_MIN_LOG_LEVEL", "3")
os.environ.setdefault("TF_ENABLE_ONEDNN_OPTS", "0")

import numpy as np
import tensorflow as tf

s = np.array([2.0], dtype=np.float32)
q = np.array([np.inf], dtype=np.float32)
correct = np.array([0.0], dtype=np.float32)

gpus = tf.config.list_logical_devices("GPU")
if not gpus:
    print("missing target: TensorFlow GPU is not available")
    sys.exit(2)

with tf.device("/GPU:0"):
    wrong = tf.math.zeta(tf.constant(s), tf.constant(q)).numpy()

print("wrong:", wrong.tolist())
print("correct:", correct.tolist())

if not np.array_equal(wrong, correct, equal_nan=True):
    sys.exit(0)

sys.exit(1)
```

### Relevant log output

```shell
wrong: [nan]
correct: [0.0]
```

Pravin-raj-03:
### Summary

Fixes tensorflow/tensorflow#121501 — `tf.math.zeta(s, +inf)` returns NaN on GPU instead of `0`.

ζ(s, q) = Σ_k (k+q)^-s → 0 as q → +∞ for s > 1, but `materializeZeta`'s Euler–Maclaurin expansion returns NaN: `correctionEulerMaclaurin = qNegPower * acc / (x - 1)` evaluates to `0 * inf = NaN` when `q == +inf` (`qNegPower` underflows to 0 while `acc` is `+inf`), and the "accurate enough" early-out `abs(qNegPower) < abs(powerSum) * smallest` is `false` when both `qNegPower` and `powerSum` are `0`, so the NaN `accurateResult` is selected instead of the partial sum `powerSum` (= 0).

This is the GPU code path for `tf.math.zeta`: the `mlir_generated` kernels lower `chlo.zeta` through this decomposition. The eager-CPU path uses Eigen's Cephes zeta, which already guards this with `if (b == 0) return s;` — which is exactly why the bug is GPU-only.

### Fix

Fall back to `powerSum` whenever `qNegPower == 0`, mirroring Eigen's guard. Normal inputs are unaffected. The CHLO→StableHLO FileCheck test is updated for the two added ops (`compare EQ` + `or`).

### Verification

A float32 simulation of the exact algorithm: `zeta(2, inf)` and `zeta(3, inf)` go `nan → 0.0`; `zeta(2, 1) = π²/6 = 1.644934`, `zeta(2, 5) = 0.2213230`, `zeta(4, 2.5) = 0.0373176` unchanged.

Note: I was unable to build MLIR/LLVM locally, so the FileCheck update is logic- and numerically-verified but not build-verified; please run `bazel test //stablehlo/tests:chlo_legalize_to_stablehlo` in CI.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/stablehlo/tests/chlo/chlo_legalize_to_stablehlo.mlir
+++ b/stablehlo/tests/chlo/chlo_legalize_to_stablehlo.mlir
@@ -2538,10 +2538,12 @@ func.func @digamma_f16(%arg : tensor<f16>) -> tensor<f16> {
 // CHECK:           %[[ADD_55:.*]] = stablehlo.add %[[ADD_54]], %[[MULTIPLY_36]] : tensor<f32>
 // CHECK:           %[[ABS_0:.*]] = stablehlo.abs %[[POWER_10]] : tensor<f32>
 // CHECK:           %[[ABS_1:.*]] = stablehlo.abs %[[ADD_17]] : tensor<f32>
+// CHECK:           %[[COMPARE_ZETA_EQ:.*]] = stablehlo.compare  EQ, %[[POWER_10]], %[[CONSTANT_0]] : (tensor<f32>, tensor<f32>) -> tensor<i1>
 // CHECK:           %[[CONSTANT_38:.*]] = stablehlo.constant dense<1.401300e-45> : tensor<f32>
 // CHECK:           %[[MULTIPLY_37:.*]] = stablehlo.multiply %[[ABS_1]], %[[CONSTANT_38]] : tensor<f32>
 // CHECK:           %[[COMPARE_0:.*]] = stablehlo.compare  LT, %[[ABS_0]], %[[MULTIPLY_37]] : (tensor<f32>, tensor<f32>) -> tensor<i1>
-// CHECK:           %[[SELECT_0:.*]] = stablehlo.select %[[COMPARE_0]], %[[ADD_17]], %[[ADD_55]] : tensor<i1>, tensor<f32>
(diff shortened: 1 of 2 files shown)


GleasonK:
Hello! Thank you for the fix. If you're able to fill out the CLA please do so we can kick off the review process

abhigunj:
> Hello! Thank you for the fix. If you're able to fill out the CLA please do so we can kick off the review process

Gentle Reminder to sign the CLA to start the review process, thanks.

PR #2963, opened 2026-07-01
reference/NumPy: harden .npy header parsing against malformed input

EylonKrause:
## Description

Hardening `reference/NumPy.cpp` against malformed `.npy` headers (reachable via `stablehlo-translate --interpret` on a crafted or foreign `.npy`), in the same spirit as #2962:

1. **Uncaught `std::stoi` → abort (DoS).** `parseDescrHeader` ends with `return std::stoi(typeString.substr(2));`. `std::stoi` throws `std::invalid_argument` on a non-numeric size (e.g. a header `'descr': '<ixy'`); nothing catches it, so the interpreter terminates:
   ```
   terminate called after throwing an instance of 'std::invalid_argument'
     what():  stoi
   Aborted (core dumped)
   ```
   Replaced with `llvm::StringRef::getAsInteger`, returning `errc::invalid_argument`.

2. **`front()`/`back()` before the length check.** `typeString.front()`/`.back()` are read before any size guard; a header whose `descr` value is empty makes these a precondition violation. Guard `typeString.size() < 2` first (the existing `< 3` check comes afterward).

3. **`substr` length underflow in `parseShapeHeader`.** `header.substr(shapeOffset, shapeOffset - dimEnd)` — since `dimEnd > shapeOffset`, the length underflows `size_t` to a huge value, so the shape regex scans the whole remaining header instead of just the shape tuple (defeating the intended "tight bound" noted in the comment). Corrected to `dimEnd - shapeOffset`.

## Testing

I don't have a local StableHLO/Bazel build. The abort in (1) is demonstrated with a standalone repro (uncaught `std::stoi` on a non-numeric string aborts, as above); (2) and (3) are straightforward. Happy to add a lit/unit test for `reference/` if you can point me at the preferred harness.

### DIFF
--- a/stablehlo/reference/NumPy.cpp
+++ b/stablehlo/reference/NumPy.cpp
@@ -29,6 +29,7 @@ limitations under the License.
 #include <type_traits>
 #include <vector>
 
+#include "llvm/ADT/StringRef.h"
 #include "llvm/ADT/bit.h"
 #include "llvm/Support/Errc.h"
 #include "llvm/Support/Error.h"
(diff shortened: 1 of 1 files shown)


EylonKrause:

abhigunj (approved):

PR #2965, opened 2026-07-06
reference: order convolution result non-spatial dims as [batch, feature]

EylonKrause:
### Problem
`convolutionOp` in the reference interpreter can crash (or silently produce a wrong result) for a valid convolution whose output layout places the feature dimension before the batch dimension.

### Root cause
`resultNonSpatialDims` is built by scanning the output dimensions in ascending index order:
```cpp
Sizes resultNonSpatialDims;
for (auto i = 0; i < result.getRank(); ++i)
  if (llvm::find(outputSpatialDimensions, i) == outputSpatialDimensions.end())
    resultNonSpatialDims.push_back(result.getShape()[i]);
```
But the resulting index space is consumed as `[0]` = batch and `[1]` = feature: `concatAndPermute((*resultNonSpatialIt)[0], ..., (*resultNonSpatialIt)[1], resultPermutation)` with `resultPermutation = [outputBatchDimension, spatials..., outputFeatureDimension]`. When `outputFeatureDimension < outputBatchDimension`, the ascending scan yields `[feature_size, batch_size]`, so the batch and feature coordinates are swapped — a wrong result, and an out-of-bounds `result.set` (fatal error) when the two sizes differ. StableHLO permits an arbitrary output layout, so this is a valid, verifier-passing convolution.

### Fix
Build `resultNonSpatialDims` in `[batch, feature]` order explicitly.

### Verification
Static reasoning. (No local MLIR/stablehlo build set up — happy to add an interpreter test with an output layout whose feature dim precedes batch.)

### DIFF
--- a/stablehlo/reference/Ops.cpp
+++ b/stablehlo/reference/Ops.cpp
@@ -1662,11 +1662,14 @@ Tensor convolutionOp(
         dotGeneralOp(reversedLhsWindow, rhs, lhsContractingDimensions,
                      rhsContractingDimensions, resultType.getElementType());
 
+    // resultNonSpatialDims is consumed below as [0] = batch, [1] = feature (via
+    // concatAndPermute with resultPermutation), so it must be built in that
+    // order rather than by ascending dimension index -- otherwise an output
+    // layout whose feature dimension precedes its batch dimension swaps the two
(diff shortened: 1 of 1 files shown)


EylonKrause:

PR #2966, opened 2026-07-06
Guard against a zero-group divide in collective group-size verification

EylonKrause:
### Problem
Collective group verification can divide by zero on an empty `replica_groups`.

### Root cause
`verifyReplicaGroups` (TypeInference.cpp) checks that all groups have the expected size with
```cpp
replicaIds.size() / replicaGroupType.getShape()[0]
```
where `getShape()[0]` is the number of groups (rows). An empty rank-2 `replica_groups` is only rejected up front when `use_global_device_ids` is set, so an op that doesn't set it (e.g. `all_to_all`, with `allGroupsMustHaveSameSize` and an `expectedGroupSize`) reaches this line with zero rows → division by zero.

### Fix
Skip the check when there are no groups (`getShape()[0] != 0`).

### Verification
Static reasoning. (Happy to add a verifier test with an empty `replica_groups` on `all_to_all`.)

### DIFF
--- a/stablehlo/dialect/TypeInference.cpp
+++ b/stablehlo/dialect/TypeInference.cpp
@@ -912,6 +912,7 @@ LogicalResult verifyReplicaGroups(std::optional<Location> location,
 
   // all_to_all_c8
   if (allGroupsMustHaveSameSize && expectedGroupSize &&
+      replicaGroupType.getShape()[0] != 0 &&
       (replicaIds.size() / replicaGroupType.getShape()[0] !=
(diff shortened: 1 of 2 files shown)


EylonKrause:

EylonKrause:
Closing: this site was fixed on `main` by #2968 (the empty `replica_groups` case is now rejected explicitly before the division), so this PR is superseded. The `triangular_solve` batched-diagonal fix that was stacked underneath it remains up as #2964.

PR #2967, opened 2026-07-06
VhloBytecode: return a parse failure instead of aborting on a bad type

EylonKrause:
### Problem
Deserializing malformed VHLO bytecode can abort the whole process instead of returning a parse error.

### Root cause
`readFloatV1Attr` and `readIntegerV1Attr` pass the decoded element `type` straight to `getFloatSemantics` / `getBitWidthForIntegerType`, both of which call `llvm::report_fatal_error(...)` for any type that isn't a float / integer type. The type code comes from the (untrusted) bytecode via `reader.readType`, and no verifier runs between the decode and these calls, so a blob whose `kFloatV1Attr` / `kIntegerV1Attr` embeds e.g. a token or a mismatched integer/float type aborts the deserializer (`stablehlo-translate --deserialize`).

### Fix
Make `getFloatSemantics` return `const llvm::fltSemantics*` (`nullptr` on an unsupported type) and `getBitWidthForIntegerType` return `FailureOr<unsigned>`, and have the two readers return an empty attribute on failure — matching the file's other decode-failure paths (e.g. the `readType`-failure early return just above).

### Verification
Static reasoning; this is a robustness / DoS-on-malformed-input fix (a controlled abort, not memory corruption). No local MLIR build — happy to add a deserialize test with a crafted type code.

### DIFF
--- a/stablehlo/dialect/VhloBytecode.cpp
+++ b/stablehlo/dialect/VhloBytecode.cpp
@@ -812,24 +812,24 @@ void VhloBytecodeInterface::write(FftTypeV1Attr attr,
 
 namespace {
 /// Returns the floating semantics for the given type.
-const llvm::fltSemantics& getFloatSemantics(Type type) {
-  if (isa<FloatBF16V1Type>(type)) return APFloat::BFloat();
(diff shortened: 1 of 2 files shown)


EylonKrause:

PR #2968, opened 2026-07-06
Fix divide-by-zero verifying all_to_all with empty replica_groups

EylonKrause:
### Problem
Verifying a `stablehlo.all_to_all` with an empty `replica_groups` (0 rows) crashes with an integer divide-by-zero (SIGFPE).

### Root cause
`verifyReplicaGroups` computes the `all_to_all_c8` group-size check as:
```cpp
if (allGroupsMustHaveSameSize && expectedGroupSize &&
    (replicaIds.size() / replicaGroupType.getShape()[0] != *expectedGroupSize))
```
The only empty-groups guard (a few lines above) is gated on `useGlobalDeviceIds`:
```cpp
if (useGlobalDeviceIds &&
    (replicaGroupType.getShape()[0] * replicaGroupType.getShape()[1] == 0))
  return emitOptionalError(..., "the replica groups cannot be empty");
```
`inferAllToAllOp` is the only caller that passes a non-null `expectedGroupSize` (`split_count`), and it does so with `useGlobalDeviceIds = false`. So for `stablehlo.all_to_all` with `replica_groups = dense<0> : tensor<0x2xi64>`, the empty guard is skipped and `replicaGroupType.getShape()[0]` is `0`, making the division `0 / 0` → SIGFPE during verification. `AllToAllOp` has no separate `verify()`, so this inference path is the sole verifier.

### Fix
Reject an empty `replica_groups` before the division (it cannot have the expected group size, since `split_count >= 1`), which also short-circuits the divide-by-zero.

### Test
`all_to_all_empty_replica_groups` in `tests/ops_stablehlo.mlir` checks that an empty `replica_groups` now produces the `group size of replica_groups must be 4` diagnostic instead of crashing.


EylonKrause:

abhigunj (approved):

PR #2971, opened 2026-07-07
Guard against integer division-by-zero in stablehlo transforms

EylonKrause:
### Problem
Three transforms crash the compiler with an integer divide-by-zero on verifier-valid IR:
1. `RefineReduceScatterOpPattern` divides the scatter dimension by the shard count without checking it is nonzero — an empty-column `replica_groups` (shard count 0) crashes the shape-refinement pass.
2. `FoldDivOpPattern` folds `stablehlo.divide` on integer constants via `APInt::udiv/sdiv` with no zero-divisor guard — `divide(x, 0)` crashes the folder.
3. `FoldRemOpPattern` has the identical issue via `APInt::urem/srem`.

### Details
- **reduce_scatter refinement** (`StablehloRefineShapes.cpp`): `shardCount` comes from `denseGroups.getType().getDimSize(1)` (the `replica_groups` column count), which is 0 for `tensor<Nx0xi64>`. `verifyReplicaGroups` only rejects an empty `replica_groups` when `use_global_device_ids` is set, so a `reduce_scatter` with no `channel_handle`, `use_global_device_ids = false`, and `replica_groups = dense<0> : tensor<1x0xi64>` verifies, and then `refinement[scatterDim] /= 0` is a SIGFPE. The sibling mesh-axes branch already guards its empty case.
- **divide/remainder folders** (`StablehloAggressiveFolder.cpp`): the folders apply `APInt::udiv/sdiv/urem/srem` element-wise with no zero-divisor check; `APInt` asserts in debug and executes a hardware divide-by-zero (SIGFPE) in release. `stablehlo.divide %a, %z` / `stablehlo.remainder %a, %z` with `%z = dense<0>` integer constant is verifier-valid and crashes the folder. (Upstream `arith::DivSIOp::fold` guards the same case.)

### Fix
- reduce_scatter: bail (`notifyMatchFailure`) when the shard count is 0.
- folders: decline to fold when the integer divisor constant has any zero element (a shared `integerDivisorHasZeroElement` helper), leaving the op in place. The float path is unaffected (division by zero yields inf/nan).

### Tests
Added lit tests: `@div_no_fold_int_zero` / `@rem_no_fold_int_zero` (integer divide/remainder by a zero constant is not folded) and `@refine_reduce_scatter_empty_replica_groups` (empty `replica_groups` leaves the result unrefined). Each previously crashed the pass.

---

### DIFF
--- a/stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp
+++ b/stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp
@@ -1000,6 +1000,20 @@ struct FoldConvertOpPattern : public ShapeOpRewritePattern<ConvertOp> {
   }
 };
 
+// Returns true if the divisor (second operand) of `op` is a constant integer
+// tensor with any zero element. Integer udiv/sdiv/urem/srem by zero would crash
+// the constant folder (APInt asserts in debug, SIGFPE in release), so such ops
+// must not be folded.
+static bool integerDivisorHasZeroElement(Operation* op) {
+  TypedAttr rhsAttr;
+  matchPattern(op->getOperand(1), m_Constant(&rhsAttr));
+  auto denseRhs = dyn_cast_or_null<DenseIntElementsAttr>(rhsAttr);
+  if (!denseRhs) return false;
+  for (const APInt& value : denseRhs.getValues<APInt>())
+    if (value.isZero()) return true;
+  return false;
+}
+
 struct FoldDivOpPattern : public ShapeOpRewritePattern<DivOp> {
   using ShapeOpRewritePattern::ShapeOpRewritePattern;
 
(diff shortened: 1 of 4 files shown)


stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp
```diff
@@ -1000,6 +1000,20 @@ struct FoldConvertOpPattern : public ShapeOpRewritePattern<ConvertOp> {
   }
 };
 
+// Returns true if the divisor (second operand) of `op` is a constant integer
+// tensor with any zero element. Integer udiv/sdiv/urem/srem by zero would crash
+// the constant folder (APInt asserts in debug, SIGFPE in release), so such ops
+// must not be folded.
+static bool integerDivisorHasZeroElement(Operation* op) {
```
GleasonK, line 1007: nit for file consistency, these transforms tend to use anonymous namespace without `static`

GleasonK (approved):
Thanks for the fix with nice unit tests!

stablehlo/transforms/optimization/StablehloAggressiveFolder.cpp
GleasonK: Will fix separately, CI looks clean. Safe to land.

PR #2977, opened 2026-07-15
NFC: Use generic DenseMapInfo for NonSpatialDim

edmundshieh:
## Summary

- Remove the local `DenseMapInfoNonSpatialDim` implementation.
- Let `SmallDenseMap` use LLVM's generic enum `DenseMapInfo` specialization.
- Preserve convolution parser behavior while improving compatibility across LLVM `DenseMap` implementations.

## Motivation

The local `DenseMapInfoNonSpatialDim` predates LLVM's generic enum `DenseMapInfo` specialization. It originally implemented hashing, equality, and the empty and tombstone sentinel keys required by LLVM's `DenseMap` at the time.

StableHLO's pinned LLVM uses a [sentinel-free `DenseMap`](https://github.com/llvm/llvm-project/blob/b713aaedb6eedd7d20f666038d945e50eafd4c27/llvm/include/llvm/ADT/DenseMap.h#L12-L15) that tracks bucket occupancy separately. The recent LLVM integration adapted the local trait by removing `getEmptyKey()` and `getTombstoneKey()`, leaving only the hashing and equality operations needed by that implementation. This makes the local trait specific to the sentinel-free `DenseMap` contract and incompatible with LLVM configurations that still require sentinel keys.

LLVM provides a [generic enum `DenseMapInfo` specialization](https://github.com/llvm/llvm-project/blob/b713aaedb6eedd7d20f666038d945e50eafd4c27/llvm/include/llvm/ADT/DenseMapInfo.h#L140-L150) that delegates to the enum's underlying type. Because `NonSpatialDim` has a fixed `int64_t` underlying type, omitting the custom `SmallDenseMap` key-info argument selects `DenseMapInfo<NonSpatialDim>`, which in turn delegates to `DenseMapInfo<int64_t>`. This keeps the map and its key-info policy aligned with the LLVM version being used.

The hashing, equality, and convolution parser behavior are unchanged.

## Testing

- Built `stablehlo-opt` against an LLVM configuration using sentinel-key `DenseMap` traits.
- Focused `llvm-lit` coverage: `ops_stablehlo.mlir`, `print_stablehlo.mlir`, and `verify_convolution.mlir` (3/3 passed).
- Full StableHLO test suite (272/272 passed).
- `clang-format --dry-run --Werror stablehlo/dialect/StablehloOps.cpp`
- `git diff --check`


abhigunj (approved):

abhigunj:
LGTM, thank you for the contribution!

PR #2978, opened 2026-07-16
Preserve memory effects during call refinement

edmundshieh:
## Summary

- compute and cache transitive memory-effect summaries before shape refinement
- erase fully constant calls only when their direct and transitive callees are memory-effect-free
- retain calls that may have memory effects while propagating each constant integer-tensor result to its users
- keep token and nonconstant results connected to the retained call
- preserve call attributes when specializing global constant shape arguments
- retain existing constant folding through memory-effect-free call chains

This supersedes #2934 by covering nested func.call chains and caching function summaries.

## Testing

- focused stablehlo_refine_shapes.mlir lit test (1/1 passed)
- full StableHLO lit suite (272/272 passed)
- ./build_tools/github_actions/lint_clang_format.sh -b upstream/main
- ./build_tools/github_actions/lint_whitespace_checks.sh -b upstream/main
- ./build_tools/github_actions/lint_buildifier.sh


stablehlo/tests/transforms/stablehlo_refine_shapes.mlir
```diff
@@ -885,6 +885,169 @@ module @refine_call_dimension_argument_not_integer {
...
+
+// CHECK-LABEL: module @refine_call_token_and_constant_results
+module @refine_call_token_and_constant_results {
+  // CHECK: func.func public @main() -> tensor<4xi32>
+  func.func public @main() -> tensor<?xi32> {
+    // CHECK-DAG: %[[TOKEN:.*]] = stablehlo.create_token
+    // CHECK-DAG: %[[SHAPE:.*]] = stablehlo.constant dense<4> : tensor<1xi64>
+    // CHECK: %[[CALL:.*]]:2 = call @callee(%[[TOKEN]])
+    // CHECK-SAME: (!stablehlo.token) -> (!stablehlo.token, tensor<1xi64>)
+    // CHECK-NOT: call @callee
```
mrguenther: It looks like you're checking that `@callee` is called once and only once. Is there a reason to suppose that it could plausibly be called multiple times? It's only called once in the input MLIR, so it would be pretty odd if the optimizations emitted a duplicate call. Do we need the `CHECK-NOT` line?

edmundshieh: Thanks for catching. Removed now as it's not needed.

stablehlo/tests/transforms/stablehlo_refine_shapes.mlir
```diff
@@ -885,6 +885,166 @@ module @refine_call_dimension_argument_not_integer {
...
+    // CHECK: return %[[RESULT]] : tensor<4xi32>
+    %0 = call @callee() : () -> tensor<?xi64>
+    %1 = stablehlo.convert %0 : (tensor<?xi64>) -> tensor<1xi64>
+    %2 = stablehlo.dynamic_iota %1, dim = 0 : (tensor<1xi64>) -> tensor<?xi32>
+    return %2 : tensor<?xi32>
+  }
+  func.func private @callee() -> tensor<?xi64> {
+    // Transitive effects through func.call must prevent erasing the call.
+    // CHECK: call @effect()
+    call @effect() : () -> ()
```
mrguenther, line 980: This certainly works and is fine for now, but I think we might actually be able to relax this restriction slightly. Suppose the definition for a particular function `@foo` gets optimized down to a single `stablehlo.custom_call @bar` op with `{has_side_effect = true}`. Couldn't we safely inline all `func.call @foo` ops, replacing them with equivalent `stablehlo.custom_call @bar` ops?

I think, generally, we can't remove calls with transitive side effects _unless_ we inline the ops responsible for those transitive side effects. IDK how often that exception is relevant, and it's probably fairly low-impact when it does come up, but it's worth considering for later.

This is just a note for us to consider later; no action required right now.

stablehlo/tests/transforms/stablehlo_refine_shapes.mlir
```diff
@@ -885,6 +885,166 @@ module @refine_call_dimension_argument_not_integer {
...
+  // CHECK-NEXT: return %[[RESULT]] : tensor<4xi32>
+  // CHECK-NEXT: }
+  func.func public @main() -> tensor<?xi32> {
+    %0 = call @callee() : () -> tensor<1xi64>
+    %1 = stablehlo.dynamic_iota %0, dim = 0 : (tensor<1xi64>) -> tensor<?xi32>
+    return %1 : tensor<?xi32>
+  }
+  func.func private @callee() -> tensor<1xi64> {
+    %0 = stablehlo.constant dense<0.> : tensor<f32>
+    // CHECK: call @identity(%{{.*}})
```
mrguenther: Why do we care if the pure call op is preserved in the `@callee` body?

stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -370,14 +372,74 @@ class RefinementKey {
...
+
+ private:
+  enum class State { Visiting, EffectFree, MayHaveEffects };
+
+  bool compute(func::FuncOp func) {
+    auto [it, inserted] = functionEffects.try_emplace(func, State::Visiting);
+    if (!inserted) return it->second == State::EffectFree;
+
+    bool effectFree = !func.isExternal() && isMemoryEffectFree(func.getBody());
+    functionEffects.find(func)->second =
```
mrguenther, line 401: Can't `functionEffects.find(func)` be replaced with `it`?

mrguenther: Or are you concerned that the call to `isMemoryEffectFree` on the prior line will invalidate the iterator?

mrguenther: Actually, according to cppreference, `try_emplace` doesn't invalidate iterators or references. I think it's safe to reuse the existing iterator here despite the intervening recursive call.

stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -370,14 +372,74 @@ class RefinementKey {
...
+        if (!isMemoryEffectFree(&op)) return false;
+    return true;
+  }
+
+  bool isMemoryEffectFree(Operation* op) {
+    if (auto call = dyn_cast<func::CallOp>(op)) {
+      auto callee = symbolTables.lookupNearestSymbolFrom<func::FuncOp>(
+          call, call.getCalleeAttr());
+      return callee && compute(callee);
+    }
```
mrguenther: I think this could be generalized slightly by using `mlir::CallOpInterface` instead of `func::CallOp`:
```cpp
if (auto call = dyn_cast<mlir::CallOpInterface>(op)) {
  auto callee = dyn_cast_or_null<func::FuncOp>(
      call.resolveCallableInTable(&symbolTables));
  return callee && compute(callee);
}
```

stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -370,14 +372,74 @@ class RefinementKey {
...
+
+  bool isMemoryEffectFree(func::FuncOp func) const {
+    auto it = functionEffects.find(func);
+    return it != functionEffects.end() && it->second == State::EffectFree;
+  }
+
+ private:
+  enum class State { Visiting, EffectFree, MayHaveEffects };
+
+  bool compute(func::FuncOp func) {
```
mrguenther: nit: Could we call this function something a bit more specific?

stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -600,10 +675,57 @@ struct RefineCallOpPattern : public OpRewritePattern<func::CallOp> {
...
+    // propagate each used constant result so users can still benefit from
+    // shape refinement.
+    bool hasUsedConstantResult = llvm::any_of(
+        llvm::zip(op.getResults(), constantResults), [](auto resultAndAttr) {
+          return !std::get<0>(resultAndAttr).use_empty() &&
+                 std::get<1>(resultAndAttr).has_value();
+        });
+    if (hasUsedConstantResult) {
+      rewriter.setInsertionPoint(op);
+      Operation* retainedCall = rewriter.clone(*op);
```
mrguenther: Is there a reason why you're cloning `op` and then deleting the original? Wouldn't it be simpler to just insert the propagated constants either before or after `op` without touching `op` itself?

stablehlo/transforms/StablehloRefineShapes.cpp
```diff
@@ -370,14 +372,74 @@ class RefinementKey {
   SmallVector<Type> functionalArgumentTypes;
 };
 
+// Computes and caches whether functions may have memory effects, including
+// effects reached transitively through func.call operations.
+class FunctionEffectAnalysis {
```
mrguenther, line 378: Thanks for implementing this! @GleasonK and I have been aware of this limitation for a while and have discussed implementing a solution along these lines, but we didn't know when we'd have time to prioritize it. Much appreciated!

mrguenther (commented):
This looks great; we've wanted to address this for a while but haven't had time to prioritize it. Thanks!

mrguenther:
At some point, we might want to look into whether MLIR's DataFlow Analysis feature (https://mlir.llvm.org/docs/Tutorials/DataFlowAnalysis/, https://mlir.llvm.org/doxygen/namespacemlir_1_1dataflow.html) would be of any use here, e.g. in case it helped us generalize the API or reuse the computation between passes. I'm not sure whether it would provide any benefits here or not, but I've been meaning to look into it. Anyway, we can always look into that later; this looks great as-is.

stablehlo/tests/transforms/stablehlo_refine_shapes.mlir
edmundshieh: Ah yeah we don't need this, removed.

stablehlo/transforms/StablehloRefineShapes.cpp
edmundshieh: The concern is not the initial `try_emplace`, but subsequent insertions during recursive callee analysis. The cppreference guarantee applies to standard-library containers, whereas this is `llvm::DenseMap`. Recursive analysis may insert into the same map; `DenseMap` treats insertion as iterator-invalidating, and growth can relocate its bucket storage. Therefore the saved iterator may be invalid when recursion returns, so the second lookup is intentional.

stablehlo/transforms/StablehloRefineShapes.cpp
edmundshieh: Switched the analysis to `CallOpInterface::resolveCallableInTable`. Resolved `func::FuncOp` callees are analyzed transitively, while unresolved and indirect callees remain conservative.

stablehlo/transforms/StablehloRefineShapes.cpp
edmundshieh: Renamed it to `computeFunctionEffects`.

stablehlo/transforms/StablehloRefineShapes.cpp
edmundshieh: Ah yup, you're right. The clone was only being used to retain the call while replacing selected results. The implementation now leaves the original call in place and replaces only the uses of results proven constant.

stablehlo/transforms/StablehloRefineShapes.cpp
edmundshieh: Glad this addresses a known gap. Thanks for the careful review.

mrguenther (approved):

PR #2979, opened 2026-07-19
Support loading interpreter arguments from a file in stablehlo-translate

hamzaqureshi5:
Fixes #2912.

`--args` currently requires spelling out every input tensor inline, which
exceeds OS argument-length limits for realistically sized models. This PR
adds an IREE-style `@` prefix for reading arguments from files, in two
commits:

## 1. `--args=@file`

The file contains the same array-attribute text as the inline form:

    stablehlo-translate model.mlir --interpret --args=@inputs.mlir

## 2. `--args=@a.npy,@b.npy`

Comma-separated NumPy files, each providing one input tensor. The tensor
type is inferred from the file's header, so probe outputs
(`--probe-output-dir`) can be fed back in as interpreter inputs directly:

    stablehlo-translate model.mlir --interpret --args=@probe1.npy,@probe2.npy

Type inference covers little-endian integer and f32/f64 dtypes. Types
without a native NumPy representation (bool, f16, bf16, complex) are
serialized by `serializeTensor` under substitute descrs, so they can't be
unambiguously inferred and are rejected with a clear error. Extending
these is left for a follow-up.

## NumPy reader fixes

Wiring up inference exposed three latent bugs in `NumPy.cpp`, previously
unreachable because the parsed shape was never used. Note the last one
changes the `descr` that probe files are written with:

- `parseShapeHeader` returned an `ArrayRef` to a local vector and
  pre-filled four zero dims; it now returns an owned, correctly-bounded
  vector, and the file's shape is validated against the expected type.
- Header sizes above 127 bytes were sign-extended into negative values.
- The i64/ui64 type dispatch was inverted relative to every other width,
  so signless i64 tensors serialized with a `'u8'` descr and vice versa.
  Write and read stay consistent with each other, as before.

## Testing

- `stablehlo/tests/interpret/api_input_arguments.mlir`: `@file` positive
  and missing-file cases.
- `stablehlo/tests/interpret/api_input_arguments_npy.mlir`: writes the
  arguments to `.npy` via `interpreter.probe`, then loads them back as
  `--args` inputs (round-trips the serializer/reader pair); plus
  missing-file and mixed-file error cases.
- Manually verified against externally generated NumPy files (f32 2x2,
  `'|i1'` int8, `'<i8'` int64, and bool → rejected as unsupported).
- `ninja check-stablehlo-tests` passes.

### DIFF
--- a/stablehlo/tests/interpret/api_input_arguments.mlir
+++ b/stablehlo/tests/interpret/api_input_arguments.mlir
@@ -1,10 +1,16 @@
 // RUN: stablehlo-translate %s --interpret --args="[dense<1> : tensor<2xi32>, dense<2> : tensor<2xi32>]" | FileCheck %s
 
+// RUN: echo "[dense<1> : tensor<2xi32>, dense<2> : tensor<2xi32>]" > %t.args
+// RUN: stablehlo-translate %s --interpret --args=@%t.args | FileCheck %s
+
 // RUN: not stablehlo-translate %s --interpret --args="not_array" 2>&1 | FileCheck %s --check-prefixes=CHECK-ERROR-NOT-ARRAY
-// CHECK-ERROR-NOT-ARRAY: expectected array attribute string for args, i.e. --args=[dense<1> : tensor<2xi32>, ...]
+// CHECK-ERROR-NOT-ARRAY: expectected array attribute string for args, i.e. --args=[dense<1> : tensor<2xi32>, ...] or --args=@file
 
 // RUN: not stablehlo-translate %s --interpret --args="[4.0 : f32]" 2>&1 | FileCheck %s --check-prefixes=CHECK-ERROR-NOT-DENSE
-// CHECK-ERROR-NOT-DENSE: expected dense elements attribute for args elements, i.e. --args=[dense<1> : tensor<2xi32>, ...]
+// CHECK-ERROR-NOT-DENSE: expected dense elements attribute for args elements, i.e. --args=[dense<1> : tensor<2xi32>, ...] or --args=@file
+
+// RUN: not stablehlo-translate %s --interpret --args=@%t.does_not_exist 2>&1 | FileCheck %s --check-prefixes=CHECK-ERROR-NO-FILE
+// CHECK-ERROR-NO-FILE: failed to read args file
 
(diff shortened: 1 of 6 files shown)


hamzaqureshi5:
@GleasonK PR is ready for review

stablehlo/tests/interpret/api_input_arguments.mlir
```diff
@@ -1,10 +1,16 @@
 // RUN: stablehlo-translate %s --interpret --args="[dense<1> : tensor<2xi32>, dense<2> : tensor<2xi32>]" | FileCheck %s
 
+// RUN: echo "[dense<1> : tensor<2xi32>, dense<2> : tensor<2xi32>]" > %t.args
+// RUN: stablehlo-translate %s --interpret --args=@%t.args | FileCheck %s
```
abhigunj: keep the existing test unchanged add new test file `api_input_arguments_from_file.mlir`

stablehlo/tools/StablehloTranslateMain.cpp
```diff
@@ -102,14 +108,43 @@ stablehlo::Tensor makeBooleanTensor(MLIRContext *context, bool value) {
...
 mlir::FailureOr<SmallVector<stablehlo::InterpreterValue>>
 parseInterpreterArguments(std::string argsStr, MLIRContext *context) {
   llvm::SmallVector<stablehlo::InterpreterValue> inputs;
   auto parseError = [&](llvm::StringRef msg) {
-    std::string usage = "--args=[dense<1> : tensor<2xi32>, ...]";
+    std::string usage =
+        "--args=[dense<1> : tensor<2xi32>, ...] or --args=@file";
     return emitError(UnknownLoc::get(context), msg) << ", i.e. " << usage;
   };
+  if (!argsStr.empty() && argsStr[0] == '@') {
```
abhigunj, line 128: this logic will ignore a.npy if the input is "@a.npy,b.txt". Change it to check if all the input files has .npy extension. Otherwise return an error message.. cannot mix .npy and non-.npy args files

abhigunj (approved):

abhigunj:
LGTM, lint check is failing, fix that. ready to merge.

Issue #1794, opened 2023-10-06
Support f16 types in fft


hawkinsp:
### What happened?

FFTs with 16-bit types aren't permitted by the stablehlo verifier, which currently only allows f32 and f64 types (and their complex versions).

`stablehlo.fft` should be generalized to allow any floating point type as input (or complex floating point type, where applicable).

(XLA does not support a complex32 type, yet, but whether XLA supports this or not has no bearing on whether stablehlo should allow it.)

PR #2993, opened 2026-08-21
[RFC] Support f16- and bf16-based complex types in StableHLO

DeanTMaxim:
## Summary

This RFC proposes extending StableHLO's supported complex component types from
f32 and f64 to the closed set `{f16, bf16, f32, f64}`.

The proposal keeps one uniform StableHLO complex type domain based on MLIR's
builtin `ComplexType`. Every existing StableHLO operation whose specified type
domain already admits complex values gains f16- and bf16-based instantiations;
the RFC does not introduce FFT-only complex types or operation-specific
exceptions.

## Compatibility model

The RFC proposes a new `ComplexV2` at the feature's StableHLO minor version.
Current StableHLO complex values, including f32- and f64-based values, map to
V2. Historical `ComplexV1<f32/f64>` values upgrade to V2, and V2 f32/f64 values
can downgrade to V1 for old targets. V2 f16/bf16 values have no old-version
downgrade and must fail with a diagnostic.

The existing shallow `ComplexV1` parser and verifier are not tightened.
Nonhistorical V1 component combinations remain structurally representable but
must be rejected by version conversion and must never be emitted as portable
old-version artifacts.

## Review focus

The main decisions requested from reviewers are:

1. Use the explicit component set `{f16, bf16, f32, f64}` rather than bf16-only
   support or all of `HLO_Float`.
2. Apply the new component types uniformly to every existing StableHLO type
   domain that already supports complex values.
3. Introduce `ComplexV2` and target-aware recursive type conversion instead of
   retroactively expanding `ComplexV1`.
4. Require compatibility-atomic implementation across the specification, ODS,
   custom verification and inference, VHLO conversion, and tests.

## Scope

This is an RFC-only PR. It contains no implementation changes and does not
claim execution support in XLA, JAX, or any backend. It also does not define a
physical memory ABI, automatic promotion to `complex<f32>`, or support for
FP4, FP6, FP8, E8M0, or future floating-point component types.

Implementation PRs will be sent only after the RFC is approved. Reference
interpreter work may be a separate StableHLO-only follow-up if needed, while
XLA, framework, legalization, and backend work remain separate.

## Related issue

Related to #1794. This RFC-only PR does not close the issue because the
requested functionality is not implemented by approving the design.

### DIFF
--- a/rfcs/20260821-low-precision-complex.md
+++ b/rfcs/20260821-low-precision-complex.md
@@ -3,7 +3,7 @@
 Status: In Review<br/>
 Initial version: 08/21/2026<br/>
 Last updated: 08/21/2026<br/>
-Discussion thread: TBD
+Discussion thread: [openxla/stablehlo#2993](https://github.com/openxla/stablehlo/pull/2993)
 
 ## Overview
 
--- a/rfcs/20260821-low-precision-complex.md
+++ b/rfcs/20260821-low-precision-complex.md
@@ -1,0 +1,505 @@
+# [RFC] Support f16- and bf16-based complex types in StableHLO
+
+Status: In Review<br/>
+Initial version: 08/21/2026<br/>
+Last updated: 08/21/2026<br/>
+Discussion thread: TBD
+
+## Overview
+
+This RFC proposes extending the set of complex types supported by StableHLO
+from:
+
+```text
+complex<f32>
+complex<f64>
+```
+
+to:
+
+```text
+complex<f16>
+complex<bf16>
+complex<f32>
+complex<f64>
+```
+
+The proposal uses the existing MLIR builtin `ComplexType`; it does not add a
+new StableHLO type. It extends the set of floating-point component types that
+StableHLO accepts inside `complex<T>`.
+
+This is a StableHLO type-system and serialization proposal. It does not require
+XLA, JAX, or any particular backend to execute the new types, and it does not
+promise native kernels, a memory ABI, or improved performance.
+
+## Motivation
+
+[openxla/stablehlo#1794](https://github.com/openxla/stablehlo/issues/1794)
+requests support for 16-bit FFT input and output types. Today, the StableHLO
+verifier prevents such programs even though MLIR can represent both
+`complex<f16>` and `complex<bf16>`.
+
+There are two independent concerns:
+
+1. Whether StableHLO can represent and specify a low-precision complex value.
+2. Whether a particular consumer or backend can execute that value efficiently.
+
+StableHLO should decide the first concern based on the coherence of its type
+system and semantics rather than on the current implementation status of one
+consumer. Consumers that do not support the new types may continue to reject
+them with a diagnostic until they add a lowering, legalization, or native
+implementation.
+
+The body of #1794 suggests allowing any floating-point component type. This
+RFC intentionally narrows that suggestion to f16 and bf16, the two widely used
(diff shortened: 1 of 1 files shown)


hawkinsp:
I'm supportive of adding these types.

mrguenther:
Are you aware of any backends that either support c16 data types or plan to add support for them? If so, I'm in favor. If not, it's more of a grey area.

I do see you mentioned this in openxla/xla#6136. Does that mean you're working on (or seeking approval to work on) implementing backend support yourself?

DeanTMaxim:
cuFFT already supports FP16 and BF16 complex FFTs (`CUDA_C_16F` and `CUDA_C_16BF`) on supported NVIDIA GPUs, although XLA does not expose these types or backend paths today.

And yes: in openxla/xla#6136, I’m seeking approval to implement the XLA core type support myself. If that direction is accepted, I also intend to pursue NVIDIA GPU/cuFFT enablement as a separate follow-up, subject to maintainer agreement.

pifon2a:
Is there a plan to add these new element types to HLO as well? or only stableHLO?

hawkinsp:
> Is there a plan to add these new element types to HLO as well? or only stableHLO?

Sounds like openxla/xla#6136 is exactly that proposal.

mrguenther:
Thanks for clarifying/confirming. I personally think it sounds like a good idea in that case. I'd like to run it by @GleasonK, though I believe he's out of office today.

GleasonK:
> The RFC proposes a new ComplexV2 at the feature's StableHLO minor version.

I wonder if we can do something like [VHLO_VersionedOpConstraintInterface](https://github.com/openxla/stablehlo/blob/639335932274930617d9724fb25ec9f2f52c5e40/stablehlo/dialect/VhloOps.td#L38) for types, then we don't need complex up/downgrade logic just a verifier that checks the requested version and either errors or doesn't pending the complex types element type.

> Sounds like https://github.com/openxla/xla/issues/6136 is exactly that proposal.

This proposal talks about the element type, as long as the RFC author's plan is to implement actual support for some backend then I'm on board. But would rather we not carry around a type representation in both StableHLO/XLA that no backend can emit code for.

Also adding to StableHLO first makes for a strange compatibility story - StableHLO tends to gate compiler features, meaning we fail at the serialization layer for forward incompat, this pushes compat a layer down. I.e. if we add support for this later in XLA, stablehlo will downgrade successfully, but compiler may or may not support it.

GleasonK (approved):

DeanTMaxim:
I’ve revised the RFC to address all three points.

For the type-level constraint interface, I kept `ComplexV2`. A target-aware check would make `ComplexV1`’s valid domain depend on an external target-version predicate, conflicting with VHLO’s add-only model. The parser could still accept nonhistorical `ComplexV1<f16>` values, requiring side-condition checks at every nested type location. Such a check may still help with recursive legality checks, but it does not replace the V1-to-V2 boundary (see “Reuse `ComplexV1` with target-aware validation”). Could you confirm `ComplexV2` as the direction before implementation?

Execution support is now a coordinated sequence with [XLA #6136](https://github.com/openxla/xla/issues/6136): first agree on `C32`/`BC32` and the unsupported-execution boundary, then add the bridge, legalization, and backend support, initially targeting NVIDIA cuFFT. The compatibility rule is explicit: StableHLO serialization proves representability only; a bridge without `C32`/`BC32` support must reject these types before constructing the XLA/HLO module, so serialization cannot defer failure until execution.

PR #2994, opened 2026-08-21
Integrate LLVM at llvm/llvm-project@fdb39b1112a6

srishti-pm:


abhigunj (approved):

.bazelversion
```diff
@@ -1 +1 @@
-7.7.0
+8.7.0
```
abhigunj: it seems new bazel version disables WORKSPACE by default.

add this to .bazelrc and see if it fixes the bazel build failure. 

explicitly enable WORKSPACE since Bazel 8.X.X disables it by default.
common --enable_workspace

abhigunj: as discuss, suggested to drop the bazelversion upgrade change, we will support it later based on priority.

.bazelrc
```diff
@@ -15,6 +15,10 @@
 # Disable Bzlmod for now
 common --noenable_bzlmod
 
+# Enable assertions (matches CMake -DLLVM_ENABLE_ASSERTIONS=ON)
```
abhigunj: rewrite the comment as
"Ensure NDEBUG is undefined so the verifier runs in opt builds"

abhigunj (approved):

srishti-pm:
Could someone please submit this PR? I don't have the permissions to do so myself.

PR #2995, opened 2026-08-22
Fix sort comparison type selection

GodlyDonuts:
TensorFlow’s `tf.RandomShuffle` lowering exposed a mismatch in `createSortOp`: the helper compares only the first operand (the sort key), but selected `TOTALORDER` when any payload type was floating-point. With `i32` keys and an `f32` payload, that produces an invalid integer comparison.

This change:
- selects `TOTALORDER` only when the first element type is floating-point;
- adds FileCheck regression coverage that invokes the helper with an integer key and float payload; and
- verifies that float keys still use `TOTALORDER`.

This is the upstream prerequisite for tensorflow/tensorflow#125279. The equivalent MHLO helper was corrected by openxla/xla#40531.

Tested: `bazel build //stablehlo/tests:test_utils`.

### DIFF
--- a/stablehlo/tests/TestUtils.cpp
+++ b/stablehlo/tests/TestUtils.cpp
@@ -94,6 +94,30 @@ struct BroadcastIfNeededPattern : public RewritePattern {
   }
 };
 
+struct CreateSortOpPattern : public RewritePattern {
+  explicit CreateSortOpPattern(MLIRContext* context)
+      : RewritePattern("hlo_test_create_sort.build", 1, context) {}
+
+  LogicalResult matchAndRewrite(Operation* op,
+                                PatternRewriter& rewriter) const override {
+    SmallVector<Value> operands = llvm::to_vector(op->getOperands());
+    SmallVector<Type> elementTypes;
+    elementTypes.reserve(operands.size());
+    for (Value operand : operands) {
+      auto shapedType = dyn_cast<ShapedType>(operand.getType());
+      if (!shapedType)
+        return rewriter.notifyMatchFailure(op, "expected shaped operands");
+      elementTypes.push_back(shapedType.getElementType());
+    }
+
+    auto sort = stablehlo::createSortOp(
+        &rewriter, op->getLoc(), operands, elementTypes, /*dimension=*/0,
+        /*isStable=*/false, stablehlo::ComparisonDirection::LT);
+    rewriter.replaceOp(op, sort.getResults());
+    return success();
+  }
+};
+
 struct InferReturnTypesPattern : public RewritePattern {
   explicit InferReturnTypesPattern(MLIRContext* context)
       : RewritePattern("hlo_test_infer.get_return_types", 1, context) {}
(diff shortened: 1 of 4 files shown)


stablehlo/dialect/StablehloOpsTest.cpp
```diff
@@ -0,0 +1,86 @@
+/* Copyright 2026 The OpenXLA Authors. All Rights Reserved.
```
abhigunj: we prefer FileCheck based tests. add a test for SortOp to stablehlo/tests/ops_stablehlo.mlir and remove this file.

abhigunj (commented):
Thanks for the contribution, LGTM.

stablehlo/dialect/StablehloOpsTest.cpp
GodlyDonuts: on it

GodlyDonuts: Updated: replaced the standalone C++ test with FileCheck coverage in `ops_stablehlo.mlir` and rebased onto current `main`.

GodlyDonuts:
> Thanks for the contribution, LGTM.

@abhigunj It should be good now!

stablehlo/tests/TestUtils.cpp
```diff
@@ -94,6 +94,30 @@ struct BroadcastIfNeededPattern : public RewritePattern {
   }
 };
 
+struct CreateSortOpPattern : public RewritePattern {
+  explicit CreateSortOpPattern(MLIRContext* context)
+      : RewritePattern("hlo_test_create_sort.build", 1, context) {}
+
+  LogicalResult matchAndRewrite(Operation* op,
+                                PatternRewriter& rewriter) const override {
+    SmallVector<Value> operands = llvm::to_vector(op->getOperands());
```
abhigunj, line 103: nit: iterate over the ValueRange directly and pass it without allocating memory

auto operands = op->getOperands(); // ValueRange
SmallVector<Type> elementTypes;
elementTypes.reserve(operands.size());
for (Value operand : operands) {
  auto shapedType = dyn_cast<ShapedType>(operand.getType());
  if (!shapedType)
    return rewriter.notifyMatchFailure(op, "expected shaped operands");
  elementTypes.push_back(shapedType.getElementType());
}

auto sort = stablehlo::createSortOp(
    &rewriter, op->getLoc(), operands, elementTypes, /*dimension=*/0,
    /*isStable=*/false, stablehlo::ComparisonDirection::LT);

abhigunj (approved):

abhigunj:
> > Thanks for the contribution, LGTM.
> 
> @abhigunj It should be good now!

One nit comment and approved. Thanks for the quick turnaround.

stablehlo/tests/TestUtils.cpp
abhigunj: stablehlo::createSortOp expects the operands to be passed as an ArrayRef. Okay to allocate memory using SmallVector.

PR #2996, opened 2026-08-24
Report StableHLO revision in tool --version

vabridgers:
stablehlo-opt, stablehlo-translate, and stablehlo-lsp-server report only the LLVM version in --version, so a built binary cannot report which StableHLO commit it came from.

Embed the source revision at build time with git describe and register a cl extra version printer. A release-tag build shows e.g. v1.19.0-0-g<hash> while an off-tag or modified tree shows a nonzero commit count or -dirty. STABLEHLO_APPEND_VC_REV (default ON) controls this; STABLEHLO_FORCE_VC_REVISION overrides it for builds without git metadata. The VHLO opset version is also printed on its own line.


vabridgers:
Hi @GleasonK - would you be able to review this PR? It adds StableHLO revision and VHLO opset info to --version (via AddExtraVersionPrinter) so we can track the provenance of our deployed StableHLO builds. It follows LLVM's LLVMRevision pattern and is off-switchable via STABLEHLO_APPEND_VC_REV. Let me know if the approach works for you. Thanks!

abhigunj:
CMake changes LGTM. We also uses Bazel, this PR doesn't add the same functionality to bazel build tools. 

Two ways to go about it, 

1. Match the Bazel behavior with CMake in this PR. OR

2. This PR can be CMAKE only change, then fix the Bazel Build failures by adding a genrule in BUILD.bazel to create a fallback version, similar to the CMake fallback. 

```
genrule(
    name = "stablehlo_revision_inc_gen",
    outs = ["StablehloRevision.h"],
    cmd = "echo '#undef STABLEHLO_REVISION' > $@",
)
```
and add other missing dependencies. If you are interested in contributing, happy to review, accept Bazel changes to correctly provide the version as a follow-up PR.

vabridgers:
> CMake changes LGTM. We also uses Bazel, this PR doesn't add the same functionality to bazel build tools.
> 

Thanks @abhigunj . I'll rework the PR and update.  Thanks!

stablehlo/tests/version.mlir
```diff
@@ -0,0 +1,7 @@
+// RUN: stablehlo-opt --version | FileCheck %s
+// RUN: stablehlo-translate --version | FileCheck %s
+
+// The embedded revision string varies per build and is not matched exactly.
```
abhigunj: remove this comment, not required.

abhigunj (approved):

vabridgers:
Thanks for the approval @abhigunj . Is there anything more I need to do here or is this just waiting for integration? Thanks

abhigunj:
> Thanks for the approval @abhigunj . Is there anything more I need to do here or is this just waiting for integration? Thanks

All good.  We are in the middle of the [migration](https://github.com/openxla/stablehlo/blob/main/rfcs/20260720-repo-sot-migration.md) this week, will merge the PR early next week.

PR #2997, opened 2026-08-24
Fix crash printing invalid convolution dimension numbers

vabridgers:
printConvolutionDimensions aborted via report_fatal_error when a ConvDimensionNumbersAttr had a dimension index outside [0, rank), which can happen for attributes built programmatically or parsed from the raw form. Instead of aborting, fall back to printing the raw form, like the MLIR printer falls back to the generic op form when verification fails. The raw output round-trips and valid attributes still print compressed.


abhigunj (approved):

vabridgers:
Thanks @abhigunj , please let me know if there's anything I need to do for this to merge. Looks to me like only maintainers can merge, but just checking to be certain. Thank you!

PR #3003, opened 2026-09-01
[RFC] Add WEAKORDER to ComparisonType and Deprecate SIGNED/UNSIGNED

akuegel:
This PR adds an RFC proposing to extend `ComparisonType` in StableHLO with `WEAKORDER` comparison ordering and deprecate the redundant `SIGNED` and `UNSIGNED` attributes.

### Summary

1. **Add `WEAKORDER` to `ComparisonType`**: Supports NumPy/Python total ordering semantics (-0.0 == +0.0, NaNs sorted to the end). This eliminates the 7+ op canonicalization boilerplate in frontend sort comparators (such as JAX) and avoids fragile compiler AST pattern matching.
2. **Deprecate `SIGNED` and `UNSIGNED`**: In StableHLO, operand types explicitly encode signedness (`si32`, `ui32`, `i1`), and integer comparisons are inherently total order. `SIGNED` and `UNSIGNED` are legacy artifacts of XLA internals and should be deprecated in favor of `NOTYPE` (or omitting the attribute).
3. **Alternatives Considered**: Evaluates replacing `compare_type` with `comparison_order` (PARTIAL, TOTAL, WEAK) to align with XLA's internal migration, but chooses in-place `ComparisonType` extension to avoid syntax/API churn for existing StableHLO users. This PR introduces the RFC markdown document in `rfcs/` for community review and feedback prior to implementation.

CC @gleasonk @phawkins

### DIFF
--- a/rfcs/20260901-numpy-comparison-type.md
+++ b/rfcs/20260901-numpy-comparison-type.md
@@ -1,8 +1,8 @@
-# [RFC] Add NUMPY to ComparisonType and Deprecate SIGNED/UNSIGNED
+# [RFC] Add WEAKORDER to ComparisonType and Deprecate SIGNED/UNSIGNED
 
 Status: In Review
 Initial version: 2026-09-01
-Last updated: 2026-09-01
+Last updated: 2026-09-03
 Discussion thread:
 [GitHub PR #3003](https://github.com/openxla/stablehlo/pull/3003)
 
@@ -10,19 +10,19 @@ Discussion thread:
 
 This RFC proposes:
 
-1. Adding `NUMPY` (or `FLOAT_NUMPY`) to `ComparisonType` (`compare_type`
-   attribute on `stablehlo.compare`).
-2. Formally deprecating the `SIGNED` and `UNSIGNED` enum values in
+1. Adding `WEAKORDER` to `ComparisonType` (`compare_type` attribute on
+   `stablehlo.compare`).
+3. Formally deprecating the `SIGNED` and `UNSIGNED` enum values in
    `ComparisonType`.
-3. Updating the StableHLO specification for `compare` to support
-   NumPy/Python-style total ordering semantics.
+4. Updating the StableHLO specification for `compare` to support strict weak
+   ordering semantics for floating-point types.
 
 ## Motivation & Background
 
-### 1. The Need for NUMPY Comparison Order
+### 1. The Need for WEAKORDER Comparison Order
 
-Machine learning frontends (JAX, PyTorch, NumPy) require NumPy sorting and
-ranking semantics:
+Machine learning frontends (JAX, PyTorch, NumPy) require array sorting and
+ranking semantics that differ from standard IEEE-754:
 
 - -0.0 and +0.0 are treated as equivalent keys (preserving stability in stable
   sorts).
@@ -41,7 +41,7 @@ detect this subgraph and emit fast CUB `DeviceRadixSort` (see b/376918731).
 If any compiler optimization alters the AST, pattern matching fails and falls
 back to a 10x-25x slower bitonic sort.
 
-Adding `NUMPY` directly to `ComparisonType` eliminates frontend
+Adding `WEAKORDER` directly to `ComparisonType` eliminates frontend
 canonicalization bloat and provides a clean, robust contract for compiler
 backends.
 
@@ -74,7 +74,7 @@ def STABLEHLO_COMPARISON_TYPE_SIGNED :
     I32EnumAttrCase<"SIGNED", 3>; // Deprecated
 def STABLEHLO_COMPARISON_TYPE_UNSIGNED :
     I32EnumAttrCase<"UNSIGNED", 4>; // Deprecated
-def STABLEHLO_COMPARISON_TYPE_NUMPY : I32EnumAttrCase<"NUMPY", 5>;
+def STABLEHLO_COMPARISON_TYPE_WEAK_ORDER : I32EnumAttrCase<"WEAKORDER", 5>;
 
 def StableHLO_ComparisonType : I32EnumAttr<"ComparisonType",
     "Which comparison type to use.",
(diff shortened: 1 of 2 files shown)


abhigunj:
Thank you for the RFC. Motivation and Overall LGTM. 

Do we have a better name for enum value and avoid using `NUMPY` reference. If it is not the standard (which seemes to the case), we are bringing framework library context into the StableHLO, framework agnostic layer.

abhigunj (approved):

abhigunj:
cc @hawkinsp  for visibility. Somehow I can't add as a reviewer.  LGTM comment will work!

rfcs/20260901-numpy-comparison-type.md
```diff
@@ -0,0 +1,134 @@
...
+1. Adding `NUMPY` (or `FLOAT_NUMPY`) to `ComparisonType` (`compare_type`
+   attribute on `stablehlo.compare`).
+2. Formally deprecating the `SIGNED` and `UNSIGNED` enum values in
+   `ComparisonType`.
+3. Updating the StableHLO specification for `compare` to support
+   NumPy/Python-style total ordering semantics.
+
+## Motivation & Background
+
+### 1. The Need for NUMPY Comparison Order
```
hawkinsp: One suggestion: don't call this "numpy" comparison, because that will get confusing if NumPy ever changes it...

hawkinsp (changes requested):
The change itself seems like a fine idea. Just a naming nit...

rfcs/20260901-numpy-comparison-type.md
akuegel: Thanks, that makes sense. I changed it to WEAKORDER now. Is that ok?

rfcs/20260901-numpy-comparison-type.md
```diff
@@ -0,0 +1,134 @@
...
+      STABLEHLO_COMPARISON_TYPE_NOTYPE,
+      STABLEHLO_COMPARISON_TYPE_FLOAT,
+      STABLEHLO_COMPARISON_TYPE_FLOAT_TOTAL_ORDER,
+      STABLEHLO_COMPARISON_TYPE_SIGNED,
+      STABLEHLO_COMPARISON_TYPE_UNSIGNED,
+      STABLEHLO_COMPARISON_TYPE_WEAK_ORDER
+    ]>
+```
+
+### 2. Specification (`docs/spec.md`)
```
abhigunj, line 91: update the spec.md with the exact change, include it in this PR and refer it at the RFC as "Proposed Specification Changes
" 
see similar example. https://github.com/openxla/stablehlo/blob/main/rfcs/20240312-standardize-collective-ops.md

rfcs/20260901-numpy-comparison-type.md
mkuperst: I kinda disgree with Peter here. :) 
I think "WEAKORDER" is somewhat confusing because of the NaN behavior - e.g. std::weak_order C++ has the following NaN rules, which are different from the numpy rules:
```
If x is negative NaN and y is not negative NaN, then weak_order_less(x, y) == true;
If x is not positive NaN and y is positive NaN, then weak_order_less(x, y) == true;
```
So naming this sort after what it's supposed to be compatible with makes more sense to me.

mkuperst (approved):

rfcs/20260901-numpy-comparison-type.md
GleasonK: My only objection to numpy is that I don't think these are spec'ed numpy semantics? I couldn't locate on quick scan at least, only that numpy uses ieee-757 comparisons. If they are specced then I'm more on board with the naming

mkuperst: The signed zero behavior is derived from ieee-754 comparisons.

The nan behavior is specced in https://numpy.org/devdocs/reference/generated/numpy.sort.html:
```
Previous to numpy 1.4.0 sorting real and complex arrays containing nan values led to undefined behaviour. In numpy versions >= 1.4.0 nan values are sorted to the end. 
```
And
```
For numerical sorts, NaT and NaN always sort to the end of the array for both ascending and descending sort order.
```

mkuperst: (Note that this implies that a numpy order descending sort can not be achieved by reversing an ascending sort. jax.lax.sort doesn't have this problem because it doesn't support descending sorts, but jax.numpy.sort does. Not sure if this change will affect the jax.numpy.sort descending sort implementation. @hawkinsp ?)

rfcs/20260901-numpy-comparison-type.md
akuegel: Thanks, done.

rfcs/20260901-numpy-comparison-type.md
akuegel: > (Note that this implies that a numpy order descending sort can not be achieved by reversing an ascending sort. jax.lax.sort doesn't have this problem because it doesn't support descending sorts, but jax.numpy.sort does. Not sure if this change will affect the jax.numpy.sort descending sort implementation. @hawkinsp ?)

Maybe the wording in https://numpy.org/devdocs/reference/generated/numpy.sort.html is sort of unclear. I can imagine that they mean that NaN values are considered the biggest values, which would put them at the front of an array sorted in descending order. In this proposal, I would like to preserve what JAX currently gets, which is that for descending order NaNs are sorted to the front. If that is different than what Numpy actually does, then I guess that would be another reason not to choose NUMPY for the comparison type.

So how to resolve the naming choice? Should we make a vote?

akuegel: @hawkinsp can you please take another look at the latest discussion here?

akuegel: Now that @mkuperst did the adjustment to document that JAX deviates from Numpy behavior when doing a descending sort, I think we can resolve this conversation, as using NUMPY would not make sense then.
With the changes suggested in this RFC, we do plan to keep the behavior in JAX the same. So for descending sort order, NaN values will be at the beginning.

akuegel:
@abhigunj can you please take a look at the latest adjustments?

abhigunj:
> @abhigunj can you please take a look at the latest adjustments?

Spec change and overall LGTM. I see we have indirect approvals via comments from the reviewers. Let's wait until early next week to get gh approval stamp from all and merge the RFC.

GleasonK (approved):

pifon2a (approved):

hawkinsp (approved):

